PR #12654, opened 2020-09-15
Bug fix/enable large follower trx

jsteemann:
### Scope & Purpose

With this PR , no followers should be dropped in this situation.

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required

#### Related Information

- [x] Design document: https://github.com/arangodb/documents/blob/master/DesignDocuments/02_PLANNING/IntermediateCommitsForFollowers.md

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11838/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11854/

arangod/Transaction/Manager.cpp
```diff
@@ -312,16 +316,59 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TransactionId tid,
...
+    // network transport etc.
+    // note that this increase may overflow the value beyond what a uint64_t can hold
+    decltype(options.maxTransactionSize) adjustedSize = double(options.maxTransactionSize) * 1.1;
+    if (adjustedSize > options.maxTransactionSize) {
+      // now the transaction on the follower should be able to grow to at least the
+      // size of the transaction on the leader.
+      options.maxTransactionSize = adjustedSize;
+    }
+    // it is also important that we set this option, so that it is ok for two 
+    // different leaders to try to "their" shards to the same follower transaction.
```
jsteemann, line 354:
```suggestion
    // different leaders to add "their" shards to the same follower transaction.
```

arangod/Transaction/Manager.cpp
```diff
@@ -312,16 +316,59 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TransactionId tid,
     auto& buck = _transactions[bucket];
     auto it = buck._managed.find(tid);
     if (it != buck._managed.end()) {
+      if (tid.isFollowerTransactionId()) {
+        // it is ok for two different leaders to try to create the same
+        // follower transaction.
```
jsteemann, line 321:
```suggestion
        // follower transaction on a leader.
```

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11856/

arangod/Transaction/Manager.cpp
```diff
@@ -645,6 +697,14 @@ Result Manager::updateTransaction(TransactionId tid, transaction::Status status,
       // make sure everyone who asks gets the updated timestamp
       mtrx.updateExpiry();
       if (mtrx.finalStatus == status) {
+        if (tid.isFollowerTransactionId() &&
+            mtrx.finalStatus == transaction::Status::ABORTED && 
+            mtrx.intermediateCommits) {
+          // we are trying to abort a follower transaction (again) that had intermediate
+          // commits already. in this case we return a special error code, which makes
+          // the leader drop us as a follower for all shards in the transaction.
+          return res.reset(TRI_ERROR_CLUSTER_FOLLOWER_TRANSACTION_COMMIT_PERFORMED);
```
graetzer, line 706: the follower transaction is alreadx aborted though, not sure what this is trying to accomplish

arangod/Transaction/Manager.cpp
```diff
@@ -664,6 +724,11 @@ Result Manager::updateTransaction(TransactionId tid, transaction::Status status,
     mtrx.type = MetaType::Tombstone;
     mtrx.updateExpiry();
     mtrx.finalStatus = status;
+    if (state->numCommits() > 0) {
+      // note that we have performed a commit or an intermediate commit.
+      // this is necessary for follower transactions
+      mtrx.intermediateCommits = true;
+    }
```
graetzer, line 731: isn't the idea that you cann abort follower transactions with intermediate commits

arangod/Transaction/Manager.cpp
graetzer: this check is too late and is only there when a second leader tries to abort

arangod/Transaction/Manager.cpp
```diff
@@ -703,9 +769,15 @@ Result Manager::updateTransaction(TransactionId tid, transaction::Status status,
     }
   } else {
     res = trx.abort();
-    if (res.ok() && wasExpired) {
+    if (intermediateCommits &&
+        tid.isFollowerTransactionId()) {
+      // we are trying to abort a follower transaction (again) that had intermediate
```
graetzer, line 774:
```suggestion
      // we are trying to abort a follower transaction that had intermediate
```

graetzer (approved):

lib/Rest/GeneralResponse.cpp
```diff
@@ -419,6 +419,7 @@ rest::ResponseCode GeneralResponse::responseCode(int code) {
     case TRI_ERROR_USER_DUPLICATE:
     case TRI_ERROR_TASK_DUPLICATE_ID:
     case TRI_ERROR_GRAPH_DUPLICATE:
+    case TRI_ERROR_CLUSTER_FOLLOWER_TRANSACTION_COMMIT_PERFORMED:
```
mchacki, line 422: I would have classified this as SERVER_ERROR not CONFLICT

We do not conflict with any of the existing data in my opinion.
This server is now in a situation that it cannot solve by itself and cannot be solved by "trying again after update" we need to somehow modify the servers environment now.

mchacki: It should only be seen internally though

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+// / @author Jan Steemann
+// //////////////////////////////////////////////////////////////////////////////
+
+let jsunity = require('jsunity');
+let internal = require('internal');
+let arangodb = require('@arangodb');
+let db = arangodb.db;
+const request = require('@arangodb/request');
+
+function getEndpointById(id) {
```
mchacki, line 35: On a side note: We should clean up those methods and pull them in a util class.

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+  };
+
+  const instanceInfo = JSON.parse(internal.env.INSTANCEINFO);
+  return instanceInfo.arangods.filter(isType)
+                              .map(toEndpoint)
+                              .map(endpointToURL);
+}
+  
+/// @brief set failure point
+function debugCanUseFailAt(endpoint) {
```
mchacki, line 75: These methods shall also be a util-library in client tests.
But can be a future cleanup task

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+      // disable intermediate commits on leader
+      debugSetFailAt(leader, "noIntermediateCommits");
+      // turn on intermediate commits on follower
+      debugClearFailAt(follower, "noIntermediateCommits");
+      
+      let droppedFollowersBefore = getMetric(leader, "arangodb_dropped_followers_count");
+      let intermediateCommitsBefore = getMetric(follower, "arangodb_intermediate_commits");
+      try {
+        db._query('FOR i IN 1..10000 INSERT { _key: CONCAT("test", i), value: ASSERT(i < 10000, "peng!") } IN ' + cn, {}, {intermediateCommitCount: 1000});
+        fail();
```
mchacki, line 207: This pattern does not work!
`fail()` will throw an error if triggered. You need this error to be passed through.
But it will be watched in the catch all. This way you miss if the query fails or not.

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+      } catch (err) {
+      }
```
mchacki, line 209:
```suggestion
      } catch (err) {
        if (err === "fail") {
          throw err;
        }
      }
```

This is a bit ugly, but will work. Would be better to assert the expected error to be thrown

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+              }
+            }
+            throw "peng!";
+          },
+          params: { cn },
+          intermediateCommitCount: 1000,
+        });
+        fail();
+      } catch (err) {
+      }
```
mchacki, line 300:
```suggestion
      } catch (err) {
        if (err === "fail") {
          throw err;
        }
      }
```

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+      getEndpointsByType("dbserver").forEach((ep) => debugClearFailAt(ep));
+      db._drop(cn + "1");
+      db._drop(cn + "2");
+    },
+    
+    // make follower execute intermediate commits (before the leader), but let the
+    // transaction succeed
+    testMultiAqlIntermediateCommitsOnFollower: function () {
+      db._create(cn + "1", { numberOfShards: 1, replicationFactor: 2 });
+      let shards1 = db._collection(cn + "1").shards(true);
```
mchacki, line 378: I think this creation of collections stunt should be a helper function.

Also I think you can drastically increase the chances by setting the `avoidServers` option on the second collection.
This way you can avoid placing any shard on the leader of collection `cn1`. effectively granting a 50/50 chance.

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
```diff
@@ -0,0 +1,491 @@
...
+      // turn on intermediate commits on follower
+      debugClearFailAt(follower1, "noIntermediateCommits");
+
+      let droppedFollowersBefore1 = getMetric(leader1, "arangodb_dropped_followers_count");
+      let droppedFollowersBefore2 = getMetric(leader2, "arangodb_dropped_followers_count");
+      try {
+        db._query('FOR i IN 1..10000 INSERT { _key: CONCAT("test", i), value: ASSERT(i < 10000, "peng!") } IN ' + cn + '1 INSERT { _key: CONCAT("test", i), value: ASSERT(i < 10000, "peng!") } IN ' + cn + '2', {}, {intermediateCommitCount: 1000});
+        fail();
+      } catch (err) {
+      }
```
mchacki, line 468:
```suggestion
      } catch (err) {
        if (err === "fail") {
          throw err;
        }
      }
```

mchacki (changes requested):
In general very good and I like.
I have some comments in the code, also in the test, they may actually loose an error case.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11890/

tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js
graetzer: `avoidServers` will also avoid using the server for follower shards

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11913/

PR #12665, opened 2020-09-16
Remove idle connections

graetzer:
### Scope & Purpose

Make sure the connection pool removes idle connections before the remote closes them

- [x] :hankey: Bugfix 

#### Backports:

- [x] No backports required

#### Related Information

- [x] GitHub issue / Jira ticket number: BTS-204

### DIFF
--- a/3rdParty/fuerte/src/H1Connection.h
+++ b/3rdParty/fuerte/src/H1Connection.h
@@ -105,18 +105,20 @@ class H1Connection final : public fuerte::GeneralConnection<ST, RequestItem> {
 
   fuerte::Error translateError(asio_ns::error_code const& e,
                                fuerte::Error c) const {
+#ifdef _WIN32
     if (this->_timeoutOnReadWrite && (c == Error::ReadError ||
                                       c == Error::WriteError)) {
       return Error::RequestTimeout;
     }
+#endif
     
     if (e == asio_ns::error::misc_errors::eof ||
         e == asio_ns::error::connection_reset) {
       return fuerte::Error::ConnectionClosed;
-    } else if (e == asio_ns::error::operation_aborted ||
-               e == asio_ns::error::connection_aborted) {
+    } else if (e == asio_ns::error::operation_aborted) {
       // keepalive timeout may have expired
-      return fuerte::Error::ConnectionCanceled;
+      return _timeoutOnReadWrite ? fuerte::Error::Timeout :
+                                   fuerte::Error::Canceled;
     }
     return c;
   }
--- a/3rdParty/fuerte/src/H1Connection.h
+++ b/3rdParty/fuerte/src/H1Connection.h
@@ -117,8 +117,8 @@ class H1Connection final : public fuerte::GeneralConnection<ST, RequestItem> {
       return fuerte::Error::ConnectionClosed;
     } else if (e == asio_ns::error::operation_aborted) {
       // keepalive timeout may have expired
-      return _timeoutOnReadWrite ? fuerte::Error::RequestTimeout :
-                                   fuerte::Error::ConnectionCanceled;
+      return this->_timeoutOnReadWrite ? fuerte::Error::RequestTimeout :
+                                         fuerte::Error::ConnectionCanceled;
     }
     return c;
   }
--- a/3rdParty/fuerte/src/H1Connection.h
+++ b/3rdParty/fuerte/src/H1Connection.h
@@ -117,8 +117,8 @@ class H1Connection final : public fuerte::GeneralConnection<ST, RequestItem> {
       return fuerte::Error::ConnectionClosed;
     } else if (e == asio_ns::error::operation_aborted) {
       // keepalive timeout may have expired
-      return _timeoutOnReadWrite ? fuerte::Error::Timeout :
-                                   fuerte::Error::Canceled;
+      return _timeoutOnReadWrite ? fuerte::Error::RequestTimeout :
+                                   fuerte::Error::ConnectionCanceled;
     }
     return c;
   }
--- a/arangod/Network/ConnectionPool.cpp
+++ b/arangod/Network/ConnectionPool.cpp
@@ -41,7 +41,6 @@ ConnectionPool::ConnectionPool(ConnectionPool::Config const& config)
     : _config(config), 
       _loop(config.numIOThreads, config.name) {
   TRI_ASSERT(config.numIOThreads > 0);
-  TRI_ASSERT(_config.minOpenConnections <= _config.maxOpenConnections);
 }
 
 ConnectionPool::~ConnectionPool() { shutdownConnections(); }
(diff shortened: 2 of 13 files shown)


arangod/Network/ConnectionPool.cpp
```diff
@@ -110,18 +108,10 @@ void ConnectionPool::pruneConnections() {
...
-          (*it)->lastLeased = now;
-          // connection will be kept
-        } else if (aliveCount >= _config.minOpenConnections &&
-                   (now - (*it)->lastLeased) > ttl) {
-          // connection hasn't been used for a while, remove it
-          // if we still have enough others
-          remove = true;
-        } else if (aliveCount >= _config.maxOpenConnections) {
-          // remove superfluous connections
+      } else if ((*it)->leases.load() == 0 && (*it)->fuerte->requestsLeft() == 0) {
```
jsteemann, line 111: The old code did bump `lastLeased` in this loop for all connections that were still in use.
The new code doesn't do this anymore. Is this going to cause any problems?

arangod/Network/ConnectionPool.cpp
```diff
@@ -199,11 +189,14 @@ std::shared_ptr<fuerte::Connection> ConnectionPool::createConnection(fuerte::Con
...
+  const std::chrono::milliseconds ttl(_config.idleConnectionMilli);
+  auto now = std::chrono::steady_clock::now();
+  
   std::lock_guard<std::mutex> guard(bucket.mutex);
 
   for (std::shared_ptr<Context>& c : bucket.list) {
-    const fuerte::Connection::State state = c->fuerte->state();
-    if (state == fuerte::Connection::State::Closed) {
+    if (c->fuerte->state() == fuerte::Connection::State::Closed ||
+        (now - c->lastLeased) > ttl) {
```
jsteemann, line 199: I understand that we cannot select a connection that was already closed. But why can't we select a connection anymore that has reached its expiration time? 
It should be ok to select an expired connection here and use it if the conection is not yet closed.
Pruning expired connections and selecting connections both acquire the bucket's mutex, so the two operations should never run at the same time on the same connection.

arangod/Network/ConnectionPool.cpp
graetzer: no

arangod/Network/ConnectionPool.cpp
graetzer: no expired connections are _expired_

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/11863/

arangod/Network/ConnectionPool.cpp
jsteemann: Fine, but are expired connections non-functional anymore?
My take is that an expired connection could still be reused/selected if otherwise ok. So we can save creating/opening a new connection object in the best case.
I am happy if there is a reason why we can't reuse expired connections, it's just that I currently don't understand it completely. Can you clarify?

graetzer: Read the title:
As I said in the team meeting: "the server side closes unused connections after the specified keep alive timeout. If the client doesn't do the same you get an error when trying to use the connection"

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/11868/

jsteemann (approved):
Seems to fix both BTS-204 and all my local tests with 3 DB servers.

arangod/Network/ConnectionPool.cpp
```diff
@@ -89,11 +87,11 @@ void ConnectionPool::shutdownConnections() {
 
 /// remove unused and broken connections
 void ConnectionPool::pruneConnections() {
-  const auto ttl = std::chrono::milliseconds(_config.idleConnectionMilli + 10);
+  const std::chrono::milliseconds ttl(_config.idleConnectionMilli);
+  auto now = std::chrono::steady_clock::now();
```
mchacki, line 91: I think this now needs to be gathered under the read-lock.
Otherwise we collect a now at `t0`, then wait for someone holding the write lock.
This someone will add a lastLeased := t1 with t1 > t0. 
And will afterwards return the lock.
This way this lease will trigger a number underflow when comparing against ttl, and will be flagged as removed

mchacki (changes requested):
I can confirm the bug in the ticket does not show up any more.
However i have some inline Comments, including two number underflows, which I think are serious

In addition I have an additional question:

Which is unclear to me:
In case I have a request, and I can reuse the connection.
Is there a possibility that some error on the existing request (or usage of any kind) on the connection could trigger a connection canceled, after this connection was handed out for lease, but before the leaser could get a chance to post something?

Everything in the pool seems to be locked, but as soon as we leave the pool this get's a bit hard to follow.

From this PR I am rather sure everything within the Pool itself is save.

arangod/Network/Methods.cpp
```diff
@@ -267,13 +272,13 @@ class RequestsState final : public std::enable_shared_from_this<RequestsState> {
     if (ADB_UNLIKELY(!_pool)) {
       LOG_TOPIC("5949f", ERR, Logger::COMMUNICATION)
           << "connection pool unavailable";
-      callResponse(Error::ConnectionCanceled, nullptr, std::move(_request));
+      callResponse(Error::ConnectionCanceled, nullptr, nullptr);
```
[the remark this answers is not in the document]
graetzer: `_request` is not populated until later in the call. This is actually a design failure, which we can probably fix in a later PR

arangod/Network/ConnectionPool.cpp
```diff
@@ -199,11 +189,14 @@ std::shared_ptr<fuerte::Connection> ConnectionPool::createConnection(fuerte::Con
 
 ConnectionPtr ConnectionPool::selectConnection(std::string const& endpoint,
                                                ConnectionPool::Bucket& bucket) {
+  const std::chrono::milliseconds ttl(_config.idleConnectionMilli);
+  auto now = std::chrono::steady_clock::now();
```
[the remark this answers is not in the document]
graetzer: sure but this is not really an issue in practice there are just more connections then

mchacki (approved):
I am waiting on a response to my Windows related question.

Otherwise I am Happy with this PR.

3rdParty/fuerte/src/H1Connection.h
```diff
@@ -105,18 +105,20 @@ class H1Connection final : public fuerte::GeneralConnection<ST, RequestItem> {
 
   fuerte::Error translateError(asio_ns::error_code const& e,
                                fuerte::Error c) const {
+#ifdef _WIN32
```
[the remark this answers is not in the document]
graetzer: Windows error codes seem to be different, this is why we need the special case

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11885/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11889/

PR #12673, opened 2020-09-17
Intel aes NI support

graetzer:
### Scope & Purpose

Test code for AES NI accelerated AES-256-CTR encryption in the enterprise version.

- [x] :pizza: New feature 

### Testing & Verification

http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/11891/


graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/11896/

arangod/CMakeLists.txt
```diff
@@ -833,6 +833,10 @@ add_library(arango_rocksdb STATIC
   ${ADDITIONAL_LIB_ARANGO_ROCKSDB_SOURCES}
 )
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  target_compile_options(arango_rocksdb PUBLIC "-maes")
```
jsteemann, line 837: This will allow usage of AES instructions for all code in this library. I doubt the compiler will emit AES instructions for other code parts than the actual encryption, but it is allowed to do so.

I think the minimum required target platform for running our packages is Westmere, which should include all the AES instructions if I am not wrong. So this would be safe. But it would be good if someone else could confirm this!

jsteemann (approved):

arangod/CMakeLists.txt
graetzer: so I was not able to just add the compile option on the one relevant file. I think the compiler will never output these instructions accidentally, so it should be ok

goedderz: I'd guess the same, though I didn't find any explicit documentation. I'm fine with this. But if you want, you can add another library `arango_rocksdb_aes` or so, and add the option only to that target.

jsteemann: @goedderz: sounds like a good idea to me!

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/11907/   (blue)

danielhlarkin (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12325/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12328/

PR #12679, opened 2020-09-19
Feature/use fixed size allocator for astnodes

jsteemann:
### Scope & Purpose

Use a slab allocator for AstNodes.
This PR retargets the existing but unused FixedSizeAllocator.

- [ ] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [x] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

Additionally:

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11902/

### DIFF
--- a/lib/Basics/FixedSizeAllocator.h
+++ b/lib/Basics/FixedSizeAllocator.h
@@ -43,31 +43,42 @@ class FixedSizeAllocator {
     MemoryBlock(size_t numItems)
         : _numAllocated(0), _numUsed(0), _alloc(nullptr), _data(nullptr) {
       TRI_ASSERT(numItems >= 32);
+      // assumption is that the size of a cache line is at least 64,
+      // so we are allocating 64 bytes in addition
       _alloc = new char[(sizeof(T) * numItems) + 64];
 
       _numAllocated = numItems;
 
-      // adjust to cache line offset (assumed to be 64 bytes)
-      _data = reinterpret_cast<char*>((reinterpret_cast<uintptr_t>(_alloc) + 63) &
-                                      ~((uintptr_t)0x3fu));
+      // adjust memory address to cache line offset (assumed to be 64 bytes)
+      _data = reinterpret_cast<char*>((reinterpret_cast<uintptr_t>(_alloc) + 63u) &
+                                      ~(uintptr_t(63u)));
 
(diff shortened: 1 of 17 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/11903/

lib/Basics/FixedSizeAllocator.h
```diff
@@ -104,63 +97,48 @@ class FixedSizeAllocator {
...
-    size_t const size = 128 << (std::min)(size_t(8), _blocks.size());
-    auto block = std::make_unique<MemoryBlock>(_itemSize, size);
-    _blocks.emplace_back(std::move(block));
+    // minimum block size is for 64 items
+    // maximum block size is for 4096 items
+    size_t const numItems = 64 << std::min<size_t>(6, _blocks.size() + 1);
+    _blocks.emplace_back(std::make_unique<MemoryBlock>(numItems));
   }
 
   std::vector<std::unique_ptr<MemoryBlock>> _blocks;
```
graetzer, line 141: not that you should change this now, but I wonder whether you could have used std::deque here because I think it sort of does the same thing

graetzer (approved):

lib/Basics/FixedSizeAllocator.h
mpoeter: I would directly link the blocks via next pointers (i.e., create a singly linked list) where head points to the latest block. And I would allocate a larger block memory to contain the `MemoryBlock` together with its items (so only a single allocation is required). I have already made a mental note to create an according PR. ;)

jsteemann: These are both good ideas, which will be most helpful for the first few MemoryBlocks.
The current implementation in this PR will do one allocation for the MemoryBlock, one for its data array, and another one for the _blocks vector.
Same for the second block. It will eventually go down to 2 allocations on average, because the vector will double its capacity every time.
But still very useful for massive ASTs!

jsteemann: @mpoeter : merging this now. Looking forward to seeing an improving PR from you! Thanks!

Issue #12683, opened 2020-09-22
SIGSEGV on installation 3.7.2-1


tarun-wadhwa-mmt:
## My Environment

* __ArangoDB Version__:        3.7.2-1
* __Storage Engine__:            RocksDB
* __Deployment Mode__:       Single Server, Cluster
* __Deployment Strategy__:   Manual Start , ArangoDB Starter
* __Infrastructure__:               own
* __Operating System__:        Centos 7
* __Total RAM in your machine__:        2 machines of 8GB 2 core configuration
* __Disks in use__:        SSD
* __Used Package__:              rpm

## Component, Query & Data

__Affected feature__:

Installation. Basically setting up DBserver

## Steps to reproduce

1. Ran this command to setup DBServer after running agency - 
`arangod --server.authentication=false --server.endpoint tcp://0.0.0.0:8530 --cluster.my-address tcp://m_dvp_arangodb_1.mmt.mmt:8530 --cluster.my-role COORDINATOR --cluster.agency-endpoint tcp://m_dvp_arangodb_1.mmt.mmt:8531`

__Problem__: We have a 2 node cluster wherein we tried setting up arango db in cluster mode with agency size = 1
I am able to run agency service with the command mentioned [here](https://www.arangodb.com/docs/stable/deployment-cluster-manual-start.html) . While trying to setup DBServer, i get SIGSEGV fault

`FATAL [a7902] {crash} ArangoDB 3.7.2-1 [linux], thread 0 caught unexpected signal 11 (SIGSEGV) accessing address 0x00000000000000b8: signal handler invoked`

Full stack trace - 
```
2020-09-22T09:18:05Z [17098] INFO [e52b0] ArangoDB 3.7.2-1 [linux] 64bit, using jemalloc, build tags/v3.7.2.1-0-g2c56dec4b2c-dirty, VPack 0.1.33, RocksDB 6.8.0, ICU 64.2, V8 7.9.317, OpenSSL 1.1.1g  21 Apr 2020
2020-09-22T09:18:05Z [17098] INFO [75ddc] detected operating system: Linux version 3.10.0-1062.12.1.el7.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC) ) #1 SMP Tue Feb 4 23:02:59 UTC 2020
[...]

Issue #12597, opened 2020-09-03
Upgrade from 3.6.6 and new install with restored data of 3.6.6 fails on Ubuntu and MacOs


goldsky:
## My Environment

* __ArangoDB Version__:        3.7.2
* __Storage Engine__:             RocksDB
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:   Manual Start
* __Infrastructure__:               own
* __Operating System__:        Ubuntu 20.04 & MacOS 10.13.4
* __Total RAM in your machine__:        4GB & 8GB
* __Disks in use__:        HHD & SSD
* __Used Package__:              .deb & .dmg

## Component, Query & Data

__Affected feature__:
Installation

__Size of your Dataset on disk__:
Under 2MB

## Steps to reproduce

1. Ubuntu: brand new installation, arangorestore; Macos: 3.6.6, replace Application file
2. Start the database
3. On application, connect to the database

__Problem__:

When the application opened the connection to database, the database suddenly shut down.

**Ubuntu core dump:**
```
Core was generated by `/usr/sbin/arangod --uid arangodb --gid arangodb --pid-file /var/run/arangodb3/a'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fdb012e5abc in ?? ()
[Current thread is 1 (LWP 30881)]
(gdb) <RET>
Undefined command: "<RET>".  Try "help".
(gdb) backtrace full
#0  0x00007fdb012e5abc in ?? ()
No symbol table info available.
#1  0x00007fdb012e3843 in ?? ()
No symbol table info available.
#2  0x00007fdb02d02cc4 in ?? ()
No symbol table info available.
#3  0x0000000000000000 in ?? ()
No symbol table info available.
(gdb) thread apply all bt

Thread 27 (LWP 31017):
#0  0x00007fdb012d48e9 in ?? ()
#1  0x0000000000000000 in ?? ()

Thread 26 (LWP 30888):
#0  0x00007fdb012e5abc in ?? ()
#1  0x00007fdb012e3843 in ?? ()
#2  0x00007fdaf71f0b64 in ?? ()
#3  0x0000000000000000 in ?? ()

Thread 25 (LWP 31015):
#0  0x00007fdb012e5abc in ?? ()
#1  0x00007fdb012e3843 in ?? ()
#2  0x00007fdade2ebb64 in ?? ()
#3  0x0000000000000000 in ?? ()

Thread 24 (LWP 30891):
[...]

PR #12686, opened 2020-09-22
Output CPU context and image base in signal handler (if available).

mpoeter:
### Scope & Purpose

Some error reports like #12683 or #12597 would greatly benefit from additional information. In particular for SIGSEGV crashes the full CPU context and image base can provide a lot of information, even if no debug symbols are available.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

Though it might make sense to have this in older versions as well (up to you).

After loading the according executable into [IDA Pro Free](https://www.hex-rays.com/products/ida/support/download_freeware/) and rebasing the image to the according base address from the process, we can check the instruction pointer (RIP) and the addresses from any callstacks. This is how I analyzed the issue in #12597.


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/11917/

jsteemann:
@mpoeter: Thanks for this PR!
I had started a similar change to show the executables base address, though I didn't have time to finish it yet.
It is not based on `dladdr`. but on `getauxval(AT_PHDR)`.
I am pretty sure both variants will work fine when compiling the executables with glibc.
However, our release packages are built statically and are linked to libmusl, and unfortunately it has a lot less to offer here.
So I will pull this PR now and check whatever it will do to a libmusl-based build.
This will take a bit of time, but I will make sure it will be pursued!

jsteemann:
@mpoeter : just built the branch into a release build, statically linked against libmusl, started the executable and sent it a SIGSEGV signal.
Here is the log output:
```
2020-09-23T11:38:38Z [90550] FATAL [a7902] {crash} ArangoDB 3.8.0-devel enterprise [linux], thread 0 caught unexpected signal 11 (SIGSEGV) accessing address 0x000003e800016259: signal handler invoked\nCPU context:\n  rip: 0x00007f057e34136b rsp: 0x00007ffe9d9bb188\n  efl: 0x0000000000000246 rbp: 0x0000000000000080\n  rsi: 0x0000000000000080 rdi: 0x00007ffe9d9bb264\n  rax: 0xfffffffffffffffc rbx: 0x00007f057ac88c80\n  rcx: 0x00007f057e34136b rdx: 0x0000000000000002\n  r8:  0x0000000000000000 r9:  0x0000000000000000\n  r10: 0x00007ffe9d9bb1e0 r11: 0x0000000000000246\n  r12: 0x00000000000000ca r13: 0x00007ffe9d9bb264\n  r14: 0x0000000000000080 r15: 0x00007ffe9d9bbb70\nimage base address: 0x0000000000000000
2020-09-23T11:38:38Z [90550] INFO [308c3] {crash} frame 1 [0x00007f057c94ac9f] (anonymous namespace)::crashHandlerSignalHandler(int, siginfo_t*, void*) (+0x000000000000003f)
2020-09-23T11:38:38Z [90550] INFO [308c3] {crash} frame 2 [0x00007f057e338e7c] sigprocmask (+0x0000000000000021)
2020-09-23T11:38:38Z [90550] INFO [ded81] {crash} available physical memory: 41722023936, rss usage: 304959488, vsz usage: 1206083584, threads: 46
```
While I have no reason to question the register output, the image base address is reported as `0x0`. I think this is expected when statically linking with libmusl.

mpoeter:
@jsteemann I am more than happy if you have an alternative to `dladdr`! On Windows I would simply use `GetModuleHandle(NULL)` to get the base address of the executable, but for Linux `dladdr` was basically the only thing I could find after a lot of googling (except for parsing the content from `/proc/self/maps`).

mpoeter:
@jsteemann I like the `getauxval(AT_PHDR)` approach much more (not only because it should work independent of glibc). One minor gotcha was that it does not return the base address directly, but the the address of the program header, which is located after the ELF header. So to get the actual base address we have to subtract the size of the ELF header. I have changed the code accordingly and hope this works correctly with libmusl now.

jsteemann:
@mpoeter : Thanks! I think I will abandon my own changes then in favor of your PR.
I will also use your PR as a basis for incorporating a few minimal changes of mine that may be still relevant.
Just to manage expectations: I think I can do this and also test it with libmusl only next week, so please expect a slight delay here.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12102/

jsteemann (approved):
LGTM!

jsteemann:
Tests blue

jsteemann:
Backport for 3.7: https://github.com/arangodb/arangodb/pull/12767

PR #12687, opened 2020-09-22
Lazily allocate revision trees

danielhlarkin:
### Scope & Purpose

Instead of automatically allocating a fixed revision tree for an empty collection, only allocate the tree when data is added to the collection. This should save a few MB of RAM for each empty collection, which can greatly benefit instances with lots of databases (e.g. multi-tenant applications).

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

Additionally:

Jenkins:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/11921/ (all blue except Windows, which failed to start)
https://jenkins.arangodb.biz/job/arangodb-matrix-pr-windows/5629/ (all blue)


arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -478,10 +459,16 @@ rocksdb::SequenceNumber RocksDBMetaCollection::serializeRevisionTree(
 
 Result RocksDBMetaCollection::rebuildRevisionTree() {
   std::unique_lock<std::mutex> guard(_revisionTreeLock);
-  _revisionTree = std::make_unique<containers::RevisionTree>(
-      6, _logicalCollection.minRevision().id());
-
-  Result res = basics::catchToResult([this]() -> Result {
+  struct {
+    std::unique_ptr<containers::RevisionTree> tree;
+    std::atomic<rocksdb::SequenceNumber> applied;
```
jsteemann, line 464: As far as I can see, the `applied` member variable is only updated inside this function for a function-local variable. So there should be no other threads modifying it at the same time. Just wondering why the variable is declared atomic here. It would probably be ok to declare it non-atomic, i.e.
```suggestion
    rocksdb::SequenceNumber applied;
```
and then of course adjust the 2 or 3 places that store or load its value.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -884,3 +881,52 @@ int RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
...
+    }
+    _revisionTreeCreationSeq = rocksutils::globalRocksDB()->GetLatestSequenceNumber();
+    _revisionTreeSerializedSeq = _revisionTreeCreationSeq;
+    _revisionTree = allocateEmptyRevisionTree();
+    return true;
+  } catch (...) {
+    _revisionTree.reset();  // should be unnecessary, but just in case
+    return false;
+  }
+}
```
jsteemann, line 932: Looks like this method's return value is never checked, so we may simplify it to just:
```suggestion
void RocksDBMetaCollection::ensureRevisionTree() {
  if (!_revisionTree) {
    // if the allocation throws, _revisionTree` will still be a nullptr.
    _revisionTree = allocateEmptyRevisionTree();
    _revisionTreeCreationSeq = rocksutils::globalRocksDB()->GetLatestSequenceNumber();
    _revisionTreeSerializedSeq = _revisionTreeCreationSeq;
  }
}
```

arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -470,14 +470,26 @@ Result RocksDBMetadata::deserializeMeta(rocksdb::DB* db, LogicalCollection& coll
...
-      if (res.fail()) {
-        LOG_TOPIC("ecdbd", WARN, Logger::ENGINES)
-            << "failed to rebuild revision tree for collection '"
-            << coll.id().id() << "'";
+      // no tree, check if collection is non-empty
+      auto bounds = RocksDBKeyBounds::CollectionDocuments(rcoll->objectId());
+      auto cmp = RocksDBColumnFamily::documents()->GetComparator();
+      std::unique_ptr<rocksdb::Iterator> it{
+          db->NewIterator(ro, RocksDBColumnFamily::documents())};
+      it->Seek(bounds.start());
```
jsteemann, line 478: This will add an extra `Seek` for each collection/shard at server startup, which could be expensive for very large datasets. Is there a way to avoid this, or at least use the just-created iterator inside `rebuildRevisionTree()`?

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -1307,6 +1307,10 @@ Result DatabaseInitialSyncer::fetchCollectionSyncByRevisions(arangodb::LogicalCo
   //std::pair<std::size_t, std::size_t> fullRange = treeLeader->range();
   std::unique_ptr<containers::RevisionTree> treeLocal =
       physical->revisionTree(*trx);
+  if (!treeLocal) {
+    // local collection does not support syncing by revision, fall back to keys
+    return fetchCollectionSyncByKeys(coll, leaderColl, maxTick);
```
jsteemann, line 1312: This will make `fetchCollectionSyncByKeys()` run inside the transaction started in line 1293.
Do we need to abort this transaction via `guard.fire();` before we branch into `fetchCollectionSyncByKeys()` here?

jsteemann (changes requested):

arangod/Replication/DatabaseInitialSyncer.cpp
danielhlarkin: Good catch.

danielhlarkin:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/11944/

jsteemann (approved):

PR #12294, opened 2020-07-24
Create leader shards on agency cache events directly.

neunhoef:
This is a first try to use event-based methods to react to
Plan changes more quickly. So far, this is a proof of concept.
We may or may not adopt this approach in the future.

PR #12709, opened 2020-09-24
[3.7] back porting database subscription

kvahed:
### Scope & Purpose

- [ ] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required

#### Related Information

*(Please reference tickets / specification etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Additionally:

### Documentation

> All new Features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the changelog so that
> developers and users have a concise overview. 

- [ ] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12007/

KVS85:
Should be the part of https://github.com/arangodb/arangodb/pull/12294.

KVS85:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12305/

KVS85:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12305/

CHANGELOG
```diff
@@ -83,6 +94,9 @@ v3.7.3 (XXXX-XX-XX)
   - `arangodb_http_request_statistics_user_requests`: Total number of HTTP
     requests executed by clients
 
+* Make scheduler react and start new threads slightly faster in case a lot
+  of new work arrives.
+
```
jsteemann, line 99: I don't think this belongs here.

arangod/Agency/Constituent.cpp
```diff
@@ -746,19 +744,41 @@ void Constituent::run() {
...
+              << followerId << " needed: " << needed;
             if (needed) {
+              auto startTime = std::chrono::steady_clock::now();
               _agent->sendEmptyAppendEntriesRPC(followerId);
+              auto endTime = std::chrono::steady_clock::now();
+              if (endTime - startTime > std::chrono::milliseconds(100)) {
+                LOG_TOPIC("ddeeb", WARN, Logger::AGENCY)
+                  << "Call to sendEmptyAppendEntriesRPC took longer than 0.1 s: time needed: "
+                  << std::chrono::duration<double>(endTime - startTime).count();
+              }
```
jsteemann, line 757: Isn't this somewhat expected that this can take a while?
At least `sendEmptyAppendEntriesRPC()` does network communication, which can take an arbitrary amount of time.
I am concerned this just adds more log spam in case things are already slow anyway. 
If we want to keep this log message, I would prefer setting the log level to INFO instead of WARN.

arangod/Agency/Node.cpp
```diff
@@ -1196,6 +1196,16 @@ void Node::clear() {
   }
 }
 
+std::vector<std::string> Node::keys() const {
+  std::vector<std::string> result;
+  if (!_isArray) {
+    for (auto const& i : _children) {
```
jsteemann, line 1202: potential performance optimization:
```suggestion
    result.reserve(_children.size());
    for (auto const& i : _children) {
```

arangod/Agency/State.cpp
```diff
@@ -347,6 +347,7 @@ void State::logEmplaceBackNoLock(log_t&& l) {
   }
 
   try {
+    _log_size += l.entry->byteSize();
```
jsteemann, line 350: Can this now count a log entry multiple times?
For example, the method directly before it does:
```
  logEmplaceBackNoLock(log_t(idx, term, buf, clientId, millis));
  _log_size += byteSize;
```
meaning that we will call the size of that log entry multiple times.

arangod/VocBase/Methods/Databases.cpp
```diff
@@ -186,11 +186,15 @@ Result Databases::createCoordinator(CreateDatabaseInfo const& info) {
     return Result(TRI_ERROR_ARANGO_DATABASE_NAME_INVALID);
   }
 
+  LOG_TOPIC("56372", DEBUG, Logger::CLUSTER) << "createDatabase on coordinator: Starting, name: " << info.getName();
```
jsteemann, line 189: I suggest that this log entry (and the following ones that are added in this PR) are downgraded to log level TRACE.

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -462,6 +462,8 @@ TRI_voc_rid_t LogicalCollection::revision(transaction::Methods* trx) const {
 }
 
 bool LogicalCollection::usesRevisionsAsDocumentIds() const {
+  // TODO: switch off for now to lower memory consumption:
+  // return false;
```
jsteemann, line 466: This comment should be removed.
```suggestion
```

arangod/Cluster/ClusterFeature.cpp
```diff
@@ -784,3 +784,61 @@ void ClusterFeature::allocateMembers() {
...
+    notify();
+  }
+}
+
+void ClusterFeature::addDirty(std::string const& database) {
+  MUTEX_LOCKER(guard, _dirtyLock);
+  if (_dirtyDatabases.emplace(database).second) {
+    LOG_TOPIC("357b9", DEBUG, Logger::MAINTENANCE) << "adding " << database << " to dirty databases";
+  }
+  notify();
```
jsteemann, line 821: shall we call notify here only if we actually added the database to `_dirtyDatabases`?

arangod/Cluster/ClusterFeature.cpp
```diff
@@ -784,3 +784,61 @@ void ClusterFeature::allocateMembers() {
...
+  if (databases.size() > 0) {
+    MUTEX_LOCKER(guard, _dirtyLock);
+    for (auto const& database : databases) {
+      if (_dirtyDatabases.emplace(database).second) {
+        LOG_TOPIC("35b75", DEBUG, Logger::MAINTENANCE)
+          << "adding " << database << " to dirty databases";
+      }
+    }
+    if (callNotify) {
+      notify();
```
jsteemann, line 798: does it make sense here to only call `notify()` if `callNotify` is set *and* we succeeded in adding a database to `_dirtyDatabases`?

arangod/Cluster/ClusterFeature.cpp
```diff
@@ -784,3 +784,61 @@ void ClusterFeature::allocateMembers() {
...
+void ClusterFeature::addDirty(std::unordered_map<std::string,std::shared_ptr<VPackBuilder>> const& databases) {
+  if (databases.size() > 0) {
+    MUTEX_LOCKER(guard, _dirtyLock);
+    for (auto const& database : databases) {
+      if (_dirtyDatabases.emplace(database.first).second) {
+        LOG_TOPIC("35b77", DEBUG, Logger::MAINTENANCE)
+          << "adding " << database << " to dirty databases";
+      }
+    }
+    notify();
```
jsteemann, line 812: does it make sense here to only call `notify()` in case we added at least one database to `_dirtyDatabases`?

arangod/Cluster/AgencyCache.cpp
```diff
@@ -151,18 +173,75 @@ void AgencyCache::handleCallbacksNoLock(VPackSlice slice, std::unordered_set<uin
...
         toCall.push_back(it->second);
       }
       ++it;
     }
+
+    if (k.size() > 8) {
+      std::string_view r(k.c_str() + 8, k.size() - 8);
+      auto rs = r.size();
+
+      if (rs > strlen(PLAN) && r.compare(0, strlen(PLAN), PLAN) == 0) {
```
gnusi, line 195: why don't to use `std::string_view` and avoid tons of `strlen` calls?

arangod/Cluster/AgencyCache.cpp
```diff
@@ -256,10 +337,19 @@ void AgencyCache::run() {
                         std::lock_guard g(_storeLock);
                         _readDB.applyTransaction(i); // apply logs
                         _commitIndex = i.get("index").getNumber<uint64_t>();
-                      }
-                      {
-                        std::lock_guard g(_callbacksLock);
-                        handleCallbacksNoLock(i.get("query"), uniq, toCall);
+
+                        std::unordered_set<std::string> pc, cc;
```
gnusi, line 341: what `pc` and `cc` mean? why can't we move these variables outside for loop above and just clear them at every iteration to reduce a number of memory allocations?

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
   }
   invokeCallbacks(toCall);
 }
+
+
+AgencyCache::change_set_t AgencyCache::changedSince(
+  std::string const& what, consensus::index_t const& last) const {
+
+  static std::vector<std::string> const planGoodies ({
+      AgencyCommHelper::path(PLAN_ANALYZERS) + "/",
```
gnusi, line 576: isn't `PLAN_ANALYZER` already contains trailing `/`?

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
   }
   invokeCallbacks(toCall);
 }
+
+
+AgencyCache::change_set_t AgencyCache::changedSince(
+  std::string const& what, consensus::index_t const& last) const {
+
+  static std::vector<std::string> const planGoodies ({
```
gnusi, line 575: would be nice to move local static variables outside function context to speedup access

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
...
+  std::shared_lock g(_storeLock);
+
+  auto tmp = _readDB.nodePtr()->hasAsUInt(AgencyCommHelper::path(what) + "/" + VERSION);
+  uint64_t version = tmp.second ? tmp.first : 0;
+
+  if (last < _lastSnapshot) {
+    get_rest = true;
+    auto keys = _readDB.nodePtr(AgencyCommHelper::path(what) + "/" + DATABASES)->keys();
+    for (auto& i : keys) {
+      databases.emplace(std::move(i));
```
gnusi, line 601: why not to reserve space for all keys before entering the loop?

gnusi:
In general I don't understand why in many places we allocate `std::shared_ptr<VPackBuilder>` for temporary variables

arangod/Cluster/Maintenance.cpp
```diff
@@ -554,65 +636,110 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
...
+      } else {
+        path.push_back(INDEXES);
+        VPackSlice indexes = planit->second->slice()[0].get(path);
+        TRI_ASSERT(indexes.isArray());
+        if (indexes.isArray()) {
+          for (auto& p : shard.second) {
+            std::string const& id = p.first;
+            bool found = false;
+            for (auto const& ind : VPackArrayIterator(indexes)) {
+              if (ind.get(ID).copyString() == id) {
```
gnusi, line 712: why `copyString()` instead of `stringView()`?

arangod/Cluster/Maintenance.cpp
```diff
@@ -476,76 +506,128 @@ VPackBuilder getShardMap(VPackSlice const& plan) {
...
+  for (auto const& dbname : dirty) {
+    if (local.find(dbname) != local.end()) {
+      bool needDrop = false;
+      auto it = plan.find(dbname);
+      if (it == plan.end()) {
+        needDrop = true;
+      } else {
+        VPackSlice pb = it->second->slice()[0];
+        VPackSlice pdb = pb.get(
+            std::vector<std::string>{AgencyCommHelper::path(), PLAN, DATABASES, dbname});
```
gnusi, line 561: why to copy `std::string`s, can't it just be a `std::vector<std::string_view>` or `std::vector<arangodb::StringRef>`?

gnusi, line 561: @jsteemann 
of course it'd be better for velocypack slice to accept just a range of iterators instead of vector

gnusi (changes requested):

arangod/Cluster/Maintenance.cpp
```diff
@@ -928,265 +1076,311 @@ static VPackBuilder assembleLocalDatabaseInfo(std::string const& database,
...
+    auto pit = plan.find(dbName);
+    VPackSlice pdb;
+    if (pit == plan.end()) {
+      LOG_TOPIC("47e23", TRACE, Logger::MAINTENANCE)
+        << dbName << " missing in plan";
+    } else {
+      TRI_ASSERT(pit->second->slice().isArray());
+      TRI_ASSERT(pit->second->slice().length() == 1);
+      pdb = pit->second->slice()[0];
+      std::vector<std::string> ppath{
```
gnusi, line 1118: again... tons of useless `std::string` copies + `std::vector` allocation

gnusi (changes requested):

gnusi:
also tons of `std::string` copies in a vector just to lookup a field in `VPackSlice` should be avoided

KVS85 (changes requested):
Reproducible failure in cluster:
```
* Test "shell_client_3"
    [FAILED]  tests/js/client/shell/shell-admin-status.js

      "testOverview" failed: Error: at assertion #1: assertTrue: (false) does not evaluate to true
(Error
    at assertTrue (/work/ArangoDB/js/common/modules/jsunity/jsunity.js:71:19)
    at Object.testOverview (tests/js/client/shell/shell-admin-status.js:68:7)
    at Object.run (/work/ArangoDB/js/common/modules/jsunity/jsunity.js:549:27)
    at Object.Run [as run] (/work/ArangoDB/js/common/modules/jsunity.js:276:24)
    at tests/js/client/shell/shell-admin-status.js:82:9
    at tests/js/client/shell/shell-admin-status.js:85:3
    at RunTest (/work/ArangoDB/js/common/modules/jsunity.js:390:12)
    at testFunc (eval at runInLocalArangosh (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:885:3), <anonymous>:4:9)
    at runInLocalArangosh (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:889:18)
    at Object.performTests (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:233:21))
    at Object.run (/work/ArangoDB/js/common/modules/jsunity/jsunity.js:569:23)
    at Object.Run [as run] (/work/ArangoDB/js/common/modules/jsunity.js:276:24)
    at tests/js/client/shell/shell-admin-status.js:82:9
    at tests/js/client/shell/shell-admin-status.js:85:3
    at RunTest (/work/ArangoDB/js/common/modules/jsunity.js:390:12)
    at testFunc (eval at runInLocalArangosh (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:885:3), <anonymous>:4:9)
    at runInLocalArangosh (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:889:18)
    at Object.performTests (/work/ArangoDB/js/client/modules/@arangodb/test-utils.js:233:21)
    at Object.shellClient [as shell_client] (/work/ArangoDB/js/client/modules/@arangodb/testsuites/aql.js:60:13)
    at iterateTests (/work/ArangoDB/js/client/modules/@arangodb/testing.js:506:36) - test failed

   Suites failed: 1 Tests Failed: 1
```

KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12309/

arangod/Cluster/DBServerAgencySync.cpp
```diff
@@ -157,14 +174,42 @@ DBServerAgencySyncResult DBServerAgencySync::execute() {
...
+  auto moreDirt = mfeature.pickRandomDirty(
+    static_cast<size_t>(std::ceil(static_cast<double>(mfeature.lastNumberOfDatabases()) / 720.)));
+  auto dirty = mfeature.dirty();
+  // Add `moreDirt` to `dirty` but remove from `moreDirt`, if already dirty anyway.
+  // Then we can reasonably be surprised if we find anything in a database in `moreDirt`.
+  for (auto it = moreDirt.begin(); it != moreDirt.end();) {
+    if (dirty.find(*it) != dirty.end()) {
+      dirty.insert(*it);
+      it++;
+    } else {
```
jsteemann, line 192: This part confuses me. We are looking up `*it` in `dirty`, and if it is already in there, we insert it into... `dirty`??

arangod/Cluster/DBServerAgencySync.cpp
```diff
@@ -157,14 +174,42 @@ DBServerAgencySyncResult DBServerAgencySync::execute() {
...
+
+  AgencyCache::databases_t plan = clusterInfo.getPlan(planIndex, dirty);
+  if (!dirty.empty() && plan.empty()) {
     // TODO increase log level, except during shutdown?
     LOG_TOPIC("0a6f2", DEBUG, Logger::MAINTENANCE)
-        << "DBServerAgencySync::execute no plan";
-    result.errorMessage = "DBServerAgencySync::execute no plan";
+        << "DBServerAgencySync::execute no plan change set";
+    result.success = true;
+    result.errorMessage = "DBServerAgencySync::execute no reconciliation plan";
```
jsteemann, line 212: We are returning an error message but also set `success` to `true` here.
This is a bit counter-intuitive, but the same happens a few lines above. So I am wondering if this is intentional or not.
If it is intentional, we should leave a code comment why encountering an error allows us to still set `success` to true here.

arangod/Cluster/DBServerAgencySync.cpp
```diff
@@ -157,14 +174,42 @@ DBServerAgencySyncResult DBServerAgencySync::execute() {
...
+      it = moreDirt.erase(it);
+    }
+  }
+
 
-  if (plan == nullptr) {
+  if (dirty.empty()) {
+    LOG_TOPIC("0a62f", DEBUG, Logger::MAINTENANCE)
+      << "DBServerAgencySync::execute no dirty collections";
+    result.success = true;
```
jsteemann, line 201: We are returning an error message but also set `success` to `true` here.
This is a bit counter-intuitive, but the same happens a few lines below. So I am wondering if this is intentional or not.
If it is intentional, we should leave a code comment why encountering an error allows us to still set `success` to true here.

arangod/Cluster/DBServerAgencySync.cpp
```diff
@@ -298,16 +346,15 @@ DBServerAgencySyncResult DBServerAgencySync::execute() {
...
+        VPackSlice planSlice = report.get("Plan");
+        VPackSlice currentSlice = report.get("Current");
         result = DBServerAgencySyncResult(
-            true,
-            report.hasKey("Plan") ? report.get("Plan").get("Version").getNumber<uint64_t>() : 0,
-            report.hasKey("Current")
-                ? report.get("Current").get("Version").getNumber<uint64_t>()
-                : 0);
+          true, planSlice.isObject() ? planSlice.get("Index").getNumber<uint64_t>() : 0,
+          currentSlice.isObject() ? currentSlice.get("Index").getNumber<uint64_t>() : 0);
```
jsteemann, line 353: The new code new looks up Plan/Index and Current/Index, whereas the old code checked Plan/Version and Current/Version.
This is probably intentional, just wanted to highlight this in case it is unintentional.

arangod/Agency/State.cpp
kvahed: This has not been altered. And I would hope that we could collect all general remarks into a separate PR, that would make its way into all versions.

arangod/VocBase/Methods/Databases.cpp
kvahed: Again, if you don't mind please collect theses as suggestions for a PR, that could be changed everywhere. We use the DEBUG on client support cases and TRACE holds way too much. This is good information in DEBUG level

CHANGELOG
kvahed: this came with the merge, but will remove, if you say so. cannot comment, if it belongs there

arangod/Cluster/AgencyCache.cpp
```diff
@@ -151,18 +173,75 @@ void AgencyCache::handleCallbacksNoLock(VPackSlice slice, std::unordered_set<uin
...
     auto it = _callbacks.lower_bound(k);
     while (it != _callbacks.end() && it->first.compare(0, k.size(), k) == 0) {
       if (uniq.emplace(it->second).second) {
         toCall.push_back(it->second);
       }
       ++it;
     }
+
+    if (k.size() > 8) {
+      std::string_view r(k.c_str() + 8, k.size() - 8);
```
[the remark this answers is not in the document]
kvahed: Sure thing.

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
...
+
+  if (get_rest) { // All the rest, i.e. All keys excluding the usual suspects
+    static std::vector<std::string> const exc {
+      "Analyzers", "Collections", "Databases", "Views"};
+    auto keys = _readDB.nodePtr(AgencyCommHelper::path(what))->keys();
+    keys.erase(
+      std::remove_if(
+        std::begin(keys), std::end(keys),
+        [&] (auto x) {
+          return std::binary_search(std::begin(exc), std::end(exc),x);}),
```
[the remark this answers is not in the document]
kvahed: It does not hurt either.

arangod/Cluster/ClusterFeature.cpp
kvahed: yes

arangod/Cluster/ClusterFeature.cpp
kvahed: yes

arangod/Cluster/ClusterFeature.cpp
kvahed: yes

arangod/Cluster/RestClusterHandler.cpp
```diff
@@ -124,11 +127,36 @@ void RestClusterHandler::handleAgencyCache() {
...
+  if (af->isActive() && !_request->user().empty()) {
+    auth::Level lvl;
+    if (af->userManager() != nullptr) {
+      lvl = af->userManager()->databaseAuthLevel(_request->user(), "_system", true);
+    } else {
+      lvl = auth::Level::RW;
+    }
+    if (lvl < auth::Level::RW) {
+      generateError(rest::ResponseCode::FORBIDDEN, TRI_ERROR_HTTP_FORBIDDEN,
+                    "you need admin rights to produce an agency dump");
```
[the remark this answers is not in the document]
kvahed: correct

arangod/Cluster/AgencyCache.cpp
kvahed: Plan changed and Current changed. Compiler are very efficient at optimising these allocations. I just tested this in a standalone program with google benchmark. No runtime difference in 10000 repetitions on either clang or gcc.

arangod/Cluster/AgencyCache.cpp
kvahed: correct, thank you

arangod/Cluster/AgencyCache.cpp
kvahed: also tested just now with google benchmark not measureable difference in runtime. in that case would rather not make the change because we lack the time to run complete chaos for over a week again.

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
...
+  }
+
+  if (databases.empty()) {
+    return change_set_t(_commitIndex, version, std::move(db_res), std::move(rest_res));
+  }
+
+  {
+    for (auto const& i : databases) {
+      if (!i.empty()) { // actual database
+        auto query = std::make_shared<arangodb::velocypack::Builder>();
```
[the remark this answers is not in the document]
kvahed: sure

arangod/Cluster/AgencyCache.cpp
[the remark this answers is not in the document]
kvahed: changes

arangod/Cluster/Maintenance.cpp
kvahed: not implemented in Slice

kvahed: same here

arangod/Cluster/DBServerAgencySync.cpp
kvahed: nice catch. many thx.

arangod/Cluster/DBServerAgencySync.cpp
kvahed: this is less of an error and more of a message,

arangod/Cluster/DBServerAgencySync.cpp
kvahed: again, not an error but a message

arangod/Cluster/DBServerAgencySync.cpp
kvahed: correct is intentional, as versions may remain unchanged in many cases

arangod/Cluster/AgencyCache.cpp
kvahed: sure. good point

arangod/Cluster/AgencyCache.cpp
```diff
@@ -469,3 +567,129 @@ void AgencyCache::invokeAllCallbacks() const {
...
+    static std::vector<std::string> const exc {
+      "Analyzers", "Collections", "Databases", "Views"};
+    auto keys = _readDB.nodePtr(AgencyCommHelper::path(what))->keys();
+    keys.erase(
+      std::remove_if(
+        std::begin(keys), std::end(keys),
+        [&] (auto x) {
+          return std::binary_search(std::begin(exc), std::end(exc),x);}),
+      std::end(keys));
+    auto query = std::make_shared<arangodb::velocypack::Builder>();
```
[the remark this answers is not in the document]
kvahed: done

arangod/Cluster/Maintenance.cpp
kvahed: done, thank you

arangod/Cluster/AgencyCache.cpp
kvahed: have taken the suggestion for a general overhaul that also goes into devel

KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12311/

arangod/Agency/Agent.cpp
```diff
@@ -253,11 +253,11 @@ AgentInterface::raft_commit_t Agent::waitFor(index_t index, double timeout) {
     ///  (_waitForCV's mutex stops writes to _commitIndex)
     CONDITION_LOCKER(guard, _waitForCV);
     if (leading()) {
-      if (lastCommitIndex != _commitIndex) {
+      if (lastCommitIndex != _commitIndex.load(std::memory_order_relaxed)) {
```
jsteemann, line 256: Here we are loading `_commitIndex` for a comparison, and 3 or 4 lines later we are loading it again unconditionally. We could fuse the two loads into just one.

arangod/Agency/Agent.cpp
```diff
@@ -327,6 +328,29 @@ index_t Agent::confirmed(std::string const& agentId) const {
 void Agent::reportIn(std::string const& peerId, index_t index, size_t toLog) {
   auto startTime = steady_clock::now();
 
+  if (index == 0) {
+    // This is only the empty case (=heartbeat)
+    MUTEX_LOCKER(locker, _emptyAppendLock);
+    auto n = steady_clock::now();
+    auto lastTime = _lastEmptyAcked[peerId];
```
jsteemann, line 335: This could create a new entry for `peerId` if it does not yet exist in the map. 
This is probably intentional, but it is worth a code comment.
In addition, in case the map value is updated later, we currently perform an extra lookup in the map.
Suggestion: do the initial lookup and assign the result variable by reference, so later the update does not have to perform an extra lookup.

arangod/Agency/Agent.cpp
```diff
@@ -327,6 +328,29 @@ index_t Agent::confirmed(std::string const& agentId) const {
...
+    if (lastTime < n) {
+      std::chrono::duration<double> d = n - lastTime;
+      auto secsSince = d.count();
+      if (secsSince < 1.5e9 && peerId != id() &&
```
jsteemann, line 339: This `1.5e9` looks very magic. Is the intention here that check if `lastTime` actually has a non-default value?

arangod/Agency/Agent.cpp
```diff
@@ -486,8 +502,8 @@ priv_rpc_ret_t Agent::recvAppendEntriesRPC(term_t term, std::string const& leade
           << " with term " << term;
       {
         WRITE_LOCKER(oLocker, _outputLock);
-        index_t const tmp = std::max(_commitIndex, std::min(leaderCommitIndex, lastIndex));
-        if (tmp > _commitIndex) {
+        index_t const tmp = std::max(_commitIndex.load(std::memory_order_relaxed), std::min(leaderCommitIndex, lastIndex));
+        if (tmp > _commitIndex.load(std::memory_order_relaxed)) {
```
jsteemann, line 506: `_commitIndex` is loaded twice in a row here.
If it doesn't have additional synchronization, then both loads could produce different values.
Even if not, it will be more efficient to load the value just once, and use it in both expressions.

arangod/Agency/Agent.cpp
```diff
@@ -515,8 +531,8 @@ priv_rpc_ret_t Agent::recvAppendEntriesRPC(term_t term, std::string const& leade
   {
     WRITE_LOCKER(oLocker, _outputLock);
     CONDITION_LOCKER(guard, _waitForCV);
-    index_t const tmp = std::max(_commitIndex, std::min(leaderCommitIndex, lastIndex));
-    if (tmp > _commitIndex) {
+    index_t const tmp = std::max(_commitIndex.load(std::memory_order_relaxed), std::min(leaderCommitIndex, lastIndex));
```
jsteemann, line 534: `_commitIndex` is loaded twice in a row here.
If it doesn't have additional synchronization, then both loads could produce different values.
Even if not, it will be more efficient to load the value just once, and use it in both expressions.

arangod/Agency/Agent.cpp
```diff
@@ -816,24 +816,31 @@ void Agent::sendEmptyAppendEntriesRPC(std::string const& followerId) {
...
+  double now = TRI_microtime();
   network::sendRequest(cp, _config.poolAt(followerId), fuerte::RestVerb::Post, "/_api/agency_priv/appendEntries",
                        std::move(buffer), reqOpts).thenValue([=](network::Response r) {
     ac->operator()(r);
   });
+  double diff = TRI_microtime() - now;
+  if (diff > 0.01) {
+    LOG_TOPIC("cfb7c", DEBUG, Logger::AGENCY)
+        << "Calling network::sendRequest took more than 1/100 of a second: " << diff;
+  }
```
jsteemann, line 836: If something takes longer than 10 ms, that is at least not completely unexpected on a loaded server.

arangod/Agency/Agent.cpp
```diff
@@ -864,29 +871,32 @@ void Agent::advanceCommitIndex() {
...
   {
     WRITE_LOCKER(oLocker, _outputLock);
-    if (index > _commitIndex) {
+    if (index > _commitIndex.load(std::memory_order_relaxed)) {
 
       CONDITION_LOCKER(guard, _waitForCV);
       LOG_TOPIC("e24a9", TRACE, Logger::AGENCY)
-          << "Critical mass for commiting " << _commitIndex + 1 << " through "
+          << "Critical mass for commiting "
+          << _commitIndex.load(std::memory_order_relaxed) + 1 << " through "
```
jsteemann, line 879: if `_commitIndex` has no external synchronization, then the log message may show a different commit index value than was used for the comparison in the `if`

arangod/Agency/Agent.cpp
```diff
@@ -864,29 +871,32 @@ void Agent::advanceCommitIndex() {
...
           << index << " to read db";
       // Change _readDB and _commitIndex atomically together:
       _readDB.applyLogEntries(_state.slices(/* inform others by callbacks */
-                                            _commitIndex + 1, index),
-                              _commitIndex, t, true);
+                                            _commitIndex.load(std::memory_order_relaxed) + 1, index),
+                              _commitIndex.load(std::memory_order_relaxed), t, true);
```
jsteemann, line 884: if `_commitIndex` has no external synchronization, then the two values of `_commitIndex` can be different here, and also different from the commit index value than was used for the comparison in the `if` above.
It would be better to load the `_commitIndex` value into a variable just once, and then use it for the if, the log message, and the function call to `applyLogEntries`.

arangod/Agency/Agent.cpp
```diff
@@ -909,17 +927,18 @@ futures::Future<query_t> Agent::poll(
...
+    } else if (index <= _commitIndex.load(std::memory_order_relaxed)) {
+      // deliver immediately all logs since index
       builder = std::make_shared<VPackBuilder>();
       VPackObjectBuilder r(builder.get());
       builder->add(VPackValue("result"));
       VPackObjectBuilder r2(builder.get());
-      logs = _state.get(index, _commitIndex);
-      builder->add("commitIndex", VPackValue(_commitIndex));
+      logs = _state.get(index, _commitIndex.load(std::memory_order_relaxed));
+      builder->add("commitIndex", VPackValue(_commitIndex.load(std::memory_order_relaxed)));
```
jsteemann, line 941: There is again a potential race here in case `_commitIndex` changes between the `if` statement and the next two usage of it.

arangod/Agency/Agent.cpp
```diff
@@ -1838,11 +1852,12 @@ void Agent::rebuildDBs() {
...
   LOG_TOPIC("b12cb", DEBUG, Logger::AGENCY)
       << "Rebuilding key-value stores from index " << lastCompactionIndex
-      << " to " << _commitIndex << " " << _state;
+      << " to " << _commitIndex.load(std::memory_order_relaxed) << " " << _state;
 
   {
-    auto logs = _state.slices(lastCompactionIndex + 1, _commitIndex);
-    _readDB.applyLogEntries(logs, _commitIndex, term, false /* do not send callbacks */);
+    auto logs = _state.slices(lastCompactionIndex + 1, _commitIndex.load(std::memory_order_relaxed));
+    _readDB.applyLogEntries(logs, _commitIndex.load(std::memory_order_relaxed),
```
jsteemann, line 1859: potential race between the two values of `_commitIndex` here.

arangod/Agency/Agent.cpp
```diff
@@ -2283,10 +2293,10 @@ query_t Agent::buildDB(arangodb::consensus::index_t index) {
 
   {
     READ_LOCKER(oLocker, _outputLock);
-    if (index > _commitIndex) {
+    if (index > _commitIndex.load(std::memory_order_relaxed)) {
       LOG_TOPIC("88754", INFO, Logger::AGENCY)
-          << "Cannot snapshot beyond leaderCommitIndex: " << _commitIndex;
-      index = _commitIndex;
+          << "Cannot snapshot beyond leaderCommitIndex: " << _commitIndex.load(std::memory_order_relaxed);
+      index = _commitIndex.load(std::memory_order_relaxed);
```
jsteemann, line 2299: would be more efficient to load the value of `_commitIndex` just once into a variable, and then use it from there.

jsteemann:
I think there is a potential data race in `Agent::waitFor(...)`.
This loads the value of `_commitIndex` twice, without holding `_outputLock` in read or write mode. That means another thread that modifies `_commitIndex` (while holding `_outputLock`) can modify the value of `_commitIndex` while `waitFor(...)` is still ongoing.

arangod/Cluster/MaintenanceFeature.cpp
```diff
@@ -950,17 +953,80 @@ void MaintenanceFeature::proceed() {
...
+  std::unordered_set<std::string> const& more) {
+  auto& clusterFeature = server().getFeature<ClusterFeature>();
+  auto ret = clusterFeature.dirty(); // plan & current in first run
+  if (_firstRun) {
+    auto all = allDatabases();
+    ret.insert(std::make_move_iterator(all.begin()),std::make_move_iterator(all.end()));
+    _firstRun = false;
+  } else {
+    if (!more.empty()) {
+      ret.insert(std::make_move_iterator(more.begin()),std::make_move_iterator(more.end()));
```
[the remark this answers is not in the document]
kvahed: done

arangod/Cluster/MaintenanceFeature.h
```diff
@@ -78,6 +78,9 @@ class MaintenanceFeature : public application_features::ApplicationFeature {
   void collectOptions(std::shared_ptr<options::ProgramOptions>) override;
   void validateOptions(std::shared_ptr<options::ProgramOptions>) override;
 
+  // @brief #databases last time we checked allDatabases
+  size_t lastNumberOfDatabases();
```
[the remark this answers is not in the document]
kvahed: thx

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1027,145 +1040,147 @@ void ClusterInfo::loadPlan() {
...
 
-    for (auto const& databasePairSlice : velocypack::ObjectIterator(planCollectionsSlice, true)) {
-      auto const& collectionsSlice = databasePairSlice.value;
+    std::vector<std::string> collectionsPath{
+      AgencyCommHelper::path(), "Plan", "Collections", databaseName};
+    if (!collectionsSlice.hasKey(collectionsPath)) {
+      auto it = newCollections.find(databaseName);
+      if (it != newCollections.end()) {
+        for (auto collection : *(it->second)) {
+          auto collectionId = collection.first;
```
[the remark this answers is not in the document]
kvahed: of course

arangod/Agency/Agent.cpp
kvahed: commitIndex is only changed under _outputLock. But still makes sense to load once.

arangod/Agency/Agent.cpp
kvahed: _commitIndex is only changed under _outputLock. the atomicity was only introduced for reads

arangod/Agency/Agent.cpp
kvahed: agree

kvahed:
> I think there is a potential data race in `Agent::waitFor(...)`.
> This loads the value of `_commitIndex` twice, without holding `_outputLock` in read or write mode. That means another thread that modifies `_commitIndex` (while holding `_outputLock`) can modify the value of `_commitIndex` while `waitFor(...)` is still ongoing.
> All places that modify `_commitIndex` hold the `_outputLock` in write mode. All places that modify it except `recvAppendEntriesRPC` also hold a condition variable mutex.
> All places that read `_commitIndex` multiple times either hold the `_outputLock` in read mode or the condition variable. The only place doesn't it `waitFor(...)`.

again, _commitIndex must not be changes without _outputLock.

arangod/Agency/Agent.cpp
kvahed: _outputLock

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1402,10 +1433,11 @@ void ClusterInfo::loadCurrent() {
   }
 
   _currentProt.isValid = true;
+  feature.addDirty(changeSet.dbs);
```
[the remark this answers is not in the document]
kvahed: this is just a const ref no advantage to move.

arangod/Agency/Agent.cpp
kvahed: no _outputLock must be held to change _commitIndex

arangod/Agency/Agent.cpp
kvahed: This is debug output. Please leave as is as we have made this changes recently for being able to understand customer issues.

arangod/Agency/Agent.cpp
kvahed: intentional yes

arangod/Cluster/AgencyCache.cpp
gnusi: that's actually not true
https://godbolt.org/z/5zTvjo

please consider lines 416, 418 of disassembly

jsteemann:
> > I think there is a potential data race in `Agent::waitFor(...)`.
> > This loads the value of `_commitIndex` twice, without holding `_outputLock` in read or write mode. That means another thread that modifies `_commitIndex` (while holding `_outputLock`) can modify the value of `_commitIndex` while `waitFor(...)` is still ongoing.
> > All places that modify `_commitIndex` hold the `_outputLock` in write mode. All places that modify it except `recvAppendEntriesRPC` also hold a condition variable mutex.
> > All places that read `_commitIndex` multiple times either hold the `_outputLock` in read mode or the condition variable. The only place doesn't it `waitFor(...)`.
> 
> again, _commitIndex must not be changes without _outputLock.

I understand this. But the code in this PR was *reading* it multiple times in one place (Agent::waitFor) without holding this lock, so this place was prone to data races. As you have changed that code via https://github.com/arangodb/arangodb/commit/8b134c68ea4df278014819e8fe4901e38dc20e09#diff-a08c67ca9d8698cb38a963fd20dd4999d9163875d86fb777944a888db0222700 I guess you agree to this finding.

kvahed:
I changed it to appease.

jsteemann:
> I changed it to appease.

Sounds a bit like if I would be making up this data race out of thin air :thinking:

kvahed:
I'm pulling you leg. that's what it sounds like.

jsteemann (approved):

PR #12715, opened 2020-09-25
feature/frozen-features

gnusi:
Compile time server feature lookup


lib/ApplicationFeatures/ApplicationServer.h
```diff
@@ -227,46 +225,42 @@ class ApplicationServer {
...
-   bool hasFeature(std::type_index type) const noexcept {
-     return (_features.find(type) != _features.end());
-   }
-
    // checks for the existence of a feature. will not throw when used for
    // a non-existing feature
    template <typename Type, typename std::enable_if<std::is_base_of<ApplicationFeature, Type>::value, int>::type = 0>
    bool hasFeature() const noexcept {
-     return hasFeature(std::type_index(typeid(Type)));
+     return nullptr == _features[Feature<Type>::Index];
```
jsteemann, line 244:
```suggestion
     return nullptr != _features[Feature<Type>::Index];
```

lib/ApplicationFeatures/ApplicationServer.h
```diff
@@ -29,11 +29,11 @@
 #include <memory>
 #include <string>
 #include <type_traits>
-#include <typeindex>
```
jsteemann, line 32: I think we should also include `<array>` here, as this header is now using it.

danielhlarkin:
I don't necessarily object to a compile-time feature map, but I do have some serious concerns about the current implementation centralizing all the types from server, shell, enterprise, etc. into `lib` via forward declarations. Is it possible instead to build some template class in lib that gets instantiated where the server is actually instantiated, e.g. in `arangod/RestServer/arangod.cpp`, `arangosh/Restore/arangorestore.cpp`, etc.?

gnusi:
@danielhlarkin absolutely, but all types have to be known per binary beforehand.

gnusi:
array vs unordered_map lookup
https://quick-bench.com/q/6S8N9fmH2dHOEICiRW24R4vl37c

gnusi:
superseded with #15610

PR #12716, opened 2020-09-25
Use dynamically-sized revision trees

danielhlarkin:
### Scope & Purpose

Currently, we lazily allocate fixed-size revision trees. For collections without so many documents, this is an overuse of memory. This PR starts out allocating small trees, and then grows them to full-size as the collection gets more documents.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

Jenkins: https://jenkins.arangodb.biz/job/arangodb-matrix-pr/11989/


lib/Containers/MerkleTree.cpp
```diff
@@ -474,16 +461,67 @@ std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> MerkleTree<BranchingBits
       new MerkleTree<BranchingBits, LockStripes>(*this));
 }
 
+template <std::size_t const BranchingBits, std::size_t const LockStripes>
+std::unique_ptr<MerkleTree<BranchingBits, LockStripes>>
+MerkleTree<BranchingBits, LockStripes>::cloneWithDepth(std::size_t newDepth) {
```
jsteemann, line 466: It looks like this method is not modifying the original tree (good!), so we could make it `const`.

lib/Containers/MerkleTree.cpp
```diff
@@ -474,16 +461,67 @@ std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> MerkleTree<BranchingBits
...
+  std::shared_lock<std::shared_mutex> guard(_bufferLock);
+
+  if (newDepth == meta().maxDepth) {
+    return std::unique_ptr<MerkleTree<BranchingBits, LockStripes>>(
+        new MerkleTree<BranchingBits, LockStripes>(*this));
```
jsteemann, line 471:
```suggestion
    return std::make_unique<MerkleTree<BranchingBits, LockStripes>>(*this);
```

lib/Containers/MerkleTree.cpp
```diff
@@ -474,16 +461,67 @@ std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> MerkleTree<BranchingBits
...
+  }
+
+  if (newDepth < meta().maxDepth) {
+    std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> newTree =
+        std::make_unique<MerkleTree<BranchingBits, LockStripes>>(newDepth,
+                                                                 meta().rangeMin,
+                                                                 meta().rangeMax);
```
jsteemann, line 478:
```suggestion
    auto newTree =
        std::make_unique<MerkleTree<BranchingBits, LockStripes>>(newDepth,
                                                                 meta().rangeMin,
                                                                 meta().rangeMax);
```

lib/Containers/MerkleTree.cpp
```diff
@@ -474,16 +461,67 @@ std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> MerkleTree<BranchingBits
...
+    return newTree;
+  }
+
+  // Otherwise let's grow the tree deeper. We're going to grow one level at a
+  // time, recursively. Typically we'll only be requesting to grow one level
+  // deeper at a time anyway, and we should very, very rarely be growing more
+  // than a couple levels at a time.
+  std::uint64_t newRangeMax =
+      meta().rangeMin + ((meta().rangeMax - meta().rangeMin) * BranchingFactor);
+  std::unique_ptr<MerkleTree<BranchingBits, LockStripes>> newTree =
```
jsteemann, line 493:
```suggestion
  auto newTree =
```

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -869,17 +867,67 @@ bool RocksDBMetaCollection::haveBufferedOperations() const {
...
+
+      bufferRemovals += pair.second.size();
+    }
+  }
+
+  std::size_t const treeDepth = _revisionTree ? _revisionTree->maxDepth() : 3;
+
+  if (bufferInserts > bufferRemovals) {
+    std::size_t const count = treeCount + bufferInserts - bufferRemovals;
+    std::size_t targetDepth = 3;
```
jsteemann, line 905: Should `targetDepth` start with the value of `treeDepth`?
I guess otherwise it would be possible to make this function return a smaller tree depth that the tree's current size.
```suggestion
    std::size_t targetDepth = 3;
```

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -869,17 +867,67 @@ bool RocksDBMetaCollection::haveBufferedOperations() const {
...
+    for (auto& pair : _revisionRemovalBuffers) {
+      if (pair.first <= ignoreSeq) {
+        continue;
+      }
+
+      bufferRemovals += pair.second.size();
+    }
+  }
+
+  std::size_t const treeDepth = _revisionTree ? _revisionTree->maxDepth() : 3;
```
jsteemann, line 901: this calculation is only used after the following `if` block, for the function's final return. So it could be moved below the `if` block for better readability locality.

lib/Containers/MerkleTree.cpp
danielhlarkin: To be honest, I'd like to avoid the use of `auto`.

lib/Containers/MerkleTree.cpp
danielhlarkin: To be honest, I'd like to avoid the use of `auto`.

lib/Containers/MerkleTree.cpp
danielhlarkin: `make_unique` requires the constructor to be public, while this one is currently protected. We don't really want that constructor to part of the public API, so I'm going to leave this as-is.

danielhlarkin:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/12001/

lib/Containers/MerkleTree.cpp
goedderz: Feel free to do that. But I'd like to point to this post: https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/
There are some good (also some less convincing) arguments for `auto` there.

danielhlarkin: @goedderz To clarify, I'm not against `auto` in general, just in this file. The type specificity for e.g. `size_t` vs `uint64_t` matters here, and has been a problem with `auto` in the past. Obviously the pointers are different, but it makes it a lot easier to double-check if I can just grep for `auto`.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -869,17 +867,67 @@ bool RocksDBMetaCollection::haveBufferedOperations() const {
...
+
+  if (bufferInserts > bufferRemovals) {
+    std::size_t const count = treeCount + bufferInserts - bufferRemovals;
+    std::size_t targetDepth = treeDepth;
+    while (targetDepth < 6 &&
```
graetzer, line 906: so if `treeDepth == 6` the whole loop above is not necessary?

danielhlarkin: Good point. I'll add in a shortcut.

danielhlarkin:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/12013/

PR #12719, opened 2020-09-28
Fix LIMIT with COLLECT WITH COUNT INTO

goedderz:
### Scope & Purpose

A `LIMIT` operation before a `COUNT WITH COLLECT INTO` did consume all input from above, resulting in an unnecessarily large runtime.

There is still a similar open problem with two adjacent top-level `LIMIT` operations when the query has the `fullCount` option set. This is more complicated to fix and will be addressed later.

- [X] :hankey: Bugfix 
- [X] :muscle: The behavior in this PR was *manually tested*
- [X] :computer: The behavior change can be verified via automatic tests

#### Backports:

- [X] Backport for *3.7*: https://github.com/arangodb/arangodb/pull/12744

#### Related Information

- [X] Jira ticket number: ES-692

### Testing & Verification

- [X] This change is already covered by existing tests, such as gtest *LimitExecutorTests*.
  - [X] Added **Regression Tests**

### Documentation

- [x] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)


goedderz:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12033/

arangod/Aql/LimitExecutor.cpp
```diff
@@ -69,13 +71,37 @@ auto LimitExecutor::calculateUpstreamCall(AqlCall const& clientCall) const -> Aq
...
+    // need to skip both our offset and our limit, regardless of the client's
+    // offset.
+    // If we ever got a client limit > 0 and thus produced rows, we may not send
+    // a non-zero offset, and thus must avoid this branch.
+    TRI_ASSERT(!useSoftLimit);
+    TRI_ASSERT(clientCall.hasHardLimit());
+
+    upstreamCall.offset = upstreamCall.offset + remainingLimit();
+    upstreamCall.hardLimit = 0;
+    // We need to send fullCount upstream iff this is fullCount-enabled LIMIT
```
hkernbach, line 90:
```suggestion
    // We need to send fullCount upstream if this is fullCount-enabled LIMIT
```

arangod/Aql/LimitExecutor.cpp
```diff
@@ -69,13 +71,37 @@ auto LimitExecutor::calculateUpstreamCall(AqlCall const& clientCall) const -> Aq
...
     upstreamCall.hardLimit = limit;
-    // We need the fullCount either if we need to report it ourselfes.
-    // or if the clientCall needs to report it.
-    upstreamCall.fullCount = infos().isFullCountEnabled() || clientCall.fullCount;
+    // We need the fullCount if we need to report it ourselves.
+    // If the client needs full count, we need to skip up to our limit upstream.
+    // But currently the execute API does not allow limited skipping after
+    // fetching rows, so we must pass fullCount upstream, even if that's
+    // inefficient.
+    // This is going to be fixed in a later PR.
```
hkernbach, line 103: 👍🏼

hkernbach (approved):
LGTM, just a minor typo found. Nice tests!

arangod/Aql/LimitExecutor.cpp
mchacki: @hkernbach I think the original is actually correct.

`iff` is a short-hand for `if and only if` which is correct here.

mchacki (approved):
LGTM!

PR #12720, opened 2020-09-28
take a copy before moving from the same object

jsteemann:
### Scope & Purpose

The following existing code
```
    auth::User user = auth::User::fromDocument(s);
    result.try_emplace(user.username(), std::move(user));
```
moves frod `user` and accesses a member function of `user` in the same expression.

In practice, this should be safe, because the map's insert key should be calculated first, and the value is only moved from if the key does not exist in the map.
But I couldn't find any proof that this is guaranteed to be safe.
So changing it to an intentional and explicit copy to avoid potential UB:
```
    auth::User user = auth::User::fromDocument(s);
    // intentional copy, as we are about to move-away from user
    std::string username = user.username();
    result.try_emplace(std::move(username), std::move(user));
```

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *any authentication/user tests*.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12004/

### DIFF
--- a/arangod/Aql/QueryRegistry.cpp
+++ b/arangod/Aql/QueryRegistry.cpp
@@ -22,7 +22,6 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "QueryRegistry.h"
-#include <Cluster/CallbackGuard.h>
 
 #include "Aql/AqlItemBlock.h"
 #include "Aql/ClusterQuery.h"
@@ -30,6 +29,7 @@
 #include "Basics/ReadLocker.h"
 #include "Basics/WriteLocker.h"
 #include "Basics/system-functions.h"
+#include "Cluster/CallbackGuard.h"
 #include "Cluster/ServerState.h"
 #include "Cluster/TraverserEngine.h"
 #include "Logger/LogMacros.h"
--- a/arangod/Auth/UserManager.cpp
+++ b/arangod/Auth/UserManager.cpp
@@ -107,8 +107,8 @@ static auth::UserMap ParseUsers(VPackSlice const& slice) {
   for (VPackSlice const& authSlice : VPackArrayIterator(slice)) {
     VPackSlice s = authSlice.resolveExternal();
 
-    if (s.hasKey("source") && s.get("source").isString() &&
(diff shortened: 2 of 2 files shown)


graetzer (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12009/

mpoeter:
FWIW - the original code was perfectly valid, because - despite its name - `std::move` does not actually move anything. In fact, at runtime, `std::move` doesn't do anything at all, it doesn't even generate any executable code. `std::move` is merely a function that casts its argument to an rvalue, that's it. (Similarly, `std::forward` performs a conditional cast.) The result type of `std::move` is what allows the compiler to pass the object to a move-constructor or move-assigment-operator - that is were the actual move happens. So for the code in question, the `std::move` simply changes deduced type of the second parameter for the `try_emplace` call from an lvalue reference to an rvalue reference. The move itself is performed _inside_ `try_emplace`.

goedderz:
Yes, but `user.username()` returns a reference to `User._username`, and `try_emplace` does take a reference as well. So if in `try_emplace` the actual move of the value did happen before copying the key, it would be broken. Or am I overlooking something?

jsteemann:
@goedderz: that was exactly my concern.
I don't think that `try_emplace` would be smart to first move/modify the value object before accessing the key part (and making sure it does not yet exist in the map or reserve space for it in case it doesn't). But at least it could be _possible_ that it does so:
```
try_emplace(K const& key, V&& value) -> auto:
  if not key_exists(key):
    value_node = create_value_node(move(value))
    # access of key after value is modified
    key_node = create_key(key)
    link(key_node, value_node)
...
```

mpoeter:
Ah, now I see the problem - it's not that you were worried that the move occurred before the evaluation of `user.username()`, but that you end up with a reference to a member of the moved-from object. In that case I agree that this is potentially unsafe. 👍

danielhlarkin (approved):

PR #12725, opened 2020-09-29
[3.7] efficient empty appends

kvahed:
### Scope & Purpose

- [ ] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required

#### Related Information

*(Please reference tickets / specification etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Additionally:

### Documentation

> All new Features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the changelog so that
> developers and users have a concise overview. 

- [ ] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)

### DIFF
--- a/arangod/Agency/Agent.cpp
+++ b/arangod/Agency/Agent.cpp
@@ -824,17 +824,23 @@ void Agent::sendEmptyAppendEntriesRPC(std::string const& followerId) {
          .param("leaderCommit", std::to_string(commitIndex))
          .param("senderTimeStamp", std::to_string(std::llround(steadyClockToDouble() * 1000)));
 
+  double now = TRI_microtime();
   network::sendRequest(cp, _config.poolAt(followerId), fuerte::RestVerb::Post, "/_api/agency_priv/appendEntries",
                        std::move(buffer), reqOpts).thenValue([=](network::Response r) {
     ac->operator()(r);
   });
+  double diff = TRI_microtime() - now;
+  if (diff > 0.01) {
+    LOG_TOPIC("cfb7c", DEBUG, Logger::AGENCY)
+        << "Calling network::sendRequest took more than 1/100 of a second: " << diff;
+  }
 
   _constituent.notifyHeartbeatSent(followerId);
 
-  double now = TRI_microtime();
+  now = TRI_microtime();
   LOG_TOPIC("54798", DEBUG, Logger::AGENCY)
       << "Sending empty appendEntriesRPC to follower " << followerId;
-  double diff = TRI_microtime() - now;
+  diff = TRI_microtime() - now;
   if (diff > 0.01) {
     LOG_TOPIC("cfb7b", DEBUG, Logger::AGENCY)
         << "Logging of a line took more than 1/100 of a second, this is bad:" << diff;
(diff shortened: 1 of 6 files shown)


arangod/Agency/Agent.cpp
```diff
@@ -327,6 +327,15 @@ index_t Agent::confirmed(std::string const& agentId) const {
 void Agent::reportIn(std::string const& peerId, index_t index, size_t toLog) {
   auto startTime = steady_clock::now();
 
+  if (index == 0) {
```
neunhoef, line 330: Let's add a comment here to say that this happens if and only if we are receiving a response for an empty appendentriesRPC.

neunhoef (approved):
Just one comment to add, please. See comment. LGTM

lib/Logger/LogThread.cpp
```diff
@@ -29,7 +29,7 @@
 using namespace arangodb;
 
 LogThread::LogThread(application_features::ApplicationServer& server, std::string const& name)
-    : Thread(server, name), _messages(16) {}
+    : Thread(server, name), _messages(16384) {}
```
jsteemann, line 32: I am not sure if this change is justified.
The boost::lockfree::queue we are using _can_ already grow dynamically. So I think we can get away with a much smaller reserve size, e.g. 32 or 64, and then have it grow dynamically if it actually needs more space (it will never do in most environments).
```suggestion
    : Thread(server, name), _messages(64) {}
```

maierlars: If that is actually the case, then I think we need a new theory why a `LOG_TOPIC` is blocking the thread issuing this log message for a long time, when the disk is slow and busy. @neunhoef

kvahed: I understand that @neunhoef has measured a case, where LOG_TOPIC took 1.4 seconds to finish.

mpoeter: The boost::lockfree::queue is an implementation of the Michael-Scott queue, which is based on a linked list and is effectively unbounded. However, the boost::lockfree::queue uses a freelist to manage the available nodes (and thereby work around the memory reclamation problem), and by limiting the available nodes on the free list, it can effectively limit the number of elements in the queue, making it bounded/fixed size. But for that the queue has to be defined using the `boost::lockfree::fixed_sized<true>` policy, which is not the case here. The number passed to the constructor only defines the number of preallocated nodes on the freelist. For a non-fixed sized queue like this, new nodes will be allocated on demand if the freelist is empty, so I agree with @jsteemann here.

neunhoef: We can go to 64, no problem. But we still need to understand how a LOG_TOPIC operation can take 1.4 seconds in situations with a bad IO system...

neunhoef: Done.

neunhoef (approved):
LGTM

lib/Logger/LogThread.cpp
mpoeter: @neunhoef can you share some details under which circumstances you observe this behavior? I'm wondering if it might be possible to reproduce this locally so I could investigate.

PR #12731, opened 2020-09-30
Slightly speed up initial replication of collections/shards

jsteemann:
### Scope & Purpose

Slightly speed up initial replication of collections/shards data by not wrapping each document in a seperate `{"type":2300,"data":...}` envelope. The "type" attribute is from the MMFiles era, when collections included document and removal markers, but is superfluous in the RocksDB era, where a dump response will only contain documents.

Stripping of the envelopes allows slightly smaller data sizes when exchanging the documents, and thus can less to small savings in document packing and unpacking as well as slightly reduce the number of required HTTP requests.
The non-wrapping of the documents is opt-in, so the existing dump API stays downwards-compatible. The current replication code on the follower will ask the follower to produce a dump result without envelopes, which the leader is free to ignore. The follower code will automatically figure out which format the response has, so there should be no compatibility issues here in any case.
The complexity added by this PR is tolerable IMHO.

The actual savings depend on the number of documents and their sizes, the available processing power and the network speed/capacity. In some scenarios, a speedup of a single-digit percentage value can be expected.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *replication tests*.
  - [x] Added new **integration tests** (i.e. in http_replication)

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12019/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,15 @@
 devel
 -----
 
+* Slightly speed up initial replication of collections/shards data by not
+  wrapping each document in a seperate `{"type":2300,"data":...}` envelope.
+  Stripping of the envelopes allows for slightly smaller data sizes when 
+  exchanging the documents, and thus can less to small savings in document
+  packing and unpacking as well as slightly reduce the number of required
+  HTTP requests. 
+  The non-wrapping of the documents is opt-in, so the existing dump API 
+  stays downwards-compatible.
+
 * Data definition reconsiliation in cluster has been modified
   extensively to greatly accelerate the creation of 1000s of
   databases through following means:
--- a/arangod/Replication/DatabaseInitialSyncer.cpp
+++ b/arangod/Replication/DatabaseInitialSyncer.cpp
@@ -534,28 +534,27 @@ Result DatabaseInitialSyncer::parseCollectionDumpMarker(transaction::Methods& tr
   }
 
   TRI_replication_operation_e type = REPLICATION_INVALID;
-  VPackSlice doc;
-
-  for (auto it : VPackObjectIterator(marker, true)) {
-    if (it.key.isEqualString(kTypeString)) {
-      if (it.value.isNumber()) {
-        type = static_cast<TRI_replication_operation_e>(it.value.getNumber<int>());
-      }
-    } else if (it.key.isEqualString(kDataString)) {
-      if (it.value.isObject()) {
-        doc = it.value;
-      }
-    }
-    if (type != REPLICATION_INVALID && doc.isObject()) {
-      break;
-    }
-  }
-
-  if (!doc.isObject()) {
-    return TRI_ERROR_REPLICATION_INVALID_RESPONSE;
-  }
-  // key must not be empty, but doc can otherwise be empty
+  VPackSlice doc = marker;
   VPackSlice key = doc.get(StaticStrings::KeyString);
+
+  if (key.isString()) {
+    // input is just a document, without any {"type":2300,"data":{...}} envelope
+    type = REPLICATION_MARKER_DOCUMENT;
+  } else {
+    // input is wrapped in a {"type":2300,"data":{...}} envelope
+    VPackSlice s = marker.get(kTypeString);
+    if (s.isNumber()) {
+      type = static_cast<TRI_replication_operation_e>(s.getNumber<int>());
+    }
+    s = marker.get(kDataString);
+    if (!s.isObject()) {
+      return TRI_ERROR_REPLICATION_INVALID_RESPONSE;
+    }
+    doc = s;
+    key = doc.get(StaticStrings::KeyString);
+  }
+
+  // key must not be empty, but doc can otherwise be empty
   if (!key.isString() || key.getStringLength() == 0) {
     return TRI_ERROR_REPLICATION_INVALID_RESPONSE;
   }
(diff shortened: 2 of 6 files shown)


arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -534,28 +534,27 @@ Result DatabaseInitialSyncer::parseCollectionDumpMarker(transaction::Methods& tr
...
-  for (auto it : VPackObjectIterator(marker, true)) {
-    if (it.key.isEqualString(kTypeString)) {
-      if (it.value.isNumber()) {
-        type = static_cast<TRI_replication_operation_e>(it.value.getNumber<int>());
-      }
-    } else if (it.key.isEqualString(kDataString)) {
-      if (it.value.isObject()) {
-        doc = it.value;
-      }
+  if (key.isString()) {
```
graetzer, line 540: sort of a brittle detection isn't it, shouldn't you check whether you set useEnvelope?
Also why not just dump vpack directly instead of using json? Should that not be even less traffic

graetzer (commented):
this lacks a benchmark

arangod/Replication/DatabaseInitialSyncer.cpp
jsteemann: It is deliberately not using the value of `useEnvelope` here, because it is unknown if the leader supports it or not.
This allows an auto-detection of the format.
An alternative would be to make any leader that supports this format return a new response header, so that it can be detected without checking in each document.
Both will work.

graetzer:
```suggestion
  if (useEnvelope && key.isString()) {
```

graetzer: why not set the accept header to use velocypack responses, should be even faster no?

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -267,7 +267,7 @@ Result RocksDBReplicationContext::getInventory(TRI_vocbase_t& vocbase, bool incl
 // creating a new iterator if one does not exist for this collection
 RocksDBReplicationContext::DumpResult RocksDBReplicationContext::dumpJson(
```
graetzer, line 268: whyt about dumpVpack

jsteemann:
> this lacks a benchmark

I ran a manual test locally, but this obviously highly depends on document count, size, CPU and network speed.
In my case it showed savings between around 5 and 10%, but this was on a localhost laptop.

What kind of benchmark would you like to see?

arangod/Replication/DatabaseInitialSyncer.cpp
graetzer: unless vpack validation is slower than JSON serialization / deserialization

graetzer:
> > this lacks a benchmark
> 
> I ran a manual test locally, but this obviously highly depends on document count, size, CPU and network speed.
> In my case it showed savings between around 5 and 10%, but this was on a localhost laptop.
> 
> What kind of benchmark would you like to see?

thats ok then

CHANGELOG
```diff
@@ -1,6 +1,15 @@
 devel
 -----
 
+* Slightly speed up initial replication of collections/shards data by not
+  wrapping each document in a seperate `{"type":2300,"data":...}` envelope.
+  Stripping of the envelopes allows for slightly smaller data sizes when 
+  exchanging the documents, and thus can less to small savings in document
```
danielhlarkin, line 7:
```suggestion
  exchanging the documents, and thus can lead to small savings in document
```

danielhlarkin (approved):

graetzer:
```bash
diff --git a/arangod/Replication/DatabaseInitialSyncer.cpp b/arangod/Replication/DatabaseInitialSyncer.cpp
index 5edc53cef4a..1baeef7ee45 100644
--- a/arangod/Replication/DatabaseInitialSyncer.cpp
+++ b/arangod/Replication/DatabaseInitialSyncer.cpp
@@ -686,12 +686,10 @@ void DatabaseInitialSyncer::fetchDumpChunk(std::shared_ptr<Syncer::JobSynchroniz
     }
 
     auto headers = replutils::createHeaders();
-#if VPACK_DUMP
     int vv = _config.leader.majorVersion * 1000000 + _config.leader.minorVersion * 1000;
     if (vv >= 3003009) {
       headers[StaticStrings::Accept] = StaticStrings::MimeTypeVPack;
     }
-#endif
 
     _config.progress.set(
         std::string("fetching leader collection dump for collection '") +
```

why not additionally try out the vpack encoded response?

jsteemann:
> ```shell
> diff --git a/arangod/Replication/DatabaseInitialSyncer.cpp b/arangod/Replication/DatabaseInitialSyncer.cpp
> index 5edc53cef4a..1baeef7ee45 100644
> --- a/arangod/Replication/DatabaseInitialSyncer.cpp
> +++ b/arangod/Replication/DatabaseInitialSyncer.cpp
> @@ -686,12 +686,10 @@ void DatabaseInitialSyncer::fetchDumpChunk(std::shared_ptr<Syncer::JobSynchroniz
>      }
>  
>      auto headers = replutils::createHeaders();
> -#if VPACK_DUMP
>      int vv = _config.leader.majorVersion * 1000000 + _config.leader.minorVersion * 1000;
>      if (vv >= 3003009) {
>        headers[StaticStrings::Accept] = StaticStrings::MimeTypeVPack;
>      }
> -#endif
>  
>      _config.progress.set(
>          std::string("fetching leader collection dump for collection '") +
> ```
> 
> why not additionally try out the vpack encoded response?

Will try!

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12104/

arangod/Replication/DatabaseInitialSyncer.cpp
jsteemann: Done now.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12108/

jsteemann:
Tests blue

graetzer (approved):

PR #12735, opened 2020-09-30
Deterministically lock exclusive shards

graetzer:
### Scope & Purpose

Make sure exclusive shards are not lazily locked

- [x] :hankey: Bugfix 

#### Backports:

### Testing & Verification

http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12067/

- [x] This change is already covered by existing tests, such as `tests/js/client/shell/shell-transaction.js`
  - [x] Added **Regression Tests**

### Documentation

- [x] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)
- [x] Added a new section in the *Manual* 
- [x] Added a new section in the *HTTP API*


graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12066/

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12067/

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12074/

Documentation/DocuBlocks/Rest/Transactions/post_api_transaction_begin.md
```diff
@@ -23,8 +23,7 @@ Allow reading from undeclared collections.
 @RESTBODYPARAM{lockTimeout,integer,optional,int64}
 an optional numeric value that can be used to set a
 timeout for waiting on collection locks. If not specified, a default
-value will be used. Setting *lockTimeout* to *0* will make ArangoDB
-not time out waiting for a lock.
+value will be used. The maximum value for stream transactions is 30 seconds.
```
mchacki, line 26: So i do understand this as:
* lockTimeout = 0 means we do not wait? (please clarify 0 here)
* lockTimeout = 5 => we wait 5 sec (ok)
* lockTimeout = 55 => we wait 30 sec?

What is the defaultValue we use?

arangod/Aql/ShardLocking.cpp
```diff
@@ -248,6 +248,7 @@ std::vector<ServerID> ShardLocking::getRelevantServers() {
                    TRI_ASSERT(!item.first.empty());
                    return item.first;
                  });
+  std::sort(result.begin(), result.end());
```
mchacki, line 251: I think this is a very important part of the fix.

mchacki: Can you please add a code comment on this function and in the header file that the Result is sorted, and that this sorting is required to avoid distributed dead-locking.
This increases the chance to not lose the sortedness in the future.

arangod/Transaction/Manager.h
```diff
@@ -63,6 +63,7 @@ class Manager final {
   static constexpr double idleTTLDBServer = 3 * 60.0;              //  3 minutes
   static constexpr double tombstoneTTL = 10.0 * 60.0;              // 10 minutes
   static constexpr size_t maxTransactionSize = 128 * 1024 * 1024;  // 128 MiB
+  static constexpr double maxLockTimeout = 60.0;                   // 60 seconds
```
mchacki, line 66: The documentation claimed this to be 30sec not 60s

arangod/Aql/ShardLocking.cpp
mchacki: ok now I am bit unsure if the sort here is actually required.
In the critical locking path (in ClusterTransaction state) the servers are sorted and contacted in serialized ordering.
Do both places need to be sorted?

arangod/Cluster/ClusterTrxMethods.cpp
```diff
@@ -307,33 +334,22 @@ Future<Result> beginTransactionOnLeaders(TransactionState& state,
...
-          network::Response const& resp = tryRes.get();  // throws exceptions upwards
-          Result res = ::checkTransactionResult(tid, transaction::Status::RUNNING, resp);
-          if (res.fail()) {  // remove follower from all collections
-            return res;
-          }
-        }
-        return Result();  // all good
-      });
+  auto servers = std::make_shared<std::vector<ServerID>>(leaders);
+  std::sort(servers->begin(), servers->end(), std::greater<ServerID>());
```
mchacki, line 351: You are sorting the servers here and also sorting them on the getRelevantServers function above.

If I am not mistaken this code place here is the critical one, the other can work with non-sortedness of Servers.
Can you please verify if I am off here, or if we can get away with the sorting here alone.

arangod/ClusterEngine/ClusterTransactionState.cpp
```diff
@@ -83,6 +83,35 @@ Result ClusterTransactionState::beginTransaction(transaction::Hints hints) {
 
   transaction::ManagerFeature::manager()->registerTransaction(id(), isReadOnlyTransaction(), false /* isFollowerTransaction */);
   setRegistered();
+  
+  // simon: deadlock-mitiagtion, still unsafe in general
```
mchacki, line 87:
```suggestion
  // simon: deadlock-mitigation, still unsafe in general
```

arangod/ClusterEngine/ClusterTransactionState.cpp
```diff
@@ -83,6 +83,35 @@ Result ClusterTransactionState::beginTransaction(transaction::Hints hints) {
...
+  // we have to include write-level transactions too for correctness
```
mchacki, line 88: I think this code is actually correct, but has a chance to get into a now limited dead-lock situation.
And this only happens if you mix Exclusive Transactions and Write Transactions on the same collection.
If you only use one of them it works without a chance for dead-locking and the write transactions being lazy and potentially faster.

tests/js/client/communication/test-communication.js
```diff
@@ -229,6 +234,19 @@ function CommunicationSuite () {
...
+    testStreamTrxInParallel: function() {
+      let tests = [
+        [ 'stream-trx-1', 'let t = db._createTransaction({ collections: { exclusive: "UnitTestsExc" } }); let tc = t.collection("UnitTestsExc"); let doc = tc.document("entityName"); tc.update(doc._key, {nextOID: doc.nextOID + 1}); t.commit();'],
+        [ 'stream-trx-2', 'let t = db._createTransaction({ collections: { exclusive: "UnitTestsExc" } }); t.query(\'LET d = DOCUMENT("UnitTestsExc/entityName") UPDATE d WITH {nextOID: d.nextOID + 1} IN UnitTestsExc\'); t.commit();'],
+        [ 'stream-trx-3', 'let t = db._createTransaction({ collections: { exclusive: "UnitTestsExc" } }); let tc = t.collection("UnitTestsExc"); for (i = 0; i < 50; i++) {tc.save({nextOID: 0})} t.commit();'],
+        [ 'stream-trx-4', 'let t = db._createTransaction({ collections: { exclusive: "UnitTestsExc" } }); t.query(\'FOR d IN UnitTestsExc UPDATE d WITH {nextOID: d.nextOID + 1} IN UnitTestsExc\'); t.commit();']
+      ];
+
+      // run the suite for 1 minutes
+      runTests(tests, 1 * 60);
```
mchacki, line 248: Does this test-suite really catch what you are trying to get here?
This PR limits the max lock time to 30s (or 60s see comments above) this test code tries to perform as many of the above operations as possible, and then tries to shut-down the clients.
In the shutdown process it will wait for at least 30s in worst-case.

So even if the above queries get into a "dead-lock" and are blocking for <maxLocktime> now this test will succeed, and not operation has taken place.

So this test checks if we will eventually get unlocked.
It will not check if we still get into a locked situation.
Can we somehow track the time spend in locks (or spend within the transaction), and assert that it is at least close to 1 * 60 on exclusive case. And above 1 * 60 on write only case? This will still not be perfect as locking here is still a game of chance, but it will eventually hit.

mchacki (changes requested):
Code in general looks good, I have some comments.
Especially I am not convinced by the test.

arangod/ClusterEngine/ClusterTransactionState.cpp
graetzer: that was the point

Documentation/DocuBlocks/Rest/Transactions/post_api_transaction_begin.md
danielhlarkin: In order to maintain something resembling backwards compatibility, we probably need to interpret an input of 0 as the max. I believe that's exactly what we do now, but we should probably document it here.

arangod/Cluster/ClusterTrxMethods.cpp
graetzer: no, this does not use getRelevantServers()

arangod/Aql/ShardLocking.cpp
graetzer: This is used in the AQL setup code

Documentation/DocuBlocks/Rest/Transactions/post_api_transaction_begin.md
graetzer: 60 seconds

tests/js/client/communication/test-communication.js
graetzer: my assumption was that if an arangosh instance gets into a dead-lock then it will not terminate by itself within the 30s time-frame allowed by the `runTests(...)` function. 
This should then manifest as an error of this testsuite

jsteemann:
Largely covered and obsoleted by PR https://github.com/arangodb/arangodb/pull/13207.
However, some small changes from this PR still need to be salvaged.

jsteemann:
Obsoleted by other PRs in devel, which addressed the same problem slightly differently.

Issue #5311, opened 2018-05-10
Script results are not escaped for HTML in the web interface


zocky:
## my environment running ArangoDB
I'm using the latest ArangoDB of the respective release series:

- [x ] 3.3

When the UI is used to run a foxx service script which returns a string which contains html/xml tags, the results are displayed as JSON inside a `<pre>` tag, but they are not HTML-escaped.

### Steps to reproduce:
(1) Create this script in a foxx service:
```js
module.exports = `foo <b>bar</b>`;
```
(2) Go to the service settings, choose the script from the dropdown and run it.

### Produced results:

The UI displays `"foo `**`bar`**`"`

### Expected results:

The UI should display `"foo <b>bar</b>"`

hkernbach:
Sorry for the late reply. Applied bugfix PR's for all supported versions.

pluma:
Closing this as per above comment.

PR #12737, opened 2020-09-30
now escaping the result of a foxx services setup script

hkernbach:
### Scope & Purpose

Added missing UI HTML escaping.

Old behaviour can be seen here:
- https://github.com/arangodb/arangodb/issues/5311

New (correct) behaviour:
image

- [x] :hankey: Bugfix 

#### Backports:

- [x] Backports required for: 3.6 + devel

#### Related Information

- [x] GitHub issue / Jira ticket number: https://github.com/arangodb/arangodb/issues/5311

### Testing & Verification

Additionally:

http://jenkins01:8080/view/PR/job/arangodb-matrix-pr/12027/


js/apps/system/_admin/aardvark/APP/frontend/js/views/applicationDetailView.js
```diff
@@ -192,7 +192,7 @@
             } else if (result) {
               info = (
                 '<p>Script results:</p>' +
-                '<pre>' + JSON.stringify(result, null, 2) + '</pre>'
+                '<pre>' + JSON.stringify(arangoHelper.escapeHtml(result), null, 2) + '</pre>'
```
pluma, line 195: Shouldn't this be `arangoHelper.escapeHtml(JSON.stringify(result), null, 2)`?

pluma: i.e. it should also escape `{"xss": "<script>alert('hi')</script>"}` correctly but to me it looks like this would result in that being converted to `"[object Object]"` (assuming `arangoHelper.escapeHtml` expects a string).

hkernbach: > Shouldn't this be `arangoHelper.escapeHtml(JSON.stringify(result), null, 2)`?

Why? Result needs to be escaped as it is the input with potential html code. We could remove the stringify if any input is expected. But it will not harm. But sure, if you have a better approach just inline edit and push.

pluma: What is the purpose of `JSON.stringify` here? I assumed it's so we can represent whatever arbitrary data the script returns, which might be a JSON object (i.e. whatever you assigned to `module.exports` in the script).

In that case running `arangoHelper.escapeHtml` on it would coerce it to a string, try to escape any HTML in that string and then pass the resulting escaped string to `JSON.stringify` which would generate a JSON string representation of that string, which is then inserted into the HTML.

The previous behavior would have been to pass the arbitrary data to `JSON.stringify`, which would generate a JSON string representation, which is then inserted into the HTML.

As I understand it the new could would always output `"[object Object]"` if the script returns an object instead of a string. Do I misunderstand what the code is doing?

pluma: Looks like `arangoHelper.escapeHtml` actually doesn't even coerce its argument to a string, it just assumes it's a string and will blow up if it's not something with a `replace` method as all it does is call `str.replace(/&/g, "&#38;").replace(/"/g, "&#34;").replace(/'/g, "&#39;").replace(/</g, "&#60;")` on its argument.

hkernbach: I'll add a stringify method in case of a non-string to the escapeHtml method. Then the behaviour should be correct.

mchacki: Don't we have it now String-escaped twice? (assume the Replace does not replace anything here)

```
> let a = {b: "foo"}
undefined
> c = String(JSON.stringify(a, null, 2)).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;').replace(/'/g, '&#39;');
'{\n  &quot;b&quot;: &quot;foo&quot;\n}'
> d = JSON.stringify(c, null, 2)
'"{\\n  &quot;b&quot;: &quot;foo&quot;\\n}"'
```

now d adds an additional set of "quotes", and re-escapes the line breaks.

I think the output of escapeHTML cannot be a JSON anymore, but only a string, without JSON meaning (quotes are replaced).

mchacki (changes requested):
I am not convinced that this now works as intended see comment

js/apps/system/_admin/aardvark/APP/frontend/js/views/applicationDetailView.js
```diff
@@ -192,7 +192,7 @@
             } else if (result) {
               info = (
                 '<p>Script results:</p>' +
-                '<pre>' + JSON.stringify(result, null, 2) + '</pre>'
+                '<pre>' + arangoHelper.escapeHtml(result) + '</pre>'
```
pluma, line 195: Note that moving the `JSON.stringify` into `escapeHtml` will result in string results being displayed differently than before:

Previously:
```
# script output
module.export = "success"

# generated html
<pre>"success"</pre>
```

Now:
```
# script output
module.export = "success"

# generated html
<pre>success</pre>
```

This results in some ambiguity if the string output looks like JSON, e.g.

```
module.exports = "[]"

# now looks the same as

module.exports = []
```

mchacki: I agree this is a change, but do not really think this is an actual issue.

mchacki (approved):
LGTM.

We have a slight optical change, but I think overall it is better.

PR #12741, opened 2020-09-30
Use dynamically-sized response queue to prevent full queues

danielhlarkin:
### Scope & Purpose

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required
- [x] Backports required for: Forward, devel

### Testing & Verification

- [x] This change is already covered by existing tests, such as java driver concurrent transaction tests

Jenkins: https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/12028/

### DIFF
--- a/arangod/GeneralServer/H2CommTask.cpp
+++ b/arangod/GeneralServer/H2CommTask.cpp
@@ -208,6 +208,7 @@ H2CommTask<T>::H2CommTask(GeneralServer& server, ConnectionInfo info,
     : GeneralCommTask<T>(server, std::move(info), std::move(so)) {
   this->_connectionStatistics.SET_HTTP();
   initNgHttp2Session();
+  _responses.reserve(32);
 }
 
 template <SocketType T>
@@ -580,9 +581,24 @@ void H2CommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> res,
 
   // this uses a fixed capacity queue, push might fail (unlikely, we limit max streams)
   unsigned retries = 512;
-  while (ADB_UNLIKELY(!_responses.push(tmp))) {
-    std::this_thread::yield();
-    if (--retries == 0) {
+  try {
+    while (ADB_UNLIKELY(!_responses.push(tmp))) {
+      std::this_thread::yield();
+      if (--retries == 0) {
+        LOG_TOPIC("924dc", WARN, Logger::REQUESTS)
+            << "was not able to queue response" << (void*)this;
+        // we are overloaded close stream
+        asio_ns::post(this->_protocol->context.io_context,
+                      [self(this->shared_from_this()), mid(res->messageId())] {
+                        auto& me = static_cast<H2CommTask<T>&>(*self);
+                        nghttp2_submit_rst_stream(me._session, NGHTTP2_FLAG_NONE,
+                                                  static_cast<int32_t>(mid),
+                                                  NGHTTP2_ENHANCE_YOUR_CALM);
+                      });
+        return;
+      }
+    }
+    catch (...) {
       LOG_TOPIC("924dc", WARN, Logger::REQUESTS)
           << "was not able to queue response" << (void*)this;
       // we are overloaded close stream
--- a/arangod/GeneralServer/H2CommTask.cpp
+++ b/arangod/GeneralServer/H2CommTask.cpp
@@ -582,35 +582,24 @@ void H2CommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> res,
   // this uses a fixed capacity queue, push might fail (unlikely, we limit max streams)
   unsigned retries = 512;
   try {
-    while (ADB_UNLIKELY(!_responses.push(tmp))) {
+    while (ADB_UNLIKELY(!_responses.push(tmp) && --retries > 0)) {
       std::this_thread::yield();
-      if (--retries == 0) {
-        LOG_TOPIC("924dc", WARN, Logger::REQUESTS)
-            << "was not able to queue response" << (void*)this;
-        // we are overloaded close stream
-        asio_ns::post(this->_protocol->context.io_context,
-                      [self(this->shared_from_this()), mid(res->messageId())] {
-                        auto& me = static_cast<H2CommTask<T>&>(*self);
-                        nghttp2_submit_rst_stream(me._session, NGHTTP2_FLAG_NONE,
-                                                  static_cast<int32_t>(mid),
-                                                  NGHTTP2_ENHANCE_YOUR_CALM);
-                      });
-        return;
-      }
     }
-    catch (...) {
-      LOG_TOPIC("924dd", WARN, Logger::REQUESTS)
-          << "was not able to queue response" << (void*)this;
-      // we are overloaded close stream
-      asio_ns::post(this->_protocol->context.io_context,
-                    [self(this->shared_from_this()), mid(res->messageId())] {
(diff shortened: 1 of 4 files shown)


arangod/GeneralServer/H2CommTask.cpp
```diff
@@ -580,9 +581,24 @@ void H2CommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> res,
...
   // this uses a fixed capacity queue, push might fail (unlikely, we limit max streams)
   unsigned retries = 512;
-  while (ADB_UNLIKELY(!_responses.push(tmp))) {
-    std::this_thread::yield();
-    if (--retries == 0) {
+  try {
+    while (ADB_UNLIKELY(!_responses.push(tmp))) {
+      std::this_thread::yield();
+      if (--retries == 0) {
+        LOG_TOPIC("924dc", WARN, Logger::REQUESTS)
```
jsteemann, line 588: this will fail the Jenkins test as it has the same logID as the log message below.

arangod/GeneralServer/H2CommTask.cpp
```diff
@@ -580,9 +581,24 @@ void H2CommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> res,
 
   // this uses a fixed capacity queue, push might fail (unlikely, we limit max streams)
   unsigned retries = 512;
-  while (ADB_UNLIKELY(!_responses.push(tmp))) {
-    std::this_thread::yield();
-    if (--retries == 0) {
+  try {
```
jsteemann, line 584: We seem to have the same handling for when an exception is thrown and when `retries` reaches zero.
We could thus change the code to:
```
unsigned retries = 512;
try {
  while (ADB_UNLIKELY(!_responses.push(tmp)) && --retries > 0) {
    std::this_thread::yield();
  }
} catch (...) {
  retries = 0;
}
if (retries == 0) {
  LOG_TOPIC("924dc", WARN, Logger::REQUESTS) << "was not able to queue response" << (void*)this;
  // we are overloaded close stream
  asio_ns::post(this->_protocol->context.io_context,
                [self(this->shared_from_this()), mid(res->messageId())] {
                  auto& me = static_cast<H2CommTask<T>&>(*self);
                  nghttp2_submit_rst_stream(me._session, NGHTTP2_FLAG_NONE,
                                            static_cast<int32_t>(mid),
                                            NGHTTP2_ENHANCE_YOUR_CALM);
                });
  return;
}

File: arangod/GeneralServer/H2CommTask.h [hunk 3]
```diff
@@ -113,7 +113,7 @@ class H2CommTask final : public GeneralCommTask<T> {
   velocypack::Buffer<uint8_t> _outbuffer;
 
   // no more than 64 streams allowed
```
jsteemann, line 115:
```suggestion
  // the queue is dynamically sized because we can't guarantee that
  // only a fixed number of responses are active at the same time.
  // the reason is that producing responses may happen faster than
  // fetching responses from the queue. this is done by different threads
  // and depends on thread scheduling, so it is somewhat random how
  // fast producing and consuming happen.
  // effectively the length of the queue is bounded by the fact that the
  // scheduler queue length is also bounded, so that we will not see
  // an endless growth of the queue in a single connection.
```

arangod/GeneralServer/VstCommTask.cpp
```diff
@@ -61,7 +61,9 @@ VstCommTask<T>::VstCommTask(GeneralServer& server, ConnectionInfo info,
       _authToken("", false, 0),
       _authenticated(!this->_auth->isActive()),
       _authMethod(rest::AuthenticationMethod::NONE),
-      _vstVersion(v) {}
+      _vstVersion(v) {
+  _writeQueue.reserve(32);
```
jsteemann, line 65:
```suggestion
  // arbitrary initial reserve value to save a few memory allocations
  // in the most common cases.
  _writeQueue.reserve(32);
```

arangod/GeneralServer/VstCommTask.cpp
```diff
@@ -382,14 +384,21 @@ void VstCommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> baseRes,
...
   // this uses a fixed capacity queue, push might fail (unlikely, we limit max streams)
   unsigned retries = 512;
-  while (ADB_UNLIKELY(!_writeQueue.push(resItem.get()))) {
-    std::this_thread::yield();
-    if (--retries == 0) {
-      LOG_TOPIC("a3bfc", WARN, Logger::REQUESTS)
-          << "was not able to queue response this=" << (void*)this;
-      this->stop();  // stop is thread-safe
-      return;
+  try {
```
jsteemann, line 387: same comment as in H2CommTask.cpp applies here.

arangod/GeneralServer/VstCommTask.h
```diff
@@ -104,7 +104,7 @@ class VstCommTask final : public GeneralCommTask<T> {
 
  private:
   std::map<uint64_t, Message> _messages;
-  boost::lockfree::queue<ResponseItem*, boost::lockfree::capacity<32>> _writeQueue;
```
jsteemann, line 107: same comment as in H2CommTask.h applies here.

graetzer:
Hey are you changing the H2CommTask, there is a limit of active streams enforced in the protocol

graetzer:
Why are you changing the H2CommTask, there is a limit of active streams enforced in the protocol

danielhlarkin:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/12030/

danielhlarkin:
@graetzer So if there's a limit of n streams, it's not allowed to send the (n+1)th request on the connection before getting at least one response back?

jsteemann:
@graetzer : I guess you are referring to this code in H2CommTask.cpp?
````
249 void submitConnectionPreface(nghttp2_session* session) {
250   std::array<nghttp2_settings_entry, 4> iv;
251   // 32 streams matches the queue capacity
252   iv[0] = {NGHTTP2_SETTINGS_MAX_CONCURRENT_STREAMS, 32};
253   // typically client is just a *sink* and just process data as
254   // much as possible.  Use large window size by default.
255   iv[1] = {NGHTTP2_SETTINGS_INITIAL_WINDOW_SIZE, window_size};
256   iv[2] = {NGHTTP2_SETTINGS_MAX_FRAME_SIZE, (1 << 14)};  // 16k
257   iv[3] = {NGHTTP2_SETTINGS_ENABLE_PUSH, 0};
258 
259   nghttp2_submit_settings(session, NGHTTP2_FLAG_NONE, iv.data(), iv.size());
```
So it looks like it is guaranteed that the number of concurrent streams will not exceed 32.
We should however turn the 32 into a constexpr, so it can be used from here and in the queue size declaration.

jsteemann:
@graetzer : can you re-check this PR after my latest push? I hope it addresses your concern about the maximum number of concurrent streams.

graetzer (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12049/

KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-ANY-java-driver-matrix/1171/

KVS85:
Test blue.

KVS85 (approved):

PR #12754, opened 2020-10-02
Fix shard leader behaviour if a follower does not play ball.

neunhoef:
There are two cases in which a shard follower refuses to execute a
synchronous replication request by a shard leader.

One is if the follower has in the meantime taken over leadership for the
shard. The other is if it is simply no longer following that leader, for
example because of a restart or because it was dropped in the meantime
by another thread.

In the first case it is correct behaviour for the leader to run into an
error, because most likely it is no longer actually the leader.
In the second case it is better to drop the follower (even if already
dropped) and move on without an error.

This PR implements fixes this behaviour.
- Fix behaviour of a shard leader if follower does not play ball.
- Improve comment.
- CHANGELOG.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Fixed a bug in handling of followers which refuse to replicate operations.
+  In the case that the follower has simply been dropped in the meantime, we now
+  avoid an error reported by the shard leader.
+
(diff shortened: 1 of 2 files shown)


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12079/

jsteemann (approved):

PR #12760, opened 2020-10-03
avoid usage of potentially unsafe slices() method

jsteemann:
### Scope & Purpose

calling slices() on the `response` member in arangodb::network::Response is potentially unsafe if there are no prior checks, as `response` can be a nullptr.

This is in response to a test failure observed in http://172.16.10.101:8080/job/arangodb-matrix-pr-linux/13716/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=cluster,limit=(linux&&test)%7C%7Cgce/

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/558

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 
- [ ] :book: CHANGELOG entry made
- [ ] :muscle: The behavior in this PR was *manually tested*
- [x] :computer: The behavior change can be verified via automatic tests

#### Backports:

- [ ] No backports required
- [x] Backports required for: 3.7, potentially 3.6

### Testing & Verification

- [x] This change is already covered by existing tests, such as *all networking tests*

PR #12761, opened 2020-10-03
avoid usage of potentially unsafe slices() method

jsteemann:
### Scope & Purpose

calling slices() on the `response` member in arangodb::network::Response is potentially unsafe if there are no prior checks, as `response` can be a nullptr.

This is in response to a test failure observed in http://172.16.10.101:8080/job/arangodb-matrix-pr-linux/13716/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=cluster,limit=(linux&&test)%7C%7Cgce/

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/559
Backport of https://github.com/arangodb/arangodb/pull/12760

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required
- [] Backports required for: potentially 3.6

### Testing & Verification

- [x] This change is already covered by existing tests, such as *all networking tests*

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12096/

### DIFF
--- a/arangod/Cluster/ClusterMethods.cpp
+++ b/arangod/Cluster/ClusterMethods.cpp
@@ -215,6 +215,7 @@ OperationResult handleResponsesFromAllShards(
       result.reset(commError);
       break;
     } else {
+      TRI_ASSERT(res.response);
       std::vector<VPackSlice> const& slices = res.response->slices();
       if (!slices.empty()) {
         VPackSlice answer = slices[0];
--- a/arangod/IResearch/IResearchAnalyzerFeature.cpp
+++ b/arangod/IResearch/IResearchAnalyzerFeature.cpp
@@ -841,7 +841,7 @@ arangodb::Result visitAnalyzers(
         continue;
       }
 
-      if (response.response->statusCode() == arangodb::fuerte::StatusNotFound) {
+      if (response.statusCode() == arangodb::fuerte::StatusNotFound) {
         return { TRI_ERROR_ARANGO_DATA_SOURCE_NOT_FOUND };
       }
 
@@ -849,8 +849,8 @@ arangodb::Result visitAnalyzers(
         return { arangodb::network::fuerteToArangoErrorCode(response) };
       }
 
-      std::vector<VPackSlice> slices = response.response->slices();
-      if (slices.empty() || !slices[0].isObject()) {
+      VPackSlice answer = response.slice();
+      if (!answer.isObject()) {
         return {
           TRI_ERROR_INTERNAL,
           "got misformed result while visiting Analyzer collection'" +
@@ -859,7 +859,6 @@ arangodb::Result visitAnalyzers(
         };
       }
 
-      VPackSlice answer = slices[0];
       auto result = arangodb::network::resultFromBody(answer, TRI_ERROR_NO_ERROR);
       if (result.fail()) {
         return result;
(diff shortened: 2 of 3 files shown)


arangod/Cluster/ClusterMethods.cpp
```diff
@@ -203,29 +203,27 @@ OperationResult handleResponsesFromAllShards(
...
+
   pre(result, builder);
-  if (result.fail()) {
-    return OperationResult(result, builder.steal());
-  }
-  for (Try<arangodb::network::Response> const& tryRes : responses) {
-    network::Response const& res = tryRes.get();  // throws exceptions upwards
-    ShardID sId = res.destinationShard();
-    int commError = network::fuerteToArangoErrorCode(res);
-    if (commError != TRI_ERROR_NO_ERROR) {
```
graetzer, line 214: so this is really a waste of everones time. if `res.error == fuerte::Error::NoError` then `res.response != nullptr`

jsteemann: Except that it crashed there

graetzer: thats a bug then which is nicely covered up now

jsteemann: I can't see any guarantee in the code that `response` is not a nullptr if `error` is `NoError`. 
Creating a new netowrk::Response object has a nullptr `response` value, and an undefined `error` value.

jsteemann: In fact, the existing code suggests this is **not** always true, for example:
```
 54   // returns a slice of the payload if there was no error
 55   [[nodiscard]] velocypack::Slice slice() const {
 56     if (error == fuerte::Error::NoError && response) {
 57       return response->slice();
 58     }
 59     return velocypack::Slice();  // none slice
 60   } 
```
Why would someone check if `response` is set here if it were guaranteed that in case of `error == NoError` it is set?

graetzer: i guess it was a precaution. however getting NoError withou a response does not make much sense

jsteemann: That may be the desired behavior. But it's not guaranteed right now.
Just look at this, for example:
```
1302 struct AsyncRequest {
1303   network::Response response;
1304   std::string destination;
1305   CoordTransactionID coordTransactionID;
1306   OperationID operationID;
1307   bool done = false;
1308 };

...
1553   auto ar = std::make_shared<::AsyncRequest>();
1554   ar->destination = destination;
1555   ar->operationID = opId;
1556   ar->coordTransactionID = coordTransactionID;
1557 
1558   network::sendRequest(pool, destination, reqType, path, std::move(body), reqOpts)
1559   .thenValue([ar] (network::Response&& r) {
1560     {
1561       std::lock_guard<std::mutex> guard(::_requestMutex);
1562       ar->response = std::move(r);
1563       ar->done = true;
1564     }
1565     ::_requestCV.notify_all();
1566   });
```

graetzer: so what thats different code

jsteemann: Ìt's from arangod/Cluster/v8-cluster.cpp
It defines a struct AsyncRequest with a default-constructed `network::Response` member.
As network::Response doesn't have any user defined constructors, its members will be default-initialized.
`error` is a uint16_t, so it will have an arbitrary value, while `response` will be properly constructed as a unique_ptr pointing to 0.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12295/

graetzer (approved):

KVS85:
Tests blue.

PR #12766, opened 2020-10-03
Make sure response is set if no error is indicated

graetzer:
### Scope & Purpose

Make sure to never have a null request member and actually set the response if `response.error == NoError`.
Additionally we now avoid recreating the request object everytime

- [x] :hankey: Bugfix 

#### Backports:

- [ ] No backports required

### Testing & Verification

http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12101/

### DIFF
--- a/arangod/Network/Methods.cpp
+++ b/arangod/Network/Methods.cpp
@@ -234,7 +234,7 @@ class RequestsState final : public std::enable_shared_from_this<RequestsState> {
         _startTime(std::chrono::steady_clock::now()),
         _endTime(_startTime + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
                                   options.timeout)) {
-    _tmp_req = prepareRequest(type, path, payload, _options, headers);
+    _tmp_req = prepareRequest(type, std::move(path), std::move(payload), _options, std::move(headers));
   }
 
   ~RequestsState() = default;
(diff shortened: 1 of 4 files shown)


arangod/Network/Methods.cpp
```diff
@@ -178,148 +176,149 @@ FutureRes sendRequest(ConnectionPool* pool, DestinationId dest, RestVerb type,
...
-        _headers(std::move(headers)),
+      : _destination(std::move(destination)),
         _options(options),
         _pool(pool),
-        _type(type),
         _startTime(std::chrono::steady_clock::now()),
         _endTime(_startTime + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
-                                  options.timeout)) {}
+                                  options.timeout)) {
+    _tmp_req = prepareRequest(type, path, payload, _options, headers);
```
jsteemann, line 237: I think we can use std::move here, because `prepareRequest` expects these parameters by value
```suggestion
    _tmp_req = prepareRequest(type, std::move(path), std::move(payload), _options, std::move(headers));
```

jsteemann (approved):
LGTM

jsteemann:
Can we also assign a default `error` code to `network::Response`, at prevent it from being default-constructed?

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12112/

jsteemann (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12119/

PR #12779, opened 2020-10-05
Refactor expression callstyle

graetzer:
### Scope & Purpose

Refactor the callstyle of AQL functions, this is in preparation for time series

- [x] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12121/

hopefully all the functions are properly covered already


gnusi (changes requested):
there are 2 issues from my POV:
1) it's impossible to simplify an `Expression` by extracting a function pointer anymore, we always need a useless argument which is bounded to `Ast` which may not be present in general.
2) currently there are no functions using this argument and it set to `nullptr` in tests. Does it mean all functions have to check it for `nullptr`?

danielhlarkin (changes requested):

graetzer:
> there are 2 issues from my POV:
> 
>     1. it's impossible to simplify an `Expression` by extracting a function pointer anymore, we always need a useless argument which is bounded to `Ast` which may not be present in general.
> 
>     2. currently there are no functions using `AstNode const*` argument and it set to `nullptr` in tests. Does it mean all functions have to check it for `nullptr`?
1. You can still do that without any trouble. The AstNode*, this is just the node the ptr is attached to.
2. The tests are just for functions that do not use the AstNode* so it might as well be nullptr

In general it is not nullptr just like the ExpressionContext is generally never a nullptr

gnusi:
1. I can't because usually I need `Ast` to allocate a certain `AstNode`. There are some cases where we don't need that, e.g. computed attributes stuff. 
2. I believe we have to define our contract first, e.g. normally we don't expect transaction is `nullptr` when calling a function. Same must be done for `AstNode` argument otherwise that makes hard to use it correctly.

Why do we need this `AstNode` at all as a part of function signature given there are no functions using it?

graetzer:
its intended for future use, thats why its a refactoring. I still don't understand your problem with 1. the function is part of the Ast already as a node with type `NODE_TYPE_FCALL`. We just pass in the node ptr 
2. no not really, would be nicer to use a reference but not harder to use at all

arangod/Aql/Function.h
```diff
@@ -55,7 +55,10 @@ struct Function {
 
     /// @brief exclude the function from being evaluated during AST
     /// optimizations evaluation of function will only happen at query runtime
-    NoEval = 8
+    NoEval = 8,
+    
+    /// @brief is this function position dependent
+    NonMoveable = 16
```
jsteemann, line 61: Right now, the new enum value is not used anywhere. I understand this may be used in the future, but I suggest we introduce it only when it is actually being used.
```suggestion
```

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12140/

jsteemann (approved):

jsteemann:
@gnusi : from within an AQL Expression that calls into an AQL function, the `AstNode const*` parameter will be set correctly on every function call invocation.
We currently don't need this parameter, but in the future it will allow us to build function-local caches, even if the same AQL function is used multiple times in the same expression.
One idea is to use the AstNode pointer as a key/identifier in some cache object.
Currently there is no harm when setting the AstNode pointer to a nullptr when calling into any AQL function, but I agree we should define whether setting this to a valid pointer is now required or still optional.

gnusi:
it's harder to use because presence of underlying `AstNode` in a signature allows me to access a function name and avoid having nasty function name duplication, like in many functions from `Functions.cpp`. But if I do so, tests will just fail.
If `AstNode` is just needed for some caching don't you consider to make it a part of `ExpressionContext`, e.g. `AstNode const* ExpressionContext::cacheKey()`? In general I tend to agree that accessing an `AstNode` from a function implementation is nice, but let's either make it mandatory (i.e. use reference) or move it to `ExpressionContext` as an optional thing.

graetzer:
I cannot parse your text. Are gonna be happy if `AstNode const*` -> `AstNode const&` ?

I  mean I **can** invest the 2h changing it if it gives you pleasure

gnusi:
@graetzer yes, let's change it to references :-)

graetzer:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/12145/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12278/

jsteemann:
Tests blue

PR #12792, opened 2020-10-06
don't respond with misleading error in smart vertex collections

jsteemann:
### Scope & Purpose

When inserting a document with a non-conforming key pattern into a smart vertex collection, the response error code and message are 1466 (ERROR_CLUSTER_MUST_NOT_SPECIFY_KEY) and "must not specify _key
for this collection".
This is misleading, because it is actually allowed to specify a key value for documents in such collection. However, there are some
restrictions for valid key values (e.g. the key must be a string and contain the smart graph attribute value at the front, followed by a
colon.
If any of these restrictions are not met, the server currently responds with "must not specify key for this collection", which is
misleading. This PR rectifies it so that the server responds with error 4003 (ERROR_KEY_MUST_BE_PREFIXED_WITH_SMART_GRAPH_ATTRIBUTE) and message "in smart vertex collections _key must be a string and prefixed with the value of the smart graph attribute". This should make it a lot easier to understand what the actual problem is.

This is an error code change, so it will not be backported.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/560

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_server*.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,25 @@
 devel
 -----
 
+* Don't respond with misleading error in smart vertex collections.
+
+  When inserting a document with a non-conforming key pattern into
+  a smart vertex collection, the response error code and message are
+  1466 (ERROR_CLUSTER_MUST_NOT_SPECIFY_KEY) and "must not specify _key
+  for this collection".
+  This is misleading, because it is actually allowed to specify a key
+  value for documents in such collection. However, there are some
+  restrictions for valid key values (e.g. the key must be a string and
+  contain the smart graph attribute value at the front, followed by a
+  colon.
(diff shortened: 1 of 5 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12261/

dothebart (approved):

jsteemann:
Tests blue

mchacki (approved):
LGTM

PR #12793, opened 2020-10-07
Added optional verbose logging for agency write operations

jsteemann:
### Scope & Purpose

Added optional verbose logging for agency write operations.
This logging is configurable by using the new log topic "agencystore".

The following log levels can be used for for the "agencystore" log topic to log writes to the agency:
* DEBUG: will log all writes on the leader
+ TRACE: will log all writes on both leaders and followers
The default log level for the "agencystore" log topic is WARN, meaning no agency writes will be logged.
Turning on this logging can be used for auditing and debugging, but it is not recommended in the general case, as it can lead to large amounts of data being logged, which can have a performance impact and will lead to higher disk space usage.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [ ] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12164/

### DIFF
--- a/arangod/Agency/State.cpp
+++ b/arangod/Agency/State.cpp
@@ -293,10 +293,21 @@ index_t State::logNonBlocking(index_t idx, velocypack::Slice const& slice,
                               term_t term, uint64_t millis, std::string const& clientId,
                               bool leading, bool reconfiguration) {
   _logLock.assertLockedByCurrentThread();
-
-  auto byteSize = slice.byteSize();
-  auto buf = std::make_shared<Buffer<uint8_t>>();
-  buf->append((char const*)slice.begin(), byteSize);
+  
+  // verbose logging for all agency operations
+  // there are two different log levels in use here for the AGENCYSTORE topic
+  // - DEBUG: will log writes only on the leader
+  // - TRACE: will log writes on both leaders and followers
+  // the default log level for the AGENCYSTORE topic is WARN
+  if (leading) {
+    LOG_TOPIC("b578f", DEBUG, Logger::AGENCYSTORE)
+        << "leader: true, client: " << clientId << ", index: " << idx << ", term: " << term
+        << ", data: " << slice.toJson();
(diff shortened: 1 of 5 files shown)


arangod/Agency/State.cpp
```diff
@@ -293,10 +293,21 @@ index_t State::logNonBlocking(index_t idx, velocypack::Slice const& slice,
                               term_t term, uint64_t millis, std::string const& clientId,
                               bool leading, bool reconfiguration) {
   _logLock.assertLockedByCurrentThread();
-
-  auto byteSize = slice.byteSize();
-  auto buf = std::make_shared<Buffer<uint8_t>>();
-  buf->append((char const*)slice.begin(), byteSize);
+
```
neunhoef, line 296: As far as I see this is - contrary to the comment above it - called for both the leader and the follower log appending. This means we would log it 3 times, if we do the logging here. There is something to be said for this, since when the leader appends something to its log, we don't know if it will stay in the Raft log, before it is replicated to a majority of followers. On the other hand, it uses three times as much disk space as needed.
Furthermore, I think at this stage the preconditions are already fulfilled and stripped. In some circumstances it could be useful to also log failed attempts to write something to the raft log. For me it is unclear what the best way to proceed here is.

neunhoef (approved):
LGTM but see comment.

arangod/Agency/State.cpp
jsteemann: As far as I can see setting the log level to `DEBUG` will make it log only on the leader, but not on followers. At least I am checking the `leading` flag here, and if it is false (i.e. we are on a follower), we log only if log level is `TRACE`.
That means one should be able to run it _only_ on leaders, if so desired.

But I am not an expert here, and if there is a better place to do the logging, I would be happy to know it!

PR #12811, opened 2020-10-09
If we fail to switch the database, flush the cache too.

dothebart:
### Scope & Purpose
The client (arangosh) needs to re-assure with the server what the current state
is after the error. A cache must not stop that from happening - hence
flush it.

This miss-behaviour was discovered while using arangosh for load-testing. 

- [x] :hankey: Bugfix 
#### Backports:
- [ ] No backports required


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12217/

dothebart:
tests green.

jsteemann (approved):

Issue #12649, opened 2020-09-14
Arangoimport does not complain for not used commandline arguments


meetme2meat:
in this commandline:

`arangoimport --threads 4 --type json --create-collection true --collection records --server.password  --server.database cdrs --server.username root --server.endpoint 'http+ssl://6e505c53b91a.arangodb.cloud:18529' --server.password myserverpassword  --create-database true --file /Users/admin/Documents/goProject/src/mongo_import/NV100827.cdr.json`
one `--` argument is to much, and hence `cdrs` becomes an unused argument, since arangoim doesn't use it at all. It shold exit & complain if this happenes. 

## My Environment

architecture: 64bit
arm: false
asan: false
asm-crc32: true
assertions: false
avx2: false
boost-version: 1.71.0
build-date: 2020-08-28 11:05:28
compiler: clang [4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.62)]
cplusplus: 201703
curl-version: none
debug: false
endianness: little
failure-tests: false
fd-client-event-handler: poll
fd-setsize: 1024
full-version-string: ArangoDB 3.7.2 [darwin] 64bit, using build , VPack 0.1.33, RocksDB 6.8.0, ICU 64.2, V8 7.9.317, OpenSSL 1.1.1g  21 Apr 2020
icu-version: 64.2
ipo: true
iresearch-version: 1.0.0.0
jemalloc: false
libunwind: false
license: community
maintainer-mode: false
ndebug: true
openssl-version-compile-time: OpenSSL 1.1.1g  21 Apr 2020
openssl-version-run-time: OpenSSL 1.1.1g  21 Apr 2020
optimization-flags: -march=nehalem -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mno-sse4a -mno-avx -mno-fma -mno-bmi2 -mno-avx2 -mno-xop -mno-fma4 -mno-avx512f -mno-avx512vl -mno-avx512pf -mno-avx512er -mno-avx512cd -mno-avx512dq -mno-avx512bw -mno-avx512ifma -mno-avx512vbmi
pic: 2
pie: none
platform: darwin
reactor-type: kqueue
rocksdb-version: 6.8.0
server-version: 3.7.2
sizeof int: 4
sizeof long: 8
sizeof void*: 8
sse42: true
tsan: false
unaligned-access: true
v8-version: 7.9.317
vpack-version: 0.1.33
zlib-version: 1.2.11

[...]

PR #12816, opened 2020-10-09
better validate commandline options

dothebart:
- validate whether the import file exists early so we can error out
  before attempting to connect the server
- if we have a `--file` argument don't swallow unused option.

this improves the useability as seen in https://github.com/arangodb/arangodb/issues/12649


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12227/

jsteemann (approved):

PR #12827, opened 2020-10-09
Feature/calulation analyzer

Dronplane:
### Scope & Purpose
This PR adds  new type of analyzer 'aql' capable of running arbitrary (with few restrictions) AQL query and putting this particular query result into ArangoSearch index. Input data (currently analyzed document field value) wired to the query via @param  parameter. Query is running in a sandbox vocbase without any data. Optimizer also does not work with this queries.

Query restrictions are set to make query self-contained and runnable on DBServers (as in cluster indexing happens on DBServers)
 - No datasource access (collections/views)
 - No DML
 - No TRAVERSAL
 - No COLLECT 
 - All used AQL functions should be DBServer runnable
 - No V8
 - No UDF
 - No other analyzer access (to make it possible to run during recovery)

- [x] :pizza: New feature 

#### Backports:

- [x] No backports required

#### Related Information
Internal ticket: https://github.com/arangodb/backlog/issues/761
Also covers:
https://github.com/arangodb/backlog/issues/195
https://github.com/arangodb/backlog/issues/630
https://github.com/arangodb/backlog/issues/765
https://github.com/arangodb/backlog/issues/766

### Testing & Verification

http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12445/

### Documentation
Docs PR: https://github.com/arangodb/docs/pull/568


arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,507 @@
...
+    // TODO: call engine refill!
+    while (_queryResults->numRows() > _resultRowIdx) {
+      AqlValue const& value = _queryResults->getValueReference(_resultRowIdx++, _engine->resultRegister());
+      if (value.isString()) {
+        arangodb::velocypack::Builder builder;
+        value.toVelocyPack(&arangodb::velocypack::Options::Defaults, builder, true, false);
+        _str = builder.slice().copyString();
+        _term.value = irs::ref_cast<irs::byte_type>(irs::string_ref(_str));
+        return true;
+      }
```
gnusi, line 448: `#include "IResearch/AqlHelper.h"`
+
```suggestion
if (value.isString()) {
  _term.value = arangodb::iresearch::getStringRef(value);
}
```

arangod/IResearch/IResearchCalculationAnalyzer.h
```diff
@@ -0,0 +1,109 @@
...
+
+  std::unique_ptr<arangodb::aql::Ast> _ast;
+  std::unique_ptr<arangodb::aql::ExecutionEngine> _engine;
+  std::unique_ptr<CalculationQueryContext> _query;
+  std::unique_ptr<arangodb::aql::ExecutionPlan> _plan;
+  arangodb::aql::SharedAqlItemBlockPtr _queryResults;
+  size_t _resultRowIdx{ 0 };
+  std::vector<arangodb::aql::AstNode*> _bindedNodes;
+  /// @brief Artificial vocbase for executing calculation queries
+  static std::unique_ptr<TRI_vocbase_t> CalculationAnalyzer::_calculationVocbase;
```
gnusi, line 104: I'd prefer to move it into local namespace in `cpp`

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,507 @@
...
+      });
+    _plan = ExecutionPlan::instantiateFromAst(_ast.get());
+  } else {
+    for (auto node : _bindedNodes) {
+      node->setStringValue(field.c_str(), field.size());
+    }
+  }
+
+  _engine = std::make_unique<ExecutionEngine>(0, *_query, _query->itemBlockManager(),
```
gnusi, line 486: it seems that we can avoid using `unique_ptr` as `QueryCalculationContext` can be already initialized.

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+bool CalculationAnalyzer::next() {
+  do {
+    if (_queryResults.get()) {
+      while (_queryResults->numRows() > _resultRowIdx) {
+        AqlValue const& value =
+            _queryResults->getValueReference(_resultRowIdx++, _engine.resultRegister());
+        if (value.isString() || (value.isNull(true) && _options.keepNull)) {
+          if (value.isString()) {
+            _term.value = irs::ref_cast<irs::byte_type>(arangodb::iresearch::getStringRef(value.slice()));
```
gnusi, line 288:
```suggestion
            _term.value = arangodb::iresearch::getBytesRef(value.slice()));
```

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+    return true;
+  }
+  return false;
+}
+} // namespace
+
+namespace arangodb {
+namespace iresearch {
+
+class CalculationQueryContext: public QueryContext {
```
gnusi, line 133: `final`?

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+  virtual bool killed() const override { return false; }
+
+  /// @brief whether or not a query is a modification query
+  virtual bool isModificationQuery() const noexcept override { return false; }
+
+  virtual bool isAsyncQuery() const noexcept override { return false; }
+
+  virtual void enterV8Context() override { TRI_ASSERT(FALSE); }
```
gnusi, line 179: maybe `throw` not implemented?

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+      _state(vocbase) {}
+
+/// @brief get transaction state, determine commit responsiblity
+std::shared_ptr<TransactionState> CalculationAnalyzer::CalculationTransactionContext::acquireState(
+    transaction::Options const& options, bool& responsibleForCommit) {
+  return std::shared_ptr<TransactionState>(std::shared_ptr<TransactionState>(), &_state);
+}
+
+std::shared_ptr<arangodb::transaction::Context> CalculationAnalyzer::CalculationTransactionContext::clone() const {
+  TRI_ASSERT(FALSE);
```
gnusi, line 481: maybe `throw` not implemented?

arangod/IResearch/IResearchCalculationAnalyzer.h
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  virtual bool next() override;
+  virtual bool reset(irs::string_ref const& field) noexcept override;
+
+ private:
+
+  /// @brief Dummmy transaction state which doesn nothing but provides valid statuses
+  /// to keep ASSERT happy
+  class CalculationTransactionState final : public arangodb::TransactionState {
```
gnusi, line 97: can we move it to `cpp`?

arangod/IResearch/IResearchCalculationAnalyzer.h
```diff
@@ -0,0 +1,216 @@
...
+
+    /// @brief number of commits, including intermediate commits
+    uint64_t numCommits() const override { return 0; }
+
+   private:
+    transaction::Options _options;
+  };
+
+  /// @brief Dummy transaction context which just gives dummy state
+  struct CalculationTransactionContext final : public arangodb::transaction::SmartContext {
```
gnusi, line 138: can we move it to `cpp`?

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+    _engine(0, _query, _query.itemBlockManager(),
+            SerializationFormat::SHADOWROWS, nullptr) {
+  TRI_ASSERT(validateQuery(_options.queryString,
+                           arangodb::DatabaseFeature::getCalculationVocbase())
+                 .ok());
+}
+
+bool CalculationAnalyzer::next() {
+  do {
+    if (_queryResults.get()) {
```
gnusi, line 354:
```suggestion
    if (nullptr != _queryResults) {
```

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+          _inc.value = _nextIncVal;
+          _nextIncVal = !_options.collapseArrayPositions;
+          return true;
+        }
+      }
+    }
+    if (_executionState == ExecutionState::HASMORE) {
+      try {
+        _executionState = ExecutionState::DONE; // set to done to terminate in case of exception
+        _resultRowIdx = 0;
```
gnusi, line 373: can we move it outside `try/catch` block?

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+            << "error executing calculation query: " << e.message();
+      } catch (std::exception const& e) {
+        LOG_TOPIC("c92eb", ERR, iresearch::TOPIC)
+            << "error executing calculation query: " << e.what();
+      } catch (...) {
+        LOG_TOPIC("bf89b", ERR, iresearch::TOPIC)
+            << "error executing calculation query";
+      }
+    }
+  } while (_executionState != ExecutionState::DONE || (_queryResults.get() &&
```
gnusi, line 388:
```suggestion
  } while (_executionState != ExecutionState::DONE || (nullptr != _queryResults &&
```

arangod/IResearch/IResearchCalculationAnalyzer.cpp
```diff
@@ -0,0 +1,486 @@
...
+bool CalculationAnalyzer::reset(irs::string_ref const& field) noexcept {
+  try {
+    if (!_plan) {  // lazy initialization
+      // important to hold a copy here as parser accepts reference!
+      auto queryString = arangodb::aql::QueryString(_options.queryString);
+      auto ast = _query.ast();
+      TRI_ASSERT(ast);
+      Parser parser(_query, *ast, queryString);
+      parser.parse();
+      AstNode* astRoot = const_cast<AstNode*>(ast->root());
```
gnusi, line 402: shall we add something like `AstNode* Ast::mutableRoot()`?

arangod/Aql/QueryContext.cpp
```diff
@@ -56,15 +56,18 @@ QueryContext::QueryContext(TRI_vocbase_t& vocbase)
       _vocbase(vocbase),
       _execState(QueryExecutionState::ValueType::INVALID_STATE),
       _numRequests(0) {
-  if (!AqlFeature::lease()) {
+  if (_vocbase.id() != std::numeric_limits<uint64_t>::max() && // FIXME: find better way to check
```
gnusi, line 59: consider
```
if (&_vocbase() != &server.getFeature<DatabaseFeature>().getCalculationVocbase()) 
```

arangod/Aql/ExecutionEngine.cpp
```diff
@@ -716,6 +726,15 @@ ExecutionBlock* ExecutionEngine::addBlock(std::unique_ptr<ExecutionBlock> block)
   return block.release();
 }
 
+void arangodb::aql::ExecutionEngine::reset()
+{
```
jsteemann, line 730:
```suggestion
void arangodb::aql::ExecutionEngine::reset() {
```

arangod/Aql/QueryContext.cpp
```diff
@@ -56,15 +57,18 @@ QueryContext::QueryContext(TRI_vocbase_t& vocbase)
       _vocbase(vocbase),
       _execState(QueryExecutionState::ValueType::INVALID_STATE),
       _numRequests(0) {
-  if (!AqlFeature::lease()) {
+  if (&_vocbase != &_vocbase.server().getFeature<DatabaseFeature>().getCalculationVocbase() &&
```
jsteemann, line 60: Is this a performance optimization only, or are there any other reasons to allow calculations to be executed during shutdown? This is unclear from the code alone, so we should add a code comment here.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+  /// @brief get transaction state, determine commit responsiblity
+  std::shared_ptr<arangodb::TransactionState> acquireState(arangodb::transaction::Options const& options,
+                                                 bool& responsibleForCommit) override {
+    return std::shared_ptr<arangodb::TransactionState>(std::shared_ptr<arangodb::TransactionState>(),
+                                                       &_state);
+  }
+
+  /// @brief unregister the transaction
+  void unregisterTransaction() noexcept override{};
```
jsteemann, line 183:
```suggestion
  void unregisterTransaction() noexcept override {}
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+  VPackBuilder builder;
+  if (normalize_slice(src->slice(), builder)) {
+    out = builder.toString();
+    return true;
+  }
+  return false;
+}
+
+/*static*/ irs::analysis::analyzer::ptr AqlAnalyzer::make_vpack(irs::string_ref const& args) {
+  Options options;
```
jsteemann, line 393: Options don't seem to be used here.
```suggestion
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+  auto const slice = arangodb::iresearch::slice(args);
+  return make_slice(slice);
+}
+
+
+/*static*/ irs::analysis::analyzer::ptr AqlAnalyzer::make_json(irs::string_ref const& args) {
+  Options options;
```
jsteemann, line 400: Options don't seem to be used here.
```suggestion
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+              }
+            } break;
+            case arangodb::aql::NODE_TYPE_COLLECT: // COLLECT nodes requires optimizer rule to work properly
+            case arangodb::aql::NODE_TYPE_COLLECT_COUNT:
+              errorMessage = "COLLECT is forbidden for calculation analyzer";
+              return false;
+            case arangodb::aql::NODE_TYPE_PARAMETER: {
+              irs::string_ref parameterName(node->getStringValue(), node->getStringLength());
+              if (parameterName != CALCULATION_PARAMETER_NAME) {
+                errorMessage = "Invalid parameter found '";
```
jsteemann, line 309:
```suggestion
                errorMessage = "Invalid bind parameter found '";
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // no traversal (also data access is forbidden) but to give mo clear error message
+    if (ast->containsTraversal()) {
+      return {TRI_ERROR_BAD_PARAMETER,
+              "Traversal usage is forbidden for calculation analyzer"};
+    }
+
+    std::string errorMessage;
+    // Forbid to use functions that reference analyzers -> problems on recovery as analyzers are not available for querying.
+    // Forbid all non-Dbserver runable functions as it is not available on DBServers where analyzers run.
+    arangodb::aql::Ast::traverseReadOnly(
```
jsteemann, line 289: It is unclear to me why an "aql" analyzer so much AQL parts to be used here, e.g. LIMIT, FOR loops, SORT etc.
I don't think we should allow this, because it may lead to users doing very unexpected things with it.
Can't we strict the "aql" analyzer capabilties to AQL expressions, which are initiated by a RETURN statement, e.g. `RETURN @param + 1` or `RETURN SOUNDEX(REVERSE(@param))` or such?
That is IMHO the only way to effectively limit what users can do with the analyzer. Leaving it all open and then adding a few checks here works now, but in the future we may forget to add restrictions to the "aql" analyzer when we add new features.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+        ast->root(), [&errorMessage](arangodb::aql::AstNode const* node) -> bool {
+          switch (node->type) {
+            case arangodb::aql::NODE_TYPE_FCALL: {
```
jsteemann, line 292: It seems we are currently allowing subqueries in "aql" analyzers, which I don't think we should do.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+              auto func = static_cast<arangodb::aql::Function*>(node->getData());
+              if (!func->hasFlag(arangodb::aql::Function::Flags::CanRunOnDBServer) ||
+                  forbiddenFunctions.find(func->name) != forbiddenFunctions.end()) {
+                errorMessage = "Function '";
+                errorMessage.append(func->name)
+                    .append("' is forbidden for calculation analyzer");
```
jsteemann, line 298: Several messages refer to a "calculation" analyzer, but in other places the analyzer is named "aql" analyzer.
I suggest we should be consistent with the naming.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,516 @@
...
+    }
+    if (_executionState == ExecutionState::HASMORE) {
+      _executionState = ExecutionState::DONE;  // set to done to terminate in case of exception
+      _resultRowIdx = 0;
+      _queryResults = nullptr;
+      try {
+        std::tie(_executionState, _queryResults) = _engine.getSome(_options.batchSize);
+        TRI_ASSERT(_executionState != ExecutionState::WAITING);
+      } catch (basics::Exception const& e) {
+        LOG_TOPIC("b0026", ERR, iresearch::TOPIC)
```
jsteemann, line 443: IMHO in user-defined queries we should not log errors to the logfile, but return the error to the caller, so the caller can handle it.
Otherwise the DBAs may see logfile errors but will not have any context about which query triggered it.

jsteemann (changes requested):

arangod/IResearch/IResearchAqlAnalyzer.cpp
Dronplane: Good point. Initially working name was "calculation" but than it was decided to switch to "aql". But auto-replace seems to fail somewhere 🤦

arangod/IResearch/IResearchAqlAnalyzer.cpp
Dronplane: Meaningful error will be generated during query validation phase. But if query passed validation and later this broke - we for sure should notify DBAs. And the context is here: "error executing calculation query" :) So log reader will know for sure - error is related to analyzers. I`ll add query string to log to make message more specific.

arangod/IResearch/IResearchAqlAnalyzer.cpp
Dronplane: Why not after all ? Data access is forbidden but why to forbid calculating some filtered average for example ?

arangod/Aql/QueryContext.cpp
Dronplane: added comments to make intentions clear

arangod/IResearch/IResearchAqlAnalyzer.cpp
Dronplane: Ok, I`ll switch validation to 'whitelist' mode - all nodes except listed are forbidden.

Dronplane:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12455/

arangod/Aql/QueryContext.cpp
```diff
@@ -56,15 +57,20 @@ QueryContext::QueryContext(TRI_vocbase_t& vocbase)
       _vocbase(vocbase),
       _execState(QueryExecutionState::ValueType::INVALID_STATE),
       _numRequests(0) {
-  if (!AqlFeature::lease()) {
+  // aql analyzers should be able to run even during recovery when aql deature
```
jsteemann, line 60:
```suggestion
  // aql analyzers should be able to run even during recovery when AqlFeature
```

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -73,6 +73,18 @@ using namespace arangodb::application_features;
 using namespace arangodb::basics;
 using namespace arangodb::options;
 
+namespace {
+arangodb::CreateDatabaseInfo createExpressionVocbaseInfo(arangodb::application_features::ApplicationServer& server) {
+  arangodb::CreateDatabaseInfo info(server, arangodb::ExecContext::current());
+  auto rv = info.load("Z", std::numeric_limits<uint64_t>::max());
```
jsteemann, line 79: Why is this database named "Z"?
I guess it has to do with sort order, but there could still be other databases sorted after "Z"...? :thinking:

Dronplane: Name is actually does not matter.  Database even is not added to database list. So sorting is also not relevant. I`ll add a comment.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+            << "error executing calculation query: " << e.message()
+            << " AQL query:" << _options.queryString;
+      } catch (std::exception const& e) {
+        LOG_TOPIC("c92eb", ERR, iresearch::TOPIC)
+            << "error executing calculation query: " << e.what()
+            << " AQL query:" << _options.queryString;
+      } catch (...) {
+        LOG_TOPIC("bf89b", ERR, iresearch::TOPIC)
+            << "error executing calculation query"
+            << " AQL query:" << _options.queryString;
```
jsteemann, line 493:
```suggestion
            << " AQL query: " << _options.queryString;
      } catch (std::exception const& e) {
        LOG_TOPIC("c92eb", ERR, iresearch::TOPIC)
            << "error executing calculation query: " << e.what()
            << " AQL query: " << _options.queryString;
      } catch (...) {
        LOG_TOPIC("bf89b", ERR, iresearch::TOPIC)
            << "error executing calculation query"
            << " AQL query: " << _options.queryString;
```

jsteemann: Ok, if actual validation happens before and the user gets a proper error returned, then it is fine.

arangod/IResearch/IResearchAqlAnalyzer.cpp
jsteemann: Should we allow all of this? I would favor restricting things to just `RETURN expression`, where `expression` would also be very limited. That would allow us to store the expression in an Aql `Expression` object, without the need to create execution plans etc.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+    // no traversal (also data access is forbidden) but to give mo clear error message
+    if (ast->containsTraversal()) {
+      return {TRI_ERROR_BAD_PARAMETER,
+              "Traversal usage is forbidden for aql analyzer"};
+    }
+
+    std::string errorMessage;
+    // Forbid to use functions that reference analyzers -> problems on recovery as analyzers are not available for querying.
+    // Forbid all non-Dbserver runable functions as it is not available on DBServers where analyzers run.
```
jsteemann, line 288:
```suggestion
    // Forbid all non-Dbserver runnable functions as it is not available on DBServers where analyzers run.
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
jsteemann: Yes, for a filtered average, you will need either subqueries or `[* FILTER ...]`.
But this means you will also get the whole complexity of ExecutionPlans and ExecutionNodes, which you probably don't need by limiting it to `Expression`s. Anyway, if this is a requirement, so be it.

jsteemann (approved):

Dronplane:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12460/

arangod/IResearch/IResearchAnalyzerFeature.cpp
```diff
@@ -2461,7 +2464,10 @@ Result IResearchAnalyzerFeature::loadAnalyzers(
       auto res = emplaceAnalyzer(result, analyzers, normalizedName, type, properties, features, revision);
 
       if (!res.ok()) {
-        return res; // caught error emplacing analyzer (abort further processing)
+        LOG_TOPIC("7cc7f", ERR, arangodb::iresearch::TOPIC)
+            << "analyzer '" << name
+            << "' ignored as emplace failed with reason:" << res.errorMessage();
+        return {};  // caught error emplacing analyzer  - skip it
```
mchacki, line 2470: Is this change intended?
On the left you have the outer place the chance to see that something goes wrong here.
Now you log an error, but report Success.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+#include <Containers/HashSet.h>
+#include "VPackDeserializer/deserializer.h"
+#include <frozen/set.h>
+
+namespace {
+using namespace arangodb::velocypack::deserializer;
+using namespace arangodb::aql;
+
+constexpr const char QUERY_STRING_PARAM_NAME[] = "queryString";
```
mchacki, line 61: We typically place those in lib/basics/StaticStrings.h|.cpp

Happy that you use this :+1:

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+          std::string("Value of '")
+              .append(BATCH_SIZE_PARAM_NAME)
+              .append("' should be less or equal to ")
+              .append(std::to_string(MAX_BATCH_SIZE))};
+    }
+    return {};
+  }
+};
+
+using OptionsDeserializer = utilities::constructing_deserializer<Options, parameter_list<
```
mchacki, line 92: \o/

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+              "V8 usage is forbidden for aql analyzer"};
+    }
+
+    // no modification (as data access is forbidden) but to give more clear error message
+    if (ast->containsModificationNode()) {
+      return {TRI_ERROR_BAD_PARAMETER,
+              "DML is forbidden for aql analyzer"};
+    }
+
+    // no traversal (also data access is forbidden) but to give mo clear error message
```
mchacki, line 280:
```suggestion
    // no traversal (also data access is forbidden) but to give more clear error message
```

arangod/IResearch/IResearchAqlAnalyzer.cpp
mchacki: i am also a bit resistant against allowing very complex things here.

Subqueries, For Loops, Sort, Collect I understand each of them is desirable to have, but will also create the possibility that we and up with very expensive analyzers, which may take up a lot of resources.

I am not finished with everything yet, and probably have not seen it, but I can imagine that it could get hard to stop / abort / identify / remove an analyzer that explodes in complexity and blocks a thread.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+/*static*/ irs::analysis::analyzer::ptr AqlAnalyzer::make_json(irs::string_ref const& args) {
+  auto builder = VPackParser::fromJson(args);
+  return make_slice(builder->slice());
+}
+
+AqlAnalyzer::AqlAnalyzer(Options const& options)
+  : irs::analysis::analyzer(irs::type<AqlAnalyzer>::get()),
+    _options(options),
+    _query(new CalculationQueryContext(arangodb::DatabaseFeature::getCalculationVocbase())),
+    _itemBlockManager(&_resourceMonitor, SerializationFormat::SHADOWROWS),
```
mchacki, line 449: I think the default memory limit of a resourceMonitor is unbound.
Now every AQL analyzer will get it's own resourceMonitor and can potentially use unbounded amount of main memory.

It is unclear to me what happens if the user manages to write an Analyzer that triggers out of memory situation.
I guess this will potentially crash the server.
If the server reboots, will it find the same Analyzer again and retry => end up in out of memory again ?

mchacki: I do not see a way to limit maximum amount of used memory from outside

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -0,0 +1,562 @@
...
+          return true;
+        }
+      }
+    }
+    if (_executionState == ExecutionState::HASMORE) {
+      _executionState = ExecutionState::DONE;  // set to done to terminate in case of exception
+      _resultRowIdx = 0;
+      _queryResults = nullptr;
+      try {
+        std::tie(_executionState, _queryResults) = _engine.getSome(_options.batchSize);
```
mchacki, line 480: Please use `_engine.execute()` instead of getSome.
Get some is the old API that we just keep around for backwards compatibility and will remove it soon.
Sorry we forgot to add a deprecated Note there.

arangod/IResearch/IResearchAqlAnalyzer.h
```diff
@@ -0,0 +1,109 @@
...
+    /// if set to false each array members is set at positions serially
+    bool collapsePositions{ false };
+
+    /// @brief do not emit empty token if query result is NULL
+    /// this could be used fo index filtering.
+    bool keepNull{ true };
+
+    /// @brief  batch size for running query. Set to 1 as most of the cases
+    /// we expect just 1 to 1 modification query.
+    uint32_t batchSize{ 1 };
```
mchacki, line 65: this highly differs from the default batchSize in AQL queries (1000 rows).
Many of the Aql Blocks do not define how many rows they need to produce, and will result to ask for 1000 rows upstream, causing the upstream to pre-allocate a 1000 lines and then just use 1.
There are some places in AQL where we can scale this down, but not always.

So in order to make this more efficient I think we need to make the default-batch-size internally adjustable to a more sensible number for this case. (I would not set it to 1 that will be too small, but maybe ~10 is best value for this case)

tests/IResearch/IResearchAqlAnalyzer-test.cpp
```diff
@@ -0,0 +1,574 @@
...
+    auto ptr = irs::analysis::analyzers::get(AQL_ANALYZER_NAME,
+                                             irs::type<irs::text_format::vpack>::get(),
+                                             arangodb::iresearch::ref<char>(
+                                               VPackParser::fromJson("{\"queryString\": \"RETURN '1'\"}")->slice()),
+                                             false);
+    ASSERT_NE(nullptr, ptr);
+    assert_analyzer(ptr.get(), "2", {{"1", 0}});
+  }
+  // just parameter
+  {
```
mchacki, line 87: Can we please put these test blocks into single `TEST_F` cases?
We have seen that one large TEST_F with many of them takes ages to compile.
Also if only of them does not work, you only see one at a time, you do not see if it is only one, or all of them that do not work.

tests/IResearch/IResearchAqlAnalyzer-test.cpp
```diff
@@ -0,0 +1,574 @@
...
+            ->slice()),
+          false);
+    ASSERT_NE(nullptr, ptr);
+    assert_analyzer(ptr.get(), "a", {{"", 0}, {"", 0}, {"a", 0}, {"b", 0}});
+  }
+}
+
+TEST_F(IResearchAqlAnalyzerTest, test_create_invalid) {
+  // Forbidden function TOKENS
+  ASSERT_FALSE(
```
mchacki, line 260: Same here, I would prefer to have one TEST_F per forbidden type.

tests/IResearch/IResearchAqlAnalyzer-test.cpp
```diff
@@ -0,0 +1,574 @@
...
+  auto actualVPack = VPackParser::fromJson(actual);
+  auto actualSlice = actualVPack->slice();
+  ASSERT_EQ(actualSlice.get("queryString").stringView(), "RETURN '1'");
+  ASSERT_EQ(actualSlice.get("keepNull").getBool(), true);
+  ASSERT_EQ(actualSlice.get("collapsePositions").getBool(), false);
+  ASSERT_EQ(actualSlice.get("batchSize").getInt(), 1);
+}
+
+TEST_F(IResearchAqlAnalyzerTest, test_normalize) {
+  {
```
mchacki, line 418: Again I prefer to have separate cases.

tests/VPackDeserializer/BasicTests.cpp
```diff
@@ -293,6 +293,55 @@ TEST_F(VPackDeserializerBasicTest, test06) {
...
+  {
+    struct TestStruct {
+      int field1;
+      int field2;
+    };
+    using TestDeserializer = utilities::constructing_deserializer<
+        TestStruct, parameter_list<factory_simple_parameter<field1_name, int, true, values::numeric_value<int, 0>>,
+                                   factory_simple_parameter<field2_name, int, false, values::numeric_value<int, 0>>>>;
+
+    {
```
mchacki, line 309: again I would prefer separate TEST_Fs

tests/js/common/aql/aql-view-arangosearch-feature.js
```diff
@@ -1618,6 +1618,141 @@ function iResearchFeatureAqlTestSuite () {
       }
     },
     
+    testCustomAqlAnalyzer : function() {
```
mchacki, line 1621: This test looks more like a test suite with a tearDown removing the analyzer.
And each testBlock it's own test.

mchacki (changes requested):
I have left a lot of comments.

Special word to the tests:
I like the amount of tests, :+1: and also I think they cover a lot.
I only dislike the way the are added to the test framework.

gnusi:
@jsteemann @mchacki 
Thanks for reviewing! I'd like to elaborate a bit on your concerns about the functionality we allow using in expressions.

We do need FILTER/VARIABLES/SORT/LIMIT/SUBQUERY to provide users a handy and convenient way of **computing** and **filtering** on input values. We do perform streaming thus we're safe from memory consumption perspective.
 As per @jsteemann suggestion, we've introduced a whitelist of nodes, so new functionality has to be validated for compliance first. Also we do restrict total size of analyzer definition, so really huge queries are prohibited.

I don't think we have to worry about really complex queries and thread blocking, since in arangosearch one may explicitly set an arbitrary number of analyzers for each field which with the high probability will lead to the same result, i.e. higher resource consumption and processing time. Same is true for AQL queries in general.

Given all this, I think we're safe enough to proceed with the allowed functionality for computations.

arangod/Aql/Ast.h
```diff
@@ -65,6 +65,14 @@ struct Variable;
 
 typedef std::unordered_map<Variable const*, std::unordered_set<std::string>> TopLevelAttributes;
 
+/// @brief type for Ast flags
+using AstFlagsType = uint32_t;
+
+/// @brief flags for customizing Ast building process
+enum AstFlagType : AstFlagsType {
```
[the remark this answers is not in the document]
Dronplane: Well, I`ve just followed existing notation in AstNodeFlagsType -> AstNodeFlagType  But I`ll ok to make names more distinctive. BTW here (in AstNodeFlagsType) also no `Default`. Should I fix this also? As I`m anyway in it.

arangod/IResearch/IResearchAnalyzerFeature.cpp
Dronplane: Yes. It`s a visitor for analyzers. It should return "Ok" or iteration will stop. So to skip it should still return OK.

arangod/IResearch/IResearchAqlAnalyzer.cpp
Dronplane: It is "local use" string. I prefer to keep it local.

tests/js/common/aql/aql-view-arangosearch-feature.js
Dronplane: As we have many analyzers to test - putting each case for each analyzer into separate suite will make tests run much longer as instantiating suite takes time. So we decided to keep analyzers in separate test methods and do cleanup in `finally` blocks to speed up test runs

tests/IResearch/IResearchAqlAnalyzer-test.cpp
Dronplane: Currently all ArangoSearch tests follows this pattern. If this pattern gives measurable complile/link performance degradation it worth changing it everywhere and by separate PR,

Dronplane:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/12515/

Dronplane:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/12527/

tests/IResearch/IResearchAqlAnalyzer-test.cpp
mchacki: Yes it does, and yes I know ;)

The tests that take longest to compile all follow this pattern and splitting it up as I described helped to speed up by a significant factor (got factor ~100 on one of them)

tests/js/common/aql/aql-view-arangosearch-feature.js
mchacki: Isn't this the use-case for:
setupAll() << initialize the shared constant part
setup() << initiliaze the local test part
teardown() << drop local test part (what is now in finally)
teardownAll() << drop shared constant part

arangod/Aql/Ast.cpp
```diff
@@ -204,7 +204,7 @@ Ast::SpecialNodes::SpecialNodes()
 }
 
 /// @brief create the AST
-Ast::Ast(QueryContext& query, AstFlagsType flags /* = 0 */)
```
mchacki: :+1:

arangod/IResearch/IResearchAqlAnalyzer.h
```diff
@@ -60,9 +60,12 @@ class AqlAnalyzer final : public irs::analysis::analyzer{
     /// this could be used fo index filtering.
     bool keepNull{ true };
 
-    /// @brief  batch size for running query. Set to 1 as most of the cases
-    /// we expect just 1 to 1 modification query.
-    uint32_t batchSize{ 1 };
+    /// @brief  batch size for running query. Set to 10 as most of the cases
+    /// we expect just simple query.
+    uint32_t batchSize{ 10 };
```
mchacki, line 65: oh I think this was a misunderstanding.
BatchSize 1 here is correct. (You want one line), and afaik you pretty-much ensure that you get one line for the index right?
I was talking about the internal AQL BatchSize used by the ExecutionBlocks.
You cannot control it from here.

mchacki (approved):
LGTM.
There is a really minor item in the BatchSize, see comment.
Nothing I would block this PR for, I think 1 or 10 does not really matter performance wise there, changing the internal default batch size is a bit more work and should be done in a different PR.
Sorry that I caused confusion.

PR #12887, opened 2020-10-21
since ARGUMENTS is const for .js 

dothebart:
its easier to have the last argument be the other node than the first (which obviously is the same in the case of 1 argument)


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12394/

dothebart:
tests blue except unrelated error.

jsteemann (approved):

PR #12866, opened 2020-10-16
remove obsoleted collection attributes

jsteemann:
### Scope & Purpose

Remove remainders of non-functional collection attributes `indexBuckets`, `journalSize`, `doCompact`, `isVolatile` from the code.
The attributes were useful for the MMFiles storage engine only, and can be completely ignored for the RocksDB engine.

- [ ] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [x] :hammer: Refactoring 
- [x] :book: CHANGELOG entry made
- [ ] :muscle: The behavior in this PR was *manually tested*
- [x] :computer: The behavior change can be verified via automatic tests

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *all*.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12334/

PR #12903, opened 2020-10-26
Docs: Remove indexBuckets, journalSize, doCompact, isVolatile from docs

Simran-B:
Addendum to https://github.com/arangodb/docs/pull/565 and #12866

@jsteemann What about the `--database.maximal-journal-size` option?
Is it supposed to be removed as well, and hence this DocuBlock?

https://github.com/arangodb/arangodb/blob/devel/Documentation/DocuBlocks/databaseMaximalJournalSize.md
(referenced by arangodb/docs!)

### DIFF
--- a/Documentation/DocuBlocks/Rest/Collections/1_structs.md
+++ b/Documentation/DocuBlocks/Rest/Collections/1_structs.md
@@ -2,14 +2,6 @@
 If *true* then creating, changing or removing
 documents will wait until the data has been synchronized to disk.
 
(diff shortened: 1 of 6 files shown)


jsteemann (approved):

Simran-B:
For posterity: The server ignores additional parameters in the REST API, but unknown startup options are a fatal error. `--database.maximal-journal-size` should thus not be removed for compatibility. It was marked as obsolete, which means it won't show in docs and arangod will warn you if it's used.

PR #12932, opened 2020-10-28
Fix #12778: Typo in special property string for schema validation

### LINKED ISSUE #12778: Edge fails validation if additionalProperties: false
## My Environment

* __ArangoDB Version__:        3.7.2-1
* __Storage Engine__:             RocksDB
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:   Daemon
* __Infrastructure__:               AWS
* __Operating System__:        Ubuntu 20.04
* __Total RAM in your machine__:        2GB
* __Disks in use__:        SSD
* __Used Package__:              Debian or Ubuntu .deb

## Component, Query & Data

__Affected feature__:
Schema validation

## Steps to reproduce

1. Create edge collection
2. Add schema to disallow additional properties
3. Attempt to add document to edge collection

__Problem__:
Adding a schema to validate on an edge collection with `"additionalProperties": false` causes collection to fail validation on any insert or update.

For example, this schema rule:
```
{
  "message": "test",
  "level": "strict",
  "rule": {
    "type": "object",
    "additionalProperties": false
  }
}
```

causes an insert or update with empty contents (other than _from, _to etc.) to fail (`{}`).

The same happens if at least one property is required, and that property is present in the document to be inserted.

However, setting `additionalProperties` to `true` or a schema instead of `false` works as expected: Documents pass validation if their properties fit that schema and fail otherwise.

This seems to only affect Edge collections. Regular collections with `"additionalProperties": false` behave as expected.

__Expected result__:
The document should be allowed, as it does fit the given schema.

Simran-B:
### Scope & Purpose

Fixes `additionalProperties: false` not working correctly for edge collections (`_from` attribute not exempt from validation)

- [x] :hankey: Bugfix 

#### Backports:

#### Related Information

- [x] GitHub issue / Jira ticket number: #12778

### Testing & Verification

  - [x] Added **Regression Tests**

https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/12609/


maxkernbach (approved):
LGTM, the fix works

TODO: add Unit tests + Changelog entry

joerg84 (approved):
Should be backported to 3.7

PR #12933, opened 2020-10-28
fix spurious error when accessing a nullptr transaction context

jsteemann:
### Scope & Purpose

Fixed very spurious errors if the `holdReadLockCollection` replication API for the getting-in-sync procedure of shards was called during server shutdown. In this case that method could ask the transaction manager for a specific transaction, but wasn't return one due to the server shutddown. 

A new failure point has been added to make the transaction manager always return a nullptr, so the problematic behavior can be reproduced. A test was added to set the failure point and trigger the issue.
Test invcation: `scripts/unittest shell_client --cluster true --test tests/js/client/shell/test-collection-counts-cluster.js`

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] Backports required for: 3.7: https://github.com/arangodb/arangodb/pull/12934 - older versions are not affected.

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12535/

### DIFF
--- a/arangod/RestHandler/RestReplicationHandler.cpp
+++ b/arangod/RestHandler/RestReplicationHandler.cpp
@@ -3488,28 +3488,36 @@ Result RestReplicationHandler::createBlockingTransaction(
 
   std::string vn = _vocbase.name();
   try {
-    std::function<void(void)> f = [=]() {
-      try {
-        // Code does not matter, read only access, so we can roll back.
-        transaction::Manager* mgr = transaction::ManagerFeature::manager();
-        if (mgr) {
-          mgr->abortManagedTrx(id, vn);
-        }
-      } catch (...) {
-        // All errors that show up here can only be
-        // triggered if the query is destroyed in between.
-      }
-    };
-
-    std::string comment = std::string("SynchronizeShard from ") + serverId +
-                          " for " + col.name() + " access mode " +
-                          AccessMode::typeString(access);
-    
     if (!serverId.empty()) {
+      std::string comment = std::string("SynchronizeShard from ") + serverId +
+                            " for " + col.name() + " access mode " +
+                            AccessMode::typeString(access);
+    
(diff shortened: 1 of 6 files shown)


jsteemann:
Tests blue

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -3488,28 +3488,36 @@ Result RestReplicationHandler::createBlockingTransaction(
...
   std::string vn = _vocbase.name();
   try {
-    std::function<void(void)> f = [=]() {
-      try {
-        // Code does not matter, read only access, so we can roll back.
-        transaction::Manager* mgr = transaction::ManagerFeature::manager();
-        if (mgr) {
-          mgr->abortManagedTrx(id, vn);
+    if (!serverId.empty()) {
+      std::string comment = std::string("SynchronizeShard from ") + serverId +
```
Dronplane, line 3492: If we declare it befor reclaring `f` below - we will force `message` to be copied into capture. Is that intended, given `f` does not use `message`?

Dronplane (approved):
LGTM  - only minor possible  redundant copy comment

arangod/RestHandler/RestReplicationHandler.cpp
Dronplane: Ah no, seems that I`m wrong. Only used is copied into capture.Pls, ignore this

jsteemann: I think `=` does not do unnecessary copies. According ot https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture it only copies *used* variables:

>      = (implicitly capture the used automatic variables by copy).

PR #12934, opened 2020-10-28
fix spurious error when accessing a nullptr transaction context

jsteemann:
### Scope & Purpose

Fixed very spurious errors if the `holdReadLockCollection` replication API for the getting-in-sync procedure of shards was called during server shutdown. In this case that method could ask the transaction manager for a specific transaction, but wasn't return one due to the server shutddown. 

A new failure point has been added to make the transaction manager always return a nullptr, so the problematic behavior can be reproduced. A test was added to set the failure point and trigger the issue.
Test invcation: `scripts/unittest shell_client --cluster true --test tests/js/client/shell/test-collection-counts-cluster.js`

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12537/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 v3.7.5 (XXXX-XX-XX)
 -------------------
 
+* Fixed very spurious errors if the `holdReadLockCollection` replication API for
+  the getting-in-sync procedure of shards was called during server shutdown.
+  In this case that method could ask the transaction manager for a specific
+  transaction, but wasn't return one due to the server shutddown. 
+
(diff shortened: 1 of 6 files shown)


jsteemann:
Tests blue

Dronplane (approved):
LGTM

danielhlarkin (approved):

KVS85 (approved):
LGTM.

PR #12937, opened 2020-10-29
reduce avoidable noise caused by replication log messages

jsteemann:
### Scope & Purpose

Reduce the noise caused by replication log messages

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 

### Testing & Verification

- [x] This change is already covered by existing tests, such as *replication_sync, replication_static*.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12544/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12549/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12558/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12564/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12575/

js/client/modules/@arangodb/testsuites/replication.js
```diff
@@ -529,7 +529,7 @@ function replicationSync (options) {
       let shutdownState = true;
       let res = true;
       print("starting replication slave: ");
-      let slave = pu.startInstance('tcp', options, {"log.level" : "replication=trace", "--log.level": "replication=trace"}, 'slave_sync');
+      let slave = pu.startInstance('tcp', options, {"log.level" : "replication=info", "--log.level": "replication=info"}, 'slave_sync');
```
dothebart, line 532: please make this configurable by `if (options.extremeVerbosity)`

dothebart (approved):
please make the noise configurable, since it may be desireable. 
otherwise LGTM.

js/client/modules/@arangodb/testsuites/replication.js
jsteemann: None of the other replication testsuites has something like this.
I remember we added this a long time ago when debugging some replication issues, and never removed it later.
As I said, there are numerous other replication suites which all don't have this, and my position is that we should remove it here too. It is just a leftover from long-ago debugging.

dothebart: `extremeVerbosity` has to be specified from the commandline. So it will not be there in jenkins runs. 
If you specify it, you want it noisy.

jsteemann: I understand, but why should we honor the flag here, in this single testsuite, but in none of the other replication testsuites?
I think we don't need this at all.
It was there as a leftover from debugging a specific problem, which has been solved long ago.

jsteemann, line 532:
```suggestion
      let slave = pu.startInstance('tcp', options, {}, 'slave_sync');
```

jsteemann:
@dothebart : now removed all special handling for logging in the one affected test.
Now it is exactly as all other tests in this file: no specials.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12626/

dothebart:
ok.

PR #12940, opened 2020-10-29
Feature/snappy 1.1.8

jsteemann:
### Scope & Purpose

Upgrade version of bundled Snappy compression/decompression library from 1.1.7 to 1.1.8.
1.1.8 is from January 2020, 1.1.7 is from August 2017.
The Snappy `NEWS` file mentions "small performance improvements", their git log has a bit more details. It is still unclear if the new version provides relevant performance improvements for us. This PR is created to figure this out, so we can run a side-by-side comparison.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 

#### Backports:

- [x] No backports required

### Testing & Verification

Basically everything needs to compile on all platforms, and installers need to work too.
If that is achieved, we can compare the performance.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12556/


jsteemann:
Tests blue

dothebart (approved):

PR #12950, opened 2020-10-30
added memory profiles, fixed memory leak

fceller:
### Scope & Purpose

#### Forward ports:

- [x] Forwardports required for: devel

#### Related Information

- [X] GitHub issue / Jira ticket number: OASIS-352

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,5 @@
-v3.7.3 (XXXX-XX-XX)
--------------------
+v3.7.3.1 (XXXX-XX-XX)
+---------------------
 
 * Fix a memory leak because cluster internal connections were not cleaned up for
   agency communication.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,11 +1,14 @@
 v3.7.3 (2020-10-14)
 -------------------
-      
+
+* Fix a memory leak because cluster internal connections were not cleaned
+  up for agency communication.
+
 * Added the following metrics for synchronous replication in the cluster:
 
-  - `arangodb_refused_followers_count`: Number of times a shard leader received 
+  - `arangodb_refused_followers_count`: Number of times a shard leader received
     a refusal answer from a follower during synchronous replication.
-  - `arangodb_sync_wrong_checksum`: Number of times a mismatching shard 
+  - `arangodb_sync_wrong_checksum`: Number of times a mismatching shard
     checksum was detected when syncing shards. In case this happens, a resync
     will be triggered for the shard.
 
@@ -314,6 +317,8 @@ v3.7.3 (2020-10-14)
 
 * Fixed issue BTS-169: cost estimation for LIMIT nodes showed wrong number of
   estimated items.
+
+* Added compile option USE_JEMALLOC_PROF to enable memory profiling
 
 
 v3.7.2.2 (2020-10-07)
--- a/arangod/Cluster/HeartbeatThread.cpp
+++ b/arangod/Cluster/HeartbeatThread.cpp
@@ -213,7 +213,7 @@ HeartbeatThread::HeartbeatThread(application_features::ApplicationServer& server
       _invalidateCoordinators(true),
       _lastPlanVersionNoticed(0),
       _lastCurrentVersionNoticed(0),
-      _DBServerUpdateCounter(0),
+      _updateCounter(0),
       _agencySync(_server, this),
       _heartbeat_send_time_ms(server.getFeature<arangodb::MetricsFeature>().histogram(
           StaticStrings::HeartbeatSendTimeMs, log_scale_t<uint64_t>(2, 4, 8000, 10),
@@ -373,11 +373,11 @@ void HeartbeatThread::getNewsFromAgencyForDBServer() {
 
   // Periodically update the list of DBServers and prune agency comm
   // connection pool:
-  if (++_DBServerUpdateCounter >= 60) {
+  if (++_updateCounter >= 60) {
     auto& clusterFeature = server().getFeature<ClusterFeature>();
     auto& ci = clusterFeature.clusterInfo();
     ci.loadCurrentDBServers();
-    _DBServerUpdateCounter = 0;
+    _updateCounter = 0;
     clusterFeature.pruneAsyncAgencyConnectionPool();
   }
 }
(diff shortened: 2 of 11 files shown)


arangod/Cluster/HeartbeatThread.cpp
```diff
@@ -697,6 +707,8 @@ void HeartbeatThread::getNewsFromAgencyForCoordinator() {
   if (++_DBServerUpdateCounter >= 60) {
     ci.loadCurrentDBServers();
     _DBServerUpdateCounter = 0;
```
jsteemann, line 709: The variable name `_DBServerUpdateCounter` is a bit misleading here, as it is obviously also used for coordinators. 
So we should probably rename it to `_updateIteration` or so.

arangod/Cluster/HeartbeatThread.cpp
```diff
@@ -697,6 +707,8 @@ void HeartbeatThread::getNewsFromAgencyForCoordinator() {
...
+    auto& clusterFeature = server().getFeature<ClusterFeature>();
+    clusterFeature.pruneAsyncAgencyConnectionPool();
   }
 }
```
jsteemann, line 713: Do we need to add the call to `pruneAsyncAgencyConnectionPool()` also in the `runSingleServer()` method of the HeartbeatThread? That method is used in active failover deployments, and it also uses the agency...

arangod/RestHandler/RestStatusHandler.cpp
```diff
@@ -274,3 +285,27 @@ RestStatus RestStatusHandler::executeOverview() {
...
+
+  if (!found) {
+    outDirectory = TRI_GetTempPath();
+  }
+
+  std::string const fileName = FileUtils::buildFilename(outDirectory, "profile.out");
+  char const* f = fileName.c_str();
+  mallctl("prof.dump", NULL, NULL, &f, sizeof(const char *));
+  std::string const content = FileUtils::slurp(fileName);
+  TRI_UnlinkFile(f);
```
jsteemann, line 302: I think allowing to specify arbitrary directories names via URL parameters is a security issue.
In addition, I am not sure what happens if the target directory does not exist. Will the memory profile dump fail?
This is also not thread-safe, because if no directory is specified, multiple callers will write to the same output file concurrently.

Instead of all this, we should also create a temporary filename in the temporary directory, write the dump into it and then remove it again.

Last but not least, the profile dump file is not cleaned up if an exception occurs in FileUtils::slurp(...). We should clean up the profile dump file in all cases, so I guess we should remove it via a scopeGuard.

CHANGELOG
```diff
@@ -315,6 +318,8 @@ v3.7.3 (2020-10-14)
 * Fixed issue BTS-169: cost estimation for LIMIT nodes showed wrong number of
   estimated items.
 
+* Added compile option USE_JEMALLOC_PROF to enable memory profiling
```
jsteemann, line 321: This adds the new option to the CHANGELOG of the already-released version 3.7.3.
I guess the CHANGELOG entry should go into the section for version 3.7.5 instead.

CHANGELOG
```diff
@@ -1,11 +1,14 @@
 v3.7.3 (2020-10-14)
 -------------------
-      
+
+* Fix a memory leak because cluster internal connections were not cleaned
+  up for agency communication.
```
jsteemann, line 5: This adds the new option to the CHANGELOG of the already-released version 3.7.3.
I guess the CHANGELOG entry should go into the section for version 3.7.5 instead.

KVS85: 3.7.3.1 will have it's own section.

CHANGELOG
KVS85: Moved to 3.7.3.1 section instead.

KVS85:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12584/

arangod/RestHandler/RestStatusHandler.cpp
fceller: done

CHANGELOG
```diff
@@ -1,11 +1,20 @@
+v3.7.3 (XXXX-XX-XX)
```
jsteemann, line 1:
```suggestion
v3.7.3.1 (XXXX-XX-XX)
```

arangod/RestHandler/RestStatusHandler.cpp
```diff
@@ -274,3 +285,26 @@ RestStatus RestStatusHandler::executeOverview() {
   generateResult(rest::ResponseCode::OK, result.slice());
   return RestStatus::DONE;
 }
+
+RestStatus RestStatusHandler::executeMemoryProfile() {
+#if defined(USE_MEMORY_PROFILE)
+  std::string const outDirectory = TRI_GetTempPath();
+  std::string const fileName = FileUtils::buildFilename(outDirectory, "profile.out");
```
jsteemann, line 292: I think it should be changed further, as `fileName` will the same for all threads that enter this method, so it is still not thread-safe. I think we should produce a random filename every time or add a mutex.

arangod/RestHandler/RestStatusHandler.cpp
```diff
@@ -274,3 +285,26 @@ RestStatus RestStatusHandler::executeOverview() {
...
+
+    _response->setContentType(rest::ContentType::TEXT);
+    _response->addRawPayload(velocypack::StringRef(content));
+  } catch (...) {
+    TRI_UnlinkFile(f);
+    throw;
+  }
+
+#endif
+  return RestStatus::DONE;
```
jsteemann, line 309: In case we get here, but `USE_MEMORY_PROFILE` was not defined, the handler will not return any response (not even an error). This is currently prevented by an earlier check in https://github.com/arangodb/arangodb/pull/12950/files#diff-61ec4959a4f0db4cf223f986abc0033d454368ff6825efb23c30857402f58b64R78-R81, but it requires looking at two different places. 
I suggest we add an explicit `not implemented` or `not found` response here for the case in which `USE_MEMORY_PROFILE` is not defined. Then we can also remove the other check and have all the handling in one place.

fceller: Done

arangod/RestHandler/RestStatusHandler.cpp
fceller: done

jsteemann (approved):

arangod/Cluster/HeartbeatThread.cpp
kvahed: yes

kvahed (commented):
LGTM

arangod/Cluster/HeartbeatThread.cpp
fceller: We will address this in a second PR

PR #12957, opened 2020-10-31
fix potential deadlock in cluster transactions

jsteemann:
### Scope & Purpose

Fix a potential deadlock in cluster transaction code in `transactions::Manager::returnManagedTrx()`.
This method acquired the write-lock on the transactions bucket, and then can call into `abortManagedTrx` while the lock is still being held.
`abortManagedTrx` directly calls `statusChangeWithTimeout`, which itself unconditionally calls `updateTransaction`, which will also unconditionally try to acquire the same bucket lock. This would be a deadlock.
We have never seen this in practice, and according to the code it looks like it can only happening when `beginShutdown()` or `stop()` are called on the TransactionManagerFeature. If this deadlock ever happened, it should be pretty rare.

Backport of the actual fix in https://github.com/arangodb/arangodb/pull/12955

#### Backports:

- [x] Backports required for: 3.5

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12591/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.6.9 (XXXX-XX-XX)
 -------------------
 
+* Fixed potential deadlock in cluster transactions if a transaction is
+  returned that was soft-aborted by transaction garbage collection before.
+
 * Bug-fix: Allow to unlink a view created on a SmartGraphEdge collection.
 
 * Fixed an ownership issue which could lead to undefined behavior in case an AQL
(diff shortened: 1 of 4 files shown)


jsteemann:
http://172.16.10.101:8080/job/arangodb-matrix-pr-linux/14152/

KVS85:
Tests blue.

danielhlarkin (approved):

KVS85 (approved):

PR #12963, opened 2020-11-02
backport for large follower transactions

jsteemann:
### Scope & Purpose

With this PR, no followers should be dropped in this situation.

The PR adds a new test suite in file `tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js` to test the changes. Some of the tests are only executed if there are at least 3 DB servers, so the PR also changes the number of DB servers for some of the tests to 3 (from previously 2).

Backport of https://github.com/arangodb/arangodb/pull/12654

#### Backports:

- [ ] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12600/


jsteemann:
Obsoleted by https://github.com/arangodb/arangodb/pull/12987

PR #12967, opened 2020-11-02
Fix #12778: Typo in special property string for schema validation

### LINKED ISSUE #12778: Edge fails validation if additionalProperties: false

Simran-B:
Backport of #12932

https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/12615/

### DIFF
--- a/3rdParty/json-schema-validation/src/validation.cpp
+++ b/3rdParty/json-schema-validation/src/validation.cpp
@@ -84,7 +84,7 @@ template<template<typename...> class Traits>
 std::string const key_string = "_key";
 std::string const rev_string = "_rev";
 std::string const id_string = "_id";
-std::string const from_string = "_form";
+std::string const from_string = "_from";
 std::string const to_string = "_to";
 
 } // namespace
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,14 @@
 v3.7.5 (XXXX-XX-XX)
 -------------------
 
+* Fixed issue #12778: fails validation if additionalProperties: false
+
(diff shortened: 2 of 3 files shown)


KVS85 (approved):

KVS85:
Tests blue.

jsteemann (approved):
LGTM! Thanks!

PR #12972, opened 2020-11-02
Feature/simplify thread exceptions

jsteemann:
### Scope & Purpose

Simplify exception catching in Thread
    
It is currently up to the implementor to catch exceptions and handle server shutdown during the execution of `Thread::run()`. Proper handling needs to be implemented for all classes derived from `Thread` separately.

This is tedious and error-prone, so this PR tries to simplify this by adding a class `TaskThread` (derived from `Thread`), which takes over some of the boilerplate work.
It will automatically log and consume exceptions thrown during thread execution, and will automatically check for server shutdown.

It is not a 1:1 replacement for the existing `Thread::run()`. The key differences are:
* `Thread::run()` will normally contain a loop and run until server shutdown. This is what we want to simplify.
* `TaskThread::run()` (which is declared final) will run one iteration of such loop, which must be implemented in derived classes in the `runTask()` method. If any exception occurs during `runTask()`, `TaskThread::run()` will log it and consume it. `TaskThread::run()` will call `runTask()` as long as the thread is not stopping. Premature exit can be achieved by returning false from `runTask()`.
* Execution of prologue and epilogue work that should only run once per thread and not on every iteration of `runTask()` can be achieved by overloading `runSetup()` and `runShutdown()`.

Usage example:
```
class FooThread : public TaskThread {
 public:
  explicit FooThread(application_featues::ApplicationServer& server) : TaskThread(server, "foo") {}
  ~FooThread() { shutdown(); }

  // code that is executed once in this thread. if it throws, runTask() and runTeardown() will not be called,
  // and the exception is propagated!
  // it is optional to overload this. the default implementation does nothing
  void runSetup() override { ... }

  // will be called repeatedly while the thread is not shutting down, and while runTask() itself will return true;
  bool runTask() override {
    // do some work
    ...
    // any waiting before next iteration must be performed here!
    std::this_thread::sleep_for(std::chrono::milliseconds(50)); 
    // return true to run this in a loop, false to exit 
    return true;
  }

  // code that is executed once in this thread after the main runTask() loop is left.
  // if it throws an exception, it is logged and consumed
  // it is optional to overload this. the default implementation does nothing
  void runTeardown() override { ... }
};
```

In this PR, a few existing threads have been modified to use the new interface, but not all. Some threads do more complex processing and are not really suitable to be used with new interface, or only with some moderate migration efforts.

Finally, this PR ensures that the supervision, heartbeat and heartbeat background threads do not abort processing when encountering an exception, but simply continue. This also removes the need for tracking failed threads via the CriticalThread class, which previously was only used for the supervision and the heartbeat, but not elsewhere.

The changes are best viewed when ignoring all whitespace changes, because a substantial amount of changes is caused by simply indenting code so it can go inside new try...catch blocks.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *scripts/unittest all*.
  - [x] Added new C++ **Unit tests**: added tests for TaskThread
 
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12623/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12624/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12637/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12638/

jsteemann:
Closing in favor of https://github.com/arangodb/arangodb/pull/12983

PR #12977, opened 2020-11-03
Fix deadlocks between write transactions and hotbackup, take 2

neunhoef:
Make sure all follower trxs know that they are follower trxs.

Before this patch, follower transactions created from AQL or other
El Cheapo transactions did not know they are follower transactions.
This could cause deadlocks between write transactions and hotbackups.

# Pull Request Guidelines

Pull requests are an essential collaborative tool for modern software development. 

The below list is intended to help you figure out whether your code is ready to be reviewed
and merged into ArangoDB. The overarching goal is to:

- Reduce the amount of *recurring* defects
- Reduce the impact to the other developer’s time and energy spent on defects in your code
- Increase the overall autonomy and productivity of individual developers

## Acceptance Checklist

*The below list is not exhaustive, think thoroughly whether the provided information is sufficient.*
*Remove options that do not apply*

### Scope & Purpose

*(Can you describe what functional change your PR is trying to effect?)*

- [ ] Bug-Fix for *devel-branch* (i.e. no need for backports?)
- [+] Bug-Fix for a *released version* (did you remember to port this to all relevant release branches?)
- [ ] Strictly *new functionality* (i.e. a new feature / new option, no need for porting)
- [+] The behavior in this PR can be (and was) *manually tested* (support / qa / customers can test it)
- [ ] The behavior change can be verified via automatic tests

#### Related Information

*(Please reference tickets / specification etc )*

- [ ] There is a GitHub Issue reported by a Community User: 
- [ ] There is an internal planning ticket:
- [ ] There is a *JIRA Ticket number* (In case a customer was affected / involved): 
- [ ] There is a design document: 

### Testing & Verification

This should be put under automatic tests. These are not yet implemented
in this PR. Our test framework is not particularly well suited to test
this.

Test that should be done:

Create a one shard database.

Run this in `arangosh`:

```
c = db._create("c");
for (i = 0; i < 1000; ++i) { c.insert({Hallo:i}); }
for (i = 0; i < 1000; ++i) { 
  db._query(`LET m = (FOR d IN c SORT d.Hallo DESC LIMIT 1 RETURN d)
             INSERT {Hallo: m[0].Hallo+1} INTO c RETURN NEW 
             OPTIONS {exclusive:true}`,).toArray();
  require("internal").wait(0.01);
  print(i);
}
```

Constantly create hotbackups during the time. Hotbackups should succeed
quickly and the above loop should not be slowed down too much.

### DIFF
--- a/arangod/RestHandler/RestDocumentHandler.cpp
+++ b/arangod/RestHandler/RestDocumentHandler.cpp
@@ -199,9 +199,6 @@ RestStatus RestDocumentHandler::insertDocument() {
   if (!isMultiple && !opOptions.overwrite) {
      _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
   }
-  if (!opOptions.isSynchronousReplicationFrom.empty()) {
-    _activeTrx->addHint(transaction::Hints::Hint::IS_FOLLOWER_TRX);
-  }
 
   Result res = _activeTrx->begin();
   if (!res.ok()) {
@@ -489,9 +486,6 @@ RestStatus RestDocumentHandler::modifyDocument(bool isPatch) {
   if (!isArrayCase) {
     _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
   }
-  if (!opOptions.isSynchronousReplicationFrom.empty()) {
-    _activeTrx->addHint(transaction::Hints::Hint::IS_FOLLOWER_TRX);
-  }
 
   // ...........................................................................
   // inside write transaction
@@ -610,9 +604,6 @@ RestStatus RestDocumentHandler::removeDocument() {
   _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
   if (suffixes.size() == 2 || !search.isArray()) {
     _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
-  }
-  if (!opOptions.isSynchronousReplicationFrom.empty()) {
-    _activeTrx->addHint(transaction::Hints::Hint::IS_FOLLOWER_TRX);
   }
 
   Result res = _activeTrx->begin();
--- a/arangod/RestHandler/RestDocumentHandler.cpp
+++ b/arangod/RestHandler/RestDocumentHandler.cpp
@@ -193,7 +193,7 @@ RestStatus RestDocumentHandler::insertDocument() {
                          opOptions.isSynchronousReplicationFrom);
 
   // find and load collection given by name or identifier
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
   bool const isMultiple = body.isArray();
 
   if (!isMultiple && !opOptions.overwrite) {
@@ -298,7 +298,7 @@ RestStatus RestDocumentHandler::readSingleDocument(bool generateBody) {
   VPackSlice search = builder.slice();
 
   // find and load collection given by name or identifier
-  _activeTrx = createTransaction(collection, AccessMode::Type::READ);
+  _activeTrx = createTransaction(collection, AccessMode::Type::READ, options);
 
   _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
 
@@ -484,7 +484,7 @@ RestStatus RestDocumentHandler::modifyDocument(bool isPatch) {
   }
 
   // find and load collection given by name or identifier
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
 
   if (!isArrayCase) {
     _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
@@ -607,7 +607,7 @@ RestStatus RestDocumentHandler::removeDocument() {
     return RestStatus::DONE;
   }
 
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
   if (suffixes.size() == 2 || !search.isArray()) {
     _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
   }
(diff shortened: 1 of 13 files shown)


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12629/

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -193,7 +193,7 @@ RestStatus RestDocumentHandler::insertDocument() {
                          opOptions.isSynchronousReplicationFrom);
 
   // find and load collection given by name or identifier
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
```
mchacki, line 196: isn't the `addHint(IS_FOLLOWER_TRX)` a couple of lines be below now twice? 
I would prefer to have it centralized

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -484,7 +484,7 @@ RestStatus RestDocumentHandler::modifyDocument(bool isPatch) {
   }
 
   // find and load collection given by name or identifier
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
```
mchacki, line 488: Again a couple if lines below we add IS_FOLLOWER_TRX twice

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -607,7 +607,7 @@ RestStatus RestDocumentHandler::removeDocument() {
     return RestStatus::DONE;
   }
 
-  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE);
+  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
   if (suffixes.size() == 2 || !search.isArray()) {
```
mchacki, line 611: Again twice

tests/Transaction/Manager-test.cpp
```diff
@@ -95,12 +95,12 @@ class TransactionManagerTest : public ::testing::Test {
 
 TEST_F(TransactionManagerTest, parsing_errors) {
   auto json = arangodb::velocypack::Parser::fromJson("{ \"write\": [33] }");
-  Result res = mgr->createManagedTrx(vocbase, tid, json->slice());
```
mchacki, line 98: Can we add tests with `true` for follower Transaction and assert that the HINT is set, and otherwise behavior is identical.
Maybe we can run this suite in a templated Test variant for true/false ?

mchacki (changes requested):
In general I think this fix is ok and works.

I have a couple of cleanup comments and would wish for additional tests. I think some should be easy, see comments.

Also I would like to see integration tests that on follower shards only FOLLOWER transactions are created. (AQL / DocumentOperations, JSTransactions etc).
I am not 100% sure how to write them, maybe we can activate a failurePoint on the follower server, that makes the server crash if ever a transaction is started as non-follower.
This should work if we move the shards round s.t. one DBServer is not responsible for any system shard, and follower of the Shard in Test.
Then we fire our APIs against this shard.

kvahed (approved):
LGTM. First @mchacki comment is right.

arangod/RestHandler/RestDocumentHandler.cpp
neunhoef: Changed.

arangod/RestHandler/RestDocumentHandler.cpp
neunhoef: Changed.

arangod/RestHandler/RestDocumentHandler.cpp
neunhoef: Changed.

arangod/RestHandler/RestVocbaseBaseHandler.cpp
```diff
@@ -586,12 +586,16 @@ void RestVocbaseBaseHandler::extractStringParameter(std::string const& name,
...
 std::unique_ptr<transaction::Methods> RestVocbaseBaseHandler::createTransaction(
-    std::string const& collectionName, AccessMode::Type type) const {
+    std::string const& collectionName, AccessMode::Type type, OperationOptions const& opOptions) const {
   bool found = false;
   std::string const& value = _request->header(StaticStrings::TransactionId, found);
   if (!found) {
-    return std::make_unique<SingleCollectionTransaction>(transaction::StandaloneContext::Create(_vocbase),
+    auto tmp = std::make_unique<SingleCollectionTransaction>(transaction::StandaloneContext::Create(_vocbase),
                                                          collectionName, type);
+    if (!opOptions.isSynchronousReplicationFrom.empty()) {
```
jsteemann, line 595: can we also add a check here that we are on a DB server?
```suggestion
    if (!opOptions.isSynchronousReplicationFrom.empty() && ServerState::instance()->isDBServer()) {
```

arangod/Transaction/Manager.cpp
```diff
@@ -346,6 +346,9 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TRI_voc_tid_t tid,
     return res.reset(TRI_ERROR_BAD_PARAMETER,
                      "<lockTimeout> needs to be positive");
   }
+  if (isFollowerTransaction) {
```
jsteemann, line 349:
```suggestion
  if (isFollowerTransaction && ServerState::instance()->isDBServer()) {
```

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12645/

tests/Transaction/Manager-test.cpp
neunhoef: Done.

neunhoef:
We will now go ahead with this without further integration tests, since a fixed version is desperately needed. We will add further integration tests when we port this to 3.7 and `devel`.

arangod/Transaction/Manager.cpp
neunhoef: I had to revert these changes since in the unit tests the server does not have role DBServer. :-(

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12647/

KVS85:
Tests blue.

mchacki (approved):
I am not happy with test-coverage yet.
But as @neunhoef says they will be added.

PR #12985, opened 2020-11-04
add some instrumentation for collection locking

jsteemann:
### Scope & Purpose

* Added metrics for collection locks:
  - `arangodb_collection_lock_timeouts_exclusive`: Number of lock timeouts when trying to acquire collection exclusive locks 
  - `arangodb_collection_lock_timeouts_write`: Number of lock timeouts when trying to acquire collection write locks 
  - `arangodb_collection_lock_acquisition_micros`: Total amount of collection lock acquisition time [μs]
  - `arangodb_collection_lock_acquisition_time`: Total collection lock acquisition time histogram [s]

* Reduce lock timeout on followers to 15 seconds.
  Rationale: we should not have any locking conflicts on followers, generally. Any shard locking should be performed on leaders first, which will then, eventually replicate changes to followers. replication to followers is only done once the locks have been acquired on the leader(s). 

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [x] Backports required for: 3.7

### Testing & Verification

- [x] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12643/

PR #12987, opened 2020-11-04
Feature 3.7/add locking instrumentation

jsteemann:
### Scope & Purpose

This PR contains several improvements for cluster transactions and collection locking, which are somewhat related:

With this PR, no followers should be dropped in this situation.

The PR adds a new test suite in file `tests/js/client/shell/shell-transaction-intermediate-commit-cluster.js` to test the changes. Some of the tests are only executed if there are at least 3 DB servers, so the PR also changes the number of DB servers for some of the tests to 3 (from previously 2).

Backport of https://github.com/arangodb/arangodb/pull/12654

This PR also adds metrics for collection locks:
- `arangodb_collection_lock_timeouts_exclusive`: Number of lock timeouts when trying to acquire collection exclusive locks 
- `arangodb_collection_lock_timeouts_write`: Number of lock timeouts when trying to acquire collection write locks 
- `arangodb_collection_lock_acquisition_micros`: Total amount of collection lock acquisition time [μs]
- `arangodb_collection_lock_acquisition_time`: Total collection lock acquisition time histogram [s]

And it reduces lock timeout on followers to 15 seconds.
Rationale: we should not have any locking conflicts on followers, generally. Any shard locking should be performed on leaders first, which will then, eventually replicate changes to followers. replication to followers is only done once the locks have been acquired on the leader(s). 

The latter two things are a backport of https://github.com/arangodb/arangodb/pull/12985

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/584
Docs PR: https://github.com/arangodb/docs/pull/575

#### Backports:

- [x] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12646/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12654/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12656/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12657/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12664/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12667/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12672/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12674/

jsteemann:
Tests blue

arangod/Transaction/Manager.cpp
```diff
@@ -312,15 +316,71 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TRI_voc_tid_t tid,
     auto& buck = _transactions[bucket];
     auto it = buck._managed.find(tid);
     if (it != buck._managed.end()) {
+      if (ServerState::instance()->isDBServer() && isFollowerTransactionId(tid)) {
```
neunhoef, line 319: Do we need a case here that we are looking at a legacy transaction id and yet this is a follower transaction? What if the transaction has the hint IS_FOLLOWER_TRANSACTION?

arangod/Transaction/Manager.cpp
```diff
@@ -312,15 +316,71 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TRI_voc_tid_t tid,
...
-                       "' already used, (before creating).");
+                       std::string("transaction id ") + std::to_string(tid) +
+                       " already used, (before creating)");
     }
   }
 
   // enforce size limit per DBServer
-  options.maxTransactionSize =
-      std::min<size_t>(options.maxTransactionSize, Manager::maxTransactionSize);
+  if (ServerState::instance()->isDBServer() && isFollowerTransactionId(tid)) {
```
neunhoef, line 341: See above.

arangod/Transaction/Manager.cpp
```diff
@@ -720,9 +800,16 @@ Result Manager::updateTransaction(TRI_voc_tid_t tid, transaction::Status status,
     }
   } else {
     res = trx.abort();
-    if (res.ok() && wasExpired) {
+    if (intermediateCommits &&
+        ServerState::instance()->isDBServer() &&
+        isFollowerTransactionId(tid)) {
```
neunhoef, line 805: See above.

neunhoef (changes requested):
Generally: LGTM. I have added 3 comments in places, where I think we need to distinguish another case: The case that we are looking at a follower transaction which has a "legacy" transaction ID.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12737/

arangod/Transaction/Manager.cpp
jsteemann: Done

arangod/Transaction/Manager.cpp
jsteemann: Done

arangod/Transaction/Manager.cpp
jsteemann: Done

neunhoef (approved):
LGTM

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12750/

jsteemann:
Yet another run, due to MacOS compile warning:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12752/

jsteemann:
Tests blue

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12987/

KVS85:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13228/

KVS85:
Tests blue.

KVS85 (approved):
LGTM.

PR #12997, opened 2020-11-06
add option to run arangobench for a duration, instead of a count of operations

dothebart:
### Scope & Purpose

this PR enables arangobench to be ran for a specific time duration instead of a specified count. 
So if you want to load your system for a while, this way its more convenient.

It may now also create the database it operates on on start.

- [x] :pizza: New feature

#### Backports:

- [ ] No backports required
- [x] Backports required for: 3.7, 3.6

### Testing & Verification

  - [x] Added new **integration tests** a new arangobench test checks, whether the specified run-time is respected.

### DIFF
--- a/arangosh/Benchmark/BenchFeature.cpp
+++ b/arangosh/Benchmark/BenchFeature.cpp
@@ -43,6 +43,8 @@
 #include "Basics/application-exit.h"
 #include "Basics/files.h"
 #include "Basics/system-functions.h"
+#include "Logger/Logger.h"
+#include "Logger/LogMacros.h"
 #include "Benchmark/BenchmarkCounter.h"
 #include "Benchmark/BenchmarkOperation.h"
 #include "Benchmark/BenchmarkThread.h"
@@ -74,11 +76,14 @@ BenchFeature::BenchFeature(application_features::ApplicationServer& server, int*
       _async(false),
       _concurrency(1),
       _operations(1000),
+      _realOperations(0),
       _batchSize(0),
+      _duration(0),
       _keepAlive(true),
       _collection("ArangoBenchmark"),
       _testCase("version"),
       _complexity(1),
+      _createDatabase(false),
       _delay(false),
       _progress(true),
       _verbose(false),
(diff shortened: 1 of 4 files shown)


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12680/

arangosh/Benchmark/BenchFeature.cpp
```diff
@@ -147,6 +152,14 @@ void BenchFeature::collectOptions(std::shared_ptr<ProgramOptions> options) {
                      "use waitForSync for created collections",
                      new BooleanParameter(&_waitForSync));
 
+  options->addOption("--create-database",
+                     "whether we should create the database specified via the server connection",
+                     new BooleanParameter(&_createDatabase));
+
+  options->addOption("--duration",
+                     "test for duration seconds instead of a fixed test count",
+                     new UInt64Parameter(&_duration));
```
jsteemann, line 161: We normally also allow fractional values for seconds, at least in many other options. So we should maybe change this to a double parameter type?

arangosh/Benchmark/BenchFeature.cpp
```diff
@@ -218,11 +231,35 @@ void BenchFeature::start() {
...
+
+    if (!result || !result->isComplete()) {
+      std::string msg;
+      if (result) {
+        msg = result->getHttpReturnMessage();
+        delete result;
+      }
+
+      LOG_TOPIC("5cda8", FATAL, arangodb::Logger::FIXME)
+        << "failed to create the specified database " << msg;
```
jsteemann, line 256:
```suggestion
        << "failed to create the specified database: " << msg;
```

arangosh/Benchmark/BenchFeature.cpp
dothebart: hm, the usecase here is rather to go into the range of hours, than fractions of seconds. 
Since current requests may need to finish first, it won't be completely "on time" anyways, so fragments would pretend a precision that can't be met anyways.

PR #13009, opened 2020-11-09
[3.7] Pregel improvements

danielhlarkin:
### Scope & Purpose

Improvements for the Pregel distributed graph processing feature:

* during the loading/startup phase, the in-memory edge cache is now intentionally bypassed. The reason for this is that any edges are looked up exactly once, so caching them is not beneficial, but would only lead to cache pollution.
* the loading/startup phase can now load multiple collections in parallel, whereas previously it was only loading multiple shards of the same collection in parallel. This change helps to reduce load times in case there are many collections with few shards, and on single server.
* the loading and result storage phases code has been overhauled so that it runs slightly faster, regardless of parallelism settings.
* for Pregel runs that are based on named graphs (in contrast to explicit naming of the to-be-used vertex and edge collections), only those edge collections are considered that, according to the graph definition, can have connections with the vertex. This change can reduce the loadingtime substantially in case the graph contains many edge definitions.
* the number of executed rounds for the underlying Pregel algorithm now does not vary for different `parallelism` values.
* memory usage for storing intermediate vertex data has been reduced.
* the internal Pregel algorithm APIs have been cleaned up a bit.
* the existing shell tests for Pregel have been modified so that they now run in both arangod and arangosh (also covering the Pregel HTTP API a bit more)

A forward port for devel is currently in the works.

#### Backports:

- [x] No backports required. We will however forward port this to devel.

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell-pregel.js*.
  - [x] Added new **integration tests** (e.g. in shell_server, shell_client)
- [x] I ensured this code runs with ASan / TSan or other static verification tools (TSan)

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12729/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12727/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12729/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12731/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12732/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12733/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12735/

arangod/Pregel/Worker.cpp
```diff
@@ -168,9 +168,16 @@ void Worker<V, E, M>::setupWorker() {
     // of time. Therefore this is performed asynchronous
     TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
     Scheduler* scheduler = SchedulerFeature::SCHEDULER;
-    auto self = shared_from_this();
-    bool queued = scheduler->queue(RequestLane::INTERNAL_LOW, [self, this, cb] {
-      _graphStore->loadShards(&_config, cb);
+    bool queued = scheduler->queue(RequestLane::INTERNAL_LOW, [self = shared_from_this(), this, cb] {
```
maierlars, line 171: This `shared_from_this` confused me, because the lambda captured `cb` which itself also holds a shared_reference to `this`. So you could drop one of them? In particular in the case of `lazyLoading()` `cb` is executed immediately. And afaik having `self` and `this` prevents small function optimization and thus allocated memory?

arangod/Pregel/Worker.cpp
```diff
@@ -778,7 +785,12 @@ void Worker<V, E, M>::finalizeRecovery(VPackSlice const& data) {
 
 template <typename V, typename E, typename M>
 void Worker<V, E, M>::_callConductor(std::string const& path, VPackBuilder const& message) {
-  if (ServerState::instance()->isRunningInCluster() == false) {
+  application_features::ApplicationServer& server = _config.vocbase()->server();
+  if (server.isStopping()) {
+    THROW_ARANGO_EXCEPTION(TRI_ERROR_SHUTTING_DOWN);
```
maierlars, line 790: Have you tried this? From my experience Pregel is not very exception safe... :disappointed:

maierlars (commented):
Just some random thoughts. Could all be wrong.

arangod/Pregel/Algos/EffectiveCloseness/EffectiveCloseness.cpp
```diff
@@ -94,15 +94,14 @@ struct ECGraphFormat : public GraphFormat<ECValue, int8_t> {
...
-  size_t estimatedEdgeSize() const override { return 0; };
+  size_t estimatedEdgeSize() const override { return 0; }
 
-  void copyVertexData(std::string const& documentId, arangodb::velocypack::Slice document,
-                      ECValue& targetPtr) override {}
+  void copyVertexData(std::string const& /*documentId*/, arangodb::velocypack::Slice /*document*/,
+                      ECValue& /*targetPtr*/, uint64_t& /*vertexIdRange*/) override {}
 
-  void copyEdgeData(arangodb::velocypack::Slice document, int8_t& targetPtr) override {}
+  void copyEdgeData(arangodb::velocypack::Slice /*document*/, int8_t& /*targetPtr*/) override {}
```
maierlars, line 102: One could add a empty default implementation in the base class. Its almost always empty.

arangod/Pregel/Graph.h
```diff
@@ -83,51 +82,103 @@ class Edge {
...
-  uint16_t _keyLength; // uint16_t
+
+  // the number of edges per vertex is limited to 4G.
+  // this should be more than enough
+  uint32_t _edgeCount; // uint32_t
+
+  // combined attribute fusing the active bit (as its highest significant bit)
+  // and the length of the key in its lowest 15 significants bits. we do this
+  // to save RAM/space (we may have _lots_ of vertices)
+  uint16_t _keyLengthAndActive; // uint16_t
```
maierlars, line 109: Wouldn't that be the perfect place for using bit fields? *hust*

arangod/IResearch/IResearchLink.h
```diff
@@ -96,16 +96,6 @@ class IResearchLink {
   void afterTruncate(TRI_voc_tick_t tick,
                      arangodb::transaction::Methods* trx); // arangodb::Index override
 
-  ////////////////////////////////////////////////////////////////////////////////
-  /// @brief insert a set of ArangoDB documents into an iResearch View using
-  ///        '_meta' params
-  /// @note arangodb::Index override
-  ////////////////////////////////////////////////////////////////////////////////
-  virtual void batchInsert(
```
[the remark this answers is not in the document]
jsteemann: No, they are unrelated. This just deletes the batchInsert methods that are not in use since MMFiles was removed.
Could be done in a seperate PR of course, but I think it doesn't do any harm removing them here.

arangod/Pregel/CommonFormats.h
```diff
@@ -117,6 +117,9 @@ struct SenderMessage {
   SenderMessage() {}
   SenderMessage(PregelID const& pid, T const& val)
       : senderId(pid), value(val) {}
+  
+  SenderMessage(PregelID&& pid, T const& val)
+      : senderId(std::move(pid)), value(val) {}
```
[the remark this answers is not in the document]
jsteemann: Ok, will change it.

arangod/Pregel/Graph.h
jsteemann: Maybe... I have never used them in practice, so I didn't think of this option at all.
Catching up with my knowledge deficits here, on cppreference.com I read that for bitfields it is not _guaranteed_ that adjacent members are packed tighly together:

> Multiple adjacent bit fields are usually packed together (although this behavior is implementation-defined):

It looks like if you want to guarantee tightly packed structs, you have to implement it yourself. 
But maybe all our target compilers are already good enough. We can figure this out by using bitfields and then using static_assert at compile time.

maierlars: :+1: for `static_assert`

arangod/Pregel/Algos/EffectiveCloseness/EffectiveCloseness.cpp
jsteemann: Ok, I will adjust the default implementation of `copyEdgeData` to do nothing, and remove the empty overrides for it.
`buildEdgeDocument` I already removed in this PR, because I could not find any override of it that would do anything.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12739/

arangod/Pregel/Worker.cpp
jsteemann: Fixed

arangod/Pregel/Graph.h
jsteemann: Added a bitfield and a static_assert. That will show us if our target compilers properly pack adjacent bitfield variables.

arangod/Pregel/Worker.cpp
jsteemann: Yes, tried it.
I agree overall it currently does not respond nicely to errors, exception, server shutdown. We should improve this further (in a seperate PR).

arangod/IResearch/IResearchLink.h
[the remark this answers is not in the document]
jsteemann: Marking this as resolved, as the deletions are intentional (though unrelated).

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12740/

arangod/IResearch/IResearchLink.h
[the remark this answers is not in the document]
danielhlarkin: Some of the `LocalTaskQueue` APIs changed, and since this code uses `LocalTaskQueue` but is completely obsolete, I figured remove it rather than update it.

jsteemann:
Tests blue

maierlars (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13236/

KVS85:
Tests blue.

KVS85 (approved):
LGTM.

PR #13017, opened 2020-11-11
don't double the timeout per test, but only have tests the double timeout

dothebart:


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12707/

dothebart:
tests ok except unrelated windows error.

jsteemann (approved):

PR #13018, opened 2020-11-11
Add deadlock test for hotbackup

neunhoef:
Port a test from 3.7 to devel.

This test does concurrently:
  - repeated hotbackups
  - AQL writing transactions with exclusive lock

### DIFF
--- a/tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
+++ b/tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
@@ -48,11 +48,7 @@ function trxWriteHotbackupDeadlock () {
     },
 
     tearDown: function () {
-      if (c !== null) {
-        c.drop();
-      }
-      c = null;
-      internal.wait(0);
+      db._drop(cn);
     },
 
     testRunAQLWritingTransactionsDuringHotbackup: function () {
@@ -81,7 +77,7 @@ function trxWriteHotbackupDeadlock () {
                         bad++;
                       }
                     }
-                    return {good,bad};
+                    return {good, bad};
                   }`,
           collections: {} }, {"x-arango-async":"store"});
       assertFalse(res.error, "Could not POST transaction.");
@@ -115,11 +111,13 @@ function trxWriteHotbackupDeadlock () {
           assertEqual(200, res.code, "Response code bad.");
           assertEqual(0, res.result.bad, "Not all hotbackups went through!");
           print("Managed to perform", res.result.good, "hotbackups.");
-          break;
+          return;
         }
         wait(1.0);
         print("Waiting for hotbackups to finish...");
       }
+      arango.DELETE(`/_api/job/${jobid}`);
+      assertFalse(true, "Did not finish in expected time.");
     }
   };
 
--- a/tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
+++ b/tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
@@ -1,0 +1,129 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global fail, assertTrue, assertFalse, assertEqual, assertNotUndefined, arango, print */
+
+// //////////////////////////////////////////////////////////////////////////////
+// / @brief ArangoDeadLockTests
+// /
+// /
+// / DISCLAIMER
+// /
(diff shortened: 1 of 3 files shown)


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12711/

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+      c = db._create(cn, {numberOfShards: 1, replicationFactor: 2},
+                     {waitForSyncReplication: true});
+    },
+
+    tearDown: function () {
+      if (c !== null) {
+        c.drop();
+      }
+      c = null;
+      internal.wait(0);
```
jsteemann, line 55:
```suggestion
      db._drop(cn);
```

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+        if (res.code !== 204) {
+          assertEqual(200, res.code, "Response code bad.");
+          assertEqual(0, res.result.bad, "Not all hotbackups went through!");
+          print("Managed to perform", res.result.good, "hotbackups.");
+          break;
+        }
+        wait(1.0);
+        print("Waiting for hotbackups to finish...");
+      }
+    }
```
jsteemann, line 123: I think we should have an assertion after the `while` loop to check that we didn't use the full 80 seconds.
In addition, we should add a `finally` block to kill/cancel the async job should it still be ongoing.

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+      let res = arango.POST_RAW("/_api/transaction",
+        { action: `function() {
+                    let console = require("console");
+                    let internal = require("internal");
+                    let wait = internal.wait;
+                    let time = internal.time;
+                    let start = time();
+                    let good = 0;
+                    let bad = 0;
+                    while (time() - start < 60) {
```
jsteemann, line 74: Suggestion: I would let the two threads communicate via a special document in the collection.
For example, the async job should do the following:
```
collection.insert({ _key: "hotbackup" });
while (!collection.exists("write")) {
  internal.sleep(0.1);
}
while (collection.exists("write")) {
  do hot backups as before...
}
```
and the main thread should do:
```
post hotbackup job
collection.insert({ _key: "write" });
while (!collection.exists("hotbackup")) {
  internal.sleep(0.1);
}
start write tests as before...
collection.remove("write");
```
That way we can ensure the two threads will run overlapping, and not one after the after (which theoretically can happen).

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
neunhoef: Done.

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
neunhoef: I find this rather theoretical. We post something asynchronously and then start immediately. If we assume that there could be a huge delay, then the same delay could happen when we check for the document and we would be back to square one. I do not think that this is worth the additional complexity. If you don't mind, I would not follow this suggestion.

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
neunhoef: Done.

tests/js/client/shell/shell-deadlock-trx-hotbackup-cluster.js
neunhoef: However, I have used this idea to stop the hotbackups early when all transactions are done, this decreases the test time considerably.

neunhoef: Unfortunately, the whole approach does not work since we are in a transactions and thus do not see any outside change. LOL. Reverted.

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12714/

neunhoef:
Tests green.

PR #13019, opened 2020-11-11
for code coverage we want all log-strings to be created and later discarded.

dothebart:


CMakeLists.txt
```diff
@@ -750,6 +750,14 @@ endif (ASM_OPTIMIZATIONS)
 ## MAINTAINER MODE
 ################################################################################
 
+option(UNCONDITIONALLY_BUILD_LOG_MESSAGES
+  "whether log strings should be generated regardles of active loglevel"
+  OFF
+  )
+if (USE_LOG_MAINTAINER_MODE)
```
jsteemann, line 757:
```suggestion
option(UNCONDITIONALLY_BUILD_LOG_MESSAGES
  "whether log strings should be generated regardless of active loglevel"
  OFF
  )
if (UNCONDITIONALLY_BUILD_LOG_MESSAGES)
```

lib/Logger/LogMacros.h
```diff
@@ -58,16 +58,36 @@
...
 
 #define ARANGO_INTERNAL_LOG_HELPER(id)                        \
   ::arangodb::Logger::LINE(__LINE__)                          \
   << ::arangodb::Logger::FILE(__FILE__)                       \
   << ::arangodb::Logger::FUNCTION(__FUNCTION__)               \
   << ::arangodb::Logger::LOGID((id))
 
+#define ARANGODB_UNCONDITIONALLY_BUILD_LOG_MESSAGES 1
+#if ARANGODB_UNCONDITIONALLY_BUILD_LOG_MESSAGES
+// these versions *allways* generate the log string, only to discard it
```
jsteemann, line 70:
```suggestion
// these versions *always* generate the log string, only to discard it
```

lib/Logger/LogMacros.h
```diff
@@ -58,16 +58,36 @@
...
+// just before logging them. This is good for ASAN or coverage.
+// DONT use in production.
+#define LOG_TOPIC(id, level, topic)                                     \
+  ::arangodb::LoggerStream() <<  (::arangodb::LogLevel::level)          \
+                             << (topic)                                 \
+                             << ARANGO_INTERNAL_LOG_HELPER(id)
+
+#define LOG_TOPIC_IF(id, level, topic, cond)                            \
+  false && (cond)                                                       \
```
jsteemann, line 79: Now this will never execute `(cond)` which could again lower coverage etc.
It is also not safe to always generate the log message regardless of whether the condition is true.
Consider:
```
LOG_TOPIC_IF("abcde", WARN, Logger::FIXME, (result != nullptr)) << "RESULT: " << result->thisWillCrashNow();
```

```suggestion
  !(cond)                                                       \
```

goedderz:
My first intuition is that this is not a good idea. If the log string is not created because the condition is false, that means there's a code path that's never tested. A real improvement to test coverage would be to extend the tests so this path is executed. This change mostly just covers it up. And as @jsteemann pointed out, it may break things - or it may as well work when the condition is false, but not when its true.

I'm aware that we can't make any universal statement about whether this is a good idea or not, and it will vary in different individual cases. However, I think overall its better to leave it be.

dothebart:
In the end this patch makes them being produced - and tested - hence coverage isn't betrayed.
If something is really wrong here, it will show up in these testruns. 
The suggestion was, to also enable this in i.e. ASAN runs, so if it has uninitialized memory access, it will show up. 

I modified the `IF` macro to also always be executed, but log TRACE so it will also be discarded later on.

lib/Logger/LogMacros.h
```diff
@@ -58,16 +58,35 @@
...
+  ::arangodb::LoggerStream() << (::arangodb::LogLevel::level)           \
+                             << (topic)                                 \
+                             << ARANGO_INTERNAL_LOG_HELPER(id)
+
+#define LOG_TOPIC_IF(id, level, topic, cond)                            \
+  ::arangodb::LoggerStream()                                            \
+  << ::arangodb::LogLevel::TRACE                                        \
+  << (topic)                                                            \
+  << ARANGO_INTERNAL_LOG_HELPER(id)                                     \
+  << (true && cond)
```
jsteemann, line 83: Macro parameters containing arbitrary expressions should be put inside parentheses.
Additionally, I don't see the value of `true && cond` here. Shouldn't it be just `(cond)`?
```suggestion
  << (cond)
```

goedderz:
> In the end this patch makes them being produced - and tested - hence coverage isn't betrayed.
> If something is really wrong here, it will show up in these testruns.
> The suggestion was, to also enable this in i.e. ASAN runs, so if it has uninitialized memory access, it will show up.
> 
> I modified the `IF` macro to also always be executed, but log TRACE so it will also be discarded later on.

That sounds sensible.

dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12801/

jsteemann (approved):

dothebart:
testrun blue.

fceller (approved):

fceller (approved):
LGTM

PR #13031, opened 2020-11-13
make test fit to be run on pre-configured replication, and dc2dc

dothebart:
### Scope & Purpose

The test now can be run in more scenarios:
 - the SUT has replication already configured, the test shouldn't manipulate it
 - DC2DC:
    - collection renaming isn't supported in clusters
    - instead of checking the replication applier as detection for whether the slave picked up, 
      we count (non-system)-collections and their documents. If counts are equal within a certain
      timeframe, the follower-cluster is considered "in sync" and thorough test status is created.

- [ ] :hankey: Bugfix 
- [x] :pizza: New test feature 

#### Backports:

- [x] No backports required

#### Related Information
https://github.com/arangodb/release-test-automation/pull/61 - here it is used too.

### Testing & Verification

- [x] This is an enhancement to the integration tests.

Link to Jenkins PR run:


dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12746/

dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12772/

tests/js/server/replication/fuzz/replication-fuzz-global.js
```diff
@@ -59,7 +62,12 @@ function ReplicationSuite() {
   };
 
   var collectionChecksum = function(name) {
-    return db._collection(name).checksum(false, true).checksum;
+    if (isCluster) {
+      let csa = db._query('RETURN md5(FOR doc IN @@col SORT doc._key RETURN [doc._key, doc._rev])', {'@col':name}).toArray();
```
jsteemann, line 66: This will be expensive to run on large collections.
What is the approximate size of data we will run this on?

jsteemann (approved):

tests/js/server/replication/fuzz/replication-fuzz-global.js
dothebart: we currently don't have another way to tell collections are identical on both sides in the cluster case. Since this needs to be running on 3.6 as well (and is not triggered via testing.js atm) I think we should leave it that way until we get a better idea.

dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13038/ - testrun blue.

PR #13049, opened 2020-11-17
Bug fix 3.6/cleanup 20201117

jsteemann:
### Scope & Purpose

This is a multi-purpose PR:

It makes the number of network I/O threads properly configurable via the startup option `--network.io-threads`. This option existed before, but its configured value was effectively clamped to a value of `1`. 
This change requires forward ports to 3.7 and devel.

It also disables cluster maintenance background threads on coordinators.
The maintenance threads don't have any purpose on coordinators, so there is no need to start them there. This saves 2 or more background threads on coordinators. These threads should have mostly been idle before, so saving them will not have a huge impact except saving a few MB of thread stack memory.
The maintenance threads are already turned off on coordinators in 3.7 and devel for a while, so this does not need to be ported.

In addition this PR removes the `_allowDirectExecution` instance variable from RestHandlers, which in 3.6 was only set but never read back. This variable lost its meaning at point in the 3.5 cycle, so it is now useless and can be removed.
The instance variable has already been removed in 3.7 and devel a while ago, so this does not need to be ported.

#### Backports:

- [x] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12786/

### DIFF
--- a/arangod/Network/NetworkFeature.cpp
+++ b/arangod/Network/NetworkFeature.cpp
@@ -104,7 +104,7 @@ void NetworkFeature::collectOptions(std::shared_ptr<options::ProgramOptions> opt
 }
 
 void NetworkFeature::validateOptions(std::shared_ptr<options::ProgramOptions>) {
-  _numIOThreads = std::min<unsigned>(1, std::max<unsigned>(_numIOThreads, 8));
+  _numIOThreads = std::max<unsigned>(1, std::min<unsigned>(_numIOThreads, 8));
   if (_maxOpenConnections < 8) {
     _maxOpenConnections = 8;
   }
--- a/arangod/Network/NetworkFeature.cpp
+++ b/arangod/Network/NetworkFeature.cpp
@@ -127,7 +127,6 @@ void NetworkFeature::prepare() {
   config.verifyHosts = _verifyHosts;
   config.clusterInfo = ci;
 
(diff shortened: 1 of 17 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12788/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12803/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12806/

arangod/Network/NetworkFeature.cpp
```diff
@@ -127,6 +127,7 @@ void NetworkFeature::prepare() {
   config.verifyHosts = _verifyHosts;
   config.clusterInfo = ci;
 
+  LOG_DEVEL << "NUM: " << _numIOThreads;
```
neunhoef, line 130: This should be removed before merging.

neunhoef (changes requested):
LGTM. I found one LOG_DEVEL, which should be removed.

arangod/Network/NetworkFeature.cpp
jsteemann, line 130:
```suggestion
```

fceller (approved):

KVS85 (approved):
LGTM.

PR #13058, opened 2020-11-18
Bug fix 3.7/aql cluster query shutdown

jsteemann:
### Scope & Purpose

Change cluster AQL query shutdown on coordinators to be less blocking.

The shutdown procedure for coordinator queries executes a busy wait loop in the destructor of an aql::Query, trying to delete all ExecutionEngine snippets from the query registry that are currently not in use by any concurrent operation on the same query. 
If one of the snippets was busy, the shutdown code couldn't make progress and continued trying to shut down. This effectively blocked the shutdown thread until all concurrent operations on the query had finished. This was problematic when the concurrent operations executing on the query were also waiting for other threads to make progress.

This behavior has been improved now, so that the shutdown operation only blocks while a concurrent operation on the query is currently ongoing and not waiting for other threads. In practice, this should reduce wait times and remove any multi-level shutdown dependencies.

The query shutdown procedure in 3.6 is vastly different, so this does not need a 3.6 (or 3.5) backport, but a forward-port to devel.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/589

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_server_aql --cluster true*


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12834/

jsteemann:
This PR's approach is superseded by https://github.com/arangodb/arangodb/pull/14755.

PR #13066, opened 2020-11-18
Fuerte shutdown TLS fix and jeprof support.

fceller:
Manual & Jenkins tested

### DIFF
--- a/3rdParty/fuerte/src/AsioSockets.h
+++ b/3rdParty/fuerte/src/AsioSockets.h
@@ -27,7 +27,7 @@
 #include <fuerte/loop.h>
 
 namespace arangodb { namespace fuerte { inline namespace v1 {
-  
+
 namespace {
 template <typename SocketT, typename F>
 void resolveConnect(detail::ConnectionConfiguration const& config,
@@ -39,7 +39,7 @@ void resolveConnect(detail::ConnectionConfiguration const& config,
       done(ec);
       return;
     }
-    
+
     // A successful resolve operation is guaranteed to pass a
     // non-empty range to the handler.
     asio_ns::async_connect(socket, it,
@@ -47,7 +47,7 @@ void resolveConnect(detail::ConnectionConfiguration const& config,
       std::forward<F>(done)(ec);
     });
   };
-  
+
   // windows does not like async_resolve
 #ifdef _WIN32
   asio_ns::error_code ec;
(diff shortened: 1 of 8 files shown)


jsteemann (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12833/

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -137,12 +144,45 @@ struct Socket<fuerte::SocketType::Ssl> {
...
+      asio_ns::error_code ec; // ignored
+      socket.lowest_layer().cancel(ec);
+      // Set a timeout, since the shutdown could block forever, if we
+      // reach the timeout, we simply give up on TLS and close the TCP
+      // layer. Note the cleanupDone is necessary, since it is possible
+      // that both the completion handler of the timeout and the one
+      // of the async_shutdown are called without error!
+      timer.expires_from_now(std::chrono::seconds(3));
+      timer.async_wait([connection, this](asio_ns::error_code ec) {
+        if (!cleanupDone && !ec) {
```
mpoeter, line 164: Is it possible that the callbacks for `async_wait` and `async_shutdown` are called concurrently by different IO threads? In that case we would have a race here since `cleanupDone` is only set after the call to `shutdownTcp`.

KVS85:
Tests blue except unrelated failures.

3rdParty/fuerte/src/AsioSockets.h
neunhoef: As far as I know we only ever put one thread in an io_context, so that this cannot happen.

danielhlarkin: Well, currently yes, we only put one in. But we just made it so that you can use more: https://github.com/arangodb/arangodb/pull/13049

mpoeter: @danielhlarkin Yes, we may have multiple IO threads, but each thread has its own io_context. And since the timer and socket are both on the same context, the callbacks cannot run concurrently. So this should be fine.

danielhlarkin: Ah, okay, I hadn't dug into the details there. Good.

PR #13087, opened 2020-11-24
provide a cross-database list of current/slow queries

jsteemann:
### Scope & Purpose

Added support for fetching the list of currently running and slow AQL queries from all databases at once, by adding an `all` parameter to the following query APIs:

* `require("@arangodb/aql/queries").current({ all: true })`: will return the currently running queries from all databases, not just the currently selected database.
* HTTP GET `/_api/query/current?all=true`: same, but for the HTTP REST API.
* `require("@arangodb/aql/queries").slow({ all: true })`: will return the slow query history from all databases, not just the currently selected database.
* HTTP GET `/_api/query/slow?all=true`: same, but for the HTTP REST API.
* `require("@arangodb/aql/queries").clearSlow({ all: true })`: will clear the slow query history for all databases, not just the currently selected database.
* HTTP DELETE `/_api/query/slow?all=true`: same, but for the HTTP REST API.

Using the `all` parameter is only allowed when making the call inside the `_system` database and with superuser privileges.

This PR also contains the fix contained in https://github.com/arangodb/arangodb/pull/13190.

#### Backports:

- [ ] No backports required
- [x] Backports required for: 3.7, 3.6

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in shell_client, server_permissions)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12865/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12878/

jsteemann:
Tests blue

arangod/VocBase/Methods/Queries.cpp
```diff
@@ -60,16 +63,71 @@ network::Headers buildHeaders() {
...
+  if (res.fail()) {
+    return res;
+  }
+
+  arangodb::DatabaseFeature& databaseFeature = vocbase.server().getFeature<DatabaseFeature>();
+
+  std::vector<arangodb::aql::QueryEntryCopy> queries;
   
   // local case
+  if (action == "slow") {
```
danielhlarkin, line 102: Should we replace this with an enum? Or at least a `StaticString`?

jsteemann: Ok, I have turned this into an enum.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13104

jsteemann:
Tests blue

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13135/

jsteemann:
Tests blue

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13170/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13177/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13201/

arangod/VocBase/Methods/Queries.cpp
```diff
@@ -60,16 +64,77 @@ network::Headers buildHeaders() {
   return headers;
 }
 
-/// @brief return the list of currently running queries
-void getQueries(TRI_vocbase_t& vocbase, std::vector<aql::QueryEntryCopy> const& queries,
-                velocypack::Builder& out, char const* action, bool fanout) {
-  out.openArray();
+arangodb::Result checkAuthorization(TRI_vocbase_t& vocbase, bool allDatabases) {
```
dothebart, line 67: hm, can we find a less generic name for this function? and should it be static?

dothebart (approved):
please rething that function name, else LGTM.

arangod/VocBase/Methods/Queries.cpp
jsteemann: It does not need to be static, because it is already in the anonymous namespace of the TU.
That will not make it visible to anything else but the translation unit, so it is the same as static linkage.

danielhlarkin: It does seem to be specifically authorizing for access to the query list, so I guess it could be `checkAuthorizationForQueryList` if you really wanted to be more specific. But I think given the context it's not really necessary.

PR #13092, opened 2020-11-25
use -Wsuggest-override option conditionally for clang

jsteemann:
### Scope & Purpose

Use compiler option `-Wsuggest-override` option only conditionally for clang, because it is only supported from version 11 onwards. Older versions don't have it and produce warnings.
Apple clang 12.0 does not seem to have it either, so disable it there too.

This hopefully fixes a glitch introduced only recently in devel, so I am intentionally not making a CHANGELOG entry for this.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *compilation with clang in Oskar*.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12872/


fceller (approved):
LGTM

mpoeter (approved):

PR #13094, opened 2020-11-25
Bug fix/executionblockimpl hasmore assertion

mchacki:
### Scope & Purpose

This PR fixes a false assertion in AQL ExecutionBlocks.
The assertion tests that something should be written in an AQL block if an executor is finished with it's work and claims there is even more data available.
This assertion tested for the actually existence of a value entry. If this value is NOT used later, it could actually be optimized away here, causing this assertion to trigger. (An empty row is written, but no value)
Now using the number of written rows.
This covers empty rows as well.

This PR does NOT change any production behavior and therefor has no CHANGELOG entry. The code modified here is not compiled into our release.

Jenkins:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12883/

### DIFF
--- a/tests/js/server/aql/aql-shadowrow-block-return-regression.js
+++ b/tests/js/server/aql/aql-shadowrow-block-return-regression.js
@@ -291,7 +291,7 @@ function nonMaterializingViewRegressionSuite() {
       // and the consumer (Subquery) cannot passthrough the Block.
       // Hence all optimzation is turned off to guarantee these two stay connected.
       const query = `FOR d IN ${vname} OPTIONS {waitForSync:true} LET p = (RETURN 1 + 1) RETURN 1`;
-      const res = db._query(query, {}, {optimizer: {rules: ["-all", "+spliced-subqueries"] }});
+      const res = db._query(query, {}, {optimizer: {rules: ["-all", "+splice-subqueries"] }});
       // This result was never wrong, the internal assertion is tested. 
       assertEqual(res.toArray().length, 1010);
     }
--- a/tests/js/server/aql/aql-shadowrow-block-return-regression.js
+++ b/tests/js/server/aql/aql-shadowrow-block-return-regression.js
@@ -250,12 +250,61 @@ function collectionInSubqeryRegressionSuite() {
     }
   };
 }
+
+function nonMaterializingViewRegressionSuite() {
+  const cname = "UnitTestCollection";
+  const vname = "UnitTestView";
+
+  const cleanup = () => {
+    try {
+      db._drop(cname);
+    } catch(e) {}
+    try {
+      db._dropView(vname);
+    } catch(e) {}
+  };
+
+  return {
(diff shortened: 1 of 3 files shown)


tests/js/server/aql/aql-shadowrow-block-return-regression.js
```diff
@@ -250,12 +250,61 @@ function collectionInSubqeryRegressionSuite() {
...
+
+    testRowsWithoutRegisters: function() {
+      // This tests a logic assertion in the ExecutionBlock.
+      // If a Block (view node) HASMORE it is required to write something
+      // In this Query `d` is optimized away, so the Block will write
+      // empty rows only.
+      // This test requires that the Producer (View) can optimize away its output
+      // and the consumer (Subquery) cannot passthrough the Block.
+      // Hence all optimzation is turned off to guarantee these two stay connected.
+      const query = `FOR d IN hv LET p = (RETURN 1 + 1) RETURN 1`;
```
Dronplane, line 293:
```suggestion
      const query = `FOR d IN hv OPTIONS {waitForSync:true} LET p = (RETURN 1 + 1) RETURN 1`;
```
Without this we could end up with less than batch size records in view,  hence violating test`s prerequisites

Dronplane (changes requested):

tests/js/server/aql/aql-shadowrow-block-return-regression.js
```diff
@@ -250,12 +250,60 @@ function collectionInSubqeryRegressionSuite() {
...
+    testRowsWithoutRegisters: function() {
+      // This tests a logic assertion in the ExecutionBlock.
+      // If a Block (view node) HASMORE it is required to write something
+      // In this Query `d` is optimized away, so the Block will write
+      // empty rows only.
+      // This test requires that the Producer (View) can optimize away its output
+      // and the consumer (Subquery) cannot passthrough the Block.
+      // Hence all optimzation is turned off to guarantee these two stay connected.
+      const query = `FOR d IN hv OPTIONS {waitForSync:true} LET p = (RETURN 1 + 1) RETURN 1`;
+      const res = db._query(query, {}, {optimizer: {rules: ["-all", "+spliced-subqueries"] }});
```
Dronplane, line 294:
```suggestion
      const res = db._query(query, {}, {optimizer: {rules: ["+spliced-subqueries"] }});
```
We still need to run ArangoSearch optimization rules

Dronplane: Just checked - I`m wrong. no materialization is hadled by `handle-arangosearch-views` rule which is non-disablable. So It is ok to have -all

Dronplane (approved):
LGTM

mchacki:
Latest Jenkins:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/12890/

Full blue.

PR #13105, opened 2020-11-27
Bug fix 3.7/executionblockimpl hasmore assertion

mchacki:
Backport of
https://github.com/arangodb/arangodb/pull/13094

### Scope & Purpose

This PR fixes a false assertion in AQL ExecutionBlocks.
The assertion tests that something should be written in an AQL block if an executor is finished with it's work and claims there is even more data available.
This assertion tested for the actually existence of a value entry. If this value is NOT used later, it could actually be optimized away here, causing this assertion to trigger. (An empty row is written, but no value)
Now using the number of written rows.
This covers empty rows as well.

This PR does NOT change any production behavior and therefor has no CHANGELOG entry. The code modified here is not compiled into our release.

Jenkins:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/12897/


Dronplane (approved):
LGTM

mchacki:
Jenkins:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/12899/
All blue

PR #13047, opened 2020-11-16
fixed gauge assignment operatorsd

kvahed:
### Scope & Purpose

*(Please describe the changes in this PR for reviewers - **mandatory**)*

- [X] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [X] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [ ] No backports required
- [X] Backports required for: *3.7*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [X] This change is a trivial rework / code cleanup without any test coverage.
- [X] The behavior in this PR was *manually tested*
- [X] This change is already covered by existing tests, such as *Guage*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:
https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/12815/

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

Please note that for legal reasons we require you to sign the [Contributor Agreement](https://www.arangodb.com/documents/cla.pdf)
before we can accept your pull requests.

PR #6196, opened 2018-08-20
Batch Handler V8-lane bug

lamai93:
When handling a batch request, queue each job with priority given by its lane instead of using the slowest lane for all.
Make the Batch handler aware of Handler that do not return a result immediately.

- [x] Execute each handler with its own priority.
- [x] Support waiting rest handlers.
- [ ] Execute all handlers in parallel.
- [ ] Clean up

PR #13108, opened 2020-11-27
Cluster overwhelm countermeasures

neunhoef:
This PR combines several measures. It tries to implement a whole lot
of countermeasures for bad behaviour of the cluster during high load
situations. The document to discuss these countermeasures is here:

    https://github.com/arangodb/documents/blob/master/DesignDocuments/02_PLANNING/FightClusterOverwhelm.md

This started as a document to collect evidence and problems, but
also ideas for counter measures. Now, it serves as a design
document for the changes actually done. A lot of discussions and
testing will still be required.

Enterprise PR: https://github.com/arangodb/enterprise/pull/600

### Scope & Purpose

- [*] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)

#### Backports:

- [ ] No backports required
- [*] Backports required for: *(Please specify versions and link PRs)*
      We need to discuss how much of this is needed in 3.7 and 3.6, since the behaviour of clusters under load is pretty bad in all versions.

#### Related Information

- [ ] Docs PR: 
- [*] Enterprise PR: https://github.com/arangodb/enterprise/pull/600
- [*] Design document: https://github.com/arangodb/documents/blob/master/DesignDocuments/02_PLANNING/FightClusterOverwhelm.md

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13136/


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12903/

neunhoef:
Note that this currently does compile on Windows with some stupid initializer list error. Will fix next week.

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12955/

neunhoef:
Still expecting Windows compile failures, will fix with the results of this run.

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12956/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12977/

danielhlarkin:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12982/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/12989/

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -305,11 +305,13 @@ bool FollowerInfo::updateFailoverCandidates() {
 // All followers can return as soon as the lock is released
 #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
     TRI_ASSERT(_failoverCandidates->size() == _followers->size());
+#if 0
```
jsteemann, line 308: I guess the changes in this file are temporary workarounds and should be removed from this PR.

arangod/Cluster/ReplicationTimeoutFeature.cpp
```diff
@@ -44,10 +56,15 @@ ReplicationTimeoutFeature::ReplicationTimeoutFeature(application_features::Appli
...
-  options->addOption(
-      "--cluster.synchronous-replication-timeout-minimum",
-      "all synchronous replication timeouts will be at least this value (in seconds)",
-      new DoubleParameter(&lowerLimit));
+  options->addOption("--cluster.synchronous-replication-timeout-minimum",
+                     "all synchronous replication timeouts will be at least "
+                     "this value (in seconds)",
+                     new DoubleParameter(&lowerLimit));
+
+  options->addOption("--cluster.synchronous-replication-timeout-maximum",
```
jsteemann, line 64: There doesn't seem to be a validation in place via `validateOptions(...)` that `lowerLimit` <= `upperLimit`

arangod/Cluster/FollowerInfo.cpp
neunhoef: Yes, there is a PR out there which fixes this and once merged, this will work again.

arangod/Cluster/ReplicationTimeoutFeature.cpp
neunhoef: Correct, this needs fixing.

arangod/Network/NetworkFeature.cpp
```diff
@@ -114,6 +132,12 @@ void NetworkFeature::collectOptions(std::shared_ptr<options::ProgramOptions> opt
   options->addOption("--network.protocol", "network protocol to use for cluster-internal communication",
                      new DiscreteValuesParameter<StringParameter>(&_protocol, protos))
                      .setIntroducedIn(30700);
+
+  options
+      ->addOption("--network.max-requests-in-flight", std::string("controls the number of internal requests that can be in flight at a given point in time"),
```
jsteemann, line 137: the extra `std::string(...)` wrapper here works but can be omitted.

arangod/Network/NetworkFeature.cpp
```diff
@@ -114,6 +132,12 @@ void NetworkFeature::collectOptions(std::shared_ptr<options::ProgramOptions> opt
...
+                  new options::UInt64Parameter(&_maxInFlight),
+                  options::makeDefaultFlags(options::Flags::Dynamic, options::Flags::Hidden))
```
jsteemann, line 139: I think the `Dynamic` option is not accurate here, as the default value for the option is hard-coded and not determined dynamically via some properties of the host system.

arangod/Network/NetworkFeature.cpp
```diff
@@ -238,4 +264,59 @@ void NetworkFeature::trackForwardedRequest() {
...
+  return _requestsInFlight.load();
+}
+
+bool NetworkFeature::isCongested() const {
+  return _requestsInFlight.load() >= (_maxInFlight * ::CongestionRatio);
+}
+
+bool NetworkFeature::isSaturated() const {
+  return _requestsInFlight.load() >= _maxInFlight;
+}
```
jsteemann, line 277: All of these use full sequential consistency memory ordering. Can we relax this a bit?

arangod/Network/NetworkFeature.cpp
```diff
@@ -238,4 +264,59 @@ void NetworkFeature::trackForwardedRequest() {
...
+                                         std::unique_ptr<fuerte::Request> req,
+                                         std::unique_ptr<fuerte::Response> res) {
+                      finishRequest(pool, err, req, res);
+                      cb(err, std::move(req), std::move(res));
+                    });
+}
+
+void NetworkFeature::prepareRequest(network::ConnectionPool const& pool,
+                                    std::unique_ptr<fuerte::Request>& req) {
+  _requestsInFlight += 1;
```
jsteemann, line 298: Can we relax the memory ordering here, e.g.
```suggestion
  _requestsInFlight.fetch_add(1, std::memory_order_relaxed);
```

arangod/Network/NetworkFeature.cpp
```diff
@@ -238,4 +264,59 @@ void NetworkFeature::trackForwardedRequest() {
...
+  if (req) {
+    req->timestamp(std::chrono::steady_clock::now());
+  }
+}
+void NetworkFeature::finishRequest(network::ConnectionPool const& pool, fuerte::Error err,
+                                   std::unique_ptr<fuerte::Request> const& req,
+                                   std::unique_ptr<fuerte::Response>& res) {
+  _requestsInFlight -= 1;
```
jsteemann, line 306: Can we relax the memory ordering here, e.g.
```suggestion
  _requestsInFlight.fetch_sub(1, std::memory_order_relaxed);
```
I also would like to see an assertion for the value not underflowing here.

arangod/Network/NetworkFeature.cpp
```diff
@@ -238,4 +264,59 @@ void NetworkFeature::trackForwardedRequest() {
...
+    _requestTimeouts.count();
+  } else if (req && res) {
+    res->timestamp(std::chrono::steady_clock::now());
+    std::chrono::milliseconds duration =
+        std::chrono::duration_cast<std::chrono::milliseconds>(res->timestamp() -
+                                                              req->timestamp());
+    std::chrono::milliseconds timeout = req->timeout();
+    double percentage = std::clamp(100.0 * static_cast<double>(duration.count()) /
+                                       static_cast<double>(timeout.count()),
+                                   0.0, 100.0);
```
jsteemann, line 317: Is it possible that `timeout` somehow is 0 here? In this case we could get a SIGFPE.

arangod/Replication/ReplicationMetricsFeature.cpp
```diff
@@ -33,94 +33,96 @@ namespace arangodb {
...
-      _waitedForTailingApply(
-        server.getFeature<arangodb::MetricsFeature>().counter(
+      _waitedForTailingApply(server.getFeature<arangodb::MetricsFeature>().counter(
           "arangodb_replication_tailing_apply_time", 0,
-          "Time needed to apply replication tailing data [ms]")) {
+          "Time needed to apply replication tailing data [ms]")),
+      _syncTimeTotal(server.getFeature<arangodb::MetricsFeature>().counter(
+          "arangodb_replication_synchronous_requests_total_time", 0,
+          "Total time needed for all synchronous replication requests "
+          "accumulated.")),
```
jsteemann, line 121: Let's add the unit, so it is more understandable for the end user.
```suggestion
          "accumulated [ns]")),
```

arangod/Replication/ReplicationMetricsFeature.cpp
```diff
@@ -33,94 +33,96 @@ namespace arangodb {
...
+      _syncOpsTotal(server.getFeature<arangodb::MetricsFeature>().counter(
+          "arangodb_replication_synchronous_requests_total_number", 0,
+          "Total number of all synchronous replication requests "
+          "accumulated.")) {
```
jsteemann, line 125:
```suggestion
          "accumulated")) {
```

arangod/Replication/ReplicationMetricsFeature.cpp
jsteemann, line 121: "Total" and "accumulated" seems to be redundant.

arangod/Replication/ReplicationMetricsFeature.cpp
jsteemann, line 125: "Total" and "accumulated" seems to be redundant.

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -316,31 +317,34 @@ RestStatus RestDocumentHandler::readSingleDocument(bool generateBody) {
...
-            generateNotModified(rid);
-            return;
-          }
-        }
-
-        // use default options
-        generateDocument(opRes.slice(), generateBody,
-                         _activeTrx->transactionContextPtr()->getVPackOptions());
-      }));
+      _activeTrx->documentAsync(collection, search, options)
```
jsteemann, line 320: Wondering if this is actually necessary.
This is a read-only operation, so we don't need to commit anything locally. The transaction can just be aborted. It still makes sense if this can be called from a mixed (RW) transaction or if this actually does communication with followers.
If it is just a local operation, we don't need to make it async by all means.

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -677,21 +681,20 @@ RestStatus RestDocumentHandler::readManyDocuments() {
...
-  return waitForFuture(_activeTrx->documentAsync(cname, search, opOptions)
-                       .thenValue([=](OperationResult opRes) {
-    auto res = _activeTrx->finish(opRes.result);
-
-    if (opRes.fail()) {
-      generateTransactionError(cname, opRes);
-      return;
-    }
+  return waitForFuture(_activeTrx->documentAsync(cname, search, opOptions).thenValue([=](OperationResult opRes) {
+    return _activeTrx->finishAsync(opRes.result).thenValue([=, opRes(std::move(opRes))](Result&& res) {
```
jsteemann, line 685: Same here. If this is just a local operation, there is no need to make it async. If it does communication with followers, it is necessary.

arangod/Scheduler/SchedulerFeature.cpp
```diff
@@ -61,8 +61,8 @@ size_t defaultNumberOfThreads() {
   // use two times the number of hardware threads as the default
   size_t result = arangodb::NumberOfCores::getValue() * 2;
   // but only if higher than 64. otherwise use a default minimum value of 64
```
jsteemann, line 63: Comment is outdated now
```suggestion
  // but only if higher than 64. otherwise use a default minimum value of 32
```

arangod/Scheduler/SchedulerFeature.cpp
```diff
@@ -104,8 +104,16 @@ void SchedulerFeature::collectOptions(std::shared_ptr<options::ProgramOptions> o
                      new UInt64Parameter(&_nrMinimalThreads),
                      arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden));
 
+    // max / min number of threads
+
+  // Concurrency throttling:
+  options->addOption("--server.ongoing-low-priority-multiplier", std::string("controls the number of low prio requests that can be ongoing at a given point in time, relative to the maximum number of request handling threads"),
```
jsteemann, line 110: Wrapping this in an `std::string(...)` works but is not necessary here.

arangod/Scheduler/SchedulerFeature.cpp
```diff
@@ -146,12 +158,19 @@ void SchedulerFeature::validateOptions(std::shared_ptr<options::ProgramOptions>
...
     LOG_TOPIC("bf034", WARN, arangodb::Logger::THREADS)
-        << "--server.minimal-threads (" << _nrMinimalThreads << ") should be at least 2";
-    _nrMinimalThreads = 2;
+        << "--server.minimal-threads (" << _nrMinimalThreads << ") must be at least 4";
+    _nrMinimalThreads = 4;
   }
 
+  if (_ongoingLowPriorityMultiplier < 1.0) {
+    LOG_TOPIC("0a93a", WARN, arangodb::Logger::THREADS)
+        << "--server.ongoing-multiplier (" << _ongoingLowPriorityMultiplier
```
jsteemann, line 169:
```suggestion
        << "--server.ongoing-low-priority-multiplier (" << _ongoingLowPriorityMultiplier
```

arangod/Scheduler/SchedulerFeature.cpp
```diff
@@ -104,8 +104,16 @@ void SchedulerFeature::collectOptions(std::shared_ptr<options::ProgramOptions> o
                      new UInt64Parameter(&_nrMinimalThreads),
                      arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden));
 
+    // max / min number of threads
+
+  // Concurrency throttling:
+  options->addOption("--server.ongoing-low-priority-multiplier", std::string("controls the number of low prio requests that can be ongoing at a given point in time, relative to the maximum number of request handling threads"),
+                     new DoubleParameter(&_ongoingLowPriorityMultiplier),
+                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Dynamic,
```
jsteemann, line 112: I think the default value for this option is hard-coded, so the `Dynamic` attribute does not seem to be justified here.

arangod/Cluster/ReplicationTimeoutFeature.cpp
```diff
@@ -44,10 +56,15 @@ ReplicationTimeoutFeature::ReplicationTimeoutFeature(application_features::Appli
...
+                     "all synchronous replication timeouts will be at most "
+                     "this value (in seconds)",
+                     new DoubleParameter(&upperLimit));
```
jsteemann, line 67: As this is a new option, we should also add a `.setIntroducedIn(30800);` here.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -155,11 +156,13 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
       _jobsDequeued(0),
       _jobsDone(0),
-      _minNumWorkers(minThreads),
-      _maxNumWorkers(maxThreads),
-      _maxFifoSize(maxQueueSize),
-      _fifo1Size(fifo1Size),
-      _fifo2Size(fifo2Size),
+      _minNumWorker(minThreads),
+      _maxNumWorker(maxThreads),
+      _maxFifoSizes{maxQueueSize, fifo1Size, fifo1Size, fifo3Size},
```
jsteemann, line 161:
```suggestion
      _maxFifoSizes{maxQueueSize, fifo1Size, fifo2Size, fifo3Size},
```

arangod/Scheduler/SupervisedScheduler.h
```diff
@@ -138,18 +143,27 @@ class SupervisedScheduler final : public Scheduler {
...
   alignas(64) std::atomic<uint64_t> _jobsDequeued;
   alignas(64) std::atomic<uint64_t> _jobsDone;
 
-  uint64_t const _minNumWorkers;
-  uint64_t const _maxNumWorkers;
-  uint64_t const _maxFifoSize;
-  uint64_t const _fifo1Size;
-  uint64_t const _fifo2Size;
+  size_t const _minNumWorker;
+  size_t const _maxNumWorker;
```
jsteemann, line 153:
```suggestion
  size_t const _minNumWorkers;
  size_t const _maxNumWorkers;
```

arangod/Scheduler/SupervisedScheduler.h
```diff
@@ -138,18 +143,27 @@ class SupervisedScheduler final : public Scheduler {
...
+  uint64_t const _maxFifoSizes[NumberOfQueues];
+  size_t const _ongoingLowPriorityLimit;
+
+  // During a queue operation there a two reasons to manually wake up a worker
```
jsteemann, line 157:
```suggestion
  // During a queue operation there are two reasons to manually wake up a worker
```

arangod/GeneralServer/CommTask.cpp
```diff
@@ -480,15 +480,23 @@ bool CommTask::handleRequestSync(std::shared_ptr<RestHandler> handler) {
...
   ContentType respType = handler->request()->contentTypeResponse();
   uint64_t mid = handler->messageId();
 
   // queue the operation in the scheduler, and make it eligible for direct execution
   // only if the current CommTask type allows it (HttpCommTask: yes, CommTask: no)
   // and there is currently only a single client handled by the IoContext
-  auto cb = [self = shared_from_this(), handler = std::move(handler)]() mutable {
+  auto cb = [self = shared_from_this(), handler = std::move(handler), prio]() mutable {
+    if (prio == RequestPriority::LOW) {
+      SchedulerFeature::SCHEDULER->increaseOngoingLowPriority();
```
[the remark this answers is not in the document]
danielhlarkin: I pushed some changes here to make it a bit more transparent. Thoughts?

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -155,11 +156,13 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+      _ongoingLowPriorityLimit(static_cast<std::size_t>(ongoingMultiplier * _maxNumWorker)),
+      _wakeupQueueLength(5),
+      _wakeupTime_ns(1000),
+      _definitiveWakeupTime_ns(100000),
```
jsteemann, line 165: Wondering why these variables are reintroduced in this PR.
I deleted them previously, because they did nothing. :thinking:

arangod/Scheduler/SupervisedScheduler.h
```diff
@@ -138,18 +143,27 @@ class SupervisedScheduler final : public Scheduler {
...
+  //  1. the queue length is bigger than _wakeupQueueLength and the last submit time
+  //      is bigger than _wakeupTime_ns.
+  //  2. the last submit time is bigger than _definitiveWakeupTime_ns.
+  //
+  // The last submit time is a thread local variable that stores the time of the last
+  // queue operation.
+  alignas(64) std::atomic<uint64_t> _wakeupQueueLength;            // q1
+  std::atomic<uint64_t> _wakeupTime_ns, _definitiveWakeupTime_ns;  // t3, t4
```
jsteemann, line 165: Wondering why these variables are reintroduced in this PR.
I deleted them previously, because they did nothing. :thinking:

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -169,30 +172,53 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
           "arangodb_scheduler_threads_stopped", 0,
           "Number of scheduler threads stopped")),
       _metricsQueueFull(server.getFeature<arangodb::MetricsFeature>().counter(
-          "arangodb_scheduler_queue_full_failures", 0, "Tasks dropped and not added to internal queue")) {
+          "arangodb_scheduler_queue_full_failures", 0,
+          "Tasks dropped and not added to internal queue")),
+      _ongoingLowPriorityGauge(_server.getFeature<arangodb::MetricsFeature>().gauge(
+          "arangodb_scheduler_ongoing_low_prio", uint64_t(0),
+          "This is the total number of ongoing RestHandlers coming from "
+          "the low prio queue.")),
```
jsteemann, line 200:
```suggestion
          "Total number of ongoing RestHandlers coming from "
          "the low prio queue")),
```

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -169,30 +172,53 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+      _metricsQueueLengths{_server.getFeature<arangodb::MetricsFeature>().gauge("arangodb_scheduler_maintenance_prio_queue_length",
+                                                                                uint64_t(0),
+                                                                                "This is current queue length of the maintenance priority queue in "
+                                                                                "the scheduler."),
```
jsteemann, line 204:
```suggestion
                                                                                "Current queue length of the maintenance priority queue in "
                                                                                "the scheduler"),
```

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -169,30 +172,53 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+                           _server.getFeature<arangodb::MetricsFeature>().gauge("arangodb_scheduler_high_prio_queue_length",
+                                                                                uint64_t(0),
+                                                                                "This is current queue length of the high priority queue in "
+                                                                                "the scheduler."),
```
jsteemann, line 208:
```suggestion
                                                                                "Current queue length of the high priority queue in "
                                                                                "the scheduler"),
```

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -169,30 +172,53 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+                           _server.getFeature<arangodb::MetricsFeature>().gauge("arangodb_scheduler_medium_prio_queue_length",
+                                                                                uint64_t(0),
+                                                                                "This is current queue length of the medium priority queue in "
+                                                                                "the scheduler."),
```
jsteemann, line 212:
```suggestion
                                                                                "Current queue length of the medium priority queue in "
                                                                                "the scheduler"),
```

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -169,30 +172,53 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+                           _server.getFeature<arangodb::MetricsFeature>().gauge("arangodb_scheduler_low_prio_queue_length",
+                                                                                uint64_t(0),
+                                                                                "This is current queue length of the low priority queue in "
+                                                                                "the scheduler.")} {
```
jsteemann, line 216:
```suggestion
                                                                                "Current queue length of the low priority queue in "
                                                                                "the scheduler")} {
```

arangod/Network/NetworkFeature.cpp
danielhlarkin: If it is, I think we have bigger problems.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -610,39 +634,71 @@ bool SupervisedScheduler::sortoutLongRunningThreads() {
...
+  if (ongoing >= _ongoingLowPriorityLimit) {
+    return false;
+  }
+
+  // Because jobs may fan out to multiple servers and shards, we also limit
+  // dequeuing based on the number of internal requests in flight
+  TRI_ASSERT(_server.hasFeature<NetworkFeature>());
+  if (_server.getFeature<NetworkFeature>().isSaturated()) {
+    return false;
   }
```
jsteemann, line 686: This will perform a lookup in the features map every time we get here.
Can we store a reference to the NetworkFeature in an instance variable somehow, to avoid the repeated lookups?

arangod/Scheduler/SupervisedScheduler.cpp
danielhlarkin: Merge artifact I think. Removed again.

arangod/RestHandler/RestDocumentHandler.cpp
danielhlarkin: I think in the case that it does everything locally then it's still basically fine, right? It adds one more callback to the future chain, but since the sync version executes the async version and waits on the future, there's not a whole lot of difference. Unless it specifically makes a network call, then all the callbacks should still happen in the same thread. It's only put back in the scheduler if it hits the network, right?

jsteemann (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13033/

mpoeter (approved):

arangod/RestServer/Metrics.h
```diff
@@ -100,29 +100,43 @@ template<typename T> class Gauge : public Metric {
   ~Gauge() = default;
   std::ostream& print (std::ostream&) const;
   Gauge<T>& operator+=(T const& t) {
-    _g.store(_g + t);
+    T tmp;
+    do {
+      tmp = _g.load(std::memory_order_acquire);
+    } while (!_g.compare_exchange_weak(tmp, tmp + t, std::memory_order_acq_rel,
```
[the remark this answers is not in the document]
danielhlarkin: After some discussion, @mpoeter and I agreed it would be better to handle these changes in a separate PR (#13047) with the memory orders parameterized and defaulted to sequential consistency.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13065/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13084/

arangod/RestHandler/RestBatchHandler.h
```diff
@@ -67,7 +67,7 @@ class RestBatchHandler : public RestVocbaseBaseHandler {
   char const* name() const override final { return "RestBatchHandler"; }
   // be pessimistic about what this handler does... it may invoke V8
   // or not, but as we don't know where, we need to assume it
-  RequestLane lane() const override final { return RequestLane::CLIENT_FAST; }
+  RequestLane lane() const override final { return RequestLane::CLIENT_SLOW; }
```
mchacki, line 70: There seems to have been a bug with this Lane (https://github.com/arangodb/arangodb/pull/6196)
Which somehow made this Handle use a faster lane.
Have we made sure we do not reintroduce the issue caused there?

mchacki: (I personally think this handler should be Client_slow just want to make sure we not reintegrate regressions)

arangod/Scheduler/SchedulerFeature.cpp
```diff
@@ -60,9 +60,9 @@ namespace {
 size_t defaultNumberOfThreads() {
   // use two times the number of hardware threads as the default
   size_t result = arangodb::NumberOfCores::getValue() * 2;
-  // but only if higher than 64. otherwise use a default minimum value of 64
-  if (result < 64) {
-    result = 64;
+  // but only if higher than 64. otherwise use a default minimum value of 32
```
mchacki, line 63:
```suggestion
  // but only if higher than 32. otherwise use a default minimum value of 32
```

CHANGELOG
```diff
@@ -1,6 +1,28 @@
 devel
 -----
 
+* Throttle work coming from low priority queue, according to a constant
```
mchacki, line 4: i am missing information about the minimal / maximal number of threads modifications done to startup parameters. (e.g. the default minThreads is now 4 instead of 2.)

mchacki: also Default number of threads is lowered from 64 to 32

arangod/GeneralServer/CommTask.cpp
```diff
@@ -480,15 +480,23 @@ bool CommTask::handleRequestSync(std::shared_ptr<RestHandler> handler) {
...
   ContentType respType = handler->request()->contentTypeResponse();
   uint64_t mid = handler->messageId();
 
   // queue the operation in the scheduler, and make it eligible for direct execution
   // only if the current CommTask type allows it (HttpCommTask: yes, CommTask: no)
   // and there is currently only a single client handled by the IoContext
-  auto cb = [self = shared_from_this(), handler = std::move(handler)]() mutable {
+  auto cb = [self = shared_from_this(), handler = std::move(handler), prio]() mutable {
+    if (prio == RequestPriority::LOW) {
+      SchedulerFeature::SCHEDULER->ongoingLowPriorityTasks() += 1;
```
mchacki, line 493: How save are we here against crashes during shutdown?
The MetricsFeature manages the gauges memory.
The Scheduler is started before the MetricsFeature, so removed after.
=> The SCHEDULER may still be available, on shutdown, when this `cb` is invoked.
But the MetricsFeture could be shutdown already.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -215,36 +240,34 @@ bool SupervisedScheduler::queue(RequestLane lane, fu2::unique_function<void()> h
...
   // queue now has ownership for the WorkItem
   (void)work.release();  // intentionally ignore return value
 
-  if (approxQueueLength > _maxFifoSize / 2) {
+  if (approxQueueLength > _maxFifoSizes[3] / 2) {
     if ((::queueWarningTick++ & 0xFFu) == 0) {
       auto const& now = std::chrono::steady_clock::now();
       if (::conditionQueueFullSince == time_point{}) {
-        logQueueWarningEveryNowAndThen(::queueWarningTick, _maxFifoSize, approxQueueLength);
+        logQueueWarningEveryNowAndThen(::queueWarningTick, _maxFifoSizes[3], approxQueueLength);
```
mchacki, line 267: Do we want to add this warning now to prio 2 queue as well?
Having 4 lanes now there is a chance for lane 2 and 3 to starve, while lane 0 and 1 are busy. (0 should be fine, but 1 certainly has the right to do heavy load (e.g. replication => funny RocksDB stalls.)

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -627,22 +651,46 @@ bool SupervisedScheduler::canPullFromQueue(uint64_t queueIndex) const noexcept {
...
-    return (jobsDequeued - jobsDone) < (_maxNumWorkers * 3 / 4);
+    size_t limit = (_maxNumWorkers >= 8) ? (_maxNumWorkers * 3 / 4) : _maxNumWorkers - 2;
+
+    return threadsWorking < limit;
+  }
+
+  // We limit the number of ongoing low priority jobs to prevent a cluster
+  // from getting overwhelmed
+  std::size_t const ongoing = _ongoingLowPriorityGauge.load();
+  if (ongoing >= _ongoingLowPriorityLimit) {
```
mchacki, line 672: For now this is good.
For the future:
As soon as we have higher parallelization in AQL, this is not enough anymore. (I am think about multiple diamonds in AQL where the upper diamond can work parallel to the lower diamond. as soon as we have this the amount of fan-out requests in AQL is not bounded by the number of DBServers, but by number diamonds * DBServers..

arangod/Scheduler/SupervisedScheduler.h
```diff
@@ -53,6 +54,10 @@ class SupervisedScheduler final : public Scheduler {
   void toVelocyPack(velocypack::Builder&) const override;
   Scheduler::QueueStatistics queueStatistics() const override;
 
+  Gauge<uint64_t>& ongoingLowPriorityTasks();
```
mchacki, line 57: i have concerns about this API, we now hand out references to Memory, that is not even under control of this class, but owned by some other place (we hold a reference here ourselves)

I would prefer to have an API that allows increment / decrement operations and can encapsulate shutdown checks, as in this situation the reference we are holding here might be invalid already, depending on the cleanup state in Metrics feature. For now this is fine, as the MetricsFeature adds itself before the Greetings Phase. and the scheduler starts after the greetings phase.

mchacki (changes requested):
I have left some inline comments, which I would like to be considered.
For the "production" code comments, none of them justifies to block this PR, so I am fine with the comments not being addressed.

But I am missing CHANGELOG entries here, which would like to sort out before accepting.

arangod/GeneralServer/CommTask.cpp
mchacki: Never mind this the Metrics Feature is actually started before the Greating, the Scheduler is started after greating.

On a side note: I would prefer to drop either startsBefore or startsAfter from features, as following one direction as a maintainer is easy to do, following two is tricky...

mchacki (changes requested):
ok, cross out everything I said before.
This PR adds a total of 0 Tests on this very critical piece of code.
How is the plan to test this?

arangod/Network/NetworkFeature.cpp
```diff
@@ -127,6 +153,8 @@ void NetworkFeature::validateOptions(std::shared_ptr<options::ProgramOptions> op
   if (_idleTtlMilli < 10000) {
     _idleTtlMilli = 10000;
   }
+
+  _maxInFlight = std::clamp(_maxInFlight, ::MinAllowedInFlight, ::MaxAllowedInFlight);
```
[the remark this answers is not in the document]
danielhlarkin: Sometimes we do, sometimes we don't (for instance, see two lines above). I don't have strong feelings either way, but I think in case someone is setting it outside the allowed values here, they really have no idea what the value means. It should already be a hidden parameter, so they should only be setting it to something non-default in the case where we suggest it.

danielhlarkin:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13092/

arangod/Network/NetworkFeature.cpp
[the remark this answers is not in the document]
mchacki: Agreed, however on the long-run we may change the upper/lower bounds of this, and someone who set this parameter may not have seen this, so does not get it set as expected.
SO even for hidden parameters I am in favor of at least logging illegal input.

arangod/GeneralServer/CommTask.cpp
```diff
@@ -490,12 +490,12 @@ bool CommTask::handleRequestSync(std::shared_ptr<RestHandler> handler) {
   // and there is currently only a single client handled by the IoContext
   auto cb = [self = shared_from_this(), handler = std::move(handler), prio]() mutable {
     if (prio == RequestPriority::LOW) {
-      SchedulerFeature::SCHEDULER->ongoingLowPriorityTasks() += 1;
+      SchedulerFeature::SCHEDULER->trackBeginOngoingLowPriorityTask();
```
mchacki, line 493: \o/

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -862,8 +862,18 @@ void SupervisedScheduler::toVelocyPack(velocypack::Builder& b) const {
   b.add("direct-exec", VPackValue(0)); // obsolete
 }
 
-Gauge<uint64_t>& SupervisedScheduler::ongoingLowPriorityTasks() {
-  return _ongoingLowPriorityGauge;
+void SupervisedScheduler::trackBeginOngoingLowPriorityTask() {
+  if (_server.isStopping()) {
```
mchacki, line 866: yay this is save, I like it

mchacki (commented):
This modifications addressed my comments very good :+1:

The topic of missing tests is still open though still no Approve from my side.
Production Code is Approved.

arangod/RestHandler/RestBatchHandler.h
danielhlarkin: Yes, since it can take arbitrary requests we should keep it in the slow lane.

danielhlarkin: It's unclear to me why the changes in #6196 (using proper per-request lanes for the child requests) was reverted, but at least for the parent handler we should keep it in the slow lane.

danielhlarkin:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13100/

danielhlarkin:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13101/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13103/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13136/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13137/

PR #13134, opened 2020-12-02
avoid usage of std::regex for constructing date/time values for log messages

jsteemann:
### Scope & Purpose

Avoid usage of expensive `std::regex`-based functionality for constructing date/time values for log messages.
This is a performance optimization only.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/597

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12958/

### DIFF
--- a/lib/Logger/LogTimeFormat.cpp
+++ b/lib/Logger/LogTimeFormat.cpp
@@ -53,9 +53,12 @@ void appendNumber(uint64_t value, std::string& out, size_t size) {
   char buffer[22];
   size_t len = arangodb::basics::StringUtils::itoa(value, &buffer[0]);
   while (len < size) {
+    // zero-padding at the beginning of the output buffer, because
+    // we haven't yet written our number into it
     out.push_back('0');
     --size;
   }
+  // now, after zero-padding, write our number into the output
   out.append(&buffer[0], len);
 }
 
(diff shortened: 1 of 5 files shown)


lib/Logger/LogTimeFormat.cpp
```diff
@@ -104,49 +107,74 @@ TimeFormat formatFromName(std::string const& name) {
...
     out.push_back('.');
-    appendNumber(uint64_t(std::chrono::duration_cast<std::chrono::microseconds>(tp - tp2).count()), out, 6);
+    appendNumber(uint64_t(duration_cast<microseconds>(tp - tp2).count()), out, 6);
   } else {
     // all date-string variants handled here
-    if (format == TimeFormat::UTCDateString) {
+    if (format == TimeFormat::UTCDateString ||
+        format == TimeFormat::UTCDateStringMillis) {
       // UTC datestring
-      out.append(arangodb::basics::formatDate("%yyyy-%mm-%ddT%hh:%ii:%ssZ", tp_sys_clock_ms(std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch()))));
```
gnusi, line 146: do we still need this slow `formatDate` function?

jsteemann: Yes, it is still called from some AQL functions that can format date/time values according to a (dynamic) format string.

lib/Logger/LogTimeFormat.cpp
```diff
@@ -104,49 +107,74 @@ TimeFormat formatFromName(std::string const& name) {
...
-  std::chrono::system_clock::time_point tp = std::chrono::system_clock::now();
-
-  if (format == TimeFormat::Uptime) {
+void writeTime(std::string& out, 
+               TimeFormat format, 
+               std::chrono::system_clock::time_point tp,
+               std::chrono::system_clock::time_point startTp) {
+  using namespace date;
+  using namespace std::chrono;
+
```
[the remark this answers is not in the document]
jsteemann: We currently have 2 call sites (apart from the unit tests) that call this function:

* https://github.com/arangodb/arangodb/blob/bug-fix/avoid-std-regex-for-datestring-logging/lib/Logger/Logger.cpp#L352-L370
* https://github.com/arangodb/enterprise/blob/bug-fix/avoid-std-regex-for-datestring-logging/Enterprise/Audit/AuditFeature.cpp#L67-L70

Both reserve enough space in the output string before calling the function.
Do you think it makes sense to add an additional `reserve` call to `LogTimeFormats::writeTime(...)`?

lib/Logger/LogTimeFormat.cpp
gnusi: then `formatDate` sounds like a good candidate for future improvements with CTRE :-)

lib/Logger/LogTimeFormat.cpp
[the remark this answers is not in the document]
gnusi: no, I think it's enough

gnusi (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12960/

PR #13136, opened 2020-12-03
Less dropped followers

jsteemann:
### Scope & Purpose

Remove a case in which followers were dropped unnecessarily in streaming transactions that replicated to the same follower for multiple shards.

#### Backports:

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12973/


arangod/Transaction/Manager.cpp
```diff
@@ -472,10 +472,8 @@ Result Manager::createManagedTrx(TRI_vocbase_t& vocbase, TransactionId tid,
   // start the transaction
   transaction::Hints hints;
   hints.set(transaction::Hints::Hint::GLOBAL_MANAGED);
-  if (options.isFollowerTransaction || tid.isFollowerTransactionId()) {
-    hints.set(transaction::Hints::Hint::IS_FOLLOWER_TRX);
-  }
   if (isFollowerTransactionOnDBServer) {
+    hints.set(transaction::Hints::Hint::IS_FOLLOWER_TRX);
```
neunhoef, line 476: In principle a good change, I fear this might break the unit tests, since there are cases in there which do not set the server role to dbserver, as far as I remember. Let's see. If the unit tests are OK, then I am fine with this.

neunhoef (approved):
LGTM. Let's see if the unit tests like it. See comment.

arangod/Transaction/Manager.cpp
jsteemann: This change was intended to be just a code simplification.
The 2 identical `if` statements are merged here, which shouldn't change anything.
Or am I missing something?

neunhoef: I fear `isFollowerTransactionOnDBServer` checks if we are on a DBServer. In the unit tests (and Mocking) this is mostly `false`, so the if will no longer be triggered.
I might be wrong, though.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/12981/

arangod/Transaction/Manager.cpp
jsteemann: You were right, the one unit test failed. I fixed it by setting the server role to DB server for this particular test.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13000/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13026/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13035/

PR #13142, opened 2020-12-04
Bug fix/reduce allocations in hashed collect executor

mpoeter:
### Scope & Purpose

The HashedCollectExecutor uses an unordered_map to gather all groups with their aggregate values. Previously the value type of the map was a `std::unique_ptr<std::vector<std::unique_ptr<Aggregator>>>`, and every group was initialized with an actual value, even there were no aggregators (i.e., the vector was empty). So without `AGGREGATE`we had one allocation per group, with `AGGREGATE` we had `2 + <NumberOfAggregators>` allocations.

The new implementation uses a single allocation for all aggregators using placement new, which should also result in better cache locality. In addition, the unnecessary allocation in case no aggregators are used is also removed. This resulted in a runtime improvement of ~4-8% on my machine.

Tested using this script in the arangod console:
```
queries = {
  group:    "FOR i IN 1..500 FOR j IN 1..500 COLLECT a = i, b = j RETURN {a,b}",
  count:    "FOR i IN 1..500 FOR j IN 1..500 COLLECT a = i, b = j WITH COUNT INTO cnt RETURN {a,b,cnt}",
  sum:      "FOR i IN 1..500 FOR j IN 1..500 COLLECT AGGREGATE sum = SUM(i), sum2 = SUM(j) RETURN {sum,sum2}",
  groupSum: "FOR i IN 1..500 FOR j IN 1..500 COLLECT a = i, b = j AGGREGATE sum = SUM(i), sum2 = SUM(j) RETURN {sum,sum2}"
}

for (var k in queries) {
  console.log(k);
  let total = 0;
  let rounds = 10;
  for (var i = 0; i < rounds; ++i) {
    let start = new Date().getTime()
    db._query(queries[k], null, {silent: true}).toArray().length;
    let diff = new Date().getTime() - start;
    total += diff;
    console.log(diff);
  }
  console.log(total / rounds);
  console.log("#############");
}
```

#### Backports:

- [x] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13004/


jsteemann (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13032/

jsteemann:
Tests blue

PR #13158, opened 2020-12-07
Feature/const time view collect count

Dronplane:
The PR adds two optimizations:

1. ArangoSearch views are now participate in "collect-in-cluster" rule just like collections.
2. skipAll for IResearchViewExecutor/IResearchViewMergeExecutor could calculate documents count with improved performance
   O(1) time if there is no data filters. O(n) but with better constant time in case of data filters.
  2.1  Added new "countApproximate" option for ArangoSearch views. ('exact' by default). Setting it to `cost` will
         make  all skipAll O(1)  even with filters by sacrificing accuracy. 

#### Backports:

- [ ] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue https://github.com/arangodb/backlog/issues/772
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:


arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -63,6 +64,28 @@ namespace {
...
 constexpr irs::payload NoPayload;
 
+size_t calculateSkipAllCount(CountApproximate approximation,
+                             size_t currentPos,
+                             irs::doc_iterator* docs) {
+  TRI_ASSERT(docs);
+  size_t skipped{0};
+  switch (approximation) {
+    case CountApproximate::Cost:
+      skipped =  irs::cost::extract(*docs, 0);
```
gnusi, line 74: I suggest we fallback to `Exact` mode if `cost` attribute is missing

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1376,11 +1415,31 @@ size_t IResearchViewMergeExecutor<ordered, materializeType>::skipAll() {
...
+       skipped +=  calculateSkipAllCount(this->infos().countApproximate(),
+                                         segment.segmentPos, segment.docs.get());
+      }
+    }
+    // Adjusting by count of docs already consumed by heap but not consumed by executor.
+    // This count is heap size minus 1 already consumed by executor after heap.next.
+    // But we should adjust by the heap size only if the heap was advanced at least once
+    // (heap is actually filled on first next) or we have nothing consumed from doc iterators!
+    if (!_segments.empty() && this->infos().countApproximate() == CountApproximate::Exact &&
+        !filterConditionIsEmpty(&this->infos().filterCondition()) &&
```
gnusi, line 1437: I suggest we store the result of `filterConditionIsEmpty(&this->infos().filterCondition())` as a member in `infos`

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -139,8 +137,15 @@ std::map<std::string, arangodb::aql::ConditionOptimization> const conditionOptim
     {"nodnf", arangodb::aql::ConditionOptimization::NoDNF},
     {"noneg", arangodb::aql::ConditionOptimization::NoNegation},
     {"none", arangodb::aql::ConditionOptimization::None}};
+
+std::map<std::string, arangodb::iresearch::CountApproximate> const countApproximationTypeMap = {
```
gnusi, line 141: frozen_map?

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -251,6 +265,25 @@ bool fromVelocyPack(velocypack::Slice optionsSlice, IResearchViewNode::Options&
     }
   }
 
+  // countApproximate
+  {
+    auto const countApproximateSlice =
+        optionsSlice.get("countApproximate");
+    if (!countApproximateSlice.isNone() && !countApproximateSlice.isNull()) {
```
gnusi, line 272:
```suggestion
    if (!countApproximateSlice.isNone()) {
```

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -845,6 +896,9 @@ constexpr size_t getExecutorIndex(bool sorted, bool ordered) {
 namespace arangodb {
 namespace iresearch {
 
+bool filterConditionIsEmpty(aql::AstNode const* filterCondition) {
```
gnusi, line 899: `noexcept`

arangod/Aql/IResearchViewNode.h
```diff
@@ -51,6 +51,8 @@ struct VarInfo;
 
 namespace iresearch {
 
+bool filterConditionIsEmpty(aql::AstNode const* filterCondition);
```
gnusi, line 54: I suggest we have a  corresponding pre-evaluated member in infos

tests/IResearch/IResearchViewCountApproximate-test.cpp
```diff
@@ -0,0 +1,640 @@
...
+      EXPECT_TRUE(arangodb::iresearch::IResearchLinkHelper::find(*logicalCollection2, *_view)
+                  ->commit().ok());
+    }
+  }
+
+  void TearDown() override {
+    _view.reset();
+  }
+
+  void executeAndCheck(std::string const& queryString,
```
gnusi, line 238: Can we somehow check count approximation mode (e.g. via explain)?

gnusi (changes requested):

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -63,6 +64,35 @@ namespace {
...
 constexpr irs::payload NoPayload;
 
+size_t calculateSkipAllCount(CountApproximate approximation,
+                             size_t currentPos,
+                             irs::doc_iterator* docs) {
+  TRI_ASSERT(docs);
+  size_t skipped{0};
+  switch (approximation) {
+    case CountApproximate::Cost:
+      if (ADB_LIKELY(irs::get<irs::cost>(*docs))) {
```
gnusi, line 74:
```suggestion
        auto* cost = irs::get<irs::cost>(*docs);
        if (ADB_LIKELY(cost)) {
          skipped =  cost->estimate();
          skipped -= std::min(skipped, currentPos);
          break;
        } 
        TRI_ASSERT(false);
        [[fallthrough]];
    default:
      // check for unknown approximation
      // fallback to exact anyway
      TRI_ASSERT(CountApproximate::Exact == approximation);
      while (docs->next()) {
        skipped++;
      }
      break;
  }
  return skipped;
```

gnusi (approved):

Dronplane:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/13132/

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1010,26 +1051,24 @@ size_t IResearchViewExecutor<ordered, materializeType>::skip(size_t limit) {
 
   size_t const toSkip = limit;
 
-  for (size_t count = this->_reader->size(); _readerOffset < count;) {
+  for (size_t count = this->_reader->size(); _readerOffset < count; ++_readerOffset) {
     if (!_itr && !resetIterator()) {
```
jsteemann, line 1055: There is a difference here compared to the previous version:
If the _if_ condition `(!_itr && !resetIterator())` is satisfied, we _now_ increase `_readerOffset` and try again.
Previously, if the _if_ condition was satisfied, we did *not* increase `_readerOffset`.
The change may be intentional, I am just pointing this out in case it is an unintended change.

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1040,20 +1079,24 @@ size_t IResearchViewExecutor<ordered, materializeType>::skipAll() {
...
-
-    while (_itr->next()) {
-      skipped++;
+  if (_readerOffset < this->_reader->size()) {
+    if (this->infos().filterConditionIsEmpty()) {
+      skipped =  this->_reader->live_docs_count();
+      TRI_ASSERT(_totalPos <= skipped);
+      skipped -= std::min(skipped, _totalPos);
+      _readerOffset = this->_reader->size();
+    } else {
```
jsteemann, line 1088: Is it required to set `_currentSegmentPos` to either 0 or increase it by `this->_reader->size()` at the end of this `if` branch, in case there are any places reading `_currentSegmentPos` afterwards?

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1040,20 +1079,24 @@ size_t IResearchViewExecutor<ordered, materializeType>::skipAll() {
...
+      for (size_t count = this->_reader->size(); _readerOffset < count;
+           ++_readerOffset, _currentSegmentPos = 0) {
+        if (!_itr && !resetIterator()) {
+          continue;
+        }
+        skipped += calculateSkipAllCount(this->infos().countApproximate(), _currentSegmentPos, _itr.get());
+        _itr.reset();
```
jsteemann, line 1095: should `_currentSegmentPos` be set to 0 here?

arangod/Aql/IResearchViewExecutor.cpp
Dronplane: It is intentional. Initial version will loop forever if resetIterator fails and failing segment will never be skipped. That was a bug actually 😅

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1376,11 +1422,31 @@ size_t IResearchViewMergeExecutor<ordered, materializeType>::skipAll() {
...
+      if (this->infos().filterConditionIsEmpty()) {
+        TRI_ASSERT(segment.segmentIndex < this->_reader->size());
+        auto const live_docs_count =  (*this->_reader)[segment.segmentIndex].live_docs_count();
+        TRI_ASSERT(segment.segmentPos <= live_docs_count);
+        skipped += live_docs_count - segment.segmentPos;
+        segment.segmentPos = live_docs_count;
+      } else {
+       skipped +=  calculateSkipAllCount(this->infos().countApproximate(),
+                                         segment.segmentPos, segment.docs.get());
+      }
```
jsteemann, line 1437: should `segment.segmentPos` be set to the size of the segment here?

tests/js/server/aql/aql-arangosearch-count-optimization.js
```diff
@@ -0,0 +1,171 @@
...
+  return {
+    setUpAll : function () {
+      db._dropView("UnitTestView");
+      db._dropView("UnitTestViewSorted");
+      db._drop("UnitTestsCollection");
+      let c = db._create("UnitTestsCollection", { numberOfShards: 3 });
+
+      for (let i = 0; i < 10010; ++i) {
+        c.save({ value: "test" + (i % 10), count: i });
+      }
```
jsteemann, line 48: Turn 10k HTTP requests in (a big) one
```suggestion
      let docs = [];
      for (let i = 0; i < 10010; ++i) {
        docs.push({ value: "test" + (i % 10), count: i });
      }
      c.insert(docs);
```

jsteemann (approved):

arangod/Aql/IResearchViewExecutor.cpp
jsteemann: Ok, that was what I expected. :+1:

arangod/Aql/IResearchViewExecutor.cpp
Dronplane: After increasing _readerOffset to this->_reader->size()  executor is "exhausted" so _currentSegmentPos does not matter, but agreed, it is better to be set to 0 as we 'switched' to next segment.

arangod/Aql/IResearchViewExecutor.cpp
Dronplane: it is done at line 1091

arangod/Aql/IResearchViewExecutor.cpp
Dronplane: To be fully consistent - it should be increased by "skipped from segment" count. Will fix.

Dronplane:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/13209/
Restarted due to GIT failure job: https://jenkins.arangodb.biz/job/arangodb-matrix-pr-linux-nonmaintainercompile/12342/

Dronplane:
Tests blue except known fuerte timeout failure.

PR #13181, opened 2020-12-09
[3.7] Fix a potential nullptr access in AsyncAgencyComm

jsteemann:
### Scope & Purpose

Fix a potential nullptr access in AsyncAgencyComm in case there was a specific error when sending an agency request.

In case `network::sendRequest(...)` returns the error code `Error::Canceled`, the `response` member of the network::Response object will always be a nullptr. Dereferencing it unconditionally in AsyncAgencyComm::agencyAsyncSend(...) thus is not safe to do.

This PR fixes that, so that the `response` member will only be dereferenced when it is safe to do. This will be fixed properly in devel by disallowing direct access to the `response` member, and only making it available via a function.

No backports are required, as the AsyncAgencyComm functionality does not exist in 3.6.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13114/


fceller (approved):
LGTM

maierlars (approved):

danielhlarkin (approved):

jsteemann:
Tests blue

KVS85 (approved):
LGTM.

PR #13057, opened 2020-11-17
[3.6] fix gauge assignment operators

kvahed:
### Scope & Purpose

*(Please describe the changes in this PR for reviewers - **mandatory**)*

- [X] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [X] :book: CHANGELOG entry made

#### Backports:

- [X] No backports required
- [ ] Backports required for: *(Please specify versions and link PRs)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [X] This change is a trivial rework / code cleanup without any test coverage.
- [X] The behavior in this PR was *manually tested*
- [X] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:
https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/13264/

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

Please note that for legal reasons we require you to sign the [Contributor Agreement](https://www.arangodb.com/documents/cla.pdf)
before we can accept your pull requests.

PR #13184, opened 2020-12-09
Connection pool many connections fix

neunhoef:
This fixes a bug in the connectionPool implementation, which could
lead to the fact that basically no connections have been reused.

Under high load and high parallelism it could happen that a connection
was leased from fuerte but could then not be leased in the connection
pool. In this case the lease from fuerte was not returned and thus the
connection could no longer be used. In k8s and with TLS we have observed
situations where the timing was so that this built up and created
thousands of connections in a very short period of time.

This let performance suffer seriously.

The solution is essentially to release the lease immediately in the
above mentioned case.

We also added some more metrics to make this behaviour better
observable.

## Acceptance Checklist

*The below list is not exhaustive, think thoroughly whether the provided information is sufficient.*
*Remove options that do not apply*

### Scope & Purpose

*(Can you describe what functional change your PR is trying to effect?)*

- [*] Bug-Fix for a *released version* (did you remember to port this to all relevant release branches?)
- [ ] The behavior in this PR can be (and was) *manually tested* (support / qa / customers can test it)

### Testing & Verification

This change is a trivial rework / code cleanup without any test coverage.


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13126/

jsteemann (commented):
IMO this PR depends on https://github.com/arangodb/arangodb/pull/13057 being merged, so we should merge these two together.

arangod/Network/ConnectionPool.cpp
```diff
@@ -196,6 +214,10 @@ std::shared_ptr<fuerte::Connection> ConnectionPool::createConnection(fuerte::Con
 
 ConnectionPtr ConnectionPool::selectConnection(std::string const& endpoint,
                                                ConnectionPool::Bucket& bucket) {
+  using namespace std::chrono;
+
+  auto start = high_resolution_clock::now();
```
jsteemann, line 219: from https://en.cppreference.com/w/cpp/chrono/high_resolution_clock:

> The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided. It is often just an alias for std::chrono::steady_clock or std::chrono::system_clock, but which one it is depends on the library or configuration. When it is a system_clock, it is not monotonic (e.g., the time can go backwards). For example, for gcc's libstdc++ it is system_clock, for MSVC it is steady_clock, and for clang's libc++ it depends on configuration.

jsteemann (approved):

neunhoef:
This might need another fix for the unittests. Will check tomorrow. Do not merge for now.

neunhoef:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13202/

neunhoef:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13207/

KVS85:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13405/

KVS85:
[Set](https://github.com/arangodb/arangodb/pull/13184#event-4148954084) `DO_NOT_MERGE` label since https://github.com/arangodb/arangodb/pull/13057 was merged into this branch according to https://github.com/arangodb/arangodb/pull/13184#pullrequestreview-548547271 and had `DO_NOT_MERGE` label at the moment.

Review is required.

neunhoef (commented):
LGTM

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13422/

KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr/13434/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13446/ new run after clang compiler bug was found.

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13448/ after making test more stable.

KVS85:
Tests blue.

KVS85 (approved):
LGTM.

PR #13195, opened 2020-12-10
Add cluster support for checksum() function.

mpoeter:
### Scope & Purpose

Add cluster support for collection.checksum() method to calculate CRC checksums for collections.

- [x] :pizza: New feature

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] The behavior in this PR was _manually tested_
- [x] This change is already covered by existing tests - existing tests have been moved from `shell-collection-noncluster.js` to `shell-collection.js`.

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13146/


mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13151/

arangod/RestHandler/RestCollectionHandler.cpp
```diff
@@ -159,8 +156,13 @@ RestStatus RestCollectionHandler::handleCommandGet() {
...
 
-        obj->add("checksum", VPackValue(std::to_string(checksum)));
-        obj->add("revision", VPackValue(revId.toString()));
+        if (raw) {
+          obj->add("checksum", VPackValue(checksum));
+          obj->add("revision", VPackValue(revId.id()));
+        } else {
+          obj->add("checksum", VPackValue(std::to_string(checksum)));
+          obj->add("revision", VPackValue(revId.toString()));
+        }
```
jsteemann, line 165: I would be in favor of not making this distinction, but keep the APIs as simple as possible.
This would mean we will need to send the "checksum" and "revision" values back as strings, as before. This should be fine, because all clients are expecting it like that already.
This only makes parsing of the individual results from DB servers slightly more complicated, because now a string needs to be turned into a number before comparing/xoring.
But it should be doable easily, and efficiency should not matter, as we are talking only about two attributes in each response.

lib/V8/v8-globals.h
```diff
@@ -241,12 +241,6 @@ static inline v8::Local<v8::String> v8Utf8StringFactory(v8::Isolate* isolate,
     return;                              \
   } while (0)
 
-/// @brief "not yet implemented" handler for sharding
-#define TRI_THROW_SHARDING_COLLECTION_NOT_YET_IMPLEMENTED(collection) \
-  if (collection && ServerState::instance()->isCoordinator()) {       \
-    TRI_V8_THROW_EXCEPTION(TRI_ERROR_NOT_IMPLEMENTED);                \
-  }
```
jsteemann, line 248: :+1:

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13167/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13171/

jsteemann (approved):

mpoeter:
Retry: http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13205/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13210/

PR #13196, opened 2020-12-10
Fixed AR-113 and added regression tests

goedderz:
### Scope & Purpose

Previously, the `geo-index-optimizer` rule could move expressions into the index which used variables which aren't available yet at the index node.

#### Backports:

- [ ] No backports required

#### Related Information

- [X] GitHub issue / Jira ticket number: Fixes [AR-113](https://arangodb.atlassian.net/browse/AR-113).

### Testing & Verification

- [X] This change is already covered by existing tests, such as *aql-optimizer-geoindex*.
  - [X] Added new **integration tests**


goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13147/

tests/js/server/aql/aql-optimizer-geoindex.js
```diff
@@ -1029,7 +1029,64 @@ function optimizerRuleTestSuite() {
...
+    // Previously, the filter was moved into the index, but one access of `dist` in `dist < dist` was left, while the
+    // variable obviously can't be available yet at the index node.
+    // When failing, throws an error like
+    //   missing variable #1 (dist) for node #8 (IndexNode) while planning registers
+    testSelfReference1: function () {
+      const query = {
+        string: `
+          FOR doc IN @@col
+            LET dist = GEO_DISTANCE(doc.geo, [16, 53])
+            FILTER dist < dist
```
danielhlarkin, line 1044: I would think this would simply get optimized to `FILTER false` and isn't a good test.

tests/js/server/aql/aql-optimizer-geoindex.js
```diff
@@ -1029,7 +1029,64 @@ function optimizerRuleTestSuite() {
...
+    },
+
+    // See testSelfReference1
+    testSelfReference2: function () {
+      const query = {
+        string: `
+          FOR doc IN @@col
+            LET dist = GEO_DISTANCE(doc.geo, [16, 53])
+            LET dist2 = GEO_DISTANCE(doc.geo, [3, 7])
+            FILTER dist < dist2
```
danielhlarkin, line 1062: So my understanding here is that we can't use the index to do the calculation here because it's two-sided. Essentially, we don't have some sort of fixed query point to feed to the index. Right?

danielhlarkin: I guess what I'm asking is, is the only reason we got into trouble before these changes (i.e. the only time we get invalid vars) is when they are dependent on the document being fetched by the index? If there's another way to generate such an invalid var without it depending on the document being served by index (if it's used), then I think we need to write a test for it.

tests/js/server/aql/aql-optimizer-geoindex.js
goedderz: This is the query @Simran-B reported in [AR-113](https://arangodb.atlassian.net/browse/AR-113), and, without the bug fix, actually results in
> Query: AQL: missing variable #1 (dist) for node #8 (IndexNode) while planning registers (exception location: /work/ArangoDB/arangod/Aql/RegisterPlan.cpp:174). Please report this error to arangodb.com (while optimizing plan) (exception location: /work/ArangoDB/arangod/RestHandler/RestCursorHandler.cpp:305). Please report this error to arangodb.com
.

tests/js/server/aql/aql-optimizer-geoindex.js
goedderz: You're right, other variables that are defined before the filter but after the enumerate collection node are also affected. I'll write another test that does this.

tests/js/server/aql/aql-optimizer-geoindex.js
danielhlarkin: Fair enough. I'd consider this an excellent opportunity to add an optimization to simplify some expressions then! (In another PR of course.)

tests/js/server/aql/aql-optimizer-geoindex.js
danielhlarkin: Yeah, I mean I definitely understand how these existing cases would trip the same rule, just I can see where somehow we might detect them separately in the future and want to make sure we still catch the cases for variables that have nothing to do with the index access itself.

tests/js/server/aql/aql-optimizer-geoindex.js
jsteemann: FILTER conditions currently only get evaluated at query compile time if all their operands are constants, with values known at query compile time. This is not the case here, so the condition will not be optimized away.
However, we could rewrite such conditions in the future as you are suggesting. Good idea!

tests/js/server/aql/aql-optimizer-geoindex.js
goedderz: Writing such a test is sadly currently not possible, until [BTS-256](https://arangodb.atlassian.net/browse/BTS-256) is also fixed.

goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13195/

jsteemann:
http://172.16.10.101:8080/job/arangodb-matrix-pr/13212/

jsteemann (approved):

goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13216/

goedderz:
image

PR #13205, opened 2020-12-11
try to exclude assertion branches from coverage

jsteemann:
### Scope & Purpose

Try to exclude assertion branches from code coverage. 
The assertions have a true branch and a false branch, but only one of them is executed during our coverage tests. The not-taken branch would raise an assertion failure and crash the server, thus we don't expect it to be executed under normal conditions. But we also do not want to count this as a non-executed branch.

This needs dedicated testing with the code coverage Jenkins job.

#### Backports:

- [x] No backports required

### Testing & Verification

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13162/

### DIFF
--- a/lib/Basics/debugging.h
+++ b/lib/Basics/debugging.h
@@ -223,14 +223,15 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
 
 #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
 
-#define TRI_ASSERT(expr)                                                               \
+#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/                                           \
   if (!(ADB_LIKELY(expr))) {                                                           \
     arangodb::CrashHandler::assertionFailure(__FILE__, __LINE__, __FUNCTION__, #expr); \
   } else {}
 
 #else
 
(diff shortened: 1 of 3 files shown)


lib/Basics/debugging.h
```diff
@@ -223,14 +223,15 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
 
 #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
 
-#define TRI_ASSERT(expr)                                                               \
+#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/                                           \
```
mchacki, line 226: This seems to not have effect on my local Clang variant, 🤞 that Jenkins likes it more. Thats the only thing to count.
Even if it does not help, this comment does not cost us anything.

mchacki: FYI: i have also tested to place this EXCL_LINE on all lines of the macro, unfortunatly also no effect on my local variant.

mchacki (approved):
LGTM

lib/Basics/debugging.h
jsteemann: Ok, let's see if our Jenkins likes this or not.

PR #13245, opened 2020-12-17
Memoize hash for group values in HashedCollectExecutor.

mpoeter:
### Scope & Purpose

`HashedCollectExecutor::findOrEmplaceGroup` first calls `find` and if we don't have a matching value we call `try_emplace`. In that case we have to calculate the hash twice (once for each function). On top of that, a rehash of the unordered_map has to recalculate the hashes of all entries. This can be quite expensive, especially for non-scalar values like lists or objects.

This PR introduces the `HashedAqlValueGroup` struct which holds a hash value in addition to the `AqlValue`s. This allows us to calculate the hash only once and store it in the struct. Since we already use our own `Hash` type for the unordered-map, we can simply extend it to return the memoized hash value. In addition to that, we can also make use of the memoized hash value in the `KeyEqual` comparer to short-circuit if the hashes do not match. 

I have tested this with the following script:
```
queries = {
  group1: "FOR i IN 1..500 FOR j IN 1..300 COLLECT a = i, b = j SORT NULL RETURN {a,b}",
  group2: "FOR p IN 1..10000 FOR i IN 1..10 FOR j IN 1..10 COLLECT a = i, b = j SORT NULL RETURN {a,b}",
  group3: "FOR i IN 1..10 FOR j IN 1..10 FOR p IN 1..10000 COLLECT a = i, b = j SORT NULL RETURN {a,b}",
  group4: "FOR i IN 1..100 FOR j IN 1..100 LET x = 1..i*10 LET y = 1..j*10 COLLECT a = x, b = y SORT NULL RETURN {a,b}",
}

for (var k in queries) {
  console.log(k);
  let total = 0;
  let rounds = 10;
  for (var i = 0; i < rounds; ++i) {
    let start = new Date().getTime();
    db._query(queries[k], null, {silent: true}).toArray().length;
    let diff = new Date().getTime() - start;
    total += diff;
    console.log(diff);
  }
  console.log(total / rounds);
}
```
which resulted in the following speedups on my machine:
```
  group1: 7,98%
  group2: 2,12%
  group3: 3,34%
  group4: 22,54%
```

Note: there are a few assertions to ensure that the hash value of the struct is set correctly, so in maintainer mode the performance can actually be worse than before.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13276/

### DIFF
--- a/arangod/Aql/AqlValueGroup.h
+++ b/arangod/Aql/AqlValueGroup.h
@@ -34,14 +34,24 @@ struct Options;
 
 namespace aql {
 
+using AqlValueGroup = std::vector<AqlValue>;
+
+struct HashedAqlValueGroup {
+  size_t hash;
+  AqlValueGroup values;
+};
+
 /// @brief hasher for a vector of AQL values
 struct AqlValueGroupHash {
   explicit AqlValueGroupHash(size_t num);
 
   size_t operator()(std::vector<AqlValue> const& value) const;
   size_t operator()(AqlValue const& value) const;
+  size_t operator()(HashedAqlValueGroup const& value) const noexcept;
 
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
   size_t const _num;
+#endif
 };
 
 /// @brief comparator for a vector of AQL values
@@ -50,6 +60,7 @@ struct AqlValueGroupEqual {
 
   bool operator()(std::vector<AqlValue> const& lhs, std::vector<AqlValue> const& rhs) const;
   bool operator()(AqlValue const& lhs, AqlValue const& rhs) const;
+  bool operator()(HashedAqlValueGroup const& lhs, HashedAqlValueGroup const& rhs) const;
 
   velocypack::Options const* _vpackOptions;
 };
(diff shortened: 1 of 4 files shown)


mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13285/

arangod/Aql/AqlValueGroup.h
```diff
@@ -34,14 +34,24 @@ struct Options;
 
 namespace aql {
 
+using AqlValueGroup = std::vector<AqlValue>;
+
+struct HashedAqlValueGroup {
+  size_t hash;
```
jsteemann, line 40: Shall we have a default initializer for `hash`, e.g. 0?
I know `hash` will be overriden later, and before it is read from, but adding a default value here may be more intuitive (at least for me it is).

jsteemann (approved):

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13290/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13295/

jsteemann:
Tests blue

PR #13247, opened 2020-12-17
Refactor FixedSizeAllocator to perform less allocations.

mpoeter:
### Scope & Purpose

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is a trivial rework / code cleanup without any test coverage.ls

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13279/


lib/Basics/FixedSizeAllocator.h
```diff
@@ -89,8 +97,8 @@ class FixedSizeAllocator {
    private:
     size_t _numAllocated;
     size_t _numUsed;
-    char* _alloc;
     char* _data;
+    std::unique_ptr<MemoryBlock> _next;
```
jsteemann, line 101: While this works, I think it could create very deep stacks on destruction of the memory block.
If, for example, there is a chain of 100 memory blocks, the dtor of each block will invoke the dtor of the next block etc., so this will cause at least 100 stack frames to be produced.
IMHO this is not ideal, as it prevents scaling the number of MemoryBlocks.

lib/Basics/FixedSizeAllocator.h
```diff
@@ -36,22 +36,24 @@ namespace arangodb {
...
       TRI_ASSERT(numItems >= 32);
-      // assumption is that the size of a cache line is at least 64,
-      // so we are allocating 64 bytes in addition
-      _alloc = new char[(std::max<size_t>(sizeof(T), alignof(T)) * numItems) + 64];
-
-      _numAllocated = numItems;
 
       // adjust memory address to cache line offset (assumed to be 64 bytes)
-      _data = reinterpret_cast<char*>((reinterpret_cast<uintptr_t>(_alloc) + 63u) &
+      _data = reinterpret_cast<char*>((reinterpret_cast<uintptr_t>(this + 1) + 63u) &
```
jsteemann, line 56: Now we do have the cache-line math here, but also take into account the cache line overhead when we create the MemoryBlock with `operator new`. It would be ideal to deal with this just once.

lib/Basics/FixedSizeAllocator.h
```diff
@@ -134,11 +144,21 @@ class FixedSizeAllocator {
...
     // maximum block size is for 4096 items
-    size_t const numItems = 64 << std::min<size_t>(6, _blocks.size() + 1);
-    _blocks.emplace_back(std::make_unique<MemoryBlock>(numItems));
+    size_t const numItems = 64 << std::min<size_t>(6, _numBlocks + 1);
+
+    // assumption is that the size of a cache line is at least 64,
+    // so we are allocating 64 bytes in addition
+    auto const dataSize = sizeof(T) * numItems + 64;
+    void* p = ::operator new(sizeof(MemoryBlock) + dataSize);
+    new (p)MemoryBlock(numItems);
```
jsteemann, line 153: If the constructor throws an exception here, we will leak the memory in `p`.
Currently the constructor does not throw, so it is safe.
But it would be good to either mark the constructor of MemoryBlock as `noexcept`, or at least leave a code comment here about it.

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13314/

lib/Basics/FixedSizeAllocator.h
```diff
@@ -125,14 +119,20 @@ class FixedSizeAllocator {
...
-  void clear() {
-    _head.reset();
+  void clear() noexcept {
+    auto* block = _head;
+    while (block != nullptr) {
+      auto* next = block->getNextBlock();
+      block->~MemoryBlock();
+      ::operator delete(block);
+      block = next;
+    }
```
jsteemann, line 129: I would prefer if this method would also set `_head` to a nullptr after having performed the cleanup. That will leave things in a consistent state, even though right now `clear()` is only called from the destructor.

jsteemann (approved):

lib/Basics/FixedSizeAllocator.h
mpoeter: Good point!

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13318/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13323/

jsteemann:
http://172.16.10.101:8080/job/arangodb-matrix-pr-windows/6826/

jsteemann:
http://172.16.10.101:8080/job/arangodb-matrix-pr-windows/6829/

PR #13251, opened 2020-12-17
Metrics documentation snippets in code

neunhoef:
### Scope & Purpose

This is an initiative to create an infrastructure to keep the
documentation for the manual for our metrics with the metrics code
itself, right in the ArangoDB repository.

The plan is that any ArangoDB version can actually create its list of
metrics with documentation in an API. This can then in turn be used to
create parts of the documentation web pages.

We essentially add one string to each metric, which can contain markdown
manual documentation.

So far this is experimental.

#### Backports:

- [*] Backports probably wanted for: 3.6, 3.7

### Testing & Verification

Does not need testing, nearly no actual code is touched.


neunhoef:
Obsoleted by https://github.com/arangodb/arangodb/pull/13668/.

PR #13268, opened 2020-12-21
do not kill thread in case of exception

fceller:
Port of 3.5 change: do not delete thread in case of exception, catch and report the exception instead.

### DIFF
--- a/arangod/GeneralServer/IoContext.cpp
+++ b/arangod/GeneralServer/IoContext.cpp
@@ -23,6 +23,9 @@
 #include "IoContext.h"
 
 #include "Basics/cpu-relax.h"
+#include "Logger/LogMacros.h"
+#include "Logger/Logger.h"
+#include "Logger/LoggerStream.h"
 
 using namespace arangodb;
 using namespace arangodb::basics;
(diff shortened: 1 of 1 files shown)


arangod/GeneralServer/IoContext.cpp
```diff
@@ -38,7 +41,12 @@ IoContext::IoThread::~IoThread() { shutdown(); }
 
 void IoContext::IoThread::run() {
   // run the asio io context
-  _iocontext.io_context.run();
+  try {
+    _iocontext.io_context.run();
+  } catch (std::exception const& ex) {
+    LOG_TOPIC("6794f", WARN, Logger::THREADS)
+      << "caught exception in IO thread '" << ex.what();
```
jsteemann, line 48:
```suggestion
      << "caught exception in IO thread: " << ex.what();
```

jsteemann (approved):

arangod/GeneralServer/IoContext.cpp
fceller: Resolved

KVS85:
http://172.16.10.101:8080/job/arangodb-matrix-pr/13416/

KVS85:
Tests blue.

KVS85 (approved):
LGTM.

PR #13271, opened 2020-12-21
do not kill thread in case of exception

fceller:
Port of 3.5 change: do not delete thread in case of exception, catch and report the exception instead.

### DIFF
--- a/arangod/GeneralServer/IoContext.cpp
+++ b/arangod/GeneralServer/IoContext.cpp
@@ -25,6 +25,10 @@
 
 #include <thread>
 
+#include "Logger/LogMacros.h"
+#include "Logger/Logger.h"
+#include "Logger/LoggerStream.h"
+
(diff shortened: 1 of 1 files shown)


arangod/GeneralServer/IoContext.cpp
```diff
@@ -25,6 +25,11 @@
 
 #include <thread>
 
+#include "Basics/cpu-relax.h"
```
jsteemann, line 28: I don't think we need to include "cpu-relax.h" here.
While waiting, the code here calls `std::this_thread::yield()` and nothing from cpu-relax.h.
Or am I missing something?

jsteemann (approved):

arangod/GeneralServer/IoContext.cpp
fceller: Done

PR #13280, opened 2020-12-22
Prevent write during RocksDB recovery

jsteemann:
### Scope & Purpose

Prevent a write to RocksDB during recovery in the case that the database already exists. If the database didn't already exist, it's okay to do the write. If it exists, we shouldn't write anything until after recovery has finished. Other than simply wanting to avoid writes during recovery for safety, we have also seen at least one case where a server shut down while in a write-stopped state, and on startup re-entered that write-stopped state; the write at startup then could not proceed, and the server was not able to proceed with recovery, start up, and serve health check requests.

#### Backports:

- [x] Backports required for: 3.6

### Testing & Verification

- [x] This change is already covered by existing tests, such as *recovery, upgrade_data*.

Link to Jenkins PR run: 
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13346/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 v3.7.6 (XXXX-XX-XX)
 -------------------
 
+* Prevent a write to RocksDB during recovery in the case that the 
+  database already exists. The write at startup is potentially blocking,
+  and will delay the startup for servers that were stopped shut down while
+  in a write-stopped state.
+
 * Added "startupTime", "computationTime" and "storageTime" to Pregel result
   statistics.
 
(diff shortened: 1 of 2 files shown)


jsteemann:
Tests blue except unrelated Pregel test failure.

CHANGELOG
```diff
@@ -1,6 +1,11 @@
 v3.7.6 (XXXX-XX-XX)
 -------------------
 
+* Prevent a write to RocksDB during recovery in the case that the 
+  database already exists. The write at startup is potentially blocking,
+  and will delay the startup for servers that were stopped shut down while
```
danielhlarkin, line 6:
```suggestion
  and will delay the startup for servers that were shut down while
```

danielhlarkin (approved):

PR #13288, opened 2020-12-22
make storage engine report its health to the agency

jsteemann:
### Scope & Purpose

Make DB servers report storage engine errors to the agency, via a new "healthy" attribute in sent to requests to Sync/ServerStates/<id>. The supervision will check this attribute if it is posted, and if it contains a value of `false`, will mark the server as BAD.
DB server health is currently determined by whether or not the storage engine (RocksDB) has reported a background error, and by whether or not the free disk space has reached a critical low amount. The current threshold for free disk space is set at 2% of the disk capacity (the disk is considered that contains the RocksDB database directory).

Failed servers are now reported consistently in the web interface, at approximately the same time in the navigation bar and in the nodes view. Previously these two places had their own, independent poll mechanism for the nodes' health, and they were updated independently, which could cause an inconsistent view of the nodes' availability. Using only one poll mechanism instead also saves some period background requests for the second availability check.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13358/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13362/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13366/

arangod/Agency/AgencyComm.cpp
```diff
@@ -643,6 +643,7 @@ AgencyCommResult AgencyComm::sendServerState(double timeout) {
     builder.add("status", VPackValue(status));
     std::string const stamp = AgencyCommHelper::generateStamp();
     builder.add("time", VPackValue(stamp));
+    builder.add("healthy", VPackValue(isHealthy));
```
neunhoef, line 646: How about adding a string in addition to describing the actual problem?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2547,4 +2580,71 @@ void RocksDBEngine::releaseTick(TRI_voc_tick_t tick) {
...
+  auto now = std::chrono::steady_clock::now();
+
+  // the following checks are executed under a mutex so that different
+  // threads can potentially call in here without messing up any data.
+  // in addition, serializing access to this function avoids stampedes
+  // with multiple threads trying to calculate the free disk space 
+  // capacity at the same time, which could be expensive.
+  MUTEX_LOCKER(guard, _healthCheckMutex);
+
+  bool lastCheckLongAgo = (now - _lastHealthCheckTimestamp > std::chrono::seconds(30));
```
neunhoef, line 2593: I doubt JPMC will like this. They create a ticket for their DB admin team for every ERROR log line we write. So if something happens, they wake up in the morning with lots of tickets. Maybe we have to build an exponential backoff here.

neunhoef: At least the log message is always the same ID.

arangod/Agency/Supervision.cpp
```diff
@@ -669,8 +688,15 @@ std::vector<check_t> Supervision::check(std::string const& type) {
...
-        if (elapsed.count() <= _okThreshold) {
+        
+        // if there is an agreement on how to handle unhealthy servers,
+        // the "healthy" flag could be used here in the future to downgrade
+        // the health of servers that report back to us, but with an
+        // unhealthy status
+        // TODO: figure this out and add to design document - until then
+        // isHealthy remains unused here!
+
+        if (elapsed.count() <= _okThreshold /*&& isHealthy*/) {
```
neunhoef, line 699: Even if we do not declare the server FAILED, we could still report any change in the `isHealthy` flag to the persisted Raft log under `/arango/Supervision/Health` and thus export it to the normal health check API. We would have to adjust `diffStat` in this file and write changes to Raft.

neunhoef (changes requested):
This is a good start. I would like to see a reason for unhealthiness sent to the agency. Yes, for now we only have the disk space and RocksDB, but already this would be helpful to distinguish. However, so far the unhealthiness report only ends up in the transient store. We should at least forward it on change to the persisted Raft log under `/arango/Supervision/Health` such that it gains visibility in the Health API.
Furthermore, we must not log an error even every 30 seconds, since JPMC will generate a ticket for their DB admin team for *every* single ERROR line in our logs.

arangod/RocksDBEngine/RocksDBEngine.cpp
neunhoef: Let's increase the 30 seconds to 24h.

neunhoef (approved):
LGTM for now, we will continue this work in https://arangodb.atlassian.net/browse/AR-117

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13390/

jsteemann:
DB servers now send a "health" attribute to the agency with more data.
There is also code in the supervision to read such data from "health" if it is posted, but currently nothing happens with the data that is read. In the future, the supervision could pick up this data and use it for its decisions to mark as server as BAD or FAILED.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13392/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13463/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13466/

PR #13289, opened 2020-12-22
Feature/collect with count transformation

mpoeter:
### Scope & Purpose

`COLLECT WITH COUNT INTO x` and `COLLECT var=expr WITH COUNT INTO x` are now internally transformed into `COLLECT AGGREGATE x = LENGTH(null)` and `COLLECT var=expr AGGREGATE x = LENGTH(null)` respectively. This not only simplified the code, but also allows more query optimizations:
    - If the variable in `COLLECT WITH COUNT INTO var` is not used, the implicit
      aggregator is now removed.
    - All queries of the form `COLLECT AGGREGATE x = LENGTH(_)` are now executed
      using the count executor, which can result in significantly improved
      performance.

- [x] :pizza: New feature

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13360/


arangod/Aql/CollectNode.cpp
```diff
@@ -124,7 +123,6 @@ void CollectNode::toVelocyPackHelper(VPackBuilder& nodes, unsigned flags,
     }
   }
 
-  nodes.add("count", VPackValue(_count));
```
jsteemann, line 127: Not sure yet if we should keep reporting this attribute for compatibility reasons, and simply always return "false" here until no clients (e.g. `db._explain(...)`) still look for it.
I suggest we keep it, but hard-code the return value to "false" here and leave a code comment about removing the attribute from the response in a future version.

arangod/Aql/Ast.cpp
```diff
@@ -598,24 +598,17 @@ AstNode* Ast::createNodeCollect(AstNode const* groups, AstNode const* aggregates
...
 }
 
-/// @brief create an AST collect node, COUNT INTO
+/// @brief create an AST collect node with a single COUNT aggregator
 AstNode* Ast::createNodeCollectCount(AstNode const* list, char const* name,
                                      size_t nameLength, AstNode const* options) {
-  AstNode* node = createNode(NODE_TYPE_COLLECT_COUNT);
-  node->reserve(2);
+  auto args = createNodeArray(1);
+  args->addMember(createNodeValueNull());
```
jsteemann, line 605: Is it necessary for the COUNT/LENGTH to have a `null` input variable? Or will it even work without?

arangod/Aql/OptimizerRules.cpp
```diff
@@ -4505,7 +4506,8 @@ void arangodb::aql::collectInClusterRule(Optimizer* opt, std::unique_ptr<Executi
 
             replaceGatherNodeVariables(plan.get(), gatherNode, replacements);
           } else if (  //! collectNode->groupVariables().empty() &&
-              (!collectNode->hasOutVariable() || collectNode->count())) {
+              (!collectNode->hasOutVariable() || false/*collectNode->count()*/)) {
+            // TODO
```
jsteemann, line 4510: This looks unready yet.
```suggestion
              (!collectNode->hasOutVariable()) {
```

tests/js/server/aql/aql-optimizer-collect-count.js
```diff
@@ -487,17 +515,18 @@ function optimizerCountTestSuite () {
       // expectation is that we exactly get one result
       // count will be 0, start will be undefined
       assertEqual(1, results.length);
-      assertEqual(0, results[0].count);
+      assertEqual(null, results[0].sum);
```
jsteemann, line 518:
```suggestion
      assertNull(results[0].sum);
```

tests/js/server/aql/aql-optimizer-rule-collect-in-cluster.js
```diff
@@ -84,7 +84,7 @@ function optimizerCollectInClusterSuite () {
       });
 
       assertNotEqual(-1, plan.rules.indexOf("collect-in-cluster"));
-      assertEqual(["SingletonNode", "EnumerateCollectionNode", "CollectNode", "RemoteNode", "GatherNode", "CollectNode", "ReturnNode"], nodeTypes);
+      assertEqual(["SingletonNode", "CalculationNode", "EnumerateCollectionNode", "CollectNode", "RemoteNode", "GatherNode", "CollectNode", "ReturnNode"], nodeTypes);
```
jsteemann, line 87: It looks like the new plan is more expensive than the previous one, as we have an additional CalculationNode now. Can we get rid of this to remove its overhead?

arangod/Aql/Ast.cpp
mpoeter: At the moment all aggregator functions require exactly one argument, even though COUNT/LENGTH simply ignore it. I am planning to improve that, but at the moment we have to pass _something_.

tests/js/server/aql/aql-optimizer-rule-collect-in-cluster.js
mpoeter: Unfortunately this is due to the fact that we are passing null to the COUNT aggregator. This CalculationNode is initializing a temporary variable with null in order to pass it to LENGHT:
```
Execution plan:
 Id   NodeType          Est.   Comment
  1   SingletonNode        1   * ROOT
  3   CalculationNode      1     - LET #2 = null   /* json expression */   /* const assignment */
  6   IndexNode          100     - FOR i IN foo   /* primary index scan, scan only */
  4   CollectNode          1       - COLLECT AGGREGATE cnt = LENGTH(#2)   /* count */
  5   ReturnNode           1       - RETURN cnt
  ```

File: arangod/Aql/Ast.cpp (hunk 2 above)
Comment (jsteemann) [reply arangod/Aql/Ast.cpp:605]: The `Aggregator` class has a static method `requiresInput` that we could leverage here or in other places that fail if no input argument is used:
```
102   /// @brief whether or not the aggregator requires any input or if the input
103   /// can be optimized away (note current: COUNT/LENGTH don't, all others do)
104   static bool requiresInput(std::string const& type);
```
This will return false for COUNT/LENGTH, and true for all other aggregate functions.

**Comment by mpoeter** [comment]
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13492/

**Comment by mpoeter** [comment]
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13497/

**Comment by mpoeter** [comment]
Next try http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13500/

**Comment by mpoeter** [comment]
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13505/

**Comment by mpoeter** [comment]
Tests blue

File: CHANGELOG [hunk 6]
```diff
@@ -1,5 +1,14 @@
 devel
 -----
+* `COLLECT WITH COUNT INTO x` and `COLLECT var=expr WITH COUNT INTO x` are now
+  internally transformed into `COLLECT AGGREGATE x = LENGTH(null)` and
+  `COLLECT var=expr AGGREGATE x = LENGTH(null)` respectively. This not only
+  simplified the code, but also allows more query optimizations:
```
```suggestion

* `COLLECT WITH COUNT INTO x` and `COLLECT var=expr WITH COUNT INTO x` are now
  internally transformed into `COLLECT AGGREGATE x = LENGTH(null)` and
  `COLLECT var=expr AGGREGATE x = LENGTH(null)` respectively. This not only
  simplified the code, but also allows more query optimizations:
```

PR #13291, opened 2020-12-22
Make internal ClusterInfo::getPlan() wait for initial plan load from agency

jsteemann:
### Scope & Purpose

Otherwise the maintenance, fired off by the heartbeat thread, may not wait for the initial data to arrive from the agency, which can cause Plan and Current to be empty when maintenance starts.
No backport for 3.6 is needed, as 3.6 has blocking agency communication here in case the Plan is not there.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *mini chaos*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13365/

### DIFF
--- a/arangod/Cluster/ClusterInfo.cpp
+++ b/arangod/Cluster/ClusterInfo.cpp
@@ -200,15 +200,6 @@ class PlanCollectionReader {
 } // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
-/// @brief a local helper to report errors and messages
-////////////////////////////////////////////////////////////////////////////////
-
-static inline int setErrormsg(int ourerrno, std::string& errorMsg) {
-  errorMsg = TRI_errno_string(ourerrno);
-  return ourerrno;
-}
-
-////////////////////////////////////////////////////////////////////////////////
 /// @brief check whether the JSON returns an error
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -257,11 +248,13 @@ CollectionInfoCurrent::~CollectionInfoCurrent() = default;
 ////////////////////////////////////////////////////////////////////////////////
 
 ClusterInfo::ClusterInfo(application_features::ApplicationServer& server,
-                         AgencyCallbackRegistry* agencyCallbackRegistry)
+                         AgencyCallbackRegistry* agencyCallbackRegistry,
+                         int syncerShutdownCode)
   : _server(server),
     _agency(server),
     _agencyCallbackRegistry(agencyCallbackRegistry),
     _rebootTracker(SchedulerFeature::SCHEDULER),
+    _syncerShutdownCode(syncerShutdownCode),
     _planVersion(0),
     _planIndex(0),
     _currentVersion(0),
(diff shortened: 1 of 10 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13369/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13373/

KVS85:
Tests blue.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -5021,7 +5021,7 @@ ClusterInfo::getPlan(uint64_t& index, std::unordered_set<std::string> const& dir
 
   // We should never proceed here, until we have seen an
   // initial agency cache through loadPlan
-  waitForPlan(1);
```
neunhoef, line 5024: We also need to address `waitForCurrent` calls, I found another one with forgotten `.wait()` or `.get()`, it is in line 5054.

neunhoef (changes requested):

neunhoef (commented):
Just for the record, here is how it is supposed to work at server start:
  1. AgencyCache and its thread starts polling from the agency. Since it does not have a State yet, it is guaranteed to get the latest snapshot and all that follows it.
  2. The PlanSyncer and CurrentSyncer in `ClusterInfo` request a local callback from AgencyCache whenever /arango/Plan/Version or /arango/Current/Version respectively change, this triggeres `loadPlan` and `loadCurrent` eventually.
  3. `loadPlan` and `loadCurrent` are the only ones which change `_planIndex`, `_planVersion`, `_currentVersion`, `_currentIndex` in `ClusterInfo`.
  4. `getPlan` and `getCurrent` are protected with a `waitForPlan(1).get()` and `waitForCurrent(1).get()` respectively. So nobody outside can get a plan or a current from `ClusterInfo` before it is set up correctly.
  5. Additionally, the `ClusterFeature::start` method has a `waitFor(1).get()` call such that this does not proceed unless the AgencyCache has got a first delivery from the agency, which will already be reasonably up to date. This is all before it even starts the HeartbeatThread (running Maintenance) and the Syncers.
  6. The SyncerThreads do a `loadPlan` or `loadCurrent` respectively before they begin their main loop.

neunhoef (approved):
LGTM. Further to my previous comment: It is safe to throw in `getPlan` and `getCurrent` when we are shutting down. This is caught in the `run` method of the `HeartbeatBackgrounThread` (which does maintenance via `DBServerAgencySync`) and the shutdown will then work properly. Furthermore, we will not do a maintenance run in this case, even if this is the very first run.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13381/

KVS85:
`IResearch*` unit tests are broken in the last run.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13394/

KVS85:
Tests blue.

KVS85 (approved):
LGTM.

fceller (approved):
LGTM

PR #13310, opened 2020-12-30
Stabilize Foxx cleanup test

neunhoef:
I have observed some instability in this test:

  tests/js/server/shell/shell-foxx-cleanup-spec.js

Namely, it relies on a directory which is only created after startup
by the foxx manager selfheal procedure. This is a race.

This PR adds a wait loop to stabilize the test.

This has been fixed for 3.6 in #13184 and in 3.7 in #13309

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Stabilize a Foxx cleanup test.
+
 * Add some more metrics to the ConnectionPool.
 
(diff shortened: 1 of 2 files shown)


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13450/

mchacki (approved):
LGTM

PR #12646, opened 2020-09-12
AQL: DATE_UTCTOLOCAL and  DATE_LOCALTOUTC

coronabytes:
### Scope & Purpose

Timezone conversion functionality for ArangoDB. Updates date library and bundles IANA timezone database.
Provides two new AQL functions DATE_UTCTOLOCAL and  DATE_LOCALTOUTC which can convert utc dates to local timezone and vice versa. tzdata is located in 3rdParty and needs to be copied to ICU_DESTINATION_DIRECTORY.

- [ ] :hankey: Bugfix 
- [x] :pizza: New feature 
- [ ] :hammer: Refactoring 
- [ ] :book: CHANGELOG entry made
- [ ] :muscle: The behavior in this PR was *manually tested*
- [ ] :computer: The behavior change can be verified via automatic tests

#### Backports:

- [x] No backports required
- [ ] Backports required for: *(Please specify versions)*

#### Related Information

*(Please reference tickets / specification etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket number: #12566 
- [ ] Design document: 

### Testing & Verification

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [x] This PR adds tests that were used to verify all changes:
  - [ ] Added **Regression Tests**
  - [ ] Added new C++ **Unit Tests**
  - [x] Added new **integration tests** (i.e. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)

Additionally:

- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

### Documentation

- [ ] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)
- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [x] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### DATE_UTCTOLOCAL()

Converts *date* assumed in Zulu time (UTC) to local *timezone*.

`DATE_UTCTOLOCAL(date, timezone) → dateString`

- **date** (number\|string): numeric timestamp or ISO 8601 date time string
[...]

PR #13314, opened 2020-12-31
AQL: DATE_TIMEZONE(S)

coronabytes:
### Scope & Purpose

New features for #12646 
Need to generate dropdowns for all timezones - current server timezone for sensible default.

* Added DATE_TIMEZONE - returns server timezone
* Added DATA_TIMEZONES - returns all timezones
* updated tzdata to 2020f
* removed obsolete tzdata files
* updated windowsZones.xml

TBD:
- Naming of the functions:
  - Although not accepting date values I prefixed them with DATE_ prefix to categorize the together with the other timezone features
  - Alternative suggestions: TIMEZONE_CURRENT(), TZ_CURRENT(), TIMEZONE_LIST(), TZ_LIST()
- Better tests
  - DATE_TIMEZONE() - can't know server timezone?
  - DATE_TIMETONES() - atleast contain Berlin / NewYork / Shanghai?

#### Backports:

- [ ] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

### Documentation

- [x] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

I'll write the docs once this is approved

CLA see #12646


dothebart (approved):
@Simran-B WDYT?

Looks good to me.

coronabytes:
Quick question:
Looking at https://en.wikipedia.org/wiki/List_of_tz_database_time_zones would you expect DATE_TIMEZONES() to:

1. return all canonical timezones
2. return all canonical, aliases and deprecated timezones
3. just include a parameter to choose - note deprecated canonicals can be filtered easily - deprecated aliases cannot

I'm happy with just the canonical ones, exluding deprecated (with the "/" filter), however others might need all aliases

```cpp
for (auto& zone : db.zones) {
  // filter deprecated zones
  if (zone.name().find("/") == std::string::npos) {
    continue;
  }
  result->add(VPackValue(zone.name()));
}

// include aliases
for (auto& link : db.links) {
  result->add(VPackValue(link.name()));
}
```

arangod/Aql/Functions.cpp
```diff
@@ -3875,13 +3873,53 @@ AqlValue Functions::DateLocalToUtc(ExpressionContext* expressionContext,
...
   const auto zoned = make_zoned(tz, local);
   const auto tp_utc = tp_sys_clock_ms{zoned.get_sys_time().time_since_epoch()};
 
   return ::timeAqlValue(expressionContext, AFN, tp_utc);
 }
 
+/// @brief function DATE_TIMEZONE
+AqlValue Functions::DateTimeZone(ExpressionContext* expressionContext,
+                                   AstNode const&,
+                                   VPackFunctionParameters const& parameters) {
```
jsteemann, line 3887:
```suggestion
                                 AstNode const&,
                                 VPackFunctionParameters const& parameters) {
```

arangod/Aql/Functions.cpp
```diff
@@ -3875,13 +3873,53 @@ AqlValue Functions::DateLocalToUtc(ExpressionContext* expressionContext,
...
+  if (zone != nullptr) {
+    return AqlValue(zone->name());
+  }
+
+  return AqlValue(AqlValueHintNull());
+}
+
+/// @brief function DATE_TIMEZONES
+AqlValue Functions::DateTimeZones(ExpressionContext* expressionContext, AstNode const&,
+                                        VPackFunctionParameters const& parameters) {
```
jsteemann, line 3901:
```suggestion
                                  VPackFunctionParameters const& parameters) {
```

arangod/Aql/Functions.h
```diff
@@ -213,6 +213,10 @@ struct Functions {
                                  AstNode const&, VPackFunctionParameters const&);
   static AqlValue DateLocalToUtc(arangodb::aql::ExpressionContext*,
                                  AstNode const&, VPackFunctionParameters const&);
+  static AqlValue DateTimeZone(arangodb::aql::ExpressionContext*,
+                                 AstNode const&, VPackFunctionParameters const&);
```
jsteemann, line 217:
```suggestion
                               AstNode const&, VPackFunctionParameters const&);
```

arangod/Aql/Functions.h
```diff
@@ -213,6 +213,10 @@ struct Functions {
...
+  static AqlValue DateTimeZones(arangodb::aql::ExpressionContext*,
+                                  AstNode const&, VPackFunctionParameters const&);
```
jsteemann, line 219:
```suggestion
                                AstNode const&, VPackFunctionParameters const&);
```

tests/js/server/aql/aql-functions-date.js
```diff
@@ -2457,6 +2457,22 @@ function ahuacatlDateFunctionsTestSuite () {
 	  assertEqual([ "2020-02-29T23:00:00.000Z" ], 
 		getQueryResults("RETURN DATE_LOCALTOUTC(DATE_ADD(DATE_UTCTOLOCAL('2020-01-31T23:00:00.000Z', 'Europe/Berlin'), 1, 'months'), 'Europe/Berlin')"));	  
     },
+	
+	// //////////////////////////////////////////////////////////////////////////////
+    // / @brief test date_timezone
+    // //////////////////////////////////////////////////////////////////////////////
+
+    testDateTimeZone: function () {
+	  getQueryResults("RETURN DATE_TIMEZONE()", {});
```
jsteemann, line 2466: IMO this just calls the `DATE_TIMEZONE` AQL function, but ignores the function's result.
I understand we cannot assume a global default timezone in the tests, but IMHO the called function should either return `null` or a string with the tz name. I think we should at least check this here. 
But in general, how is the current tz actually determined by the date/tz library? If I read its source correctly, the date/tz library will try to read some OS files to figure it out. So there is no reliable way to anticipate that value as a "normal" database user. I am thus wondering what the use case for this AQL function will be, assuming that the default tz value will not be used inside AQL, but that for all tz conversion operations the tz names must be specified explicitly.

tests/js/server/aql/aql-functions-date.js
```diff
@@ -2457,6 +2457,22 @@ function ahuacatlDateFunctionsTestSuite () {
...
+    },
+	
+	// //////////////////////////////////////////////////////////////////////////////
+    // / @brief test date_timezones
+    // //////////////////////////////////////////////////////////////////////////////
+	
+	testDateTimeZones: function () {
+	  getQueryResults("RETURN DATE_TIMEZONES()", {});
```
jsteemann, line 2474: This does not perform any checks. I know that the list of tzs can change over time, but it is at least assumed to be always an array.
In addition, as the timezones are part of the repository now, we can also assume that this function returns timezones that are present in the tzdata in 3rdParty.
At least that is my understanding of how it works: by providing the tzdata in 3rdParty we have control over which tz names will be returned by this function.
If so, IMHO we should add a test here that checks if a few select timezone names are part of the result list.

tests/js/server/aql/aql-functions-date.js
coronabytes: yes, honestly just added the function for completness
most "cloud" servers are utc nowadays, but some of our customers just run this on premise on plain old windows machines, where we can now access the correct timezone for reports (Platform independant that is - thats the tricky part for win/nix)

coronabytes: what environment do the js test use?
this could work
console.log(Intl.DateTimeFormat().resolvedOptions().timeZone)

jsteemann: Got it. 
The js tests run inside the ArangoDB process, using the bundled V8 version (which currently is V8 7.9.317 - version may be upgraded in the future).

tests/js/server/aql/aql-functions-date.js
```diff
@@ -2457,6 +2457,37 @@ function ahuacatlDateFunctionsTestSuite () {
...
+	  
+	  try {
+	    systemtz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+	  } catch {
+	  }
+		
+	  if (systemtz) {
+	    assertEqual(systemtz, res);
+	  } else {
+		assertTrue(res != null);
```
jsteemann, line 2477: Do I understand correctly that when there is no systemtz, the DATE_TIMEZONE function should still return non-null? I would assume that it will return null then.
No matter what, instead of `assertTrue(res != null);` you can use `assertNotNull(res);`, which is more explicit and intuitive. 
(apart from that, using `res != null` could make our jslint scan for the JavaScript files fail, because it may demand a `res !== null` here. Not sure exactly, but it may happen at least for other comparisons. But it would be best to avoid this and use `assertNotNull` (or `assertNull`) instead.
By the way, using either `assertNotNull` or `assertNull` requires adding these "global" helper functions to the list of global function names at the very beginning of the test case file (line 2).

tests/js/server/aql/aql-functions-date.js
```diff
@@ -2457,6 +2457,37 @@ function ahuacatlDateFunctionsTestSuite () {
...
+	
+	// //////////////////////////////////////////////////////////////////////////////
+    // / @brief test date_timezones
+    // //////////////////////////////////////////////////////////////////////////////
+	
+	testDateTimeZones: function () {
+	  var res = getQueryResults("RETURN DATE_TIMEZONES()", {})[0];
+	  assertTrue(res.indexOf("America/New_York") !== -1);
+	  assertTrue(res.indexOf("Europe/Berlin") !== -1);
+	  assertTrue(res.indexOf("Asia/Shanghai") !== -1);
```
jsteemann, line 2489: Same here, using `assertNotEqual` is a bit more descriptive:
```suggestion
	  assertNotEqual(-1, res.indexOf("America/New_York"));
	  assertNotEqual(-1, res.indexOf("Europe/Berlin"));
	  assertNotEqual(-1, res.indexOf("Asia/Shanghai"));
```

coronabytes: Hi, thanks can you please verify that these tests run?
or is there any way to just execute the js aql tests without the cygwin shell under windows?

jsteemann: Yes, I can verify the tests on this branch later today.
I guess there is a way to execute these tests under Windows too, but as I am a Linux user that is beyond my own experience. Let me try to find someone to add instructions here...

mpoeter: Do you have [WSL (Windows Subsystem for Linux)](https://docs.microsoft.com/en-us/windows/wsl/install-win10) installed? IMHO that is the easiest way to run the tests on windows. In WSL you can simply do `ARANGOSH=path/to/arangosh.exe ./scripts/unittests shell_server_aql`

If you don't have WSL (and dont want to install it), you should be able to call arangosh directly just like the unittest script does:
```
path\to\arangosh.exe -c etc\arangodb3\arangosh.conf --log.level warning --server.endpoint none --javascript.allow-external-process-control true --javascript.execute UnitTests\unittest.js -- shell_server_aql
```

jsteemann:
Thanks for all the work so far!
I moved this into a different PR so we can run the changes in our CI:
https://github.com/arangodb/arangodb/pull/13320

I preserved all your original commits, fixed a remaining issue with jslint and fixed a few whitespace inconsistencies in the tests. The tests already worked locally, so I am positive our Jenkins will also like it.

coronabytes:
k thanks,
couldn't get tests to run with WSL2 nor allow-external-process-control true

jsteemann:
Closing in favor of https://github.com/arangodb/arangodb/pull/13320
Many thanks again!

dothebart:
hm, I'm seeing this error while running it inside a docker container on ARM:
```
    [FAILED]  tests/js/server/aql/aql-functions-date.js

      "testDateTimeZone" failed: Error: at assertion #1: assertEqual: (Etc/UTC) is not equal to (UTC)
(Error
    at assertEqual (/home/ec2-user/arangodb/js/common/modules/jsunity/jsunity.js:108:19)
    at Object.testDateTimeZone (tests/js/server/aql/aql-functions-date.js:2475:9)
```
its probably part of the functionality added by this PR?
```
    // / @brief test date_timezone

    testDateTimeZone: function () {
      let systemtz = null;
      let res = getQueryResults("RETURN DATE_TIMEZONE()", {})[0];
      
      try {
        systemtz = Intl.DateTimeFormat().resolvedOptions().timeZone;
      } catch (err) {
      }
        
      if (systemtz) {
        assertEqual(systemtz, res);
      } else {
        assertNotNull(res);
      }
    },
``` 
can you have a look @coronabytes whether you can make sure the test also passes under these conditions?

coronabytes:
well the test is quite hacky, since we don't have the exact same way on getting the system timezone, but the only way to fix this reasonably is.

```javascript
if (systemtz === "UTC") {
 systemtz = "Etc/UTC"
}
```

dothebart:
yes, these were GCC-buildroots. 
Can you create a PR with this workaround? Which is the modern one? `Etc/` ?

coronabytes:
@dothebart 
done #13393, 

Etc/UTC is the canonical one, thus prefered.
Now when testing on really old systems with legacy os timezone, like SystemV this will still fail

only option then is to just check if result is not NULL...
and only test that the function does not crash

PR #13344, opened 2021-01-07
fix spurious error in shell-pregel.js test suite

jsteemann:
### Scope & Purpose

Fix spurious error in shell-pregel.js `shell_client` or `shell_server` tests.
This is an error in the test only, which assumes that the attribute "storageTime" is always present in the result if "store" is true. However, this is not guaranteed, as the status of a Pregel run can be checked _before_ the run is completed. In that case, the "storageTime" attribute is intentionally missing from the result.
This is a test fix only, so there is no CHANGELOG entry for it.

Backport of https://github.com/arangodb/arangodb/pull/13343

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13503/

### DIFF
--- a/tests/js/common/shell/shell-pregel.js
+++ b/tests/js/common/shell/shell-pregel.js
@@ -50,12 +50,14 @@ function testAlgo(a, p) {
       assertEqual(stats.vertexCount, 11, stats);
       assertEqual(stats.edgeCount, 17, stats);
       let attrs = ["totalRuntime", "startupTime", "computationTime"];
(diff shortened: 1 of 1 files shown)


hkernbach (approved):

jsteemann:
Tests blue

KVS85 (approved):
LGTM.

PR #13378, opened 2021-01-14
Feature/read write metrics

kvahed:
### Scope & Purpose

* Added general read / write io counters to metrics
* Added histograms for runtimes in RocksDB for read, insert, replace, update, remove
* Added queue time and full URL to HttpCommTask end trace output

#### Backports:

- [ ] No backports required
- [ ] Backports required for: 3.7
#### Related Information

https://github.com/arangodb/arangodb/pull/13383

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:
https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/13598/

### Documentation


arangod/GeneralServer/HttpCommTask.cpp
```diff
@@ -602,14 +600,16 @@ void HttpCommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> baseRes,
...
   // and give some request information
   LOG_TOPIC("8f555", DEBUG, Logger::REQUESTS)
       << "\"http-request-end\",\"" << (void*)this << "\",\""
       << this->_connectionInfo.clientAddress << "\",\""
-      << GeneralRequest::translateMethod(::llhttpToRequestType(&_parser))
-      << "\",\"" << static_cast<int>(response.responseCode()) << "\","
-      << Logger::FIXED(totalTime, 6);
+      << GeneralRequest::translateMethod(::llhttpToRequestType(&_parser)) << "\",\""
+      << _url << "\",\"" << static_cast<int>(response.responseCode()) << "\",tot:"
+      << Logger::FIXED(totalTime, 6) << ",que:" << Logger::FIXED(queueTime, 6) ;
```
neunhoef, line 612: the prefixes "tot:" and "que:" make it more difficult to extract the numbers with awk, I would not add them.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -646,7 +646,7 @@ bool SupervisedScheduler::canPullFromQueue(uint64_t queueIndex) const noexcept {
   // Otherwise we might have the unlucky case that we first check dequeued,
   // then a job gets done fast (eg dequeued++, done++)
   // and then we read done.
-  uint64_t jobsDone = _jobsDone.load(std::memory_order_acquire);
+   uint64_t jobsDone = _jobsDone.load(std::memory_order_acquire);
```
neunhoef, line 649: This seems to be a typo.

arangod/Statistics/ServerStatistics.h
```diff
@@ -52,6 +52,15 @@ struct TransactionStatistics {
...
   Counter& _sequentialLocks;
+  Counter& _numRead;
+  Counter& _numWrite;
+  Counter& _numReplicate;
+  Histogram<log_scale_t<float>>& _rocksdb_insert_msec;
+  Histogram<log_scale_t<float>>& _rocksdb_read_msec;
+  Histogram<log_scale_t<float>>& _rocksdb_replace_msec;
+  Histogram<log_scale_t<float>>& _rocksdb_remove_msec;
+  Histogram<log_scale_t<float>>& _rocksdb_update_msec;
+  Histogram<log_scale_t<float>>& _rocksdb_queue_msec;
```
neunhoef, line 63: Is the rocksDB queue time actually used anywhere? And if so, what does it say?

neunhoef (changes requested):
See small comments.

neunhoef (approved):
LGTM

arangod/Statistics/ServerStatistics.h
kvahed: done

arangod/Scheduler/SupervisedScheduler.cpp
kvahed: yes

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Added read write metrics
+
```
jsteemann, line 5: The convention for our CHANGELOG is to end every entry with a full stop or something equivalent.

In addition, "Added read-write metrics" is not very understandable for end users or other interested parties that browse our CHANGELOG. What about

```
* Added metrics for document read and write operations:

  - `arangodb_num_document_reads`: Total number of document reads since process start.
  - `arangodb_num_document_writes: Total number of document writes since process start
  - `arangodb_num_document_replications`: Number of document replications since process start
  ...
```

arangod/GeneralServer/H2CommTask.cpp
```diff
@@ -477,15 +477,13 @@ void H2CommTask<T>::processStream(H2CommTask<T>::Stream& stream) {
 
   // from here on we will send a response, the connection is not IDLE
   _numProcessing.fetch_add(1, std::memory_order_relaxed);
-
+  _url = std::string((req->databaseName().empty() ? "" : "/_db/" + req->databaseName())) +
```
jsteemann, line 480: `_url` is set here and read only once a few lines later.
So we would not need an instance variable for it.
In the other CommTask implementations, `_url` is read twice and not just once. Are we missing the second place in the H2CommTask?

arangod/GeneralServer/HttpCommTask.cpp
```diff
@@ -416,15 +416,13 @@ void HttpCommTask<T>::processRequest() {
   // C functions like strchr that except a C string as input
   _request->body().push_back('\0');
   _request->body().resetTo(_request->body().size() - 1);
-
+  _url = std::string((_request->databaseName().empty() ? "" : "/_db/" + _request->databaseName())) +
+    (Logger::logRequestParameters() ? _request->fullUrl() : _request->requestPath());
```
jsteemann, line 420: This pessimizes the case in which we don't log requests (not logging requests is the default, by the way).
Instead of unconditionally storing the url in an instance variable `_url`, we could make the log messages call a new function named `url()`, which computes the url value only when needed, and then stores the computed value in the `_url` instance variable.
That way we would avoid computation of `_url` in case there is no logging, and still benefit if the `_url` value is logged in 2 places.
That idea will only work however if the `url()` function can access `_request` and `_request` is still valid while `url()` is called. I don't know if that is the case or not, but we could try it.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -714,7 +721,9 @@ Result RocksDBCollection::truncate(transaction::Methods& trx, OperationOptions&
 
   // avoid OOM error for truncate by committing earlier
   uint64_t const prvICC = state->options().intermediateCommitCount;
-  state->options().intermediateCommitCount = std::min<uint64_t>(prvICC, 10000);
+  auto const tmp = std::min<uint64_t>(prvICC, 10000);
+  state->options().intermediateCommitCount = tmp;
+  _statistics._numWrite.count(tmp);
```
jsteemann, line 726: This does not make sense to me. We are setting the intermediate commit value here, but this is not a write operation that should be counted. It is just a (temporary) change of settings.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -869,18 +893,29 @@ bool RocksDBCollection::readDocument(transaction::Methods* trx,
...
 
+  if (options.isSynchronousReplicationFrom.empty()) {
+    ++_statistics._numWrite;
+  } else {
+    ++_statistics._numReplicate;
+  }
+
+  using namespace std::chrono;
+  using clock = steady_clock;
+  auto start = clock::now();
```
jsteemann, line 916: If we exit the `insert()` method early (it has mutliple exits), then we increase the write counter but don't update the timers. This is inconsistent.
It would be ok if `_numWrite` / `_numReplicate` were labeled "attempted writes" / "attempted replication writes", but currently they aren't.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -944,13 +979,25 @@ Result RocksDBCollection::insert(arangodb::transaction::Methods* trx,
...
+
+  if (options.isSynchronousReplicationFrom.empty()) {
+    ++_statistics._numWrite;
+  } else {
+    ++_statistics._numReplicate;
+  }
+
+  using namespace std::chrono;
+  using clock = steady_clock;
+  auto start = clock::now();
```
jsteemann, line 999: same here, in case of early exit we only update the counters but not the timers.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -944,13 +979,25 @@ Result RocksDBCollection::insert(arangodb::transaction::Methods* trx,
...
   return res;
 }
 
 Result RocksDBCollection::update(transaction::Methods* trx,
                                  velocypack::Slice newSlice,
                                  ManagedDocumentResult& resultMdr, OperationOptions& options,
                                  ManagedDocumentResult& previousMdr) {
+
+  if (options.isSynchronousReplicationFrom.empty()) {
+    ++_statistics._numWrite;
```
jsteemann, line 992: We now have seperate metrics for insert, update and remove when it comes to timers.
But we don't have seperate metrics for insert, update and remove when it comes to counters.
Should we also split the counters in the same way as the timers?

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -680,6 +685,8 @@ Result RocksDBCollection::truncate(transaction::Methods& trx, OperationOptions&
     _meta.adjustNumberDocuments(seq, /*revision*/ newRevisionId(),
                                 -static_cast<int64_t>(numDocs));
 
+    _statistics._numWrite.count(numDocs);
```
jsteemann, line 688: Here we don't actually write documents, but we place a DeleteRange, which is not the same as placing x remove operations.
In addition if someone wants to calculcate the write operations total time divided by the total number of write operations, then the results would be much skewed by us counting the number of writes here but no timings.
I suggest we don't increase `_numWrite` here at all.

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -210,6 +211,7 @@ class RocksDBCollection final : public RocksDBMetaCollection {
   bool _cacheEnabled;
   /// @brief number of index creations in progress
   std::atomic<int> _numIndexCreations;
+  arangodb::TransactionStatistics& _statistics;
```
jsteemann, line 214: Is it guaranteed that the TransactionStatistics object always outlives all RocksDBCollection instances? I think under normal circumstances this is the case, but I am not sure about unit tests/mocks.

arangod/Statistics/ServerStatistics.cpp
```diff
@@ -37,27 +37,51 @@ TransactionStatistics::TransactionStatistics(MetricsFeature& metrics)
...
+                         "collections to avoid deadlocking")),
+      _numRead(
+        _metrics.counter("arangodb_num_document_reads", 0,
+                         "Number of document reads since process start")),
+      _numWrite(
+        _metrics.counter("arangodb_num_document_writes", 0,
+                         "Number of document writes since process start")),
+      _numReplicate(
+        _metrics.counter("arangodb_num_document_replications", 0,
+                         "Number of document replications since process start")),
```
jsteemann, line 69: I would prefer "replication write operations" here.

arangod/Statistics/ServerStatistics.cpp
```diff
@@ -37,27 +37,51 @@ TransactionStatistics::TransactionStatistics(MetricsFeature& metrics)
...
+      _rocksdb_insert_msec(
+        _metrics.histogram("arangodb_insert_time",
+                           log_scale_t(std::exp(1.f), 10.0f, 1.e6f, 10), "Insert a document [us]")),
```
jsteemann, line 72: What about "Total document insertion time [us]"?

arangod/GeneralServer/H2CommTask.cpp
kvahed: true. fixed

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: the plan was to record the successful insert times in the histogram. this is detached from the counts for replication / write

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: That is correct was as intended.

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: that is correct. this was also intended. write / read / replicate are one class of counters. 
the much more specific insert, ... operations are recorded by the histograms, that inherently collect counters as a sum of all records.

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: the idea was to record deletions as writes. in a sense comparable to read and write operations on a posix file system, where deletions, truncations, modifications and creations are all counted as write.

arangod/RocksDBEngine/RocksDBCollection.h
kvahed: I am not sure about guarantees. Tests never failed, however. Not much of a proof. One can always break this dependency. I am not sure, what you expect how to fix this.

arangod/Statistics/ServerStatistics.cpp
kvahed: If you don't mind, let's leave this concise. It is not ambiguous.

arangod/Statistics/ServerStatistics.cpp
kvahed: Sure

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: now separately as truncations

arangod/RocksDBEngine/RocksDBCollection.cpp
kvahed: The counts are now documented to be write *operations* rather than writes and the histograms document to be *successful* operations

arangod/Statistics/ServerStatistics.cpp
kvahed: done

arangod/RestServer/Metrics.h
```diff
@@ -93,52 +93,74 @@ template<typename T> class Gauge : public Metric {
   Gauge(T const& val, std::string const& name, std::string const& help,
         std::string const& labels = std::string())
     : Metric(name, help, labels) {
-    _g.store(val);
+    _g.store(val, std::memory_order_relaxed);
```
mpoeter, line 96: You could put this into the initializer list instead of using a separate store:
`: Metric(name, help, labels), _g(val) {}`

arangod/RestServer/Metrics.h
```diff
@@ -93,52 +93,74 @@ template<typename T> class Gauge : public Metric {
...
   }
   Gauge(Gauge const&) = delete;
   ~Gauge() = default;
   std::ostream& print (std::ostream&) const;
-  Gauge<T>& operator+=(T const& t) {
+
+  T fetch_add(T t, std::memory_order mo = std::memory_order_relaxed) noexcept {
```
mpoeter, line 102: Not sure if we should use `memory_order_relaxed` as default here. I don't think we will ever have a case where we actually need something stronger than relaxed, but it feels not right. However, if we use `memory_order_seq_cst` as default order we should probably adapt all callers (even though on x86 we wouldn't gain anything), which would be a much larger change. So guess for now we can leave it as is.

mpoeter (approved):

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -714,7 +721,9 @@ Result RocksDBCollection::truncate(transaction::Methods& trx, OperationOptions&
 
   // avoid OOM error for truncate by committing earlier
   uint64_t const prvICC = state->options().intermediateCommitCount;
-  state->options().intermediateCommitCount = std::min<uint64_t>(prvICC, 10000);
+  auto const tmp = std::min<uint64_t>(prvICC, 10000);
+  state->options().intermediateCommitCount = tmp;
+  _statistics._numTruncate.count(tmp);
```
jsteemann, line 726: This is not correct. `tmp` will contain the intermediate commit threshold value, which is not the number of documents removed/truncated.
The number of documents truncated is only determined further below in this method, and is only known in total when the `for` loop is completed.

arangod/RestServer/Metrics.h
kvahed: i did see this dilemma and also decided that this was the pragmatic solution

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13710/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13711/

jsteemann:
Tests blue

PR #13385, opened 2021-01-14
Reduced the amount of test combinations in (K-)ShortestPathsTests.

mchacki:
This test does not have any User Impact.

It only reduces the amount of tested combinations of input / data in the tests.
It is fine to have the input format tested once, and the data tested once.
This speeds up initial load of tests by ~66% on my machine.
Also speeds up the tests of the specific files by factor of ~10 or more.

Jenkins:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13604/


jsteemann (approved):
Tried it out, and it works fine and fast for me!

PR #13389, opened 2021-01-15
Replaced MAINTAINERMODE with GTEST makro to include testing code

mchacki:
This PR has no User Impact.
The modified place inject some test-case preparation code.
This should only be enabled with the GTEST flag enabled, and is independent of MAINTAINER_MODE which was used here wrongly.
Jenkins runs either both on or both of, so it slipped through.

Jenkins:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13613/


dothebart (approved):

dothebart:
tests blue.

PR #13398, opened 2021-01-18
disable old-style subqueries

jsteemann:
### Scope & Purpose

Disable old-style subqueries. 
This PR makes the "splice-subqueries" optimizer rule mandatory, in the sense that it cannot be deactivated.
Any query execution plans created by 3.8 will not have contain any execution nodes of type SubqueryNode, because the "splice-subqueries" optimizer rule will replace them with a combination of SubqueryStartNode and SubqueryEndNode.
During plan creation and optimization, there can still be SubqueryNodes, as before. 

We also need to keep the executor for old-style subqueries around in 3.8, in case a 3.7 coordinator sends an execution plan with a SubqueryNode inside during rolling upgrade. These subqueries still need to work on 3.8 DB servers. With 3.9, the SubqueryExecutor can be removed, because it will not be called anymore.

Because only the old subqueries used the AqlValue internal type DOCVEC, this was also removed from AqlValue, allowing the simplfication of AqlValue's internals. The result buildup in the old SubqueryExecutor is now performed using a Velocypack Builder/Buffer instead of a DOCVEC.
There is a unit test for this, which still creates and trigger old-style subqueries by setting a specific failure point when creating a query.

This PR should not have any user-visible changes for normal setups, because the "splice-subqueries" rule was always enabled since it was introduced. A change in behavior only occurs if end users turned this optimizer rule off deliberately. In this case, the rule will now (silently) still be turned on. Query results should not change due to this.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/629

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_server_aql, shell_server, gtest*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13633/


jsteemann:
Tests blue

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13635/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13637/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13642/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13657/

tests/js/server/aql/aql-optimizer-rule-remove-filter-covered-by-index.js
```diff
@@ -287,14 +287,14 @@ function optimizerRuleTestSuite() {
         var result;
         result = AQL_EXPLAIN(query, { }, paramIndexRangeFilter);
         assertEqual([ IndexesRule, FilterRemoveRule ], 
-          removeAlwaysOnClusterRules(result.plan.rules), query);
+          removeAlwaysOnClusterRules(result.plan.rules).filter((r) => r !== "splice-subqueries"), query);
```
goedderz, line 290: Would it make sense to add something like `removeAlwaysOnRules`, which we could use everywhere instead of the `.filter((r) => r !== "splice-subqueries")`? It's fine with me to leave it as it is though, we can always add it if we get another "always on" rule that affectes a lot of tests.

tests/js/server/aql/aql-optimizer-rule-use-index-for-sort.js
```diff
@@ -1109,7 +1084,7 @@ function optimizerRuleTestSuite() {
 
     testSortDescWithFilterSplicedSubquery : function () {
       const query = `FOR i IN [123] RETURN (FOR v IN ${colName} FILTER v.a == i SORT v.b DESC RETURN v)`;
-      const plan = AQL_EXPLAIN(query, {}, {optimizer: {rules: ['+splice-subqueries']}}).plan;
+      const plan = AQL_EXPLAIN(query).plan;
```
goedderz, line 1087: I think you can delete one of `testSortDescWithFilterSubquery`, `testSortDescWithFilterSplicedSubquery`.

tests/js/server/aql/aql-profiler-cluster.js
```diff
@@ -55,7 +55,7 @@ function ahuacatlProfilerTestSuite () {
     EnumerateListNode, EnumerateViewNode, FilterNode, GatherNode, IndexNode,
     InsertNode, LimitNode, NoResultsNode, RemoteNode, RemoveNode, ReplaceNode,
     ReturnNode, ScatterNode, ShortestPathNode, SingletonNode, SortNode,
-    SubqueryNode, TraversalNode, UpdateNode, UpsertNode } = profHelper;
+    TraversalNode, UpdateNode, UpsertNode } = profHelper;
```
goedderz, line 58: Can `SubqueryBlock` (a few lines below) be removed, too?

tests/js/server/aql/aql-profiler-noncluster-nightly.js
```diff
@@ -53,7 +53,7 @@ function ahuacatlProfilerTestSuite () {
     EnumerateListNode, EnumerateViewNode, FilterNode, GatherNode, IndexNode,
     InsertNode, LimitNode, NoResultsNode, RemoteNode, RemoveNode, ReplaceNode,
     ReturnNode, ScatterNode, ShortestPathNode, SingletonNode, SortNode,
-    SubqueryNode, TraversalNode, UpdateNode, UpsertNode } = profHelper;
```
goedderz, line 56: Can `SubqueryBlock` (a few lines below) be removed, too?

tests/js/server/aql/aql-profiler-noncluster.js
```diff
@@ -51,7 +51,7 @@ function ahuacatlProfilerTestSuite () {
     EnumerateListNode, EnumerateViewNode, FilterNode, GatherNode, IndexNode,
     InsertNode, LimitNode, MutexNode, NoResultsNode, RemoteNode, RemoveNode, ReplaceNode,
     ReturnNode, ScatterNode, ShortestPathNode, SingletonNode, SortNode,
-    SubqueryNode, TraversalNode, UpdateNode, UpsertNode } = profHelper;
```
goedderz, line 54: Can `SubqueryBlock` (a few lines below) be removed, too?

tests/js/server/aql/aql-profiler.js
```diff
@@ -68,7 +68,7 @@ function ahuacatlProfilerTestSuite () {
     EnumerateListNode, EnumerateViewNode, FilterNode, GatherNode, IndexNode,
     InsertNode, LimitNode, NoResultsNode, RemoteNode, RemoveNode, ReplaceNode,
     ReturnNode, ScatterNode, ShortestPathNode, SingletonNode, SortNode,
-    SubqueryNode, TraversalNode, UpdateNode, UpsertNode } = profHelper;
```
goedderz, line 71: Can `SubqueryBlock` (a few lines below) be removed, too?

goedderz (approved):

arangod/Aql/Optimizer.cpp
```diff
@@ -370,6 +376,15 @@ void Optimizer::createPlans(std::unique_ptr<ExecutionPlan> plan,
...
     plan.first->findVarUsage();
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+    TRI_IF_FAILURE("Optimizer::allowOldSubqueries") {
+      // intentionally let old subqueries pass. this is used only in testing and can be removed in 3.9
+      continue;
+    }
+
+    NoSubqueryChecker checker;
+    plan.first->root()->walk(checker);
+#endif // ARANGODB_ENABLE_MAINTAINER_MODE
```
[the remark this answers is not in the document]
jsteemann: I don't think we do. I will test the upgrade manually now.

tests/js/server/aql/aql-profiler-cluster.js
jsteemann: Yes :+1:

tests/js/server/aql/aql-profiler-noncluster-nightly.js
jsteemann: Yes :+1:

tests/js/server/aql/aql-profiler-noncluster.js
jsteemann: Yes :+1:

tests/js/server/aql/aql-profiler.js
jsteemann: Yes :+1:

tests/js/server/aql/aql-optimizer-rule-use-index-for-sort.js
jsteemann: Yes :+1:

arangod/Aql/Optimizer.cpp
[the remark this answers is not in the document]
jsteemann: Verified 3.7->devel rolling upgrade manually, with `-splice-subqueries`. Seems to work fine.

jsteemann:
Restarted Jenkins after latest changes:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13659/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13663/

PR #13405, opened 2021-01-20
Feature/const registers for const values

mpoeter:
### Scope & Purpose

This is a RFC.

At the moment, the only execution node that can handle expressions is the `CalculationNode`. All other nodes take a variable as input. This has the effect, that even constant values are kept in a variable, which also requires a register, resulting in increased memory usage and perhaps unnecessary copies of that constant.

This PR introduces "const registers". Register id values <=1000 are _variable registers_ (same as before); register id values >= 2000 are _const registers_.  
A new optimizer rule marks variables which are set by expressions that resolve to a constant value. These variables are then assigned to const registers. The AqlItemBlockManager now holds an additional AqlItemBlock for all the const values. All operations that fetch a value from an AqlItemBlock using a const register are forwarded to that const-value-block.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:


arangod/Aql/AqlItemBlock.h
```diff
@@ -133,9 +133,10 @@ class AqlItemBlock {
   template <typename... Args>
   // std::enable_if_t<!(std::is_same<AqlValue,std::decay_t<Args>>::value || ...), void>
   void emplaceValue(size_t index, RegisterId varNr, Args&&... args) {
-    auto address = getAddress(index, varNr);
+    TRI_ASSERT(varNr.isVariableRegister());
+    auto address = getAddress(index, varNr.rawValue());
     AqlValue* p = &_data[address];
-    TRI_ASSERT(p->isEmpty());
+    //TRI_ASSERT(p->isEmpty());
```
jsteemann, line 139: Can we add this assertion back?

arangod/Aql/AqlItemBlockManager.h
```diff
@@ -110,6 +116,8 @@ class AqlItemBlockManager {
   };
 
   Bucket _buckets[numBuckets];
+
+  AqlItemBlock* _constValueBlock = nullptr;
```
jsteemann, line 120: Can we make this an `std::unique_ptr<AqlItemBlock>` instead of a raw pointer?
```suggestion
  AqlItemBlock* _constValueBlock = nullptr;
```

arangod/Aql/CalculationExecutor.cpp
```diff
@@ -147,6 +147,17 @@ bool CalculationExecutor<calculationType>::shouldExitContextBetweenBlocks() cons
...
 
+template <>
+void CalculationExecutor<CalculationType::Constant>::doEvaluation(InputAqlItemRow& input,
+                                                                   OutputAqlItemRow& output) {
+  TRI_IF_FAILURE("CalculationBlock::executeExpression") {
+    THROW_ARANGO_EXCEPTION(TRI_ERROR_DEBUG);
+  }
+
+  // TODO - really nothing to do here?
+  // we at least have to set _inputRowCopied, otherwise subsequent assertions fail
```
jsteemann, line 158: We should clarify whether this is necessary before merging this PR.

arangod/Aql/ClusterNodes.cpp
```diff
@@ -305,7 +305,8 @@ std::unique_ptr<ExecutionBlock> DistributeNode::createBlock(
     TRI_ASSERT(it != getRegisterPlan()->varInfo.end());
     regId = (*it).second.registerId;
 
-    TRI_ASSERT(regId < RegisterPlan::MaxRegisterId);
+    TRI_ASSERT(regId.isConstRegister() || regId < RegisterPlan::MaxRegisterId);
```
jsteemann, line 308:
```suggestion
    TRI_ASSERT(regId.isConstRegister() || regId < RegisterPlan::MaxRegisterId.rawValue());
```

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1802,6 +1802,12 @@ RegisterId ExecutionBlockImpl<IdExecutor<SingleRowFetcher<BlockPassthrough::Enab
   return _executorInfos.getOutputRegister();
 }
 
+template <>
+template <>
+RegisterId ExecutionBlockImpl<ReturnExecutor>::getInputRegisterId() const noexcept {
```
jsteemann, line 1807:
```suggestion
template <>
RegisterId ExecutionBlockImpl<ReturnExecutor>::getInputRegisterId() const noexcept {
```

arangod/Aql/CalculationExecutor.cpp
mpoeter: This should be obsolete once we completely remove the CalculationNode from the plan.

arangod/Aql/AqlItemBlockManager.h
mpoeter: Unfortunately not, because the `AqlItemBlock` destructor is protected.

arangod/Aql/ExecutionBlockImpl.cpp
mpoeter: This is actually correct as it is.

arangod/Aql/OptimizerRules.cpp
```diff
@@ -1589,6 +1589,57 @@ void arangodb::aql::propagateConstantAttributesRule(Optimizer* opt,
...
+  ::arangodb::containers::SmallVector<ExecutionNode*> distributeNodes{a};
+  plan->findNodesOfType(distributeNodes, EN::DISTRIBUTE, true);
+
+  // In some cases the DistributeNode needs to write into the inputVariable's
+  // register, so in these cases we cannot use a ConstRegister.
+  VarSet vars;
+  for (auto const& n : distributeNodes) {
+    auto nn = ExecutionNode::castTo<DistributeNode*>(n);
+    vars.insert(nn->variable());
+    vars.insert(nn->alternativeVariable());
```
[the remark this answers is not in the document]
mpoeter: Yes, why not - we only use this set for a simple contains check, so it doesn't hurt if we also add null values.

arangod/Aql/Variable.cpp
```diff
@@ -43,30 +43,33 @@ char const* const Variable::NAME_CURRENT = "$CURRENT";
...
 Variable::Variable(arangodb::velocypack::Slice const& slice)
-    : Variable(arangodb::basics::VelocyPackHelper::checkAndGetStringValue(slice, "name"),
-               arangodb::basics::VelocyPackHelper::checkAndGetNumericValue<VariableId>(slice, "id"),
-               arangodb::basics::VelocyPackHelper::getBooleanValue(slice, "isDataFromCollection", false)) {}
+    : id(arangodb::basics::VelocyPackHelper::checkAndGetNumericValue<VariableId>(slice, "id")),
+      name(arangodb::basics::VelocyPackHelper::checkAndGetStringValue(slice, "name")),
+      value(),
+      isDataFromCollection(arangodb::basics::VelocyPackHelper::getBooleanValue(slice, "isDataFromCollection", false)),
+      hasConstValue(arangodb::basics::VelocyPackHelper::getBooleanValue(slice, "hasConstValue", false)),
+      constantValue(slice.get("constantValue")) {}
```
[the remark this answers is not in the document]
mpoeter: I also considered that, but if we later want to introduce virtual registers it might make sense to have a separate type field. So instead I would suggest to replace the `hasConstValue` with an enum field `type`.

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13672/

arangod/Aql/Variable.cpp
```diff
@@ -79,6 +82,10 @@ void Variable::toVelocyPack(VPackBuilder& builder) const {
   builder.add("id", VPackValue(id));
   builder.add("name", VPackValue(name));
   builder.add("isDataFromCollection", VPackValue(isDataFromCollection));
+  builder.add("hasConstValue", VPackValue(hasConstValue));
+  builder.add(VPackValue("constantValue"));
+  constantValue.toVelocyPack(nullptr, builder, /*resolveExternals*/ false,
+                              /*allowUnindexed*/ true);
```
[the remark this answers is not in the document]
mpoeter: Done

arangod/Aql/Variable.cpp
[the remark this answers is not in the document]
mpoeter: Done

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13673/

arangod/Aql/types.h
```diff
@@ -60,10 +60,69 @@ struct Collection;
...
-typedef unsigned RegisterId;
 typedef size_t RegisterCount;
 
+/// @brief type for register numbers/ids
+struct RegisterId {
+  using value_t = unsigned;
+  RegisterId() = default;
+
+  static constexpr unsigned maxVariableRegisterId = 1000;
+  static constexpr unsigned constRegisterIdOffset = 2000;
```
goedderz, line 71: Can we use a way to distinguish regular and const registers that's easier to change later, if necessary? For example something like (just the idea)
```
struct Register {
    enum class Type { Regular, Const };
    Type _type : 1;
    unsigned _registerId : 10;
};
```
. That feels more closely related to a sum type (i.e. `type RegisterId = RegularRegister Int | ConstRegister Int`), and might be a little nicer to extend (e.g. for virtual registers).

arangod/Aql/types.h
```diff
@@ -60,10 +60,68 @@ struct Collection;
...
+  }
+
+  constexpr bool operator!=(const RegisterId& rhs) const noexcept {
+    return _value != rhs._value;
+  }
+
+ private:
+  static_assert(maxVariableRegisterId < constRegisterIdOffset);
+
+  unsigned _value;
```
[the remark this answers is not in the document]
goedderz: That would disallow assignments to a `RegisterId`, which will probably break some code, e.g. loops over registers.

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13699/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13939/

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+    return RegisterId(value, Type::Const);
+  }
+
+  constexpr static RegisterId makeRegular(value_t value) noexcept {
+    return RegisterId(value, Type::Regular);
+  }
+
+  constexpr static RegisterId fromUInt32(uint32_t value) noexcept {
+    auto v = static_cast<value_t>(value);
+    auto type = static_cast<Type>(value >> (sizeof(value_t) * 8));
```
jsteemann, line 95: what is the behavior if an invalid input value is passed into this function, e.g. UINT32_MAX?

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+    RegisterId result (v, type);
```
jsteemann, line 96:
```suggestion
    RegisterId result(v, type);
```

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+    ++_value;
+    TRI_ASSERT(isValid());
+    return *this;
+  }
+
+  constexpr RegisterId& operator++(int) noexcept {
+    ++_value;
+    TRI_ASSERT(isValid());
+    return *this;
+  }
```
jsteemann, line 129: Do we actually need these operators? Why would one want to increase the value of an existing Register?

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+
+  constexpr RegisterId operator+(int16_t v) noexcept {
+    return RegisterId(_value + v, _type);
+  }
+
+  constexpr RegisterId operator-(int16_t v) noexcept {
+    return RegisterId(_value - v, _type);
```
jsteemann, line 136: I am wondering what we need this operator for.
Should we check for underflow here?

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+  constexpr bool operator==(const RegisterId& rhs) const noexcept {
+    return type() == rhs.type() && _value == rhs._value;
+  }
+
+  constexpr bool operator==(const value_t& rhs) const noexcept {
+    return _value == rhs;
+  }
+
+  constexpr bool operator!=(const RegisterId& rhs) const noexcept {
+    return type() != rhs.type() || _value != rhs._value;
```
jsteemann, line 167: I suggest implementing `operator!=` as `!(*this == rhs);`, because that allows us to have the logical in a single place.

arangod/Aql/types.h
```diff
@@ -60,9 +60,121 @@ struct Collection;
...
+  }
+
+  constexpr bool operator!=(const value_t& rhs) const noexcept {
+    return _value != rhs;
+  }
+ private:
+  uint16_t _value = 0;
```
jsteemann, line 174: Wondering why we can't make the register value const here.
Does the value of a register ever change after creation?

arangod/Aql/Variable.h
```diff
@@ -76,6 +85,18 @@ struct Variable {
...
+  // on the constantValue. If constantValue.isNone, the type is Type::Regulor,
+  // otherwise it is Type::Const
+  Type type() const noexcept;
+
+  /// @brief returns the constant value of the variable.
+  AqlValue constantValue() const noexcept { return _constantValue; }
+
+  /// @brief set the constant value of the variable.
+  /// This implicitly changes the type -> see type()
+  void setConstantValue(AqlValue value) noexcept;
```
jsteemann, line 98: With that, we will have `constValue()` and `constantValue()` which is confusing from an API perspective.
Can we rename the older `constValue()` function to something completely different here, so the two are not confused? Mid-term we should also clean this up, but a short-term fix could be just renaming the functions.

arangod/Aql/RegisterPlan.h
```diff
@@ -119,12 +119,15 @@ struct RegisterPlanT final : public std::enable_shared_from_this<RegisterPlanT<T
...
+  RegisterCount nrConstRegs = 0;
+
   // We collect the subquery nodes to deal with them at the end:
   std::vector<T*> subqueryNodes;
 
   /// @brief maximum register id that can be assigned, plus one.
   /// this is used for assertions
-  static constexpr RegisterId MaxRegisterId = RegisterId{1000};
+  static constexpr RegisterId MaxRegisterId = RegisterId(RegisterId::maxRegisterId);
+  // TODO - remove MaxRegisterId in favor of RegisterId::maxRegisterId
```
jsteemann, line 130: What is keeping us from doing this now?

mpoeter: Nothing really, but a quick search shows that there are still 139 usages of `MaxRegisterId` and I was a bit hesitant to make this change in this PR so that it does not become unnecessarily big. But if you prefer I am happy to do it now.

arangod/Aql/types.h
mpoeter: These where helpful during transition, but you are right that we can get rid of them now.

arangod/Aql/types.h
mpoeter: These where helpful during transition, but you are right that we can get rid of them now.

arangod/Aql/types.h
mpoeter: Yes, there are several cases where a `RegisterId` variable is assigned a new value (copy-assignment).

jsteemann: Ah, ok

arangod/Aql/RegisterPlan.h
jsteemann: Ok, if you haven't start it yet, then let's do it later.
However, it looks like a trivial search&replace, so we should do it after this PR has been merged.

arangod/Aql/types.h
mpoeter: Then `isValid()` would be false. At the moment there is only an assertion, would you prefer to throw an exception if we get some invalid value here?

jsteemann: I would be happier if it wasn't possible to create invalid `RegisterId` objects at all.

jsteemann: Allowing to create an object with invalid state and later having to check if it is valid feels a bit like in the 80s.

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13969/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13970/

arangod/Aql/types.h
mpoeter: Done!

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13975/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13979/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13984/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13985/

arangod/Aql/types.h
```diff
@@ -109,4 +106,20 @@ enum class ExplainRegisterPlan {
 
 }  // namespace arangodb
 
+namespace std {
```
jsteemann, line 109: Probably these should be moved into RegisterId.h, too. But we can do this later.

jsteemann (approved):

jsteemann:
Ran ASan tests on this branch for single-server and cluster, and both did not report any issues!

arangod/Aql/Variable.h
```diff
@@ -76,16 +86,25 @@ struct Variable {
 
   bool isEqualTo(Variable const& other) const;
   
+  /// @brief returns the type of the variable. The type is determined based
+  // on the constantValue. If constantValue.isNone, the type is Type::Regulor,
```
goedderz, line 90:
```suggestion
  // on the constantValue. If constantValue.isNone, the type is Type::Regular,
```

arangod/Aql/RegisterId.h
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  constexpr Type type() const noexcept { return _type; }
+
+  constexpr bool isConstRegister() const noexcept { return _type == Type::Const; }
+  constexpr bool isRegularRegister() const noexcept { return _type == Type::Regular; }
+
+  constexpr bool isValid() const noexcept { return _value < maxRegisterId && _type <= Type::Const; }
+
+  constexpr bool operator<(const RegisterId& rhs) const noexcept {
```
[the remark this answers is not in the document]
mpoeter: Yeah, I'm still not fully used to that. 😉

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -201,7 +201,7 @@ std::unique_ptr<OutputAqlItemRow> ExecutionBlockImpl<Executor>::createOutputRow(
         if (!hasShadowRows || !newBlock->isShadowRow(row)) {
           for (auto const& reg : regs) {
             AqlValue const& val = newBlock->getValueReference(row, reg);
-            TRI_ASSERT(val.isEmpty());
+            TRI_ASSERT(val.isEmpty() == reg.isRegularRegister());
```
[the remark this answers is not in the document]
mpoeter: No, const registers are basically read-only, so it would not make sense to have a const output register.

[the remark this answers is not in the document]
goedderz: Shouldn't the assertion then say `&&` instead of `==`?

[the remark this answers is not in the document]
mpoeter: Of course...

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14024/

goedderz (approved):

mpoeter:
Tests blue

PR #13433, opened 2021-01-23
improve performance of shard synchronization protocol

jsteemann:
### Scope & Purpose

Improve performance of shard synchronization protocol by calling a special version of the inventory, which will only return structural information for the requested shard, and not for all shards in the requested database. This can reduce the amount of data transferred in case there are many shards present and they are mostly in sync.

The smaller response format is triggered by sending the "collection" URL parameter to the `/_api/replication/inventory` endpoint. This is an optional parameter, so there shouldn't be any issues if it is not set in any request.

In 3.8, the shard synchronization will use that smaller format by default, by setting that URL parameter when calling the inventory during shard synchronization. 
Any calls from a 3.7 follower to a 3.8 leader will get the same (verbose) answers as before, as 3.7 will need send the new URL parameter.
In addition, asking a 3.7 leader for the smaller format from a 3.8 follower will also produce the previous (verbose) answers, because 3.7 will not check for the new URL parameter.

The URL parameter is supposed to be used internally, in DB server-to-DB server communication.

In addition, this PR saves 3 cluster-internal requests per shard in the initial shard synchronization protocol by reusing already existing information across the different steps of the replication process.
The first request is saved by combining the two previously separate requests for fetching the leader's state and the starting of a replication batch in the `DatabaseInitialSyncer`.
Another request is saved in the `DatabaseTailingSyncer` by not querying the leader state again, because we can simply reuse the leader state from the DatabaseInitialSyncer her. As there are two usages of the `DatabaseTailingSyncer` during shard synchronization, we save twice.
This sums up to 3 saved requests per shard.

TODO: should be added to arangosync as well, in case it uses this API.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *all cluster tests*.
  - [x] Added new **integration tests** (e.g. in http_replication)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13712/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 devel
 -----
 
+* Improve shard synchronization protocol by only transferring the required
+  parts of the inventory from leader to follower. Previously, for each shard
+  the entire inventory was exchanged, which included all shards of the
+  respective database with all their details.
+
 * Fixed some wrong behavior in single document updates. If the option
   ignoreRevs=false was given and the precondition _rev was given in the body
   but the _key was given in the URL path, then the rev was wrongly taken
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,7 +5,7 @@ devel
   parts of the inventory from leader to follower. Previously, for each shard
   the entire inventory was exchanged, which included all shards of the
   respective database with all their details.
-  In addition, save 3 cluster-internal requests per shard in the shard
+  In addition, save 3 cluster-internal requests per shard in the initial shard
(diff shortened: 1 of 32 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13713/

jsteemann:
Tests blue

neunhoef (approved):
LGTM

PR #13437, opened 2021-01-25
improve memory efficiency of SharedPRNG

jsteemann:
### Scope & Purpose

Improve memory efficiency of the shared PRNG by not using two cache lines per PRNG bucket, and by making sure
that each bucket starts exactly at a cache line boundary (still assumed to be 64 bytes as in the previous version of the code).
The memory usage of the shared PRNG instance inside arangod is now 4 MB instead of 8 MB, while it still provides the same number of buckets and protection against false sharing as before.

In addition, the SharedPRNG now is not a static global variable anymore, which allows spinning up multiple PRNGs, and makes testing easier.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtests, at least there the PRNG is used*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13722/


arangod/Aql/QueryCache.cpp
```diff
@@ -800,7 +800,7 @@ void QueryCache::excludeSystem() {
 /// @brief determine which lock to use for the cache entries
 unsigned int QueryCache::getPart(TRI_vocbase_t const* vocbase) const {
   uint64_t v = uintptr_t(vocbase);
-  return static_cast<int>(fasthash64_uint64(v, 0xf12345678abcdef) % numberOfParts);
+  return static_cast<int>(fasthash64_uint64(v, 0xf12345678abcdef) & (numberOfParts - 1));
```
mpoeter, line 803:
```suggestion
  return static_cast<int>(fasthash64_uint64(v, 0xf12345678abcdef) % numberOfParts);
```
I would prefer to keep the % version, because it is clearer and it also works correctly should `numberOfParts` be changed to something that is not a power of two. And in case it is a power of two, compilers are smart enough to translate the modulo operation to a bitwise-and themselves.

arangod/Cache/FrequencyBuffer.h
```diff
@@ -32,7 +32,8 @@
 #include <utility>
 #include <vector>
 
-#include "Basics/SharedPRNG.h"
+#include "ApplicationFeatures/SharedPRNGFeature.h"
+#include "Basics/debugging.h"
```
mpoeter, line 36:
```suggestion
```
I did not find anything that would require this include. Did I miss something?

mpoeter (changes requested):
The `SharedPRNGFeature` can be significantly simplified. C++17 finally support dynamic allocation of overaligned types. So `_prngs` can simply be a `std::unique_ptr<PaddedPRNG[]>` which would simplify initialization and we could revert to a simple default destructor.

jsteemann:
> The `SharedPRNGFeature` can be significantly simplified. C++17 finally support dynamic allocation of overaligned types. So `_prngs` can simply be a `std::unique_ptr<PaddedPRNG[]>` which would simplify initialization and we could revert to a simple default destructor.

I know, but our allocator doesn't yet.

arangod/Cache/FrequencyBuffer.h
jsteemann: Yes, TRI_ASSERT(...) was used in this file (even before my change), so following the include-what-you-use scheme I added this include here.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13729/

mpoeter (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13734/

PR #13449, opened 2021-01-26
Various micro-optimizations

jsteemann:
### Scope & Purpose

This PR contains a zoo of small changes:
* Adds new metric `arangodb_transactions_expired` to track the total number of expired and then garbage-collected transactions
* Clarifies scope of another metric in the CHANGELOG
* Improves debug and warning log messages, by providing a bit more context
* Downgrades some log messages for recoverable errors into warnings (to comply with our log levels policy)
* micro-optimization for arangodump (no allocation of StringBuffer for masking data when no masking is used)
* micro-optimization for `/_api/import` handling: we can turn off `returnOld` even in case `onDuplicate=update` or `onDuplicate=replace`, because we are only using the "old" attribute in the return to distinguish inserts from updates/replaces. Setting `returnOld` is overkill here however, because it will return the full previous version of the document(s), which can be large. Instead, go with the much small `_oldRev` attribute that is also set when a document is updated/replaced.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_client, shell_server, importing, dump*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13760/

### DIFF
--- a/arangod/Cluster/ClusterTrxMethods.cpp
+++ b/arangod/Cluster/ClusterTrxMethods.cpp
@@ -41,6 +41,7 @@
 #include "VocBase/LogicalCollection.h"
 
 #include <velocypack/Slice.h>
+#include <velocypack/StringRef.h>
 #include <velocypack/velocypack-aliases.h>
 
 using namespace arangodb;
@@ -125,7 +126,7 @@ void buildTransactionBody(TransactionState& state, ServerID const& server,
   builder.close();  // </openObject>
 }
 
-/// @brief lazy begin a transaction on subordinate servers
+/// @brief lazily begin a transaction on subordinate servers
 Future<network::Response> beginTransactionRequest(TransactionState& state,
                                                   ServerID const& server) {
   TransactionId tid = state.id().child();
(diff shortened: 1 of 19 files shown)


jsteemann:
Tests blue

arangod/Cluster/ClusterTrxMethods.cpp
```diff
@@ -150,46 +151,53 @@ Future<network::Response> beginTransactionRequest(TransactionState& state,
...
-  if ((resp.statusCode() == fuerte::StatusOK || resp.statusCode() == fuerte::StatusCreated) &&
+  // whatever we got can contain a success (HTTP 2xx) or an error (HTTP >= 400) 
+  if (VPackSlice answer = resp.slice(); (resp.statusCode() == fuerte::StatusOK || resp.statusCode() == fuerte::StatusCreated) &&
       answer.isObject()) {
-    VPackSlice idSlice = answer.get(std::vector<std::string>{"result", "id"});
-    VPackSlice statusSlice =
-        answer.get(std::vector<std::string>{"result", "status"});
+    //VPackSlice idSlice = answer.get(std::vector<std::string>{"result", "id"});
+    //VPackSlice statusSlice =
+    //    answer.get(std::vector<std::string>{"result", "status"});
```
mpoeter, line 166:
```suggestion
```

mpoeter (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13775/

jsteemann:
Tests blue again after merging a substantial devel change.

PR #13450, opened 2021-01-27
Adds `TAKE/DROP WHILE` and array index iteration variable to the `[*]` AQL operator

jsteemann:
### Scope & Purpose

Note: this is experimental and may be deleted. PR template is not completed as this is an RFC and thus in draft state.

This is an RFC for an experimental change that adds `PRUNE` support to the `[*]` array iterator operator in AQL. 

**To be implemented**: `DROP WHILE` iterates over the array and ignores all elements for which its condition is satisfied. Once the `DROP WHILE` condition is satisfied, it is ignored in the further iteration.

**To be implemented**: `TAKE WHILE` will stop iterating over the array as soon as its condition is *not* satisfied.

The PR also adds an `INDEX` pseudo variable to the `[*]` AQL operator. `INDEX` is a pseudo variable which contains the current loop iteration index value (zero-based). The variable is made available automatically inside `[* ...]` and is only valid inside. There already exists a pseudo variable `CURRENT` to refer to the current array element, and `INDEX` extends this.

**To be implemented**: The suggestion is to also add a `LENGTH` pseudo variable, returning the length of the array that is iterated over. If there are any DROP WHILE, TAKE WHILE, FILTER, or LIMIT clauses, these will not influence the value of LENGTH.

A few usage examples:

**Stop at first array element which is not greater eql to precending**:
```
LET a = [1, 2, 4, 16, 99, 257, 4, 10000]
RETURN a[* TAKE WHILE INDEX == 0 || a[INDEX - 1] >= CURRENT RETURN CURRENT]

[
  1,
  2,
  4,
  16,
  99,
  257
]
```

**Stop at first array element which is not double the value of the preceding**:
```
LET a = [1, 2, 4, 8, 32, 64, 256]
RETURN a[* TAKE WHILE INDEX == 0 || a[INDEX - 1] * 2 != CURRENT RETURN CURRENT]

[
  1,
  2,
  4,
  8
]
```

**Same, but return array indexes**:
```
LET a = [1, 2, 4, 8, 32, 64, 256]
RETURN a[* TAKE WHILE INDEX == 0 || a[INDEX - 1] * 2 != CURRENT RETURN INDEX]

[
  0,
  1,
  2,
  3
]
```

**Don't abort early, but instead return information that compares current array element with previous**:
```
LET a = [1, 2, 4, 8, 32, 64, 256]
RETURN a[* RETURN {value: CURRENT, isDuplicate: INDEX == 0 || a[INDEX - 1] * 2 == CURRENT}]

[
  {
    "value" : 1,
    "isDuplicate" : true
  },
  {
    "value" : 2,
    "isDuplicate" : true
  },
  {
    "value" : 4,
    "isDuplicate" : true
  },
  {
    "value" : 8,
    "isDuplicate" : true
  },
  {
    "value" : 32,
    "isDuplicate" : false
  },
  {
    "value" : 64,
    "isDuplicate" : true
  },
  {
    "value" : 256,
    "isDuplicate" : false
  }
]
```

**Return only every second array element**:
```
LET a = [1, 2, 4, 8, 16, 32, 64, 256, 512]
RETURN a[* FILTER INDEX % 2 == 1 RETURN CURRENT]

[
  2,
  8,
  32,
  256
]
```

Note: as this is implemented right now, it will only work for the `[*]` array iterator, and not for arbitrary FOR loops.

Open issues: 
- [ ] do we want this, at all?
- [ ] `PRUNE` is not a good term for aborting array iteration, and should probably be renamed.
- [ ] add tests
- [ ] add documentation

The below PR template still needs filling out once we are sure this idea will be pursued.

#### Backports:

- [x] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation

### DIFF
--- a/arangod/Aql/Ast.cpp
+++ b/arangod/Aql/Ast.cpp
@@ -1032,19 +1032,15 @@ AstNode* Ast::createNodeArrayLimit(AstNode const* offset, AstNode const* count) 
 /// @brief create an AST expansion node, with or without a filter
 AstNode* Ast::createNodeExpansion(int64_t levels, AstNode const* iterator,
                                   AstNode const* expanded, AstNode const* filter,
-                                  AstNode const* limit, AstNode const* projection) {
+                                  AstNode const* limit, AstNode const* projection,
+                                  AstNode const* prune) {
   AstNode* node = createNode(NODE_TYPE_EXPANSION);
-  node->reserve(5);
+  node->reserve(6);
   node->setIntValue(levels);
 
   node->addMember(iterator);
   node->addMember(expanded);
-
-  if (filter == nullptr) {
-    node->addMember(createNodeNop());
-  } else {
-    node->addMember(filter);
-  }
+  node->addMember(filter == nullptr ? createNodeNop() : filter);
 
   if (limit == nullptr) {
     node->addMember(createNodeNop());
@@ -1053,30 +1049,27 @@ AstNode* Ast::createNodeExpansion(int64_t levels, AstNode const* iterator,
     node->addMember(limit);
   }
 
-  if (projection == nullptr) {
-    node->addMember(createNodeNop());
-  } else {
-    node->addMember(projection);
-  }
+  node->addMember(projection == nullptr ? createNodeNop() : projection);
+  node->addMember(prune == nullptr ? createNodeNop() : prune);
 
-  TRI_ASSERT(node->numMembers() == 5);
+  TRI_ASSERT(node->numMembers() == 6);
 
   return node;
 }
 
 /// @brief create an AST iterator node
-AstNode* Ast::createNodeIterator(char const* variableName, size_t nameLength,
+AstNode* Ast::createNodeIterator(std::string const& currentVariable,
+                                 std::string const& indexVariable,
                                  AstNode const* expanded) {
-  if (variableName == nullptr) {
-    THROW_ARANGO_EXCEPTION(TRI_ERROR_OUT_OF_MEMORY);
-  }
-
   AstNode* node = createNode(NODE_TYPE_ITERATOR);
-  node->reserve(2);
+  node->reserve(3);
 
-  AstNode* variable = createNodeVariable(variableName, nameLength, false);
-  node->addMember(variable);
+  // iteration variable (CURRENT)
+  node->addMember(createNodeVariable(currentVariable.data(), currentVariable.size(), false));
   node->addMember(expanded);
+
+  // index variable (INDEX)
+  node->addMember(createNodeVariable(indexVariable.data(), indexVariable.size(), false));
 
   return node;
 }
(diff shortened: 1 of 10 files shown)


goedderz:
I like these ideas.

Some random thoughts on PRUNE:

I think the PRUNE functionality would also be very useful as a high-level operation. E.g.
```
FOR d IN docs
  FILTER d.a == @a
  SORT d.t ASC
  PRUNE d.t < 5
```
could now be expressed efficiently (imagine an index on `[a, t]`). This would also be an essential feature for K-Shortest-Path.

Especially with that picture in mind, I think `WHILE` might be a good name, if we want to allow ourselves to add a new keyword. Of course, we could also negate it and choose `UNTIL`, or add both.

I think it would be wise to choose something other than `PRUNE` in the end:

> PRUNE will stop iterating over the array as soon as its condition is not satisfied.

This is the opposite behaviour of PRUNE in traversals, which stop when the condition turns `true`. This contradiction would be irritating for users. Plus, PRUNE in Traversal still emits the element on which the condition turned true, only stopping after. We probably don't want this behaviour here, so this subtle but important difference would stay, even if we'd flip the behaviour here.

It might also feel awkward to have this available as an array operator, but not in a (sub)query. This isn't an argument against adding it, but for also adding the operation (also maybe later).

jsteemann:
> Some random thoughts on PRUNE:
> I think the PRUNE functionality would also be very useful as a high-level operation. E.g.
> 
> ```
> FOR d IN docs
>   FILTER d.a == @a
>   SORT d.t ASC
>   PRUNE d.t < 5
> ```
> 
> could now be expressed efficiently (imagine an index on `[a, t]`). This would also be an essential feature for K-Shortest-Path.

I agree, adding such functionality as a high-level operation would be useful. But obviously that makes things a lot more complicated, because we will need a new keyword, an ExecutionNode, an Executor and have all optimizer rules apply the correct logic when moving nodes around. Surely this could be done, but I wanted to avoid doing a really large change now.

> > PRUNE will stop iterating over the array as soon as its condition is not satisfied.
> 
> This is the opposite behaviour of PRUNE in traversals, which stop when the condition turns `true`. This contradiction would be irritating for users. Plus, PRUNE in Traversal still emits the element on which the condition turned true, only stopping after. We probably don't want this behaviour here, so this subtle but important difference would stay, even if we'd flip the behaviour here.

I also agree. That wasn't intentional, of course. It is obviously easy to adjust my PR w.r.t., but I am still not satisfied with using `PRUNE` here, as it is still different. I would prefer WHILE or UNTIL.

> It might also feel awkward to have this available as an array operator, but not in a (sub)query. This isn't an argument against adding it, but for also adding the operation (also maybe later).

True. The reasons for _only_ adding this as an array operator is the amount of effort it would take to implement this as a generic feature for all FOR loops.
I am still unsure if just having the functionality as an array operator is a worthwhile improvement, or if we should rather go the full way and implement it as a generic feature, also available in FOR loops - with all the efforts attached. That's why there is this RFC.

ewoutp:
A couple of comments:

# Length
With INDEX & CURRENT, what about adding LENGTH? I can imagine that you want to break 2 before the end of a list.

# PRUNE keyword
To me, this keyword does not immediately explain what it does. Something like TAKE WHILE would make it clear right away for me. Also, see the similarity to https://lodash.com/docs/4.17.15#takeWhile

# Dropping ...
If I understand it correctly, there is already an `* ...` operator. Why are the `...` being dropped in this proposal?

jsteemann:
Obsolete

PR #13453, opened 2021-01-27
Feature/test repl factor 2 adjust non default tests

mchacki:
This PR only adapts existing tests.
No User-Impact at all => No CHANGELOG.

We have modified the default replication factor used in tests to `2` in a previous PR.
Now during graph creation we tested the default replication factor and a specific replication factor of `2`.
Those are not distinguishable anymore. So with this PR we use default replication factor `2` and specific replication factor `1`.
It is asserted that the specific value is different from the default value, to ensure if defaults change this test will get red either way, and needs to be adapted.

Both test suites touched here consisted of a lot of copy-paste code, so I replaced them by a generator instead to make it simpler.

NOTE:
Before we tested replicationFactor, minReplicationFactor and numberOfShards specific values in the same suite.
This is not possible any more as minReplicationFactor by default is 1, and needs to be less then replicationFactor. So they would contradict each other, so we got one test suite more.

### DIFF
--- a/tests/js/common/http/api-gharial-creation-cluster-spec.js
+++ b/tests/js/common/http/api-gharial-creation-cluster-spec.js
@@ -52,9 +52,8 @@ describe('General graph creation', function () {
 
   const clean = () => {
     try {
-      request['delete'](`/_api/gharial/${gn}?dropCollections=true`);
-    } catch (ignore) {
-    }
+      request["delete"](`/_api/gharial/${gn}?dropCollections=true`);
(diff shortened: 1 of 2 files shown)


mchacki:
Jenkins:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13767/

dothebart (approved):

mchacki:
The modified tests are blue in Jenkins.
There are unrelated tests red.

PR #13459, opened 2021-01-28
Fix and document memory profiling

neunhoef:
This adds documentation for memory profiling to README_maintainers.md.
It also fixes memory profiling for `devel`.

This is no user facing capability.


dothebart (approved):
rather only state "procedure has been tested on ubuntu linux" since we don't have jemalloc on windows in first place? 
LGTM otherwise.

PR #13206, opened 2020-12-11
Change file descriptors threshold handling at startup

jsteemann:
### Scope & Purpose

Change file descriptors threshold handling at startup

On Linux and MacOS, require at least 8192 usable file descriptors by default to start arangod.
If less file descriptors are available to the arangod process, then the startup is automatically aborted.

Even the chosen minimum value of 8192 will often not be high enough to store considerable amounts of data. However, no higher value was chosen in order to not make too many existing installations fail after upgrading.

The required number of file descriptors can be configured using the startup option `--server.descriptors-minimum`. It defaults to 8192, but it can be increased to ensure that arangod can make use of a sufficiently high number of files. 

Setting `--server.descriptors-minimum` to a value of `0` will make the startup require only an absolute minimum limit of 1024 file descriptors, effectively disabling the change.
Such low values should only be used to bypass the file descriptors check in case of an emergency, but this is not recommended for production.

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/613

### Testing & Verification

- [x] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*

http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13785/

PR #13462, opened 2021-02-01
Remove unsafe query option `inspectSimplePlans`.

jsteemann:
### Scope & Purpose

Remove potentially unsafe AQL query option `inspectSimplePlans`.

This option previously defaulted to `true`, and turning it off could make particular queries fail.
The option was ignored in the cluster previously, and turning it off only had an effect in single server, there making very simple queries (queries not containing any FOR loops) not going through the optimizer's complete pipeline as a performance optimization. However, the optimization was only
possible for a very small number of queries and even had adverse effects, so it is now removed entirely.

The option was not listed in our documentation and was not used by arangojs/arangodb-java-driver, so there is no accompanying PR to remove it from documentation/drivers.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13808/


tests/js/client/server_invalid_options/test-too-few-file-descriptors.js
```diff
@@ -0,0 +1,32 @@
...
+///
+/// @author Jan Steemann
+/// @author Copyright 2019, ArangoDB Inc, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+return {
+  // require an enormous number of file descriptors, hopefully
+  // more than can be made available
+  'server.descriptors-minimum': '10000000000000',
+};
```
mpoeter, line 32: This file should probably be part of #13206

mpoeter (approved):

tests/js/client/server_invalid_options/test-too-few-file-descriptors.js
jsteemann: Ooops. Absolutely. Thanks for the hint!

jsteemann: Pushed a fix to remove all the unrelated JS files in this PR. 
Thanks again!

PR #13466, opened 2021-02-01
Turn off option `--server.export-read-write-metrics` for now

jsteemann:
### Scope & Purpose

Turn off option `--server.export-read-write-metrics` for now, until there is certainty about the runtime overhead it introduces. I agreed on this with @neunhoef today.

There is currently some investigation ongoing by @neunhoef to measure the overhead of metrics in general, which will likely trigger some follow-up optimizations and lead to this option being turned on again by default. But we can do this all separately.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *server_parameters*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13813/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Turn off option `--server.export-read-write-metrics` for now, until there
+  is certainty about the runtime overhead it introduces.
+
 * Remove unsafe query option `inspectSimplePlans`. This option previously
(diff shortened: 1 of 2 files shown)


jsteemann:
Tests blue

neunhoef (approved):
LGTM

PR #13469, opened 2021-02-02
remove multiple debug functions for producing backtraces.

jsteemann:
### Scope & Purpose

Remove multiple debug functions for producing backtraces.

Instead, make CrashHandler::logBacktrace() available to write a
backtrace of the current thread into the log.
Calling this function should work on Linux, because there we build with
libunwind support by default, and this can produce backtraces even for
our static builds using libmusl (in contrast to `backtrace()`, which is
a no-op there). In addition, using `CrashHandler::logBacktrace()` does
not require to compile with the `-DUSE_BACKTRACE=On` CMake option.

It is likely that the change made by this PR removes backtrace
capabilities for MacOS and Windows, but it looks like those were only
enabled previously when compiling with `-DUSE_BACKTRACE=On`, which is
unlikely.

In addition, this PR removes the `USE_BACKTRACE` CMake variable and also
the internal `ARANGODB_ENABLE_BACKTRACE` define that was enabled by this
variable.

The `USE_BACKTRACE` CMake option does not seem to have been documented. 
So this is likely an internal change only.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13816/


lib/Basics/win-utils.cpp
```diff
@@ -100,16 +100,11 @@ static void InvalidParameterHandler(const wchar_t* expression,  // expression se
...
 #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
-      << buf;
-
-  std::string bt;
-  TRI_GetBacktrace(bt);
-  LOG_TOPIC("967e6", ERR, arangodb::Logger::FIXME)
-      << "Invalid handle parameter Invoked from: " << bt
+      << "Invalid handle parameter passed: " << buf;
+#else
+      << "Invalid handle parameter passed";
```
dothebart, line 106: please log a stacktrace here.

dothebart (approved):
please add stack trace logging to invalid handle value - without its completely useless. 
Else LGTM.

jsteemann:
Tests blue

lib/Basics/win-utils.cpp
jsteemann: This code will only run on Windows, but the CrashHandler can't log stacktraces on Windows.
It is only available for Linux builds.

The whole point of this PR is to remove all other implementation of back/stacktrace functionality that we had elsewhere.

We could improve the CrashHandler so that it logs stacktraces on Windows, too, but I have no idea how to do this and test this myself.

dothebart (approved):

jsteemann:
As discussed with @mpoeter and @dothebart offline, we will not add Windows stacktrace logging, but instead we will try to leverage minidumps in case an unhandled exception flies by.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13821/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13825/

PR #13477, opened 2021-02-02
Bug fix 3.7/bts 284

Dronplane:
### Scope & Purpose
PR fixes BTS-284 issue by moving  upgrading ArangoSearchLinks task to stage of DATABASE_EXISTING from DATABASE_UPGRADE as this task needs  enabled Cluster Feature which is not available during DATABASE_UPGRADE phase.
Also introduced new 'hint' flag for upgrade tasks DATABASE_ONLY_ONCE to mark tasks which should be run only once if was not run before and for new databases this task should either be run on DATABASE_INIT (if corresponding flag is set) or should be considered executed. Previously that was implicitly declared as having only DATABASE_UPGRADE flag for task. But as new name upgrade task is also 'one-shot' but has DATABASE_EXISTING this implicit consideration became unusable.

#### Backports:

- [ ] No backports required
- [x] Backports required for:  devel

### Testing & Verification

- [x] The behaviour in this PR was *manually tested*
- [x] There are tests in an external testing repository: upgrade tests in repository release-test-automation/release_tester/upgrade.py

Link to Jenkins PR run:

https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/13828/

### DIFF
--- a/arangod/IResearch/IResearchFeature.cpp
+++ b/arangod/IResearch/IResearchFeature.cpp
@@ -618,7 +618,8 @@ void registerUpgradeTasks(arangodb::application_features::ApplicationServer& ser
     task.clusterFlags = arangodb::methods::Upgrade::Flags::CLUSTER_DB_SERVER_LOCAL  // db-server
                         | arangodb::methods::Upgrade::Flags::CLUSTER_NONE           // local server
                         | arangodb::methods::Upgrade::Flags::CLUSTER_LOCAL;
-    task.databaseFlags = arangodb::methods::Upgrade::Flags::DATABASE_UPGRADE;
+    task.databaseFlags = arangodb::methods::Upgrade::Flags::DATABASE_UPGRADE |
+                         arangodb::methods::Upgrade::Flags::DATABASE_ONLY_ONCE;
     task.action = &upgradeSingleServerArangoSearchView0_1;
     upgrade.addTask(std::move(task));
   }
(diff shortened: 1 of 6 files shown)


arangod/VocBase/Methods/Upgrade.cpp
```diff
@@ -305,8 +305,8 @@ UpgradeResult methods::Upgrade::runTasks(TRI_vocbase_t& vocbase, VersionResult&
     // check that the database occurs in the database list
     if (!(t.databaseFlags & dbFlag)) {
       // special optimization: for local server and new database,
-      // an upgrade-only task can be viewed as executed.
-      if (isLocal && dbFlag == DATABASE_INIT && t.databaseFlags == DATABASE_UPGRADE) {
+      // an one-shot task can be viewed as executed.
+      if (isLocal && dbFlag == DATABASE_INIT && (t.databaseFlags & DATABASE_ONLY_ONCE)) {
```
gnusi, line 309: where is `DATABASE_UPGRADE` handled then?

arangod/IResearch/IResearchFeature.cpp
```diff
@@ -310,27 +310,39 @@ bool upgradeArangoSearchLinkCollectionName(TRI_vocbase_t& vocbase,
...
-                                    collection->name());
-    if (clusterCollection) {
-      auto name = clusterCollection->name();
+    std::string clusterCollectionName;
+    if (!collection->shardIds()->empty()) {
+      unsigned tryCount{60};
+      do {
+        LOG_TOPIC("423b3", TRACE, arangodb::iresearch::TOPIC)
+            << " Checking collection '" << collection->name() << "' in database '" << vocbase.name() << "'";
+        clusterCollectionName = clusterInfo.getCollectionNameForShard(collection->name());
```
gnusi, line 319: let's leave a comment of why do we use `getCollectionNameForShard` instead of `getCollectionNT`

arangod/VocBase/Methods/Upgrade.cpp
Dronplane: dbFlag is set to DATABASE_UPGRADE if version check will report UPGRADE_NEEDED status see Upgrade.cpp:169 that databaseFlags for task will pass check on Upgrade.cpp:306 and task will be executed

Dronplane:
Tests blue

KVS85 (commented):
LGTM.

arangod/IResearch/IResearchFeature.cpp
Dronplane: done

gnusi (approved):

PR #13478, opened 2021-02-02
feature/move-calculations-out-of-subqueries

mpoeter:
### Scope & Purpose

Extend the move-calculations-up rule so that it can move calculations out of subqueries into the outer query.

- [x] :pizza: New feature (requires CHANGELOG entry)

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13830/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 devel
 -----
 
-* Extend the "move-calculations-up optimizer" rule so that it can move
+* Extend the "move-calculations-up" optimizer rule so that it can move
   calculations out of subqueries into the outer query.
 
 * Fix warnings about suggest-override which can break builds when warnings
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Extend the "move-calculations-up optimizer" rule so that it can move
+  calculations out of subqueries into the outer query.
+
 * Fix warnings about suggest-override which can break builds when warnings
   are treated as errors.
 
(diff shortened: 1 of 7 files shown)


CHANGELOG
```diff
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Extend the "move-calculations-up optimizer" rule so that it can move
```
```suggestion
* Extend the "move-calculations-up" optimizer rule so that it can move
```

arangod/Aql/OptimizerRules.cpp
```diff
@@ -1622,6 +1639,20 @@ void arangodb::aql::moveCalculationsUpRule(Optimizer* opt,
       auto dep = current->getFirstDependency();
 
       if (dep == nullptr) {
+        auto it = subqueries.find(current);
```
jsteemann, line 1642: I think this could introduce some unsafety _here_.
In case `dep` is a nullptr, it is still possible that `current` sets a variable that is used by the CalculationNode that we are now moving beyond it.

For example, given the following initial structure:
```
(1) SubqueryNode -> (2) SingletonNode
                    (3) CalculationNode
```
Our CalculationNode (3) would be `n` here. We load its first dependency (2) into `current`. `current` points to (2), so it is not a nullptr.
Now, we check the dependency of (2), which is a nullptr, and store that in `dep`.
`dep` is a nullptr, so we move the CalculationNode out of the subquery. However, we haven't checked yet if node (2) sets a variable that is used in the CalculationNode.

To make it more robust, I think the order of things should instead be:
```
while (current != nullptr) {
  if (current->setsVariable(neededVars)) {
    // shared variable, cannot move up any more
    // done with optimizing this calculation node
    break;
  }

  // perform other checks on current, e.g. if it is a LimitNode etc.

  auto dep = current->getFirstDependency();
  if (dep == nullptr) {
    // move out of subquery or abort the search
    // ...
  }

  // move node up
  ...  
}

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13837/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13838/

mpoeter:
Tests blue

jsteemann (approved):

PR #13480, opened 2021-02-02
don't log empty JavaScript exception

jsteemann:
### Scope & Purpose

Don't log empty JavaScript exceptions.
In case we log a JavaScript exception to the logfile, unrecoverable exceptions without a message currently log "JavaScript exception:", which looks like a bug.
For example, this happens if the arangosh is ended by a script with something similar to `process.exit(1)`.
In this case, the exception that this thrown internally is unrecoverable, and does not have any message attached. So we can as well skip logging it.

`process.exit(...)` is not available in arangod, so this should mostly affect arangosh.
I found this by accident when working on https://github.com/arangodb/cluster-maintenance/pull/44, which always printed "JavaScript exception:" whenever I made mistakes with the startup parameters for the commands I tried.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13832/

### DIFF
--- a/lib/V8/v8-utils.cpp
+++ b/lib/V8/v8-utils.cpp
@@ -5169,12 +5169,15 @@ void TRI_LogV8Exception(v8::Isolate* isolate, v8::TryCatch* tryCatch) {
   // V8 didn't provide any extra information about this error; just print the
   // exception.
   if (message.IsEmpty()) {
-    if (exceptionString == nullptr) {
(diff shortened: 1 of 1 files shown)


jsteemann:
Tests blue

fceller (approved):
LGTM

dothebart (approved):

PR #13481, opened 2021-02-02
Add `Internal` flag for AQL functions.

jsteemann:
### Scope & Purpose

Add `Internal` flag for AQL functions.

Functions flagged `Internal` cannot be invoked by end users in queries, but they can be inserted by AQL-internal optimizations and transformations. That way we can add specialized internal functions for various purposes without having to provide API stability for them.
Internal functions are (intentionally) not listed in the documentation nor are they returned by any of the AQL functions inspection APIs.

#### Backports:

- [x] No backports required

### Testing & Verification

  - [x] Added new **integration tests** (e.g. shell_server_aql)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13833/

### DIFF
--- a/arangod/Aql/AqlFunctionFeature.cpp
+++ b/arangod/Aql/AqlFunctionFeature.cpp
@@ -47,12 +47,6 @@ AqlFunctionFeature::AqlFunctionFeature(application_features::ApplicationServer& 
   startsAfter<AqlFeature>();
 }
 
-// This feature does not have any options
-void AqlFunctionFeature::collectOptions(std::shared_ptr<options::ProgramOptions>) {}
-
-// This feature does not have any options
-void AqlFunctionFeature::validateOptions(std::shared_ptr<options::ProgramOptions>) {}
-
 void AqlFunctionFeature::prepare() {
   /// @brief Add all AQL functions to the FunctionDefintions map
   addTypeCheckFunctions();
@@ -88,6 +82,10 @@ void AqlFunctionFeature::addAlias(std::string const& alias, std::string const& o
 void AqlFunctionFeature::toVelocyPack(VPackBuilder& builder) const {
   builder.openArray();
   for (auto const& it : _functionNames) {
+    if (it.second.hasFlag(FF::Internal)) {
+      // don't serialize internal functions
+      continue;
+    }
     it.second.toVelocyPack(builder);
   }
   builder.close();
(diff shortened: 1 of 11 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13834/

arangod/Aql/AqlFunctionFeature.cpp
```diff
@@ -476,6 +474,11 @@ void AqlFunctionFeature::addMiscFunctions() {
   add({"WITHIN", ".h,.,.,.|.", Function::makeFlags(FF::Cacheable), &Functions::NotImplemented});
   add({"WITHIN_RECTANGLE", "h.,.,.,.,.", Function::makeFlags(FF::Cacheable), &Functions::NotImplemented});
   add({"FULLTEXT", ".h,.,.|.", Function::makeFlags(FF::Cacheable), &Functions::NotImplemented});
+  
+  // this is an internal function that is only here for testing. it cannot
+  // be invoked by end users, because refering to internal functions from user
+  // queries will pretend these functions do not exist.
+  add({"INTERNAL", "", Function::makeFlags(FF::Internal), &Functions::NotImplemented});
```
mpoeter, line 481:
```suggestion
#ifdef ARANGODB_USE_GOOGLE_TESTS
  // this is an internal function that is only here for testing. it cannot
  // be invoked by end users, because refering to internal functions from user
  // queries will pretend these functions do not exist.
  add({"INTERNAL", "", Function::makeFlags(FF::Internal), &Functions::NotImplemented});
#endif
```

mpoeter (approved):

arangod/Aql/AqlFunctionFeature.cpp
jsteemann: This is not 100% equivalent.
The `ARANGODB_USE_GOOGLE_TESTS` is defined when the CMake variable `USE_GOOGLE_TESTS` is set. It controls whether gtest C++ tests are compiled or not.
The JavaScript tests however are always available and can be run, regardless of the value of `*_GOOGLE_TESTS`. So it is possible to run them without setting the CMake variable, and in that case, they would fail because they would refer to an unknown function.

Probably nobody will hit this problem in reality, but I would like to keep it as is. The overhead of unconditionally adding the `INTERNAL` function to the list of all functions should be negligible IMHO.

mpoeter: Ok, then let's leave it as is.

jsteemann: Thanks!

PR #13482, opened 2021-02-03
Bug fix/bts 284

Dronplane:
### Scope & Purpose

PR fixes BTS-284 issue by moving upgrading ArangoSearchLinks task to stage of DATABASE_EXISTING from DATABASE_UPGRADE as this task needs enabled Cluster Feature which is not available during DATABASE_UPGRADE phase.
Also introduced new 'hint' flag for upgrade tasks DATABASE_ONLY_ONCE to mark tasks which should be run only once if was not run before and for new databases this task should either be run on DATABASE_INIT (if corresponding flag is set) or should be considered executed. Previously that was implicitly declared as having only DATABASE_UPGRADE flag for task. But as new name upgrade task is also 'one-shot' but has DATABASE_EXISTING this implicit consideration became unusable.

#### Backports:

- [x] Backports required for: 3.7  https://github.com/arangodb/arangodb/pull/13477
#### Related Information

- [x] GitHub issue / Jira ticket number: BTS-284

### Testing & Verification

- [x] There are tests in an external testing repository: upgrade tests in repository release-test-automation/release_tester/upgrade.py

Link to Jenkins PR run:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/13839/


Dronplane:
Tests blue

PR #13484, opened 2021-02-04
Feature/strict typing for error codes

goedderz:
### Scope & Purpose

This PR introduces a type for our error codes, to avoid confusions with other `int`s, accidental casts from `bool` and so on.

It also fixes several mistakes of that kind that were found while writing this PR.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/643

#### Backports:

- [X] No backports required

### Testing & Verification

- [X] This change is a trivial rework / code cleanup without any test coverage.
- [X] This change is already covered by existing tests, such as *virtually all tests*.

Link to Jenkins PR run:

http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13843/


goedderz:
Jenkins run of commit 92cec6f967dc0fa29d388345010e57c5d25d387d is blue:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13843/
image

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -2416,7 +2418,7 @@ Result ClusterInfo::createCollectionsCoordinator(
   // closure and the main thread executing this function. Note that it can
   // happen that the callback is called only after we return from this
   // function!
-  auto dbServerResult = std::make_shared<std::atomic<int>>(-1);
+  auto dbServerResult = std::make_shared<std::atomic<std::optional<ErrorCode>>>();
```
jsteemann, line 2421: This is slightly inconsistent (though not a problem):
The similar changes a few lines above initialize the result with an `std::nullopt`, but this place doesn't. Feel free to leave it as it is.

arangod/GeneralServer/RestHandler.cpp
```diff
@@ -563,8 +563,8 @@ void RestHandler::compressResponse() {
 /// @brief generates an error
 ////////////////////////////////////////////////////////////////////////////////
 
-void RestHandler::generateError(rest::ResponseCode code, int errorNumber) {
-  char const* message = TRI_errno_string(errorNumber);
+void RestHandler::generateError(rest::ResponseCode code, ErrorCode errorNumber) {
+  const auto* message = TRI_errno_string(errorNumber);
```
jsteemann, line 567: I personally find `char const*` more informative than a generic `const auto*`, but this is obviously opinionated.
```suggestion
  char const* message = TRI_errno_string(errorNumber);
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -655,7 +655,7 @@ Result DatabaseInitialSyncer::parseCollectionDump(transaction::Methods& trx,
       } catch (velocypack::Exception const& e) {
         LOG_TOPIC("746ea", ERR, Logger::REPLICATION)
             << "while parsing collection dump: " << e.what();
-        return Result(e.errorCode(), e.what());
+        return Result(TRI_ERROR_INTERNAL, e.what());
```
jsteemann, line 658: I think we should use `TRI_ERROR_HTTP_CORRUPTED_JSON` here. That seems to be the closest error code we have, and we use it in other places when there are JSON parse errors.

arangod/RestHandler/RestShutdownHandler.cpp
```diff
@@ -81,7 +81,7 @@ RestStatus RestShutdownHandler::execute() {
     builder.add(VPackValue(true));
     AgencyCommResult result = agency.setValue("Shutdown", builder.slice(), 0.0);
     if (!result.successful()) {
-      generateError(rest::ResponseCode::SERVER_ERROR, 500);
+      generateError(rest::ResponseCode::SERVER_ERROR, TRI_ERROR_HTTP_SERVER_ERROR);
```
jsteemann, line 84: Can we turn `result` into a more useful error here, e.g. by leveraging `result.asResult()`?

arangosh/Utils/ClientManager.cpp
```diff
@@ -54,7 +54,8 @@ arangodb::Result getHttpErrorMessage(arangodb::httpclient::SimpleHttpResult* res
     std::shared_ptr<VPackBuilder> parsedBody = result->getBodyVelocyPack();
     VPackSlice const body = parsedBody->slice();
 
-    int serverCode = VelocyPackHelper::getNumericValue<int>(body, "errorNum", 0);
+    auto serverCode =
+        ErrorCode{VelocyPackHelper::getNumericValue<int>(body, "errorNum", 0)};
     std::string const& serverMessage =
         VelocyPackHelper::getStringValue(body, "errorMessage", "");
```
jsteemann, line 60: Seeing this, I think this should use `StaticStrings::ErrorNum` and `StaticStrings::ErrorMessage` instead.

lib/Basics/StringBuffer.cpp
```diff
@@ -826,11 +817,12 @@ int arangodb::basics::StringBuffer::inflate(std::stringstream& out,
...
-      case Z_DATA_ERROR:
-      case Z_MEM_ERROR: {
-        return TRI_ERROR_INTERNAL;
-      }
+        case Z_NEED_DICT:
+        case Z_DATA_ERROR:
+        case Z_MEM_ERROR: {
+          return TRI_ERROR_INTERNAL;
+        }
+        default:;
```
jsteemann, line 825: should work, but looks a bit ugly. do we need this to suppress compiler warnings?

lib/Basics/StringBuffer.cpp
```diff
@@ -904,11 +895,13 @@ int arangodb::basics::StringBuffer::inflate(arangodb::basics::StringBuffer& out,
...
-      case Z_DATA_ERROR:
-      case Z_MEM_ERROR: {
-        return TRI_ERROR_INTERNAL;
-      }
+        case Z_NEED_DICT:
+        case Z_DATA_ERROR:
+        case Z_MEM_ERROR: {
+          return TRI_ERROR_INTERNAL;
+        }
+        default:
```
jsteemann, line 903: same here, although this looks slighty different (`;` on new line here) than the previous occurrence.

lib/Basics/files.cpp
```diff
@@ -455,7 +455,7 @@ int TRI_ChMod(char const* path, long mode, std::string& err) {
   if (res != 0) {
     err = "error setting desired mode " + std::to_string(mode) + " for file " +
           path + ": " + strerror(errno);
-    return errno;
+    return TRI_set_errno(TRI_ERROR_SYS_ERROR);
```
jsteemann, line 458: This was and is potentially unsafe to do.
We are referring to errno here _after_ calling std::to_string() and strerror().
But at least strerror() may clobber the value in errno:

From `man 3 strerror`:
```
        POSIX.1-2001 and POSIX.1-2008 require that a successful  call  to  str‐
       error()  or  strerror_l()  shall  leave errno unchanged, and note that,
       since no function return value is reserved to indicate an error, an ap‐
       plication  that  wishes  to check for errors should initialize errno to
       zero before the call, and then check errno after the call.
```

So it would be better to first capture errno and store the result of `TRI_set_errno(TRI_ERROR_SYS_ERROR)` somewhere, and only then build the string and call strerror().

lib/Basics/files.cpp
```diff
@@ -1465,9 +1465,9 @@ int TRI_VerifyLockFile(char const* filename) {
   // file was not yet locked; could be locked
   if (canLock == 0) {
     // lock.l_type = F_UNLCK;
-    res = fcntl(fd, F_GETLK, &lock);
+    int res2 = fcntl(fd, F_GETLK, &lock);
 
-    if (res != TRI_ERROR_NO_ERROR) {
+    if (res2 != 0) {
```
jsteemann, line 1470: `res2` is not really needed here. Maybe make this more readable by simplifying it to
```
if (0 != fcntl(fd, F_GETLK, &lock)) {
  TRI_set_errno(TRI_ERROR_SYS_ERROR);
  ...
```

lib/Basics/files.cpp
```diff
@@ -2107,9 +2111,14 @@ int TRI_Crc32File(char const* path, uint32_t* crc) {
   TRI_Free(buffer);
 
   res2 = fclose(fin);
-  if (res2 != TRI_ERROR_NO_ERROR && res2 != EOF) {
+  // According to fclose(3):
+  // > Upon  successful  completion,  0  is returned.  Otherwise, EOF is
+  // > returned and errno is set to indicate the error.
+  // That seems to indicate the condition can never be true. I don't want to
+  // delete it in this PR, but I think it should just be removed.
+  if (res2 != 0 && res2 != EOF) {
```
jsteemann, line 2119: I agree to what you are saying here. :+1:

lib/Basics/files.cpp
```diff
@@ -2107,9 +2111,14 @@ int TRI_Crc32File(char const* path, uint32_t* crc) {
...
     if (res == TRI_ERROR_NO_ERROR) {
-      res = res2;
+      res = ErrorCode{res2};
```
jsteemann, line 2121: I think this would be wrong, even though the code will possibly be never reached.
Here we would be storing whatever fclose's return value was an translate it into one of our own error codes. 
I think instead of all this, we need to query errno if fclose does not return 0, and then
```
res = TRI_set_errno(TRI_ERROR_SYS_ERROR);
```

lib/Basics/files.cpp
```diff
@@ -2537,7 +2546,7 @@ int TRI_CreateDatafile(std::string const& filename, size_t maximalSize) {
   }
 
   // no fallocate present, or at least pretend it's not there...
-  int res = TRI_ERROR_NOT_IMPLEMENTED;
+  int res = 1;
```
jsteemann, line 2549: while this line looks ok, a few lines further down (after the #ifdef) we are comparing the numeric `res` variable against an ErrorCode (TRI_ERROR_NO_ERROR).
Shouldn't that line be `if (res != 0) {` instead?

lib/Basics/tri-zip.cpp
```diff
@@ -373,8 +374,8 @@ int TRI_ZipFile(char const* filename, char const* dir,
       }
 
       if (sizeRead > 0) {
-        res = zipWriteInFileInZip(zf, buffer, sizeRead);
-        if (res != 0) {
+        int res2 = zipWriteInFileInZip(zf, buffer, sizeRead);
+        if (res2 != 0) {
```
jsteemann, line 378:
```suggestion
        if (0 != zipWriteInFileInZip(zf, buffer, sizeRead)) {
```

lib/V8/v8-utils.cpp
```diff
@@ -5316,10 +5318,10 @@ v8::Handle<v8::Value> TRI_ExecuteJavaScriptString(v8::Isolate* isolate,
...
 ////////////////////////////////////////////////////////////////////////////////
 
-void TRI_CreateErrorObject(v8::Isolate* isolate, int errorNumber) {
-  v8::HandleScope scope(isolate);
-  CreateErrorObject(isolate, errorNumber, TRI_errno_string(errorNumber));
-}
+// void TRI_CreateErrorObject(v8::Isolate* isolate, ErrorCode errorNumber) {
+//   v8::HandleScope scope(isolate);
+//   CreateErrorObject(isolate, errorNumber, TRI_errno_string(errorNumber));
+// }
```
jsteemann, line 5324: If this is not used anymore, let's remove it!
```suggestion
```

utils/generateErrorfile.py
```diff
@@ -64,15 +64,15 @@ def genCFile(errors, filename):
...
-         + "\n"\
-         + "void TRI_InitializeErrorMessages() {\n"
+  impl = prologue + """
+#include "Basics/error.h"
+#include "Basics/voc-errors.h"
+
+/// helper macro to define an error string
+#define REG_ERROR(id, label) TRI_set_errno_string(TRI_ ## id, label);
+
+void TRI_InitializeErrorMessages() {
```
jsteemann, line 74: Suggestion: instead of having a separate initialization function and registering each error message at program start, can we switch to a compile-time map of error codes to error messages?
`frozen::map` should be a good fit, I hope.

goedderz: Love that idea, will try it out.

arangod/Cluster/ClusterInfo.cpp
goedderz: Thanks, fixed.

arangod/Replication/DatabaseInitialSyncer.cpp
goedderz: That's better, changed it here and above.

lib/Basics/StringBuffer.cpp
goedderz: Not compiler, but a clang-tidy warning `hicpp-multiway-paths-covered`.

As to the uglyness, would you prefer
```suggestion
        default: {
        }
```
?

lib/Basics/files.cpp
goedderz: Good catch, thanks. Fixed.

lib/Basics/files.cpp
goedderz: Done.

lib/Basics/files.cpp
```diff
@@ -2092,7 +2092,11 @@ int TRI_Crc32File(char const* path, uint32_t* crc) {
 
     if (sizeRead < bufferSize) {
       if (feof(fin) == 0) {
-        res = errno;
+        if (errno != 0) {
+          res = TRI_set_errno(TRI_ERROR_SYS_ERROR);
+        } else {
+          res = TRI_ERROR_NO_ERROR;
+        }
```
goedderz, line 2099: @jsteemann fyi, I'm gonna change this again because it seems to be nonsense: `feof` doesn't set `errno` at all, and no system or library function sets `errno` to `0`.
```suggestion
        res = TRI_ERROR_INTERNAL;
```
Do you think we need a more specific error code? It would probably have to be introduced, or we can change the return code to something else (e.g. `ResultT<bool>` or so).

lib/Basics/files.cpp
goedderz: Right. I'll change it here, but there are most probably more places like that. The compiler will catch these as soon as I remove the implicit cast `operator int()` in `ErrorCode`.

lib/Basics/StringBuffer.cpp
jsteemann: It is not important for me. It just looked a bit strange to me. I am fine with any outcome.

lib/Basics/files.cpp
jsteemann: `TRI_ERROR_FAILED` is fine for me too.

utils/generateErrorfile.py
goedderz: https://github.com/arangodb/arangodb/pull/13496

lib/Basics/files.cpp
goedderz: fixed

arangod/RestHandler/RestShutdownHandler.cpp
goedderz: Let's do this in a different PR.

arangosh/Utils/ClientManager.cpp
goedderz: Let's do this in a different PR.

goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13873/

goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13893/

goedderz:
http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13895/

goedderz:
image

PR #13489, opened 2021-02-04
add timeout if the server seems to be alive but we can't reach it after er 5 minutes bail out.

dothebart:
### Scope & Purpose

this fixes an incident where the server comes up, but we can't reach it because of other reasons.


goedderz (approved):

dothebart:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/13857/

PR #13491, opened 2021-02-04
don't use an exclusive transaction on _fishbowl collection

jsteemann:
### Scope & Purpose

Don't use an exclusive transaction on _fishbowl collection every time the Foxx store results are written to that collection. 
Instead perform two separate operations: insert-replace and remove (the latter only if necessary).

The overall operation will not be atomic though, but it is more efficient than a truncate and then re-inserting everything. Moving away from exclusive locks is good for other reasons, too.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13862/


jsteemann:
Tests blue

dothebart (approved):

mchacki (approved):
LGTM

PR #13493, opened 2021-02-05
improve shard synchronization visibility

jsteemann:
### Scope & Purpose

Improve shard synchronization visibility in the web UI.
The web UI can now display shard synchronization progress in a more intuitive way.
It will now be displayed how many followers are actively synchronizing a shard, and what their progress is. Progress reporting is also improved when there is more than one follower for a shard.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13865/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14254/

arangod/Cluster/ActionBase.h
```diff
@@ -30,7 +30,10 @@
 #include "Basics/Common.h"
 #include "Basics/Result.h"
 
+#include <atomic>
```
hkernbach, line 33: Are those includes needed here?

js/apps/system/_admin/aardvark/APP/frontend/js/views/shardsView.js
```diff
@@ -62,25 +62,53 @@
...
-          if (value.progress.current === 0) {
-            var spin = '<span>n/A</span>';
-            $('#' + collection + '-' + shard + ' .shardProgress').html(spin);
-          } else {
-            percent = (value.progress.current / value.progress.total * 100).toString().match(/^-?\d+(?:\.\d{0,2})?/)[0] + '%';
-            $('#' + collection + '-' + shard + ' .shardProgress').html(percent);
+          if (value.progress.hasOwnProperty('followersSyncing') && 
+              value.progress.followersSyncing > 0) {
+            // number of followers currently running the synchronization for the shard
+            followersSyncing = '<span>' + value.progress.followersSyncing + ' follower';
```
hkernbach, line 87:
```suggestion
            followersSyncing = '<span>' + arangoHelper.escapeHtml(value.progress.followersSyncing) + ' follower';
```

js/apps/system/_admin/aardvark/APP/frontend/js/views/shardsView.js
```diff
@@ -62,25 +62,53 @@
...
+              typeof value.progress.followerPercent === 'number') {
+            shardProgress = percentify(value.progress.followerPercent);
+          } else if (value.progress.current !== 0) {
+            shardProgress = percentify(value.progress.current / value.progress.total * 100);
+          }
+          if (shardProgress === '' || followersSyncing === '') {
+            shardProgress = 'waiting for slot...';
+          }
+
+          shardProgress = '<span>' + shardProgress + '</span>';
```
hkernbach, line 106:
```suggestion
          shardProgress = '<span>' + arangoHelper.escapeHtml(shardProgress) + '</span>';
```

tests/Sharding/ShardDistributionReporterTest.cpp
```diff
@@ -93,6 +94,14 @@ static void VerifyNumbers(VPackSlice result, std::string const& colName,
   VPackSlice current = progress.get("current");
   ASSERT_TRUE(current.isNumber());
   ASSERT_EQ(current.getNumber<uint64_t>(), testCurrent);
+
+  VPackSlice pct = progress.get("followerPercent");
+  if (testFollowerPercent >= 0.0) {
```
hkernbach, line 99: Will this if be removed by the compiler in non-maintainer mode?

hkernbach: Ah nevermind, just test code.

hkernbach (changes requested):
Just a few minor things.

arangod/Cluster/ActionBase.h
hkernbach: (also the other ones below which have been added)

jsteemann: These includes may not be 100% necessary here (in fact, they were not present before). But this is only because previous file includes will have incldued `atomic`, `memory` etc., so we relied on other includes here to include the right headers for us.
IMHO this is not how it should be. We should follow the include-what-you-use rule, and Action.h base uses `std::shared_ptr` (memory) and `std::atomic` (atomic), std::chrono (chrono) etc. 
So yes, we should include these headers here!

hkernbach: Ty!

hkernbach (approved):

Issue #13476, opened 2021-02-02
Feature request: some optimization on pagination


razbenya:
## My Environment

* __ArangoDB Version__:         3.7.6

## Component, Query & Data

__AQL query (if applicable)__:
```
for record in v_records
    SEARCH ANALYZER(record.queries in @queries , "text_en")
    SORT record.score desc
    SORT @start,1000
    return record._key
```
__AQL explain and/or profile (if applicable)__:

profile for first page 
```
Query String (149 chars, cacheable: false):
 for record in v_records
     SEARCH ANALYZER(record.queries in @queries, "text_en")
     sort record.score desc
     limit 0,1000
     return record._key

Execution plan:
 Id   NodeType            Calls   Items   Runtime [s]   Comment
  1   SingletonNode           1       1       0.00001   * ROOT
  2   EnumerateViewNode       1    1000       0.00528     - FOR record IN v_records SEARCH ANALYZER((record.`queries` == "a"), "text_en") SORT record.`score` DESC   /* view query */
  5   LimitNode               1    1000       0.00002       - LIMIT 0, 1000
  6   CalculationNode         1    1000       0.00028       - LET #3 = record.`_key`   /* attribute expression */
  7   ReturnNode              1    1000       0.00001       - RETURN #3

Indexes used:
 none

Optimization rules applied:
 Id   RuleName
  1   handle-arangosearch-views
  2   remove-unnecessary-calculations-2

Query Statistics:
 Writes Exec   Writes Ign   Scan Full   Scan Index   Filtered   Exec Time [s]
           0            0           0         1000          0         0.00616

Query Profile:
 Query Stage           Duration [s]
 initializing               0.00000
 parsing                    0.00014
 optimizing ast             0.00001
 loading collections        0.00002
 instantiating plan         0.00004
 optimizing plan            0.00024
 executing                  0.00562
 finalizing                 0.00009
```
profile for page with 9m skip
```
Query String (155 chars, cacheable: false):
 for record in v_records
     SEARCH ANALYZER(record.queries in @queries, "text_en")
     sort record.score desc
     limit 9000000,1000
     return record._key

Execution plan:
 Id   NodeType            Calls     Items   Runtime [s]   Comment
  1   SingletonNode           1         1       0.00001   * ROOT
  2   EnumerateViewNode       1   9001000       0.44689     - FOR record IN v_records SEARCH ANALYZER((record.`queries` == "a"), "text_en") SORT record.`score` DESC   /* view query */
[...]

PR #13497, opened 2021-02-05
Docs: Correct description of streaming cursor

Simran-B:
#13476

>> the documentation recommended not using the stream options for long queries.
>
> That is an outdated comment and was targeted at the MMFiles storage engine, which is no longer supported. Unless you're using exclusive locks or so, you can safely ignore it and use streaming queries also for long running queries.

- Remove MMFiles leftovers
- As far as I understand, one should still not use streaming cursors with the RocksDB engine if a query sets `OPTIONS { exclusive: true }` for a modification query, because it will block all other queries involving the same collection. Are there additional cases?
- > [when stream is set to false, then] query results are either returned right away (if the result set is small enough), ...

  Does that mean the query result is calculated on-the-fly as long as it fits within one cursor batch?
- The description leaves open how this works in practice:
  > query statistics, warnings and profiling data will only be available after the query is finished.

  Is this information contained in the last batch of results before the cursor is exhausted?

Needs to be backported to 3.7. Should also adjust 3.6 docs and distinguish between MMFiles and RocksDB.

Related:
- https://arangodb.atlassian.net/browse/AR-21
- https://arangodb.atlassian.net/browse/AR-22


goedderz:
As a synopsis, the main difference is as follows:

A query with `stream: false` will be immediately entirely executed on the server, and its result stored in memory (on the contacted coordinator, if in a cluster). All other resources will immediately be freed (like locks or RocksDB snapshots).

A query with `stream: true` will suspend execution of the query and return a batch of results as soon as `batchSize` many results were calculated. Depending on the query, this can mean that the first result will be available much faster, and that we don't need to store the entirety of the result in memory at any point in time. Though certain queries won't benefit from either; for example, a query that `SORT`s all its results (without using an index) and has no `LIMIT` will still collect everything in memory in order to sort it, before returning it peu à peu. But on the other hands, the query is not ended but just suspended, and still holds resources. For example, a RocksDB snapshot, which will prevent compactions to some degree. Writes to RocksDB will also be in memory until the query is committed. And writes will make RocksDB hold write locks on those documents, which will add some overhead to other (write) transactions and make other queries that try to write the same documents fail with conflicts until it has ended.

One other consequence of this behavior is that if a query fails, e.g. with a write conflict, a query with `stream: false` will report that error before returning any results. A streaming query may return some results up to the point where it encounters a conflict, which may render the already returned results meaningless as the transaction was aborted.

> As far as I understand, one should still not use streaming cursors with the RocksDB engine if a query sets OPTIONS { exclusive: true } for a modification query, because it will block all other queries involving the same collection. Are there additional cases?

So the question is, when should we recommend or discourage the use of streaming cursors.

For queries returning at most a `batchSize` of results, it doesn't make a difference.

Especially read-only queries can benefit greatly from streaming cursors, unless they have to gather everything in memory anyways, e.g. due to a SORT, or COLLECT. And the cost for streaming cursors is low in that case.

For queries that write lots of documents *and* produce a lot of results, it may not be advisable to use a streaming cursor, though it may also have advantages.

> > [when stream is set to false, then] query results are either returned right away (if the result set is small enough), ...
>
> Does that mean the query result is calculated on-the-fly as long as it fits within one cursor batch?

That is actually true in both cases. The difference is:
* with `stream: false`, the result is completely calculated and the query ended before the first request is answered. If the result fits inside one batch, everything else is also cleaned up immediately. If the result doesn't fit in one batch, it is stored in memory attached to the cursor and can be fetched later, but the actual query is already finished and cleaned up.
* with `stream: true`, the query is executed only as long as necessary to produce the first batch of results, and no more. If that means the query is exhausted, it and everything else is also cleaned up immediately. If it isn't, its execution is suspended until the next request arrives.

> The description leaves open how this works in practice:
>
> >  query statistics, warnings and profiling data will only be available after the query is finished.
>
> Is this information contained in the last batch of results before the cursor is exhausted?

Yes, the final response, i.e. the one returning `hasMore: false`, contains this data.

goedderz:
@mchacki Please have a look at my response, whether I've layed out everything correctly.

Simran-B:
Thank you @goedderz!

> with `stream: true`, the query is executed only as long as necessary to produce the first batch of results, and no more. If that means the query is exhausted, it and everything else is also cleaned up immediately. If it isn't, its execution is suspended until the next request arrives.

Until the next batch is ready, or until the client requests the next batch from the cursor?

Is "first batch" defined as the first `batchSize` (1000) elements/documents?

goedderz:
> Until the next batch is ready, or until the client requests the next batch from the cursor?

Until the client requests the next batch. Currently, there is no asynchronous execution while arangod waits for the next client request. Though that would be sleek.

> Is "first batch" defined as the first batchSize (1000) elements/documents?

Yes. To clarify that:

When the request to create the (streaming) cursor arrives, the query is executed until it produced exactly `batchSize` many results, and/or the query is done. These are returned. If the query is done, that's it - no persistent cursor is created, and everything is cleaned up immediately. If the query can produce more results, its execution is suspended, and the `batchSize` many results we produced are returned, together with a cursor id to continue execution and retrieve more results later.

jsteemann:
>     * > [when stream is set to false, then] query results are either returned right away (if the result set is small enough), ...
>       
>       
>       Does that mean the query result is calculated on-the-fly as long as it fits within one cursor batch?
> 
>     * The description leaves open how this works in practice:
>       > query statistics, warnings and profiling data will only be available after the query is finished.
>      
>       
>       Is this information contained in the last batch of results before the cursor is exhausted?

`stream: false` (which is the default) means that the entire query is executed on the server, and the results will be built up in memory. The query responds with its first batch once the entire query is finished and results have been stored in memory.
Everything will be returned in a single batch if the number of query results is less equal to the batch size (default: 1000). If there are more results, a cursor is automatically created on the server, and the client application can fetch more data by sending follow-up requests to the cursor API using the cursor's id (using the "id" attribute of the response).

Query statistics and warnings are available with the first response batch already, as the query was completely executed once the first batch becomes available.

Example:
```
curl -X POST http://127.0.0.1:8529/_api/cursor --data '{"query":"FOR i IN 9..0 RETURN 1 / i","batchSize":5,"profile":true}' --dump - --user "root:"; echo
HTTP/1.1 201 Created
X-Content-Type-Options: nosniff
Server: ArangoDB
Connection: Keep-Alive
Content-Type: application/json; charset=utf-8
Content-Length: 404

{"result":[0.1111111111111111,0.125,0.14285714285714285,0.16666666666666666,0.2],"hasMore":true,"id":"27169809","extra":{"stats":{"writesExecuted":0,"writesIgnored":0,"scannedFull":0,"scannedIndex":0,"filtered":0,"httpRequests":0,"executionTime":2.593730023363605e-4,"peakMemoryUsage":38912},"warnings":[{"code":1562,"message":"in operator /: division by zero"}]},"cached":false,"error":false,"code":201}
```

`stream: true` means that the query is started on the server and can return its first batch as soon as it is available. Once there is data to fill a response batch, the server will respond it, and the query will be paused on the server until the next follow-up request comes in to fetch more data. This can go on as long as the query returns more data.
When it comes to query statistics etc., these are not known until the query is fully finished. With a streaming query this means that the statistics are returned with the batch that finishes the query. It can be the first batch if the query does not return more than this fits in a batch, or any of the follow-up batches. In any case, statistics will contained be the _last_ batch in a streaming query.

Now an example for a streaming query:
```
 curl -X POST http://127.0.0.1:8529/_api/cursor --data '{"query":"FOR i IN 9..0 RETURN 1 / i","batchSize":5,"profile":true,"stream":true}' --dump - --user "root:"; echo
HTTP/1.1 201 Created
X-Content-Type-Options: nosniff
Server: ArangoDB
Connection: Keep-Alive
Content-Type: application/json; charset=utf-8
Content-Length: 152

{"result":[0.1111111111111111,0.125,0.14285714285714285,0.16666666666666666,0.2],"hasMore":true,"id":"27169890","cached":false,"error":false,"code":201}

```
No statistics nor warnings available in the first batch. However, we got an "id" here, so we can query more data from the cursor:
```
curl -X PUT http://127.0.0.1:8529/_api/cursor/27169890 --dump - --user "root:"; echo
HTTP/1.1 200 OK
X-Content-Type-Options: nosniff
Server: ArangoDB
Connection: Keep-Alive
Content-Type: application/json; charset=utf-8
Content-Length: 353

{"result":[0.25,0.3333333333333333,0.5,1,null],"hasMore":false,"cached":false,"extra":{"stats":{"writesExecuted":0,"writesIgnored":0,"scannedFull":0,"scannedIndex":0,"filtered":0,"httpRequests":0,"executionTime":19.554455741999845,"peakMemoryUsage":39072},"warnings":[{"code":1562,"message":"in operator /: division by zero"}]},"error":false,"code":200}
```
Now the query is done ("hasMore" is false), so the statistics and warnings are returned.

@Simran-B

goedderz (approved):
LGTM

PR #13500, opened 2021-02-05
improvements for logging

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13492/

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/645
Docs PR: https://github.com/arangodb/docs/pull/616

Improvements for logging. This adds the following startup options to arangod:

- `--log.max-entry-length`: controls the maximum line length for individual log messages that are written into normal logfiles by arangod (note: this does not include audit log messages). 
  Any log messages longer than the specified value will be truncated and the suffix '...' will be added to them. The purpose of this parameter is to shorten long log messages in case there is not a lot of space for logfiles, and to keep rogue log messages from overusing resources. 
  The default value is 128 MB, which is very high and should effectively mean downwards-compatiblity with previous arangod versions, which did not restrict the maximum size of log messages.

- `--audit.max-entry-length`: controls the maximum line length for individual audit log messages that are written into audit logs by arangod. Any audit log messages longer than the specified value will be truncated and the suffix '...' will be added to them. 
  The default value is 128 MB, which is very high and should effectively mean downwards-compatiblity with previous arangod versions, which did not restrict the maximum size of log messages.

- `--log.in-memory-level`: controls which log messages are preserved in memory (in case `--log.in-memory` is set to `true`). The default value is `info`, meaning all log messages of types `info`, `warning`, `error` and `fatal` will be stored by an instance in memory (this was also the behavior in previous versions of ArangoDB). By setting this option to `warning`, only `warning`, `error` and `fatal` log messages will be preserved in memory, and by setting the option to `error` only error and fatal messages will be kept.
  This option is useful because the number of in-memory log messages is limited to the latest 2048 messages, and these slots are by default shared between informational, warning and error messages. 

Honor the value of startup option `--log.api-enabled` when set to `false`. The desired behavior in this case is to turn off the REST API for logging, but was not implemented. The default value for the option is `true`, so the REST API is enabled. This behavior did not change, and neither did the behavior when setting the option to a value of `jwt` (meaning the REST API for logging is only available for superusers with a valid JWT token).

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/616
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/645

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13874/


lib/Logger/Logger.h
```diff
@@ -97,12 +97,19 @@ struct LogMessage {
         _message(std::move(message)),
         _offset(offset) {}
 
+  void shrink(std::size_t maxLength) {
+    if (_message.size() > maxLength) {
+      _message.resize(maxLength);
+    }
+    _message.append("...", 3);
```
jsteemann, line 104:
```suggestion
      _message.resize(maxLength);
      _message.append("...", 3);
    }
```

joerg84 (approved):

CHANGELOG
```diff
@@ -1,6 +1,45 @@
...
+  
+  - `--audit.max-entry-length`: controls the maximum line length for individual
+    audit log messages that are written into audit logs by arangod. Any audit
+    log messages longer than the specified value will be truncated and the 
+    suffix '...' will be added to them. 
+    The default value is 128 MB, which is very high and should effectively
+    mean downwards-compatiblity with previous arangod versions, which did not
+    restrict the maximum size of log messages.
+
+  - `--log.in-memory-level`: controls which log messages are preserved in
```
joerg84, line 25: Just a thought:
Imo we have different semantics for level 'info' (implying all levels below) and 'error', 'warning' (implying exactly that level).
Would it make sense to rename 'info' -> 'all'?

jsteemann: Thanks, I fixed the description of the PR.
`info` is not equivalent to `all`, as there are also `debug` and `trace`, which are not mentioned in the description because they don't make so much sense here (the size of the in-memory buffer is limited, and when setting log level to debug or trace, it will very quickly be exceeded).

jsteemann:
Tests blue

PR #13502, opened 2021-02-06
Compressed index selectivity estimates

jsteemann:
### Scope & Purpose

Compressed index selectivity estimates

Decrease the size of serialized index esimtates, by introducing a compressed serialization format. The compressed format uses the previous uncompressed format internally, compresses it, and stores the compressed data instead. This makes serialized index estimates a lot smaller, which in turn decreases the size of I/O operations for index maintenance.

#### Backports:

- [x] No backports required

### Testing & Verification

  - [x] Added new C++ **Unit tests**: tests/RocksDBEngine/IndexEstimatorTest.cpp
  - [x] Added new **integration tests** recovery/index-estimator.js

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13876/


jsteemann:
Requesting review from 
- @mchacki as the author of the original index estimator format
- @cw00dw0rd as the one who brought this to our attention
- @mpoeter as being awesome

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13879/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13883/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13886/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13888/

jsteemann:
Tests blue

tests/RocksDBEngine/IndexEstimatorTest.cpp
```diff
@@ -333,12 +333,84 @@ TEST_F(IndexEstimatorTest, test_truncate_logic_2) {
...
+  }
+
+  { 
+    // compressed
+    auto est = buildEstimator();
+
+    std::string serialization;
+    est->serialize(serialization, seq, true);
+    ASSERT_EQ(10056, serialization.size());
+    ASSERT_EQ('2', serialization[sizeof(uint64_t)]);
```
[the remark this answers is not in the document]
jsteemann: Should be addressed now, although slightly differently.

tests/RocksDBEngine/IndexEstimatorTest.cpp
```diff
@@ -333,12 +333,84 @@ TEST_F(IndexEstimatorTest, test_truncate_logic_2) {
...
+  };
+
+  { 
+    // uncompressed
+    auto est = buildEstimator();
+
+    std::string serialization;
+    est->serialize(serialization, seq, false);
+    ASSERT_EQ(24641, serialization.size());
+    ASSERT_EQ('1', serialization[sizeof(uint64_t)]);
```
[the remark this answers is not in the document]
jsteemann: Should be addressed now, although slightly differently.

arangod/RocksDBEngine/RocksDBCuckooIndexEstimator.h
```diff
@@ -250,212 +204,26 @@ class RocksDBCuckooIndexEstimator {
...
-                                !_truncateBuffer.empty();
-      _needToPersist.store(havePendingUpdates, std::memory_order_release);
-    }
-
-    _appliedSeq.store(appliedSeq, std::memory_order_release);
-  }
-
+  void serialize(std::string& serialized, 
+                 rocksdb::SequenceNumber maxCommitSeq, 
+                 bool compress);
```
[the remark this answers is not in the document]
jsteemann: Done

mpoeter (approved):

tests/RocksDBEngine/IndexEstimatorTest.cpp
```diff
@@ -333,12 +344,86 @@ TEST_F(IndexEstimatorTest, test_truncate_logic_2) {
...
+    // Still all values have to be identical
+
+    EXPECT_EQ(est.nrUsed(), copy.nrUsed());
+    EXPECT_EQ(est.nrCuckood(), copy.nrCuckood());
+    EXPECT_EQ(est.computeEstimate(), copy.computeEstimate());
+  };
+
+  { 
+    // uncompressed
+    auto est = buildEstimator();
```
mchacki, line 405: I like this test :+1:

i would have written it a way that I only create the `est` once outside, and do two calls on serialize with different format, then deserialize both, and do a `three-way` equality comparison of each.
To emphasize that they behave exactly the same way, no matter how they are compressed.
This way your estimates slightly differ on the sequence, and in the pseudo random choice you take on inserting equally hashed keys.

However I think the implication of equality on COMPRESSED and UNCOMPRESSED are identical here. So I do not think it is required to add an additional test.

tests/RocksDBEngine/IndexEstimatorTest.cpp
```diff
@@ -333,12 +344,86 @@ TEST_F(IndexEstimatorTest, test_truncate_logic_2) {
...
+
+    std::string serialization;
+    auto format = RocksDBCuckooIndexEstimatorType::SerializeFormat::UNCOMPRESSED;
+    est->serialize(serialization, seq, format);
+    ASSERT_EQ(24641, serialization.size());
```
mchacki, line 410: There is some pseudo-random component in the Cuckoo algorithm.
Does this guarantee the size here? Or is there a small chance that we end up with differently filled buckets?

My guess on the UNCOMPRESSED format is that it actually is guaranteed. as we are storing deterministically many items, and the ordering of UNCOMPRESSED does not matter.

tests/RocksDBEngine/IndexEstimatorTest.cpp
```diff
@@ -333,12 +344,86 @@ TEST_F(IndexEstimatorTest, test_truncate_logic_2) {
...
+  }
+
+  { 
+    // compressed
+    auto est = buildEstimator();
+
+    std::string serialization;
+    auto format = RocksDBCuckooIndexEstimatorType::SerializeFormat::COMPRESSED;
+    est->serialize(serialization, seq, format);
+    ASSERT_EQ(10056, serialization.size());
```
mchacki, line 423: This sizing is actually more relevant.
How "Stable" is the Snappy compression here? Does it guarantee identical sizes even if the ordering of chars in the Compressed string may be modified?
Or is the pseudo-randomness under enough control that the non-compressed string between two runs is guaranteed to be identical?
This will be a randomly red test if the size differs just because the ordering of execution in unrelated tests has changed. (end thereby the random generator seed).

mchacki: Okay i have checked, the seed is constant and defined in the class alone.
=> This test will be deterministic this way, unless we change the seed, which is fine. :+1:

mchacki (approved):
LGTM.

I have added a point in the tests, which is good for now, but only works as long as nothing changes on the "random" factor of the cuckoo estimate.
As there is no intention to change randomness there, I am good with this PR :+1:

PR #13506, opened 2021-02-07
Do not create index estimator objects for proxy collection objects

jsteemann:
### Scope & Purpose

Do no create index estimator objects for proxy collection objects on coordinators and DB servers. Proxy objects are created on coordinators and DB servers for all shards, and they also make index objects available. In order to reduce the memory usage by these objects, we don't create any index estimator objects for indexes in those proxy objects. Index estimators usually take several KB of memory each, so not creating them will pay out for higher numbers of collections/shards.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13890/


CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* Do not create index estimator objects for proxy collection object on 
+  coordinators and DB servers. Proxy objects are created on coordinator and
```
```suggestion
* Do not create index estimator objects for proxy collection objects on 
  coordinators and DB servers. Proxy objects are created on coordinators and
```

jsteemann:
Tests blue

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -890,6 +892,9 @@ void RocksDBEdgeIndex::handleValNode(VPackBuilder* keys,
 
 void RocksDBEdgeIndex::afterTruncate(TRI_voc_tick_t tick,
                                      arangodb::transaction::Methods* trx) {
+  if (unique()) {
```
mchacki, line 895: I think the edge Index can never be unique, so this code in fact is unreachable.
But it would be correct if reachable :+1:

mchacki (approved):
LGTM.

SideNote:
As soon as we have better Monitoring of Ressource usage on those structs we can add tests around this.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: I agree. Will put a code comment there after merging!

PR #13507, opened 2021-02-08
more tests for replication consistency

jsteemann:
### Scope & Purpose

Fixes replication bug which leads to data corruption. In the Merkle sync protocol there is the following problem: When there are large amounts of differences between source and destination, then the revisions in the source are shipped in batches to the destination. In that case, it is possible that a smaller (=earlier) revision is inserted in the destination before a larger (=later) revision with the same primary key is removed. As such this would not be a problem and is ultimately inevitable. However, some performance optimizations lead to the fact that this condition is not detected and so the data structures are corrupted. Specifically, the operations flag `ignoreUniqueConstraint` so far prevented any unique conflict detection in the primary index or any other unique constraint index. Furthermore, even if the conflict would be detected, another optimization (`DisableIndexing`) prevented the transaction from seeing its own writes, so the delete of the wrong revision was not detected and would lead to another unique constraint violation. 

Finally, the case of multiple unique constraints being violated in this case was not considered. In the process of hunting down these problems some more minor problems (like a buffer overrun in `ridBuffer`s), a bug in RocksDB which leads to an assertion failure and a bug in document count accounting which can lead to an assertion failure were found and are fixed in this PR.

Adds more tests for incremental sync.
Additional tests can be started as follows:
```
scripts/unittest replication_sync --test malarkey
```

#### Backports:

- [x] Backports required for: 3.7 (parts of this will be backported in separate PRs, PR #1: https://github.com/arangodb/arangodb/pull/13520)

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in replication_sync)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13892/


neunhoef (approved):
Overall: LGTM, see three minor comments.

neunhoef:
One thing is still on my list: There is a possible assertion failure in maintainer mode if a thread tries to add the first document of a collection at the same time and another thread tries to remove it immediately. In this case the remove can "overtake" the insert and move the document count below zero for a brief moment, enough to trigger the assertion. Do we want to address this in this PR, too?

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13907/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13908/

neunhoef (approved):
LGTM

jsteemann:
Tests blue

PR #13509, opened 2021-02-08
Feature/distribute simplification

mpoeter:
### Scope & Purpose

Enterprise companion PR: arangodb/enterprise#647

Simplify the DistributeExecutor and avoid implicit modification of its input variable.

Previously the DistributeExecutor sometimes updated the input variable in-place, leading to unexpected results like in this example:
```
FOR i IN 1..3 LET x = { test: 1 } INSERT x INTO testi RETURN x

Execution plan:
 Id   NodeType            Site  Est.   Comment
  1   SingletonNode       COOR     1   * ROOT
  2   CalculationNode     COOR     1     - LET #3 = 1 .. 3   /* range */   /* simple expression */
  4   CalculationNode     COOR     1     - LET x = { "test" : 1 }   /* json expression */   /* const assignment */
  3   EnumerateListNode   COOR     3     - FOR i IN #3   /* list iteration */
  9   DistributeNode      COOR     3       - DISTRIBUTE  /* create keys: true, variable: x */
 10   RemoteNode          DBS      3       - REMOTE
  6   InsertNode          DBS      3       - INSERT x IN testi 
 11   RemoteNode          COOR     3       - REMOTE
 12   GatherNode          COOR     3       - GATHER   /* unsorted */
  8   ReturnNode          COOR     3       - RETURN x
```
This query should return `{test: 1}` three times. Instead it returns:
```
[ 
  { "test" : 1, "_key" : "2010078" }, 
  { "test" : 1, "_key" : "2010079" }, 
  { "test" : 1, "_key" : "2010080" } 
]
```

This PR moves the modification logic from the DistributeExecutor into three new _internal_ AQL functions (`MAKE_DISTRIBUTE_INPUT`, `MAKE_DISTRIBUTE_INPUT_WITH_KEY_CREATION`, `MAKE_DISTRIBUTE_GRAPH_INPUT`). As a post-processing step _after_ the optimization, we insert a new calculation node with the corresponding function call for each distribute node in the plan (if necessary). This is done in a post-processing step so that the calculation node does not interfere with any optimization rules operating on the distribute node and its variable,

This change not only simplifies the DistributeExecutor, but also makes any additional calculation (if necessary) _explicit_ and avoids unexpected results like in the previous example.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Main repository PR: arangodb/enterprise#647

### Testing & Verification


mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13897/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13900/

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13902/

mpoeter:
Tests blue

CHANGELOG
```diff
@@ -21,7 +30,7 @@ devel
     This API can only be used in the `_system` database of coordinators, and
     requires admin user privileges.
 
-* Decrease the size of serialized index esimtates, by introducing a 
+* Decrease the size of serialized index estimates, by introducing a
```
jsteemann, line 33: :+1:

arangod/Aql/ClusterNodes.cpp
```diff
@@ -344,13 +324,8 @@ void DistributeNode::toVelocyPackHelper(VPackBuilder& builder, unsigned flags,
...
   writeClientsToVelocyPack(builder);
 
-  builder.add("createKeys", VPackValue(_createKeys));
-  builder.add("allowKeyConversionToObject", VPackValue(_allowKeyConversionToObject));
-  builder.add("fixupGraphInput", VPackValue(_fixupGraphInput));
   builder.add(VPackValue("variable"));
-  _variable->toVelocyPack(builder);
-  builder.add(VPackValue("alternativeVariable"));
-  _alternativeVariable->toVelocyPack(builder);
+  _variable->toVelocyPack(builder);;
```
jsteemann, line 328: I guess this doesn't break in case of rolling upgrades in the cluster, as coordinators are updated last.
However, it could break an older version of the explainer, which may check any of these attributes on a `DistributeNode`. So we should prepare the explainer in 3.7 to handle the new DistributeNode type gracefully.

arangod/Aql/DistributeExecutor.cpp
```diff
@@ -161,129 +123,22 @@ auto DistributeExecutor::distributeBlock(SharedAqlItemBlockPtr block, SkipResult
...
-  _keyBuilder.add(StaticStrings::KeyString,
-                  VPackValuePair(keyPart.data(), keyPart.size(), VPackValueType::String));
-  _keyBuilder.close();
-  // If _key is invalid, we will throw here.
-  auto res = _infos.getResponsibleClient(_keyBuilder.slice());
-  THROW_ARANGO_EXCEPTION_IF_FAIL(res.result());
-  return res.get();
-}
-
 auto DistributeExecutor::getClient(SharedAqlItemBlockPtr block, size_t rowIndex)
```
jsteemann, line 126: ah yeah, this method got a lot better in the new version :partying_face:

arangod/Aql/Functions.cpp
```diff
@@ -8698,6 +8698,178 @@ AqlValue Functions::CallGreenspun(arangodb::aql::ExpressionContext* expressionCo
...
+  if (!canUseCustomKey) {
+    if (ignoreErrors) {
+      return AqlValue{};
+    }
+    THROW_ARANGO_EXCEPTION(TRI_ERROR_CLUSTER_MUST_NOT_SPECIFY_KEY);
+  }
+
+  // convert string key into object with { _key: "string" }
+  TRI_ASSERT(allowKeyConversionToObject);
+  return AqlValue{buildKeyObject(input.stringRef()).slice()};
```
jsteemann, line 8731: This will call `buildKeyObject(...)` for each input, which creates a VPackBuilder temporary.
This VPackBuilder object will be thrown away directly afterwards.

Can we pass in a temporary Builder object into `ConvertToObject`, so an existing Builder can be recycled? Alternatively, we can pass in the current transaction and use sth like
```
transaction::BuilderLeaser tempBuilder(trx);
```

arangod/Aql/Functions.cpp
```diff
@@ -8698,6 +8698,178 @@ AqlValue Functions::CallGreenspun(arangodb::aql::ExpressionContext* expressionCo
...
+    // value is set, but null
+    // check if there is a second input register available (UPSERT makes use of
+    // two input registers,
+    // one for the search document, the other for the insert document)
+    value = extractFunctionParameterValue(parameters, 1);
+    allowSpecifiedKeys = false;
+  }
+      
+  // TODO - use ignoreErrors in all error cases?
+
```
jsteemann, line 8774: Do we need to do anything here? That is unclear to me as a reader of the code.

arangod/Aql/OptimizerRules.cpp
```diff
@@ -8292,3 +8257,211 @@ void arangodb::aql::decayUnnecessarySortedGather(Optimizer* opt,
...
+          allowKeyConversionToObject = true;
+          setInVariable = [updateReplaceNode](Variable* var) {
+            updateReplaceNode->setInKeyVariable(var);
+          };
+        } else {
+          inputVariable = updateReplaceNode->inDocVariable();
+          allowKeyConversionToObject = false;
+          setInVariable = [updateReplaceNode](Variable* var) {
+            updateReplaceNode->setInDocVariable(var);
+          };
```
jsteemann, line 8328: As far as I can tell, the value of `setInVariable` is the same for the `if` and the `else` parts, so it can be moved behind them.

js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -1867,7 +1867,7 @@ function processQuery(query, explain, planIndex) {
       case 'RemoteNode':
         return keyword('REMOTE');
       case 'DistributeNode':
-        return keyword('DISTRIBUTE') + '  ' + annotation('/* create keys: ' + node.createKeys + ', variable: ') + variableName(node.variable) + ' ' + annotation('*/');
+        return keyword('DISTRIBUTE') + '  ' + annotation('/* variable: ') + variableName(node.variable) + ' ' + annotation('*/');
```
jsteemann, line 1870: Shall we change this to just
```suggestion
        return keyword('DISTRIBUTE') + ' ' + variableName(node.variable) ;
```

arangod/Aql/Functions.cpp
mpoeter: Ah, I did not know about `BuilderLeaser`, but that makes sense!

arangod/Aql/OptimizerRules.cpp
mpoeter: No, in one case we call `setInKeyVariable`, in the other `setInDocVariable`.

js/common/modules/@arangodb/aql/explainer.js
mpoeter: Fine by me!

mpoeter: Done

mpoeter:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13904/

arangod/Aql/Functions.cpp
```diff
@@ -8698,17 +8698,18 @@ AqlValue Functions::CallGreenspun(arangodb::aql::ExpressionContext* expressionCo
   }
 }
 
-static VPackBuilder buildKeyObject(VPackStringRef key, bool closeObject = true) {
-  VPackBuilder builder;
+static void buildKeyObject(VPackBuilder& builder, VPackStringRef key, bool closeObject = true) {
+  ;
```
jsteemann, line 8702:
```suggestion
```

jsteemann (approved):

PR #13512, opened 2021-02-08
micro cleanup

jsteemann:
### Scope & Purpose

Minimal code cleanup. Fixes whitespace in function declarations, adds a code comment, plus adds an early exit for case in which `_estimator` is null in indexes. This situation should never occur, however, it is better to make the code safe by not derefencing a pointer that under some (other) conditions can be a nullptr.

This is an internal-only change and should not require any CHANGELOG entry.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13901/

### DIFF
--- a/arangod/Agency/Store.h
+++ b/arangod/Agency/Store.h
@@ -181,7 +181,6 @@ class Store {
   /// @brief Clear entries, whose time to live has expired
   query_t clearExpired() const;
 
-  /// @brief Run thread
  private:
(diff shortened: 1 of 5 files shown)


jsteemann:
Tout les tests sont bleus!

mpoeter (approved):

PR #13516, opened 2021-02-09
fix the combination of profileQuery and stream: true

jsteemann:
### Scope & Purpose

Fix profiling of AQL queries with the `silent` and `stream` options sets in combination. Using the `silent` option makes a query execute, but discard all its results instantly. This led to some confusion in streaming queries, which can return the first query results once they are available, but don't necessarily execute the full query.
Now, `silent` correctly discards all results even in streaming queries, but this has the effect that a streaming query will likely be executed completely when the `silent` option is set. This is not the default however, and the `silent` option is normally not set. There is no change for streaming queries if the `silent` option is not set.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13541

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in shell_client, shell_server)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13964/


goedderz:
Just a thought: It might not be helpful in this way, because at least on the first glance it effectively changes the behavior as it is in `stream: false`. Would a loop after `stmt.execute();` to consume the cursor work here instead?

dothebart (approved):

dothebart (approved):

js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -2135,6 +2135,11 @@ function profileQuery(data, shouldPrint) {
   stringBuilder.clearOutput();
   let stmt = db._createStatement(data);
   let cursor = stmt.execute();
+  if (options.stream) {
+    while(cursor.hasNext()) {
+      cursor.next();
+    }
```
jsteemann, line 2141: Imagine this running in arangosh... This will do a lot of back-and-forth communication with arangod. No doubt it will work, but this is very likely inefficient.

goedderz: Well, it fetches all batches. But I think that's the only sensible thing to do in order to profile a streaming query. At least I currently can't think of something better.

jsteemann: We could do this server-sided instead of client-sided.

goedderz: My thought was that making this server-sided will be effectively the same as setting `stream: false`, rendering it somewhat useless. But perhaps the query profiling doesn't make much sense with a streaming cursor, and we should just forbid it, or emit a warning. I'm fine either way.

jsteemann: I will try to add a warning here. 
Apart from that, I think we can leave the change in this PR as it is, i.e. merge it. 
Let me add that warning and a test now.

jsteemann: Changed my mind and fixed it properly. See latest commit.

jsteemann:
bug-fix/streaming-profiling-fix

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13964/

jsteemann:
Tests blue

PR #13519, opened 2021-02-10
Fix potential undefined behavior

jsteemann:
### Scope & Purpose

Fix potential undefined behavior when iterating over connected nodes in an execution plan and calling callbacks for each of the nodes: if the callbacks modified the list of connected nodes of the current that they were called from, this could lead to potentially undefined behavior due to iterator invalidation. The issue occurred when using a debug STL via `_GLIBCXX_DEBUG`.

I verified that the issue (debug STL assertion failure when running the cluster AQL tests) goes away with this PR.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13521, *3.6*: https://github.com/arangodb/arangodb/pull/13522

### Testing & Verification

  - [x] Added new **integration tests** (shell_server_aql, but only with debug STL)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13911/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,12 @@
 devel
 -----
 
+* Fix potential undefined behavior when iterating over connected nodes in an
+  execution plan and calling callbacks for each of the nodes: if the callbacks
+  modified the list of connected nodes of the current that they were called
+  from, this could lead to potentially undefined behavior due to iterator
+  invalidation. The issue occurred when using a debug STL via `_GLIBCXX_DEBUG`.
+
 * Issue #13141: The `move-filters-into-enumerate` optimization, when applied to
   an EnumerateCollectionNode (i.e. full collection scan), did not do regular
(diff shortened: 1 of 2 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13922/

goedderz (approved):
LGTM, though this makes me wonder under which circumstances it makes sense to change dependencies of the current node, and just continue the walk to the old dependencies. It might indicate a bug in the walker where this happens.

neunhoef (approved):
LGTM

jsteemann:
> LGTM, though this makes me wonder under which circumstances it makes sense to change dependencies of the current node, and just continue the walk to the old dependencies. It might indicate a bug in the walker where this happens.

I think we first need clarification about how many optimizer rules this actually affects. I can run a test today to find this out.

PR #13533, opened 2021-02-10
Improvements for dropping indexes

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13532/

* Avoid a potential deadlock when dropping indexes.

  A deadlock could theoretically happen for a thread that is attempting to drop an index in case there was another thread that tried to create or drop an index in the very same collection at the very same time. We haven't managed to trigger the deadlock with concurrency tests, so it may have been a theoretical issue only. The underlying code was changed anyway to make sure this will not cause problems in reality.

* Make dropping of indexes in cluster retry in case of precondition failed.

  When dropping an indexes of a collection in the cluster, the operation could fail with a "precondition failed" error in case there were simultaneous index creation or drop actions running for the same collection. The error was returned properly internally, but got lost at the point when `<collection>.dropIndex()` simply converted any error to just `false`. We can't make `dropIndex()` throw an exception for any error, because that would affect downwards-compatibility. But in case there is a simultaneous change to the collection indexes, we can just retry our own operation and check if it succeeds then. This is what `dropIndex()` will do now.

#### Backports:

- [x] No backports required

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in shell_server)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13942/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,27 @@
 v3.6.12 (XXXX-XX-XX)
 --------------------
 
+* Avoid a potential deadlock when dropping indexes.
+
+  A deadlock could theoretically happen for a thread that is attempting to 
+  drop an index in case there was another thread that tried to create or drop 
+  an index in the very same collection at the very same time. We haven't
+  maanged to trigger the deadlock with concurrency tests, so it may have been
+  a theoretical issue only. The underlying code was changed anyway to make
+  sure this will not cause problems in reality.
+
+* Make dropping of indexes in cluster retry in case of precondition failed.
+
+  When dropping an indexes of a collection in the cluster, the operation could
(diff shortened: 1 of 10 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13950/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13955/

neunhoef (approved):
LGTM

jsteemann:
Tests blue except know foxx-cleanup-spec test failure

PR #13536, opened 2021-02-10
Make dropping of indexes in cluster retry in case of precondition failed

jsteemann:
### Scope & Purpose

When dropping an indexes of a collection in the cluster, the operation could fail with a "precondition failed" error in case there were simultaneous index creation or drop actions running for the same collection. The error was returned properly internally, but got lost at the point when `<collection>.dropIndex()` simply converted any error to just `false`. We can't make `dropIndex()` throw an exception for any error, because that would affect downwards-compatibility. But in case there is a simultaneous change to the collection indexes, we can just retry our own operation and check if it succeeds then. This is what `dropIndex()` will do now.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13532, *3.6*: https://github.com/arangodb/arangodb/pull/13533

### Testing & Verification

  - [x] Added new **integration tests** (shell_server)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13947/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13952/

jsteemann:
Tests blue

neunhoef (approved):
LGTM

PR #13538, opened 2021-02-10
potential bugfix for disjoint smartgraph queries that refer to collections on other servers

jsteemann:
### Scope & Purpose

Potential bugfix/workaround for issues with disjoint smart graph queries that have a "WITH" declaration and that can refer to other collections/shards, on different servers.

Given a query such as
```
WITH a, c
FOR v, e IN 1 .. 1 ANY "a/11:82238118" b 
  FOR doc in c
    FILTER doc.color == e.color
    RETURN doc
```
and collections with the following setup:

* `a` is a disjoint smart vertex collection, with 3 shards, on 3 different servers
* `b` is a disjoint smart edge collection, with 3 shards, on 3 different servers
* `c` is an unrelated collection, with just 1 shard, on only one server

The problem here is that `a` and `b` are distributed in the same way, but `c` is not. Especially, on some servers that there will be no shards for `c`.

Now, when the query snippets are created, all extra vertex and edge collections will be enumerated and their shards will be added. Here, this is `c`. But on some servers there are no shards for `c`, so there will be no entry in the collection->shard map for it.

This PR inserts a placeholder for such collections which have no shard attached, and on lookup just ignores them. This seems to work in my quick tests, and prevents the coordinator from running into undefined behavior.
The PR also returns a proper error in case no shard is mapped for a collection, so that it is clear what happened and the coordinator can live on.

It is unclear to me if this is a proper way to address the problem, as my knowledge about the subject is only 5 minutes of code-reading and poking with [the one true debugger](https://news.ycombinator.com/item?id=25858969).

If this fix is considered ok, we will need to add tests for this particular case, and port the fix to devel, too.

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

### Testing & Verification

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [x] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [x] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
[...]

PR #13544, opened 2021-02-12
Bug fix/disjoint smart graph heterogeneous sharding

mchacki:
### Scope & Purpose

Forward Port of:
https://github.com/arangodb/arangodb/pull/13538
There have been no merge conflicts.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/649

Potential bugfix/workaround for issues with disjoint smart graph queries that have a "WITH" declaration and that can refer to other collections/shards, on different servers.

Given a query such as
```
WITH a, c
FOR v, e IN 1 .. 1 ANY "a/11:82238118" b 
  FOR doc in c
    FILTER doc.color == e.color
    RETURN doc
```
and collections with the following setup:

* `a` is a disjoint smart vertex collection, with 3 shards, on 3 different servers
* `b` is a disjoint smart edge collection, with 3 shards, on 3 different servers
* `c` is an unrelated collection, with just 1 shard, on only one server

The problem here is that `a` and `b` are distributed in the same way, but `c` is not. Especially, on some servers that there will be no shards for `c`.

Now, when the query snippets are created, all extra vertex and edge collections will be enumerated and their shards will be added. Here, this is `c`. But on some servers there are no shards for `c`, so there will be no entry in the collection->shard map for it.

This PR inserts a placeholder for such collections which have no shard attached, and on lookup just ignores them. This seems to work in my quick tests, and prevents the coordinator from running into undefined behavior.
The PR also returns a proper error in case no shard is mapped for a collection, so that it is clear what happened and the coordinator can live on.

It is unclear to me if this is a proper way to address the problem, as my knowledge about the subject is only 5 minutes of code-reading and poking with [the one true debugger](https://news.ycombinator.com/item?id=25858969).

If this fix is considered ok, we will need to add tests for this particular case, and port the fix to devel, too.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13974/

### DIFF
--- a/arangod/Aql/QuerySnippet.cpp
+++ b/arangod/Aql/QuerySnippet.cpp
@@ -650,6 +650,16 @@ auto QuerySnippet::prepareFirstBranch(
           if (localGraphNode->isDisjoint()) {
             if (found->second == server) {
               myExp.emplace(shard);
+            } else {
+              // the target server does not have anything to do with the particular
+              // collection (e.g. because the collection's shards are all on other
+              // servers), but it may be asked for this collection, because vertex
+              // collections are registered _globally_ with the TraversalNode and
+              // not on a per-target server basis.
+              // so in order to serve later lookups for this collection, we insert
+              // an empty string into the collection->shard map.
+              // on lookup, we will react to this.
+              localGraphNode->addCollectionToShard(aqlCollection->name(), "");
             }
           } else {
             localGraphNode->addCollectionToShard(aqlCollection->name(), shard);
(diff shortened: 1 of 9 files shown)


arangod/Aql/QuerySnippet.cpp
```diff
@@ -650,14 +650,24 @@ auto QuerySnippet::prepareFirstBranch(
...
+              localGraphNode->addCollectionToShard(aqlCollection->name(), "");
             }
           } else {
             localGraphNode->addCollectionToShard(aqlCollection->name(), shard);
           }
 
           numShards++;
         }
-
+
```
goedderz, line 670:
```suggestion

```

goedderz (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14023/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14028/

PR #13399, opened 2021-01-18
add AQL bit handling functions (#13332)

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13332

This PR add the following bit handling functionality to AQL:

- BIT_AND(array): and-combined result
- BIT_OR(array): or-combined result
- BIT_XOR(array): xor-combined result
- BIT_NEGATE(value, bits): bitwise negation of `value`, with a mask of bits` length
- BIT_TEST(value, index): test if bit at position `index` is set in `value` (indexes are 0-based)
- BIT_POPCOUNT(value): return number of bits set in `value`
- BIT_SHIFT_LEFT(value, shift, bits): bitwise shift-left of `value` by `shift` bits, with a mask of `bits` length
- BIT_SHIFT_RIGHT(value, shift, bits): bitwise shift-right of `value` by `shift` bits, with a mask of `bits` length
- BIT_CONSTRUCT(array): construct a number with bits set at the positions given in the array
- BIT_DECONSTRUCT(value): deconstruct a number into an array with the set bits
- BIT_TO_STRING(value): create a bitstring representation from numeric `value`
- BIT_FROM_STRING(value): parse a bitstring representation into a number

`BIT_AND`, `BIT_OR` and `BIT_XOR` are also available as aggregate functions for usage inside COLLECT AGGREGATE.

All above bit operations support unsigned integer values with up to 32 bits. Using values outside the supported range will make any of these bit functions return `null` and register a warning.

The PR also adds binary (base 2) and hexadecimal (base 16) integer literals to AQL. 
These literals can be used where regular (base 10) integer literal can used. 
The prefix for binary integer literals is `0b`, e.g. `0b10101110`.
The prefix for hexadecimal integer literals i `0x`, e.g. `0xabcdef02`.

Binary and hexadecimal integer literals can only be used for unsigned integers.
The maximum supported value is `(2 ^ 32) - 1`, i.e. `0xffffffff` (hexadecimal) or `0b11111111111111111111111111111111` (binary).

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/607

### Testing & Verification

[...]

Issue #13117, opened 2020-12-01
Aardvark: Weird cursor offsets in query editor.


NiklasBeierl:
## My Environment

* __ArangoDB Version__:        3.5.5-1
* __Storage Engine__:             
* __Deployment Mode__:       
* __Deployment Strategy__:   
* __Configuration__:               
* __Infrastructure__:               
* __Operating System__:        Arch linux with KDE Plamsa.
* __Total RAM in your machine__:        
* __Disks in use__:        
* __Used Package__:              

## Component, Query & Data

__Affected feature__: The query editor in the web interface.
__Problem__:
I will start with a picture:
![Peek 2020-12-01 12-11](https://user-images.githubusercontent.com/22919961/100733832-0fb99700-33cf-11eb-862a-1d46831c2a5c.gif)
(I am aware that there is an error in the query, not relevant here)
So I guess the best way to describe it, is the "cursor" has an offset? This happens pretty often, usually affects some lines and some not and the offset also varies: Usually 1-5 chars. I haven't found a reliable way of getting rid off this behavior.
Browser is latest Google chrome running on Arch linux with KDE Plamsa. I have never encountered this issue in any other text field ever.
If you need additional info on the Arango server, I will have to annoy our sysadmin. 

__Expected result__:
![2020-12-01_12-23-32](https://user-images.githubusercontent.com/22919961/100734490-0bda4480-33d0-11eb-8413-f66be871a658.png)

NiklasBeierl:
Possibly related? https://github.com/ajaxorg/ace/issues/3036

dothebart:
heh, that looks fancy - and scary.
Maybe you've got some special characters in there?
Since there are control characters, that switch sequence left to right?
If you cut'n'paste the query into the github editor, does it behave similar?
does it work if you start all over with a fresh window and *type* the same again?

maxkernbach:
Hi @NiklasBeierl,

The first two characters in line 3 (word `filter`) look weird. Could you try to retype `FILTER` or check if the offset always occurs if there is a `FILTER` statement involved?

NiklasBeierl:
Ill go with the old saying, two pictures say more than two thousand words:
![Peek 2020-12-04 11-41](https://user-images.githubusercontent.com/22919961/101154564-2efe3180-3626-11eb-936b-a37a5d8e894a.gif)

[...]

PR #13558, opened 2021-02-16
Disable ligatures in CSS for Ace editor / fix mode-aql.js

Simran-B:
### Scope & Purpose

Avoid rare display issue in Web UI where `fi` is turned into a ligature that throws off the cursor position.

This seems to occur only on very few systems and we we not able to reproduce it, but it shouldn't hurt to disable ligatures.

Also fixes a mistake of [#13399](https://github.com/arangodb/arangodb/pull/13399/files#diff-f3cdfd124afe97d1e97528e938a91ca713fc35d46a65025b1ce950e2fe4b5e11) which added the AQL bit handling functions to a generated file instead of its source file. The rebuild also updates the generated file with missing AQL functions and fixes a typo.

- [ ] ~Rebuild Ardvaark~ to be done later, or maybe we can finally remove it as build artifact from the repo?

#### Backports:

- [x] No backports required
- [ ] Backports required for: 3.6, 3.7

#### Related Information

- [x] GitHub issue / Jira ticket number: https://github.com/arangodb/arangodb/issues/13117

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -63,6 +63,10 @@ devel
   statement could lead to a wrong count output when used in combination with
   an index which has been created with an array index attribute.
 
+* Fixed issue #13117: Aardvark: Weird cursor offsets in query editor.
+
+  Disabled font ligatures for Ace editor in Web UI to avoid rare display issue.
+
 * Fixed ES-784 regression related to encryption cipher propagation to
   ArangoSearch data.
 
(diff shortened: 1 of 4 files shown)


dothebart (approved):

Simran-B:
Need to change the CSS to apply to all instances of ACE, currently it's just the output editor...

PR #13569, opened 2021-02-18
Feature/optimize aql values

Dronplane:
### Scope & Purpose
PR changes memory layout of AqlValue to allow fast aligned access for stored integers/doubles
Also sorting comparison operator is improved - it does not creates additional SharedAqlItemBlockPtr but instead uses direct pointer to AqlItemBlock in sort matrix.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as Aql related tests

Link to Jenkins PR run:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14032/

Tests blue.


arangod/Aql/AqlValue.h
```diff
@@ -355,8 +464,11 @@ struct AqlValue final {
                      AqlValue const& right, bool useUtf8);
 
   /// @brief Returns the type of this value. If true it uses an external pointer
-  /// if false it uses the internal data structure
-  AqlValueType type() const noexcept;
+  /// if false it uses the internal data structure.
+  /// Do not move it to cpp file as MSVC does not inlines it.
```
gnusi, line 468:
```suggestion
  /// Do not move it to cpp file as MSVC does not inline it.
```

arangod/Aql/AqlValue.cpp
```diff
@@ -33,6 +33,7 @@
 #include "Transaction/Helpers.h"
 #include "Transaction/Methods.h"
 #include "V8/v8-vpack.h"
+#include <utils/math_utils.hpp>
```
gnusi, line 36:
```suggestion

File: arangod/Aql/AqlValue.h [hunk 3]
```diff
@@ -132,7 +134,12 @@ struct AqlValue final {
                            // not managed!
     VPACK_MANAGED_SLICE,   // contains vpack, via pointer to a managed uint8_t
                            // slice, allocated by new[] or malloc()
-    RANGE    // a pointer to a range remembering lower and upper bound, managed
+    RANGE,    // a pointer to a range remembering lower and upper bound, managed
+    VPACK_48BIT_INLINE_INT,    // contains vpack data, inline and unpacked 48bit stored as 64bit int number value (system endianess)
```
gnusi, line 138: let's follow general naming pattern, e.g.
```
VPACK_INLINE_INT48,
VPACK_INLINE_UINT48,
VPACK_INLINE_INT64,
VPACK_INLINE_UINT64,
VPACK_INLINE_DOUBLE
```

arangod/Aql/AqlValue.cpp
```diff
@@ -1473,11 +1786,29 @@ bool std::equal_to<arangodb::aql::AqlValue>::operator()(arangodb::aql::AqlValue
...
-  if (type == arangodb::aql::AqlValue::VPACK_INLINE) {
-    return arangodb::velocypack::Slice(&a._data.internal[0])
-           .binaryEquals(arangodb::velocypack::Slice(&b._data.internal[0]));
+  switch (type) {
+    case arangodb::aql::AqlValue::VPACK_INLINE:
+      return arangodb::velocypack::Slice(a._data.inlineSliceMeta.slice)
+           .binaryEquals(arangodb::velocypack::Slice(b._data.inlineSliceMeta.slice));
+    case arangodb::aql::AqlValue::VPACK_INLINE_INT48:
+    case arangodb::aql::AqlValue::VPACK_INLINE_UINT48:
+      // equal is equal. sign does not matters. So compare unsigned
```
jsteemann, line 1795:
```suggestion
      // equal is equal. sign does not matter. So compare unsigned
```

arangod/Aql/AqlValue.cpp
```diff
@@ -778,6 +898,19 @@ double AqlValue::toDouble(bool& failed) const {
...
   AqlValueType t = type();
   switch (t) {
+    case VPACK_INLINE_INT48:
+      return static_cast<int64_t>(_data.shortNumberMeta.data.int48.val);
+    case VPACK_INLINE_UINT48:
+      return static_cast<uint64_t>(_data.shortNumberMeta.data.uint48.val);
+    case VPACK_INLINE_INT64:
+      return basics::littleToHost(_data.longNumberMeta.data.intLittleEndian.val);
+    case VPACK_INLINE_UINT64:
+      return basics::littleToHost(_data.longNumberMeta.data.uintLittleEndian.val);
```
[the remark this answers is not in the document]
Dronplane: If we got this optimized type of AqlValue - we have created it from corresponding hinted constructor or from slice with length checking, so it always will fit into corresponding value. Or I`m missing the way we could get non-fitting number here ?

arangod/Aql/AqlValue.cpp
```diff
@@ -1399,35 +1656,76 @@ void AqlValue::initFromSlice(arangodb::velocypack::Slice slice, arangodb::velocy
...
-    // Use inline value
-    memcpy(_data.internal, slice.begin(), static_cast<size_t>(length));
-    setType(AqlValueType::VPACK_INLINE);
+  if (length <= sizeof(_data.inlineSliceMeta.slice)) {
+    if (slice.isDouble()) {
+      setType(AqlValueType::VPACK_INLINE_DOUBLE);
+      TRI_ASSERT(length == (sizeof(double) + 1));
+      memcpy(_data.longNumberMeta.data.slice.slice, slice.begin(), static_cast<size_t>(length));
+    } else if (slice.isInteger()) {
+      if (length > 7) {
```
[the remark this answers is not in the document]
Dronplane: No, we should use greater as 7 is slice size. so if slice is 7 bytes then value size is 6 bytes and it still fits 48 bits for shortNumberMeta layout

arangod/StorageEngine/TransactionState.cpp
```diff
@@ -131,7 +131,7 @@ TransactionState::Cookie::ptr TransactionState::cookie(void const* key,
 /// @brief add a collection to a transaction
 Result TransactionState::addCollection(DataSourceId cid, std::string const& cname,
                                        AccessMode::Type accessType, bool lockUsage) {
-#ifdef ARANGODB_ENABLE_FAILURE_TESTS
+#if defined(ARANGODB_ENABLE_MAINTAINER_MODE) && defined(ARANGODB_ENABLE_FAILURE_TESTS)
```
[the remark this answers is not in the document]
Dronplane: Yes, there was an error with building  without ARANGODB_ENABLE_MAINTAINER_MODE but with ARANGODB_ENABLE_FAILURE_TESTS. It`s is not very valid set of cmake flags, but nonetheless

arangod/Aql/AqlValue.cpp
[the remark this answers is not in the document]
jsteemann: Ok, got it.

arangod/Aql/AqlValue.cpp
```diff
@@ -1093,10 +1276,47 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
   }
-
   // if we get here, types are equal or can be treated as being equal
 
   switch (leftType) {
+    case VPACK_INLINE_INT48:
+    case VPACK_INLINE_UINT48:
+    case VPACK_INLINE_INT64:
+    case VPACK_INLINE_UINT64:
+    // if right value type is aslo optimized inline we can optimize comparison
```
goedderz, line 1286:
```suggestion
    // if right value type is also optimized inline we can optimize comparison
```

arangod/Aql/AqlValue.cpp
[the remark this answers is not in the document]
goedderz: I think the point is that the function returns an `int64_t`, but this can hold a `uint64_t`, with possibly larger values than INT64_MAX.

arangod/Aql/AqlValue.cpp
```diff
@@ -1093,10 +1276,47 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
+        if (leftType == VPACK_INLINE_UINT64) {
+          uint64_t l = static_cast<uint64_t>(left.toInt64());
+          uint64_t r = static_cast<uint64_t>(right.toInt64());
+          if (l == r) {
+            return 0;
+          }
+          return (l < r ? -1 : 1);
+        } else {
+          int64_t l = left.toInt64();
+          int64_t r = right.toInt64();
```
goedderz, line 1297: What happens here if `right` is a `uint64_t`?

arangod/Aql/AqlValue.cpp
```diff
@@ -52,10 +52,17 @@ using namespace arangodb::aql;
 // this is a copy of that functionality, because the functions in velocypack
 // are not accessible from here
 namespace {
+
+union DoubleEndianessConverter {
+  double dbl;
+  uint64_t uint;
+};
```
maierlars, line 59: This is UB. See [here](https://en.cppreference.com/w/cpp/language/union).
> It's undefined behavior to read from the member of the union that wasn't most recently written.
As far as I know there is no reason to use this, because a simple memcpy works as well - it is guaranteed that the object representation is not changed. And it produces the same [code](https://godbolt.org/z/zTzGG6).

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,119 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
 
+#ifdef __APPLE__
+#include <libkern/OSByteOrder.h>
+#include <machine/endian.h>
+#elif _WIN32
+#include <stdlib.h>
+#elif __linux__
+#include <endian.h>
+#else
+#pragma messsage("unsupported os or compiler")
```
maierlars, line 121: I don't think it is necessary to exclude systems. We can provide a default implementation [like so](https://godbolt.org/z/Wcv34T). This is always correct, without the need to detect what the host endianness is. In particular it compiles perfectly on clang and gcc, i.e. using a single `mov` for little endian and if available a `movbe` for big endian. Only MSVC can not compile this, because [MSVC](https://developercommunity.visualstudio.com/content/problem/917784/msvc-not-using-bswap-when-possible.html). We can make an `ifdef` for them.

arangod/Aql/AqlValue.cpp
Dronplane: It can`t happen. If we got here - then at line 1287 we checked that leftType is equal to rightType. So here we know that teftType and rightType are both equal and both are not UINT64

arangod/Aql/AqlValue.cpp
```diff
@@ -1093,10 +1276,47 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
+        }
+        // intentional fallthrough to double comparison
+      }
+    [[fallthrough]];
+    case VPACK_INLINE_DOUBLE:
+    // here we could only compare doubles in case of right is also inlined
+    // the same is done in VelocyPackHelper::compare for numbers. Equal types are compared
+    // directly - unequal (or doubles) as doubles
+      if (rightType >= VPACK_INLINE_INT48 &&
+          rightType <= VPACK_INLINE_DOUBLE) {
```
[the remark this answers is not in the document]
Dronplane: There is a static assertion in AqlValue.h that values are adjacent see line 146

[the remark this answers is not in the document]
goedderz: Perfect, overlooked that.

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,119 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+#elif __BYTE_ORDER == __BIG_ENDIAN
+static constexpr bool isLittleEndian() { return false; }
+#endif
+#else
+#pragma messsage("unsupported os or compiler")
+#endif
+
+
+template<typename T>
+VELOCYPACK_FORCE_INLINE T hostToLittle(T in) noexcept {
```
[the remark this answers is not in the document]
Dronplane: Done

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,119 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_ushort(in);
+    }
+#endif
+  }
+  return in;
+}
+
+template<typename T>
+VELOCYPACK_FORCE_INLINE T littleToHost(T in) noexcept {
```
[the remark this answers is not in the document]
Dronplane: Done

arangod/Aql/AqlValue.cpp
```diff
@@ -289,6 +341,12 @@ char const* AqlValue::getTypeString() const noexcept {
 size_t AqlValue::length() const {
   AqlValueType t = type();
   switch (t) {
+    case VPACK_INLINE_INT48:
+    case VPACK_INLINE_UINT48:
+    case VPACK_INLINE_INT64:
+    case VPACK_INLINE_UINT64:
+    case VPACK_INLINE_DOUBLE:
+      // FIXME: values above will immediately throw in Slice::length. Maybe throw directly or return 1?
```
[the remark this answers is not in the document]
Dronplane: Okay. I`ll just remove FIXME and let slice throw like it did before - for consistency reason.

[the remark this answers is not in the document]
Dronplane: Anyway throwing an exception is the performance killer, so throwing there or inside slice won`t make it more dead  😄

arangod/Aql/AqlValue.cpp
```diff
@@ -1093,10 +1276,47 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
+          }
+          return (l < r ? -1 : 1);
+        }
+        // intentional fallthrough to double comparison
+      }
+    [[fallthrough]];
+    case VPACK_INLINE_DOUBLE:
+    // here we could only compare doubles in case of right is also inlined
+    // the same is done in VelocyPackHelper::compare for numbers. Equal types are compared
+    // directly - unequal (or doubles) as doubles
```
goedderz, line 1317: That seems to indicate that the code in VelocyPackHelper::compare is broken. Perhaps we should seize the opportunity to look at that again (perhaps in a different PR), and fix the comparison for large integral values.

arangod/Aql/AqlValue.cpp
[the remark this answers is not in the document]
Dronplane: Indeed. I`ve overlooked that slice is throwing NumberOutOfRange if it detects overflow. I`ll do the same here

arangod/Aql/AqlValue.cpp
```diff
@@ -778,6 +898,19 @@ double AqlValue::toDouble(bool& failed) const {
...
+      return static_cast<int64_t>(_data.shortNumberMeta.data.int48.val);
+    case VPACK_INLINE_UINT48:
+      return static_cast<uint64_t>(_data.shortNumberMeta.data.uint48.val);
+    case VPACK_INLINE_INT64:
+      return basics::littleToHost(_data.longNumberMeta.data.intLittleEndian.val);
+    case VPACK_INLINE_UINT64:
+      return basics::littleToHost(_data.longNumberMeta.data.uintLittleEndian.val);
+    case VPACK_INLINE_DOUBLE: {
+      DoubleEndianessConverter tmp;
+      tmp.uint = basics::littleToHost(_data.longNumberMeta.data.uintLittleEndian.val);
```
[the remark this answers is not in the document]
Dronplane: Indeed. I`ve overlooked that slice is throwing NumberOutOfRange if it detects overflow. I`ll do the same here

arangod/Aql/AqlValue.cpp
```diff
@@ -289,6 +349,12 @@ char const* AqlValue::getTypeString() const noexcept {
 size_t AqlValue::length() const {
   AqlValueType t = type();
   switch (t) {
+    case VPACK_INLINE_INT48:
+    case VPACK_INLINE_UINT48:
+    case VPACK_INLINE_INT64:
+    case VPACK_INLINE_UINT64:
+    case VPACK_INLINE_DOUBLE:
+      //values above will immediately throw in Slice::length - let it be for consitency (one place of throwing)
```
mpoeter, line 352:
```suggestion
      //values above will immediately throw in Slice::length - let it be for consistency (one place of throwing)
```

arangod/Aql/AqlValue.cpp
Dronplane: Good point. Let`s go for memcpy then.

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+#else
+#pragma messsage("unsupported os or compiler")
+#endif
+
+
+template<typename T>
+VELOCYPACK_FORCE_INLINE T hostToLittle(T in) noexcept {
+  static_assert(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1,
+    "Type size is not supported");
+  if constexpr (sizeof(T) == 8) {
```
mpoeter, line 154:
```suggestion
  if constexpr (isLittleEndian()) {
    return in;
  }
  if constexpr (sizeof(T) == 8) {
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+#ifdef __APPLE__
+    return OSSwapHostToLittleInt64(in);
+#elif __linux__
+    return htole64(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_uint64(in);
+    }
```
mpoeter, line 162:
```suggestion
    return _byteswap_uint64(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+  }
+  if constexpr (sizeof(T) == 4) {
+#ifdef __APPLE__
+    return OSSwapHostToLittleInt32(in);
+#elif __linux__
+    return htole32(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_ulong(in);
+    }
```
mpoeter, line 173:
```suggestion
    return _byteswap_ulong(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+  }
+  if constexpr (sizeof(T) == 2) {
+#ifdef __APPLE__
+    return OSSwapHostToLittleInt16(in);
+#elif __linux__
+    return htole16(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_ushort(in);
+    }
```
mpoeter, line 184:
```suggestion
    return _byteswap_ushort(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+#endif
+  }
+  return in;
+}
+
+template<typename T>
+VELOCYPACK_FORCE_INLINE T littleToHost(T in) noexcept {
+  static_assert(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1,
+    "Type size is not supported");
+  if constexpr (sizeof(T) == 8) {
```
mpoeter, line 194:
```suggestion
  if constexpr (isLittleEndian()) {
    return in;
  }
  if constexpr (sizeof(T) == 8) {
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+#ifdef __APPLE__
+    return OSSwapLittleToHostInt64(in);
+#elif __linux__
+    return le64toh(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_uint64(in);
+    }
```
mpoeter, line 202:
```suggestion
    return _byteswap_uint64(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+  }
+  if constexpr (sizeof(T) == 4) {
+#ifdef __APPLE__
+    return OSSwapLittleToHostInt32(in);
+#elif __linux__
+    return le32toh(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_ulong(in);
+    }
```
mpoeter, line 213:
```suggestion
    return _byteswap_ulong(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -110,9 +110,123 @@ uint64_t fasthash32(void const*, std::size_t, uint32_t);
...
+  }
+  if constexpr (sizeof(T) == 2) {
+#ifdef __APPLE__
+    return OSSwapLittleToHostInt16(in);
+#elif __linux__
+    return le16toh(in);
+#elif _WIN32
+    if constexpr (!isLittleEndian()) {
+      return _byteswap_ushort(in);
+    }
```
mpoeter, line 224:
```suggestion
    return _byteswap_ushort(in);
```

3rdParty/velocypack/include/velocypack/velocypack-common.h
```diff
@@ -245,11 +356,11 @@ static inline uint64_t readUInt64(uint8_t const* start) noexcept {
...
 
 static inline void storeUInt64(uint8_t* start, uint64_t value) noexcept {
-  uint8_t const* end = start + 8;
-  do {
-    *start++ = static_cast<uint8_t>(value & 0xffU);
-    value >>= 8;
-  } while (start < end);
+ if constexpr (!isLittleEndian()) {
+   value = hostToLittle(value);
+ }
```
mpoeter, line 361:
```suggestion
 value = hostToLittle(value);
```
If we have the constexpr with the early return in `hostToLittle` this should compile down to a noop anyway.

arangod/Aql/AqlValue.cpp
```diff
@@ -71,19 +78,30 @@ static inline uint8_t intLength(int64_t value) noexcept {
...
+  // if 4 MSB bytes are all 0 then we should check only 4 LSB
+  // we actually set (5 : 1) as we at the end will anyway need to do +1 for last byte - so do it here.
+  uint8_t nSize = (x & UINT64_C(0xFFFFFFFF80000000)) ? x >>= 32, 5 : 1;
+
+  // same trick but with 4 left bytes now checking by 2 bytes
+  nSize += (x & UINT64_C(0xFFFF8000)) ? x >>= 16, 2 : 0;
+
+  // same trick with 2 last bytes
+  nSize += (x & 0xFF80) ? 1 :0;
+  return nSize;
```
mpoeter, line 87: This could be implemented even more efficient with a bit scan (`_BitScanReverse`/ `__builtin_clzll`).

arangod/Aql/AqlValue.cpp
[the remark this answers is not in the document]
mpoeter: Perhaps we should write `if (length > 6+1)` and add a short comment.

arangod/Aql/AqlValue.cpp
```diff
@@ -1473,11 +1794,29 @@ bool std::equal_to<arangodb::aql::AqlValue>::operator()(arangodb::aql::AqlValue
...
+    case arangodb::aql::AqlValue::VPACK_INLINE_UINT48:
+      // equal is equal. sign does not matter. So compare unsigned
+      return a._data.shortNumberMeta.data.uint48.val == b._data.shortNumberMeta.data.uint48.val;
+    case arangodb::aql::AqlValue::VPACK_INLINE_INT64:
+    case arangodb::aql::AqlValue::VPACK_INLINE_UINT64:
+    case arangodb::aql::AqlValue::VPACK_INLINE_DOUBLE:
+      // equal is equal. sign/endianess does not matter
+      return a._data.longNumberMeta.data.intLittleEndian.val == b._data.longNumberMeta.data.intLittleEndian.val;
+      return arangodb::velocypack::Slice(a._data.longNumberMeta.data.slice.slice)
+           .binaryEquals(arangodb::velocypack::Slice(b._data.longNumberMeta.data.slice.slice));
```
mpoeter, line 1817: Something's not right here - why are there two return statements? And the double case probably has to be treated differently.

Dronplane: Good catch 😄  Copy&paste is evil

arangod/Aql/AqlValue.cpp
Dronplane: We checked actually __builtin_clzll works surprisingly slower for small numbers (where there are many leading zeroes)

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: I prefer  to keep it clean to the reader that we do conversion only in big-endian case. An as it is constexpr it costs us nothing.

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: no need to do this. At the end of the functions stays default "return n;"

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: hmm always byteswap on windows? why?

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: hmm always byteswap on windows? why?

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: hmm always byteswap on windows? why?

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: no need to do this. At the end of the functions stays default "return n;"

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: hmm always byteswap on windows? why?

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: ??

3rdParty/velocypack/include/velocypack/velocypack-common.h
Dronplane: ??

arangod/Aql/AqlValue.cpp
Dronplane: 👍

3rdParty/velocypack/include/velocypack/velocypack-common.h
mpoeter: Right, but if we are on a little endian system we can immediately return. This way we can also avoid the `isLittleEndian()` checks for Windows, which is why I added the suggestions to remove them (sorry, I should have added a comment to clarify the intent).

arangod/Aql/AqlValue.cpp
mpoeter: Hm.. the `LZCNT` instruction should have a latency of 3 cycles and a throughput of one cycle, regardless of the input value.

3rdParty/velocypack/include/velocypack/velocypack-common.h
mpoeter: IMHO this should be the job of `hostToLittle` (that's why I suggested the early return) and it should be relatively clear that on a little endian system the function should do nothing. So even though the `if` is more explicit, it also adds some complexity to the code that IMHO is not necessary (the same actually goes for the `if`s in the two functions above). But no strong opinion here - I'm fine with leaving it as is.

Dronplane:
http://jenkins.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14043/

Dronplane:
Tests blue

arangod/Aql/AqlValue.cpp
```diff
@@ -1473,11 +1764,26 @@ bool std::equal_to<arangodb::aql::AqlValue>::operator()(arangodb::aql::AqlValue
...
+      return arangodb::velocypack::Slice(a._data.inlineSliceMeta.slice)
+           .binaryEquals(arangodb::velocypack::Slice(b._data.inlineSliceMeta.slice));
+    case arangodb::aql::AqlValue::VPACK_INLINE_INT48:
+      // equal is equal. sign does not matter. So compare unsigned
+      return a._data.shortNumberMeta.data.int48.val == b._data.shortNumberMeta.data.int48.val;
+    case arangodb::aql::AqlValue::VPACK_INLINE_INT64:
+    case arangodb::aql::AqlValue::VPACK_INLINE_UINT64:
+    case arangodb::aql::AqlValue::VPACK_INLINE_DOUBLE:
+      // equal is equal. sign/endianess does not matter
+      return a._data.longNumberMeta.data.intLittleEndian.val == b._data.longNumberMeta.data.intLittleEndian.val;
```
mpoeter, line 1778: For double this is not correct! For example, `0.0 == -0.0` but the binary representation is different.

Dronplane: I tend to agree, good point. But before here was used binaryEquals - this internally called memcmp on VPack. So having 2 slices one with double(0.0) and another with double(-0.0) we will detect non-equality.  So should we fix this or leave as is and keep old (erroneous) behaviour ? @jsteemann  WDYT ?

mpoeter: Oh, ok - so it was broken before too! I wasn't of that that... :D

jsteemann: It's true comparing by `==` or by using `.binaryEquals(...)` the values wouldn't compare equal.

I don't think this is a problem _here_ because the hash function for AqlValue previously already called `VPackSlice(...).hash()` (in contrast to `.normalizedHash()`).
And the equal_to function for AqlValue previously called `.binaryEquals()`, so it could also compare two logically equivalent values and conisider them non-equal.

For AqlValue, this is not a problem right now. AqlValues were previously compared only inside AqlItemBlock, and were placed as keys into an std::unordered_map to find duplicates. This was only important for AqlValues that used dynamic memory allocations anyway, but does not affect the "simple" AqlValues.

Because the unordered_map structure was changed a while ago and now uses the dynamic memory addresses as keys, it is possible that we don't even need the hash and equal_to functions for AqlValue anymore.
I can't tell yet if this is true, but I suggest we try removing the 2 functions and check if everything still compiles. If it does, we don't need to make any decision on the behavior.
And if this leads to compile errors, we should check from where it is called to make the decision.

mpoeter: I think these functions are still necessary. But for what it's worth, the current behavior (already before this PR) is incorrect. This query:
```
FOR i in [0.0, -0.0, 1.0, -1.0]
  COLLECT x = i WITH COUNT INTO cnt
  RETURN [x, cnt]
```
produces this result:
```
[
  [-1,1],
  [0,1],
  [0,1],
  [1,1]
]
```

PR #13394, opened 2021-01-18
Bug fix/database initial syncer needs patience at times

kvahed:
### Scope & Purpose

*Initial database sync has more robust estimate of timeout for leader keys:*
* Leader API has new `quick` option to return `#docs` if > 1 mio docs
* Follower side does a first `quick` call. If keys received => proceed as before, if count recalculate longer timeout repeat call without `quick` option 

- [X] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [X] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [ ] No backports required
- [x] Backports required for: *3.7* : https://github.com/arangodb/arangodb/pull/13359

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [X] The behavior in this PR was *manually tested*
- [X] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

Please note that for legal reasons we require you to sign the [Contributor Agreement](https://www.arangodb.com/documents/cla.pdf)
before we can accept your pull requests.

PR #13359, opened 2021-01-12
Bug fix 3.7/database initial syncer needs patience at times

kvahed:
### Scope & Purpose

*Wait for getting keys for initial sync more intelligently*

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [ ] No backports required
- [ ] Backports required for: *(Please specify versions and link PRs)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [ ] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [x] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:
https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/13549/

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

Please note that for legal reasons we require you to sign the [Contributor Agreement](https://www.arangodb.com/documents/cla.pdf)
before we can accept your pull requests.

PR #13575, opened 2021-02-18
clean up after initial syncer changes, add tests

jsteemann:
### Scope & Purpose

Cleanup after initial syncer "quick" feature addition (https://github.com/arangodb/arangodb/pull/13394) was merged:
- fixes wrong version number for new startup option
- adds code comments
- adds new test that explicitly tests the "quick" parameter
- adds missing lower bound for the wait time

Re the last point, the previous calculation used a default value of 300 implictly, but in a non-obvious way:
```
if (ndocs > 3750000) { // only bother beyond a 
  maxWaitTime = ndocs * 80;
}
```
This would fall apart if the default value of 300 seconds for the wait time was ever adjusted, and it was not obvious in the code how 3850000 and the default wait time were related.

This is now
```
// use a lower bound for maxWaitTime of 180M microseconds, i.e. 180 seconds
constexpr uint64_t lowerBoundForWaitTime = 180000000;
  
...

// calculate a wait time proportional to the number of documents on the leader
// if we get 1M documents back, the wait time is 80M microseconds, i.e. 80 seconds
// if we get 10M documents back, the wait time is 800M microseconds, i.e. 800 seconds
// ...
maxWaitTime = std::max<uint64_t>(maxWaitTime, ndocs * 80);

// there is an additional lower bound for the wait time as defined initially in
//    _config.applier._initialSyncMaxWaitTime
// we also apply an additional lower bound of 180 seconds here, in case that value
// is configured too low, for whatever reason
maxWaitTime = std::max<uint64_t>(maxWaitTime, lowerBoundForWaitTime);
```

As this is an internal cleanup only, there is intentionally no CHANGELOG entry for it.
There was a CHANGELOG entry for the original PR already, which should be enough.

#### Backports:

- [ ] No backports required
- [x] Backports required for: *3.7* (will be done in https://github.com/arangodb/arangodb/pull/13359)

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14039/

### DIFF
--- a/arangod/Replication/DatabaseInitialSyncer.cpp
+++ b/arangod/Replication/DatabaseInitialSyncer.cpp
@@ -1013,10 +1013,18 @@ Result DatabaseInitialSyncer::fetchCollectionSyncByKeys(arangodb::LogicalCollect
   std::string msg =
       "fetching collection keys for collection '" + coll->name() + "' from " + url;
   _config.progress.set(msg);
+  
+  // use a lower bound for maxWaitTime of 180M microseconds, i.e. 180 seconds
+  constexpr uint64_t lowerBoundForWaitTime = 180000000;
 
+  // note: maxWaitTime has a unit of microseconds
+  uint64_t maxWaitTime = _config.applier._initialSyncMaxWaitTime;
+  maxWaitTime = std::max<uint64_t>(maxWaitTime, lowerBoundForWaitTime);
+  
+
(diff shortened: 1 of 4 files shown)


jsteemann:
Tests blue

kvahed (approved):
LGTM

PR #13579, opened 2021-02-20
added tests for AqlValue in-memory optimization

jsteemann:
### Scope & Purpose

This PR add tests for constructing AqlValues from unsigned integer values.
It fixes a bug when constructing AqlValues from uint64_t values that are larger than INT64_MAX.
It also unifies the handling of +0.0 and -0.0 in some cases and adds a few tests for this.

It does not add tests for constructing AqlValues from signed integers or from floating-point values. These tests can be added in separate PRs.

Parts of this (handling of +0.0 and -0.0) may need to be backported to 3.7 later.

#### Backports:

- [ ] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14049/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14064/

arangod/Aql/Ast.cpp
```diff
@@ -1124,6 +1124,10 @@ AstNode* Ast::createNodeValueInt(int64_t value) {
 
 /// @brief create an AST double value node
 AstNode* Ast::createNodeValueDouble(double value) {
+  if (value == -0.0) {
+    // unify -0.0 and +0.0 
+    value = 0.0;
+  }
```
mpoeter, line 1130: This is inconsistent with the `AqlValueHintDouble` constructor which maps NaN and infinity to null. Is this intentional? If we have to support doubles with NaN values, this complicates comparison if we want to comply to the IEEE 754 standard.
But if we do want to support NaNs, we should at least make sure that we don't have signaling NaNs to avoid invalid operation exceptions!

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+  AqlValueMemory memory;
+  void* p = reinterpret_cast<void*>(&memory);
+
+  // put uint64_t value into an AqlValue directly using the AqlValueHintUInt ctor.
+  // note: we are using placement new here, so that the AqlValue uses the
+  // poisoned memory region
+  new (p) AqlValue((AqlValueHintUInt(value)));
+  // although we are using placement new here, we don't need to call the
+  // destructor here, as the AqlValue with the payloads we use here won't
+  // do anything it is destructor
```
mpoeter, line 91:
```suggestion
  // do anything in its destructor
```

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+  AqlValue& aqlValue = *reinterpret_cast<AqlValue*>(p);
+  runChecksForNumber(aqlValue, expected);
+
+  ASSERT_EQ(value != 0, aqlValue.toBoolean());
```
mpoeter, line 96:
```suggestion
  EXPECT_EQ(value != 0, aqlValue.toBoolean());
```
I prefer to use EXPECT instead of ASSERT for assertions that are not a precondition for following code, because that way the test does not abort with the first failing EXPECT and we get more information. As far as I see all ASSERTs in this test file could be converted to EXPECTs.
I only use ASSERT in cases where we definitely want to abort, because otherwise we would run into issues. E.g.:
```
  ASSERT_NE(nullptr, x);
  EXPECT_EQ(42, *x);
```

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2b, 0xff, 0xff, 0xff, 0x7f, UNINITIALIZED, UNINITIALIZED,
+    0xff, 0xff, 0xff, 0x7f, 0x00, 0x00, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(2147483647ULL), expected);
+}
+
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues48Bit_2147483648) {
+  // 7f ff ff ff
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2c, 0x00, 0x00, 0x00, 0x80, 0x00, UNINITIALIZED,
```
mpoeter, line 283: Why is this value encoded with 5 byes? 4 would be enough..?

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+    0x00, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(2147483648ULL), expected);
+}
+
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues48Bit_4294967294) {
+  // ff ff ff fe
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2c, 0xfe, 0xff, 0xff, 0xff, 0x00, UNINITIALIZED,
```
mpoeter, line 292: Why is this value encoded with 5 byes? 4 would be enough..?

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+    0xfe, 0xff, 0xff, 0xff, 0x00, 0x00, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(4294967294ULL), expected);
+}
+
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues48Bit_4294967295) {
+  // ff ff ff ff
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2c, 0xff, 0xff, 0xff, 0xff, 0x00, UNINITIALIZED,
```
mpoeter, line 301: Why is this value encoded with 5 byes? 4 would be enough..?

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2c, 0xff, 0xff, 0xff, 0xff, 0x7f, UNINITIALIZED,
+    0xff, 0xff, 0xff, 0xff, 0x7f, 0x00, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(549755813887ULL), expected);
+}
+
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues48Bit_549755813888) {
+  // 7f ff ff ff ff
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_INT48, 0x2d, 0x00, 0x00, 0x00, 0x00, 0x80, 0x00,
```
mpoeter, line 328: Why is this value encoded with 6 byes? 5 would be enough..?

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+  };
+  runChecksForUInt64(uint64_t(140737488355327ULL), expected);
+}
+
+// note: in all following tests, the value UNINITIALIZED (0xa5) has a special meaning.
+// if it used, it means that we don't compare the byte, because it is not initialized
+// in certain AqlValue configurations.
+// if you add tests, make sure that none of the test data includes 0xa5!
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues64Bit_281474976710654) {
+  // 80 ff ff ff ff ff
```
mpoeter, line 357:
```suggestion
  // ff ff ff ff ff fe
```

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_UINT64, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, 0x2f,
+    0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(281474976710654ULL), expected);
```
mpoeter, line 362: Why is this value encoded as a 64bit value? It only requires 48bits...

tests/Aql/AqlValueMemoryLayoutTest.cpp
```diff
@@ -0,0 +1,419 @@
...
+}
+
+TEST(AqlValueMemoryLayoutTest, UnsignedLargerValues64Bit_281474976710655) {
+  // ff ff ff ff ff ff
+  uint8_t const expected[] = { 
+    AqlValue::AqlValueType::VPACK_INLINE_UINT64, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, UNINITIALIZED, 0x2f,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x00
+  };
+  runChecksForUInt64(uint64_t(281474976710655ULL), expected);
```
mpoeter, line 371: Why is this value encoded as a 64bit value? It only requires 48bits...

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: Good idea, done that!

arangod/Aql/Ast.cpp
jsteemann: Fixed

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: As discussed, this is necessary because the value is encoded as a signed integer right now.

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: As discussed, this is necessary because the value is encoded as a signed integer right now.

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: As discussed, this is necessary because the value is encoded as a signed integer right now.

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: As discussed, this is necessary because the value is encoded as a signed integer right now.

tests/Aql/AqlValueMemoryLayoutTest.cpp
jsteemann: As discussed, this is necessary because the value is encoded as a signed integer right now.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14075/

gnusi (approved):

tests/js/server/aql/aql-double-conv.js
```diff
@@ -0,0 +1,82 @@
...
+      res = db._query(q2).toArray()[0];
+      assertNull(res);
+      
+      const q3 = `RETURN NOOPT(-1.0) / NOOPT(0.0)`;
+      res = db._query(q3).toArray()[0];
+      assertNull(res);
+      
+      const q4 = `RETURN NOOPT(-1.0) / NOOPT(-0.0)`;
+      res = db._query(q4).toArray()[0];
+      assertNull(res);
```
mpoeter, line 66: We should also add a test for `0.0 / 0.0` - in JavaScript this returns NaN.

mpoeter (approved):

tests/js/server/aql/aql-double-conv.js
jsteemann: Good idea. Done!

PR #13580, opened 2021-02-20
fix a potential wrong length calculation after log messages were shrunk

jsteemann:
### Scope & Purpose

Fix a potentially wrong length calculation after log messages were shrunk.

Reproduction: compile with ASan and execute:
```
scripts/unittest server_parameters --test log-max --cluster true
```

ASan errors are gone with the fix in place.

New ASan run:
http://172.16.10.101:8080/job/arangodb-ANY-linux-asan/560/

This is a fix for a yet-unreleased feature, so intentionally no CHANGELOG entry added for it.

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] The behavior in this PR was *manually tested*
- [x] This change is already covered by existing tests, such as *server_parameters*.
- [x] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14050/
http://172.16.10.101:8080/job/arangodb-ANY-linux-asan/560/

PR #13581, opened 2021-02-20
move some parts of LogMessage from header to cpp file

jsteemann:
### Scope & Purpose

Small cleanup for some logging code

* moved some parts of LogMessage from header to cpp file, so less recompilation will be required if that code is changed later on
* verify that `_offset` does not point outside the log message after shrinking. this is very unlikely, because offset is something between 20 or 30 characters only, and the minimum log message length for shrinking is 256 characters. so it is a theoretical issue right now, but the fix will be useful should the limits/bounds be changed later.

Inspired by a fix that was necessary for 3.7, but didn't affect devel: https://github.com/arangodb/arangodb/pull/13580

Intentionally no CHANGELOG entry written, as it is an internal code cleanup PR only, without effects for end users.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *server_parameters*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14051/

### DIFF
--- a/lib/Logger/Logger.cpp
+++ b/lib/Logger/Logger.cpp
@@ -102,7 +102,7 @@ void LogMessage::shrink(std::size_t maxLength) {
     if (_offset > _message.size()) {
       // we need to make sure that the offset is not outside of the message
       // after shrinking
-      _offset = _message.size();
+      _offset = static_cast<uint32_t>(_message.size());
     }
     _message.append("...", 3);
     _shrunk = true;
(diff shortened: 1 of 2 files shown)


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14065/

mpoeter (approved):

PR #13585, opened 2021-02-20
Cleanup hotbackup transfer jobs in agency

neunhoef:
This PR has two purposes:

(1) Automatically clean up /arango/Target/HotBackup/TransferJobs
(2) Reduce memory footprint of agency store (class Node).

### Scope & Purpose

#### Backports:

- [ ] No backports required
- [*] Backports required for: 3.6 and 3.7

### Testing & Verification

- [*] This change is already covered by existing tests, such as *(please describe tests)*.


neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14055/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14056/

neunhoef:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14068/

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
 
 class Store;
 
+class SmallBuffer {
+  uint8_t* _start;
+  size_t _size;
+ public:
+  SmallBuffer() : _start(nullptr), _size(0) {}
```
jsteemann, line 82: this should be marked `noexcept`.
```suggestion
  SmallBuffer() noexcept : _start(nullptr), _size(0) {}
```

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
...
+  SmallBuffer(SmallBuffer const& other) {
+    if (other.empty()) {
+      _start = nullptr;
+      _size = 0;
+    } else {
+      _start = new uint8_t[other._size];
+      _size = other._size;
+      memcpy(_start, other._start, other._size);
+    }
+  }
```
jsteemann, line 93: nitpicking... it was already fine as it was before, but we could make sure we always initialize the instance variables in the initializer list. but it is fine as it was before.
```suggestion
  SmallBuffer(SmallBuffer const& other) 
      : SmallBuffer() {
    if (!other.empty()) {
      _start = new uint8_t[other._size];
      _size = other._size;
      memcpy(_start, other._start, other._size);
    }
  }
```

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
...
+  SmallBuffer(SmallBuffer&& other) {
+    _start = other._start;
+    other._start = nullptr;
+    _size = other._size;
+    other._size = 0;
+  }
```
jsteemann, line 99:
```suggestion
  SmallBuffer(SmallBuffer&& other) noexcept
      : _start(other._start), _size(other._size) {
    other._start = nullptr;
    other._size = 0;
  }
```

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
...
+  SmallBuffer& operator=(SmallBuffer const& other) {
```
jsteemann, line 100: This would be broken if `this == &other`.
Should we add the usual `if (this != &other) { ... }` check here?

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
...
+      _start = nullptr;
+      _size = 0;
+    } else {
+      _start = new uint8_t[other._size];
+      _size = other._size;
+      memcpy(_start, other._start, other._size);
+    }
+    return *this;
+  }
+  SmallBuffer& operator=(SmallBuffer&& other) {
```
jsteemann, line 114:
```suggestion
  SmallBuffer& operator=(SmallBuffer&& other) noexcept {
```

arangod/Agency/Supervision.cpp
```diff
@@ -1684,6 +1688,89 @@ void Supervision::cleanupFinishedAndFailedJobs() {
...
+    Node const& snapshot,
+    std::shared_ptr<VPackBuilder> envelope) {
+  // This deletes old Hotbackup transfer jobs in 
+  // /Target/HotBackup/TransferJobs according to their time stamp.
+  // We keep at most 100 transfer jobs which are completed.
+  constexpr uint64_t maximalNumberTransferJobs = 100;
+  constexpr char const* prefix = "/Target/HotBackup/TransferJobs/";
+
+  auto const& jobs = snapshot.hasAsChildren(prefix).first;
+  if (jobs.size() <= maximalNumberTransferJobs + 6) {
```
jsteemann, line 1702: I think the intention behind the "+ 6" here should be explained by a code comment.

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
...
+  }
+  SmallBuffer(SmallBuffer&& other) {
+    _start = other._start;
+    other._start = nullptr;
+    _size = other._size;
+    other._size = 0;
+  }
+  SmallBuffer& operator=(SmallBuffer const& other) {
+    if (!empty()) {
+      delete[] _start;
```
jsteemann, line 102: we could even optimize this case when `other` is not empty and we already have a buffer allocated. In case our own buffer is at least as large as the other buffer, we can copy the other's buffer into ours and save a `delete[]` and a `new[]`. However, that will also require tracking the buffer _capacity_ in addition to the buffer size, so it would be a somewhat larger change.

arangod/Agency/Node.h
```diff
@@ -174,7 +228,10 @@ class Node final {
   VPackBuilder toBuilder() const;
 
   /// @brief Access children
-  Children& children();
+  //Children& children();
```
jsteemann, line 231:
```suggestion
```

arangod/Agency/Node.h
```diff
@@ -75,6 +75,60 @@ typedef std::chrono::steady_clock::time_point SteadyTimePoint;
 
 class Store;
 
+class SmallBuffer {
+  uint8_t* _start;
+  size_t _size;
+ public:
+  SmallBuffer() : _start(nullptr), _size(0) {}
+  SmallBuffer(size_t size) : _start(new uint8_t[size]), _size(size) {}
```
jsteemann, line 83: Do we want to handle the case of `size == 0` here?
Not sure if it is really worth it.
```suggestion
  explicit SmallBuffer(size_t size) : SmallBuffer() {
    if (size > 0) {
      _start = new uint8_t[size];
      _size = size;
    }
  }
```

arangod/Agency/Node.cpp
```diff
@@ -82,11 +82,14 @@ void Node::rebuildVecBuf() const {
...
-        tmp.add(Slice(i.data()));
+      if (_value != nullptr) {
+        for (auto const& i : *_value) {
+          tmp.add(Slice(i.data()));
+        }
       }
     }
-    _vecBuf = *tmp.steal();
+    _vecBuf = std::make_unique<SmallBuffer>(tmp.size());
+    memcpy(_vecBuf->data(), tmp.data(), tmp.size());
```
jsteemann, line 92: We have a few such places that create a SmallBuffer object and then memcpy into it.
I would prefer it we added a SmallBuffer constructor that would do this automatically, i.e.
```
SmallBuffer(uint8_t const* data, size_t size) : SmallBuffer() {
  if (size > 0) {
    _start = new uint8_t[size];
    _size = size;
    memcpy(_start, data, size);
  }
}
```
Or maybe a `char const*` instead of `uint8_t const*` in the signature, or both.
Anyway, IMHO all the copying should be done by the SmallBuffer object and not by callers.

arangod/Agency/Node.cpp
```diff
@@ -133,15 +138,22 @@ Node::Node(Node const& other)
...
+      _value(nullptr),
+      _vecBuf(other._vecBuf ? std::make_unique<SmallBuffer>(*other._vecBuf) : nullptr),
       _vecBufDirty(other._vecBufDirty),
       _isArray(other._isArray) {
-  for (auto const& p : other._children) {
-    auto copy = std::make_shared<Node>(*p.second);
-    copy->_parent = this;  // new children have us as _parent!
-    _children.insert(std::make_pair(p.first, copy));
+  if (other._children) {
+    _children = std::make_unique<Children>();
```
jsteemann, line 148: We could reserve enough space for all children to save a few vector reallocations:
```
_children->reserve(other._children->size());
```

arangod/Agency/Node.cpp
```diff
@@ -152,19 +164,22 @@ Node::Node(Node const& other)
...
   if (slice.isArray()) {
     _isArray = true;
-    _value.resize(slice.length());
+    _value = std::make_unique<std::vector<SmallBuffer>>();
+    _value->reserve(slice.length());
     for (size_t i = 0; i < slice.length(); ++i) {
-      _value.at(i).append(reinterpret_cast<char const*>(slice[i].begin()),
-                          slice[i].byteSize());
+      _value->emplace_back(slice[i].byteSize());
+      memcpy((*_value)[i].data(), slice[i].start(), slice[i].byteSize());
```
jsteemann, line 175: IMO copying should be performed by a SmallBuffer constructor

arangod/Agency/Node.cpp
```diff
@@ -152,19 +164,22 @@ Node::Node(Node const& other)
...
     }
   } else {
     _isArray = false;
-    _value.resize(1);
-    _value.front().append(reinterpret_cast<char const*>(slice.begin()), slice.byteSize());
+    _value = std::make_unique<std::vector<SmallBuffer>>();
+    _value->reserve(1);
```
jsteemann, line 180: We don't need to reserve one element here. The `emplace_back` will do this already. The `reserve(1)` doesn't harm here, but it will also not help.

arangod/Agency/Node.cpp
```diff
@@ -201,14 +218,40 @@ Node& Node::operator=(Node const& rhs) {
...
-  }
-  _value = rhs._value;
-  _vecBuf = rhs._vecBuf;
+  if (rhs._children) {
+    if (_children) {
+      _children->clear();
+    } else {
+      _children = std::make_unique<Children>();
+    }
+    for (auto const& p : *rhs._children) {
```
jsteemann, line 227: here we could again reserve enough space for the children to come.

arangod/Agency/Node.cpp
```diff
@@ -1196,22 +1267,22 @@ Slice Node::getArray() const {
...
 }
 
 void Node::clear() {
-  _children.clear();
+  _children.reset();
   removeTimeToLive();
-  _value.clear();
-  _vecBuf.clear();
+  _value.reset();
+  _vecBuf.reset();
```
jsteemann, line 1277: Not sure how often this is called, but instead of `reset`ing `_value` and `_children`, we may also call `->clear()` on them. This will leave the objects in place with empty vectors, so the vectors can be recycled later if the objects are refilled.
But I can't tell if that is a frequent use case and worth it.

tests/Agency/NodeTest.cpp
```diff
@@ -44,6 +46,118 @@ namespace arangodb {
...
+  SmallBuffer sb;
+  ASSERT_EQ(nullptr, sb.data());
+  ASSERT_EQ(0, sb.size());
+  ASSERT_TRUE(sb.empty());
+}
+
+TEST(SmallBufferTest, sizeConstructor) {
+  for (size_t i : std::vector<size_t>({10, 123, 2049, 65536, 1237235})) {
+    SmallBuffer sb(i);
+    ASSERT_TRUE(sb.data() != nullptr);
```
jsteemann, line 59:
```suggestion
    ASSERT_NE(sb.data(), nullptr);
```

tests/Agency/NodeTest.cpp
```diff
@@ -44,6 +46,118 @@ namespace arangodb {
...
+}
+
+TEST(SmallBufferTest, copyConstructor) {
+  SmallBuffer sb1(123);
+  for (size_t i = 0; i < 123; ++i) {
+    sb1.data()[i] = (uint8_t)i;
+  }
+  SmallBuffer sb2(sb1);
+  ASSERT_TRUE(sb2.data() != nullptr);
+  ASSERT_TRUE(sb2.data() != sb1.data());
```
jsteemann, line 72:
```suggestion
  ASSERT_NE(sb2.data(), nullptr);
  ASSERT_NE(sb2.data(), sb1.data());
```

tests/Agency/NodeTest.cpp
```diff
@@ -44,6 +46,118 @@ namespace arangodb {
...
+
+TEST(SmallBufferTest, copyAssignment) {
+  SmallBuffer sb1(123);
+  for (size_t i = 0; i < 123; ++i) {
+    sb1.data()[i] = (uint8_t)i;
+  }
+  SmallBuffer sb2;
+  sb2 = sb1;
+  ASSERT_TRUE(sb2.data() != nullptr);
+  ASSERT_TRUE(sb2.data() != sb1.data());
```
jsteemann, line 108:
```suggestion
  ASSERT_NE(sb2.data(), nullptr);
  ASSERT_NE(sb2.data(), sb1.data());
```

tests/Agency/NodeTest.cpp
```diff
@@ -44,6 +46,118 @@ namespace arangodb {
...
+  ASSERT_FALSE(sb2.empty());
+  ASSERT_EQ(123, sb2.size());
+  for (size_t i = 0; i < 123; ++i) {
+    ASSERT_EQ((uint8_t) i, sb2.data()[i]);
+  }
+  SmallBuffer sb3(147);
+  sb3 = sb1;
+  ASSERT_TRUE(sb3.data() != nullptr);
+  ASSERT_TRUE(sb3.data() != sb1.data());
```
jsteemann, line 117:
```suggestion
  ASSERT_NE(sb3.data(), nullptr);
  ASSERT_NE(sb3.data(), sb1.data());
```

kvahed (approved):
Agree with all of jan's points and otherwise approve.

arangod/Agency/Node.cpp
neunhoef: I think it can happen that a particular `Node` object switches between being a JSON object (in which case it allocates `_children` unless it is empty) and something else (JSON array or other value) but I do not expect it to happen very frequently. Therefore I have chosen to always clear up all possible allocations.

arangod/Agency/Node.h
neunhoef: Yes, this is a good idea, will do.

arangod/Agency/Node.h
neunhoef: I wanted to avoid exactly this and since we do not have realloc in C++, I chose against it.

arangod/Agency/Supervision.cpp
neunhoef: Already added a comment.

arangod/Agency/Node.h
neunhoef: Done.

arangod/Agency/Node.cpp
neunhoef: Good point...

arangod/Agency/Node.h
jsteemann: Fine

arangod/Agency/Node.cpp
neunhoef: Changed.

arangod/Agency/Node.cpp
neunhoef: Done.

arangod/Agency/Node.cpp
neunhoef: Done.

jsteemann (approved):

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14180/

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14196/

PR #13597, opened 2021-02-22
Prevent arangosh from trying to connect after every executed command.

jsteemann:
### Scope & Purpose

Prevent arangosh from trying to connect after every executed command.

This fixes the case when arangosh is started with default options, but no server is running on localhost:8529. In this particular case, arangosh will try to connect on startup and after every executed shell command. The connect attempts all fail and time out after 300ms.
In this case we now don't try to reconnect after every command.

Reproduction instructions:
- start an arangosh with default options
- make sure no server is running on localhost:8529
- execute a few trivial commands in arangosh (e.g. `1`)
- without this change, they will each take 300ms. with this change, they will be instant.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14082/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14118/

hkernbach (approved):

PR #13603, opened 2021-02-24
fix replication fuzz test

dothebart:
### Scope & Purpose

this PR fixes the replication fuzzing test and synchronizes it with the release-test-automation.

### DIFF
--- a/tests/js/server/replication/fuzz/replication-fuzz-global.js
+++ b/tests/js/server/replication/fuzz/replication-fuzz-global.js
@@ -138,6 +138,7 @@ function ReplicationSuite() {
           db._flushCache();
         }
       }
+      db._useDatabase('_system');
     } else {
 
       var printed = false;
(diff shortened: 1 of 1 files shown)


dothebart:
http://jenkins01.arangodb.biz:8080/view/Windows/job/arangodb-ANY-windows-matrix/1821/

tests/js/server/replication/fuzz/replication-fuzz-global.js
```diff
@@ -118,14 +122,18 @@ function ReplicationSuite() {
...
-          }
+        db._flushCache();
+        db._useDatabase('_system');
+        db._databases().forEach(function(d) {
+          db._useDatabase(d);
+          db._collections().forEach(col => {
+            if (col.name()[0] !== '_') {
+              lastLogTick += col.count() + 1;
+            }
+          });
```
jsteemann, line 133: After iteration over all available databases, we may end up being in a different database than before. Should we add an `db._useDatabase('_system');` after this loop, or even go back to the database that we left off in?

jsteemann (approved):

PR #13614, opened 2021-02-25
Enable statistics in web UI in non-`_system` databases in cluster mode.

jsteemann:
### Scope & Purpose

In cluster mode, the web UI dashboard did not display statistics properly when not being logged in to the `_system` database. For all other databases than `_system`, no statistics were displayed but just some "No data..." placeholders.
Statistics for non-`_system` databases were not properly displayed since 3.7.6 (and 3.6.10) due to an internal change in the statistics processing.

In addition, a new startup option `--server.statistics-all-databases` was added that controls whether cluster statistics are displayed in the web interface for all databases (if the option is set to `true`) or just for the system database (if the option is set to `false`).
The default value for the option is `true`, meaning statistics will be displayed in the web interface for all databases.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13615, *3.6*: https://github.com/arangodb/arangodb/pull/13616

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in authentication)

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14129/


jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14195/

jsteemann:
Test failure is a known and unrelated hot backup test failure.

CHANGELOG
```diff
@@ -1,11 +1,26 @@
...
+  controls whether cluster statistics are displayed in the web interface for
+  all databases (if the option is set to `true`) or just for the system
+  database (if the option is set to `false`).
+  The default value for the option is `true`, meaning statistics will be
+  displayed in the web interface for all databases. 
 
 * Added logging of elapsed time of ArangoSearch commit/consolidation/cleanup 
   jobs.
 
+* Updated OpenSSL to 1.1.1j and OpenLDAP to 2.4.57.
```
joerg84, line 22: I assume this is not related to this PR?

arangod/Statistics/StatisticsFeature.cpp
```diff
@@ -337,14 +350,22 @@ void StatisticsFeature::collectOptions(std::shared_ptr<ProgramOptions> options)
...
   options->addOption("--server.statistics",
                      "turn statistics gathering on or off",
-                     new BooleanParameter(&_statistics),
-                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden));
+                     new BooleanParameter(&_statistics));
+
   options->addOption("--server.statistics-history",
                      "turn storing statistics in database on or off",
-                     new BooleanParameter(&_statisticsHistory),
-                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden))
```
joerg84, line 345: Just curious: why do we remove arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden)) here?

joerg84 (approved):
It looks good to me, but we should have another review as I am unfamiliar with this parts.

Dronplane (approved):
LGTM only minor comment

CHANGELOG
jsteemann: No, it is unrelated. It just moves the most recent change to the top.

arangod/Statistics/StatisticsFeature.cpp
jsteemann: I think it is a useful option that deserves more attention. Using it can greatly reduce idle write load on an instance, so making people aware of that it exists is useful IMHO.

js/apps/system/_admin/aardvark/APP/statistics.js
```diff
@@ -406,46 +406,9 @@ router.use((req, res, next) => {
...
-            deletesPerSecond: s.http.requestsDeletePerSecond,
-            othersPerSecond: s.http.requestsOptionsPerSecond,
-            patchesPerSecond: s.http.requestsPatchPerSecond
-          }
-      }`;
-    
-    const params = { start: startOffsetSchema - 2 * STATISTICS_INTERVAL, clusterIds: coordinators };
-    const stats15 = db._query(stats15Query, params).toArray();
-    const statsSamples = db._query(statsSampleQuery, params).toArray();
+    let start = internal.time() - 20 * STATISTICS_INTERVAL;
```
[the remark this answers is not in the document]
jsteemann: Before it was actually `startOffsetSchema - 2 * STATISTICS_INTERVAL`, i.e.
`function - 2 * 10`, i.e. `NaN - 2 * 10`, i.e. `NaN`.
(as `startOffsetSchema` is a function and the previous code did not call it, just reference it in the calculation).
I think using 20 times the interval is ok here, as that would mean 20 * 10 = 200 seconds. This should be enough backlog to fetch to display most recent statistics.
12 * 10 = 120 seconds would probably also be ok, as probably would be 6 * 10 = 60 seconds.
So the number used here is pretty arbitrary. But I will change it to 120.

jsteemann, line 410:
```suggestion
    let start = internal.time() - 12 * STATISTICS_INTERVAL;
```

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14204/

PR #13625, opened 2021-02-25
make hot backup test clean up all its temp backups and stop as soon as possible

jsteemann:
### Scope & Purpose

This makes a hotbackup test clean up all hotbackups taken during the test (plus all other ones that may exist - we don't care).
Otherwise all these hotbackups may end up in the tar archive that our CI creates when a test run fails. We don't want all the hotbackups to be included there, as they will only take a lot of space.

In addition, this change makes the test stop as early as possible. Previously it kept running for a while even though the to-be-tested conditions were already covered enough.

Furthermore, this PR makes the same test less time critical by considering ASAN and TSAN and coverage builds in the choice of the timeout.

This is only a test change, so there is intentionally no CHANGELOG.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13627

### Testing & Verification

Link to Jenkins PR run:


KVS85 (commented):
LGTM.

dothebart (approved):

PR #13633, opened 2021-02-28
Allow dumping of individual shards with arangodump

jsteemann:
### Scope & Purpose

Added `--shard` option to arangodump, so that dumps can be restricted to one or multiple shards only.

In addition fix an unrelated assertion failure in arangorestore when using `--create-collection false` and clean up an old option in arangodump.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14202/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 devel
 -----
 
-* Add `--shard` option to arangodump, so that dumps can be restrict to one or
+* Added `--shard` option to arangodump, so that dumps can be restricted to one or
   multiple shards only.
 
 * Updated OpenSSL to 1.1.1j and OpenLDAP to 2.4.57.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Add `--shard` option to arangodump, so that dumps can be restrict to one or
+  multiple shards only.
+
 * Updated OpenSSL to 1.1.1j and OpenLDAP to 2.4.57.
 
 * Added logging of elapsed time of ArangoSearch commit/consolidation/cleanup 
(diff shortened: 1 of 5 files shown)


CHANGELOG
```diff
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Add `--shard` option to arangodump, so that dumps can be restrict to one or
```
```suggestion
* Added `--shard` option to arangodump, so that dumps can be restricted to one or
```

arangosh/Dump/DumpFeature.cpp
```diff
@@ -625,6 +633,12 @@ void DumpFeature::collectOptions(std::shared_ptr<options::ProgramOptions> option
       "--collection",
       "restrict to collection name (can be specified multiple times)",
       new VectorParameter<StringParameter>(&_options.collections));
+  
+  options->addOption(
+      "--shard",
+      "restrict to shard name (can be specified multiple times)",
```
jsteemann, line 639:
```suggestion
      "restrict dump to shard (can be specified multiple times)",
```

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14205/

arangosh/Restore/RestoreFeature.cpp
```diff
@@ -1164,7 +1164,9 @@ arangodb::Result processInputDirectory(
         if (result.fail()) {
           return result;
         }
+      }
 
+      if (progressTracker.getStatus(name.copyString()).state < arangodb::RestoreFeature::CREATED) {
```
maierlars, line 1169: we could change `getStatus` to expect a `std::string_view` but I guess this is fine.

maierlars (approved):

arangosh/Restore/RestoreFeature.cpp
jsteemann: Yes, we should do this, but seperately.

PR #13560, opened 2021-02-17
fixed BTS-309

hkernbach:
### Scope & Purpose

From our old changelog:
```
The General Graph document API is now consistent with the document API in its error messages.When attempting to create / modify edges pointing to non existing vertex collections HTTP 400 is returned instead of 404.
```

**The described change above was not implemented. This PR fixes it.** 

Proper statusCode is now returned, additionally, two new Errors were implemented to more precise and to distinguish between a "not found collection / returns 404" and a "not found collection in parameter but the route, in general, is valid / return 400 (as the request is invalid here)".

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [ ] No backports required
- [x] Backports required for: 3.6, 3.7

#### Related Information

- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/BTS-309

### Testing & Verification

- [x] The behavior in this PR was *manually tested*
- [x] This PR adds tests that were used to verify all changes:
  - [x] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

PR #13644, opened 2021-03-02
BTS-309: Rename error message 1947 to match actual behavior

Simran-B:
Follow-up PR to #13560 to correct the error message to "referenced vertex collection is not part of graph" instead of "... does not exist". In the latter case, a different error and HTTP status 404 is returned. The actual check that raises error 1947 only looks at the edge definitions to determine if the vertex collection is part of the graph.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,7 +4,7 @@ devel
 * Fixed issue BTS-309: The Graph API (Gharial) did not respond with the correct
   HTTP status code when validating edges. It now responds with 400 (Bad Request)
   as documented and a new, more precise error code (1947) and message if a vertex
-  collection referenced in the _from or _to attribute cannot be found.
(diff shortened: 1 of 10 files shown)


Simran-B:
About the naming of the error code `TRI_ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_IN_EDGE_DEF`:
There is a similar error message for edge collections which has the suffix "NOT_USED":

`ERROR_GRAPH_EDGE_COLLECTION_NOT_USED,1930,"edge collection not used in graph","The edge collection is not used in any edge definition of the graph.",`

I thus renamed the new one to `TRI_ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_USED`, but kept the phrasing "is not part of the graph".

hkernbach (approved):

PR #13659, opened 2021-03-05
Fix lost coordinators in UI after long maintenance

neunhoef:
This PR fixes the following problem: If one switches the agency
supervision to maintenance mode and leaves this on for more than
24h, then the first run of the supervision maintenance would erase all
coordinators (and dbservers which are not shard leaders) in the agency
data. This leads to the coordinators (and potentially dbservers) become
invisible in the UI.

The reason for this is that after 24h the heartbeats arriving in the
agency have not been processed in the agency supervision (due to it
being in maintenance mode), therefore the subroutine
cleanupExpiredServers would erase the servers as a cleanup job.

The fix is to run this subroutine after the most recent heartbeats
have been processed and to update the transient store snapshot in
between.

### Scope & Purpose

    [*] hankey Bugfix (requires CHANGELOG entry)

Backports:

    [*] Backports required for: 3.6 and 3.7
        This is a backport of https://github.com/arangodb/arangodb/pull/13658

### Testing & Verification

    [*] This change is a trivial rework / code cleanup without any test coverage.

        The behaviour was manually tested by setting the times for the cleanup
        to much smaller values. We would have to run a 24h test to test this
        behaviour or mess with the numbers, which is both not desirable.


kvahed (commented):
LGTM

neunhoef:
This is good to go. The Jenkins Windows failure was fake, another run: http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr-windows/7726/ worked.

PR #13662, opened 2021-03-05
Feature/UI metrics view

hkernbach:
### Scope & Purpose

Adds internal forwarding of metrics.
Adds standalone and cluster views for metrics (coordinator and dbserver specific in cluster case)
Standalone: Dashboard -> Metrics
Cluster: Nodes -> Select a node -> Metrics 

#### Backports:

- [x] No backports required

#### Related Information

https://arangodb.atlassian.net/browse/GORDO-1031

### Testing & Verification

### Documentation


hkernbach:
@jsteemann currently it seems that API does not provide server specific metrics. Need to check this after the weekend.

arangod/CMakeLists.txt
```diff
@@ -967,6 +967,7 @@ target_link_libraries(arango_iresearch arango_indexes)
 target_link_libraries(arango_iresearch arango_cluster_engine)
 
 target_link_libraries(arango_metrics arango)
+target_link_libraries(arango_metrics arango::validation)
```
jsteemann, line 970: Wondering what this change has to do with the metrics UI changes... :thinking:

arangod/RestServer/Metrics.cpp
```diff
@@ -102,8 +102,8 @@ void Counter::store(uint64_t const& n) {
 
 void Counter::toPrometheus(std::string& result) const {
   _b.push();
-  result += "\n#TYPE " + name() + " counter\n";
-  result += "#HELP " + name() + " " + help() + "\n";
+  result += "\n# HELP " + name() + " " + help() + "\n";
+  result += "# TYPE " + name() + " counter\n";
```
jsteemann, line 106: Ok... why do we need to swap the order of `TYPE` and `HELP` here, and why do we need the extra space?
According to https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#grouping-and-sorting these changes are valid, but not necessary. 
To keep our API output as close to previous ArangoDB versions as possible, I would favor reverting this change.

arangod/V8Server/v8-statistics.cpp
```diff
@@ -218,6 +218,21 @@ static void JS_EnabledStatisticsAllDatabases(v8::FunctionCallbackInfo<v8::Value>
...
+
+static void JS_EnabledMetrics(v8::FunctionCallbackInfo<v8::Value> const& args) {
+  TRI_V8_TRY_CATCH_BEGIN(isolate)
+    v8::HandleScope scope(isolate);
+
+    TRI_GET_GLOBALS();
+    bool enabled = v8g->_server.getFeature<MetricsFeature>().exportAPI();
+    v8::Handle<v8::Value> result = v8::Boolean::New(isolate, enabled);
+    TRI_V8_RETURN(result);
+  TRI_V8_TRY_CATCH_END
```
jsteemann, line 233:
```suggestion
  TRI_V8_TRY_CATCH_BEGIN(isolate)
  v8::HandleScope scope(isolate);

  TRI_GET_GLOBALS();
  bool enabled = v8g->_server.getFeature<MetricsFeature>().exportAPI();
  v8::Handle<v8::Value> result = v8::Boolean::New(isolate, enabled);
  TRI_V8_RETURN(result);
  TRI_V8_TRY_CATCH_END
```

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
```diff
@@ -1066,9 +1066,10 @@
       var dlType;
       if (type === 'csv') {
         dlType = 'text/csv; charset=utf-8';
-      }
-      if (type === 'json') {
+      } else if (type === 'json') {
         dlType = 'application/json; charset=utf-8';
+      } else if (type === 'text') {
+        dlType = 'text/html; charset=utf8';
```
jsteemann, line 1072: If this is for our metrics out, I think the correct Content-Type would be `text/plain`, not `text/html`.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
```diff
@@ -1090,7 +1090,12 @@
       var self = event.currentTarget;
 
       var url;
-      var doc = $(self).attr('id').substr(4);
+      var doc;
+      if ($(self).attr('id')) {
+        doc = $(self).attr('id').substr(4);
+      } else {
+        return;
+      }
```
jsteemann, line 1098: The following would be easier to read:
```suggestion
      if (!$(self).attr('id')) {
        return;
      }
      var doc = $(self).attr('id').substr(4);
```

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
hkernbach: Done, ty.

arangod/RestServer/Metrics.cpp
hkernbach: Well, yes true. The definition is a bit clumsy. I just wanted to stick to the examples given there. 
Also: Reading name + definition first makes more sense from a readers perspective than reading the type first.
Additionally, libraries I've found do want that extra space for parsing. If someone is not using prom UI, and wanted to use those libraries for parsing as well, they will run into the same problems I ran into. 

So I vote for keeping those changes ( at least the extra space - but IMHO also the order of attributes)

arangod/CMakeLists.txt
hkernbach: Without that addition, the target arango_metrics is not able to compile.

PR #13674, opened 2021-03-10
Added `--documents-per-batch` option to arangoexport.

jsteemann:
### Scope & Purpose

Added `--documents-per-batch` option to arangoexport.

This option allows to control the number of documents to be returned by each server-side batch. It can be used to limit the number of documents per batch when exporting collections with large documents.

#### Backports:

- [x] No backports required

### Testing & Verification

Link to Jenkins PR run:


mpoeter (approved):

goedderz:
One thing I noted: `arangoimport`, `arangodump` and `arangorestore` all have a `--batch-size` option. Just that in these cases, it is the size in bytes, not number of documents.

Now we probably can't do that here in a useful way. And I do feel that `--batch-size` is the natural name for this. Just wanted to put it on the table, maybe you do have another idea.

jsteemann:
@goedderz: we could rename it to `--documents-per-batch` here, because it is slightly different here than it is for all the other cases.
Although in the AQL query API the controlling attribute is actually called `batchSize`...

goedderz:
Yeah, it will be inconsistent with *something*, no matter what we do here. I think I like your `--documents-per-batch` proposal best, though.

jsteemann:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14304/

goedderz (approved):

mpoeter (approved):

PR #13682, opened 2021-03-11
Update OptimizeForArchitecture.cmake files and add support for Zen 3 CPUs

mpoeter:
### Scope & Purpose

On my machine I could not build with gcc/clang on Linux because of an issue in s2geometry due to the missing `__SSE2__` macro. The reason for that is that I have a new Ryzen 5950X CPU (Zen 3) which was not recognized by the `OptimizeForArchitecture.cmake` script.

This PR updates `OptimizeForArchitecture.cmake` ~to the latest version which also includes some updates for ARM (found here https://github.com/tunabrain/tungsten/blob/master/cmake/OptimizeForArchitecture.cmake)~ and adds support for Zen 3 CPUs.
For some reason the linked version does not build correctly. It would probably make sense to investigate this further, but due to lack of time I have reversed these changes and added the Zen3 support to the current version; will get back to this at a later point.

#### Backports:

- [x] No backports required

### Testing & Verification


dothebart (approved):

mpoeter:
Blue except for unrelated test errors.

Issue #13665, opened 2021-03-08
Regression between mmfiles and rocksdb when using indexes 


AoJ:
## My Environment

* __ArangoDB Version__:        3.6.9-3.7.8
* __Storage Engine__:             RocksDB
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:   Manual Start
* __Configuration__:               default configuration
* __Infrastructure__:               own
* __Operating System__:        Debian 9.13
* __Total RAM in your machine__:   16gb
* __Disks in use__:        SSD
* __Used Package__:              official arangodb3-linux-3.6.9.tar.gz

## Component, Query & Data

__Affected feature__:
AQL query on rocksdb engine

__AQL query (if applicable)__:
```
Query String (164 chars, cacheable: false):
 FOR a IN test1 FILTER a._ci == 'vCollect/28692498-f5cf-44b2-87c0-d0448070489e' FILTER a.netDev !=
 true AND a.ipLong > 172125031 AND a.ipLong < 172126032 RETURN a.ip
```

__AQL profile__:
```
Execution plan:
 Id   NodeType          Calls   Items   Runtime [s]   Comment
  1   SingletonNode         1       1       0.00000   * ROOT
  9   IndexNode             1       0       0.00007     - FOR a IN test1   /* hash index scan, projections: `ip`, `netDev`, `ipLong` */    FILTER (((a.`netDev` != true) && (a.`ipLong` > 172125031)) && (a.`ipLong` < 172126032))   /* early pruning */
  7   CalculationNode       1       0       0.00001       - LET #5 = a.`ip`   /* attribute expression */   /* collections used: a : test1 */
  8   ReturnNode            1       0       0.00000       - RETURN #5

Indexes used:
 By   Name                      Type   Collection   Unique   Sparse   Selectivity   Fields             Ranges
  9   idx_1692009215052218368   hash   test1        false    false        33.33 %   [ `_ci`, `mac` ]   (a.`_ci` == "vCollect/28692498-f5cf-44b2-87c0-d0448070489e")

Optimization rules applied:
 Id   RuleName
  1   move-calculations-up
  2   move-filters-up
  3   move-calculations-up-2
  4   move-filters-up-2
  5   use-indexes
  6   remove-filter-covered-by-index
  7   remove-unnecessary-calculations-2
  8   reduce-extraction-to-projection
  9   move-filters-into-enumerate

Query Statistics:
[...]

PR #13686, opened 2021-03-12
fix issue #13665: Improve index selection when there are multiple candidate indexes

jsteemann:
### Scope & Purpose

Fix issue #13665: Improve index selection when there are multiple candidate indexes

This PR improves index selection when there are multiple candidate indexes with common prefixes.

#### Backports:

#### Related Information

- [x] GitHub issue / Jira ticket number: https://github.com/arangodb/arangodb/issues/13665

### Testing & Verification

- [x] This change is already covered by existing tests, such as *aql-projections*.
  - [x] Added new **integration tests** (aql-index-choice.js)


arangod/Indexes/SortedIndexAttributeMatcher.cpp
```diff
@@ -388,19 +417,11 @@ Index::SortCosts SortedIndexAttributeMatcher::supportsSortCondition(
       if (costs.coveredAttributes >= sortCondition->numAttributes()) {
         // sort is fully covered by index. no additional sort costs!
         // forward iteration does not have high costs
-        costs.estimatedCosts = itemsInIndex * 0.001;
-        if (idx->isPersistent() && sortCondition->isDescending()) {
-          // reverse iteration has higher costs than forward iteration
```
mpoeter, line 393: Why was this removed? Is this statement no longer true?

tests/js/server/aql/aql-index-choice.js
```diff
@@ -0,0 +1,513 @@
...
+        let nodes = AQL_EXPLAIN(q).plan.nodes;
+        assertEqual(1, nodes.filter((n) => n.type === 'IndexNode').length);
+        assertEqual(0, nodes.filter((n) => n.type === 'SortNode').length);
+        let indexNode = nodes.filter((n) => n.type === 'IndexNode')[0];
+        assertEqual([], indexNode.projections);
+        assertFalse(indexNode.indexCoversProjections);
+        assertEqual(1, indexNode.indexes.length);
+        let index = indexNode.indexes[0];
+        assertEqual("persistent", index.type);
+        assertEqual(["uid", "dt"], index.fields);
```
mpoeter, line 131: Why would be pick the index with two fields in this scenario? `dt` is not used, neither in a filter nor in a projection.

tests/js/server/aql/aql-index-choice.js
```diff
@@ -0,0 +1,513 @@
...
+        let nodes = AQL_EXPLAIN(q).plan.nodes;
+        assertEqual(1, nodes.filter((n) => n.type === 'IndexNode').length);
+        assertEqual(0, nodes.filter((n) => n.type === 'SortNode').length);
+        let indexNode = nodes.filter((n) => n.type === 'IndexNode')[0];
+        assertEqual(["uid"], indexNode.projections);
+        assertTrue(indexNode.indexCoversProjections);
+        assertEqual(1, indexNode.indexes.length);
+        let index = indexNode.indexes[0];
+        assertEqual("persistent", index.type);
+        assertEqual(["uid", "dt"], index.fields);
```
mpoeter, line 153: Why would be pick the index with two fields in this scenario? `dt` is not used, neither in a filter nor in a projection.

tests/js/server/aql/aql-index-choice.js
```diff
@@ -0,0 +1,513 @@
...
+      db[cn].ensureIndex({ type: "persistent", fields: ["uid"] });
+      
+      [
+        `FOR doc IN ${cn} FILTER doc.uid == 1234 RETURN doc`,
+        `FOR doc IN ${cn} FILTER doc.uid > 1234 RETURN doc`,
+        `FOR doc IN ${cn} FILTER doc.uid >= 1234 RETURN doc`
+      ].forEach((q) => {
+        let nodes = AQL_EXPLAIN(q).plan.nodes;
+        assertEqual(1, nodes.filter((n) => n.type === 'IndexNode').length);
+        assertEqual(0, nodes.filter((n) => n.type === 'SortNode').length);
```
mpoeter, line 81: Why the check that no SortNode is found?

tests/js/server/aql/aql-index-choice.js
```diff
@@ -0,0 +1,513 @@
...
+        assertEqual(1, nodes.filter((n) => n.type === 'SortNode').length);
+        let indexNode = nodes.filter((n) => n.type === 'IndexNode')[0];
+        assertEqual([], indexNode.projections);
+        assertFalse(indexNode.indexCoversProjections);
+        assertEqual(1, indexNode.indexes.length);
+        let index = indexNode.indexes[0];
+        assertEqual("persistent", index.type);
+        assertEqual(["uid", "dt"], index.fields);
+      });
+    },
```
mpoeter, line 425: How about another test case where we filter on `dt` and sort on `uid`? That would be covered by the index, so we should not have a SortNode.

tests/js/server/aql/aql-index-choice.js
```diff
@@ -0,0 +1,513 @@
...
+  };
+}
```
mpoeter, line 468: How about a test where we have two indexes, one one `uid` and one on `dt`; in the query we filter on `uid` and sort on `dt` or vice versa. Which filter should be picked? The one that covers the filter or the one that covers the sort?

arangod/Indexes/SortedIndexAttributeMatcher.cpp
jsteemann: Added this back.

tests/js/server/aql/aql-index-choice.js
jsteemann: When all other things are equal, we will always pick the index with more attributes.

tests/js/server/aql/aql-index-choice.js
jsteemann: Removed this superfluous check :+1:

tests/js/server/aql/aql-index-choice.js
jsteemann: Added

tests/js/server/aql/aql-index-choice.js
jsteemann: Added

mpoeter:
How about adding a few more tests with `COLLECT` instead of `SORT`? With the index the optimizer should always pick the sorted executor.

jsteemann:
@mpoeter : ok, will add such tests!

PR #13696, opened 2021-03-12
Retry request when connection from pool stale

neunhoef:
If an `arangod` in the cluster is restarted (as for example dbservers on
hotbackup restore), all cluster-internal TCP connections to it break.
However, some connection objects in other instances might still linger
in the connection pool for some time. In this case the next network
request can run into a spurious error, despite the fact that the
server in question is already completely up and running again.

This PR implements a retry in exactly this case, only if a connection
comes from the pool and immediately runs into a ConnectionClosed error
we do a single retry.

### Scope & Purpose

#### Backports:

- [ ] No backports required
- [*] Backports required for: the problem is also in 3.6 and 3.7, it is
  not decided yet if we will backport it.

### Testing & Verification


maierlars (approved):

maierlars (approved):

neunhoef:
Jenkins run is "only" a shutdown hanger in a coordinator, will investigate, but we can still merge this.

PR #13699, opened 2021-03-12
Put Sync/LatestId into hotbackup

neunhoef:
When a hotbackup is restored to a cluster which is a lot "younger" than
the one where it was taken, the entry Sync/LatestID in the agency in the
new cluster is a lot smaller than the one in the old cluster. Therefore,
the unique key generation does no longer work well. Therefore, it makes
sense to put the Sync/LatestID into the hotbackup and restore it with a
hotbackup restore. This solves the key generation problem at least in
this hotbackup related case.

This PR implements this.

Old hotbackups can still be restored, even if they do not contain
Sync/LatestID.

### Scope & Purpose

#### Backports:

- [*] Backports required for: 3.6 and 3.7

### Testing & Verification

- [*] This change is already covered by existing tests, such as *(please describe tests)*.
       The hotbackup test was extended to check for the LatestID in the hotbackup.


tests/Cluster/ClusterInfoTest.cpp
```diff
@@ -120,3 +121,53 @@ TEST_F(ClusterInfoTest, testServerAliasExists) {
   }
 }
 
+TEST_F(ClusterInfoTest, plan_will_provide_latest_id) {
+  auto &cache {server.getFeature<ClusterFeature>().agencyCache()};
+  auto [acb, index] {cache.read({AgencyCommHelper::path("Sync/LatestID")})};
+  auto expectedLatestId {acb->slice().at(0).get("arango").get("Sync").get("LatestID").getInt()};
```
neunhoef, line 127: Note that this can throw, which will then break the test. So it is OK here, but in actual production code we need to be more careful.

neunhoef: The variant `get(std::vector<std::string>)` does not throw if the top level is an object, though, it simply returns a `NoneSlice` if someway on the path an attribute is not found.

tests/Cluster/ClusterInfoTest.cpp
```diff
@@ -120,3 +121,53 @@ TEST_F(ClusterInfoTest, testServerAliasExists) {
...
+  auto builder {std::make_shared<VPackBuilder>()};
+  auto result {ci.agencyPlan(builder)};
+  ASSERT_TRUE(result.ok());
+  ASSERT_EQ(builder->slice().at(0).get("arango").get("Sync").get("LatestID").getInt(), expectedLatestId);
+}
+
+TEST_F(ClusterInfoTest, replan_will_use_latest_id) {
+  VPackBuilder builder;
+  VPackParser parser(builder);  // add json as vpack to the builder
+  std::string test_plan{R"([{"arango":
```
neunhoef, line 138: I think a much smaller `Plan` could be used here, since the contents of the plan does not actually matter. As far as I know the top level components `Analyzers`, `Databases`, `Collections` and `Views` must exist (whatever `replanAgency` needs), but the gory details do not matter.

neunhoef (commented):
LGTM

tests/js/server/dump/dump-modify.js
```diff
@@ -488,6 +488,20 @@ function dumpTestSuite () {
...
+/// @brief test latestId
+////////////////////////////////////////////////////////////////////////////////
+
+    testLatestId : function () {
+      if (arango.getRole() === "COORDINATOR") {
+        // Only executed in the cluster
+        let next = JSON.parse(db._connection.POST("/_admin/execute?returnAsJSON=true", "return global.ArangoAgency.uniqid(100000000)"));
+        assertTrue(next < 100000000, "expected next uniqid in agency to be less than 100000000, not " + next);
+        next = JSON.parse(db._connection.POST("/_admin/execute?returnAsJSON=true", "return global.ArangoAgency.uniqid(100000000)"));
+        assertTrue(next > 100000000, "expected next uniqid in agency to be greater than 150000000, not " + next);
```
jsteemann, line 501:
```suggestion
        assertTrue(next > 100000000, "expected next uniqid in agency to be greater than 100000000, not " + next);
```

jsteemann (approved):

PR #13703, opened 2021-03-12
 Log actual signal instead of "control-c" and also include sender process id

mpoeter:
### Scope & Purpose

Log the actual signal instead of "control-c" and also include the process id of the process that sent the signal.

#### Backports:

- [x] No backports required

### Testing & Verification


jsteemann:
LGTM, but as this PR changes a very central log message text (the one for controlled shutdown), we should check internally if anyone or any tools relies on the old message text somehow.

jsteemann (approved):

dothebart (approved):

goedderz (approved):

PR #13709, opened 2021-03-14
Fix slower-than-necessary arangoimport behavior:

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13704

arangoimport has a built-in rate limiter, which can be useful for importing
data with a somewhat constant rate. However, it is enabled by default and
limits imports to 1MB per second. These settings are not useful.

This change turns the rate limiting off by default, and sets the default
chunk size to 8MB (up from 1MB) as well. This means that arangoimport will
send larger batches to the server by default. The already existing `--batch-size`
option can be used to control the maximum size of each batch.

The new parameter `--auto-rate-limit` can now be used to toggle rate limiting.
It defaults to off, whereas previously rate limiting was enabled by default
unless `--batch-size` was specified when arangoimport was invoked.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/634

### Testing & Verification


fceller (approved):
LGTM

PR #13710, opened 2021-03-15
Compact collection and index ranges in background

jsteemann:
### Scope & Purpose

When dropping a collection or an index with a larger amount of documents, the
key range for the collection/index in RocksDB gets compacted. Previously, the
compaction was running in foreground and thus would block the deletion operations.
Now, the compaction is running in background, so that the deletion operations
can return earlier.
The maximum number of compaction jobs that are executed in background can be
configured using the new startup parameter `--rocksdb.max-parallel-compactions`,
which defaults to 2.

Performance implications of this change still need to be determined.

#### Backports:

- [x] Backports required for: *3.7*: https://github.com/arangodb/arangodb/pull/13711

### Testing & Verification

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -84,6 +84,7 @@
 #include "RocksDBEngine/RocksDBV8Functions.h"
 #include "RocksDBEngine/RocksDBValue.h"
 #include "RocksDBEngine/RocksDBWalAccess.h"
+#include "Scheduler/SchedulerFeature.h"
 #include "Transaction/Context.h"
 #include "Transaction/Manager.h"
 #include "Transaction/Options.h"
@@ -178,6 +179,7 @@ RocksDBEngine::RocksDBEngine(application_features::ApplicationServer& server)
       _maxTransactionSize(transaction::Options::defaultMaxTransactionSize),
       _intermediateCommitSize(transaction::Options::defaultIntermediateCommitSize),
       _intermediateCommitCount(transaction::Options::defaultIntermediateCommitCount),
+      _maxParallelCompactions(2),
       _pruneWaitTime(10.0),
       _pruneWaitTimeInitial(180.0),
       _maxWalArchiveSizeLimit(0),
@@ -200,7 +202,8 @@ RocksDBEngine::RocksDBEngine(application_features::ApplicationServer& server)
 #else
       _createShaFiles(false),
 #endif
-      _lastHealthCheckSuccessful(false) {
+      _lastHealthCheckSuccessful(false),
+      _runningCompactions(0) {
   server.addFeature<RocksDBOptionFeature>();
 
   startsAfter<BasicFeaturePhaseServer>();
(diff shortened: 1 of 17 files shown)


arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -1325,6 +1333,90 @@ RecoveryState RocksDBEngine::recoveryState() noexcept {
...
+      _pendingCompactions.pop_front();
+      // set it to running already, so that concurrent callers of this method
+      // will not kick off additional jobs
+      ++_runningCompactions;
+    }
+    
+    LOG_TOPIC("6ea1b", TRACE, Logger::ENGINES) 
+          << "scheduling compaction for execution";
+    
+    bool queued = scheduler->queue(arangodb::RequestLane::CLIENT_SLOW, [this, bounds]() {
```
mpoeter, line 1383: Is it guaranteed that `this` lives long enough? In other words - can it happen that this lambda gets executed after the `RocksDBEngine` instance has been deleted?

jsteemann: That is a good point :thinking:

jsteemann: Addressed.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -938,6 +938,25 @@ void RocksDBEngine::stop() {
...
+      }
+    }
+      
+    if (!compactionWarning) {
+      LOG_TOPIC("9cbfd", INFO, Logger::ENGINES) << "waiting for compaction jobs to finish...";
+      compactionWarning = true;
+    }
+    // unfortunately there is not much we can do except waiting for
+    // RocksDB's compaction job(s) to finish.
+    std::this_thread::yield();
```
mpoeter, line 958: I would rather use a sleep instead of a `yield`. The number of compaction jobs probably less than the number of available cores, so there is a good chance that there are no other threads waiting to be scheduled and the OS scheduler might immediately reschedule this thread...

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -1381,26 +1400,32 @@ void RocksDBEngine::processCompactions() {
...
-        rocksdb::Slice b = bounds.start(), e = bounds.end();
-        _db->CompactRange(opts, bounds.columnFamily(), &b, &e);
-      } catch (std::exception const& ex) {
-        LOG_TOPIC("a4c42", WARN, Logger::ENGINES) 
-              << "compaction for range " << bounds 
-              << " failed with error: " << ex.what();
-      } catch (...) {
-        // whatever happens, we need to count down _runningCompactions in all cases
-      }
+      if (server().isStopping()) {
```
mpoeter, line 1403: `server()` is a method of `RocksDBEngine`, so that does not really solve the problem. I guess we need to capture a `shared_from_this` here.

jsteemann: I think that is fine, because the RocksDBEngine ApplicationFeature now blocks in its shutdown until all running compaction jobs have finished.
So whenever we execute the callback, `this` is guaranteed to be still alive and kicking.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Ok, will change it.

jsteemann: Done

arangod/RocksDBEngine/RocksDBEngine.cpp
mpoeter: You are right, my bad. I realized this a few minutes after I wrote that comment, but I was already AFK then.

jsteemann: No worries, good that you mentioned the issue, because originally there _was_ a problem! :+1:

PR #13711, opened 2021-03-15
Compact collection and index ranges in background

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13710

When dropping a collection or an index with a larger amount of documents, the
key range for the collection/index in RocksDB gets compacted. Previously, the
compaction was running in foreground and thus would block the deletion operations.
Now, the compaction is running in background, so that the deletion operations
can return earlier.
The maximum number of compaction jobs that are executed in background can be
configured using the new startup parameter `--rocksdb.max-parallel-compactions`,
which defaults to 2.

Performance implications of this change still need to be determined.

#### Backports:

- [x] No backports required.

### Testing & Verification


fceller (approved):
LGTM

PR #13719, opened 2021-03-15
Don't store selectivity estimate values for newly created system collections

jsteemann:
### Scope & Purpose

Don't store selectivity estimate values for newly created system collections

Not storing the estimates has a benefit especially for the `_statistics`
system collections, which are written to periodically even on otherwise
idle servers. In this particular case, the actual statistics data was way
smaller than the writes caused by the index estimate values, causing a
disproportional overhead just for maintaining the selectivity estimates.
The change now turns off the selectivity estimates for indexes in all newly
created system collections, and for new user-defined indexes of type
"persistent", "hash" or "skiplist", there is now an attribute "estimates"
which can be set to `false` to disable the selectivity estimates for the index.
The attribute is optional. Not setting it will lead to the index being
created with selectivity estimates, so this is a downwards-compatible change
for user-defined indexes.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: to be done

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in recovery, shell_server)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,7 +11,7 @@ devel
   The change now turns off the selectivity estimates for indexes in all newly 
   created system collections, and for new user-defined indexes of type
   "persistent", "hash" or "skiplist", there is now an attribute "estimates"
-  which can be set to `true` to disable the selectivity estimates for the index.
+  which can be set to `false` to disable the selectivity estimates for the index.
   The attribute is optional. Not setting it will lead to the index being
   created with selectivity estimates, so this is a downwards-compatible change
   for user-defined indexes.
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,27 @@
 devel
 -----
 
+* Don't store selectivity estimate values for newly created system collections.
+
+  Not storing the estimates has a benefit especially for the `_statistics`
+  system collections, which are written to periodically even on otherwise
+  idle servers. In this particular case, the actual statistics data was way
+  smaller than the writes caused by the index estimate values, causing a
+  disproportional overhead just for maintaining the selectivity estimates.
+  The change now turns off the selectivity estimates for indexes in all newly 
+  created system collections, and for new user-defined indexes of type
+  "persistent", "hash" or "skiplist", there is now an attribute "estimates"
+  which can be set to `true` to disable the selectivity estimates for the index.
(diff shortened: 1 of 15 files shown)


CHANGELOG
```diff
@@ -1,12 +1,27 @@
...
+
+  Not storing the estimates has a benefit especially for the `_statistics`
+  system collections, which are written to periodically even on otherwise
+  idle servers. In this particular case, the actual statistics data was way
+  smaller than the writes caused by the index estimate values, causing a
+  disproportional overhead just for maintaining the selectivity estimates.
+  The change now turns off the selectivity estimates for indexes in all newly 
+  created system collections, and for new user-defined indexes of type
+  "persistent", "hash" or "skiplist", there is now an attribute "estimates"
+  which can be set to `true` to disable the selectivity estimates for the index.
```
mpoeter, line 14: The attribute is just called "estimates" and a value of "true" _disables_ the estimates? I would find that very counter intuitive! I suppose this should be:
```suggestion
  which can be set to `false` to disable the selectivity estimates for the index.
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -130,10 +149,11 @@ bool ClusterIndex::hasSelectivityEstimate() const {
...
     return _indexType == Index::TRI_IDX_TYPE_PRIMARY_INDEX ||
            _indexType == Index::TRI_IDX_TYPE_EDGE_INDEX ||
-           _indexType == Index::TRI_IDX_TYPE_HASH_INDEX ||
-           _indexType == Index::TRI_IDX_TYPE_SKIPLIST_INDEX ||
            _indexType == Index::TRI_IDX_TYPE_TTL_INDEX ||
-           _indexType == Index::TRI_IDX_TYPE_PERSISTENT_INDEX;
+           (_estimates && 
+            (_indexType == Index::TRI_IDX_TYPE_HASH_INDEX ||
+             _indexType == Index::TRI_IDX_TYPE_SKIPLIST_INDEX ||
+              _indexType == Index::TRI_IDX_TYPE_PERSISTENT_INDEX));
```
mpoeter, line 156:
```suggestion
             _indexType == Index::TRI_IDX_TYPE_PERSISTENT_INDEX));
```

mpoeter (approved):

PR #13722, opened 2021-03-16
validate metrics API output against promtool

jsteemann:
### Scope & Purpose

This PR adds a test to validate the output of our metrics API against Prometheus' `promtool` command.
The test intentionally only executes on Linux, and only if the `PROMTOOL_PATH` environment variable is set to the path containing the `promtool` executable. If `PROMTOOL_PATH` is set, but the executable cannot be found, the test will error out.
The test also requires `sh` to be a shell that supports input/output redirection, and `true` to be an executable that returns exit code 0 (we use `sh -c true` as a test to check the shell functionality).

Example invocation:
```
PROMOTOOL_PATH=/home/jsteemann/Downloads/prometheus/ scripts/unittest shell_client --test shell-promtool.js --cluster true
```

The promtool executable is part of a Prometheus distribution, and can be found in each Prometheus release distributable (e.g. in https://github.com/prometheus/prometheus/releases/tag/v2.25.1). 

The promtool source code can be found here (should not be required, just putting it here for reference): https://github.com/prometheus/prometheus/tree/main/cmd/promtool

In order for this test to make sense, we will need to make `promtool` available in our (Linux) CI and set the `PROMTOOL_PATH` environment variable to the directory that contains the `promtool` executable.

#### Backports:

- [x] No backports required

### Testing & Verification

  - [x] Added new **integration tests** (see above)

### DIFF
--- a/tests/js/client/shell/shell-promtool.js
+++ b/tests/js/client/shell/shell-promtool.js
@@ -1,0 +1,109 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global assertTrue, assertFalse, assertEqual, arango, fail */
+
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2021 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
+///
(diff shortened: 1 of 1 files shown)


neunhoef (approved):
LGTM

PR #13724, opened 2021-03-16
In a cluster, don't create collections `_statistics*` on DB servers

jsteemann:
### Scope & Purpose

These collections should only be created by the coordinator, and should translate into 2 shards each on DB servers. But there shouldn't be shards named `_statistics*` on DB servers.

#### Backports:

- [x] No backports required

### Testing & Verification


tests/js/client/shell/shell-statistics-collection-cluster.js
```diff
@@ -0,0 +1,104 @@
...
+///
+/// @author Jan Steemann
+////////////////////////////////////////////////////////////////////////////////
+
+const jsunity = require('jsunity');
+const arangodb = require('@arangodb');
+const db = arangodb.db;
+const request = require("@arangodb/request");
+
+let getServers = function(role) {
```
hkernbach, line 32: I've seen things like this all over our code. We should move that to a JS helper file/class.

neunhoef (approved):
LGTM

tests/js/client/shell/shell-statistics-collection-cluster.js
```diff
@@ -0,0 +1,104 @@
...
+    },
+
+    testStatisticsCollectionsOnDBServer: function () {
+      let dbservers = getServers("dbserver");
+      dbservers.forEach((server) => {
+        let result = request({ method: "GET", url: server.url + "/_api/collection", body: {} });
+        assertEqual(200, result.json.code);
+        let shards = result.json.result;
+        assertTrue(Array.isArray(shards));
+        // verify that there are no shards named _statistics, _statistics15, _statisticsRaw
```
hkernbach, line 90: Where's the actual verify here? Maybe I do not get it, but I am not able to see it. 
L94 does shard id checks.
L96 checks then the appearance of `s` in `shardNames`?
Ah and while reading it I see that there are several `s`es defined with different meanings. So maybe change naming here ..

hkernbach: Understood now. The test is correct. Nevertheless `s` confused me, would change the `s` -> `collectionName`

tests/js/client/shell/shell-statistics-collection-cluster.js
```diff
@@ -0,0 +1,104 @@
...
+        collections.forEach((s) => {
+          let shardNames = shards.map((s) => s.name);
+          shardNames.forEach((s) => {
+            assertMatch(/^s[0-9]+$/, s);
+          });
+          assertEqual(-1, shardNames.indexOf(s));
+        });
```
hkernbach, line 97:
```suggestion
        collections.forEach((collectionName) => {
          let shardNames = shards.map((s) => s.name);
          shardNames.forEach((s) => {
            assertMatch(/^s[0-9]+$/, s);
          });
          assertEqual(-1, shardNames.indexOf(collectionName));
        });
```

PR #13729, opened 2021-03-16
Fix duplicate leaving of V8 contexts when returning streaming cursors.

jsteemann:
### Scope & Purpose

The `exitContext` call done on query shutdown could previously try to exit
the V8 context multiple times, which would cause undefined behavior. Now
we are tracking if we already left the context to prevent duplicate invocation.

Such error was triggered by running `scripts/unittest shell_server` when running the tests with ASan. With the change, the error is gone.

#### Backports:

- [x] No backports required

### Testing & Verification


goedderz (approved):

neunhoef (approved):
LGTM

PR #13730, opened 2021-03-16
Fixed two bogus messages about hotbackup restore:

jsteemann:
### Scope & Purpose

Fixed two bogus messages about hotbackup restore:

- Coordinators unconditionally logged the message "Got a hotbackup restore
  event, getting new cluster-wide unique IDs..." on shutdown. This was not
  necessarily related to a hotbackup restore.
- DB servers unconditionally logged the message "Strange, we could not
  unregister the hotbackup restore callback." on shutdown, although this was
  meaningless.

These messages were only introduced one day ago in devel, so there is no need to backport the change.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/Cluster/ClusterFeature.cpp
```diff
@@ -685,10 +689,14 @@ void ClusterFeature::stop() {
     return;
   }
 
-  if (!_agencyCallbackRegistry->unregisterCallback(_hotbackupRestoreCallback)) {
-    LOG_TOPIC("84152", INFO, Logger::BACKUP) << "Strange, we could not "
-      "unregister the hotbackup restore callback.";
+#ifdef USE_ENTERPRISE
+  if (_hotbackupRestoreCallback != nullptr) {
+    if (!_agencyCallbackRegistry->unregisterCallback(_hotbackupRestoreCallback)) {
+      LOG_TOPIC("84152", INFO, Logger::BACKUP) << "Strange, we could not "
```
maierlars, line 695: Is this even worth logging? The use can nothing do about it and we neither and nobody cares actually.

arangod/Cluster/ClusterFeature.cpp
```diff
@@ -645,13 +645,16 @@ void ClusterFeature::start() {
   AsyncAgencyCommManager::INSTANCE->setSkipScheduler(false);
   ServerState::instance()->setState(ServerState::STATE_SERVING);
 
+#ifdef USE_ENTERPRISE
   // If we are on a coordinator, we want to have a callback which is called
   // whenever a hotbackup restore is done:
   if (role == ServerState::ROLE_COORDINATOR) {
     auto hotBackupRestoreDone = [this](VPackSlice const& result) -> bool {
```
maierlars, line 652: I know its not related to that PR but does the ClusterFeature and the ClusterInfo outlive this lambda?

maierlars (approved):

arangod/Cluster/ClusterFeature.cpp
jsteemann: The lambda is registered as an AgencyCallback and stored in the ClusterFeature's `_hotbackupRestoreCallback` instance member. I guess this means the lambda will not outlive the ClusterFeature, at least.

arangod/Cluster/ClusterFeature.cpp
jsteemann, line 695:
```suggestion
      LOG_TOPIC("84152", DEBUG, Logger::BACKUP) << "Strange, we could not "
```

jsteemann: I changed the log level to debug here.

PR #13732, opened 2021-03-17
Bug fix 3.6/save unique id counter in agency with each hotback

ignacionr:
### Scope & Purpose

When a hotbackup is restored to a cluster which is a lot "younger" than
the one where it was taken, the entry Sync/LatestID in the agency in the
new cluster is a lot smaller than the one in the old cluster. Therefore,
the unique key generation does no longer work well. Therefore, it makes
sense to put the Sync/LatestID into the hotbackup and restore it with a
hotbackup restore. This solves the key generation problem at least in
this hotbackup related case.

This PR is a back port of the implementation of that, from #13699 .

Old hotbackups can still be restored, even if they do not contain
Sync/LatestID.

#### Backports:

- [x] This is a backport of #13699 into 3.6

#### Related Information

- [x] GitHub issue / Jira ticket number: TC-18

### Testing & Verification

Link to Jenkins PR run:

### DIFF
--- a/arangod/Cluster/ClusterFeature.cpp
+++ b/arangod/Cluster/ClusterFeature.cpp
@@ -565,9 +565,11 @@ void ClusterFeature::start() {
   // whenever a hotbackup restore is done:
   if (role == ServerState::ROLE_COORDINATOR) {
     auto hotBackupRestoreDone = [this](VPackSlice const& result) -> bool {
-      LOG_TOPIC("12636", INFO, Logger::BACKUP) << "Got a hotbackup restore "
-        "event, getting new cluster-wide unique IDs...";
-      this->_clusterInfo->uniqid(1000000);
+      if (!server().isStopping()) {
+        LOG_TOPIC("12636", INFO, Logger::BACKUP) << "Got a hotbackup restore "
+          "event, getting new cluster-wide unique IDs...";
+        this->_clusterInfo->uniqid(1000000);
+      }
       return true;
     };
     _hotbackupRestoreCallback =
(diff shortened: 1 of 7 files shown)


arangod/Cluster/ClusterFeature.cpp
```diff
@@ -560,11 +560,40 @@ void ClusterFeature::start() {
...
+        LOG_TOPIC("12636", INFO, Logger::BACKUP) << "Got a hotbackup restore "
+          "event, getting new cluster-wide unique IDs...";
+        this->_clusterInfo->uniqid(1000000);
+      }
+      return true;
+    };
+    _hotbackupRestoreCallback =
+      std::make_shared<AgencyCallback>(
+        server(), "Sync/HotBackupRestoreDone", hotBackupRestoreDone, true, false);
+    auto r =_agencyCallbackRegistry->registerCallback(_hotbackupRestoreCallback);
```
neunhoef, line 578: Let's use `bool` here instead of `auto` to make it easier to read.

arangod/Cluster/ClusterFeature.cpp
```diff
@@ -560,11 +560,40 @@ void ClusterFeature::start() {
...
+           "to problems after a restore!";
+    }
+  }
 }
 
 void ClusterFeature::beginShutdown() { ClusterComm::instance()->disable(); }
 
 void ClusterFeature::stop() {
+  #ifdef USE_ENTERPRISE
+  if (_hotbackupRestoreCallback != nullptr && !_agencyCallbackRegistry->unregisterCallback(_hotbackupRestoreCallback)) {
```
neunhoef, line 591: We need to add a check if this is a coordinator here as well, otherwise dbservers and agents issue a strange info message. This was forgotten in the `devel` PR, too.

neunhoef (changes requested):
See comments, these changes are all minor.

arangod/Cluster/ClusterFeature.cpp
ignacionr: Done.

arangod/Cluster/ClusterFeature.cpp
ignacionr: Done.

ignacionr:
Jenkins had a transient hiccup; the Windows part was re-run successfully. See http://172.16.10.101:8080/job/arangodb-matrix-pr-windows/7899/

goedderz (approved):

neunhoef (approved):
LGTM

PR #13734, opened 2021-03-17
Fix return type of ManagedDirectory::File::offset().

mpoeter:
### Scope & Purpose

On Windows `TRI_read_return_t` is defined as `int`, so previously `ManagedDirectory::File::offset()` was broken for files > 2GB.

This manifested itself as missing progress reports in arangoimport,

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Fix progress reporting for arangoimport with large files on Windows.
+  Previously, progress was only reported for the first 2GB of data due to a
+  int overflow.
+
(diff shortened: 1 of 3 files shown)


dothebart:
hm, wasn't the TRI-type the one used by the windows API functions in the game?

mpoeter:
@dothebart what do you mean? `File::offset` uses `TRI_SEEK` which on Windows resolves to `::_lseeki64` which returns `__int64`.

dothebart:
hm, shouldn't all these functions be used, and `TRI_read_return_t` rather become that type then?

PR #13736, opened 2021-03-17
fix a race in the agency-restart test

jsteemann:
### Scope & Purpose

Fixed a race in `agency-restart` tests, that assumed no compaction of data had happened after the restart. This assumption is not safe, because compaction can be started before the recovery code in JavaScript can disable it.
This PR only adjusts a test, so there is intentionally no CHANGELOG entry for it.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *agency-restart*.


neunhoef (approved):
LGTM

PR #13737, opened 2021-03-17
improve error message when dropping a follower

jsteemann:
### Scope & Purpose

Improve error message in case a truncate operation drops a follower.
The error message now also contains the database name in addition to the shard name, plus the reason for why the follower was dropped.
This is a change that only affects an error message, so there is intentionally no CHANGELOG entry for it.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/Transaction/Methods.cpp
```diff
@@ -1880,16 +1880,18 @@ Future<OperationResult> transaction::Methods::truncateLocal(std::string const& c
              responses[i].get().statusCode() == fuerte::StatusOK);
         if (!replicationWorked) {
           auto const& followerInfo = collection->followers();
+          LOG_TOPIC("0e2e0", WARN, Logger::REPLICATION)
+              << "truncateLocal: dropping follower " << (*followers)[i]
+              << " for shard" << collection->vocbase().name() << "/" << collectionName
```
neunhoef, line 1885: I suggest to put one more space behind the word "shard" here.

neunhoef (approved):
LGTM. See one small comment, no need to run jenkins again or me to look again after the change.

arangod/Transaction/Methods.cpp
jsteemann, line 1885:
```suggestion
              << " for shard " << collection->vocbase().name() << "/" << collectionName
```

jsteemann: Sure, thanks for the hint. I overlooked this initially.

hkernbach (approved):

PR #13744, opened 2021-03-17
Fix passing around an uninitialized struct member

jsteemann:
### Scope & Purpose

Fix passing around an uninitialized struct member

Fix potentially undefined behavior when creating a `CalculationTransactionContext` for an arangosearch analyzer. An uninitialized struct member was passed as an argument to its base class. This potentially had no observable effects, but should be fixed.

It seems that the affected `CalculationTransactionContext` struct does not exist in previous versions, so there is no need to backport the change.

#### Backports:

- [x] No backports required.

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtest*.


Dronplane (approved):
LGTM.
BTW what  tool has helped finding the issue? Or it was just noticed accidently?

jsteemann:
@Dronplane : I found this by accident, when I added a new member to `transaction::Options`. The new member was an `std::string`, and this caused an `std::bad_alloc` exception to be thrown when used that way.

PR #13750, opened 2021-03-18
Remove some leftover replication tailing clients from internal bookkeeping

jsteemann:
### Scope & Purpose

Remove some leftover replication tailing clients from internal bookkeeping once they are finished with getting a shard in sync. 
Previously, such replication clients could have been left in the internal bookkeeping until they were removed by the automatic garbage collection after 2 hours. While the client entries were still in the bookkeeping tables, they could prevent archived RocksDB WAL files from being dropped, leading to leader DB servers keeping more WAL files around than necessary.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,14 @@
 devel
 -----
 
+* Remove some leftover replication tailing clients from internal bookkeeping
+  once they are finished with getting a shard in sync. Previously, such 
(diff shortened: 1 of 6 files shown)


jsteemann:
Not ready yet because arangodump needs adjustment, too. It doesn't pass as clientId right now.

PR #13761, opened 2021-03-19
Fix coordinator blocked thread on query shutdown

neunhoef:
This PR addresses a problem which was found by a coordinator shutdown
hanger. It turned out to be the following problem:

If a Query::finalize fails during the transaction commit, then it throws
an exception which leaves the SharedQueryState in a state which seems to
be in the shutdown process but then waits for some event to happen which
never happens (because nobody ordered it). This leaves the thread
blocked on a condition variable. In the case of the coordinator shutdown
the StatisticsWorker had this problem and thus prevented the server
shutdown because the StatisticsFeature could not shut down.

### Scope & Purpose

*(Please describe the changes in this PR for reviewers - **mandatory**)*

- [*] :hankey: Bugfix (requires CHANGELOG entry)

#### Backports:

- [*] Backports required for: 3.7 and 3.6

### Testing & Verification

*(Please pick either of the following options)*

- [*] The behavior in this PR was *manually tested*
- [*] This PR adds tests that were used to verify all changes:
  - [*] Added new **integration tests** (e.g. in shell_server / shell_server_aql)

PR #13782, opened 2021-03-23
Bug fix 3.8/query stuck at shutdown

mchacki:
Backport of https://github.com/arangodb/arangodb/pull/13761

This PR addresses a problem which was found by a coordinator shutdown
hanger. It turned out to be the following problem:

If a Query::finalize fails during the transaction commit, then it throws
an exception which leaves the SharedQueryState in a state which seems to
be in the shutdown process but then waits for some event to happen which
never happens (because nobody ordered it). This leaves the thread
blocked on a condition variable. In the case of the coordinator shutdown
the StatisticsWorker had this problem and thus prevented the server
shutdown because the StatisticsFeature could not shut down.

### Scope & Purpose

#### Backports:

- [*] Backports required for: 3.7 and 3.6

### Testing & Verification

  - [*] Added new **integration tests** (e.g. in shell_server / shell_server_aql)


neunhoef (approved):
LGTM, please add a CHANGELOG entry.

mchacki:
Added a changlog entry

PR #13788, opened 2021-03-23
Docs: Correct description of streaming cursor

Simran-B:
Backport of #13497


Simran-B:
@goedderz I made a minor addition about MMFiles, but I'm not entirely sure if it's correct. Will all queries outright fail that involve the same collections as another with stream: true, even if both are read-only? Or is there some timeout involved?

goedderz:
@Simran-B Read-only queries don't interfere, as there can be multiple read-locks on the same collection. A write-lock however will prevent both read and write locks, and a write-lock cannot be obtained as long as there is any other lock, read or write, held on the collection.

IIRC the queries block until the lock can be obtained, plus there is a deadlock-detection on single-server that can abort queries if a deadlock is detected. But my knowledge is a little vague about this part and I don't know the details.

goedderz (approved):

PR #13791, opened 2021-03-23
Fix race in LogAppender::haveAppenders.

mpoeter:
### Scope & Purpose

Backport of #13789 

`haveAppenders` is called as part of audit logging. It accesses `_topics2appenders` and `_globalAppenders` but previously did not hold a lock while doing so.
This problem only showed up because previously we used `operator[]` which inserts a new entry if none is found and this caused a memory leak in the ASan tests.

- [x] :hankey: Bugfix

#### Backports:

- [x] No backports required

### Testing & Verification


jsteemann (approved):

KVS85:
Tests blue except unrelated failure.

PR #13792, opened 2021-03-23
fix cluster recovery test timeouts

dothebart:
### Scope & Purpose

set the http-timeout on the arangosh doing the `setup` for the testcase. By it aborting after 60s we don't wait 15 minutes for the coordinator to return. 

### DIFF
--- a/js/client/modules/@arangodb/testsuites/recovery_cluster.js
+++ b/js/client/modules/@arangodb/testsuites/recovery_cluster.js
@@ -121,9 +121,9 @@ function runArangodRecovery (params) {
   // arangosh has different name for parameter :(
   additionalTestParams['javascript.execute'] =  params.script;
   additionalTestParams['javascript.run-main'] = true;
-  additionalTestParams['server.request-timeout'] = '60';
 
   if (params.setup) {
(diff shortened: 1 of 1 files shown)


js/client/modules/@arangodb/testsuites/recovery_cluster.js
```diff
@@ -119,6 +121,7 @@ function runArangodRecovery (params) {
   // arangosh has different name for parameter :(
   additionalTestParams['javascript.execute'] =  params.script;
   additionalTestParams['javascript.run-main'] = true;
+  additionalTestParams['server.request-timeout'] = '60';
```
Dronplane, line 124: I guess we should do this only if params.setup is true? 
And wouldn't this introduce a problem if setup query will need to run morethan a minute? But the latter we could just document as a limitation for setup.

dothebart: yes, absolutely!

dothebart:
http://jenkins01.arangodb.biz:8080/job/arangodb-matrix-pr-linux/15963/

Dronplane (approved):
LGTM

PR #13815, opened 2021-03-24
Add documention for scheduler metrics

mpoeter:
### Scope & Purpose

Backport of #13811

#### Backports:

- [x] No backports required

### Testing & Verification


Documentation/Metrics/arangodb_scheduler_jobs_dequeued.yaml
```diff
@@ -9,5 +9,15 @@ exposedBy:
   - coordinator
   - dbserver
   - agent
+  - single
 description: |
-  Total number of jobs dequeued.
+  The total number of jobs dequeued from all scheduler queues.
```
neunhoef, line 14: I would have suggested to not document the legacy metrics at all, but I have no strong feelings. I would have suggested to remove this YAML file and use `DECLARE_LEGACY_GAUGE` such that the check script does not complain. If we leave it like this we have the strange situation that the manual has a duplication of the documentation but the v2 metrics do not have some of the entries.

Documentation/Metrics/arangodb_scheduler_jobs_done.yaml
```diff
@@ -9,5 +9,12 @@ exposedBy:
   - coordinator
   - dbserver
   - agent
+  - single
 description: |
-  Total number of queue jobs done.
+  The total number of queue jobs done.
```
neunhoef, line 14: Same here.

Documentation/Metrics/arangodb_scheduler_jobs_submitted.yaml
```diff
@@ -9,5 +9,12 @@ exposedBy:
   - coordinator
   - dbserver
   - agent
+  - single
 description: |
-  Total number of jobs submitted to the queue.
+  Total number of jobs submitted to the scheduler.
```
neunhoef, line 14: Same here.

neunhoef (changes requested):
LGTM, but see suggestion in comments. Not completely necessary, but this was so far our policy.

Documentation/Metrics/arangodb_scheduler_jobs_dequeued.yaml
mpoeter: Makes sense!

neunhoef (approved):
LGTM

PR #13598, opened 2021-02-22
Connection cache for arangosh

jsteemann:
### Scope & Purpose

* Add an experimental connection cache to arangosh, so that the number of
  connections and thus ephemeral TCP ports used can be reduced for pathologic
  use cases such as connect to A, connect to B, connect to A, connect to B, ...

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*
- [x] This change is already covered by existing tests, such as *replication_sync*.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/14083/

PR #13821, opened 2021-03-25
Feature 3.8/arangosh connection cache

dothebart:
### Scope & Purpose

    Add an experimental connection cache to arangosh, so that the number of
    connections and thus ephemeral TCP ports used can be reduced for pathologic
    use cases such as connect to A, connect to B, connect to A, connect to B, ...

backport for https://github.com/arangodb/arangodb/pull/13598


CHANGELOG
```diff
@@ -1,6 +1,9 @@
 v3.8.0-alpha.1 (XXXX-XX-XX)
 ---------------------------
 
+* arangosh may keep connections establinged and switch between them
```
mpoeter, line 4:
```suggestion
* arangosh may keep connections established and switch between them
```

mpoeter (approved):

PR #13869, opened 2021-03-29
Feature 3.8/auto check statistics metrics

mpoeter:
### Scope & Purpose

Backport of #13837

Add some assertions in maintainer mode that the metrics from the statistics feature are in sync with the new metrics used for documentation.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/Statistics/StatisticsFeature.cpp
+++ b/arangod/Statistics/StatisticsFeature.cpp
@@ -30,6 +30,7 @@
 #include "Basics/NumberOfCores.h"
 #include "Basics/PhysicalMemory.h"
 #include "Basics/StaticStrings.h"
+#include "Basics/StringUtils.h"
 #include "Basics/application-exit.h"
 #include "Basics/process-utils.h"
 #include "Cluster/ClusterFeature.h"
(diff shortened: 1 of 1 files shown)


arangod/Statistics/StatisticsFeature.cpp
```diff
@@ -339,6 +349,76 @@ std::map<std::string, std::vector<std::string>> statStrings{
...
+    -> StatBuilder {
+  auto unomap = StatBuilder{};
+  unomap.reserve(initList.size());
+  for (auto const& it : initList) {
+    unomap.emplace(it.first, it.second);
+  }
+  return unomap;
+}
+auto const statBuilder = makeStatBuilder({
+    {"bytesReceived", new arangodb_client_connection_statistics_bytes_received()},
```
jsteemann, line 366: All of these initializations are done when static variables are initialized.
Can this lead to some potential init order fiasco in case the to-be-created metrics depend on other static variables in a different TU?

mpoeter: No, everything here is fully local. This is only a temporary check to ensure everything is consistent. Longterm this should be cleaned up anyway so that this check is no longer necessary.

jsteemann (approved):

PR #13872, opened 2021-03-29
Added option to disallow usage of collections in expression

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/660

Added startup option `--query.allow-collections-in-expressions` to control whether collection names can be used in arbitrary places in AQL expressions, e.g. `collection + 1`. This was allowed before, as a collection can be seen as an array of documents. However, referring to a collection like this in a query would materialize all the collection's documents in RAM, making such constructs prohibitively expensive for medium-size to large-size collections.

The option can now be set to `false` to prohibit accidental usage of collection names in AQL expressions. With that setting, using a collection inside an arbitrary expression will trigger the error `collection used as expression operand` and make the query fail.
Even with the option being set to `false`, it is still possible to use collection names in AQL queries where they are expected, e.g. `FOR doc IN collection RETURN doc`.

#### Backports:

- [x] Backports required for: 3.8

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/660

### Testing & Verification


tests/js/client/server_parameters/test-allow-collections-in-expressions-off.js
```diff
@@ -0,0 +1,132 @@
...
+    testUseInTraversal: function() {
+      let result = db._query("WITH " + cn + " FOR v, e, p IN 1..3 OUTBOUND '" + cn + "/test0' " + en + " RETURN [v, e, p]").toArray();
+      assertEqual(0, result.length);
+    },
+    
+    testUseInShortestPath: function() {
+      let result = db._query("WITH " + cn + " FOR s IN OUTBOUND SHORTEST_PATH '" + cn + "/test0' TO '" + cn + "/test1' " + en + " RETURN s").toArray();
+      assertEqual(0, result.length);
+    },
+
```
mchacki, line 108: I think there are some Calculations that explicitly want a collection by CollectionIdentifier not by name which will require at least a test here (maybe even some modification in the production code, as I expect those tests to fail)

What I found right now:
* `LENGTH(collection)` or `COLLECTION_COUNT(collection)` which are aliases of each other.
* The deprecated GeoIndexFunctions https://www.arangodb.com/docs/stable/aql/functions-geo.html#geo-index-functions (I am open to let them not work with the flag set)
* Fulltext functions: https://www.arangodb.com/docs/stable/aql/functions-fulltext.html#fulltext

mchacki (changes requested):
This PR looks really good, however I am afraid there are some special cases to be considered.
Where I at least would request a test for.^

mchacki (approved):
LGTM, thanks for fixing :+1:

PR #13876, opened 2021-03-30
remove old SubqueryExecutor

jsteemann:
### Scope & Purpose

Remove now-unused SubqueryExecutor. This is an internal change only and should not have any effect on queries, as from 3.8 onwards only spliced subqueries should be used in query execution plans and during query execution.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtest, shell_server_aql*.


arangod/Aql/ExecutionNode.cpp
```diff
@@ -2171,33 +2168,7 @@ bool SubqueryNode::mayAccessCollections() {
...
-  auto executorInfos =
-      SubqueryExecutorInfos(*subquery, engine.getQuery(), outReg, const_cast<SubqueryNode*>(this)->isConst());
-  if (isModificationNode()) {
-    return std::make_unique<ExecutionBlockImpl<SubqueryExecutor<true>>>(
-        &engine, this, std::move(registerInfos), std::move(executorInfos));
-  } else {
-    return std::make_unique<ExecutionBlockImpl<SubqueryExecutor<false>>>(
-        &engine, this, std::move(registerInfos), std::move(executorInfos));
-  }
+  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "cannot instantiate SubqueryExecutor");
```
mpoeter, line 2171: Should we also add a `TRI_ASSERT(false)`?

mpoeter (commented):
We still have the `SubqueryNode` with its subquery + register plan. Is this still necessary? The subquery should no longer have a separate register plan which would allow some simplifications in register planning (in particular with regards to const registers).

arangod/Aql/ExecutionNode.cpp
jsteemann: Ok, done

PR #13886, opened 2021-04-01
Control unintended usage of collection names in AQL

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/13872

Docs PR: https://github.com/arangodb/docs/pull/660

Added startup option `--query.allow-collections-in-expressions` to control whether collection names can be used in arbitrary places in AQL expressions, e.g. `collection + 1`. This was allowed before, as a collection can be seen as an array of documents. However, referring to a collection like this in a query would materialize all the collection's documents in RAM, making such constructs prohibitively expensive for medium-size to large-size collections.

The option can now be set to `false` to prohibit accidental usage of collection names in AQL expressions. With that setting, using a collection inside an arbitrary expression will trigger the error `collection used as expression operand` and make the query fail.
Even with the option being set to `false`, it is still possible to use collection names in AQL queries where they are expected, e.g. `FOR doc IN collection RETURN doc`.

The option `--query.allow-collections-in-expressions` is introduced with a default value of `true` in 3.8 to ensure downwards-compatibility, but the default value will change to `false` in 3.9. Furthermore, the option will be deprecated in 3.9 and removed in later versions, in addition to making unintended usage of collection names always an error.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/660

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,22 @@
+v3.8.0 (XXXX-XX-XX)
+-------------------
+
+* Added startup option `--query.allow-collections-in-expressions` to control
+  whether collection names can be used in arbitrary places in AQL expressions,
+  e.g. `collection + 1`. This was allowed before, as a collection can be seen
+  as an array of documents. However, referring to a collection like this in a
+  query would materialize all the collection's documents in RAM, making such
+  constructs prohibitively expensive for medium-size to large-size collections.
+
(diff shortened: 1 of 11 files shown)


KVS85 (approved):
LGTM.

PR #13887, opened 2021-04-01
Deprecate option `--rocksdb.exclusive-writes`

jsteemann:
### Scope & Purpose

Deprecate option `--rocksdb.exclusive-writes`

The option was meant to serve only as a stopgap measure while porting applications from the MMFiles storage engine to RocksDB.

#### Backports:

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Deprecate option `--rocksdb.exclusive-writes`, which was meant to serve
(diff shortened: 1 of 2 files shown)


Simran-B (approved):
Would be great if you could add an article to the docs that explains how to deal with conflicts etc. that can occur with RocksDB. Some users seem to have enabled this option because they didn't know how else to fix conflict errors.

PR #13916, opened 2021-04-06
instead of silently ignoring / asserting return an error if one tries to create a collection on a dbserver

dothebart:
### Scope & Purpose

behave honest to the wrongly API on DB-Server using instead of silently ignoring the collection cretion or asserting.

#### Backports:

- [x] Backports required for: all

### Testing & Verification

use arangosh to create a collection on a dbserver.

Link to Jenkins PR run:
http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/14776/

### DIFF
--- a/arangod/RestHandler/RestCollectionHandler.cpp
+++ b/arangod/RestHandler/RestCollectionHandler.cpp
@@ -322,6 +322,24 @@ RestStatus RestCollectionHandler::handleCommandGet() {
 
 // create a collection
 void RestCollectionHandler::handleCommandPost() {
+
+  switch (ServerState::instance()->getRole()) {
+  case ServerState::ROLE_SINGLE:
+    break;
+  case ServerState::ROLE_DBSERVER:
+    generateError(rest::ResponseCode::BAD, TRI_ERROR_FORBIDDEN, "may not create collections on DB-Servers");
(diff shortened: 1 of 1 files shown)


arangod/RestHandler/RestCollectionHandler.cpp
```diff
@@ -322,6 +322,24 @@ RestStatus RestCollectionHandler::handleCommandGet() {
...
+    return;
+  case ServerState::ROLE_COORDINATOR:
+    break;
+  case ServerState::ROLE_AGENT:
+    break;
+  case ServerState::ROLE_UNDEFINED:
+    generateError(rest::ResponseCode::BAD, TRI_ERROR_FORBIDDEN, "may not create collections on who am I anyways?");
+    events::CreateCollection(_vocbase.name(), "", TRI_ERROR_FORBIDDEN);
+    return;
+  }
```
jsteemann, line 341:
```suggestion

  if (ServerState::instance()->isDBServer()) {
    generateError(Result(TRI_ERROR_CLUSTER_ONLY_ON_COORDINATOR));
    return;
  }
```

jsteemann:
We should also protect HTTP DELETE in the same way, potentially also HTTP PUT. It may be easier to do this in `execute()`.

PR #13925, opened 2021-04-07
convert test to making it use the fuerte arango connection

dothebart:
### Scope & Purpose
and the connection cache

Reduce resource usage of test by re-using tcp connections.

the previous implementation state of this test would open and close again ~1500 tcp connections.  After this change only ~500 will be used by utilizing the new arangodb connection cache.

This may be one of the responsible tests for TCP connection unavailable on windows nightly runs. 

Tcp connections are a sparse resource and re-opening them is (real-)time intense, we shouldn't be as waste full with it. 

- [x] :hankey: Bugfix
- [x] Backports: 3.7:  https://github.com/arangodb/arangodb/pull/13937 3.8: https://github.com/arangodb/arangodb/pull/13936


dothebart:
failure in other testcase, rest of the run is green.

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
+      return;
     }
+    trx = result.parsedBody.result.id;
+    const query = "FOR x IN 1..${docsPerWrite} INSERT {} INTO ${twoShardColName}";
+    result = arango.POST_RAW("/_api/cursor", { query }, { "x-arango-trx-id": trx });
+    if (result.code === 201) {
+        // Commit
+        result = arango.PUT_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {});
+        if (result.code == 200) {
+          break;
```
mchacki, line 486: if this is a return, the while is obsolete

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
+        }
+    }
+    print('apiExclusive failure:');
+    print(result);
+    print(arango.DELETE_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {}));
```
mchacki, line 491: please separate the print from the call.
The call has an effect the print is "for user readability".
It would be perfectly okay to run a "verbose=false" mode that simply replaces print() with a no op, that would cause this test to fail. I know we do not have this, but better save then sorry.

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
+        result = arango.PUT_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {});
+        if (result.code === 200)  {
+          break;
+        }
+      }
     }
+    print('apiWrite failure:');
+    print(result);
+    // Abort
+    print(arango.DELETE_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {}));
```
mchacki, line 515: Please separate print and api call.

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
+    }
+    print('apiExclusive failure:');
+    print(result);
+    print(arango.DELETE_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {}));
+    return;
+  }
   `;
   const apiWrite = `
-    ${apiLibs}
+  while (true) {
```
mchacki, line 496: There is no loop

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
+    print('apiWrite failure:');
+    print(result);
+    // Abort
+    print(arango.DELETE_RAW('/_api/transaction/' + encodeURIComponent(trx), {}, {}));
+    return;
+  }
   `;
   const apiRead = `
-    ${apiLibs}
+  while(true) {
```
mchacki, line 520: There is no loop

tests/js/client/communication/test-communication.js
```diff
@@ -578,14 +555,15 @@ function GenericAqlSetupPathSuite(type) {
...
-    let args = ['--javascript.execute-string', cmd];
+    let args = {'javascript.execute-string': cmd};
     let pid = runShell(args, key);
     debug("started client with key '" + key + "', pid " + pid + ", args: " + JSON.stringify(args));
     return { key, pid };
   };
 
   const singleRun = (tests) => {
     let clients = [];
+    reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 566: if the helper methods do this reconnectRetry internally this Is not required at all

tests/js/client/communication/test-communication.js
```diff
@@ -732,6 +710,7 @@ function GenericAqlSetupPathSuite(type) {
 
   const testSuite = {
     setUpAll: function () {
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 713: if the helper methods do this reconnectRetry internally this Is not required at all

tests/js/client/communication/test-communication.js
```diff
@@ -767,6 +746,7 @@ function GenericAqlSetupPathSuite(type) {
     },
 
     tearDownAll: function () {
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 749: if the helper methods do this reconnectRetry internally this Is not required at all

mchacki (changes requested):
I am happy that you got rid of the large internal API requirement in server-side code :+1:
I do not understand the argv refactoring, but okay.

However I have left some comments.

tests/js/client/communication/test-communication.js
```diff
@@ -72,28 +75,27 @@ function getEndpointById(id) {
...
+    'log.foreground-tty': 'false',
+    'log.output': 'file://' + prefix + '.log'
+  };
+  _.assign(args, moreArgs);
+  let argv = toArgv(args);
+
   for (let o in options['javascript.module-directory']) {
-    args.push('--javascript.module-directory');
-    args.push(options['javascript.module-directory'][o]);
+    argv.push('--javascript.module-directory');
```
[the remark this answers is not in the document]
dothebart: no, moreArgs is only working for uniq options. these may occur multiple times.

tests/js/client/communication/test-communication.js
```diff
@@ -470,96 +466,77 @@ function GenericAqlSetupPathSuite(type) {
...
-          res.body = JSON.parse(res.body);
-        }
-      }
-      return res;
-    }
-  `;
   // Note: A different test checks that the API works this way
   const apiExclusive = `
-    ${apiLibs}
+  while (true) {
```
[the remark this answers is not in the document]
dothebart: its for having `return` abort, and break cut. Goto so to say.

tests/js/client/communication/test-communication.js
dothebart: we only get there on error, hence verbosity by default is ok here imho.

tests/js/client/communication/test-communication.js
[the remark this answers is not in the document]
mchacki: ah yes, now I see.
This is bad.
After this code is executed there is an insert into a collection, on which we wait
So in case of the error this insert is NOT happening anymore, causing the test to get stuck instead of fail.
We cannot do a return here, the code needs to complete until the end.

tests/js/client/communication/test-communication.js
```diff
@@ -130,6 +130,7 @@ function CommunicationSuite() {
     `);
 
     let args = {'javascript.execute': file};
+    // print(file);
```
mchacki, line 133:
```suggestion
```

tests/js/client/communication/test-communication.js
```diff
@@ -309,19 +310,20 @@ function GenericAqlSetupPathSuite(type) {
       reconnectRetry(endpoint, db._name(), "root", "");
       
       let res = arango.GET_RAW('/_admin/debug/failat');
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
       return res.code === 200;
     } catch(ex) {
       reconnectRetry(primaryEndpoint, "_system", "root", "");
       throw (ex);
     }
```
mchacki:
```suggestion
      let res = arango.GET_RAW('/_admin/debug/failat');
      return res.code === 200;
    } finally {
      reconnectRetry(primaryEndpoint, "_system", "root", "");
    }
```

tests/js/client/communication/test-communication.js
```diff
@@ -309,19 +310,20 @@ function GenericAqlSetupPathSuite(type) {
...
-    return false;
   }
 
   /// @brief set failure point
   function debugSetFailAt(endpoint, failAt) {
     try {
       reconnectRetry(endpoint, db._name(), "root", "");
       let res = arango.PUT_RAW('/_admin/debug/failat/' + failAt, {});
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 326: This will now reconnect twice in case of error.
Please replace this by a simple `finally` (see above)
Unfortunately GitHub does not allow me to da a multiline comment here

tests/js/client/communication/test-communication.js
```diff
@@ -330,13 +332,13 @@ function GenericAqlSetupPathSuite(type) {
...
       throw (ex);
     }
-    return false;
   }
 
   function debugResetRaceControl(endpoint) {
     try {
       reconnectRetry(endpoint, db._name(), "root", "");
       let res = arango.DELETE_RAW('/_admin/debug/raceControl');
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 341: Replace by a finally.

tests/js/client/communication/test-communication.js
```diff
@@ -346,14 +348,14 @@ function GenericAqlSetupPathSuite(type) {
...
     }
-    return false;
   };
 
   /// @brief remove failure point
   function debugRemoveFailAt(endpoint, failAt) {
     try {
       reconnectRetry(endpoint, db._name(), "root", "");
       let res = arango.DELETE_RAW('/_admin/debug/failat/' + failAt);
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 358: Replace by a finally

tests/js/client/communication/test-communication.js
```diff
@@ -363,13 +365,13 @@ function GenericAqlSetupPathSuite(type) {
...
       throw (ex);
     }
-    return false;
   }
 
   function debugClearFailAt(endpoint) {
     try {
       reconnectRetry(endpoint, db._name(), "root", "");
       let res = arango.DELETE_RAW('/_admin/debug/failat');
+      reconnectRetry(primaryEndpoint, "_system", "root", "");
```
mchacki, line 374: Replace by a finally

tests/js/client/communication/test-communication.js
```diff
@@ -503,32 +504,27 @@ function GenericAqlSetupPathSuite(type) {
...
     let result = arango.POST_RAW("/_api/transaction/begin", obj);
-    if (result.code !== 201) {
-      print(result.parsedBody);
-      break;
-    }
-    trx = result.parsedBody.result.id;
-    const query = "FOR x IN 1..${docsPerWrite} INSERT {} INTO ${twoShardColName}";
-    result = arango.POST_RAW("/_api/cursor", { query }, { "x-arango-trx-id": trx });
     if (result.code === 201) {
+      trx = result.parsedBody.result.id;
```
mchacki, line 511: This looks good \o/

tests/js/client/communication/test-communication.js
```diff
@@ -592,6 +585,7 @@ function GenericAqlSetupPathSuite(type) {
 
     let args = {'javascript.execute-string': cmd};
     let pid = runShell(args, key);
+    // print(cmd);
```
mchacki, line 588:
```suggestion
```

mchacki (changes requested):
Much better now, like it.
Still some suggestions open

mchacki:
On another note:
I would wish for a more descriptive Issue description.
* Why are you doing this? Where does it originate from?
* Does it need back ports?

mchacki (approved):
LGTM

PR #13927, opened 2021-04-08
Experimentally switch to wyhash (from xxhash) for velocypack

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/678

Add wyhash function to velocypack, in addition to either xxhash or fasthash.
In our builds, the built-in velocypack hash function is still xxhash.

wyhash is added as a (hopefully) faster alternative, but it is not used in most places by default.
It _is_ used for the std::hash operator for StringRef/HashedStringRef however, because the hash values here are supposed to stay somewhat opaque and will not be persisted anywhere.

We unfortunately cannot use wyhash as a replacement for xxhash for velocypack::Slice etc., because ArangoDB relies on the generated hash values for Slices to determine the responsible shard(s). Changing the hash values here would be disastrous.

So for now we only use it for StringRef and HashedStringRef, and to calculate hashes of velocypack Slices in places that don't store these hashes anywhere or use them for any decisions. One place we can use it is the hash function used when storing AqlValues in hashed containers.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] There are tests in an external testing repository: arangodb/velocypack


Dronplane (approved):
Let`s try it! 🤞

PR #13939, opened 2021-04-09
Bug fix/bts 346

mchacki:
### Scope & Purpose

This PR adds many tests for killing AQL queries at specific "dangerous" positions.
It also fixes all sub-optimal handlings of those Kill commands.
* Sometimes internal ASSERTIONS were wrong (they expect the query to be active although it could be killed)
* A Cluster query will only trigger cleanup once. This was handled later before but caused the test to endless loop.
* A dropped database will now clear and garbage collect all lingering cursors, without this an AQLFeature lease was kept forever prohibiting the Server to shutdown. (Only if a query is killed, and the DB was dropped before the garbage collection kicked in, so unlikely)
* I killed cursor can now be removed from the Cursor list on demand as long as it is not in use. Before we had to wait for garbage collection to kick in for this operation, as our specific remove now call was ignored. This could cause a 5sec delay to release Locks.
* Fixes BTS-346 which actually kills a query during it's setup and got into a wrong assertion.

#### Backports:

- [ ] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation

This does not require external documentation, everything behaves as described, just faster.


arangod/Aql/Query.cpp
```diff
@@ -1448,3 +1469,58 @@ aql::ExecutionState Query::cleanupTrxAndEngines(ErrorCode errorCode) {
...
+    collections().visit([&usingSystemCollection](std::string const&, Collection& col) -> bool {
+      if (col.getCollection()->system()) {
+        usingSystemCollection = true;
+        return false;
+      }
+      return true;
+    });
+
+    if (usingSystemCollection) {
+      LOG_DEVEL << "Using system collection: " << std::boolalpha << usingSystemCollection;
```
jsteemann, line 1492: I am in favor of deleting this log line.
```suggestion
```

arangod/Aql/Query.cpp
```diff
@@ -1448,3 +1469,58 @@ aql::ExecutionState Query::cleanupTrxAndEngines(ErrorCode errorCode) {
...
+      return;
+    }
+
+    _wasDebugKilled = true;
+    // A query can only be killed under certain circumstances.
+    // We assert here that one of those is true.
+    // a) Query is in the list of current queries, this can be requested by the user and the query can be killed by user
+    // b) Query is in the query-registery. In this case the query registry can hit a timeout, which triggers the kill
```
jsteemann, line 1500:
```suggestion
    // b) Query is in the query registry. In this case the query registry can hit a timeout, which triggers the kill
```

arangod/Aql/Query.cpp
```diff
@@ -1448,3 +1469,58 @@ aql::ExecutionState Query::cleanupTrxAndEngines(ErrorCode errorCode) {
...
+        }
+      }
+    }
+
+    QueryRegistry* registry = QueryRegistryFeature::registry();
+    if (registry != nullptr) {
+      isInRegistry = registry->queryIsRegistered(vocbase().name(), _queryId);
+    }
+    LOG_DEVEL << "Query killed for debugging in current: " << std::boolalpha << isInList << " ,streaming: " << isStreaming << ", in QueryRegistry: " << isInRegistry;
+    LOG_DEVEL << "Query string is: " << queryString();
```
jsteemann, line 1521: Do we need these message? If yes, let's make them proper `LOG_TOPIC(...)` messages.
```suggestion
    LOG_DEVEL << "Query killed for debugging in current: " << std::boolalpha << isInList << ", streaming: " << isStreaming << ", in QueryRegistry: " << isInRegistry;
    LOG_DEVEL << "Query string is: " << queryString();
```

arangod/Aql/Query.h
```diff
@@ -206,6 +206,13 @@ class Query : public QueryContext {
   aql::SnippetList& snippets() { return _snippets; }
   aql::ServerQueryIdList& serverQueryIds() { return _serverQueryIds; }
   aql::ExecutionStats& executionStats() { return _execStats; }
+
+
+  // Debug method to kill a query at a specific position
+  // during execution. It internally asserts that the query
+  // is actually visible through other APIS (e.g. current queries)
+  // so user actually has a chance to kill it here.
+  void debugKillQuery() override;
```
jsteemann, line 215: Wondering if we should wrap this method and its implementation into a 
```
#ifdef ARANGODB_ENABLE_FAILURE_TESTS
...
#endif
```
as well.

arangod/Aql/QueryCursor.h
```diff
@@ -90,6 +90,12 @@ class QueryStreamCursor final : public arangodb::Cursor {
 
   void kill() override;
 
+  // Debug method to kill a query at a specific position
+  // during execution. It internally asserts that the query
+  // is actually visible through other APIS (e.g. current queries)
+  // so user actually has a chance to kill it here.
+  void debugKillQuery() override;
```
jsteemann, line 97: Same here, should we wrap it into a 
```
#ifdef ARANGODB_ENABLE_FAILURE_TESTS
...
#endif
```
?

arangod/Aql/QueryRegistry.h
```diff
@@ -122,6 +122,10 @@ class QueryRegistry {
   /// @brief return the default TTL value
   TEST_VIRTUAL double defaultTTL() const { return _defaultTTL; }
 
+#ifdef ARANGODB_ENABLE_FAILURE_TESTS
+  bool queryIsRegistered(std::string const& dbName, QueryId id);
```
jsteemann, line 126: It would be nice if we could make this method `const`.

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -290,6 +290,8 @@ class CalculationQueryContext final : public arangodb::aql::QueryContext {
 
   virtual bool killed() const override { return false; }
 
+  virtual void debugKillQuery() override {};
```
jsteemann, line 293:
```suggestion
  virtual void debugKillQuery() override {}
```

arangod/V8Server/v8-actions.cpp
```diff
@@ -1738,8 +1738,16 @@ static void JS_DebugClearFailAt(v8::FunctionCallbackInfo<v8::Value> const& args)
...
+      pathUrl = "_admin/debug/failat";
+    } else {
+      std::string const point = TRI_ObjectToString(isolate, args[0]);
+      pathUrl = "_admin/debug/failat/" + point;
+    }
+
     auto res =
-        clusterSendToAllServers(isolate, dbname, "_admin/debug/failat",
+        clusterSendToAllServers(isolate, dbname, pathUrl,
                                 arangodb::rest::RequestType::DELETE_REQ, "");
```
jsteemann, line 1751: Ok, now `JS_DebugClearFailAt` can do what `JS_DebugRemoveFailAt` can already do:
https://github.com/arangodb/arangodb/pull/13939/files?diff=unified&w=1#diff-284d4d1b634fa1caa50716432cab64ca7e6c6679df6b2c35db7a30103615604eR1677

Previously we had 2 distinct methods:
* "clear": unconditionally remove all failure points
* "remove": remove a single failure point by name
Now "clear" supports both modes, which is ok for me. Just mentioning it here.

tests/js/common/shell/test-kill-queries.js
```diff
@@ -0,0 +1,328 @@
...
+    // stream execution
+    let localQuery;
+    let stateForBoth = false; // marker that we expect either a kill or a result
+
+    try {
+      localQuery = db._query(exlusiveWriteQueryString, null, null, {stream: true});
+      localQuery.dispose();
+      // 1.) dispose muss funktionieren
+      // 2.) OHNE dispose, Cursor hält noch daten: Erwartet das der Cursor lebt => timeout "rennen"
+      // 3.) OHNE DISPOSE, cursor hält KEINE weiteren Daten mehr (ist fertig) => timeout sollte nicht mehr auftreten.
```
jsteemann, line 85: Let's convert these comments to English.

tests/js/common/shell/test-kill-queries.js
```diff
@@ -0,0 +1,328 @@
...
+    setUp: function () {
+      db._useDatabase("_system");
+      db._createDatabase(databaseName);
+      db._useDatabase(databaseName);
+
+      db._create(collectionName, {numberOfShards: 4});
+    },
+
+    tearDown: function () {
+      console.warn("CALLING TEAR DOWN, DROPPING DB");
```
jsteemann, line 220: Calling `tearDown` is expected behavior. Shall we remove the `console.warn(...)` here?

tests/js/common/shell/test-kill-queries.js
```diff
@@ -0,0 +1,328 @@
...
+    localQuery.dispose();
+  };
+
+  testSuite["test_positiveDefaultQueryExecution"] = function () {
+    let localQuery = db._query(exlusiveWriteQueryString);
+    assertTrue(localQuery);
+  };
+
+  if (!isServer && !internal.isCluster()) {
+    testSuite["test_export_api"] = function () {
```
jsteemann, line 290: This will very likely fail once PR https://github.com/arangodb/arangodb/pull/13929 is merged.

arangod/Aql/Query.cpp
mchacki: aarggh!!!
Shame on me!

arangod/Aql/Query.cpp
mchacki: No we do not need this message, we used it during development to make sure we actually hit the FailurePoints when we expect them. And stupid me forgot to cleanup properly.

arangod/Aql/Query.h
mchacki: I would love that (and initial did so), BUT:
You cannot compile without FAILURE tests anymore.

Here is why:

```
TRI_IF_FAILURE("foo") {
  query->debugKillQuery();
}
```

will be translated to:
```
if (false) {
  query->debugKillQuery()
}
```
without FAILURE_TESTS

And compilers are not clever enough to figure out that this code is in-fact unreachable and will complain that the function does not exist.

I tried

```
if constexpr (false) {
```

as well here, but this only works within templated code, otherwise it will also try to compile the unreachable branch.

So we need some way to reformulate the TRI_IF_FAILURE macro to remove the non-compiled code first.

jsteemann: Or alternatively wrap all users of `TRI_IF_FAILURE(...) { ... }` into `#ifdef`s as well.
This is also not a great alternative.

Ok, then let's leave it as is.

arangod/Aql/QueryRegistry.h
mchacki: This would require to make the lock mutable.
I am up for this, but I did not want to modify the "production" path in unnecessary ways here.

arangod/V8Server/v8-actions.cpp
mchacki: That is a good catch.
On client there was only clearFailAt(?name) which could take an optional failurePoint Name. The server side did not.
The server offers removeFailAt(name) and clearFailAt().
I have aliased removeFailAt and clearFailAt on client, and will remove the special JS_ implementation here.

jsteemann (approved):

PR #13953, opened 2021-04-13
convert foxx tests into using the cached arangosh connection

dothebart:
### Scope & Purpose
rework tests for better performance and using the arangosh tcp connection instead of `internal.download`

- [x] :hankey: Bugfix


pluma (approved):
Can't test this locally right now but JS LGTM.

arangod/V8Server/v8-actions.cpp
```diff
@@ -921,14 +921,7 @@ static void ResponseV8ToCpp(v8::Isolate* isolate, TRI_v8_global_t const* v8g,
...
       case Endpoint::TransportType::VST: {
-        VPackBuffer<uint8_t> buffer;
-        VPackBuilder builder(buffer);
-        builder.add(VPackValuePair(reinterpret_cast<uint8_t const*>(content), length));
-        TRI_FreeString(content);
-
-        // create vpack from file
-        response->setContentType(rest::ContentType::VPACK);
-        response->setPayload(std::move(buffer));
+        response->addRawPayload(velocypack::StringRef(content, length));
```
jsteemann, line 924: we'll now leak the memory of `content` here.

arangosh/Shell/V8ClientConnection.cpp
```diff
@@ -1863,7 +1863,11 @@ v8::Local<v8::Value> V8ClientConnection::requestDataRaw(
...
     if (boost::iequals(StaticStrings::ContentTypeHeader, pair.first)) {
-      req->header.contentType(fu::ContentType::Custom);
+      if (pair.second == StaticStrings::MimeTypeJsonNoEncoding) {
+        req->header.contentType(fu::ContentType::Json);
+      } else {
+        req->header.contentType(fu::ContentType::Custom);
+      }
     } else if (boost::iequals(StaticStrings::Accept, pair.first)) {
       req->header.acceptType(fu::ContentType::Custom);
     }
```
jsteemann, line 1873: I don't see how this will help much.
For both headers `Content-Type` and `Accept` we will do something special here (i.e. set the `_contentType` or `_acceptType` in `req->header`).
However, after we do this we unconditionally call `req->header.addMeta(...)`, which for these headers will unconditionally overwrite `_contentType` or `_acceptType`.

If this is true, we can remove the whole special handling here, and simply call `req->header.addMeta(...)` here without _any_ special handling.

js/common/modules/@arangodb/test-helper.js
```diff
@@ -103,11 +102,8 @@ exports.waitForFoxxInitialized = function () {
       const coordinators = exports.getEndpointsByType('coordinator');
       let done = 0;
       coordinators.forEach((endpoint) => {
-        let res = request.get({
-          url: endpoint + '/wenn-der-fuxxmann-zweimal-klingelt',
-          timeout: 5
-        });
-        if (res.status === 404) {
+        let res = arango.GET_RAW('/wenn-der-fuxxmann-zweimal-klingelt');
```
jsteemann, line 105: The previous implementation was able to send data to _multiple_ coordinators with multiple endpoints. 
The changed implementation now sends all requests to the same endpoint.

pluma: Oops, you're right. This is broken now.

pluma (changes requested):

dothebart:
reverted the `test-helper` part.

pluma (approved):

arangod/V8Server/v8-actions.cpp
jsteemann, line 924: we still leak the memory of `content` here.

PR #13973, opened 2021-04-15
Fix shard error multiplication

neunhoef:
This fixes a problem found with the new testagent. In document batch
operations (which are also used for import jobs) there is a problem in
the error handling if a shard is completely off line (all replicas
cannot be reached for 15 minutes). In that case, the errors for the
individual documents of the batch for that shard are multiplied, for
each document, as many errors are created as the batch size is.
This essentially squares the number of bad responses.

 - Do not multiply shard errors.
 - CHANGELOG.

Scope & Purpose

Backports:

[*] Backports required for: 3.6, 3.7, 3.8

Related Information

[*] This is a backport from: https://github.com/arangodb/arangodb/pull/13970/
[*] Jira ticket number: https://arangodb.atlassian.net/browse/BTS-369

Testing & Verification

[*] This change is a trivial rework / code cleanup without any test coverage.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,8 @@
 v3.6.14 (XXXX-XX-XX)
 --------------------
 
-* Fixed a problem in document batch operations, where errors from one shard
-  were reported multiple times, if the shard is completely off line.
+* Fixed a problem in document batch operations, where errors from one shard were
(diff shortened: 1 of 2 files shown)


jsteemann (approved):
LGTM!

PR #13976, opened 2021-04-15
Add experimental asynchronous prefetching

mpoeter:
### Scope & Purpose

This is an experimental feature that asynchronously prefetches the next batch while processing the current batch. This works on node-level, so theoretically the level of parallelism can grow linear with the size of the execution plan. However, at the moment this is only enabled for queries with out modification nodes and nodes with a MultiDependencyFetcher are also excluded.

On my machine I could get almost 10x speedup for a stupid test query. Without async-prefetch:
```
Execution plan:
 Id   NodeType                  Calls      Items   Runtime [s]   Comment
  1   SingletonNode                 1          1       0.00000   * ROOT
  2   EnumerateCollectionNode   10060   10060000       2.45257     - FOR doc IN col   /* full collection scan, projections: `v` */
  3   CalculationNode           10060   10060000       1.72142       - LET a = NOOPT((doc.`v` + 1))   /* simple expression */   /* collections used: doc : col */
  4   CalculationNode           10060   10060000       1.22835       - LET b = NOOPT((a - 1))   /* simple expression */
  5   CalculationNode           10060   10060000       1.26388       - LET c = NOOPT((b + 1))   /* simple expression */
  6   CalculationNode           10060   10060000       1.26516       - LET d = NOOPT((c - 1))   /* simple expression */
  7   CalculationNode           10060   10060000       1.26299       - LET e = NOOPT((d + 1))   /* simple expression */
  8   CalculationNode           10060   10060000       1.26822       - LET f = NOOPT((e - 1))   /* simple expression */
  9   CalculationNode           10060   10060000       1.26687       - LET g = NOOPT((f + 1))   /* simple expression */
 10   CalculationNode           10060   10060000       1.26434       - LET h = NOOPT((g - 1))   /* simple expression */
 11   CalculationNode           10060   10060000       1.26330       - LET i = NOOPT((h + 1))   /* simple expression */
 12   CalculationNode           10060   10060000       1.26409       - LET j = NOOPT((i - 1))   /* simple expression */
 13   CalculationNode           10060   10060000       1.27056       - LET k = NOOPT((j + 1))   /* simple expression */
 14   CalculationNode           10060   10060000       1.26509       - LET l = NOOPT((k - 1))   /* simple expression */
 15   CalculationNode           10060   10060000       1.26286       - LET m = NOOPT((l + 1))   /* simple expression */
 16   CalculationNode           10060   10060000       1.26868       - LET n = NOOPT((m - 1))   /* simple expression */
 17   CalculationNode           10060   10060000       1.26722       - LET o = NOOPT((n + 1))   /* simple expression */
 18   CalculationNode           10060   10060000       1.27412       - LET p = NOOPT((o - 1))   /* simple expression */
 19   CalculationNode           10060   10060000       1.26283       - LET q = NOOPT((p + 1))   /* simple expression */
 20   CalculationNode           10060   10060000       1.26476       - LET r = NOOPT((q - 1))   /* simple expression */
 21   CalculationNode           10060   10060000       1.26212       - LET s = NOOPT((r + 1))   /* simple expression */
 22   CalculationNode           10060   10060000       1.26353       - LET t = NOOPT((s - 1))   /* simple expression */
 23   CalculationNode           10060   10060000       1.26609       - LET u = NOOPT((t + 1))   /* simple expression */
 24   CalculationNode           10060   10060000       1.26403       - LET v = NOOPT((u - 1))   /* simple expression */
 25   CalculationNode           10060   10060000       1.26957       - LET w = NOOPT((v + 1))   /* simple expression */
 26   CalculationNode           10060   10060000       1.27171       - LET x = NOOPT((w - 1))   /* simple expression */
 27   CalculationNode           10060   10060000       1.26751       - LET y = NOOPT((x + 1))   /* simple expression */
 28   CalculationNode           10060   10060000       1.27170       - LET z = NOOPT((y - 1))   /* simple expression */
 29   CalculationNode           10060   10060000       1.21351       - LET #28 = (a.`v` - z)   /* simple expression */
 30   CollectNode                   1          1       0.45767       - COLLECT AGGREGATE ss = SUM(#28)   /* sorted */
 31   ReturnNode                    1          1       0.00001       - RETURN ss

Indexes used:
 none

Optimization rules applied:
 Id   RuleName
  1   reduce-extraction-to-projection

Query Statistics:
 Writes Exec   Writes Ign   Scan Full   Scan Index   Filtered   Peak Mem [b]   Exec Time [s]
           0            0    10060000            0          0         131072        37.46554
```

With async prefetch:
```
Execution plan:
 Id   NodeType                  Calls      Items   Runtime [s]   Comment
  1   SingletonNode                 1          1       0.00000   * ROOT
  2   EnumerateCollectionNode   10060   10060000       3.16023     - FOR doc IN col   /* full collection scan, projections: `v` */
  3   CalculationNode           10060   10060000       0.03844       - LET a = NOOPT((doc.`v` + 1))   /* simple expression */   /* collections used: doc : col */
  4   CalculationNode           10060   10060000       0.01592       - LET b = NOOPT((a - 1))   /* simple expression */
  5   CalculationNode           10060   10060000       0.01456       - LET c = NOOPT((b + 1))   /* simple expression */
  6   CalculationNode           10060   10060000       0.01407       - LET d = NOOPT((c - 1))   /* simple expression */
  7   CalculationNode           10060   10060000       0.00635       - LET e = NOOPT((d + 1))   /* simple expression */
  8   CalculationNode           10060   10060000       0.06911       - LET f = NOOPT((e - 1))   /* simple expression */
  9   CalculationNode           10060   10060000       0.00125       - LET g = NOOPT((f + 1))   /* simple expression */
 10   CalculationNode           10060   10060000       0.00647       - LET h = NOOPT((g - 1))   /* simple expression */
 11   CalculationNode           10060   10060000       0.02432       - LET i = NOOPT((h + 1))   /* simple expression */
 12   CalculationNode           10060   10060000       0.03538       - LET j = NOOPT((i - 1))   /* simple expression */
 13   CalculationNode           10060   10060000       0.09294       - LET k = NOOPT((j + 1))   /* simple expression */
 14   CalculationNode           10060   10060000       0.06355       - LET l = NOOPT((k - 1))   /* simple expression */
 15   CalculationNode           10060   10060000       0.05288       - LET m = NOOPT((l + 1))   /* simple expression */
 16   CalculationNode           10060   10060000       0.03506       - LET n = NOOPT((m - 1))   /* simple expression */
 17   CalculationNode           10060   10060000       0.03658       - LET o = NOOPT((n + 1))   /* simple expression */
 18   CalculationNode           10060   10060000       0.01674       - LET p = NOOPT((o - 1))   /* simple expression */
 19   CalculationNode           10060   10060000       0.03715       - LET q = NOOPT((p + 1))   /* simple expression */
 20   CalculationNode           10060   10060000       0.08411       - LET r = NOOPT((q - 1))   /* simple expression */
 21   CalculationNode           10060   10060000       0.07831       - LET s = NOOPT((r + 1))   /* simple expression */
 22   CalculationNode           10060   10060000       0.06015       - LET t = NOOPT((s - 1))   /* simple expression */
 23   CalculationNode           10060   10060000       0.12552       - LET u = NOOPT((t + 1))   /* simple expression */
 24   CalculationNode           10060   10060000       0.01644       - LET v = NOOPT((u - 1))   /* simple expression */
 25   CalculationNode           10060   10060000       0.07114       - LET w = NOOPT((v + 1))   /* simple expression */
 26   CalculationNode           10060   10060000       0.16518       - LET x = NOOPT((w - 1))   /* simple expression */
 27   CalculationNode           10060   10060000       0.02767       - LET y = NOOPT((x + 1))   /* simple expression */
 28   CalculationNode           10060   10060000       0.11072       - LET z = NOOPT((y - 1))   /* simple expression */
 29   CalculationNode           10060   10060000       0.22476       - LET #28 = (a.`v` - z)   /* simple expression */
 30   CollectNode                   1          1       0.24735       - COLLECT AGGREGATE ss = SUM(#28)   /* sorted */
 31   ReturnNode                    1          1       0.00001       - RETURN ss

Indexes used:
 none

Optimization rules applied:
 Id   RuleName
  1   reduce-extraction-to-projection

Query Statistics:
 Writes Exec   Writes Ign   Scan Full   Scan Index   Filtered   Peak Mem [b]   Exec Time [s]
           0            0    10060000            0          0        1146880         3.90308
```
However, there are still a few open points that need to be sorted out. In particular we need to ensure that the internal transaction handling is thread-safe. Another point that needs to be investigated in more detail is behavior with multiple parallel paths (i.e., with gather/scatter nodes).

At the moment, the async tasks are posted on the scheduler, but that means we might flood the queue and potentially delay processing of incoming user REST requests, so we should consider ways to avoid this. Obviously, the more async requests are processed concurrently, the higher the memory usage as more blocks are processed concurrently and less blocks can be reused. So we should also consider ways to limit parallelism and think of strategies on how to identify for which nodes async prefetching should be enabled in such a scenario to achieve the optimal result. Another open point is visibility - we need to think of ways to make the degree of parallelism that actually occurred (just because we schedule asynchronous task does not mean that these are picked up by other worker threads; in the worst case all work is still performed by the thread that started the query) visible in the profiling output, e.g., per node wall-clock time + CPU time.

**For this reason, the according optimizer rule is disabled by default ATM!**

No changelog entry yet since this feature is not yet visible to users.

- [x] :pizza: New feature

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] I ensured this code runs with TSan


arangod/Aql/AqlItemBlockInputMatrix.h
```diff
@@ -37,13 +37,14 @@ class AqlItemBlockInputMatrix {
...
 
-  AqlItemBlockInputMatrix(arangodb::aql::SharedAqlItemBlockPtr const&);
-
   AqlItemBlockInputMatrix(ExecutorState state, AqlItemMatrix* aqlItemMatrix);
 
   std::pair<ExecutorState, ShadowAqlItemRow> nextShadowRow();
   ShadowAqlItemRow peekShadowRow() const;
 
+  void reset() noexcept { _lastRange.reset(); }
+  bool hasBlock() noexcept { return _lastRange.hasBlock(); }
```
jsteemann, line 46:
```suggestion
  bool hasBlock() const noexcept { return _lastRange.hasBlock(); }
```

arangod/Aql/AqlItemBlockInputRange.h
```diff
@@ -47,6 +47,9 @@ class AqlItemBlockInputRange {
                          arangodb::aql::SharedAqlItemBlockPtr&&,
                          std::size_t startIndex) noexcept;
 
+  void reset() noexcept { _block.reset(nullptr); }
+  bool hasBlock() noexcept { return _block.get() != nullptr; }
```
jsteemann, line 51:
```suggestion
  bool hasBlock() const noexcept { return _block.get() != nullptr; }
```

arangod/Aql/OptimizerRulesFeature.cpp
```diff
@@ -474,6 +474,14 @@ void OptimizerRulesFeature::addRules() {
   registerRule("splice-subqueries", spliceSubqueriesRule, OptimizerRule::spliceSubqueriesRule,
                OptimizerRule::makeFlags());
 
+  // allow nodes to asynchronously prefetch the next batch while processing the current batch.
+  // this effectively allows parts of the query to run in parallel, but as some internal details
+  // are currently not guaranteed to be thread safe (e.g., TransactionState), this is currently
+  // disabled, and should only be activated for experimental usage at one's own risk.
+  registerRule("asynch-prefetch", asynchPrefetchRule, OptimizerRule::asynchPrefetch,
```
jsteemann, line 481: Let's rename this to "async-" (not "asynch-"). I think "async" is more common than "asynch", although both would be understood.

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1820,6 +1855,82 @@ auto ExecutionBlockImpl<Executor>::testInjectInputRange(DataRange range, SkipRes
...
+      auto& subQueryCall = clientCallList.modifyNextCall();
+      // Overwrite with old state.
+      subQueryCall = block._outputItemRow->getClientCall();
+    }
+    stack.pushCall(std::move(clientCallList));
+    clientCallList = AqlCallList{AqlCall{}, AqlCall{}};
+    clientCall = clientCallList.popNextCall();
+  } else {
+    clientCall = clientCallList.popNextCall();
+  }
```
jsteemann, line 1878:
```suggestion
  }
  clientCall = clientCallList.popNextCall();
  
```

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1820,6 +1855,82 @@ auto ExecutionBlockImpl<Executor>::testInjectInputRange(DataRange range, SkipRes
...
+  // Caller is wrong fix it.
+  TRI_ASSERT(clientCall.getSkipCount() == 0);
+  
+  TRI_ASSERT(!(clientCall.getOffset() == 0 && clientCall.softLimit == AqlCall::Limit{0u}));
+  TRI_ASSERT(!(clientCall.hasSoftLimit() && clientCall.fullCount));
+  TRI_ASSERT(!(clientCall.hasSoftLimit() && clientCall.hasHardLimit()));
+}
+
+template <class Executor>
+bool ExecutionBlockImpl<Executor>::PrefetchTask::isConsumed() const {
```
jsteemann, line 1889: Can we mark this as `noexcept`?

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1820,6 +1855,82 @@ auto ExecutionBlockImpl<Executor>::testInjectInputRange(DataRange range, SkipRes
...
+  return state.load(std::memory_order_relaxed) == State::Consumed;
+}
+
+template <class Executor>
+bool ExecutionBlockImpl<Executor>::PrefetchTask::tryClaim() {
```
jsteemann, line 1894: Can we mark this as `noexcept`?

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1820,6 +1855,82 @@ auto ExecutionBlockImpl<Executor>::testInjectInputRange(DataRange range, SkipRes
...
+  auto expected = State::Pending;
+  return state.load(std::memory_order_relaxed) == expected &&
+    state.compare_exchange_strong(expected, State::InProgress, std::memory_order_relaxed);
+}
+
+template <class Executor>
+void ExecutionBlockImpl<Executor>::PrefetchTask::waitFor() {
```
jsteemann, line 1901: This method is called in destructors, which requires noexcept-ness. 
I think the method is effectively `noexcept`, although `guard` and `bell` may throw under some adverse conditions. 
We could either wrap this method in a try...catch, or mark it `noexcept` and have `std::terminate()` kill the process if an exception escapes from here. What shall we do, if anything?

mpoeter: Since `std::mutex` and `std::condition_variable` can potentially throw, I would not mark this function as `noexcept`, but add a try-catch in the destructor. IMHO that would be the cleanest solution.

jsteemann: Fine for me.

mpoeter: Actually, after checking the standard under which circumstances `std::mutex` might throw, I would go with your last suggestion and mark the function `noexcept` and have the process terminate if we really have an exception.

arangod/Aql/ExecutionBlockImpl.h
```diff
@@ -218,13 +218,20 @@ class ExecutionBlockImpl final : public ExecutionBlock {
 #endif
 
  private:
+  struct Context {
+    Context(ExecutionBlockImpl& block, AqlCallStack const& callstack);
```
goedderz, line 222: Perhaps we can change the name to something a little less generic? Perhaps `AqlCallContext` or so, or maybe you have another idea?

goedderz (commented):
I've commented on some minor points, LGTM in general!

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -687,16 +704,60 @@ auto ExecutionBlockImpl<Executor>::executeFetcher(AqlCallStack& stack,
...
+        auto result = std::move(_prefetchTask->result.value());
+        _prefetchTask->result.reset();
+        return result;
+      } else {
+        return _rowFetcher.execute(ctx.stack);
+      }
+    });
+
+    // TODO - we should also consider the limit here, but unfortunately the
+    // softLimit is currently also used for the batchSize
```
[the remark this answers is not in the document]
mpoeter: I also have an item in my private TODO list... would you prefer if I create an internal Jira ticket as well?

arangod/Aql/ExecutionBlockImpl.h
mpoeter: This `Context` is only used when executing `executeWithoutTrace` and captures variables that are passed to functions that are called as part of `executeWithoutTrace`. That should now also make it easier to extract some parts of `executeWithoutTrace` and move them to separate functions.
So how about `ExecutionContext`?

goedderz: Sounds good!

arangod/Aql/ExecutionBlockImpl.cpp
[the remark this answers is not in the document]
mpoeter: Alright! 😄

arangod/Aql/ExecutionBlockImpl.h
```diff
@@ -297,24 +297,54 @@ class ExecutionBlockImpl final : public ExecutionBlock {
...
+    void reset() noexcept;
+    Result stealResult() noexcept;
     
     void execute(ExecutionBlockImpl& block, AqlCallStack& stack);
+    
+   private:
+    std::atomic<State> _state{State::Pending};
+    std::mutex _lock;
+    std::condition_variable _bell;
+    std::optional<Result> _result;
```
goedderz: The typedef `Result` here can be a little bit misleading, thinking of our `arangodb::Result` type.

goedderz (commented):
Thanks for the changes, especially the very helpful comment! One last note from me:

goedderz (approved):

PR #13983, opened 2021-04-15
Fix ordering of Database and FlushFeatures

mchacki:
### Scope & Purpose

The FlushFeature uses the Instance of DatabaseFeature during start().
This instance is only set during start() of DatabaseFeature but the dependency is not defined.
As the startup is consistent after a build it by accident worked, now we force it to be correct.
Hence no changelog as it will not have visible effect.


jsteemann (approved):
LGTM

PR #13985, opened 2021-04-15
Fix cluster internal network request retry behaviour

neunhoef:
Backport of: https://github.com/arangodb/arangodb/pull/13942
The behaviour of network::sendRequestRetry is found lacking in multiple aspects:

 - the code for a follower refusing leader work should be on the list of cases to retry,
   since a former leader will respond with this and a coordinator could
   still talk to the old leader and then should move on to the new leader
   and retry, we decided to use a new code 421 Misdirected Request for this.

 - sendRequestRetry currently does not distinguish the cases of
   "Connection refused" (nothing can have happened on the other side)
   and "Connection Closed" (potentially broken pipe during the send or
   when waiting for response, in which case the request might have gone
   through, must not retry in most cases) and "Timeout" (request could have
   gone through, must not retry in most cases) and "Connection Cancelled"
   (request was locally cancelled, should retry to allow for failover).

Lot's of sendRequest calls should actually be sendRequestRetry.

And: Make sure that we cover the case of a stale connection coming from
the connection pool and do a retry if the connection breaks on first
using it.

Furthermore, we improve the retry behavior for agency communications
in the inquire case.

Finally, a coordinator redirect (for a query cursor or so) which finds that the coordinator
to redirect to is no longer member of a cluster must return 404 instead of 503, otherwise
the client would retry.
Scope & Purpose

    [*] This is a bug fix backport of: https://github.com/arangodb/arangodb/pull/13942

Testing & Verification

    [*] This change is already covered by existing tests, such as (please describe tests).
    [*] Additional integration and unit tests have been added for the retry behavior around the
    421 status code (ex-leader refuses leader operation).


mchacki (approved):
LGTM.
No changes to DEVEL in this code.
Good PR :+1:

PR #13986, opened 2021-04-15
added escaping of the filename string when printing it in the web ui

hkernbach:
### Scope & Purpose

When a zip file is being uploaded, the file name of the file (or filelist in case multiple files are supplied) will be printed during /after the upload process. That name (string) is not being escaped., but it needs to be.

This PR fixes this behaviour. 

#### Backports:

### Testing & Verification


fceller (approved):
LGTM

PR #13989, opened 2021-04-15
WET -> DRY.

dothebart:
### Scope & Purpose
- async cluster foxx startup (self healing) is now disabled in most tests, since it could take quiet some time to happen
- hence waitForFoxxInitialized isn't needed anymore and has been removed from `test-helper.js`
- test-helper.js has been split into 
  - common/test-helper-common.js
  - server/test-helper.js now containing the simple client implementations for setting debug points
  - client/test-helper.js now containing the new `arango.GET_RAW` implementations
- tests have removed the `waitForFoxxInitialized` invocations
- tests have removed their own debug-point-setter methods, but instead use the `test-helper` ones.

- [x] :hankey: Bugfix


fceller (approved):
LGTM, removes duplicate code for failure points

Issue #13655, opened 2021-03-04
Inconsistent comparison behavior in combination with OR


pfu3tz:
## My Environment

* __ArangoDB Version__:        3.7.8
* __Storage Engine__:            RocksDB 6.8.0
* __Deployment Mode__:      Single Server
* __Deployment Strategy__:   Manual Start
* __Configuration__:               -
* __Infrastructure__:               Own
* __Operating System__:        Ubuntu 18.04.5 LTS
* __Total RAM in your machine__:        16 GB
* __Disks in use__:       SSD
* __Used Package__:         Ubuntu .deb

## Component, Query & Data

__Affected feature__:
arangosh, but was originally detected with Java driver version 6.9.0

__AQL query (if applicable)__:
The affected query as I understand the documentation:
```
FOR doc IN table FILTER ((true >= doc.column) OR doc.column) RETURN doc
```
__AQL explain and/or profile (if applicable)__:
Query String (71 chars, cacheable: true):
 FOR doc IN table FILTER ((true >= doc.column) OR doc.column) RETURN doc

Execution plan:
 Id   NodeType                  Est.   Comment
  1   SingletonNode                1   * ROOT
  2   EnumerateCollectionNode      1     - FOR doc IN table   /* full collection scan */   FILTER (doc.`column` <= true)   /* early pruning */
  5   ReturnNode                   1       - RETURN doc

Indexes used:
 none

Optimization rules applied:
 Id   RuleName
  1   remove-redundant-or
  2   move-filters-into-enumerate

__Dataset__:
// This is what I ran:

// DB Creation and Insertion
db._dropDatabase("test")
db._createDatabase("test")
db._useDatabase("test")

db._create("table")
db._query("INSERT {column: 1} IN table")

// Without the Filter 
db._query("FOR doc IN table RETURN doc")
// db._explain("FOR doc IN table RETURN doc")

// The Query
db._query("FOR doc IN table FILTER (true >= doc.column) OR doc.column RETURN doc")
// db._explain("FOR doc IN table FILTER (true >= doc.column) OR doc.column RETURN doc")

// The negated query, in case I misunderstood how the comparison is evaluated
db._query("FOR doc IN table FILTER !(((true >= doc.column) OR doc.column)) RETURN doc")
// db._explain("FOR doc IN table FILTER !((true >= doc.column) OR doc.column) RETURN doc")

## Steps to reproduce
1. Run the script
2. Check if the single document shows up in either the negated or the query

__Problem__:
[...]

PR #13993, opened 2021-04-16
Fix BTS-340: Prevent overzealous remove-redundant-or

goedderz:
### Scope & Purpose

There are cases where the remove-redundant-or rule removes a part of a disjunction where it should not. Specifically,

```
val < 3 || val
```

and similar terms would erroneously be reduced to `val < 3`.

The actual target of the rule are expressions like `val < 3 || val < 5`, which can be simplified to `val < 5`.

The optimizer rule inspects an expression and expect it to be a tree with a binary or at its root. Children of a binary or are now either expected to be another binary or, or a comparison. The children of a comparison are then inspected for whether one of them is a variable access, which is then looked for in other parts of the expression, and whether the other one is a constant. All parts of the expression must currently match that structure, or the rule won't apply - there is room for improvement here.

Now if a variable access was already encountered previously in the expression (e.g. in `val < 3` in `val < 3 || val`), and a single variable access was encountered where the rule actually expected a comparison (e.g. the second `val` in `val < 3 || val`), the code erroneously continued successfully. This behaviour is fixed by this PR.

#### Backports:

- [ ] No backports required
- [X] Backports required for: *3.8* #14034
- [X] Backports required for: *3.7* #14035
- [X] Backports required for: *3.6* #14036

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [X] GitHub issue: https://github.com/arangodb/arangodb/issues/13655
- [X] Jira ticket number: BTS-340
- [ ] Design document: 

### Testing & Verification

  - [X] Added new **integration tests** (e.g. in shell_server / shell_server_aql)


kvahed (approved):
Reviewed and manually tested.

goedderz:
I'll open backports soon.

PR #14010, opened 2021-04-18
improve parallelism capabilities of arangorestore

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/688

Improve parallelism capabilities of arangorestore.

arangorestore can now dispatch restoring data chunks of a collection to idle background threads, so that multiple restore requests can be in flight for the same collection concurrently.

This can improve restore speed in situations when there are idle arangorestore threads left (number of threads can be configured via arangorestore's `--threads` option) and the dump file for the collection is large.
Previously the thread that read the data for file a collection would send HTTP requests to the server and block until the response came back. So even when there were idle threads available they could not be used.
With the change in this PR, the thread that reads the input file for a collection can either dispatch sending the data to another idle thread or send the data itself in case there is no other idle thread. In the former case the original thread continues to read data from the input file and the other thread does the request-response handling. The original thread can even dispatch requests to multiple idle threads if it can read the input file quicker than the request-response handling takes (which is to be expected).
In case there are no other idle threads available, the behavior is effectively unchanged, and the arangorestore thread that reads the input file is also sending requests and blocks until the response comes back.

The improved parallelism is only used when restoring dumps that are in the non-enveloped format. This format has been introduced with ArangoDB 3.8. The reason is that dumps in the non-enveloped format only contain the raw documents, which can be restored independent of each other, i.e. in any order. However, the enveloped format may contain documents and remove operations, which need to be restored in the original order. Thus we can't parallelize the enveloped format's restores.

The non-enveloped format is available for arangodump since ArangoDB 3.8, and is opt-in in 3.8.
In order to improve restore performance in more use cases, this PR will also change the default value of arangodump's `--envelope` option from `true` to `false`. As a side-effect, this will also decrease the size of dumps taken with arangodump, and should slightly improve their speed.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/688

### Testing & Verification


dothebart:
please note that this PR should be revalidated using this technique with clusters: https://www.arangodb.com/2019/01/using-the-linux-kernel-and-cgroups-to-simulate-starvation/

jsteemann:
@dothebart : I don't see any connection between the article and this PR, but feel free to do it anyway.

dothebart:
it needs to be ensured, that this patch doesn't end up breaking the import with clusters once they reach their I/O saturation. 
The CGroups test can be used to ensure this, you can also use the specified lower end GCP or AWS machines to revalidate. 

I'd name this as crucial DOD condition for working on restore / import.

jsteemann:
@dothebart : well, was there any speed/rate limit in place before my changes? I don't see any.

fceller (approved):
LGTM

arangosh/Restore/RestoreFeature.h
```diff
@@ -132,30 +141,119 @@ class RestoreFeature final : public application_features::ApplicationFeature {
...
+    SharedState() : readCompleteInputfile(false) {}
+
+    Mutex mutex;
+
+    /// @brief this contains errors produced by background send operations
+    /// (i.e. RestoreSendJobs)
+    Result result;
+
+    /// @brief data chunk offsets (start offset, length) of requests that are currently
+    /// ongoing. it is safe the resume restore only up the minimum value contained
```
mpoeter, line 156:
```suggestion
    /// ongoing. Resumption of the restore must start at the smallest key value contained
```

mpoeter (approved):

PR #14018, opened 2021-04-19
Try to resume RocksDB operations after bg error

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/14017

Try to automatically resume RocksDB operations after a RocksDB background error. 

On a background error, RocksDB may have gone into read-only mode and not try to recover from it automatically in some situations, e.g. if the background error happened during compaction. In this case it will stay in read-only mode until shutdown, or until its Resume() API is called.

Now we try to automatically resume RocksDB operations about every 15 seconds in case a background error has happened. Resuming may not succeed in all cases, but it can help in some cases if the original error condition is fixed.

This may fix a few cases of RocksDB going into read-only mode and not automatically recovering from it.

#### Backports:

- [x] No backports required.

### Testing & Verification

### DIFF
--- a/3rdParty/rocksdb/6.8/util/status.cc
+++ b/3rdParty/rocksdb/6.8/util/status.cc
@@ -135,6 +135,9 @@ std::string Status::ToString() const {
   }
 
   if (state_ != nullptr) {
+    if (subcode_ != kNone) {
+      result.append(": ");
+    }
     result.append(state_);
   }
   return result;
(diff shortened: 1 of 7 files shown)


jsteemann:
RocksDB still maps some potentially recoverable BG errors to non-recoverable errors.
Waiting for RocksDB upgrade...

PR #14019, opened 2021-04-19
Retry if an ex leader cannot drop a follower

neunhoef:
To be extended.

- Retry if an ex-leader can no longer drop a follower
- CHANGELOG.


arangod/Cluster/ClusterTrxMethods.cpp
```diff
@@ -297,7 +297,11 @@ Future<Result> commitAbortTransaction(arangodb::TransactionState* state,
                       << "synchronous replication: could not drop follower " << follower
                       << " for shard " << cc->vocbase().name() << "/" << tc.collectionName()
                       << ": " << r.errorMessage();
-                  res.reset(TRI_ERROR_CLUSTER_COULD_NOT_DROP_FOLLOWER);
+                  if (res.is(TRI_ERROR_CLUSTER_NOT_LEADER)) {
+                    THROW_ARANGO_EXCEPTION(TRI_ERROR_CLUSTER_SHARD_LEADER_RESIGNED);
+                  } else {
+                    THROW_ARANGO_EXCEPTION(TRI_ERROR_CLUSTER_COULD_NOT_DROP_FOLLOWER);
+                  }
```
jsteemann, line 304: Is it intentional to throw an exception here?
This code is in a loop that iterates over all collections in the transaction.
Now we may abort after dropping the first follower we could not remove, and don't even try removing others even though that may be necessary.

goedderz (commented):
To see if I understand the changes correctly: In these cases replication (of any modification operation, truncate, or commit/abort) to a follower failed. In case the consequential follower removal fails because we are (no longer) the leader, this makes the thrown error more correct.

The CHANGELOG entry seems to indicate that this will result in a retry, but I don't see where or why that happens. Can you give me a hint here?

arangod/Cluster/ClusterTrxMethods.cpp
kvahed: Yes, as it rolls back the local transaction. This behaviour has been in place since before this fix.

kvahed:
@goedderz precisely. The intention is to to respond with "Could not drop follower" when you realised that you have not even done that, as you found out that you have lost the leadership

arangod/Cluster/ClusterTrxMethods.cpp
jsteemann: Is it intentional to drop out of the loop over all followers when removing one of the followers fails? Shall the loop continue and try to remove the other followers too, and only fail afterwards?

jsteemann: Plus, I think it _is_ a behavior change. Because previously the loop continued, and now it is aborted early.

goedderz:
Thanks.

goedderz (commented):
Could you add a regression test for this?

jsteemann (approved):
LGTM

PR #14020, opened 2021-04-19
Improve parallelism capabilities of arangorestore.

jsteemann:
### Scope & Purpose

Partial backport of https://github.com/arangodb/arangodb/pull/14010
Docs PR: https://github.com/arangodb/docs/pull/688

Improve parallelism capabilities of arangorestore.

arangorestore can now dispatch restoring data chunks of a collection to idle background threads, so that multiple restore requests can be in flight for the same collection concurrently.

This can improve restore speed in situations when there are idle arangorestore threads left (number of threads can be configured via arangorestore's `--threads` option) and the dump file for the collection is large.
Previously the thread that read the data for file a collection would send HTTP requests to the server and block until the response came back. So even when there were idle threads available they could not be used.
With the change in this PR, the thread that reads the input file for a collection can either dispatch sending the data to another idle thread or send the data itself in case there is no other idle thread. In the former case the original thread continues to read data from the input file and the other thread does the request-response handling. The original thread can even dispatch requests to multiple idle threads if it can read the input file quicker than the request-response handling takes (which is to be expected).
In case there are no other idle threads available, the behavior is effectively unchanged, and the arangorestore thread that reads the input file is also sending requests and blocks until the response comes back.

The improved parallelism is only used when restoring dumps that are in the non-enveloped format. This format has been introduced with ArangoDB 3.8. The reason is that dumps in the non-enveloped format only contain the raw documents, which can be restored independent of each other, i.e. in any order. However, the enveloped format may contain documents and remove operations, which need to be restored in the original order. Thus we can't parallelize the enveloped format's restores.

The non-enveloped format is available for arangodump since ArangoDB 3.8, and is opt-in in 3.8. It will become the default dump format in ArangoDB 3.9.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/688

### Testing & Verification

### DIFF
--- a/arangosh/Restore/RestoreFeature.cpp
+++ b/arangosh/Restore/RestoreFeature.cpp
@@ -39,9 +39,11 @@
 #include "Basics/NumberOfCores.h"
 #include "Basics/Result.h"
 #include "Basics/StaticStrings.h"
+#include "Basics/StringBuffer.h"
 #include "Basics/StringUtils.h"
 #include "Basics/VelocyPackHelper.h"
 #include "Basics/application-exit.h"
+#include "Basics/debugging.h"
 #include "Basics/files.h"
 #include "Basics/system-functions.h"
 #include "FeaturePhases/BasicFeaturePhaseClient.h"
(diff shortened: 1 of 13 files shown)


arangosh/Restore/RestoreFeature.cpp
```diff
@@ -1306,65 +967,534 @@ arangodb::Result processInputDirectory(
...
+  if (res.fail()) {
+    // error
+    LOG_TOPIC("a595a", WARN, Logger::RESTORE)
+        << "Error while restoring data into collection '" << collectionName
+        << "': " << res.errorMessage();
+    // leave readOffset in place in readOffsets, because the job did not succeed
+    {
+      // store error
+      MUTEX_LOCKER(locker, sharedState->mutex);
+      sharedState->result = res;
```
mpoeter, line 1037: Here we are potentially overwriting previous errors. Should we consider collecting the errors instead so we can report them all? However, this would probably require rather big changes...

arangosh/Restore/RestoreFeature.cpp
```diff
@@ -1306,65 +967,534 @@ arangodb::Result processInputDirectory(
...
+    progressTracker.updateStatus(
+          collectionName, arangodb::RestoreFeature::CollectionStatus{arangodb::RestoreFeature::RESTORING, readOffset});
+  } else if (sharedState->readCompleteInputfile) {
+    // we are done with restoring the entire collection
+    
+    // progressTracker has its own lock
+    locker.unlock();
+
+    progressTracker.updateStatus(
+        collectionName, arangodb::RestoreFeature::CollectionStatus{arangodb::RestoreFeature::RESTORED, 0});
```
mpoeter, line 1088: Why do we report 0 here?

mpoeter (approved):

arangosh/Restore/RestoreFeature.cpp
jsteemann: The actual value `0` does not matter here. The more important thing is that we store status `RESTORED`, which means "fully restored".
This is copy&paste from the old code, and should be just fine.

arangosh/Restore/RestoreFeature.cpp
jsteemann: Yes, we could be overwriting an already stored error with another error.
But we will never overwrite an error with a non-error here.
I don't think it is a problem, because with default startup parameters the restore will stop on an error. The error is also reported to the user. Which error it is does not really matter, and if multi-threading is involved, will likely depend on scheduling and such.
It may be useful to collect all the errors here, but it is a bigger change and I didn't want to do it in context of this PR.

arangosh/Restore/RestoreFeature.cpp
mpoeter: Got it!

PR #14029, opened 2021-04-20
Bug fix/fix tsan issues

mpoeter:
### Scope & Purpose

Fixes various issues (mainly data races) reported by running the integration tests with a TSan build. At least some of those fixes need to be backported to 3.8.

#### Backports:

- [x] Backports required for: *3.8* (#14092)

### Testing & Verification


arangod/Cluster/HeartbeatThread.cpp
```diff
@@ -1088,6 +1086,7 @@ bool HeartbeatThread::init() {
   if (ServerState::instance()->isClusterRole() && !sendServerState()) {
     return false;
   }
+  _maintenanceThread = std::make_unique<HeartbeatBackgroundJobThread>(_server, this);
```
jsteemann, line 1089: As far as I can see, this starts the maintenanceThread now in all server modes (i.e. coordinator and DB server).
But AFAIK only the DB server should have this thread. So I think initializing in the original place was better.

arangod/RestServer/Metrics.h
```diff
@@ -539,9 +547,10 @@ template<typename Scale> class Histogram : public Metric {
...
  private:
   Metrics::hist_type _c;
-  Scale _scale;
-  value_type _lowr, _highr;
-  size_t _n;
+  Scale const _scale;
+  size_t const _n;
+  std::atomic<value_type> _lowr;
+  std::atomic<value_type> _highr;
   std::atomic<value_type> _sum;
```
jsteemann, line 554: I don't think this change is required. Lower and higher should newer be modified after creating a histogram. If we do have code that does that, it seem wrong.

arangod/GeneralServer/AcceptorTcp.cpp
```diff
@@ -151,7 +152,10 @@ void AcceptorTcp<SocketType::Tcp>::asyncAccept() {
   // _asioSocket is nullptr.
   TRI_ASSERT(_endpoint->encryption() == Endpoint::EncryptionType::NONE);
 
-  _asioSocket.reset(new AsioSocket<SocketType::Tcp>(_server.selectIoContext()));
+  {
+    std::unique_lock<std::mutex> guard(_mutex);
+    _asioSocket = std::make_unique<AsioSocket<SocketType::Tcp>>(_server.selectIoContext());
```
jsteemann, line 157: In line 192 we still access `_asioSocket` without the mutex and read its `socket` and `peer` members. I think this is safe, because even though `_asioSocket` is initialized after line 157, no other thread can call `close()` on it and reset the `_asioSocket` member.
But it is a bit tricky to figure all this out.

I think keeping the `_asioSocket` member (in contrast to passing the asio socket in local variables) still makes sense, because I don't see a good other way of how `close()` would work.

arangod/GeneralServer/AcceptorTcp.cpp
```diff
@@ -165,8 +169,12 @@ void AcceptorTcp<SocketType::Tcp>::asyncAccept() {
     info.encryptionType = _endpoint->encryption();
     info.serverAddress = _endpoint->host();
     info.serverPort = _endpoint->port();
-
-    std::unique_ptr<AsioSocket<SocketType::Tcp>> as = std::move(_asioSocket);
+    
+    std::unique_ptr<AsioSocket<SocketType::Tcp>> as;
+    {
+      std::unique_lock<std::mutex> guard(_mutex);
+      as = std::move(_asioSocket);
```
jsteemann, line 176: I wonder if it is possible that another thread can call `close()` between the beginning of this callback and this line. If so, then the `close()` could reset the `_asioSocket` pointer, which we then use here and dereference.
Wouldn't it be better to lock the mutex at the beginning of the callback, move out `_asioSocket` under the lock, and then release it?
We would also need an extra check that the moved-out `_asioSocket` is not a nullptr, of course.

arangod/GeneralServer/AcceptorTcp.cpp
```diff
@@ -265,14 +273,22 @@ void AcceptorTcp<SocketType::Ssl>::asyncAccept() {
...
       return;
     }
 
-    performHandshake(std::move(_asioSocket));
+    decltype(_asioSocket) as;
+    {
+      std::unique_lock<std::mutex> guard(_mutex);
+      as = std::move(_asioSocket);
+    }
+    performHandshake(std::move(as));
```
jsteemann, line 291: We should check here if `as` is a nullptr. This could happen IMO if some other thread calls `close()` between the first few lines of the handler and line 286.

arangod/GeneralServer/AcceptorTcp.cpp
```diff
@@ -126,6 +126,7 @@ void AcceptorTcp<T>::open() {
 
 template <SocketType T>
 void AcceptorTcp<T>::close() {
+  std::unique_lock<std::mutex> guard(_mutex);
   if (_asioSocket) {
     _asioSocket->timer.cancel();
   }
```
jsteemann, line 132: I don't like holding the mutex for the entire duration of `close()`, but it looks safe and like it won't block any other place that also grabs this mutex. So I guess this works.

arangod/Cluster/HeartbeatThread.cpp
mpoeter: The problem is that this object is accessed by some other thread that is already running at the time that old code would initialize it. So the intention was to initialize it _before_ this other thread is started. If this thread is only needed in the agency we can add an according if.

arangod/RestServer/Metrics.h
mpoeter: No, these track the highest/lowest value ever recorded. Since these were not updated atomically, the previous implementation might report incorrect values.

mpoeter: These values are not required, so I changed the code to only track them in maintainer mode for debugging.

arangod/GeneralServer/AcceptorTcp.cpp
mpoeter: As discussed offline, all the issues related to `_asioSocket` can be avoided by completely removing that member since it is not required. The timer is only initialized once the the CommTask has been created, but at that point the `_asioSocket` has been moved from the acceptor into the CommTask anyway, so there is no point in cancelling the timer in `close()`.

3rdParty/iresearch/core/index/index_writer.hpp
```diff
@@ -845,7 +845,7 @@ class IRESEARCH_API index_writer
     std::atomic<size_t> active_count_; // number of active in-progress operations (insert/replace) (e.g. document instances or replace(...))
     std::atomic<size_t> buffered_docs_; // for use with index_writer::buffered_docs() asynchronous call
     format::ptr codec_; // the codec to used for flushing a segment writer
-    bool dirty_; // true if flush_all() started processing this segment (this segment should not be used for any new operations), guarded by the flush_context::flush_mutex_
```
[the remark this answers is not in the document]
mpoeter: Will do.

3rdParty/iresearch/core/utils/object_pool.hpp
```diff
@@ -131,7 +132,7 @@ class concurrent_stack : private util::noncopyable {
         return nullptr;
       }
 
-      new_head.node = head.node->next;
+      new_head.node = head.node->next.load(std::memory_order_relaxed);
```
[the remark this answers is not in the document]
mpoeter: `next` is set in `push`:
https://github.com/arangodb/arangodb/blob/215bf7cab476a7ea9d4250bbde86f40b9771e69f/3rdParty/iresearch/core/utils/object_pool.hpp#L147-L153
The problem is that nodes are _reused_. Suppose a thread is currently trying to pop a node, so it loaded the current `head`, but is interrupted before it can read `next`. Now some other thread is faster, pops the node and puts it on the free list. Then the node can be taken from the free list and reinserted into the stack (that's why we need ABA protection on the head). This push operation will overwrite the value of `next` in the reused node, and this write operation conflicts with the read in the first thread, resulting in a data race.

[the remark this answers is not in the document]
Dronplane: Yes, I see, but `head` and `new_head` are local variables on stack. And once load from `_head` is finished we have a copy here and no other writes to `_head` could affect us. Indeed `next` value may became totally invalid but as we don't dereference it we are safe.

arangod/Pregel/PregelFeature.h
```diff
@@ -90,6 +88,7 @@ class PregelFeature final : public application_features::ApplicationFeature {
  private:
   Mutex _mutex;
   std::unique_ptr<RecoveryManager> _recoveryManager;
+  std::atomic<RecoveryManager*> _recoveryManagerPtr;
```
neunhoef, line 91: I suggest to add a comment here to explain that `_recoveryManagerPtr` always points to the same as `_recoveryManager` and why we need a separate, atomic pointer here. For example, it was not clear to me on first reading.

arangod/Sharding/ShardingInfo.h
```diff
@@ -114,11 +115,11 @@ class ShardingInfo {
   size_t _numberOfShards;
 
   // @brief replication factor (1 = no replication, 0 = smart edge collection)
-  size_t _replicationFactor;
+  std::atomic<size_t> _replicationFactor;
```
neunhoef, line 118: Why does this need to be atomic? As far as I know it is only ever written to in the beginning and then never changed? Maybe add a comment to explain this.

arangod/Sharding/ShardingInfo.h
```diff
@@ -114,11 +115,11 @@ class ShardingInfo {
...
 
   // @brief write concern (_writeConcern <= _replicationFactor)
   // Writes will be disallowed if we know we cannot fulfill minReplicationFactor.
-  size_t _writeConcern;
+  std::atomic<size_t> _writeConcern;
```
neunhoef, line 122: Same here.

neunhoef (approved):
LGTM.See comments, some questions and a few suggestions to add a few more comments for the reader.

3rdParty/iresearch/core/utils/object_pool.hpp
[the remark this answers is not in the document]
mpoeter: True, we never dereference the value we read from `next`, and the subsequent CAS operation is doomed to fail anyway. However, that doesn't change the fact that we have two conflicting non-atomic operations (write + read) on `next` that are not ordered by a happens-before relation.

arangod/Pregel/PregelFeature.h
mpoeter: Added a comment.

arangod/Sharding/ShardingInfo.h
mpoeter: Added a comment.

arangod/Sharding/ShardingInfo.h
mpoeter: Added a comment.

3rdParty/iresearch/core/utils/object_pool.hpp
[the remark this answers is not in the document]
Dronplane: Ah, I thing I`ve got it. This conflict could happen if we are popping the same `node` object we are pushing from the other thread. But that could not happen due to logic of stack usage. Even more  - if due to some circumstances this will happen  - we are still not using obtained value. And you are correct to notice that CAS will fail and protect us. So adding atomic here would be a pessimization. Could we put here some TSan suppress comment?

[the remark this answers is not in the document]
mpoeter: > This conflict could happen if we are popping the same node object we are pushing from the other thread. But that could not happen due to logic of stack usage.

Actually it can (and it does) happen, because nodes are reused. That is also the reason why we need cmpxchg16 to protect against the ABA problem. Using atomic here is not really a pessimization since `memory_order_relaxed` is sufficient, so it will translate to the same instructions on virtually all architectures anyway. This is not just about TSan complaining, but this really is a data race as defined in the standard. Even though it is _most likely_ benign in almost all cases, strictly speaking it is still UB.

Dronplane (approved):
LGTM. Thx for elaboration with race! 👍

3rdParty/iresearch/core/utils/object_pool.hpp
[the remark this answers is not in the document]
mpoeter: For the sake of completeness here is a sequence that shows how the data race occurs:
  - suppose we have a stack a node A
  - thread 1 executes pop, so it reads `head` and receives a reference to A, but before it can read A's `next` pointer, the thread is interrupted
  - thread 2 also executes pop and removes A from the stack. The stack is now empty.
     -> A has now been completely removed from the stack, but thread 1 does not yet know that!
  - thread 2 now reuses A and pushes a new value to the stack. _This requires a write to A's `next` pointer._
  - thread 1 now continues and reads A's `next` -> this is a data race with the write in the previous step

gnusi:
@mpoeter thank you, looking forward to see a corresponding backport to IResearch upstream :heart:

neunhoef (approved):
LGTM

jsteemann (approved):
LGTM!

jsteemann (approved):

PR #14037, opened 2021-04-21
Feature/hybrid smart graph api

hkernbach:
### Scope & Purpose

Basic API adjustments to be able to create hybrid smart graphs.


arangod/Graph/Graph.cpp
```diff
@@ -116,6 +116,20 @@ Graph::Graph(velocypack::Slice const& slice, ServerDefaults const& serverDefault
   TRI_ASSERT(!_rev.empty());
 
   if (slice.hasKey(StaticStrings::GraphEdgeDefinitions)) {
+    // TODO: Check why we're landing here and not in SmartGraphEE - Cleanup!
+    LOG_DEVEL << "do not forget me - 1";
```
mchacki, line 120: You forgot me :P

arangod/Graph/Graph.cpp
```diff
@@ -116,6 +116,20 @@ Graph::Graph(velocypack::Slice const& slice, ServerDefaults const& serverDefault
...
+    if (slice.isObject()) {
```
mchacki, line 121: This if looks like it is in the wrong position.
I would have expected it on top level, before the check for GraphEdgeDefinitions.

arangod/Graph/Graph.cpp
```diff
@@ -146,6 +160,20 @@ Graph::Graph(TRI_vocbase_t& vocbase, std::string&& graphName,
   TRI_ASSERT(_rev.empty());
 
   if (info.hasKey(StaticStrings::GraphEdgeDefinitions)) {
+    // TODO: Check why we're landing here and not in SmartGraphEE - Cleanup!
+    LOG_DEVEL << "do not forget me - 2";
```
mchacki, line 164: You also forgot me!

arangod/Graph/Graph.cpp
```diff
@@ -146,6 +160,20 @@ Graph::Graph(TRI_vocbase_t& vocbase, std::string&& graphName,
...
+    if (options.isObject()) {
```
mchacki, line 165: This also looks like it is in the Wrong position.
The satellites are independent of GraphEdgeDefinitions

arangod/Graph/Graph.cpp
```diff
@@ -212,10 +240,31 @@ std::set<std::string> const& Graph::orphanCollections() const {
...
 
+std::set<std::string> const& Graph::satelliteCollections() const {
+  return _satelliteColls;
+}
+
 std::set<std::string> const& Graph::edgeCollections() const {
   return _edgeColls;
 }
 
+bool Graph::needsToBeSatellite(std::string edge) const {
```
mchacki, line 251:
```suggestion
bool Graph::needsToBeSatellite(std::string const& edge) const {
```

mchacki, line 251: Please move this into GraphEE enterprise code.

The community variant should always return false, no matter what

arangod/Graph/Graph.cpp
```diff
@@ -417,9 +466,10 @@ void EdgeDefinition::toVelocyPack(VPackBuilder& builder) const {
     builder.add(VPackValue(to));
   }
   builder.close();  // array
+  builder.add("type", VPackValue(getType()));
 }
 
-ResultT<EdgeDefinition> EdgeDefinition::createFromVelocypack(VPackSlice edgeDefinition) {
+ResultT<EdgeDefinition> EdgeDefinition::createFromVelocypack(VPackSlice edgeDefinition, std::set<std::string> const& satCollections) {
```
mchacki, line 472:
```suggestion
ResultT<EdgeDefinition> EdgeDefinition::createFromVelocypack(VPackSlice edgeDefinition, std::unordered_set<std::string> const& satCollections) {
```

mchacki: There is no need for the collections to be sorted. So we would loose some speed here

arangod/Graph/Graph.cpp
```diff
@@ -212,10 +240,31 @@ std::set<std::string> const& Graph::orphanCollections() const {
...
   return _edgeColls;
 }
 
+bool Graph::needsToBeSatellite(std::string edge) const {
+  bool useAsSatellite = false;
+
+  auto edgeDefinition = edgeDefinitions().at(edge);
+  for (auto const& from: edgeDefinition.getFrom()) {
+    if (satelliteCollections().find(from) != satelliteCollections().end()) {
+      useAsSatellite = true;
```
mchacki, line 257: I think you can just `return true` no need to check anything else as the value will never be set to false later

arangod/Graph/Graph.cpp
```diff
@@ -430,22 +480,47 @@ ResultT<EdgeDefinition> EdgeDefinition::createFromVelocypack(VPackSlice edgeDefi
...
+    TRI_ASSERT(it.isString());
+    if (satCollections.find(it.copyString()) != satCollections.end()) {
+      // found a sat collection defined in to's
+      foundToSat = true;
+    }
+
     toSet.emplace(it.copyString());
   }
 
+  if (foundFromSat && !foundToSat) {
```
mchacki, line 507: I think we need to move this special handling to GraphEE as well.

arangod/Graph/Graph.h
```diff
@@ -46,12 +46,23 @@ struct ServerDefaults;
 namespace graph {
 
 class EdgeDefinition {
+ private:
+  enum EdgeDefinitionType {
+    DEFAULT,
+    SMARTTOSMART,
```
mchacki, line 52: I would prefer if we can spend some `_`
e.g.
```suggestion
    SMART_TO_SMART,
```

that would make it more readable imo

arangod/Graph/Graph.h
```diff
@@ -46,12 +46,23 @@ struct ServerDefaults;
...
+    SATTOSAT,
+    SMARTTOSAT,
+    SATTOSMART
+  };
+
  public:
   EdgeDefinition(std::string edgeCollection_, std::set<std::string>&& from_,
-                 std::set<std::string>&& to_)
+                 std::set<std::string>&& to_,
```
mchacki, line 60: I do not want to change it here, just wondering, why is it a `std::set` ?
Do we need this to be sorted?

arangod/Graph/Graph.h
```diff
@@ -70,7 +81,7 @@ class EdgeDefinition {
   /// types of values.
   static Result validateEdgeDefinition(const velocypack::Slice& edgeDefinition);
 
-  static ResultT<EdgeDefinition> createFromVelocypack(velocypack::Slice edgeDefinition);
+  static ResultT<EdgeDefinition> createFromVelocypack(velocypack::Slice edgeDefinition, std::set<std::string> const& satCollections);
```
mchacki, line 84:
```suggestion
  static ResultT<EdgeDefinition> createFromVelocypack(velocypack::Slice edgeDefinition, std::unordered_set<std::string> const& satCollections);
```

I think sortedness is not required.

arangod/Graph/GraphManager.cpp
```diff
@@ -550,7 +555,18 @@ Result GraphManager::ensureCollections(Graph* graph, bool waitForSync) const {
       return res;
     } else {
       if (edgeCollectionsToCreate.find(vertexColl) == edgeCollectionsToCreate.end()) {
-        documentCollectionsToCreate.emplace(vertexColl);
+        if (graph->isHybrid()) {
+          if (satelliteDocumentCollectionsToCreate.find(vertexColl) ==
+                  satelliteDocumentCollectionsToCreate.end() &&
```
mchacki, line 560:
```suggestion
          if (
```

Please run format after this!

It is not necessary to check for existence in the set we are going to insert into, it actually is wrong here.
In case we find the collection in the set of satellites we also add it to document collections, which would be wrong.

Inserting into unordered_set twice is not an issue, the second time will be ignored.

arangod/Graph/GraphManager.cpp
```diff
@@ -550,7 +555,18 @@ Result GraphManager::ensureCollections(Graph* graph, bool waitForSync) const {
...
+              graph->satelliteCollections().find(vertexColl) !=
```
mchacki, line 561: Marge this if with the one above:
if hybrid() && collection found in satellites
this way you can eliminate one of the false cases below.

mchacki (changes requested):
Please see comments.
We may need to talk offline about some please ping me

arangod/Graph/GraphManager.cpp
hkernbach: changed manually.

arangod/Graph/GraphManager.cpp
hkernbach: thanks! done

arangod/Graph/Graph.h
hkernbach: True, but I would like to keep it sorted here to have sorted velocypack output (we're doing this for edgeColls, vertexColls as well)

arangod/Graph/Graph.h
hkernbach: Think this one might be same as your mention below:
> True, but I would like to keep it sorted here to have sorted velocypack output (we're doing this for edgeColls, vertexColls as well)

arangod/Graph/Graph.cpp
hkernbach: done ty, much better yes.

arangod/Graph/GraphManager.cpp
```diff
@@ -554,19 +554,13 @@ Result GraphManager::ensureCollections(Graph* graph, bool waitForSync) const {
...
-              graph->satelliteCollections().find(vertexColl) !=
-                  graph->satelliteCollections().end()) {
-            satelliteDocumentCollectionsToCreate.emplace(vertexColl);
-          } else {
-            documentCollectionsToCreate.emplace(vertexColl);
-          }
-        } else {
-          documentCollectionsToCreate.emplace(vertexColl);
-        }
+      if (graph->isHybrid() &&
```
mchacki, line 557: I think this is now broken.
You merged 3 ifs not 2.

The behavior should be the following:
if vertexCol is in edgeCollectionsToCreate => Do nothing
else if hybrid() and vertexCol is in satelliteCollections => emplace in satelliteToCreate
else emplace in documentsToCreate

mchacki (changes requested):
One bug inside.
Everything else LGTM

arangod/Graph/GraphManager.cpp
hkernbach: Thanks for spotting, done.

mchacki (approved):
LGTM

PR #14043, opened 2021-04-21
Bug fix/testing cluster startup wait for foxx

dothebart:
### Scope & Purpose

it seems we should do this to make sure foxx is actually up and running properly?

- [x] :hankey: Bugfix


dothebart:
added to other PR.

PR #14058, opened 2021-04-23
Fix assertion failure in fuerte

neunhoef:
We had an assertion failure in fuerte in 

  3rdParty\fuerte\src\H2Connection.cpp:746

The reason is that there may be requests with an expiry time of
Clock::time_point::max() and at destruction time, when all requests
must be aborted, these must be aborted, too. So far, we compared with
< and thus lost those with Clock::time_point::max() as expire time. Now we
use <=.

This is a trivial change and is covered by existing tests.
Otherwise, the PR speaks for itself.


jsteemann (approved):
LGTM!

PR #14061, opened 2021-04-23
Fix return code on shutting down in sendRequestRetry

neunhoef:
Currently, if a server that is shutting down is using sendRequestRetry,
an immediate error "Timeout" is reported. This is wrong and should
better be a 503 TRI_ERROR_SHUTTING_DOWN.

This was found in chaos tests with the new test agent.

- Report proper error on shutdown in sendRequestRetry.
- CHANGELOG.

### Scope & Purpose

#### Backports:

- [ ] No backports required
- [*] Back-/Forwardports required for: 3.7, devel, 3.6

### Testing & Verification

- [*] This change is a trivial rework / code cleanup without any further 
      test coverage.
- [*] This change is already covered by existing tests, such as chaos
      tests.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 v3.8.0 (XXXX-XX-XX)
 -------------------
 
+* Fixed proper return value in sendRequestRetry if server is shutting down.
+
 * Fixed issue BTS-373: ASan detected possible heap-buffer-overflow at
   arangodb::transaction::V8Context::exitV8Context().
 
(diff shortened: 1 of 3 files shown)


goedderz (approved):

mchacki (changes requested):
Code looks good and sensible to me.

However I am Missing a test.
The Method there has a unit-test file, I think it should be possible to simulate a Shutdown in there to make sure it is returned properly.

neunhoef:
So far, we cannot simulate a shutdown situation in the server mocks, as far as I see. It seems excessive to me to implement this just to test this particular return value in a unit test, since the benefit is smaller than the effort. And if I add any kind of failure point in `sendRequestRetry` (which is also not available for unittests, as far as I know), I change the very code I want to test. Therefore, I vote for not adding this test. What do you think?

mchacki (approved):
LGTM!

PR #14065, opened 2021-04-23
improve performance in case of conflicts in batch CRUD operation (#14…

jsteemann:
…008)

### Scope & Purpose

Improve performance of batch CRUD operations (insert, update, replace, remove) if some of the documents in the batch run into write-write conflicts. Rolling back partial operations in case of a failure is very expensive because it requires rebuilding RocksDB write batches for the transaction from scratch. Rebuilding write batches takes time proportional to the number of operations in the batch, and for larger batches the cost can be prohibitive. 

Now we are not rolling back write batches in some situations when this is not required, so that in many cases running into a conflict does not have that high overhead. There can still be issues when conflicts happen for index entries, but a lot of previously problematic cases should now work better.

This change also reduces the RocksDB-internal lock timeout for writing to keys locked by another transaction from 1s to 1ms. This will mean that operations that ran into a write-write conflict may fail quicker than before, and not wait and retry to acquire the locked key(s).

#### Backports:

- [x] No backports required

### Testing & Verification


neunhoef (approved):
LGTM. I think the change is a good one. The critical question is if this will lead to a lot more conflicts which are reported to the client. In existing applications using ArangoDB, these conflicts could have been brushed under the carpet by us waiting for a whole second to resolve the conflict. This will now no longer happen. I think not brushing these things under the carpet is the right behaviour for a database. If we get too many complaints, we will have to see what we do. I approve of the change.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1742,20 +1764,24 @@ Result RocksDBCollection::updateDocument(transaction::Methods* trx,
...
+      auto rIdx = static_cast<RocksDBIndex*>(it->get());
+      // if we already performed the preflight checks, there is no need to repeat the
+      // checks once again here
+      res = rIdx->update(*trx, mthds, oldDocumentId, oldDoc, newDocumentId, newDoc, options, /*performChecks*/ !performPreflightChecks);
+      needReversal = needReversal || rIdx->needsReversal();
+      if (!res.ok()) {
+        if (needReversal && !trx->isSingleOperationTransaction()) {
+          ::reverseIdxOps(_indexes, it,
+                          [&](RocksDBIndex* rid) {
+                            return rid->update(*trx, mthds, newDocumentId, newDoc, oldDocumentId, oldDoc, options, /*performChecks*/ true);
```
mpoeter, line 1780: I suppose we don't need to perform any checks in this case?
```suggestion
                            return rid->update(*trx, mthds, newDocumentId, newDoc, oldDocumentId, oldDoc, options, /*performChecks*/ false);
```

mpoeter (approved):

PR #14081, opened 2021-04-27
human-readable sizes outputs in client tools

jsteemann:
### Scope & Purpose

Make client tool output easier to digest for humans.

#### Backports:

- [x] No backports required

### Testing & Verification

  - [x] Added new C++ **Unit tests** (for formatting function)


lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+  if (value < 1000) {
+    if (value == 1) {
+      out = "1";
+      label ="byte";
+    } else {
+      out = std::to_string(value);
+      label = "bytes";
+    }
+  } else if (value < 1000000ULL) {
+    out = std::to_string((double) value / double(1000));
```
Simran-B, line 1977: We seem to use a mix of binary and decimal prefix units:
https://www.arangodb.com/docs/stable/administration-configuration.html#suffixes-for-numeric-options
Would it make sense to use a factor of 1024 and KiB, MiB, etc. here?

jsteemann: Yes, this would also work.
But I am not clear about why you sent me this docs link, because it lists _both_ types of units (base 10 and base 2) as valid and supported. So I am not clear why we should go with base 2 outputs here and not base 10.
Regardless of for which units (base 2 or base 10) we decide here, there will be someone who doesn't like it.

arangosh/Export/ExportFeature.cpp
```diff
@@ -304,7 +304,9 @@ void ExportFeature::start() {
     }
   }
 
-  std::cout << "Processed " << _collections.size() << " collection(s), wrote " << exportedSize
+  using arangodb::basics::StringUtils::formatSize;
+
+  std::cout << "Processed " << _collections.size() << " collection(s), wrote " << formatSize(exportedSize)
             << " byte(s), " << _httpRequestsDone << " HTTP request(s)" << std::endl;
```
mpoeter, line 310:
```suggestion
            << ", " << _httpRequestsDone << " HTTP request(s)" << std::endl;
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+  std::string out, label;
+  if (value < 1000) {
+    if (value == 1) {
+      out = "1";
+      label ="byte";
+    } else {
+      out = std::to_string(value);
+      label = "bytes";
+    }
+  } else if (value < 1000000ULL) {
```
mpoeter, line 1976:
```suggestion
  } else if (value < 1000'000ULL) {
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    out = std::to_string((double) value / double(1000));
+    label = "KB";
+  } else if (value < 1000000000ULL) {
```
mpoeter, line 1979:
```suggestion
  } else if (value < 1'000'000'000ULL) {
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    out = std::to_string((double) value / double(1000000));
```
mpoeter, line 1980:
```suggestion
    out = std::to_string((double) value / 1e6);
```
If you don't like the scientific notation I would vote for using `double(1'000'000)` instead.

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    label = "MB";
+  } else if (value < 1000000000000ULL) {
```
mpoeter, line 1982:
```suggestion
  } else if (value < 1'000'000'000'000ULL) {
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    out = std::to_string((double) value / double(1000000000));
```
mpoeter, line 1983:
```suggestion
    out = std::to_string((double) value / 1e9);
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    label = "GB";
+  } else if (value < 1000000000000000ULL) {
```
mpoeter, line 1985:
```suggestion
  } else if (value < 1'000'000'000'000'000ULL) {
```

lib/Basics/StringUtils.cpp
```diff
@@ -1963,6 +1963,37 @@ std::string escapeRegexParams(std::string const& in) {
...
+    out = std::to_string((double) value / double(1000000000000));
```
mpoeter, line 1986:
```suggestion
    out = std::to_string((double) value / 1e12);
```

mpoeter (approved):
LGTM! One minor issue, the other suggestions are only for improved readability.

lib/Basics/StringUtils.cpp
Simran-B: The base 2 variant is what I perceive as "correct" for everything computer-related. Base 10 doesn't make much sense to me, but unfortunately drive capacity is specified using base 10. I'm not sure how different OSes handle it, but under Windows, a 1 TB drive shows up with a capacity of ~930 GB (actually GiB) for instance. Personally, I'm in favor of base 2, but as you say, people have different opinions and we already have both bases in use in ArangoDB, so it's unclear what the preferred one should be. The binary prefixes are at least less ambiguous (see Windows).

PR #14084, opened 2021-04-27
Fix maintenance amok and HTTP responses

neunhoef:
This is a PR with very tiny fixes.

It does 2 things:

1. Fix removeDBErrors busy loop error.
2. Respond correctly on wrong HTTP version and too large body.

The first is a bug in the error reporting in case of a create database
has not worked. This error is currently reported in a busy loop to the
agency, if it happens. The fix clears the error after the first report.

The second is a bug in the parsing of HTTP headers and the corresponding
responses. In case the HTTP version is not 1.1 or 1.0, we now report
505 HTTP VERSION NOT SUPPORTED (and do not simply close off the
connection). In case that the content length is larger than 1 GB, we
now respond with a proper 413 REQUEST ENTITY TOO LARGE instead of
simply closing the connection.

This is the first easy step to fix these error responses. Therefore, we will not test
the correct responses in this PR. A future PR will test the complete behaviour under
wrong HTTP requests (which is not entirely fixed in this PR!).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,12 @@
 v3.8.0 (XXXX-XX-XX)
 -------------------
 
+* Fixed bug in error reporting when a database create did not work, which
+  lead to a busy loop reporting this error to the agency.
+
+* Fixed the error response if the HTTP version is not 1.0 or 1.1 and if
+  the Content-Length is too large (> 1 GB).
+
 * Guarded access only to ActionBase::_result.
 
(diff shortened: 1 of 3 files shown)


dothebart (approved):

jsteemann (approved):
LGTM!

neunhoef:
The last change is only the CHANGELOG, so no new Jenkins run is needed.

PR #14086, opened 2021-04-27
AQL micro optimizations

jsteemann:
### Scope & Purpose

AQL code micro optimizations

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/Aql/AqlCall.h
+++ b/arangod/Aql/AqlCall.h
@@ -217,9 +217,6 @@ struct AqlCall {
   }
 
   auto requestLessDataThan(AqlCall const& other) const noexcept -> bool;
-  
-  /// @brief reusable const empty call object
-  static AqlCall const emptyCall;
 };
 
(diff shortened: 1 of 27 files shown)


arangod/Aql/AqlCall.h
```diff
@@ -217,6 +217,9 @@ struct AqlCall {
   }
 
   auto requestLessDataThan(AqlCall const& other) const noexcept -> bool;
+  
+  /// @brief reusable const empty call object
+  static AqlCall const emptyCall;
```
mpoeter, line 222: I don't think this helps performancewise since `AqlCall` should be trivially copyable and therefore a default constructed instance should be similarly cheap. However, if you prefer it for readability I am fine with that.

mpoeter (approved):

arangod/Aql/AqlCall.h
jsteemann: Ok, I reverted all of the `AqlCall::emptyCall` changes.
This also simplifies the changes in this PR a bit.

PR #14098, opened 2021-04-29
Adding Unit Tests for Agency Store (TC-34)

ignacionr:
### Scope & Purpose

#### Backports:

- [x] No backports required

#### Related Information

- [x] GitHub issue / Jira ticket number: [TC-34](https://arangodb.atlassian.net/browse/TC-34)

### Testing & Verification

Link to Jenkins PR run:
https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr/15383/

### Documentation


tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,1347 @@
...
+     )=");
+  ASSERT_TRUE(velocypack::NormalizedCompare::equals(j->slice(), result.slice()));
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief Test transact interface
+////////////////////////////////////////////////////////////////////////////////
+/*
+TEST_F(StoreTestAPI, testPoll) {
```
neunhoef, line 107: This test is not needed. It only tests the log poll API, which is not present in the Store.

ignacionr: Yes, after our conversation there will be this and another one that I will remove. Thank you.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,1239 @@
...
+      writeAndCheck(
+        "[[" + localObj + R"(, {"baz":{"old":)" + baz_text + R"(},"foo":)" + foo_value + R"(,"qux":)" + qux + R"(}]])");
+      writeAndCheck(
+        "[[" + localObj + R"(, {"baz":{"old":)" + baz_text + R"(},"qux":)" + qux + R"(,"foo":)" + foo_value + R"(}]])");
+      writeAndCheck(
+        "[[" + localObj + R"(, {"qux":)" + qux + R"(,"baz":{"old":)" + baz_text + R"(},"foo":)" + foo_value + R"(}]])");
+
+      std::vector<std::string> localKeys;
+      std::transform(baz.begin(), baz.end(), std::back_inserter(localKeys), [](auto kv){ return kv.first; });
+      for (int permutation_count{}; permutation_count < 2000; ++permutation_count) {
```
neunhoef, line 306: 2000 is way too much and not necessary. 3 permutation will be more than enough. The code now seems to work.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,973 @@
...
+
+class StoreTestAPI : public ::testing::Test {
+ public:
+  StoreTestAPI() : _server(), _store(_server.server(), nullptr) {
+  }
+ protected:  
+  arangodb::tests::mocks::MockCoordinator _server;
+  arangodb::consensus::Store _store;
+
+  auto readAndCheck(std::string const &json)
```
neunhoef, line 49: The name seems to be misleading: It only reads. It does not check.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,973 @@
...
+      insert_value_array(ss, operations.front());
+      for (auto oi {operations.begin() + 1}; oi != operations.end(); ++oi) {
+        insert_value_array(ss, *oi);
+      }
+    }
+    ss << "]";
+    return write(ss.str());
+  }
+
+  auto transactAndCheck(std::string const &json) {
```
neunhoef, line 97: Name is misleading.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,973 @@
...
+    auto results { _store.applyTransactions(q)};
+    return results;
+  }
+
+  auto writeAndCheck(std::string const &json)
+  {
+    try {
+      auto r {write(json)};
+      auto applied_all = std::all_of(r.begin(), r.end(), [](auto const &result){ return result == consensus::apply_ret_t::APPLIED; });
+      if (!applied_all)
```
neunhoef, line 108: If you have a test, the `ASSERT_EQ` is enough, so you can remove the `if`. Besides, throwing here will be caught be the `catch` below which does not really seem to be sensible.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,973 @@
...
+        }
+      ASSERT_TRUE(applied_all);
+    }
+    catch (std::exception &ex) {
+      throw std::runtime_error(std::string(ex.what()) + " processing " + json);
+    }
+  }
+
+  void assertEqual(std::shared_ptr<velocypack::Builder> result, std::string const &expected_result) const {
+    try {
```
neunhoef, line 120: Again: Why `try` if you throw again anyway?

neunhoef (changes requested):
So far, I looked at the infrastructure. See comments, I suggest at least a rename.
I have not yet looked at the actual tests. Will try later today.

tests/Agency/StoreTestAPI.cpp
ignacionr: Because I am adding contextual information to whatever exception was generated on the called functions. Makes the error report a lot more useful.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+
+class StoreTestAPI : public ::testing::Test {
+ public:
+  StoreTestAPI() : _server(), _store(_server.server(), nullptr) {
+  }
+ protected:  
+  arangodb::tests::mocks::MockCoordinator _server;
+  arangodb::consensus::Store _store;
+
+  auto read(std::string const &json)
```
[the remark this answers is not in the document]
neunhoef: It is not necessary to change this, if you prefer the `auto`, just give it a thought.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+  }
+ protected:  
+  arangodb::tests::mocks::MockCoordinator _server;
+  arangodb::consensus::Store _store;
+
+  auto read(std::string const &json) 
+  {
+    try {
+      auto q {VPackParser::fromJson(json)};
+      auto result {std::make_shared<VPackBuilder>()};
```
neunhoef, line 53: Here, it is totally fine (and even preferable), since one can see that the type is a shared_ptr to a VPackBuilder.

tests/Agency/StoreTestAPI.cpp
neunhoef: ok

tests/Agency/StoreTestAPI.cpp
neunhoef: Fine, but then the `ASSERT_TRUE` is unnecessary, since the code will never be reached if `applied_all` is `false`.

tests/Agency/StoreTestAPI.cpp
neunhoef: No checks are done.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+    "[[" + localObj + R"(, {"foo":)" + foo_value + R"(,"baz":{"old":)" + baz_text + R"(},"qux":)" + qux + "}]]");
+  writeAndCheck(
+    "[[" + localObj + R"(, {"baz":{"old":)" + baz_text + R"(},"foo":)" + foo_value + R"(,"qux":)" + qux + R"(}]])");
+  writeAndCheck(
+    "[[" + localObj + R"(, {"baz":{"old":)" + baz_text + R"(},"qux":)" + qux + R"(,"foo":)" + foo_value + R"(}]])");
+  writeAndCheck(
+    "[[" + localObj + R"(, {"qux":)" + qux + R"(,"baz":{"old":)" + baz_text + R"(},"foo":)" + foo_value + R"(}]])");
+
+  std::vector<std::string> localKeys;
+  std::transform(baz.begin(), baz.end(), std::back_inserter(localKeys), [](auto kv){ return kv.first; });
```
neunhoef, line 327: Again, a matter of taste. I know this is considered "modern C++", but I prefer this style:

```
localKeys.reserve(baz.size));
for (auto const& p : baz ) {
  localKeys.push_back(p.first);
}
```

Yes, it is more lines, but it is easier to understand, IMHO.
No need to change, this is your style. Just wanted to mention it.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+  auto rd {std::random_device{}()};
+  auto g {std::mt19937(rd)};
+  for (int permutation_count{}; permutation_count < 5; ++permutation_count) {
+    std::shuffle(localKeys.begin(), localKeys.end(), g);
+    std::string const permuted = to_json_object(localKeys, 
+      [](std::string const &k){ return k; },
```
neunhoef, line 333: Another issue of style: For me, the type of `k` is `std::string const&` and therefore I put the `&` next to `const` and not next to `k`. Same for pointers, I would always write `int* x;` and not `int *x;`. I know, many do not do this, but here you go. And I would always refrain from doing things like:
```
  int *x, y;
```
We used to have coding guidelines in this way, but I do not know if they are still in force.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+      writeAndCheck(R"([[{"/version/c":{"op":"pop"}}]])"); // int before
+      assertEqual(read(R"([["version"]])"), R"([{"version":{"c":["hello"]}}])");
+      std::this_thread::sleep_for(std::chrono::milliseconds{3100});
+      assertEqual(read(R"([["version"]])"), R"([{}])");
+      writeAndCheck(R"([[{"/version/c":{"op":"pop"}}]])"); // int before
+      assertEqual(read(R"([["version"]])"), R"( [{"version":{"c":[]}}])");
+    }
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief Test "pop" operator
```
neunhoef, line 616: Wrong heading, needs to say "erase".

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+      assertEqual(read(R"([["/a"]])"), R"( [{"a":[2,4,9]}])");
+      writeAndCheck(R"([[{"a":{"op":"erase","pos":0}}]])");
+      assertEqual(read(R"([["/a"]])"), R"( [{"a":[4,9]}])");
+      writeAndCheck(R"([[{"a":{"op":"erase","pos":1}}],
+                     [{"a":{"op":"erase","pos":0}}]])");
+      assertEqual(read(R"([["/a"]])"), R"( [{"a":[]}])");
+    }
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief Test "pop" operator
```
neunhoef, line 671: Wrong heading, please check all of them again.

tests/Agency/StoreTestAPI.cpp
ignacionr: Agreed. The semantics were changed to match the name in this case.

tests/Agency/StoreTestAPI.cpp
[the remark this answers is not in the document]
ignacionr: Yes, it is a bit puzzling. On one hand, I heard you and I heard Bjarne say that it is friendlier on the reader (both human and compiler) to express the type you want at some point. On the other hand, experience and other authoritative voices such as Jason Turner tell me that ```auto``` makes refactoring in the long run much easier.

tests/Agency/StoreTestAPI.cpp
```diff
@@ -0,0 +1,977 @@
...
+////////////////////////////////////////////////////////////////////////////////
+/// @brief test preconditions
+////////////////////////////////////////////////////////////////////////////////
+TEST_F(StoreTestAPI, precondition) {
+  writeAndCheck(R"([[{"/a":12}]])");
+  assertEqual(read(R"([["/a"]])"), R"([{"a":12}])");
+  writeAndCheck(R"([[{"/a":13},{"/a":12}]])");
+  assertEqual(read(R"([["/a"]])"), R"([{"a":13}])");
+  auto res = write(R"([[{"/a":14},{"/a":12}]])"); // fail precond {"a":12}
+  ASSERT_EQ(consensus::apply_ret_t::PRECONDITION_FAILED, res.front());
```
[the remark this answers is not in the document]
ignacionr: Yes. It is better to have an assertion failed, than a stack trace dump.

tests/Agency/StoreTestAPI.cpp
ignacionr: Actually, I used to follow your convention (which is consistent with MFC, which occuppied my first decade of C++ practice). Then I noticed someone who I fail to remember who it was, around 2008, doing this ```std::string const &``` which at first seemed strange. At some point he gave an explanation that I thought was very smart, and started following this style naturally. Let me try to find who this person was, and what was the specific motiviation that made it better to write like that.

neunhoef (approved):
LGTM

PR #14105, opened 2021-04-29
fix some smaller issues in RecursiveLockers

jsteemann:
### Scope & Purpose

Fix smaller issues in RecursiveLocker implementations:
- honor the fact that a `lock()` operation can fail when the locker type is `TRY`. In this case it is not safe to unconditionally assume we have acquired the lock, but we need to check if we have acquired it.
- for the RecursiveReadLocker, the lock type is implicitly `BLOCKING` (as the class does not provide any API for checking if the lock was acquired successfully). However, it was not guaranteed that the constructor of RecursiveReadLocker actually acquired the lock successfully. If it didn't, the constructor was left _without_ having acquired the lock :no_mouth: 

#### Backports:

- [x] Backports required for: *3.8*, *3.7*

### Testing & Verification

### DIFF
--- a/lib/Basics/RecursiveLocker.h
+++ b/lib/Basics/RecursiveLocker.h
@@ -36,14 +36,15 @@ namespace arangodb {
 template <typename T>
 class RecursiveMutexLocker {
  public:
-  RecursiveMutexLocker( // recursive locker
-      T& mutex, // mutex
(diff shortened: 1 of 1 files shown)


jsteemann:
Will be merged into #14101.

PR #14113, opened 2021-04-30
Fix move after free bug in connection pool

neunhoef:
This fixes a relatively subtle use-after-free bug in the connection
pool. It only showed up in a specific cluster failure test and only
if the glibc malloc (as opposed to jemalloc) was used. A lock guard
destructor was using already freed memory and destroyed some
glibc/malloc internal thread local data structures.

 - Fix a use-after free.
 - CHANGELOG.

The fix speaks for itself and does not need any additional tests. It
was manually verified that this fixes the observed problem.

This is a forward port of #14110 from 3.8 to devel.


dothebart (approved):

jsteemann (approved):
LGTM!

PR #14119, opened 2021-04-30
Reduce the public surface of Agency Store (TC-62)

ignacionr:
### Scope & Purpose

As a first step before introducing backend changes on the Agency Store, reduce its public interface.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Jira ticket number: [TC-62](https://arangodb.atlassian.net/browse/TC-62)

### Testing & Verification

Link to Jenkins PR run: https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr/15453/

### Documentation

- [x] Added entry to *Release Notes*


arangod/Agency/Store.h
```diff
@@ -173,7 +163,17 @@ class Store {
   /// and ignoring multiple subsequent forward slashes
   static std::vector<std::string> split(std::string const& str);
 
+#if defined(ENABLE_MOCKS)
+public:
```
neunhoef, line 167: As far as I see this is not doing anything since we are anyway in `public` here.

arangod/Agency/Store.h
```diff
@@ -155,10 +149,6 @@ class Store {
   /// @brief Remove time to live entries for uri
```
neunhoef, line 149: As far as I see `applies` is only used in `Store.cpp` and could thus be private. Please check.

arangod/Agency/Store.h
neunhoef: Furthermore, this means that in the mocks we would make everything below public, too. Do we want this?

tests/IResearch/AgencyMock.cpp
```diff
@@ -22,6 +22,7 @@
 /// @author Vasiliy Nabatchikov
 ////////////////////////////////////////////////////////////////////////////////
 
+#define ENABLE_MOCKS
```
neunhoef, line 25: Please make the name more specific and add a comment what this is for.

neunhoef (changes requested):
See comments.

arangod/Agency/Store.h
ignacionr: Good catch with the first part (it was already in the public section); flipping the condition now.

tests/IResearch/AgencyMock.cpp
ignacionr: The original thought was that there could be more instances where code would be made public to enable mocking, so that it could apply elsewhere. On second thought, making it specific instead, poses less burden on the reader. Will incorporate your suggestion and make it MAKE_NOTIFY_OBSERVERS_PUBLIC.

arangod/Agency/Store.h
ignacionr: Yes, testing shows that the method is not needed outside of the class. Removing this one too.

neunhoef (approved):
LGTM

Issue #14122, opened 2021-05-03
Traversal broken


rm-minus-r-star:
## My Environment

* __ArangoDB Version__:        3.7.5
* __Storage Engine__:             RocksDB
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:   
* __Configuration__:               Single
* __Infrastructure__:               Install on GCP VM
* __Operating System__:        Ubuntu 20.04
* __Total RAM in your machine__:        128 GB
* __Disks in use__:        HDD
* __Used Package__:              Ubuntu .deb
* 
## Component, Query & Data

__Affected feature__:

__AQL query (if applicable)__:
```
LET J0 = (
    FOR junction in junctions
        FILTER HAS (junction['centerline_aliases'], '11929684')
           AND junction['centerline_aliases']['11929684'] == 0
    RETURN {
        'pfi': TO_NUMBER('11929684')
        , 'seq 0': junction['_id']
    }
)

FOR zp in J0
    FOR v, e, p IN 1 OUTBOUND zp['seq 0'] GRAPH 'Roads'
        PRUNE v['tr_road PFI'] != zp['pfi']
          AND NOT HAS(v['centerline_aliases'], TO_STRING(zp['pfi']))
    RETURN {
        'pfi': zp['pfi']
        , 'v0': p.vertices
    }
```

__AQL explain and/or profile (if applicable)__:
Not available

__Dataset__:
See detail in expected result below

__Size of your Dataset on disk__:
3-4 GB

__Replication Factor & Number of Shards (Cluster only)__:
N/A

## Steps to reproduce

1. Run the query

__Problem__:
"Query: AQL: missing variable #4 (zp) for node #11 (TraversalNode) while planning registers (exception location: /workl/Arangod/Aql/RegisterPlan.cpp:138. Please report this error to arangodb.com (while optimizing plan)(exception location: /work/ArangoDB/arangod/RestHandler/RestCursorHandler.cpp:311). Please report this error to arangodb.com"

__Expected result__:
Query above extracts a single point in the first stage:
_id:junctions/0356412
_rev:_cJOLQ9m---
_key:0356412
_type                    :  junctions
centerline_aliases:
    11929425 : 11
    11929684 : 0
    11929684 : 0

and yields: 
([
  {
    "pfi": 11929684,
    "seq 0": "junctions/0356412"
  }
]

I expect it to be able to traverse to
_id:centerlines/11929684-0001
_rev:_cI7br_K---
[...]

PR #14126, opened 2021-05-03
Fixed issue #14122

jsteemann:
### Scope & Purpose

Fixed issue #14122: when the optimizer rule "inline-subqueries" is applied, it may rename some variables in the query. The variable renaming was however not carried out for traversal PRUNE conditions, so the PRUNE conditions could still refer to obsolete variables, which would make the query fail with errors such as

    Query: AQL: missing variable ... for node ... while planning registers

#### Backports:

- [x] No backports required

#### Related Information

- [x] GitHub issue / Jira ticket number: #14122 

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,13 +1,13 @@
 v3.6.14 (XXXX-XX-XX)
 --------------------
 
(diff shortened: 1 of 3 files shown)


mchacki (approved):
LGTM!
good Catch

PR #14129, opened 2021-05-03
Fix potential deadlock in RecursiveReadLocker.

mpoeter:
### Scope & Purpose

Fixed a potential deadlock in RecursiveReadLocker. This could happen if 
  some thread A has acquired a read-lock, then some thread B signals that
  it wants to acquire a write-lock and now thread A recursively wants to
  acquire another read-lock. Since our ReadWriteLocks usually prefer writers,
  the second attempt to acquire the read-lock would block, waiting for the
  writer to finish while the writer would block, waiting for the first read
  to finish. To solve this, the RecursiveReadLock does no longer follow the
  principal that writes have precedence over reads.

#### Backports:

- [x] Backports required for: 3.8, 3.7, 3.6

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,20 @@
 devel
 -----
 
+* Fixed a potential deadlock in RecursiveReadLocker. This could happen if 
+  some thread A has acquired a read-lock, then some thread B signals that
+  it wants to acquire a write-lock and now thread A recursively wants to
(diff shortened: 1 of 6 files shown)


MBkkt:
Only I think that the problem solution -- removes the recursive capture lock usage?

PR #14141, opened 2021-05-05
update Snappy from 1.1.8 to 1.1.9

jsteemann:
### Scope & Purpose

Update bundled version of Snappy from 1.1.8 to 1.1.9.
Also require AVX support for building Snappy now. This should be safe, as ArangoDB 3.9 requires AVX support too.

#### Backports:

- [x] No backports required

### Testing & Verification

Must compile and link fine on all target platforms in CI.


3rdParty/CMakeLists.txt
```diff
@@ -36,13 +36,17 @@ function (add_snappy)
...
     set(SNAPPY_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/snappy/snappy-${SNAPPY_VERSION}")
     set(SNAPPY_SOURCE_DIR "${SNAPPY_SOURCE_DIR}" PARENT_SCOPE)
     set(SNAPPY_BUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/snappy/snappy-${SNAPPY_VERSION}" PARENT_SCOPE)
     set(SNAPPY_LIB "snappy" PARENT_SCOPE)
-    set(SNAPPY_BUILD_TESTS OFF CACHE BOOL "disable Snappy tests")
+    set(SNAPPY_FUZZING_BUILD OFF CACHE BOOL "build Snappy for fuzzing")
+    set(SNAPPY_BUILD_TESTS OFF CACHE BOOL "disable Snappy tests" FORCE)
+    set(SNAPPY_BUILD_BENCHMARKS OFF CACHE BOOL "disable Snappy benchmarks" FORCE)
     set(SNAPPY_INSTALL OFF CACHE BOOL "disable Snappy installation")
+    set(SNAPPY_REQUIRE_AVX ON CACHE BOOL "target processors with AVX support" FORCE)
```
dothebart, line 48: hm, though *our* packages will have AVX etc from now on, this will permanently trash the ability to build without?

CMakeLists.txt
```diff
@@ -1238,7 +1238,7 @@ include_directories(${ROCKSDB_INCLUDE_DIR})
 
 set(ROCKSDB_LIBS rocksdb;${SNAPPY_LIB})
 
-add_dependencies(rocksdb snappy-static)
+add_dependencies(rocksdb snappy)
```
dothebart, line 1241: afair we patched the snappy lib to have a distinguishable target for this?

3rdParty/CMakeLists.txt
jsteemann: @dothebart: yes, unless the CMakeLists.txt file(s) are modified.
See https://github.com/arangodb/docs/blob/main/3.9/release-notes-upgrading-changes39.md#architecture-requirements

CMakeLists.txt
jsteemann: I don't know. I don't see what we need it for. At least that library target is not available in upstream Snappy.

maierlars (approved):

CMakeLists.txt
jsteemann: Discussed this in a side channel with @dothebart : it is possible that we need an extra target for this to build with Ninja. However, building with Ninja is currently unsupported and fails for various other reasons. Thus we will not make a potential adjustment for Ninja in the context of this PR, so we can get rid of one extra patch on top of Snappy.

PR #14143, opened 2021-05-05
Improve exception safety for maintenance thread & shard unlocking

jsteemann:
### Scope & Purpose

Improve exception safety for cluster maintenance threads.
If the `nextState()` method in MaintenanceWorker.cpp threw an exception, the maintenance thread was stopped and not restarted. Now this is properly caught, and the action will be set to `FAILED` state.

In addition, shard unlock in maintenance actions that had to do with shards was potentially unsafe, because it was happening in an action's `setState()` method but could require several memory allocations, with potential to throw exceptions before unlocking is carried out.
Shard unlocking is now ensured by storing the shard name as an instance variable, so there do not need to be lookups and temporary string creations to find out the shard name. In SynchronizeShard.cpp, shard unlocking was also put into a scopeGuard so we will see a guaranteed execution even with exceptions.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *cluster tests*.

### DIFF
--- a/arangod/Cluster/ActionBase.h
+++ b/arangod/Cluster/ActionBase.h
@@ -233,6 +233,30 @@ private:
 
 };  // class ActionBase
 
+class ShardDefinition {
+ public:
+  ShardDefinition(ShardDefinition const&) = delete;
+  ShardDefinition& operator=(ShardDefinition const&) = delete;
+
+  ShardDefinition(std::string const& database, std::string const& shard);
+
+  std::string const& getDatabase() const noexcept {
+    return _database;
+  }
+  
+  std::string const& getShard() const noexcept {
+    return _shard;
+  }
+
+  bool isValid() const noexcept {
+    return !_database.empty() && !_shard.empty();
+  }
+
+ private:
+  std::string const _database;
+  std::string const _shard;
+};
+
 }  // namespace maintenance
 
 Result actionError(ErrorCode errorCode, std::string const& errorMessage);
--- a/arangod/Cluster/ActionBase.h
+++ b/arangod/Cluster/ActionBase.h
@@ -239,6 +239,8 @@ class ShardDefinition {
   ShardDefinition& operator=(ShardDefinition const&) = delete;
 
   ShardDefinition(std::string const& database, std::string const& shard);
+  
+  virtual ~ShardDefinition() = default;
 
   std::string const& getDatabase() const noexcept {
     return _database;
--- a/arangod/Cluster/CreateCollection.h
+++ b/arangod/Cluster/CreateCollection.h
@@ -33,7 +33,7 @@
 namespace arangodb {
 namespace maintenance {
 
-class CreateCollection : public ActionBase {
+class CreateCollection : public ActionBase, public ShardDefinition {
  public:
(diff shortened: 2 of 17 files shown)


arangod/Cluster/TakeoverShardLeadership.cpp
```diff
@@ -71,25 +71,20 @@ std::string stripServerPrefix(std::string const& destination) {
...
   if (!desc.has(COLLECTION)) {
     error << "collection must be specified. ";
   }
   TRI_ASSERT(desc.has(COLLECTION));
-
-  if (!desc.has(SHARD)) {
-    error << "shard must be specified";
+  
+  if (!ShardDefinition::isValid()) {
+    error << "database and shard must be specified. ";
```
kvahed, line 86:
```suggestion
    error << "database and shard must be specified.";
```

kvahed (approved):
LGTM

arangod/Cluster/TakeoverShardLeadership.cpp
jsteemann: The whitespace at the end is actually intentional, because error messages can be concatenated. This is the case for all maintenace action constructors, which simply do
```
error << "new error message. ";
```
The idea of the ". " at the end is to make concatenated messages easier to read. It is not my invention, it is already like that in all ctors already.

arangod/Cluster/CreateCollection.h
```diff
@@ -33,7 +33,7 @@
 namespace arangodb {
 namespace maintenance {
 
-class CreateCollection : public ActionBase {
+class CreateCollection : public ActionBase, public ShardDefinition {
```
gnusi, line 36: As `~ShardDefinition()` isn't marked as `virtual`, the following code won't destroy `CreateCollection` object correctly 
```
ShardDefinition *p = new CreateCollection(...);
delete p;
```
I suggest to use aggregation instead of inheritance for this case.

gnusi (changes requested):

arangod/Cluster/CreateCollection.h
jsteemann: We don't have any code that will delete maintenance actions via a pointer to a ShardDefinition, and I don't think we will add it any soon.
Anyway, I can make it virtual easily. 
Using aggregation will also work but requires adjusting all call sites that access `getDatabase()` and `getShard()`. Possible, but more code, more work, and potentially not necessary.

arangod/Cluster/ActionBase.h
```diff
@@ -233,6 +233,30 @@ class ActionBase {
 
 };  // class ActionBase
 
+class ShardDefinition {
+ public:
+  ShardDefinition(ShardDefinition const&) = delete;
+  ShardDefinition& operator=(ShardDefinition const&) = delete;
+
+  ShardDefinition(std::string const& database, std::string const& shard);
```
jsteemann, line 241:
```suggestion
  ShardDefinition(std::string const& database, std::string const& shard);
  
  virtual ~ShardDefinition() = default;
```

fceller (approved):
Virtual issue addressed and solved. LGTM

PR #14144, opened 2021-05-05
Improve exception safety for maintenance thread & shard unlocking

jsteemann:
### Scope & Purpose

Improve exception safety for cluster maintenance threads.
If the `nextState()` method in MaintenanceWorker.cpp threw an exception, the maintenance thread was stopped and not restarted. Now this is properly caught, and the action will be set to `FAILED` state.

In addition, shard unlock in maintenance actions that had to do with shards was potentially unsafe, because it was happening in an action's `setState()` method but could require several memory allocations, with potential to throw exceptions before unlocking is carried out.
Shard unlocking is now ensured by storing the shard name as an instance variable, so there do not need to be lookups and temporary string creations to find out the shard name. In SynchronizeShard.cpp, shard unlocking was also put into a scopeGuard so we will see a guaranteed execution even with exceptions.

#### Backports:

- [x] Backports required for 3.8: https://github.com/arangodb/arangodb/pull/14143

### Testing & Verification

- [x] This change is already covered by existing tests, such as *cluster tests*.

### DIFF
--- a/arangod/Cluster/ActionBase.h
+++ b/arangod/Cluster/ActionBase.h
@@ -232,6 +232,30 @@ private:
 
 };  // class ActionBase
 
+class ShardDefinition {
+ public:
+  ShardDefinition(ShardDefinition const&) = delete;
+  ShardDefinition& operator=(ShardDefinition const&) = delete;
+
+  ShardDefinition(std::string const& database, std::string const& shard);
+
+  std::string const& getDatabase() const noexcept {
+    return _database;
+  }
+  
+  std::string const& getShard() const noexcept {
+    return _shard;
+  }
+
+  bool isValid() const noexcept {
+    return !_database.empty() && !_shard.empty();
+  }
+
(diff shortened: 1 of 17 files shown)


arangod/Cluster/ActionBase.h
```diff
@@ -232,6 +232,30 @@ class ActionBase {
 
 };  // class ActionBase
 
+class ShardDefinition {
+ public:
+  ShardDefinition(ShardDefinition const&) = delete;
+  ShardDefinition& operator=(ShardDefinition const&) = delete;
+
+  ShardDefinition(std::string const& database, std::string const& shard);
```
jsteemann, line 240:
```suggestion
  ShardDefinition(std::string const& database, std::string const& shard);
  
  virtual ~ShardDefinition() = default;
```

neunhoef (approved):
LGTM

PR #14146, opened 2021-05-05
Fix bugs in fuerte with H2 and VST connections

neunhoef:
This fixes two bugs in fuerte with HTTP/2 and VST connections.

- Do not call callbacks more than once. This could happen since in HTTP/2 a request lingers in form of an HTTP/2 stream for some time longer after its callback has been called by a timeout. If the next timeout (for a different request) is hit, the callback could be called a second time. This is now prevented by keeping some more state.
- Fix bug with timeouts in H2 connections in fuerte. This was a simple typo in the tracking if we are reading and/or writing.
- CHANGELOG.

#### Backports:

- [*] Backports required for: 3.6, 3.7 and 3.8

### Testing & Verification

- [*] This change is already covered by existing tests: Fuerte tests


jsteemann (approved):
LGTM!

PR #14151, opened 2021-05-06
Small improvements

jsteemann:
### Scope & Purpose

A few small improvements made during debugging other issues:
- add a proper log message in takeOverLeadership if it encounters an unexpected situation. this log message can make aware of that there is an issue, and also provides details about the state
- in takeOverLeadership, make the changes to the internal followerinfo state atomic, so we do not have an invalid state if any modification goes wrong
- require at least 3.8 for sync-by-revisions protocol. technically this should not be necessary, as 3.7 will not offer collections with that protocol enabled. however, this was not reflected properly in the code.
- fail with an exception if we get an invalid pointer back from rocksdb. this is a precaution for an unlikely (potentially even impossible) situation, but it doesn't hurt.
- added hidden option --default-language-check so that the startup language check (`--default-language` value vs. persisted language value) can be bypassed. the default value for the option is `true`, meaning the check will take place normally. it is not recommended to turn it off except when debugging binary data from other deployments.

#### Backports:

- [x] Backports required for 3.8: https://github.com/arangodb/arangodb/pull/14150

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,10 @@
 devel
 -----
+
+* Log a proper message if an unexpected state is encountered when taking over
+  shard leadership. In addition, make the change to the internal followerinfo 
+  state atomic so that it cannot be semi-changed.
+
 * Fix response when isBuilding could not be removed from newly created
   collection, when agency precondition fails. This can happen, when own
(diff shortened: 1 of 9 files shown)


neunhoef (approved):
LGTM

PR #14163, opened 2021-05-08
wip for shard synchronization issues

jsteemann:
### Scope & Purpose

This is a work-in-progress branch to improve error detection and self-repair for shard synchronization.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/Cluster/MaintenanceFeature.h
+++ b/arangod/Cluster/MaintenanceFeature.h
@@ -304,6 +304,23 @@ class MaintenanceFeature : public application_features::ApplicationFeature {
    */
   arangodb::Result removeDBError(std::string const& database);
 
+
+  /// @brief remove all replication errors for a particular database
+  void removeReplicationError(std::string const& database);
+  
+  /// @brief remove all replication errors for a particular shard.
+  /// this will be called after a successful SynchronizeShard job for the shard
+  void removeReplicationError(std::string const& database, std::string const& shard);
+
+  /// @brief store a replication error for a particular shard.
+  /// this will be called after a failed SynchronizeShard job for the shard
+  void storeReplicationError(std::string const& database, std::string const& shard);
+
+  /// @brief return the number of replication errors for a particular shard.
+  /// note: we will return only those errors which happened not longer than
+  /// maxReplicationErrorsPerShardAge  ago
+  size_t replicationErrors(std::string const& database, std::string const& shard) const;
+
   /**
    * @brief copy all error maps (shards, indexes and databases) for Maintenance
    *
@@ -338,7 +355,15 @@ class MaintenanceFeature : public application_features::ApplicationFeature {
   /// @brief get n random db names
   std::unordered_set<std::string> pickRandomDirty(size_t n);
 
-
+  /// @brief maximum number of replication error occurrences that are kept per 
+  /// shard.
+  static constexpr size_t maxReplicationErrorsPerShard = 20;
+    
+  /// @brief maximum age of replication error occurrences that are kept per shard.
+  /// error occurrences older than this max age will be removed only lazily and
+  /// will not be considered when counting the number of errors.
+  static constexpr auto maxReplicationErrorsPerShardAge = std::chrono::minutes(20);
+  
  protected:
   void initializeMetrics();
 
@@ -451,12 +476,19 @@ class MaintenanceFeature : public application_features::ApplicationFeature {
   mutable arangodb::Mutex _dbeLock;
   /// @brief pending errors raised by CreateDatabase
   std::unordered_map<std::string, std::shared_ptr<VPackBuffer<uint8_t>>> _dbErrors;
-
+  
   /// @brief lock for shard version map
   mutable arangodb::Mutex _versionLock;
   /// @brief shards have versions in order to be able to distinguish between
   /// independant actions
   std::unordered_map<std::string, size_t> _shardVersion;
+  
+  /// @brief lock for replication error bucket
+  mutable arangodb::Mutex _replLock;
+  /// @brief shard replication errors { database => { shard => [ timestamps ] } }
+  /// we store up to  maxReplicationErrorsPerShard  errors per shard.
+  /// all errors for a shard will be cleared after a successful SynchronizeShard job
+  std::unordered_map<std::string, std::unordered_map<std::string, std::vector<std::chrono::steady_clock::time_point>>> _replErrors;
 
   std::atomic<std::chrono::steady_clock::duration> _pauseUntil;
 
--- a/arangod/Cluster/SynchronizeShard.cpp
+++ b/arangod/Cluster/SynchronizeShard.cpp
@@ -665,67 +665,59 @@ bool SynchronizeShard::first() {
   std::string planId = _description.get(COLLECTION);
   std::string const& shard = getShard();
   std::string leader = _description.get(THE_LEADER);
+ 
+  size_t failuresInRow = feature().replicationErrors(database, shard);
   
-  auto const& props = properties();
-  TRI_ASSERT(props.isObject());
-  TRI_ASSERT(props.hasKey("failuresInRow"));
+  if (failuresInRow >= 10) {
+    //      MaintenanceFeature::maxReplicationErrorsPerShard) 
+    auto& df = _feature.server().getFeature<DatabaseFeature>();
+    DatabaseGuard guard(df, database);
+    auto vocbase = &guard.database();
+    
+    auto collection = vocbase->lookupCollection(shard);
+    if (collection != nullptr) {
+      LOG_TOPIC("7a2cf", WARN, Logger::MAINTENANCE)
+          << "SynchronizeShard: synchronizing shard '" << database << "/" << shard
+          << "' for central '" << database << "/" << planId << "' encountered "
+          << failuresInRow << " failures in a row. now dropping follower shard for "
+          << "a full rebuild";
 
-  if (props.isObject() && props.hasKey("failuresInRow")) {
-    VPackSlice f = props.get("failuresInRow"); 
-    if (f.isNumber()) {
-      size_t failuresInRow = f.getNumber<size_t>();
-      if (failuresInRow >= 10) {
-        //      MaintenanceFeature::maxReplicationErrorsPerShard) 
-        auto& df = _feature.server().getFeature<DatabaseFeature>();
-        DatabaseGuard guard(df, database);
-        auto vocbase = &guard.database();
-        
-        auto collection = vocbase->lookupCollection(shard);
-        if (collection != nullptr) {
-          LOG_TOPIC("7a2cf", WARN, Logger::MAINTENANCE)
-              << "SynchronizeShard: synchronizing shard '" << database << "/" << shard
-              << "' for central '" << database << "/" << planId << "' encountered "
-              << failuresInRow << " failures in a row. now dropping follower shard for "
-              << "a full rebuild";
+      // remove all recorded failures, so in next run we can start with a clean state
+      _feature.removeReplicationError(getDatabase(), getShard());
 
-          // remove all recorded failures, so in next run we can start with a clean state
-          _feature.removeReplicationError(getDatabase(), getShard());
+      // drop shard (💥)
+      methods::Collections::drop(*collection, false, 3.0); 
+      result(TRI_ERROR_REPLICATION_WRONG_CHECKSUM);
+      return false;
+    }
+  }
 
-          // drop shard (💥)
-          methods::Collections::drop(*collection, false, 3.0); 
-          result(TRI_ERROR_REPLICATION_WRONG_CHECKSUM);
-          return false;
-        }
+  if (failuresInRow >= 4) {
+    double sleepTime = 2.0;
+    double add = 0.1;
+    for (size_t i = 0; i < failuresInRow; ++i) {
+      sleepTime += add;
+      add += 0.1;
+    }
+
+    sleepTime = std::min<double>(sleepTime, 15.0);
+    
+    LOG_TOPIC("40376", WARN, Logger::MAINTENANCE)
+        << "SynchronizeShard: synchronizing shard '" << database << "/" << shard
+        << "' for central '" << database << "/" << planId << "' encountered "
+        << failuresInRow << " failures in a row. delaying next sync by " 
+        << sleepTime << " s";
+
+    while (sleepTime > 0.0) {
+      if (feature().server().isStopping()) {
+        result(TRI_ERROR_SHUTTING_DOWN);
+        return false;
       }
-
-      if (failuresInRow >= 4) {
-        double sleepTime = 2.0;
-        double add = 0.1;
-        for (size_t i = 0; i < failuresInRow; ++i) {
-          sleepTime += add;
-          add += 0.1;
-        }
-
-        sleepTime = std::min<double>(sleepTime, 15.0);
-        
-        LOG_TOPIC("40376", WARN, Logger::MAINTENANCE)
-            << "SynchronizeShard: synchronizing shard '" << database << "/" << shard
-            << "' for central '" << database << "/" << planId << "' encountered "
-            << failuresInRow << " failures in a row. delaying next sync by " 
-            << sleepTime << " s";
-
-        while (sleepTime > 0.0) {
-          if (feature().server().isStopping()) {
-            result(TRI_ERROR_SHUTTING_DOWN);
-            return false;
-          }
-           
-          constexpr double sleepPerRound = 0.5;
-          // sleep only for up to 0.5 seconds at a time so we can react quickly to shutdown
-          std::this_thread::sleep_for(std::chrono::duration<double>(std::min(sleepTime, sleepPerRound)));
-          sleepTime -= sleepPerRound;
-        }
-      }
+       
+      constexpr double sleepPerRound = 0.5;
+      // sleep only for up to 0.5 seconds at a time so we can react quickly to shutdown
+      std::this_thread::sleep_for(std::chrono::duration<double>(std::min(sleepTime, sleepPerRound)));
+      sleepTime -= sleepPerRound;
     }
   }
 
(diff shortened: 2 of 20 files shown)


arangod/Cluster/MaintenanceFeature.h
```diff
@@ -338,7 +355,15 @@ class MaintenanceFeature : public application_features::ApplicationFeature {
...
 
-
+  /// @brief maximum number of replication error occurrences that are kept per 
+  /// shard.
+  static constexpr size_t maxReplicationErrorsPerShard = 20;
+    
+  /// @brief maximum age of replication error occurrences that are kept per shard.
+  /// error occurrences older than this max age will be removed only lazily and
+  /// will not be considered when counting the number of errors.
+  static constexpr auto maxReplicationErrorsPerShardAge = std::chrono::minutes(20);
```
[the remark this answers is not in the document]
jsteemann: Ok, I have bumped it up to 24 hours as suggested.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -665,6 +665,63 @@ bool SynchronizeShard::first() {
   std::string planId = _description.get(COLLECTION);
   std::string const& shard = getShard();
   std::string leader = _description.get(THE_LEADER);
+ 
+  size_t failuresInRow = feature().replicationErrors(database, shard);
+  
+  if (failuresInRow >= MaintenanceFeature::maxReplicationErrorsPerShard) {
```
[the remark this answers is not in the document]
jsteemann: :heavy_check_mark:

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -665,6 +665,63 @@ bool SynchronizeShard::first() {
...
+  if (failuresInRow >= 4) {
+    // shard synchronization has failed several times in a row.
+    // now step on the brake a bit. this blocks our maintenance thread, but currently
+    // there seems to be no better way to delay the execution of maintenance tasks.
+    double sleepTime = 2.0;
+    double add = 0.1;
+    for (size_t i = 0; i < failuresInRow; ++i) {
+      sleepTime += add;
+      add += 0.1;
+    }
```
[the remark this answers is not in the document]
jsteemann: Your formula looks a lot fancier than my loop, so I am happy to use your suggestion.

lib/Containers/MerkleTree.cpp
```diff
@@ -455,46 +537,66 @@ void MerkleTree<Hasher, BranchingBits, LockStripes>::remove(std::vector<std::uin
...
+      new MerkleTree<Hasher, BranchingBits>(*this));
 }
 
-template <typename Hasher, std::uint64_t const BranchingBits, std::uint64_t const LockStripes>
-std::unique_ptr<MerkleTree<Hasher, BranchingBits, LockStripes>>
-MerkleTree<Hasher, BranchingBits, LockStripes>::cloneWithDepth(std::uint64_t newDepth) const {
+template <typename Hasher, std::uint64_t const BranchingBits>
+std::unique_ptr<MerkleTree<Hasher, BranchingBits>>
+MerkleTree<Hasher, BranchingBits>::cloneWithDepth(std::uint64_t newDepth) const {
+  // arangod does not call this method!
```
neunhoef, line 531:
```suggestion
  // arangod does not call this method! That is, because it is rather dangerous:
  // To grow the depth, it has to modify `rangeMax` by multiplying `rangeMax-rangeMin` by
  // the `branchingFactor`. This can quickly lead to integer overflow! Do not use without
  // knowing exactly what you are doing!
```

neunhoef (commented):
So far LGTM, see one comment.

arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -416,11 +416,13 @@ Result RocksDBMetadata::serializeMeta(rocksdb::WriteBatch& batch,
 
   // Step 4. store the revision tree
   if (rcoll->needToPersistRevisionTree(maxCommitSeq)) {
+    output.clear();
+      
+    rocksdb::SequenceNumber seq =
+        rcoll->serializeRevisionTree(output, maxCommitSeq, force);
+    appliedSeq = std::min(appliedSeq, seq);
+
     if (coll.useSyncByRevision()) {
```
neunhoef, line 425: If `needToPersistRevisionTree` returns true, then `useSyncByRevision` will always return `true`. So we can leave out the check here and just assert it.

arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -468,6 +466,8 @@ Result RocksDBMetadata::serializeMeta(rocksdb::WriteBatch& batch,
         << "collection with objectId '" << rcoll->objectId() << "'";
```
neunhoef, line 466: It looks to me as if this `else` is now executed for a collection which does not have `useSyncByRevision` which will not really work, or am I mistaken?

neunhoef (changes requested):
Basically LGTM. See a few comments. One more, which did not fit anywhere: The RocksDB background thread logs when its sync takes more than 0.75s. Is this time still appropriate, since it is entirely possible that lots of revision trees have been hibernated?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -918,7 +918,10 @@ void RocksDBEngine::stop() {
     _backgroundThread->beginShutdown();
 
     if (_settingsManager) {
-      _settingsManager->sync(true);
+      try {
+        _settingsManager->sync(true);
+      } catch (...) {
```
gnusi, line 923: shouldn't we log an error?

jsteemann: Fixed

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -527,56 +524,84 @@ rocksdb::SequenceNumber RocksDBMetaCollection::serializeRevisionTree(
...
+    return {};
   });
 }
 
 void RocksDBMetaCollection::rebuildRevisionTree(std::unique_ptr<rocksdb::Iterator>& iter) {
-  std::unique_lock<std::mutex> guard(_revisionTreeLock);
+  auto newTree = allocateEmptyRevisionTree(revisionTreeDepth);
 
-  _revisionTree = allocateEmptyRevisionTree();
+  std::unique_lock<std::mutex> guard(_revisionTreeLock);
```
mpoeter, line 604: Wouldn't it be sufficient to acquire the lock later once we actually update our members?

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -592,40 +617,75 @@ void RocksDBMetaCollection::rebuildRevisionTree(std::unique_ptr<rocksdb::Iterato
...
   _revisionTreeApplied.store(seq);
   _revisionTreeCreationSeq = seq;
   _revisionTreeSerializedSeq = seq;
 }
 
-void RocksDBMetaCollection::revisionTreeSummary(VPackBuilder& builder) {
+void RocksDBMetaCollection::revisionTreeSummary(VPackBuilder& builder, bool fromCollection) {
   if (!_logicalCollection.useSyncByRevision()) {
     return;
   }
```
mpoeter, line 659: I suppose this check is obsolete.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -726,56 +791,115 @@ void RocksDBMetaCollection::applyUpdates(rocksdb::SequenceNumber commitSeq) {
...
-      if (!removals.empty()) {
-        _revisionTree->remove(removals);
-        removals.clear();
+        // apply removals, without holding the lock
+        // if this throws we will not have modified _revisionRemovalBuffers
+        try {
+          _revisionTree->remove(removeIt->second);
+        } catch (...) {
+          // this should never fail
+          TRI_ASSERT(false);
```
mpoeter, line 884: Ok, so this _should_ not fail, but shouldn't we at least write a log if it actually does?

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
jsteemann: Will be incorporated into https://github.com/arangodb/arangodb/pull/14169

lib/Containers/MerkleTree.cpp
```diff
@@ -920,15 +1171,112 @@ std::pair<std::uint64_t, std::uint64_t> MerkleTree<Hasher, BranchingBits, LockSt
...
+        output.append(reinterpret_cast<const char*>(&value), sizeof(uint64_t));
+        
+        value = basics::hostToLittle(src.hash);
+        output.append(reinterpret_cast<const char*>(&value), sizeof(uint64_t));
+      }
+    }
+  }
+
+  TRI_ASSERT(output.size() >= 3 * sizeof(uint64_t));
+  TRI_ASSERT((output.size() - 3 * sizeof(uint64_t)) % (sizeof(uint32_t) + 2 * sizeof(uint64_t)) == 0);
```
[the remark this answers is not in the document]
jsteemann: @neunhoef: I may not understand this, but doesn't 2^18 exceed the range of a uint16_t?

lib/Containers/MerkleTree.cpp
```diff
@@ -200,14 +227,106 @@ MerkleTree<Hasher, BranchingBits, LockStripes>::fromBuffer(std::string_view buff
...
+
+      ++prevOffset;
+    }
+    if (d == 0) { 
+      break;
+    }
+  }
+
+#ifdef PARANOID_TREE_CHECKS
+  tree->checkInternalConsistency();
```
mpoeter, line 320: Can we enable this for maintainer mode in general?

lib/Containers/MerkleTree.cpp
```diff
@@ -256,7 +375,9 @@ MerkleTree<Hasher, BranchingBits, LockStripes>::deserialize(velocypack::Slice sl
   }
 
   // allocate the tree
-  tree.reset(new MerkleTree<Hasher, BranchingBits, LockStripes>(maxDepth, rangeMin, rangeMax));
+  TRI_ASSERT(rangeMin < rangeMax);
+
+  tree.reset(new MerkleTree<Hasher, BranchingBits>(maxDepth, rangeMin, rangeMax));
```
mpoeter, line 380:
```suggestion
  tree = std::make_unique<MerkleTree<Hasher, BranchingBits>>(maxDepth, rangeMin, rangeMax);
```

lib/Containers/MerkleTree.cpp
```diff
@@ -290,30 +410,51 @@ MerkleTree<Hasher, BranchingBits, LockStripes>::deserialize(velocypack::Slice sl
...
+  if (rangeMax <= rangeMin) {
+    throw std::invalid_argument("rangeMax must be larger than rangeMin");
   }
+ 
   TRI_ASSERT(((rangeMax - rangeMin) / nodeCountAtDepth(maxDepth)) * nodeCountAtDepth(maxDepth) ==
              (rangeMax - rangeMin));
-  new (&meta()) Meta{rangeMin, rangeMax, maxDepth};
+  
+  // no lock necessary here
+  _buffer.reset(new uint8_t[allocationSize(maxDepth)]);
```
mpoeter, line 437:
```suggestion
  _buffer = std::make_unique<uint8_t[]>(allocationSize(maxDepth));
```

lib/Containers/MerkleTree.cpp
```diff
@@ -290,30 +410,51 @@ MerkleTree<Hasher, BranchingBits, LockStripes>::deserialize(velocypack::Slice sl
...
+  // cycles, so we only do it if absolutely necessary
+  checkInternalConsistency();
+#endif
 }
 
-template <typename Hasher, std::uint64_t const BranchingBits, std::uint64_t const LockStripes>
-MerkleTree<Hasher, BranchingBits, LockStripes>::~MerkleTree() {
+template <typename Hasher, std::uint64_t const BranchingBits>
+MerkleTree<Hasher, BranchingBits>::~MerkleTree() {
   std::unique_lock<std::shared_mutex> guard(_bufferLock);
```
mpoeter, line 458: No need to lock in the destructor!

lib/Containers/MerkleTree.cpp
```diff
@@ -455,46 +608,69 @@ void MerkleTree<Hasher, BranchingBits, LockStripes>::remove(std::vector<std::uin
...
+      new MerkleTree<Hasher, BranchingBits>(*this));
+}
+
+template <typename Hasher, std::uint64_t const BranchingBits>
+std::unique_ptr<MerkleTree<Hasher, BranchingBits>>
+MerkleTree<Hasher, BranchingBits>::cloneWithDepth(std::uint64_t newDepth) const {
+  // arangod does not call this method! That is, because it is rather dangerous:
+  // To grow the depth, it has to modify `rangeMax` by multiplying `rangeMax-rangeMin` by
+  // the `branchingFactor`. This can quickly lead to integer overflow! Do not use without
+  // knowing exactly what you are doing!
```
mpoeter, line 646: Perhaps we should remove this method if it is not used anymore?

lib/Containers/MerkleTree.cpp
```diff
@@ -635,40 +836,46 @@ void MerkleTree<Hasher, BranchingBits, LockStripes>::serialize(velocypack::Build
...
+
   TRI_ASSERT(output.empty());
+    
   if (compress) {
-    snappy::Compress(reinterpret_cast<char*>(_buffer.get()),
-                     allocationSize(meta().maxDepth), &output);
-    output.push_back(::CompressedCurrent);
+    if (this->node(0).count <= 15'000) {
+      // 15'000 is an arbitrary cutoff value for when to move from
+      // bottom-most level compression to full tree compression with Snappy
```
mpoeter, line 856: 15k seems pretty low to me. With depth 6 the bottom-most level has ~262k nodes. Even if every document is assigned to a different node, only ~6% of the nodes have any data.

lib/Containers/MerkleTree.cpp
```diff
@@ -920,15 +1171,112 @@ std::pair<std::uint64_t, std::uint64_t> MerkleTree<Hasher, BranchingBits, LockSt
...
+    if (src.count > 0) {
+      // index
+      {
+        uint32_t value = basics::hostToLittle(static_cast<uint32_t>(index));
+        output.append(reinterpret_cast<const char*>(&value), sizeof(uint32_t));
+      }
+      // count / hash
+      {
+        uint64_t value = basics::hostToLittle(src.count);
+        output.append(reinterpret_cast<const char*>(&value), sizeof(uint64_t));
```
mpoeter, line 1259: Since this encoding is only used for trees with a rather small number of entries we could squeeze offset and count into the same 64-bit value. If we keep the cutoff below 16k entries even a 32-bit value would suffice. This would further decrease the required memory.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -954,81 +1076,178 @@ bool RocksDBMetaCollection::haveBufferedOperations() const {
...
 
-std::size_t RocksDBMetaCollection::revisionTreeDepth() const {
-  std::size_t const treeCount = _revisionTree ? _revisionTree->count() : 0;
-  std::size_t bufferInserts = 0;
-  std::size_t bufferRemovals = 0;
-  {
-    std::unique_lock<std::mutex> guard(_revisionBufferLock);
+std::unique_ptr<containers::RevisionTree> RocksDBMetaCollection::allocateEmptyRevisionTree(std::size_t depth) const {
+  // should have _revisionTreeLock held outside
+  return std::make_unique<containers::RevisionTree>(depth, _logicalCollection.minRevision().id());
```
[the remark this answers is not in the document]
jsteemann: Indeed, the comment is now outdated. It used to be necessary before, because the function previously checked the tree's current depth. Will remove the comment.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
```diff
@@ -206,9 +206,14 @@ static void JS_CollectionRevisionTreeSummary(v8::FunctionCallbackInfo<v8::Value>
     TRI_V8_THROW_EXCEPTION_INTERNAL("cannot extract collection");
   }
 
+  bool fromCollection = false;
+  if (args.Length() > 0) {
+    TRI_ObjectToBoolean(isolate, args[0]);
```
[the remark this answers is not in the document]
jsteemann: Indeed. Thanks

arangod/VocBase/Identifiers/RevisionId.cpp
```diff
@@ -93,6 +93,14 @@ void RevisionId::toPersistent(std::string& buffer) const {
   rocksutils::uint64ToPersistent(buffer, id());
 }
 
+/// @brief create a revision id with a lower-bound HLC value
+RevisionId RevisionId::lowerBound() { 
+  // "2020-01-01T00:00:00.000Z" => 1577836800000 milliseconds since the epoch
+  RevisionId value{uint64_t(1577836800000) << 20ULL};
```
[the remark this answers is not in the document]
jsteemann: This will be used as an HLC value, and in our HLC implementation the highest 44 bits contain the timestamp, and the lowest 20 bits contain a counter value in case we need multiple values for the same timestamp.

lib/Containers/MerkleTree.cpp
jsteemann: Yes, we could, but this will make tests a lot slower, and I am not sure if we want to do it or not.

lib/Containers/MerkleTree.cpp
jsteemann: Gut, will use it.

lib/Containers/MerkleTree.cpp
jsteemann: Yes. But at 15k nodes to store, we would need 15kb x 20 bytes, i.e. 300kb already.
Snappy will compress this to about the same size, and is potentially faster than our own implementation. But definitely the cut-off value is very arbitrary.

lib/Containers/MerkleTree.cpp
jsteemann: Yes, we can still improve here.
I think it makes sense to improve the bottom-most format in devel when there is time, and then backport it to 3.8 as an additional encoding variant if it is significantly better.

PR #14168, opened 2021-05-11
Bug fix 3.8/pregel lifetime management

mpoeter:
### Scope & Purpose

Backport of #14157 

Fixes pregel lifetime management. Previously shutting down the server while a pregel job was still running could result in a segfault or a shutdown hanger.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/Pregel/Conductor.h
```diff
@@ -133,7 +133,8 @@ class Conductor {
             std::vector<CollectionID> const& vertexCollections,
```
gnusi, line 133: since you've started using `shared_from_this()` please, make `Conductor` constructor private, introduce a factory method returning `std::shared_ptr<Conductor>` and leave corresponding comment.

gnusi (changes requested):

arangod/Pregel/Conductor.h
mpoeter: Unfortunately when making the constructor private we can no longer use `make_shared`.

goedderz: You could use this workaround: https://godbolt.org/z/K85cv1xGq

PR #14177, opened 2021-05-12
Bug fix3.8/window operation issues

mpoeter:
### Scope & Purpose

Fix various issues related to the new WINDOW operation (see BTS-402)
  - Improved explain output for ISO 8601 duration strings and fixed missing week component.
  - Improved validation of input data and error messages.
  - Prevent FILTERs from being moved beyond a WINDOW.

#### Backports:

- [ ] No backports required
- [x] Forwardport required for: *devel*

#### Related Information

- [x] Jira ticket number: BTS-402
- [X] Docu PR: https://github.com/arangodb/docs/pull/695 

### Testing & Verification

### DIFF
--- a/arangod/Aql/WindowNode.cpp
+++ b/arangod/Aql/WindowNode.cpp
@@ -292,6 +292,8 @@ void WindowBounds::toVelocyPack(VPackBuilder& b) const {
             } else {
               result.append(std::to_string(duration.seconds)).push_back('.');
               auto ms = std::to_string(duration.milliseconds);
+              // parseIsoDuration already limits the number of decimals in milliseconds
+              TRI_ASSERT(ms.size() <= 3);
               result.append(3 - ms.size(), '0').append(ms).push_back('S');
             }
           }
--- a/arangod/Aql/WindowNode.cpp
+++ b/arangod/Aql/WindowNode.cpp
@@ -291,7 +291,8 @@ void WindowBounds::toVelocyPack(VPackBuilder& b) const {
               append(duration.seconds, 'S');
             } else {
               result.append(std::to_string(duration.seconds)).push_back('.');
-              append(duration.milliseconds, 'S');
+              auto ms = std::to_string(duration.milliseconds);
+              result.append(3 - ms.size(), '0').append(ms).push_back('S');
             }
           }
           if (result == "P") {
--- a/arangod/Aql/WindowNode.cpp
+++ b/arangod/Aql/WindowNode.cpp
@@ -241,26 +241,50 @@ WindowBounds::Row WindowBounds::calcRow(AqlValue const& input, QueryWarnings& w)
 
 void WindowBounds::toVelocyPack(VPackBuilder& b) const {
   switch (_type) {
-    case Type::Row:
-      b.add("preceding", VPackValue(_numPrecedingRows));
-      b.add("following", VPackValue(_numFollowingRows));
+    case Type::Row: {
+      auto translate = [](int64_t v) {
+        if (v == std::numeric_limits<int64_t>::max()) {
+          return VPackValue("unbounded");
+        }
+        return VPackValue(v);
+      };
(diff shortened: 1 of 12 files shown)


arangod/Aql/WindowNode.cpp
```diff
@@ -241,26 +256,50 @@ WindowBounds::Row WindowBounds::calcRow(AqlValue const& input, QueryWarnings& w)
...
+          append(duration.days, 'D');
+          if (duration.hours != 0 || duration.minutes != 0 || duration.seconds != 0 || duration.milliseconds != 0) {
+            result.push_back('T');
+            append(duration.hours, 'H');
+            append(duration.minutes, 'M');
+            if (duration.milliseconds == 0) {
+              append(duration.seconds, 'S');
+            } else {
+              result.append(std::to_string(duration.seconds)).push_back('.');
+              append(duration.milliseconds, 'S');
```
jsteemann, line 294: I think we have to zero-pad the milliseconds to 3 decimal places, so that we stringify one millisecond as `0.001S` and not `0.1S`.

arangod/Aql/WindowNode.cpp
```diff
@@ -241,26 +256,50 @@ WindowBounds::Row WindowBounds::calcRow(AqlValue const& input, QueryWarnings& w)
...
+            if (v != 0) {
+              result.append(std::to_string(v)).push_back(suffix);
+            }
+          };
+          append(duration.years, 'Y');
+          append(duration.months, 'M');
+          append(duration.weeks, 'W');
+          append(duration.days, 'D');
+          if (duration.hours != 0 || duration.minutes != 0 || duration.seconds != 0 || duration.milliseconds != 0) {
+            result.push_back('T');
```
jsteemann, line 287: If we only have accumulated `P` in the result here, we would next append the `T`.
Is `PT1H` a valid ISO duration, or does one have to add a date part to it?

arangod/Aql/grammar.y
```diff
@@ -201,6 +201,49 @@ bool validateAggregates(Parser* parser, AstNode const* aggregates,
   return true;
 }
 
+
+/// @brief validate the aggregate variables expressions
```
jsteemann, line 205:
```suggestion
/// @brief validate the WINDOW specification
```

arangod/Aql/WindowNode.cpp
mpoeter: Yes, `PT1H` is a valid duration string. Everything is optional, but there has to be at least one component. And if that is a time component, the string must include the `T` prefix.

arangod/Aql/WindowNode.cpp
mpoeter: Good catch!

mpoeter: Should be fixed.

arangod/Aql/WindowNode.cpp
```diff
@@ -291,7 +291,8 @@ void WindowBounds::toVelocyPack(VPackBuilder& b) const {
               append(duration.seconds, 'S');
             } else {
               result.append(std::to_string(duration.seconds)).push_back('.');
-              append(duration.milliseconds, 'S');
+              auto ms = std::to_string(duration.milliseconds);
+              result.append(3 - ms.size(), '0').append(ms).push_back('S');
```
jsteemann, line 295: Not sure if this is safe in all cases, e.g. if milliseconds has a value of 1000 or more.
Maybe this is not possible because milliseconds are bounded elsewhere. But it is not immediately clear here if that is the case. So I suggest we limit `ms` to a length of 3 characters here to be on the safe side.

mpoeter: Yes, the number of decimals is limited to 3 in `parseIsoDuration`, but I can at least add an assertion.

jsteemann (approved):
LGTM!

KVS85 (approved):
LGTM.

PR #14182, opened 2021-05-12
Agency Store - Performance Tests and Baseline

ignacionr:
### Scope & Purpose

#### Backports:

- [x] No backports required

#### Related Information

- [x] GitHub issue / Jira ticket number: TC-70

### Testing & Verification

Link to Jenkins PR run: https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr/15510/

### Documentation


tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2020 ArangoDB GmbH, Cologne, Germany
```
neunhoef, line 4: Should be 2021 by now.
```suggestion
/// Copyright 2014-2021 ArangoDB GmbH, Cologne, Germany
```

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Ignacio Rodriguez
+/// @author Copyright 2017, ArangoDB GmbH, Cologne, Germany
```
neunhoef, line 22: 2021
```suggestion
/// @author Copyright 2021, ArangoDB GmbH, Cologne, Germany
```

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+}
+
+TEST_F(StorePerformanceTest, single_deep_key_writes_and_reads) {
+  std::map<std::string, std::string> obj{};
+  auto const key {"a/b/c/d/e/f/g/h"};
+  auto const json_read_key {std::string("[[\"") + key + "\"]]"};
+  for (int i{}; i < hammer_times; ++i) {
+    obj[key] = std::to_string(i);
+    write(std::vector<std::vector<std::string>>{{to_json_object(obj)}});
```
neunhoef, line 189: Better take the construction of the JSON string and the vectors out of the loop, we only want to measure the Store operations.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+}
+
+// write lots of different keys in different places
+//  Random-access performance
+
+static std::string rand_path(const size_t len = 10u) {
+  static auto const alphabet {"abcdefghijklmnopqrstuvwxyz0123456789"};
+  std::string result;
+  std::generate_n(std::back_inserter(result), 1 + std::rand() % (len-1), [](){
```
neunhoef, line 211: From cppreference: "rand() is not recommended for serious random-number generation needs. It is recommended to use C++11's random number generation facilities to replace rand(). (since C++11)"

neunhoef: And: I would generate the keys first and store them beforehand such that the actual measurement only measures the Store operations themselves.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+// do a lot of small transactions
+// Performance in situations that should smartly use caches and primary memory.
+TEST_F(StorePerformanceTest, small_tx_r) {
+  writeAndCheck("[[{\"a\": 1}]]");
+  writeAndCheck("[[{\"b/b/c\": 2}]]");
+  writeAndCheck("[[{\"d\": 3}]]");
+  for (int i{}; i < hammer_times; ++i) {
+    auto const result {read("[[\"a\"], [\"b/b/c\"], [\"d\"]]")};
+    ASSERT_EQ(result->toJson(), "[{\"a\":1},{\"b\":{\"b\":{\"c\":2}}},{\"d\":3}]");
```
neunhoef, line 271: We should not assert in a tight loop which is supposed to measure the Store performance.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+// multiple threads manipulate values
+TEST_F(StorePerformanceTest, multiple_threads_all_separate_keys) {
+  std::vector<std::thread> threads;
+  std::generate_n(std::back_inserter(threads), hammer_times, [this](){return std::thread([this](){
+    auto const key {rand_path()};
+    auto const json_object {"{\"" + key + "\": " + std::to_string(rand()) + "}"};
+    auto result {write(std::vector<std::vector<std::string>>{{json_object}})};
+    ASSERT_EQ(result.front(), consensus::apply_ret_t::APPLIED);
+  });});
+  for (auto& thread: threads) {
```
neunhoef, line 403: I think we should make sure that all threads start at the same time and use a condition variable or barrier to let them actually start work at the same time.

neunhoef: I would be interested in individual time measurements of each thread in addition to an overall measurement.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+#include <cstdlib>
+#include <algorithm>
+
+namespace arangodb {
+namespace tests {
+namespace store_performance_test {
+
+auto const scattered_times{50};
+auto const hammer_times {200};
```
neunhoef, line 44: This is way too little for any meaningful measurement.

tests/Agency/StorePerformanceTest.cpp
neunhoef: It is a mistake to do only one call to the Store in each thread. In this way, thread creation and joining will vastly dominate the actual execution time rendering any measurement useless to judge the actual Store performance. I think we should vary the number of threads and each thread should - concurrently - run many executions. We probably also need some kind of multi-threaded protection, since the Store itself is not thread-safe as far as I remember.

neunhoef (changes requested):
I do not see any actual time measurements. I would have thought that we use the `std::chrono::steady_clock` or maybe the `std::chrono::high_resolution_clock` and then do some sensible statistics to get a good performance measurement, but I do not see any.

I know that the testing framework measures the time of each test, but this includes time to set up the test, to set up the Store, and in the end shutdown everything again. This is probably not accurate enough. And for the multi-threaded case we have the thread startup and joining time which we want to take out of the measurement.

tests/Agency/StorePerformanceTest.cpp
ignacionr: About the first comment, actually this was a conscious option. In this case I want to easily obtain the same items time and again, which makes rand() ideal (we only leave the default seed or force it with srand, and we get the exact same pseudorandomness). "Serious" in terms of randomness is not what I am after here.

ignacionr: To the second comment, of course; actually if you look at some of the tests that is the approach I am taking.

tests/Agency/StorePerformanceTest.cpp
ignacionr: I thought about that, but then I also thought that a smart implementation could eventually delay some operations until the result is actually used, in which case the test would become invalid. What are your thoughts?

tests/Agency/StorePerformanceTest.cpp
ignacionr: I agree. This numbers were separated into constants specifically to be able to tune on our expectations. If these tests are to run with every Jenkins build, we still need to be cautious.

tests/Agency/StorePerformanceTest.cpp
ignacionr: Agreed! I can spare a bit of memory for it.

gnusi:
@neunhoef @ignacionr it seems a bit weird that you use gtest for micro-benchmarks.
 There is awesome library from Google 
https://github.com/google/benchmark

I suggest we introduce a separate target for such micro-benchmarks as that'd be useful for others as well.

ignacionr:
> @neunhoef @ignacionr it seems a bit weird that you use gtest for micro-benchmarks.
> There is awesome library from Google
> https://github.com/google/benchmark
> 
> I suggest we introduce a separate target for such micro-benchmarks as that'd be useful for others as well.

Even though this is a great suggestion, the topic we are tackling right now is a back-end replacement of the Agency Store. As we are heavily using tests to make sure that the established semantics of the current store implementation are kept consistent when the change happens, it is simple and direct to add other tests that at least to some extent make sure that performance is not lost.

So this is not truly aimed as a general solution for profiling, but rather as part of an ongoing specific effort.

tests/Agency/StorePerformanceTest.cpp
ignacionr: This last point is truly important; will address it.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,651 @@
...
+  }
+
+  template <typename TSource>
+  static std::string to_json_object(TSource const& src) {
+    return to_json_object(
+        src, [](auto const& element) { return element.first; },
+        [](auto const& element) { return element.second; });
+  }
+
+  void threaded_writes(std::function<std::string()> value_generator) {
```
neunhoef, line 289: Unfortunatly, this measurement is not good for multiple threads. Let me explain why:
1. You measure the points in time at which any thread **finishes** an operation. That means that you do not measure the individual latency of each thread independently with your class. Rather, you measure the time when the first thread finishes, then the time from when the first thread finished to the time when the second finished etc. This is not what we want.
2. Furthermore, even if you correct this, you only measure the complete time a thread took and not the individual latency of each write. I think each thread should have its own `OperationMeasurement` class and then you want to do some reporting for all threads.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,651 @@
...
+        points.at(1 + thread_index) = OperationMeasurement::Clock::now();
+      });});
+    std::unique_lock<std::mutex> lk{m_ready};
+    cv_ready.wait(lk, [&worker_count, count](){ 
+      return worker_count.load() == count;
+    });
+    
+    std::this_thread::sleep_for(std::chrono::milliseconds{50});
+    points[0] = OperationMeasurement::Clock::now();
+    cv_start.notify_all();
```
neunhoef, line 315: This startup and finish is nice. It takes the thread creation latency out of the equation, which is exactly, what we want.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,651 @@
...
+  for (auto const& wq: write_queries) {
+    write(wq);
+    ++op;
+  }
+  op.report();
+}
+
+// test for contention:
+// multiple threads manipulate values
+TEST_F(StorePerformanceTest, multiple_threads_all_separate_keys) {
```
neunhoef, line 635: We probably want to vary a bit the number of threads and maybe try with 1, 2, 4, 8, 16 and 32 threads to see contention building up (or not).

neunhoef (changes requested):
See comments, the multithreading needs some changes.

tests/Agency/StorePerformanceTest.cpp
ignacionr: Let's try that.

tests/Agency/StorePerformanceTest.cpp
mpoeter: In that case you should also consider thread numbers that are not a power of two - sometimes this generates some surprising results.

tests/Agency/StorePerformanceTest.cpp
ignacionr: Oh, sorry I just re-read your comment, it is not correct. Please look at the new parameter ("sequential") on the report() method.

tests/Agency/StorePerformanceTest.cpp
ignacionr: Sorry to re-iterate, what are your thoughts about this?

tests/Agency/StorePerformanceTest.cpp
ignacionr: I can do the latency of each write, if you think it bears value (I understand they are important for understanding minimums and maximums).

tests/Agency/StorePerformanceTest.cpp
ignacionr: Thanks.

tests/Agency/StorePerformanceTest.cpp
ignacionr: Sure, can do.

tests/Agency/StorePerformanceTest.cpp
ignacionr: This was incorporated.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+        }
+      }},
+    }};
+
+    std::vector<OperationMeasurement::Clock::duration> durations;
+    auto begin{observations_.begin()};
+    auto from {*begin};
+    auto const absolute_start {*begin};
+    size_t block_position {0};
+    std::transform(++begin, observations_.end(), std::back_inserter(durations),
```
neunhoef, line 123: This code is rather obfuscated. I had a hard time to understand it. We either need better comments or a different approach.

neunhoef (changes requested):
Let's discuss in a meeting. Basically the setup works now. I am still puzzled about the very few very large measurements at the beginning. Probably an indication that the Store needs some time of setup time at the beginning.
The `report` method either needs a rewrite or a lot more comments.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+        if (begin != end) { 
+          std::vector<Duration> copy {begin, end};
+          std::sort(copy.begin(), copy.end());
+          inform (copy[copy.size() / 2]);
+        }
+      }},
+      {"max10", [](DurationIterator begin, DurationIterator end, InformValue inform) { 
+        if (begin != end) { 
+          std::vector<Duration> copy {begin, end};
+          std::sort(copy.begin(), copy.end(), [](Duration const& a, Duration const& b){ return b < a; });
```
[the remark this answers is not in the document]
ignacionr: I realised that sorting before hand, didn't affect any of these basic statistics, so I did that. Thanks for the comment, it was incorporated.

tests/Agency/StorePerformanceTest.cpp
neunhoef: I guess this could be true, and the assert is probably not too expensive, let's leave it as it is.

tests/Agency/StorePerformanceTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+///
+/// @author Ignacio Rodriguez
+/// @author Copyright 2021, ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#include "gtest/gtest.h"
+
+#include "Agency/Store.h"
+#include "Mocks/Servers.h"
+#include "fakeit.hpp"
```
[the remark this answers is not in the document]
neunhoef: We do not need this, I have removed it.

neunhoef (approved):
LGTM, let's just wait for the successful jenkins run.

neunhoef:
Given up on this.

PR #14189, opened 2021-05-14
MerkleTree left growth

neunhoef:
### Scope & Purpose

This implements growth to the left for MerkleTrees.

We noticed that this is necessary because a new cluster might have a rangeMin for its MerkleTrees of new collections which is larger than some `_rev` values of data in another cluster. Since DC2DC replication must be able to sync the old data, we must be able to lower rangeMin, that is, grow to the left.

As we currently only use the root node of the tree (summary information about count and hash) and the bottommost level (level 6), we also decided to remove all intermediate levels to save some memory and to reduce the cost of updates to the tree. So effectively every tree modification operation will modify only level 6 and the root node. The levels in between are not present anymore and not used.

This PR also adjusts some test code to call the new metrics API (`/_admin/metrics/v2`) instead of the old API (`/_admin/metrics`) and thus changes a few metrics names in other, unrelated tests.

It also exposes whether or not a collection uses Merkle trees (revision trees) in the detail view of a collection in the web UI.
That means when this PR is merged, we need to rebuild the web UI files.

### Backports:

- [x] Forward port required for: https://github.com/arangodb/arangodb/pull/14220

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in recovery, shell_client, replication_sync)


lib/Containers/MerkleTree.cpp
```diff
@@ -262,17 +197,17 @@ MerkleTree<Hasher, BranchingBits>::fromBottomMostCompressed(std::string_view buf
   char const* p = buffer.data();
   char const* e = p + buffer.size();
 
-  if (p + 3 * sizeof(uint64_t) > e) {
+  if (p + sizeof(Meta) >= e) {
```
jsteemann, line 200: I think for a completely empty tree `e` can be equal to `p + sizeof(Meta)`
```suggestion
  if (p + sizeof(Meta) > e) {
```

lib/Containers/MerkleTree.cpp
```diff
@@ -528,15 +459,16 @@ void MerkleTree<Hasher, BranchingBits>::checkInsertMinMax(std::unique_lock<std::
                                                           std::uint64_t minKey,
                                                           std::uint64_t maxKey) {
   if (minKey < meta().rangeMin) {
-    throw std::out_of_range("Cannot insert, key " + std::to_string(minKey) +
-                            " less than range minimum " +
-                            std::to_string(meta().rangeMin) + ".");
+    // unlock so we can get exclusive access to grow the range
+    guard.unlock();
+    growLeft(minKey);
```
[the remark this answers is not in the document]
jsteemann: I have renamed `checkInsertMinMax` to `prepareInsertMinMax` now.

The locking issue w.r.t. `growLeft()` and `growRight()` would be easy to fix in the container alone, but we have unit tests that call `growLeft()` and `growRight()` from the outside, and expects the functions to lock properly. So I tend to stick with the existing locking even though it is not great. But it should work fine as it is, and the performance impact of removing a sequence of unlock-lock steps should be negligible. Growing the trees will not happen extremely often.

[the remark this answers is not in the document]
mpoeter: I like the new name! 👍 I'm fine with keeping the unlock-lock sequence if changing it would be more of a hassle.

lib/Containers/MerkleTree.h
```diff
@@ -54,6 +54,51 @@ template <typename Hasher,
           std::uint64_t const BranchingBits = 3  // 8 children per internal node,
           >
 class MerkleTree {
+  // A MerkleTree has three parameters which define its semantics:
+  //  - rangeMin: lower bound (inclusive) for _rev values it can take
+  //  - rangeMax: upper bound (exclusive) for _rev values it can take
+  //  - maxDepth: depth of the tree (root plus so many levels below)
```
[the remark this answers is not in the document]
jsteemann: Good point. Working on it...

lib/Containers/MerkleTree.cpp
```diff
@@ -665,143 +597,104 @@ std::unique_ptr<MerkleTree<Hasher, BranchingBits>> MerkleTree<Hasher, BranchingB
...
+    ++index1;
+    ++index2;
+  }
+  // And finally the rest of tree2:
+  for ( ; pos < tree2->meta().rangeMax; pos += keysPerBucket) {
+    Node const& node2 = tree2->node(offsetBottom + index2);
+    if (node2.count != 0) {
+      addRange(pos, pos + keysPerBucket - 1);
+    }
+    ++index2;
```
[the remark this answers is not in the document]
jsteemann: We indeed only make use of the lowest level now.
Thus we are removing the nodes above via https://github.com/arangodb/arangodb/pull/14210 to save maintaining and storing them.

arangod/Aql/grammar.h
```diff
@@ -1,4 +1,4 @@
-/* A Bison parser, made by GNU Bison 3.7.  */
+/* A Bison parser, made by GNU Bison 3.5.1.  */
```
joerg84, line 1: Any reason to go back to this version?

jsteemann: It's a generated file, and whatever is in there will likely get overwritten soon by a new build. So I wouldn't worry about it.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
```diff
@@ -221,6 +223,89 @@ static void JS_CollectionRevisionTreeCorrupt(v8::FunctionCallbackInfo<v8::Value>
...
+      = manager->createContext(engine, ttl, SyncerId{17}, ServerId{17}, "");
+    if (ctx == nullptr) {
+      TRI_V8_THROW_EXCEPTION_INTERNAL("Could not create RocksDBReplicationContext");
+    }
+    RocksDBReplicationContextGuard guard(manager, ctx);
+    auto* physical = toRocksDBCollection(*collection);
+    auto batchId = ctx->id();
+    storedTree = physical->revisionTree(batchId);
+    computedTree = physical->computeRevisionTree(batchId);
+    ctx->setDeleted();
```
mpoeter, line 256: Is this exception safe? If `computeRevisionTree` etc. can throw then `ctx` would be leaked.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
```diff
@@ -221,6 +223,89 @@ static void JS_CollectionRevisionTreeCorrupt(v8::FunctionCallbackInfo<v8::Value>
...
+    } else {
+      builder.add("computedTree", VPackValue(false));
+    }
+    if (storedTree != nullptr && computedTree != nullptr) {
+      try {
+        std::vector<std::pair<uint64_t, uint64_t>> diff
+          = computedTree->diff(*storedTree);
+        builder.add("equal", VPackValue(diff.empty()));
+      } catch(std::exception const& exc) {
+        builder.add("equal", VPackValue(std::string(exc.what())));
```
mpoeter, line 280: We store the exception message in the attribute `equal`?

lib/Containers/MerkleTree.cpp
```diff
@@ -57,7 +57,7 @@
 #endif
 
 // can turn this on for more aggressive and expensive consistency checks
-// #define PARANOID_TREE_CHECKS
+#define PARANOID_TREE_CHECKS
```
mpoeter, line 60: I suppose we don't want this to be enabled by default
```suggestion
// #define PARANOID_TREE_CHECKS
```

lib/Containers/MerkleTree.cpp
```diff
@@ -262,199 +175,224 @@ MerkleTree<Hasher, BranchingBits>::fromBottomMostCompressed(std::string_view buf
...
     std::uint64_t count = read.getNumber<std::uint64_t>();
 
     read = nodeSlice.get(StaticStrings::RevisionTreeHash);
     if (!read.isString()) {
-      tree.reset();
-      return tree;
+      return nullptr;
     }
     p = nodeSlice.get(StaticStrings::RevisionTreeHash).getString(l);
     std::uint64_t hash = basics::HybridLogicalClock::decodeTimeStamp(p, l);
```
mpoeter, line 319: Why is the hash encoded as a HLC timestamp? The hash value is obtained by passing the revision ids through some arbitrary hash function, so the hash value can be anything, or am I missing something?

lib/Containers/MerkleTree.cpp
```diff
@@ -262,199 +175,224 @@ MerkleTree<Hasher, BranchingBits>::fromBottomMostCompressed(std::string_view buf
...
-    return tree;
+    return nullptr;
+  }
+
+  read = slice.get(StaticStrings::RevisionTreeInitialRangeMin);
+  if (!read.isString()) {
+    return nullptr;
+  }
+  p = read.getString(l);
+  std::uint64_t initialRangeMin = basics::HybridLogicalClock::decodeTimeStamp(p, l);
```
mpoeter, line 276: This code could be simplified since there is also a `decodeTimeStamp` function that takes a slice.

lib/Containers/MerkleTree.cpp
```diff
@@ -1038,162 +951,317 @@ template <typename Hasher, std::uint64_t const BranchingBits>
...
-    }
+ 
+  // adjust bucket node
+  bool success = modifyLocal(key, value, isInsert);
+  if (ADB_UNLIKELY(!success)) {
+    throw std::invalid_argument("Tried to remove key that is not present.");
   }
+    
+  // adjust summary node
+  modifyLocal(meta().summary, 1, value, isInsert);
```
mpoeter, line 987: It is probably unlikely that this fails, but should we at least add an assertion?

lib/Containers/MerkleTree.h
```diff
@@ -54,6 +54,51 @@ template <typename Hasher,
...
+  // of the tree. This turned out to be not good enough, since we cannot
+  // estimate the smallest _rev value a collection will ever receive
+  // (for example, in DC2DC we might need to replicate old data into a
+  // newly created collection). Therefore, we can now also grow the width
+  // to the left by decreasing rangeMin.
+  //
+  // Unfortunately, we can only compare two different MerkleTrees, if
+  // the difference of their rangeMin values is a multiple of the number
+  // of _rev values in a leaf node, which is 
+  //   (rangeMax-rangeMin)/(BranchingFactor^depth),
```
mpoeter, line 78: BranchingFactor is not used anymore, so anything related to it can actually be removed.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -597,21 +626,54 @@ void RocksDBMetaCollection::corruptRevisionTree(uint64_t count, uint64_t hash) {
...
+    auto&& [newTree, beginSeq] = result.get();
 
-      _revisionTree = std::make_unique<RevisionTreeAccessor>(std::move(newTree), _logicalCollection);
-      _revisionTreeApplied = beginSeq;
-      _revisionTreeCreationSeq = beginSeq;
-      _revisionTreeSerializedSeq = beginSeq;
+    if (_meta.hasBlockerUpTo(beginSeq)) {
+      return {TRI_ERROR_LOCKED, "cannot rebuild revision tree now as there are still transaction blockers in place"};
     }
+
```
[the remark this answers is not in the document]
jsteemann: We do 2 checks, one before we start scanning entire collection (which is indeed in `revisionTreeFromCollection`) and one after the scanning is finished.
The idea is that scanning can take so long that there is some chance that new blocker arrive during the scan.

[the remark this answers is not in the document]
mpoeter: Ah, makes sense since blockers can be updated concurrently.

tests/Containers/MerkleTreeTest.cpp
```diff
@@ -505,14 +470,166 @@ TEST(MerkleTreeTest, test_serializeBinary) {
 
   std::string t1s;
   t1.serializeBinary(t1s, true);
+  ASSERT_EQ(520, t1s.size());
+
+  std::unique_ptr<::arangodb::containers::MerkleTree<::arangodb::containers::FnvHashProvider, 3>> t2 =
+      ::arangodb::containers::MerkleTree<::arangodb::containers::FnvHashProvider, 3>::fromBuffer(t1s);
+  ASSERT_NE(t2, nullptr);
+  ASSERT_TRUE(t1.diff(*t2).empty());
+  ASSERT_TRUE(t2->diff(t1).empty());
```
mpoeter, line 479: In such cases I prefer `EXPECT_xxx` instead of `ASSERT_xxx` as that way we have more information if the test fails (but no need to change this now).

arangod/Aql/grammar.h
jsteemann: Anyway, reverted all these changes in this PR.

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -2968,15 +2968,37 @@ void RestReplicationHandler::handleCommandRevisionTree() {
...
+  if (withVerification) {
+    auto tree2 = ctx.collection->getPhysical()->computeRevisionTree(ctx.batchId);
+    if (!tree2) {
+      generateError(rest::ResponseCode::SERVER_ERROR, TRI_ERROR_INTERNAL,
+                    "could not generate revision tree from collection");
+      return;
+    }
+
+    VPackObjectBuilder guard(&result);
+    result.add(VPackValue("computedTree"));
```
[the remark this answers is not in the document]
jsteemann: Renamed the attribute to just "computed". No "tree" in the attribute name anymore.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -820,9 +906,17 @@ void RocksDBMetaCollection::applyUpdates(rocksdb::SequenceNumber commitSeq) {
...
-  TRI_ASSERT(_revisionTree->maxDepth() == revisionTreeDepth);
-
+  TRI_ASSERT(_revisionTree->depth() == revisionTreeDepth);
+  
   Result res = basics::catchVoidToResult([&]() -> void {
+    // bump sequence number upwards
+    auto bumpSequence = [this](rocksdb::SequenceNumber seq) noexcept {
+      rocksdb::SequenceNumber applied = _revisionTreeApplied.load();
+      while (seq > applied) {
+        _revisionTreeApplied.compare_exchange_strong(applied, seq);
```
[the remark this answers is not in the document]
jsteemann: Changed this, and also added an early abort to the loop in case the CAS succeeds.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
jsteemann: I am now storing it in `error` instead.

lib/Containers/MerkleTree.cpp
jsteemann: Good catch. Done!

lib/Containers/MerkleTree.cpp
jsteemann: Applied as suggested. Thanks!

lib/Containers/MerkleTree.cpp
jsteemann: I have added such assertion.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
jsteemann: Good catch. Fixed!

lib/Containers/MerkleTree.cpp
jsteemann: Fixed as well.

tests/Containers/MerkleTreeTest.cpp
jsteemann: Ok, then for simplicity we will leave it as is here :sweat_smile:

lib/Containers/MerkleTree.h
jsteemann: BranchingFactor is currently hard-coded to 8. So we can remove it any time later from the code. It shouldn't change any user-visible parts of APIs.

mpoeter (approved):
A few suggestions, but otherwise LGTM!

There is quite a bit of code duplication in the test code that could be reduced, but given tight time frame I am fine with keeping it as is.

lib/Containers/MerkleTree.h
mpoeter: `BranchingFactor` is only stored and read from vpack slices, but otherwise not used _at all_, so I would simply remove it now (together with the static string).
In fact, I think this comment is no longer entirely accurate. @neunhoef can you comment on that?

neunhoef: We still store the "depth" in the Meta information. It is currently a constant 6 everywhere. The number of buckets we use is defined to be `BranchingFactor^depth`, which is currently a constant `2^18`.
If we want to clean this up everywhere, we would have to:
 - remove `BranchingBits` and `BranchingFactor` in `MerkleTree.h`
 - change Meta and put `NumberOfBuckets` or so instead of `depth`
 - change all places which use depth
 - change all comments where depth and BranchingFactor occur
 - there are 33 occurrences of `nodeCountAtDepth` alone in the code, so this is not an insignificant change.

neunhoef: Actually, I was wrong, in the unittests we do use MerkleTree objects with depth set to 2 and 3, in addition to the normal 6.

tests/Containers/MerkleTreeTest.cpp
```diff
@@ -740,3 +831,572 @@ TEST(MerkleTreeTest, test_check_consistency) {
...
+  t2.insert(M + 22);
+  expected.emplace_back(std::pair(M + 20, M + 23));
+  t1.insert(M + W - 8);
+  t2.insert(M + W - 5);
+  expected.emplace_back(std::pair(M + W - 8, M + W - 5));
+  t2.insert(M + W);
+  t2.insert(M + W + 5);
+  expected.emplace_back(std::pair(M + W, M + W + 7));
+  t2.clear();
+}
```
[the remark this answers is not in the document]
neunhoef: You are right, I will check if the tests pass with the correct diffs... :-)

lib/Containers/MerkleTree.h
neunhoef: The word BranchingFactor is removed from the code and from comments. We only use BranchingBits now to work with depth.

tests/Containers/MerkleTreeTest.cpp
```diff
@@ -740,3 +831,572 @@ TEST(MerkleTreeTest, test_check_consistency) {
...
+
+  std::vector<std::pair<std::uint64_t, std::uint64_t>> expected;
+  ASSERT_TRUE(::diffAsExpected(t1, t2, expected));
+
+  // Now insert something into t1 left of tree 2 as well as in the overlap:
+  t1.insert(M);      // first bucket in t1
+  expected.emplace_back(std::pair(M, M+3));
+  t1.insert(M + 8);  // third bucket in t1
+  expected.emplace_back(std::pair(M+8, M+11));
+  t1.insert(M + 16); // firth bucket in t1, first bucket in t2
```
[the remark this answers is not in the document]
neunhoef: Done.

tests/Containers/MerkleTreeTest.cpp
```diff
@@ -740,3 +831,572 @@ TEST(MerkleTreeTest, test_check_consistency) {
...
+  ASSERT_TRUE(::diffAsExpected(t1, t2, expected));
+  t1.clear();
+  expected.clear();
+
+  // Now insert something into t1 left of tree 2 as well as in the overlap, but expect one contiguous interval:
+  t1.insert(M);      // first bucket in t1
+  t1.insert(M + 4);  // second bucket in t1
+  t1.insert(M + 8);  // third bucket in t1
+  t1.insert(M + 12); // fourth bucket in t1
+  t1.insert(M + 16); // firth bucket in t1, first bucket in t2
```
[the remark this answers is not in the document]
neunhoef: Done.

tests/Containers/MerkleTreeTest.cpp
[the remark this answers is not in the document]
neunhoef: Done.

tests/Containers/MerkleTreeTest.cpp
```diff
@@ -740,3 +831,572 @@ TEST(MerkleTreeTest, test_check_consistency) {
   }
 #endif
 }
+
+class MerkleTreeGrowTests
+    : public ::arangodb::containers::MerkleTree<::arangodb::containers::FnvHashProvider, 3>,
```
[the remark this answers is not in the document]
neunhoef: An easy change is to make the local variables into public members of the derived class. If we make the container a member, then we do no longer have access to protected methods and members and need to call lots of methods differently.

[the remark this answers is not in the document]
neunhoef: For now, I will do the smaller solution.

tests/js/client/shell/shell-collection-counts-cluster.js
```diff
@@ -64,10 +64,19 @@ let getCoordinators = function() {
   return getServers('coordinator');
 };
 
+// this array will be modified by tests
+let failurePointsToKeep = [];
```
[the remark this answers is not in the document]
neunhoef: Done.

PR #14194, opened 2021-05-18
Fix index range calculation for queries where both bounds are expressed using the same operator.

mpoeter:
### Scope & Purpose

Fixes BTS-417

In some cases an index did not consider both bounds (lower and upper) for a close range scan if both bounds are expressed using the same operator, e.g., `FILTER doc.beginDate >= lb AND ub >= doc.beginDate`.

The newly added tests revealed yet another issue with the condition optimization in case of redundant conditions which result in suboptimal range calculation with an additional filter. But that needs to be fixed in a separate PR.

#### Backports:

- [x] Backports required for: 3.7, devel  (#14196, #14195)

#### Related Information

- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/BTS-417

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 v3.8.0 (XXXX-XX-XX)
 -------------------
 
+* Fixes BTS-417. In some cases an index did not consider both bounds (lower and
+  upper) for a close range scan if both bounds are expressed using the same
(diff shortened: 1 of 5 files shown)


jsteemann (approved):
LGTM!

PR #14195, opened 2021-05-18
Fix index range calculation for queries where both bounds are expressed using the same operator.

mpoeter:
### Scope & Purpose

Fixes BTS-417

In some cases an index did not consider both bounds (lower and upper) for a close range scan if both bounds are expressed using the same operator, e.g., `FILTER doc.beginDate >= lb AND ub >= doc.beginDate`.

The newly added tests revealed yet another issue with the condition optimization in case of redundant conditions which result in suboptimal range calculation with an additional filter. But that needs to be fixed in a separate PR.

#### Backports:

- [x] Backports required for: 3.7, 3.8  (#14196, #14194)

#### Related Information

- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/BTS-417

### Testing & Verification

### DIFF
--- a/tests/js/server/aql/aql-index-choice.js
+++ b/tests/js/server/aql/aql-index-choice.js
@@ -564,6 +564,105 @@ function BaseTestConfig () {
         assertEqual("sorted", collectNode.collectOptions.method);
       });
     },
+    
+    testSingleFilterUpperAndLowerBound: function() {
+      db[cn].ensureIndex({ type: "persistent", fields: ["uid"] });
+      [
+        `
+          FOR doc IN ${cn}
+            LET lowerBound = 1
+            LET upperBound = 100
+            FILTER doc.uid >= lowerBound AND doc.uid <= upperBound
+            RETURN doc
+        `,
+        `
+          FOR doc IN ${cn}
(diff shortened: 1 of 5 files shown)


tests/js/server/aql/aql-index-choice.js
```diff
@@ -564,6 +564,105 @@ function BaseTestConfig () {
...
+    },
+    
+    testSingleFilterUpperAndLowerBoundWithRedunantCondition: function() {
+      db[cn].ensureIndex({ type: "persistent", fields: ["uid"] });
+      // generate the cross product of the given arrays
+      const product = (...a) => a.reduce((a, b) => a.flatMap(d => b.map(e => [d, e].flat())));
+      // generate a list of all permutations of the given array
+      const perm = a => a.length
+        ? a.reduce((r, v, i) => [...r, ...perm([...a.slice(0, i), ...a.slice(i + 1)]).map(x => [v, ...x])], [])
+        : [[]];
```
jsteemann, line 613: :scream_cat:  :sweat_smile:

jsteemann (approved):
LGTM!

PR #14206, opened 2021-05-19
Add tree verification at run time triggered by API, add tests

neunhoef:
Sub-PR to add runtime verification of MerkleTrees.

### DIFF
--- a/arangod/RestHandler/RestReplicationHandler.cpp
+++ b/arangod/RestHandler/RestReplicationHandler.cpp
@@ -2968,15 +2968,44 @@ void RestReplicationHandler::handleCommandRevisionTree() {
     return;
   }
 
+  // shall we do a verification?
+  bool found{false};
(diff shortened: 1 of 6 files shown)


arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -2968,15 +2968,44 @@ void RestReplicationHandler::handleCommandRevisionTree() {
...
 
+  // shall we do a verification?
+  bool found{false};
+  bool withVerification{false};
+  std::string const& verificationString = _request->value("verification", found);
+  if (found) {
+    if (verificationString.compare("true") == 0) {
+      withVerification = true;
+    }
+  }
```
jsteemann, line 2979: Can be simplified to:
```suggestion
  bool withVerification = _request->parsedValue("verification", false);
```

PR #14207, opened 2021-05-19
Fixed code to be compatible with gcc-11.1

maierlars:
### Scope & Purpose

# I do not intend to change the compiler used for releases to gcc-11.1. Please read below.

I use Arch Linux. Thus my default system compiler is now gcc-11.1. I made some adjustments to make ArangoDB compile on Arch Linux. I explained all changes below. Almost all of them are adding **missing** includes.

- `lib/Basics/StringHeap.h` uses `std::size_t` but does not include `cstddef`.
- `3rdParty/iresearch/external/unicorn/unicorn/utility.hpp` uses `std::shared_ptr` but does not include `memory`.
- `arangosh/Utils/ProgressTracker.h` uses `std::unique_lock` but does not include `mutex`.

   - See https://gcc.gnu.org/gcc-11/porting_to.html#header-dep-changes:
 >  The following headers are used less widely in libstdc++ and may need to be included explicitly when compiled with GCC 11:
 >
 >    <limits> (for std::numeric_limits)
 >    <memory> (for std::unique_ptr, std::shared_ptr etc.)
 >    <utility> (for std::pair, std::tuple_size, std::index_sequence etc.)
 >    <thread> (for members of namespace std::this_thread.)

- `tests/IResearch/IResearchTestCompressor.h` changes meaning of `compressor`/`decompressor`. After declaration of static functions the unqualified name lookup does not return the same thing as before. (see https://godbolt.org/z/xjMPsWard)

- `arangod/Cluster/TraverserEngine.cpp` uses `EdgeDocumentToken` but does not include `Graph/EdgeDocumentToken.h`. It is only forward
declared and `std::function` requires all parameter types to be complete when invoked.

### DIFF
--- a/3rdParty/iresearch/external/unicorn/unicorn/utility.hpp
+++ b/3rdParty/iresearch/external/unicorn/unicorn/utility.hpp
@@ -54,6 +54,7 @@
 #include <istream>
 #include <iterator>
 #include <limits>
+#include <memory>
 #include <mutex>
 #include <new>
 #include <optional>
--- a/3rdParty/velocypack/include/velocypack/Slice.h
+++ b/3rdParty/velocypack/include/velocypack/Slice.h
@@ -25,18 +25,19 @@
 #ifndef VELOCYPACK_SLICE_H
 #define VELOCYPACK_SLICE_H 1
 
+#include <algorithm>
 #include <cstdint>
 #include <cstring>
-#include <string>
-#include <vector>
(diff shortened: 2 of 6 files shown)


maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/15493/

mchacki (approved):
LGTM!

PR #14234, opened 2021-05-21
Add `fillBlockCache` option to bypass RocksDB block cache

jsteemann:
### Scope & Purpose

https://arangodb.atlassian.net/browse/APM-86
Docs PR: https://github.com/arangodb/docs/pull/703
Added `fillBlockCache` option to bypass RocksDB block cache

* Implemented APM-86: add query option `fillBlockCache` to control population
  of RocksDB block cache with data read by the query. The default value for
  this per-query option is `true`, which mimics the previous behavior.
  Setting the option to off allows not storing data in RocksDB's block cache
  for queries that are known to read only semi-relevant or unimportant data.

This is currently experimental, and does not have documentation, tests, nor driver support.

#### Backports:

- [x] Backports required for: *to be discussed*

### Testing & Verification


mpoeter (approved):
It is not clear to me why the tests are split into three separate files. The setup seems to be identical in all cases, and the tests are also very similar, so IMHO it would make sense to put them into a single file to keep it DRY (unless there are other reasons which I am not aware of?).
Otherwise LGTM!

Simran-B:
Just double checking: `fillBlockCache` is a cursor option, not a query option?

`@RESTSTRUCT{fillBlockCache,post_api_cursor_opts,boolean,optional,}`

I'm a bit confused that `batchSize` isn't a cursor option by the look of things:

`@RESTBODYPARAM{batchSize,integer,optional,int64}`

jsteemann:
@Simran-B : I don't know what the difference is, but it can be used as follows:
https://github.com/arangodb/arangodb/pull/14234/files#diff-71885788211640a079769272c37b97fab5c379c088246697cfcab33e063f6f12R62

jsteemann:
@mpoeter : it is split into 3 separate files because a test that populates the block cache basically "wastes" the environment. With 3 different files we can at least have 3 separate and clean server instances that don't interfere with each other.

mpoeter (approved):

PR #14235, opened 2021-05-21
Small change on creation scripts for metrics

neunhoef:
This change is necessary for metrics on k8s. For counters, the
averaging period is changed from 1 minute to 3 minutes, since on k8s
scraping is done less frequently.


jsteemann (approved):
LGTM!

PR #14238, opened 2021-05-21
Feature/use validators on smart edge collections

mchacki:
### Scope & Purpose

This is mainly an Enterprise PR.

It adds the ability to define internal Validators alongside the User-Given Schema Validator.
The internal validators are meant to be constant and not to be modified by the user.
They are executed on the Shard so we have higher confidence that nothing slips past them.

Right now they are used to only to validate Smart-Graph sharding consistency, which has been done on Coordinator only in the past.

This is a prerequisite for Hybrid SmartGraphs.

There are some side-effects to this PR:
1) We introduced C++ Test Files in Enterprise CodeBase.
2) We introduced Enterprise-Only Upgrade Tasks
3) There will be a new Property Entry on every Collection, which is added during upgrade to this version on existing clusters.

- [X] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)

#### Backports:

- [X] No backports required

#### Related Information

- [ ] Docs PR: 
- [X] Enterprise PR: https://github.com/arangodb/enterprise/pull/680
- [ ] Design document: 

### Testing & Verification

- [X] This change is already covered by existing tests, such as *SmartGraph modification tests*
  - [X] Added new C++ **Unit tests** (See Enterprise PR)
- [X] There are tests in an external testing repository:
https://github.com/arangodb/release-test-automation/pull/98

Link to Jenkins PR run:

### Documentation

This is just an internal change, not controllable by the user, and the behavior should not change visibly.


arangod/Agency/AsyncAgencyComm.h
```diff
@@ -90,8 +90,11 @@ struct AsyncAgencyCommResult {
       return Result{fuerteToArangoErrorCode(error), to_string(error)};
     } else {
       auto code = statusCode();
-      return Result{fuerteStatusToArangoErrorCode(code),
-                    fuerteStatusToArangoErrorMessage(code)};
+      auto internalCode = fuerteStatusToArangoErrorCode(code);
```
mchacki, line 93: This is a drive-by "fix".
If the FuerteResult was OK, and someone asked for the Result.
If was created with `TRI_ERROR_NO_ERROR` and message "200 OK".
The Result asserts that NO_ERROR has NO message and crashed.

arangod/Cluster/Maintenance.cpp
```diff
@@ -88,12 +87,15 @@ static std::shared_ptr<VPackBuilder> createProps(VPackSlice const& s) {
 
 static std::shared_ptr<VPackBuilder> compareRelevantProps(VPackSlice const& first,
                                                           VPackSlice const& second) {
+  static std::vector<std::string> const compareProperties{WAIT_FOR_SYNC, SCHEMA, CACHE_ENABLED,
```
mchacki, line 90: Moved into this function to guarantee that the StaticString used in this static value is generated before.

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -992,6 +1003,27 @@ arangodb::Result LogicalCollection::properties(velocypack::Slice const& slice, b
...
+        Helper::getNumericValue<InternalCollectionType>(slice, StaticStrings::InternalCollectionType,
+                                                        _collectionType);
+    if (nextType != _collectionType) {
+      // We can only apply this operation If we are write locked.
+      // The WriteLock is guaranteed by the maintenance.
+      // Otherwise someone may insert a document and perform an unprotected
+      // access to _internalValidator.
+
+      // TODO Check if we need to make sure we are write locked.
+      // TRI_ASSERT(statusLock().isLockedWrite());
```
mchacki, line 1020: This TODO is still under investigation

arangod/VocBase/Validators.cpp
```diff
@@ -127,6 +129,7 @@ void ValidatorBase::toVelocyPack(VPackBuilder& b) const {
   VPackObjectBuilder guard(&b);
   b.add(StaticStrings::ValidationParameterMessage, VPackValue(_message));
   b.add(StaticStrings::ValidationParameterLevel, VPackValue(to_string(this->_level)));
+  b.add(StaticStrings::ValidationParameterType, VPackValue(this->type()));
```
mchacki, line 132: I actually do not need this anymore.
However I think knowing the Type of a schema could be benefitial in the future.
So i am happy with keeping and removing this line.

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -1066,28 +1116,58 @@ TEST_F(MaintenanceTestActionPhaseOne,
...
       ASSERT_EQ(action->name(), "UpdateCollection");
       ASSERT_EQ(action->get("shard"), shname);
       ASSERT_EQ(action->get("database"), dbname);
       auto const props = action->properties();
-
     }
   }
 }
 
+TEST_F(MaintenanceTestActionPhaseOne,
```
mchacki, line 1131: This is new code

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -742,36 +754,77 @@ class MaintenanceTestActionPhaseOne : public SharedMaintenanceTest {
     ASSERT_TRUE(action.has(SHARD));
     ASSERT_EQ(action.get(DATABASE), dbName);
   }
-};
 
+  // Modify the internal collection type in Plan.
+  // This is basically used to simulate a Database Upgrade
+  // as the value is not modifiyable by the user.
+  auto changeInternalCollectionTypePlan(std::string const& dbName, std::string const& planId,
```
mchacki, line 761: This function is new

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -742,36 +754,77 @@ class MaintenanceTestActionPhaseOne : public SharedMaintenanceTest {
...
+    ASSERT_TRUE(plan.has(vec)) << "The underlying test plan is modified, it "
+                                  "does not contain Database '"
+                               << dbName << "' and Collection '" << planId << "' anymore.";
+    auto& props = plan(vec);
+    VPackBuilder v;
+    v.add(VPackValue((int)type));
+    props(StaticStrings::InternalCollectionType) = v.slice();
+  }
+
+  auto assertIsChangeInternalCollectionTypeAction(ActionDescription const& action,
```
mchacki, line 778: This function is new

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -543,12 +539,18 @@ class MaintenanceTestActionPhaseOne : public SharedMaintenanceTest {
...
+  enum class PLAN_LEADERSHIP_TYPE {
+    SELF,
+    RESIGNED_SELF,
+    OTHER,
+    RESIGNED_OTHER
+  };
 
   enum class LOCAL_LEADERSHIP_TYPE { SELF, OTHER, RESIGNED, REBOOTED };
 
   auto getShardsForServer(std::string const& dbName, std::string const& planId,
```
mchacki, line 551: This function is modified

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -98,8 +98,6 @@ size_t localId = 1016002;
 std::string S("s");
```
mchacki, line 98: Sorry this file suffered from reformatting, I have added comments an actually modified code

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -404,7 +410,7 @@ ErrorCode LogicalCollection::getResponsibleShard(arangodb::velocypack::Slice sli
 }
 
 /// @briefs creates a new document key, the input slice is ignored here
-std::string LogicalCollection::createKey(VPackSlice) {
+std::string LogicalCollection::createKey(VPackSlice) const {
```
jsteemann, line 413: It may work like this, but it feels wrong that `createKey()` is const.
I guess the keyGenerator will need to modify its internal state somehow, so it feels non-consty.

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -1184,16 +1214,66 @@ void LogicalCollection::schemaToVelocyPack(VPackBuilder& b) const {
...
   }
   return {};
 }
+
+void LogicalCollection::setInternalCollectionType(InternalCollectionType type) {
+  _collectionType = type;
+}
+
+void LogicalCollection::setInternalValidator(std::unique_ptr<arangodb::ValidatorBase> validator) {
+  // We can only set the InteralSchema once.
```
jsteemann, line 1249:
```suggestion
  // We can only set the internalValidator once.
```

arangod/VocBase/LogicalCollection.h
```diff
@@ -88,6 +88,21 @@ class LogicalCollection : public LogicalDataSource {
...
+   * Mostly used for Enterprise/Smart collection types which have
+   * some specialities over community collections.
+   * For Backwards Compatibility a Value can never be reused
+   */
+  enum class InternalCollectionType {
+    Community = 1,
+    SmartEdge = 2,
+    LocalSmartEdge = 3,
+    RemoteSmartEdge = 4
+  };
```
jsteemann, line 104: I am not a huge fan of this, because with this a collection now has a `_type` and an `_internalCollectionType`.
It would be good to fuse them into just one. I understand this will be a lot of work and affects all drivers etc., so I understand using a second internal variable for it.
However, as far as I can see it is only used for edge collections, so it should be named appropriately, e.g. `_edgeCollectionType`. I also think `Community` is not a good name for the non-smart edge collections, because these collections can also be created with the EE. It would rename it to `Normal`, even if that sounds boring.

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -742,36 +754,77 @@ class MaintenanceTestActionPhaseOne : public SharedMaintenanceTest {
     ASSERT_TRUE(action.has(SHARD));
     ASSERT_EQ(action.get(DATABASE), dbName);
   }
-};
 
+  // Modify the internal collection type in Plan.
+  // This is basically used to simulate a Database Upgrade
+  // as the value is not modifiyable by the user.
```
jsteemann, line 760:
```suggestion
  // as the value is not modifyable by the user.
```

arangod/VocBase/LogicalCollection.h
mpoeter: How about `Standard` instead of `Normal`?

mchacki: The reason why I did this is to not expose this non-modifyable value to the user, it is of pure internal use only.
Having this exposed limits us in what we can do with it (see drivers) having this not-exposed (and separated from document/edge type) gives us freedom to only be backwards compatible by 1 version, or even erase it at all if we do not need it anymore.

I a open to store this value into the "old" type attribute internally and translate it before we send it out to the user, to not change the drivers. But I do not know if it is worth it.

I am happy to rename the first to `Standard` or `Normal`.

Also we may want to introduce e.g. "SmartJoin" type here as well, as this collections also requires the same speciality as the SmartEdges, so.i am against moving it to an edge only type, this are just the first entries here.

arangod/VocBase/LogicalCollection.cpp
mchacki: I agree, no idea why it compiled this way.
Removed the const

neunhoef (approved):
LGTM.

mchacki:
FYI: The last two Jenkins runs each showed a different known error, not caused by this PR.

arangod/VocBase/LogicalCollection.cpp
hkernbach: Remove TODO and add different assertion as discussed.

tests/Maintenance/MaintenanceTest.cpp
```diff
@@ -742,36 +754,77 @@ class MaintenanceTestActionPhaseOne : public SharedMaintenanceTest {
...
+                    ->collections()
+                    ->database(dbName)
+                    ->collection(planId);
+    auto vec = path->vec(2);
+    ASSERT_TRUE(plan.has(vec)) << "The underlying test plan is modified, it "
+                                  "does not contain Database '"
+                               << dbName << "' and Collection '" << planId << "' anymore.";
+    auto& props = plan(vec);
+    VPackBuilder v;
+    v.add(VPackValue((int)type));
```
hkernbach, line 774:
```suggestion
    v.add(VPackValue(type));
```

hkernbach (approved):

PR #14240, opened 2021-05-21
Feature/arangosearch inverted index

Dronplane:
### Scope & Purpose

New type of  index - ArangoSearch backed index  on collection. Index uses inverted index under the hood

#### Backports:

- [x] No backports required

#### Related Information
- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] Design document:  https://arangodb.atlassian.net/wiki/spaces/REED/pages/1492287725/DD%2BInverted%2Bindex

### Testing & Verification

Link to Jenkins PR run:

### Documentation


arangod/IResearch/IResearchCommon.cpp
```diff
@@ -60,6 +60,13 @@ arangodb::LogTopic& logTopic() { return TOPIC; }
 /*static*/ std::string const StaticStrings::PrimarySortCompressionField("primarySortCompression");
 /*static*/ std::string const StaticStrings::StoredValuesField("storedValues");
 /*static*/ std::string const StaticStrings::CollectionNameField("collectionName");
+/*static*/ std::string const StaticStrings::ConsolidationIntervalMsec("consolidationIntervalMsec");
```
MBkkt, line 63: https://en.wikipedia.org/wiki/Millisecond Ms better than Msec

arangod/IResearch/IResearchCommon.cpp
```diff
@@ -60,6 +60,13 @@ arangodb::LogTopic& logTopic() { return TOPIC; }
...
+/*static*/ std::string const StaticStrings::CommitIntervalMsec("commitIntervalMsec");
```
MBkkt, line 64: https://en.wikipedia.org/wiki/Millisecond Ms better than Msec

arangod/IResearch/IResearchDataStore.h
```diff
@@ -0,0 +1,378 @@
+////////////////////////////////////////////////////////////////////////////////
```
MBkkt, line 1: Why are these files duplicated IResearchLink?

I understand this is a renaming. There is a problem with this. 
When you merge you will not get a conflict even if IResearchLink changes

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -0,0 +1,1547 @@
+////////////////////////////////////////////////////////////////////////////////
```
MBkkt, line 1: Why are these files duplicated IResearchLink?

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,272 @@ bool StoredValue::write(irs::data_output& out) const {
...
+  if (_currentTypedAnalyzer) {
+    if (_currentTypedAnalyzer->next()) {
+      TRI_ASSERT(_primitiveTypeResetter);
+      TRI_ASSERT(_currentTypedAnalyzerValue);
+      TRI_ASSERT(_value._analyzer.get());
+      _primitiveTypeResetter(_value._analyzer.get(), _currentTypedAnalyzerValue->value);
+      return;
+    } else {
+      _currentTypedAnalyzer.reset();
+    }
```
MBkkt, line 726:
```suggestion
    }
    _currentTypedAnalyzer.reset();
```

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,272 @@ bool StoredValue::write(irs::data_output& out) const {
...
+  }
+  _valueSlice = VPackSlice::noneSlice();
+  while (_begin != _end) {
+    while (!_arrayStack.empty() && _valueSlice.isNone()) {
+      if (_arrayStack.back().valid()) {
+        _valueSlice = *_arrayStack.back();
+        _arrayStack.back()++;
```
MBkkt, line 733:
```suggestion
        ++_arrayStack.back();
```

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,272 @@ bool StoredValue::write(irs::data_output& out) const {
...
+          }
+          THROW_ARANGO_EXCEPTION_FORMAT(
+              TRI_ERROR_NOT_IMPLEMENTED,
+              "Inverted index does not support indexing objects and configured "
+              "analyzer does "
+              " not accept objects. Please use another analyzer to process an "
+              "object or exclude field '%s'"
+              " from index definition",
+              _nameBuffer.c_str());
+          break;
```
MBkkt, line 779:
```suggestion
          return;
```

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,272 @@ bool StoredValue::write(irs::data_output& out) const {
...
+            THROW_ARANGO_EXCEPTION_FORMAT(
+                TRI_ERROR_NOT_IMPLEMENTED,
+                "Configured analyzer does not accept arrays and field has no "
+                "expansion set. "
+                "Please use another analyzer to process an array or exclude "
+                "field '%s'"
+                " from index definition or enable expansion",
+                _nameBuffer.c_str());
+          }
+        } break;
```
MBkkt, line 795:
```suggestion
          if (_begin->first.back().shouldExpand) {
            _arrayStack.push_back(VPackArrayIterator(_valueSlice));
          } else if (!setValue(_valueSlice, _begin->second)) {
            THROW_ARANGO_EXCEPTION_FORMAT(
                TRI_ERROR_NOT_IMPLEMENTED,
                "Configured analyzer does not accept arrays and field has no "
                "expansion set. "
                "Please use another analyzer to process an array or exclude "
                "field '%s'"
                " from index definition or enable expansion",
                _nameBuffer.c_str());
          }
          return;
```

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,272 @@ bool StoredValue::write(irs::data_output& out) const {
...
+        _currentTypedAnalyzerValue = irs::get<VPackTermAttribute>(*analyzer);
+        TRI_ASSERT(_currentTypedAnalyzerValue);
+        setNumericValue(_currentTypedAnalyzerValue->value);
+        _primitiveTypeResetter = [](irs::token_stream* stream, VPackSlice slice) -> void {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+          auto number_stream = dynamic_cast<irs::numeric_token_stream*>(stream);
+          TRI_ASSERT(number_stream);
+#else
+          auto number_stream = static_cast<irs::numeric_token_stream*>(stream);
+#endif
```
MBkkt, line 947: Maybe you should write a macro for such a pattern?
Better yet, a `constexpr` function

```C++
template <typename To, typename From>
constexpr ptrCast(From&& from) {
  auto* to = dynamic_cast<To*>(std::forward<From>(from));
  TRI_ASSERT(to);
  return to;
#else
  return static_cast<To*>(std::forward<From>(from));
#endif
}
```

js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -114,6 +114,15 @@ function attributeUncolored(v) {
   return '`' + v + '`';
 }
 
+function indexFieldToName(v) {
+  'use strict';
+  if (typeof v === 'object') {
+    return v.name;
+  } else {
+    return v;
+  }
```
MBkkt, line 123:
```suggestion
  }
  return v;
```

arangod/IResearch/IResearchViewStoredValues.h
```diff
@@ -55,6 +55,12 @@ class IResearchViewStoredValues {
     bool operator!=(StoredColumn const& rhs) const noexcept {
       return !(*this == rhs);
     }
+
+    bool sameName(std::string_view str) const noexcept {
+      return (str.size() + 1) == name.size() && 
+        strncmp(str.data(), name.data() + 1, str.size()) == 0;
```
MBkkt, line 61:
```suggestion
      return std::string_view{name.data(), name.size()} == str;
```
https://godbolt.org/z/9rfE45Wz9

MBkkt (commented):
too big

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,219 @@
...
+      lazy_bitset& bitset,
+      irs::cost::cost_t estimation) noexcept
+    : cost_(estimation),
+      bitset_(bitset) {
+    reset();
+  }
+
+  bool next() override final;
+  irs::doc_id_t seek(irs::doc_id_t target) override final;
+  irs::doc_id_t value() const noexcept override final { return doc_.value; }
```
MBkkt, line 82:
```suggestion
  bool next() final;
  irs::doc_id_t seek(irs::doc_id_t target) final;
  irs::doc_id_t value() const noexcept final { return doc_.value; }
```

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,219 @@
...
+ public:
+
+  struct proxy_cache {
+    std::map<const irs::sub_reader*, std::unique_ptr<lazy_bitset>> readers_;
+    irs::filter::prepared::ptr prepared_real_filter_;
+  };
+
+  proxy_query(proxy_cache& cache, irs::filter::ptr&& filter,
+              const irs::index_reader& index, const irs::order::prepared& order)
+    :cache_(cache), real_filter_(std::move(filter)), index_(index), order_(order) {}
```
MBkkt, line 105: Use clang-format

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  proxy_filter& set_cache(proxy_query::proxy_cache* cache) {
+    cache_ = cache;
+    return *this;
+  }
+
+ private:
+   mutable irs::filter::ptr real_filter_{nullptr};
+   mutable proxy_query::proxy_cache* cache_{nullptr};
```
MBkkt, line 166: Wrong code style

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+      return false;
+    }
+    return true;
+  }
+
+  mutable std::vector<arangodb::basics::AttributeName> _parsed;
+  QueryContext const& _ctx;
+  aql::Variable const& _ref;
+  using atr_ref = std::reference_wrapper<std::vector<basics::AttributeName> const>;
+  std::unordered_set<atr_ref,
```
MBkkt, line 79: Maybe you can use `absl::flat_hash_set`?

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+      // extra is expected to be the _from/_to attribute. So don't bother with the full
+      // path inspection
+      for (size_t i = 0; i < columnSize; ++i) {
+        if (index->_meta._sort.fields()[i].size() == 1 &&
+            index->_meta._sort.fields()[i].front().name == extraFieldName) {
+          _extraIndex = static_cast<int64_t>(i);
+          break;
+        }
+      }
+      if (_extraIndex < 0) { // try to find in other stored
```
MBkkt, line 121: early return has better readability

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+                    [&rdr](decltype(_coverage)::value_type& v) {
+                      v.second.reset(rdr);
+                    });
+    }
+
+    void seek(irs::doc_id_t doc) {
+      _doc = doc;
+    }
+
+    VPackSlice at(size_t i) override {
```
MBkkt, line 447: Write `final` instead of `override` if the class is final

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+  const irs::payload* _pkValue{};
+  irs::index_reader const* _reader{0};
+  IResearchInvertedIndex* _index;
+  aql::Variable const* _variable;
+  size_t _readerOffset{0};
+  int64_t _mutableConditionIdx;
+  irs::order::prepared _order;
+
+  CoveringVector _projections;
+  int64_t _extraIndex{-1};
```
MBkkt, line 500: Maybe all initialization in ctor?

arangod/IResearch/IResearchDataStore.h
Dronplane: The idea is to move all "common" part for the  inverted index and  the link here. So it will be not duplicate but move. Currently yes. Some conflicts are expected to happen,

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+/// @brief lookup referenced analyzer
+////////////////////////////////////////////////////////////////////////////////
+AnalyzerPool::ptr IResearchInvertedIndex::findAnalyzer(AnalyzerPool const& analyzer) const {
+  auto const it = _meta._analyzerDefinitions.find(irs::string_ref(analyzer.name()));
+
+  if (it == _meta._analyzerDefinitions.end()) {
+    return nullptr;
+  }
+
+  auto const pool = *it;
```
MBkkt, line 569: const remove auto move https://godbolt.org/z/f71aKxb8q

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,846 @@
...
+    }
+    translate.clear();
+  }
+  return matched == count;
+}
+
+std::unique_ptr<IndexIterator> arangodb::iresearch::IResearchInvertedIndex::iteratorForCondition(
+    LogicalCollection* collection, transaction::Methods* trx, aql::AstNode const* node, aql::Variable const* reference,
+    IndexIteratorOptions const& opts, int mutableConditionIdx) {
+  if (node) {
```
MBkkt, line 669:
```C++
if (!node) {
  ...
  return
}
...
```

arangod/IResearch/IResearchRocksDBEncryption.h
```diff
@@ -0,0 +1,102 @@
...
+  static std::shared_ptr<RocksDBEncryptionProvider> make(rocksdb::EncryptionProvider& encryption,
+                                                         rocksdb::Options const& options) {
+    return std::make_shared<RocksDBEncryptionProvider>(encryption, options);
+  }
+
+  explicit RocksDBEncryptionProvider(rocksdb::EncryptionProvider& encryption,
+                                     rocksdb::Options const& options)
+      : _encryption(&encryption), _options(options) {}
+
+  virtual size_t header_length() override {
```
MBkkt, line 66:
```suggestion
  size_t header_length() final {
```

arangod/Aql/IndexExecutor.h
```diff
@@ -117,6 +118,10 @@ class IndexExecutorInfos {
     return _outNonMaterializedIndRegs;
   }
 
+  bool isOneIndexCondition() const noexcept {
+    return _oneIndexCondition;
+  }
```
jsteemann, line 123: As far as I can see, the `_oneIndexCondition` boolean is used only as a performance optimization if there is only a single lookup condition (in contrast to having multiple sub-conditions combined with logical OR).
What we will gain from it is to save a call to `getMember(...)` on the conditions root node.
If this is the only thing it does but is otherwise not required, I think it may be worthwhile to drop the boolean because it creates more complicated code and special handling in a few files.
At least from the logical point of view, having only a single condition should be a subset of having n ORed conditions.

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -126,11 +132,123 @@ bool latematerialized::getReferencedAttributes(AstNode* node,
...
+        if (latematerialized::isPrefix<indexDataOnly>(field.second, nodeAttr.attr, false, postfix)) {
+          tmpSlot.emplace_back(&nodeAttr.afData, fieldNum, &field.second, std::move(postfix));
+          if constexpr (std::is_arithmetic_v<decltype(postfix)>) {
+            postfix = 0;
+          }
+          found = true;
+          break;
+        }
+        ++fieldNum;
+      }
```
jsteemann, line 179: The loop body looks identical to the loop body above (lines 152 to 162), just the input fields and `tmpSlot` different.
Is it possible to implement the loop body just once and call it from both places in this function?

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -126,11 +132,123 @@ bool latematerialized::getReferencedAttributes(AstNode* node,
...
-                                std::vector<std::string>& postfix) {
-  TRI_ASSERT(postfix.empty());
+                                typename ColumnVariant<indexDataOnly>::PostfixType& postfix) {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+  if constexpr (!indexDataOnly) {
+    TRI_ASSERT(postfix.empty());
+  } else {
+    TRI_ASSERT(postfix == 0);
+  }
+#endif
```
jsteemann, line 251: The `#ifdef` here seems unnecessary.

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -61,11 +63,15 @@ struct TraversalState {
 
 }
 
+namespace arangodb {
+namespace aql {
+namespace latematerialized {
```
jsteemann, line 68:
```suggestion
namespace arangodb::aql::latematerialized {
```

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -150,16 +268,23 @@ bool latematerialized::isPrefix(std::vector<arangodb::basics::AttributeName> con
...
+      });
+    }
   }
-
   return true;
 }
 
+} // namespace latematerialized
+} // namespace aql
+} // namespace arangodb
```
jsteemann, line 286:
```suggestion
} // namespace arangodb::aql::latematerialized
```

arangod/Aql/OptimizerRules.cpp
```diff
@@ -3012,13 +3012,15 @@ struct SortToIndexNode final
...
-
+        TRI_ASSERT(usedIndexes.size() == 1);
         IndexIteratorOptions opts;
         opts.ascending = sortCondition.isAscending();
         auto newNode =
             std::make_unique<IndexNode>(_plan, _plan->nextId(),
                                         enumerateCollectionNode->collection(), outVariable,
-                                        usedIndexes, std::move(condition), opts);
+                                        usedIndexes,
+                                        false, // here we could always assume false as there is one index anyway
```
jsteemann, line 3022:
```suggestion
                                        false, // here we could always assume false as there is no lookup condition here
```

arangod/Aql/OptimizerRules.cpp
```diff
@@ -3109,7 +3111,20 @@ struct SortToIndexNode final
...
+      auto support = index->supportsSortCondition(&sortCondition, outVariable, 1);
+      if (support.supportsCondition) {
+        // sort condition is fully covered by index... now we can remove the
+        // sort node from the plan
+        _plan->unlinkNode(_plan->getNodeById(_sortNode->id()));
+        // we need to have a sorted result later on, so we will need a sorted
+        // GatherNode in the cluster
+        indexNode->needsGatherNodeSort(true);
+        _modified = true;
+        handled = true;
```
jsteemann, line 3125: The code that is executed here for an inverted index that covers the sort condition is completely identical to the code below that is executed if a regular index covers the sort condition.
Can we simplify it to something like:
```
bool indexCoversSortCondition = false;
if (invertex index supports sort condition) {
  indexCoversSortCondition = true;
} else if (regular index supports sort condition) {
  indexCoversSortCondition = true;
}

if (indexCoversSortCondition) {
  // sort condition is fully covered by index... now we can remove the
  // sort node from the plan
  _plan->unlinkNode(_plan->getNodeById(_sortNode->id()));
  // we need to have a sorted result later on, so we will need a sorted
  // GatherNode in the cluster
  indexNode->needsGatherNodeSort(true);
  _modified = true;
  handled = true;
}
```

arangod/Aql/OptimizerUtils.cpp
```diff
@@ -443,7 +443,9 @@ std::pair<bool, bool> findIndexHandleForAndNode(
 
   if (bestIndex == nullptr) {
     for (auto const& idx : indexes) {
-      considerIndex(idx);
+      if (!Index::onlyHintForced(idx->type())) {
```
jsteemann, line 446: Why is this a static method call right now? Can we make this a virtual function on the index object?
Or do we need the same function elsewhere in static context too?

arangod/Aql/OptimizerUtils.cpp
```diff
@@ -688,6 +691,28 @@ std::pair<bool, bool> getBestIndexHandlesForFilterCondition(
...
+      if (costs.supportsCondition) {
+        // we need to find 'root' in 'ast' and replace it with specialized version
+        // but for now we know that index will not alter the node, so just an assert
+        index.get()->specializeCondition(root, reference);
+        usedIndexes.emplace_back(index);
+        isAllCoveredByIndex = true;
+        // FIXME: we should somehow consider other indices and calculate here "overall" score
+        // Also a question: if sort is covered but filter is not ? What is more optimal?
+        auto const sortSupport = index.get()->supportsSortCondition(sortCondition, reference, itemsInCollection);
+        return std::make_pair(true, sortSupport.supportsCondition);
```
jsteemann, line 711: It looks like we would always prefer inverted index now. Not saying this is wrong, but if there is a matching inverted index, it will always be preferred over other index types (and even over other index types that support both the filter and the sort condition).

arangod/IResearch/IResearchDocument.h
```diff
@@ -220,6 +220,62 @@ class FieldIterator {
   bool _isDBServer;
 }; // FieldIterator
 
+////////////////////////////////////////////////////////////////////////////////
+/// @brief allows to iterate over the provided VPack accoring the specified
```
jsteemann, line 224:
```suggestion
/// @brief allows to iterate over the provided VPack according to the specified
```

arangod/IResearch/IResearchDocument.h
```diff
@@ -220,6 +220,62 @@ class FieldIterator {
...
+  // but we require it here just to match signature of "FieldIterator" in general
+  explicit InvertedIndexFieldIterator(arangodb::transaction::Methods&, irs::string_ref collection, IndexId indexId);
+
+  bool valid() const noexcept { return _fieldsMeta && _begin != _end;}
+
+  void reset(velocypack::Slice slice,
+             InvertedIndexFieldMeta const& fieldsMeta) {
+    _slice = slice;
+    _fieldsMeta = &fieldsMeta;
+    _begin = _fieldsMeta->_fields.data() - 1;
```
jsteemann, line 247: Looks like we are intentionally pointing to a potentially invalid memory address.
This will work if we always do `++_begin` before dereferencing it, but that needs to be remembered throughout the code. Additionally, with that setup `_begin != _end` even if `_fieldsMeta->_fields.empty()` (not sure if the latter is possible, but it feels a bit strange).

arangod/IResearch/IResearchLink.cpp
```diff
@@ -2017,59 +577,61 @@ AnalyzerPool::ptr IResearchLink::findAnalyzer(AnalyzerPool const& analyzer) cons
...
-    return true;
-  };
+const std::string& IResearchLink::getViewId() const noexcept {
+  return _viewGuid;
+}
 
-  reader->meta().meta.visit_segments(visitor);
-  return stats;
+std::string IResearchLink::getDbName() const {
+  return std::to_string(_collection.vocbase().id());
```
jsteemann, line 593: Is it expected that we return the database ID here and not the database name?

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,257 @@
...
+                                                          engine.rocksDBOptions())};
+    }
+    return irs::directory_attributes{};
+  });
+
+  if (initRes.fail()) {
+    LOG_TOPIC("9c9ac", ERR, iresearch::TOPIC)
+        << std::string(
+               "failed to do an init for index from definition, error setting "
+               "index properties ")
```
jsteemann, line 103: Technically, this is not exactly right. The "index properties" are only set afterwards (line 109). So probably we should change the error message here to better reflect what's going on?
```suggestion
        << "failed to do an init for index from definition, error setting index properties: "
```

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,257 @@
...
+        << initRes.errorMessage();
+    return nullptr;
+  }
+
+  // separate call as here object should be fully constructed and could be safely used by async tasks
+  initRes = index->properties(indexMeta);
+  if (initRes.fail()) {
+    LOG_TOPIC("9c949", ERR, iresearch::TOPIC) <<
+      std::string("failed to initialize index from definition, error setting index properties ")
+      << initRes.errorMessage();
```
jsteemann, line 113:
```suggestion
      "failed to initialize index from definition, error setting index properties: "
      << initRes.errorMessage();
```

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,257 @@
...
+           : (std::string("failed to initialize index from definition, error in attribute '")
+               + errField + "': " + definition.toString()));
+    return nullptr;
+  }
+  auto nameSlice = definition.get(arangodb::StaticStrings::IndexName);
+  std::string indexName;
+  if (!nameSlice.isNone()) {
+    if (!nameSlice.isString() || nameSlice.getStringLength() == 0) {
+      LOG_TOPIC("91ebd", ERR, iresearch::TOPIC) <<
+        std::string("failed to initialize index from definition, error in attribute '")
```
jsteemann, line 76:
```suggestion
        "failed to initialize index from definition, error in attribute '"
```

arangod/Indexes/IndexIterator.h
```diff
@@ -69,8 +69,47 @@ class IndexIterator {
   friend class MultiIndexIterator;
 
  public:
+  class CoveringData {
+   public:
+    virtual VPackSlice at(size_t i) = 0;
```
jsteemann, line 74: Not having a virtual destructor will work, as we don't seem to have any data members. But it may be better for the future to add a virtual destructor here, just in case this struct gets extended.

arangod/RocksDBEngine/RocksDBIndexFactory.cpp
```diff
@@ -390,6 +392,8 @@ RocksDBIndexFactory::RocksDBIndexFactory(application_features::ApplicationServer
   emplace("skiplist", skiplistIndexFactory);
   emplace("ttl", ttlIndexFactory);
   emplace("zkd", zkdIndexFactory);
+  emplace(arangodb::iresearch::IRESEARCH_INVERTED_INDEX_TYPE.data(),
```
jsteemann, line 395: IRESEARCH_INVERTED_INDEX_TYPE is an `std::string_view`, which AFAIK does not guarantee to be null-terminated.

From https://en.cppreference.com/w/cpp/string/basic_string_view/data:

> Unlike std::basic_string::data() and string literals, data() may return a pointer to a buffer that is not null-terminated. Therefore it is typically a mistake to pass data() to a routine that takes just a const CharT* and expects a null-terminated string.

arangod/RocksDBEngine/RocksDBPrimaryIndex.cpp
```diff
@@ -156,7 +156,8 @@ class RocksDBPrimaryIndexEqIterator final : public IndexIterator {
     LocalDocumentId documentId =
         _index->lookupKey(_trx, arangodb::velocypack::StringRef(_key->slice()), canReadOwnWrites());
     if (documentId.isSet()) {
-      cb(documentId, _key->slice());
+      auto data = SliceArrayCoveringData(_key->slice());
```
jsteemann, line 159: Do we really need `SliceArrayCoveringData` here?
The primary index does not store VPack arrays, but stores the indexed `_key` values as simple strings (not even vpack).

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -143,7 +143,8 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
       coveringBuilder->add(_lastKey);
       coveringBuilder->add(fromTo);
       coveringBuilder->close();
-      cb(docId, coveringBuilder->slice());
+      auto data = SliceArrayCoveringData(coveringBuilder->slice());
```
jsteemann, line 146: Not sure if we need `SliceArrayCoveringData` here. The edge index stores the edge id as a simple string, and not as vpack.

tests/Mocks/StorageEngineMock.cpp
```diff
@@ -866,7 +867,7 @@ class HashIndexMock final : public arangodb::Index {
   std::unique_ptr<arangodb::IndexIterator> iteratorForCondition(
       arangodb::transaction::Methods* trx, arangodb::aql::AstNode const* node,
       arangodb::aql::Variable const*, arangodb::IndexIteratorOptions const&,
-      arangodb::ReadOwnWrites) override {
+      arangodb::ReadOwnWrites,  int) override {
```
jsteemann, line 870:
```suggestion
      arangodb::ReadOwnWrites, int) override {
```

arangod/IResearch/IResearchCommon.cpp
Dronplane: But we already use Msec in code and documentation. So let it be as is

arangod/IResearch/IResearchDocument.h
Dronplane: Yes, this is intentionally  for simplifying `next` implementation. So it could always  increment `_begin` without check if it is initial step or not. But you are right - we need an assertion that `_fieldsMeta->_fields.empty()` is false (Index factory forbids to create index with empty fields list)

arangod/IResearch/IResearchLink.cpp
Dronplane: Good catch! Db name is much more user-friendly

arangod/RocksDBEngine/RocksDBIndexFactory.cpp
Dronplane: IRESEARCH_INVERTED_INDEX_TYPE is initialized as 
`constexpr std::string_view IRESEARCH_INVERTED_INDEX_TYPE = "inverted";`
And string literal is always zero terminated. So here this is safe. But in general yes - string_view could be without zero terminator

arangod/IResearch/IResearchDocument.cpp
MBkkt: .

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -715,6 +705,290 @@ bool StoredValue::write(irs::data_output& out) const {
...
+          // fo array subobjects we index "null" in case of absence as declared for other indicies
+          _valueSlice = get(*_arrayStack.back(), _begin->expansion,
+                            arangodb::velocypack::Slice::nullSlice());
+        }
+        ++_arrayStack.back();
+        _nameBuffer.resize(_prefixLength); // FIXME: just clear should work!
+        break;
+      } else {
+        _arrayStack.pop_back();
+      }
```
MBkkt, line 744:
```suggestion
      _arrayStack.pop_back();
```

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
Dronplane: renamed to SliceCoveringData to exclude misleading Array from name

arangod/RocksDBEngine/RocksDBPrimaryIndex.cpp
Dronplane: renamed to SliceCoveringData to exclude misleading Array from name

arangod/IResearch/IResearchCommon.h
```diff
@@ -39,6 +39,9 @@ arangodb::LogTopic& logTopic();
 ADB_IGNORE_UNUSED static auto& DATA_SOURCE_TYPE = dataSourceType();
 ADB_IGNORE_UNUSED extern arangodb::LogTopic TOPIC;
 
+
+constexpr std::string_view IRESEARCH_INVERTED_INDEX_TYPE = "inverted";
```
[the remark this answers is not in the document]
Dronplane: experimets proved inline makes no sense here

arangod/IResearch/IResearchDocument.cpp
Dronplane: We should continue the cycle here. Iterator added to the _arrayStack  should be immediately used.

arangod/Aql/OptimizerUtils.cpp
Dronplane: For now this function only checks index type so I want to avoid virtual call.

arangod/Aql/OptimizerUtils.cpp
Dronplane: No, inverted index for now is only permitted by explicit hint.

arangod/IResearch/IResearchDataStore.h
```diff
@@ -0,0 +1,467 @@
...
+    try {
+      // hold references even after transaction
+      auto filter = std::make_unique<PrimaryKeyFilterContainer>(std::move(_removals));
+      _ctx.remove(std::unique_ptr<irs::filter>(std::move(filter)));
+    } catch (std::exception const& e) {
+      LOG_TOPIC("eb463", ERR, arangodb::iresearch::TOPIC)
+          << "caught exception while applying accumulated removals: " << e.what();
+    } catch (...) {
+      LOG_TOPIC("14917", WARN, arangodb::iresearch::TOPIC)
+          << "caught exception while applying accumulated removals";
```
MBkkt, line 111: Why one of them is warn, but another is error?

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
Dronplane: Fixed. Good catch.

arangod/IResearch/IResearchViewStoredValues.h
Dronplane: The suggested code has different behaviour than original. But as we are on c++20 there is even better way  - `ends_with`

arangod/IResearch/IResearchDocument.cpp
MBkkt: We decide make separate PR for this

gnusi (changes requested):

arangod/IResearch/AqlHelper.h
```diff
@@ -526,6 +526,54 @@ bool nameFromAttributeAccess(std::string& name, aql::AstNode const& node,
...
+  if (!node) {
+    return false;
+  }
+  std::string name;
+  switch (node->type) {
+    case aql::NODE_TYPE_ATTRIBUTE_ACCESS:
+    case aql::NODE_TYPE_INDEXED_ACCESS:
+    case aql::NODE_TYPE_EXPANSION:
+      if (checkAttributeAccess(node, ref) &&
+          nameFromAttributeAccess(name, *node, ctx)) {
```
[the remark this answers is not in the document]
Dronplane: Done

arangod/Aql/IndexExecutor.cpp
```diff
@@ -76,15 +76,15 @@ static void resolveFCallConstAttributes(Ast* ast, AstNode* fcall) {
...
+template<bool checkUniqueness, bool skip>
+IndexIterator::CoveringCallback getCallback(
     DocumentProducingFunctionContext& context,
     transaction::Methods::IndexHandle const& index,
     IndexNode::IndexValuesVars const& outNonMaterializedIndVars,
     IndexNode::IndexValuesRegisters const& outNonMaterializedIndRegs) {
   return [&context, &index, &outNonMaterializedIndVars,
           &outNonMaterializedIndRegs](LocalDocumentId const& token,
-                                      VPackSlice slice) {
+                                      IndexIterator::CoveringData* covering) {
```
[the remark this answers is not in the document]
Dronplane: Done

arangod/IResearch/IResearchLinkMeta.cpp
```diff
@@ -902,5 +896,548 @@ size_t IResearchLinkMeta::memory() const noexcept {
...
+            return false;
+          }
+
+          name = value.get(subFieldName).copyString();
+          if (!defaultVocbase.null()) {
+            name =
+                IResearchAnalyzerFeature::normalize(name, defaultVocbase, true);
+          }
+        }
+        irs::string_ref type;
```
MBkkt, line 996: std::string_view?

arangod/Indexes/IndexFactory.cpp
```diff
@@ -316,9 +317,12 @@ std::shared_ptr<Index> IndexFactory::prepareIndexFromSlice(
 
 /// same for both storage engines
 std::vector<std::string> IndexFactory::supportedIndexes() const {
```
MBkkt, line 319: string_view?

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,314 @@
...
+  InvertedIndexFieldMeta const& meta() const noexcept { return _meta; }
+
+ protected:
+  void invalidateQueryCache(TRI_vocbase_t* vocbase) override;
+
+ private:
+  InvertedIndexFieldMeta _meta;
+};
+
+class IResearchInvertedClusterIndex : public IResearchInvertedIndex,
```
MBkkt, line 221: Also, cluster inheritance by datastore

arangod/IResearch/IResearchDataStore.h
```diff
@@ -0,0 +1,476 @@
...
+  //////////////////////////////////////////////////////////////////////////////
+  /// @brief remove statistic from MetricsFeature
+  //////////////////////////////////////////////////////////////////////////////
+  virtual void removeStats(){};
+
+  //////////////////////////////////////////////////////////////////////////////
+  /// @brief update store statistics in MetricsFeature
+  //////////////////////////////////////////////////////////////////////////////
+  virtual void updateStats(Stats const&){};
+
```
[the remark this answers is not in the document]
Dronplane: No, virtual methods are about  "publishing" stats. So the descendants should decide how to do that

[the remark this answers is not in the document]
MBkkt: Ok, why not pure virtual?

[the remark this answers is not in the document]
Dronplane: To not force metrics usage if  not applicable for descendants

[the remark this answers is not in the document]
MBkkt: I think in that case 
> if not applicable for descendants
class doesn't want have metrics in their layout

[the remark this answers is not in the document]
Dronplane: But still may need stats `Stats` to provide by other means like `IResearchRocksDBInvertedIndex::memory` BTW forgot to update this method. 😓

arangod/Aql/IndexExecutor.cpp
```diff
@@ -379,9 +392,20 @@ IndexExecutor::CursorReader::CursorReader(
...
+                                         _infos.getOutNonMaterializedIndVars(),
+                                         _infos.getOutNonMaterializedIndRegs());
+  } else {
+    _documentSkipper = checkUniqueness
+                           ? buildDocumentCallback<true, true>(context)
+                           : buildDocumentCallback<false, true>(context);
+  }
 }
 
 IndexExecutor::CursorReader::CursorReader(CursorReader&& other) noexcept
```
gnusi, line 411: can we use default move ctor?

arangod/Aql/OptimizerRulesReplaceFunctions.cpp
```diff
@@ -557,6 +557,8 @@ AstNode* replaceFullText(AstNode* funAstNode, ExecutionNode* calcNode,
       new IndexNode(plan, plan->nextId(), aqlCollection, indexOutVariable,
                     std::vector<transaction::Methods::IndexHandle>{
                         transaction::Methods::IndexHandle{index}},
+                    false,  // here we are not using inverted index so for sure
```
gnusi, line 560: maybe create a constant to avoid comment on the same line

arangod/Aql/OptimizerRules.cpp
```diff
@@ -7160,6 +7170,8 @@ static bool applyGeoOptimization(ExecutionPlan* plan, LimitNode* ln,
                              info.collectionNodeOutVar,
                              std::vector<transaction::Methods::IndexHandle>{
                                  transaction::Methods::IndexHandle{info.index}},
+                             false,  // here we are not using inverted index so
```
gnusi, line 7173: maybe create a constant to avoid comment on the same line

arangod/Aql/OptimizerUtils.cpp
```diff
@@ -720,6 +723,33 @@ std::pair<bool, bool> getBestIndexHandlesForFilterCondition(
 
   TRI_ASSERT(usedIndexes.empty());
 
+  // we might have an inverted index - it could cover whole condition at once.
+  // Give it a try
+  for (auto& index : indexes) {
+    if (index.get()->type() == Index::TRI_IDX_TYPE_INVERTED_INDEX &&
+        ((hint.type() ==
+              IndexHint::Simple &&  // apply this index only if the hint-а
```
gnusi, line 731: move comment to the line above

arangod/IResearch/AqlHelper.cpp
```diff
@@ -720,11 +726,12 @@ bool nameFromAttributeAccess(std::string& name, aql::AstNode const& node,
       (*str_) += NESTING_LIST_OFFSET_SUFFIX;
     }
 
+    bool _expansion;
```
gnusi, line 729: move `_expansion` after `buf_` to reduce memory footprint

arangod/IResearch/AqlHelper.h
```diff
@@ -516,15 +535,67 @@ bool attributeAccessEqual(aql::AstNode const* lhs, aql::AstNode const* rhs,
...
+/// @param node node to visit
+/// @param ref variable to check for access
+/// @param ctx current query context
+/// @param allowExpansion allow to have [*] in path
+/// @param visitor visitor to apply
+/// @return true if all access were valid and all visitor calls returned true,
+/// false otherwise
+////////////////////////////////////////////////////////////////////////////////
+template<typename Visitor>
+bool visitAllAttributeAccess(aql::AstNode const* node, aql::Variable const& ref,
```
gnusi, line 564: nameFromAttributeAccess`es`?

arangod/IResearch/IResearchAnalyzerFeature.cpp
```diff
@@ -725,14 +725,17 @@ bool analyzerInUse(application_features::ApplicationServer& server,
 
       for (auto const& index : collection->getIndexes()) {
         if (!index ||
-            arangodb::Index::TRI_IDX_TYPE_IRESEARCH_LINK != index->type()) {
+            (arangodb::Index::TRI_IDX_TYPE_IRESEARCH_LINK != index->type() &&
```
gnusi, line 728: can remove `arangodb::`

arangod/IResearch/IResearchCommon.h
```diff
@@ -151,7 +153,9 @@ struct StaticStrings {
   /// @brief the name of the field in the IResearch Link definition denoting the
   ///        corresponding collection name in cluster (not shard name!)
   ////////////////////////////////////////////////////////////////////////////////
-  static constexpr std::string_view CollectionNameField{"collectionName"};
+==== BASE ====
```
gnusi: leftover?

arangod/Aql/IResearchViewOptimizerRules.h
```diff
@@ -26,6 +26,8 @@
 
 #include <memory>
 
+#include "Aql/LateMaterializedOptimizerRulesCommon.h"
+
```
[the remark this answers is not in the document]
Dronplane: Necessary. Some stuff were moved to that file.

arangod/Aql/IndexExecutor.cpp
```diff
@@ -139,45 +139,45 @@ IndexIterator::DocumentCallback getCallback(
...
+      AqlValue v(AqlValueHintUInt(token.id()));
+      AqlValueGuard guard{v, true};
+      TRI_ASSERT(!output.isFull());
+      output.moveValueInto(registerId, input, guard);
+
+      // hash/skiplist/persistent
+      if (covering.isArray()) {
+        for (auto const& indReg : outNonMaterializedIndRegs.second) {
+          TRI_ASSERT(indReg.first < covering.length());
+          if (ADB_UNLIKELY(indReg.first >= covering.length())) {
```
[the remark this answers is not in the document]
Dronplane: ADB_UNLIKELY here is the common approach

arangod/Aql/IndexExecutor.cpp
```diff
@@ -139,45 +139,45 @@ IndexIterator::DocumentCallback getCallback(
...
+          auto s = covering.at(indReg.first);
+          AqlValue v(s);
+          AqlValueGuard guard{v, true};
+          TRI_ASSERT(!output.isFull());
+          output.moveValueInto(indReg.second, input, guard);
+        }
+      } else {  // primary/edge
+        auto indReg = outNonMaterializedIndRegs.second.cbegin();
+        TRI_ASSERT(indReg != outNonMaterializedIndRegs.second.cend());
+        if (ADB_UNLIKELY(indReg == outNonMaterializedIndRegs.second.cend())) {
```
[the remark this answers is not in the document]
Dronplane: ADB_UNLIKELY here is the common approach

arangod/Aql/IndexNode.cpp
```diff
@@ -424,13 +430,14 @@ CostEstimate IndexNode::estimateCost() const {
   double totalCost = 0.0;
 
   auto root = _condition->root();
-
+  TRI_ASSERT(!_allCoveredByOneIndex || _indexes.size() == 1);
   for (size_t i = 0; i < _indexes.size(); ++i) {
     Index::FilterCosts costs =
         Index::FilterCosts::defaultCosts(itemsInCollection);
 
     if (root != nullptr && root->numMembers() > i) {
```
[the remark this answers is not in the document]
Dronplane: Yes, empty root node is useless. Anyway I want to rethink the concept of the "allCoveredByOneIndex" in the next PR. As it looks like for conjunctions this is not needed. And for disjunction I guess there is a way to get rid of this "kludgy" solution

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -60,14 +62,41 @@ struct TraversalState {
   bool wasAccess;
 };
 
+template<bool indexDataOnly, typename Fields, typename Slot, typename Node>
+bool findMatchedField(Fields const& fields, Slot& tmpSlot, Node& nodeAttr) {
+  size_t fieldNum = 0;
+  typename arangodb::aql::latematerialized::ColumnVariant<
+      indexDataOnly>::PostfixType postfix{};
+  for (auto const& field : fields) {
```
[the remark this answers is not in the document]
Dronplane: I prefer to keep const to keep it clear to the reader.

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -60,14 +62,41 @@ struct TraversalState {
...
+    std::vector<arangodb::basics::AttributeName> const* fieldValue;
+    if constexpr (std::is_same_v<
+                      std::vector<arangodb::basics::AttributeName> const&,
+                      decltype(field)>) {
+      fieldValue = &field;
+    } else {
+      fieldValue = &field.second;
+    }
+    TRI_ASSERT(fieldValue);
```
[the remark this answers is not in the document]
Dronplane: Indeed useless.

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -133,11 +164,113 @@ bool latematerialized::getReferencedAttributes(AstNode* node,
...
+  static_assert(std::is_move_constructible_v<ColumnVariant<indexDataOnly>>,
+                "To efficiently move from temp variable we need working move "
+                "for ColumnVariant");
+  // store only on successful exit, otherwise pointers to afData will be
+  // invalidated as Node will be not stored!
+  size_t current = 0;
+  for (auto it = tmpUsedColumnsCounter.begin();
+       it != tmpUsedColumnsCounter.end(); ++it) {
+    std::move(it->begin(), it->end(),
+              irs::irstd::back_emplacer(usedColumnsCounter[current++]));
```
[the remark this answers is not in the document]
Dronplane: Well, it`s not my code, so can't say about initial intention. But at first glance your suggestion looks viable.

arangod/Aql/LateMaterializedOptimizerRulesCommon.cpp
```diff
@@ -133,11 +164,113 @@ bool latematerialized::getReferencedAttributes(AstNode* node,
...
+  }
+  return true;
+}
+
+template<bool indexDataOnly>
+void setAttributesMaxMatchedColumns(
+    std::vector<std::vector<ColumnVariant<indexDataOnly>>>& usedColumnsCounter,
+    size_t columnsCount) {
+  TRI_ASSERT(columnsCount <= usedColumnsCounter.size());
+  std::vector<ptrdiff_t> idx(columnsCount);
```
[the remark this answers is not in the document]
Dronplane: I guess I know. )  See AstAndColumnFieldData::columnNumber - it is ptrdiff_t as -1 is "sort" column index. So the author of initial code decided to have ptrdiff_t in all places. Just my speculation though.

arangod/ClusterEngine/ClusterIndexFactory.cpp
```diff
@@ -144,6 +147,64 @@ struct PrimaryIndexFactory : public DefaultIndexFactory {
...
+            << "failed to initialize index from definition, error in attribute "
+               "'" +
+                   arangodb::StaticStrings::IndexName +
+                   "': " + definition.toString();
+        return nullptr;
+      }
+      indexName = nameSlice.copyString();
+    }
+    auto objectId = basics::VelocyPackHelper::stringUInt64(
+        definition, arangodb::StaticStrings::ObjectId);
```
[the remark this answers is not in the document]
Dronplane: There are "arangodb::StaticStrings" and "arangodb::iresearch::StaticStrings" so here we have to be explicit

arangod/ClusterEngine/ClusterIndexFactory.cpp
```diff
@@ -144,6 +147,64 @@ struct PrimaryIndexFactory : public DefaultIndexFactory {
...
+                  ? (std::string("failed to initialize index fields from "
+                                 "definition: ") +
+                     definition.toString())
+                  : (std::string(
+                         "failed to initialize index fields from definition, "
+                         "error in attribute '") +
+                     errField + "': " + definition.toString()));
+      return nullptr;
+    }
+    auto nameSlice = definition.get(arangodb::StaticStrings::IndexName);
```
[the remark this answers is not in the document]
Dronplane: There are "arangodb::StaticStrings" and "arangodb::iresearch::StaticStrings" so here we have to be explicit

arangod/IResearch/IResearchCommon.cpp
```diff
@@ -62,6 +62,19 @@ arangodb::LogTopic& logTopic() { return TOPIC; }
...
+    "commitIntervalMsec");
+/*static*/ std::string const StaticStrings::CleanupIntervalStep(
+    "cleanupIntervalStep");
+/*static*/ std::string const StaticStrings::ConsolidationPolicy(
+    "consolidationPolicy");
+/*static*/ std::string const StaticStrings::WritebufferActive(
+    "writebufferActive");
+/*static*/ std::string const StaticStrings::WritebufferIdle("writebufferIdle");
+/*static*/ std::string const StaticStrings::WritebufferSizeMax(
+    "writebufferSizeMax");
```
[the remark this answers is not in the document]
Dronplane: Already fixed

arangod/IResearch/AqlHelper.h
```diff
@@ -421,11 +421,30 @@ bool visitAttributeAccess(aql::AstNode const*& head, aql::AstNode const* node,
...
+          return aql::NODE_TYPE_ITERATOR == itr->type && var &&
+                 aql::NODE_TYPE_VARIABLE == var->type &&
+                 visitAttributeAccess(head, root, visitor) &&  // 1st visit root
+                 visitor.expansion(*node) &&  // 2nd visit current node
+                 visitAttributeAccess(tailHead, ref,
+                                      visitor);  // 3rd visit the tail
+        } else {                                 // unexpected expansion form
+          TRI_ASSERT(false);
+          return false;
+        }
```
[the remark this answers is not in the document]
Dronplane: If  NODE_TYPE_EXPANSION node structure will be changed and we will forget to update this function we definitely want to have an assertion triggered. And for prod we at least will fail parsing (this will most likely fail the query but in graceful manner)

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -746,6 +736,296 @@ bool StoredValue::write(irs::data_output& out) const {
...
+      valueRef = iresearch::ref<char>(value);
+      valueType = AnalyzerValueType::Object;
+    } break;
+    case VPackValueType::String: {
+      valueRef = iresearch::getStringRef(value);
+      valueType = AnalyzerValueType::String;
+    } break;
+    default:
+      TRI_ASSERT(false);
+      return false;
```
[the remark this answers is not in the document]
Dronplane: Potentially, if we will forget to update function after  some changes in the AQL. So prefer to keep assert in place

arangod/IResearch/IResearchFilterFactory.h
```diff
@@ -49,14 +50,19 @@ namespace iresearch {
...
   ////////////////////////////////////////////////////////////////////////////////
   /// @brief determine if the 'node' can be converted into an iresearch filter
   ///        if 'filter' != nullptr then also append the iresearch filter there
   ////////////////////////////////////////////////////////////////////////////////
   static arangodb::Result filter(irs::boolean_filter* filter,
                                  QueryContext const& ctx,
-                                 arangodb::aql::AstNode const& node);
+                                 arangodb::aql::AstNode const& node,
+                                 bool forSearch = true,
+                                 AnalyzerProvider const* provider = nullptr);
```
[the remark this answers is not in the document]
Dronplane: Hm. You have any particular reason in mind to restrict ability to make stateful providers? As to me  it looks useful.

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+  explicit CoveringValue(irs::string_ref col) : column(col) {}
+  CoveringValue(CoveringValue&& other) noexcept : column(other.column) {}
+
+  void reset(irs::sub_reader const& rdr) {
+    itr.reset();
+    value = &NoPayload;
+    // FIXME: this is cheap. Keep it here?
+    auto extraValuesReader = column.empty() ? rdr.sort() : rdr.column(column);
+    // FIXME: this is expensive - move it to get and do lazily?
+    if (ADB_LIKELY(extraValuesReader)) {
```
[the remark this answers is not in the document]
Dronplane: ADB_LIKELY is the common pattern

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+      itr = extraValuesReader->iterator(false);
+      TRI_ASSERT(itr);
+      if (ADB_LIKELY(itr)) {
```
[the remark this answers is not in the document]
Dronplane: ADB_LIKELY

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        value = irs::get<irs::payload>(*itr);
+        TRI_ASSERT(value);
+        if (ADB_UNLIKELY(!value)) {
```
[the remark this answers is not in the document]
Dronplane: ADB_LIKELY is the common pattern

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+      if (index == 0 && totalSize == 0) {
+        // one empty field optimization
+        return VPackSlice::nullSlice();
+      }
+      TRI_ASSERT(totalSize > 0);
+      size_t size = 0;
+      VPackSlice slice(value->value.c_str());
+      TRI_ASSERT(slice.byteSize() <= totalSize);
+      while (i < index) {
+        if (ADB_LIKELY(size < totalSize)) {
```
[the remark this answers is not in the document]
Dronplane: ADB_LIKELY is the common pattern

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+      doc_.value = base_ - 1;
+      continue;
+    }
+    doc_.value = irs::doc_limits::eof();
+    word_ = 0;
+    return false;
+  }
+  const irs::doc_id_t delta =
+      irs::doc_id_t(irs::math::math_traits<lazy_bitset::word_t>::ctz(word_));
+  assert(delta < irs::bits_required<lazy_bitset::word_t>());
```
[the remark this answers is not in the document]
Dronplane: No. This code eventually will land in the upstream iresearch. so "assert"

arangod/IResearch/IResearchInvertedIndex.cpp
```diff
@@ -0,0 +1,1265 @@
...
+  if (!supportsFilterNode(id(), indexedFields, node, reference,
+                          &analyzerProvider)) {
+    TRI_ASSERT(node->type == aql::AstNodeType::NODE_TYPE_OPERATOR_NARY_AND);
+    std::vector<arangodb::aql::AstNode const*> children;
+    size_t const n = node->numMembers();
+    for (size_t i = 0; i < n; ++i) {
+      auto part = node->getMemberUnchecked(i);
+      if (supportsFilterNode(id(), indexedFields, part, reference,
+                             &analyzerProvider)) {
+        children.push_back(part);
```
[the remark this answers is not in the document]
Dronplane: We don't know exact size in advance. And reserving for the upper bound  will not give us much.

arangod/Aql/IndexExecutor.cpp
Dronplane: Indeed

arangod/IResearch/IResearchFilterFactory.h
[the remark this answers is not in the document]
MBkkt: Only that in the code I haven't seen stateful providers yet.
If we need its in the future we will can change it.

arangod/IResearch/IResearchKludge.cpp
```diff
@@ -46,27 +46,75 @@ const char ANALYZER_DELIMITER = '\1';
...
 irs::string_ref const NUMERIC_SUFFIX("\0_d", 3);
+irs::string_ref const STRING_SUFFIX("\0_s", 3);
 
-void mangleType(std::string& name) { name += TYPE_DELIMITER; }
+void mangleType(std::string& name) {
+  // remove the last expansion as it could be omitted accodring to our
+  // indicies behaviour
+  if (name.ends_with("[*]")) {
+    name.resize(name.size() - 3);
+  }
```
gnusi, line 56: move to separate inline function?

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,314 @@
...
+          *cached_bitset, rdr.docs_count());
+    } else {
+      if (!cache_.prepared_real_filter_) {
+        cache_.prepared_real_filter_ = real_filter_->prepare(index_, order_);
+      }
+      auto inserted = cache_.readers_.insert(
+          {&rdr, std::make_unique<lazy_bitset>(
+                     rdr, *(cache_.prepared_real_filter_.get()))});
+      return irs::memory::make_managed<lazy_filter_bitset_iterator>(
+          *inserted.first->second.get(), rdr.docs_count());
```
gnusi, line 120: I suggest we simplify this
```
auto& [unused, cache] = cache_.reader.emplace(&rdr, nullptr);

if (!cached) {
   if (!cache_.prepared_real_filter_) {
        cache_.prepared_real_filter_ = real_filter_->prepare(index_, order_);
   }
        
   cached = std::make_unique<lazy_bitset>(rdr, *cache_.prepared_real_filter_);
}

assert(cached);
return irs::memory::make_managed<lazy_filter_bitset_iterator>(
          *cached, rdr.docs_count());
```

gnusi (changes requested):

arangod/IResearch/IResearchFilterFactory.cpp
```diff
@@ -4123,18 +4382,21 @@ namespace iresearch {
...
+    AnalyzerProvider const* provider /*= nullptr*/) {
   if (node.willUseV8()) {
     return {TRI_ERROR_NOT_IMPLEMENTED,
             "using V8 dependent function is not allowed in SEARCH statement"};
   }
 
   // The analyzer is referenced in the FilterContext and used during the
   // following ::filter() call, so may not be a temporary.
-  FieldMeta::Analyzer analyzer{IResearchAnalyzerFeature::identity()};
+==== BASE ====
```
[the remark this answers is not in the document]
Dronplane: already fixed )

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,320 @@
...
+  Result res;
+  try {
+    res = shutdownDataStore();
+  } catch (...) {
+    res = {TRI_ERROR_INTERNAL};
+  }
+  if (!res.ok()) {
+    LOG_TOPIC("2b41f", ERR, iresearch::TOPIC)
+        << "failed to unload arangosearch rockdb inverted index destructor: "
+        << res.errorNumber() << " " << res.errorMessage();
```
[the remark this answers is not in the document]
Dronplane: I guess yes. Good point. But as we want to not have shutdownDataStore call  in dtors  - I could just remove this code here.

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,320 @@
...
+  //   normalized.add(arangodb::StaticStrings::IndexId,
+  //   definition.get(arangodb::StaticStrings::IndexId));
+  // }
+
+  normalized.add(arangodb::StaticStrings::IndexSparse,
+                 arangodb::velocypack::Value(true));
+  normalized.add(arangodb::StaticStrings::IndexUnique,
+                 arangodb::velocypack::Value(false));
+
+  // FIXME: make indexing true background?
```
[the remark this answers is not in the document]
Dronplane: Leftover

arangod/IResearch/IResearchRocksDBInvertedIndex.cpp
```diff
@@ -0,0 +1,315 @@
...
+      RocksDBIndex(id, collection, name, IResearchInvertedIndex::fields(meta()),
+                   false, true,
+                   RocksDBColumnFamilyManager::get(
+                       RocksDBColumnFamilyManager::Family::Invalid),
+                   objectId, false) {}
+
+IResearchRocksDBInvertedIndex::~IResearchRocksDBInvertedIndex() {
+  // if triggered  - no unload was called prior to deleting index object
+  TRI_ASSERT(!_dataStore);
+}
```
gnusi, line 252: for release builds:
```
IResearchRocksDBInvertedIndex::~IResearchRocksDBInvertedIndex() = default; 
```

arangod/IResearch/IResearchView.cpp
```diff
@@ -113,7 +113,7 @@ class ViewTrxState final : public arangodb::TransactionState::Cookie,
   size_t _docs_count{};
   size_t _live_docs_count{};
   std::unordered_set<arangodb::DataSourceId> _collections;
```
gnusi, line 115: `std::unordered_set<arangodb::DataSourceId>` -> `FlastHashMap<DataSourceId>`

arangod/Indexes/Index.cpp
```diff
@@ -363,14 +372,23 @@ Index::IndexType Index::type(char const* type, size_t len) {
   if (::typeMatch(type, len, "noaccess")) {
     return TRI_IDX_TYPE_NO_ACCESS_INDEX;
   }
-
+  if (::typeMatch(type, len,
+                  arangodb::iresearch::IRESEARCH_INVERTED_INDEX_TYPE.data())) {
+    return TRI_IDX_TYPE_INVERTED_INDEX;
+  }
```
gnusi, line 378:
```
if (arangodb::iresearch::IRESEARCH_INVERTED_INDEX_TYPE == std::string_view{ type, len }
 ...
```

arangod/Pregel/IndexHelpers.cpp
```diff
@@ -64,7 +64,7 @@ std::unique_ptr<arangodb::IndexIterator> EdgeCollectionInfo::getEdges(
   opts.enableCache = false;
   return _trx->indexScanForCondition(
       _index, _searchBuilder.getOutboundCondition(),
-      _searchBuilder.getVariable(), opts, ReadOwnWrites::no);
+      _searchBuilder.getVariable(), opts, ReadOwnWrites::no, -1);
```
gnusi, line 67: constant?

tests/IResearch/IResearchDocumentTest.cpp
```diff
@@ -681,8 +698,7 @@ TEST_F(IResearchDocumentTest,
       {mangleStringIdentity("array.id"), 3},
       {mangleStringIdentity("array.subarr"), 9},
       {mangleStringIdentity("array.subobj.id"), 2},
-      {mangleStringIdentity("array.subobj.name"), 1},
-      {mangleStringIdentity("array.id"), 2}};
```
gnusi, line 685: why was it there before?

arangod/IResearch/IResearchInvertedIndex.h
Dronplane: 👍

arangod/IResearch/IResearchFilterFactory.cpp
```diff
@@ -2247,51 +2457,52 @@ Result fromFuncExists(char const* funcName, irs::boolean_filter* filter,
       irs::string_ref const TypeAnalyzer("analyzer");
 
       typedef bool (*TypeHandler)(
-          std::string&, arangodb::iresearch::FieldMeta::Analyzer const&);
+          std::string&, bool isIndexFilter,
+          arangodb::iresearch::FieldMeta::Analyzer const&);
 
       static std::map<irs::string_ref, TypeHandler> const TypeHandlers{
```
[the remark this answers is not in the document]
Dronplane: Good point. But  as this code was not introduced by this PR, let's address that in the next PR ) I just want to merge this one eventually 😅

arangod/IResearch/IResearchInvertedIndex.h
```diff
@@ -0,0 +1,312 @@
...
+  irs::document doc_;
+  irs::doc_id_t word_idx_{0};
+  lazy_bitset::word_t word_{0};
+  irs::doc_id_t base_{irs::doc_limits::invalid()};
+};
+
+class proxy_query final : public irs::filter::prepared {
+ public:
+  struct proxy_cache {
+    std::map<const irs::sub_reader*, std::unique_ptr<lazy_bitset>> readers_;
```
gnusi, line 90: maybe `absl::flat_hash_map`?

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+    TRI_ASSERT(false);
+    return false;
+  }
+
+  {
+    // clear existing definitions
+    _analyzerDefinitions.clear();
+
+    // optional object list
+    static const std::string fieldName("analyzerDefinitions");
```
gnusi, line 106: `constexpr std::string_view kFieldName{"analyzerDefinitions}`

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+    static VPackStringRef const fieldName("storedValues");
+
+    auto const field = slice.get(fieldName);
+    if (!field.isNone() && !_storedValues.fromVelocyPack(field, errorField)) {
+      return false;
+    }
+  }
+  {
+    // optional sort compression
+    static VPackStringRef const fieldName("primarySortCompression");
```
gnusi, line 53: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+namespace arangodb::iresearch {
+
+bool IResearchInvertedIndexMeta::init(
+    arangodb::application_features::ApplicationServer& server,
+    VPackSlice const& slice, bool readAnalyzerDefinition,
+    std::string& errorField, irs::string_ref const defaultVocbase) {
+  // copied part begin FIXME: verify and move to common base class if possible
+  {
+    // optional stored values
+    static VPackStringRef const fieldName("storedValues");
```
gnusi, line 44: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+    auto const field = slice.get(fieldName);
+
+    if (!field.isNone() && ((_sortCompression = columnCompressionFromString(
+                                 getStringRef(field))) == nullptr)) {
+      return false;
+    }
+  }
+  {
+    // optional primarySort
+    static VPackStringRef const fieldName("primarySort");
```
gnusi, line 63: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+    auto const field = slice.get(fieldName);
+    if (!field.isNone() && !_sort.fromVelocyPack(field, errorField)) {
+      return false;
+    }
+  }
+  {
+    // optional version
+    VPackStringRef constexpr fieldName("version");
```
gnusi, line 71: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+          errorField = fieldName + "[" + std::to_string(itr.index()) + "]";
+
+          return false;
+        }
+
+        std::string name;
+
+        {
+          // required string value
+          static const std::string subFieldName("name");
```
gnusi, line 132: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+          if (!defaultVocbase.null()) {
+            name =
+                IResearchAnalyzerFeature::normalize(name, defaultVocbase, true);
+          }
+        }
+        std::string_view type;
+
+        {
+          // required string value
+          static const std::string subFieldName("type");
```
gnusi, line 152: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+          }
+
+          type = typeSlice.stringView();
+        }
+
+        VPackSlice properties;
+
+        {
+          // optional string value
+          static const std::string subFieldName("properties");
```
gnusi, line 169: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+              return false;
+            }
+
+            properties = subField;
+          }
+        }
+        Features features;
+        {
+          // optional string list
+          static const std::string subFieldName("features");
```
gnusi, line 187: constexpr std::string_view kFieldName

arangod/IResearch/IResearchInvertedIndexMeta.cpp
```diff
@@ -0,0 +1,598 @@
...
+            errorField.append(": ").append(res.errorMessage());
+          }
+          return false;
+        }
+        _analyzerDefinitions.emplace(analyzer);
+      }
+    }
+  }
+  // end of the copied part
+  static const std::string fieldsFieldName("fields");
```
gnusi, line 254: constexpr std::string_view kFieldName

arangod/Indexes/IndexIterator.h
```diff
@@ -63,17 +63,52 @@ class Methods;
 
 struct IndexIteratorOptions;
 
+class IndexIteratorCoveringData {
+ public:
+  virtual ~IndexIteratorCoveringData() = default;
+  virtual VPackSlice at(size_t i) = 0;
+  virtual bool isArray() const noexcept = 0;
+  virtual VPackSlice value() const {
```
gnusi, line 71: `virtual VPackSlice value() const = 0;`?

Dronplane: Some implementation do not need "value" so I wanted to implement "assert only" function just once

gnusi (approved):

PR #14241, opened 2021-05-21
only build actually used attributes of the path

jsteemann:
### Scope & Purpose

Traversal optimization.
The optimizer can now detect for traversals if only certain attributes of the path output variable (e.g. `vertices`, `edges` or `weights`) are accessed later. Only those that are actually accessed later will be included in the result paths that are built up.

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/681

#### Backports:

- [x] No backports required

#### Related Information

- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/681

### Testing & Verification

  - [x] Added new **integration tests** (e.g. in shell_server_aql)


arangod/Aql/OptimizerRules.cpp
```diff
@@ -5843,14 +5843,92 @@ void arangodb::aql::optimizeTraversalsRule(Optimizer* opt,
...
+        // "vertices" or "edges" are in use (or the complete path)
+        std::unordered_set<std::string> attributes;
+        VarSet vars;
+        bool canOptimize = true;
+
+        if (options->usesPrune()) {
+          std::vector<Variable const*> pruneVars;
+          traversal->getPruneVariables(pruneVars);
+          if (std::find(pruneVars.begin(), pruneVars.end(), outVariable) != pruneVars.end()) {
+            // prune uses the path. we don't know if the optimization
```
mchacki, line 5869: I think we are able to analyze the Prune statement as well.
And figure out if / which of the items are used by prune.

arangod/Aql/OptimizerRules.cpp
```diff
@@ -5843,14 +5843,92 @@ void arangodb::aql::optimizeTraversalsRule(Optimizer* opt,
...
+    if (outVariable != nullptr) {
+      if (!n->isVarUsedLater(outVariable)) {
+        // traversal path outVariable not used later
+        if (std::find(pruneVars.begin(), pruneVars.end(), outVariable) ==
+            pruneVars.end()) {
+          options->setProducePaths(/*vertices*/ false, /*edges*/ false, /*weights*/ false);
+          traversal->setPathOutput(nullptr);
+        } else {
+          // we still need to build the path because PRUNE relies on it
+          options->setProducePaths(/*vertices*/ true, /*edges*/ true, /*weights*/ true);
```
mchacki, line 5855: I think the weights here need to be based on the WEIGHTED order

mchacki: ah it's save because the later accessor will apply the ordering

mchacki (commented):
Nice I like this PR :+1:

I have left some inline comments everything else LGTM!

As we discussed offline already tests need to be added still.

Simran-B:
Perhaps add tests for bracket notation (`p["vertices"]`, `NOEVAL(p["edges"])` etc.)?

arangod/Graph/TraverserOptions.cpp
```diff
@@ -509,9 +509,9 @@ void TraverserOptions::toVelocyPack(VPackBuilder& builder) const {
   }
 
   builder.add("produceVertices", VPackValue(_produceVertices));
-  builder.add("producePathsVertices", VPackValue(_producePathsVertices));
-  builder.add("producePathsEdges", VPackValue(_producePathsEdges));
-  builder.add("producePathsWeights", VPackValue(_producePathsWeights));
+  builder.add("producePathsVertices", VPackValue(producePathsVertices()));
+  builder.add("producePathsEdges", VPackValue(producePathsEdges()));
+  builder.add("producePathsWeights", VPackValue(producePathsWeights()));
```
mchacki, line 514: Great! I overlooked this in my first review!

tests/js/server/aql/aql-traversal-paths-optimization.js
```diff
@@ -0,0 +1,216 @@
...
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN p.edges", [ false, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN p.vertices", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN [p.edges, p.vertices]", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN [p.edges, p]", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN [p.vertices, p]", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN [p.edges, p.vertices, p]", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' RETURN p.testi", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' COLLECT pl = LENGTH(p) RETURN pl", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' COLLECT key = v._key INTO g RETURN { key, other: g[*].p.edges }", [ true, true, false ]],
```
mchacki, line 55: Nice this case is good.
Would love to move it to "false, true, false" at one Point. But I agree that is super hard to figure out.

tests/js/server/aql/aql-traversal-paths-optimization.js
```diff
@@ -0,0 +1,216 @@
...
+       
+        // PRUNE
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE v.testi == 'abc' RETURN p", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE v.testi == 'abc' RETURN p.edges", [ false, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE v.testi == 'abc' RETURN p.vertices", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE p.vertices[-1].testi == 'abc' RETURN p.vertices", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE p.vertices[-1].testi == 'abc' RETURN p.edges", [ false, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE p.edges[-1].testi == 'abc' RETURN p.vertices", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' PRUNE p.edges[-1].testi == 'abc' RETURN p.edges", [ false, true, false ]],
+
```
mchacki, line 71: I think we should add the following test-cases as well:

`FILTER p.edges[*].weight ALL >= 3 RETURN p.vertices` => `vertices:true, edges:false` (the edges should be optimized into Traversal and do not need to be produced)

`FILTER p.vertices[1].weight >= 3 RETURN p.edges` => `vertices:false, edges:true` (same as above)

`FILTER p.edges[0].weight >= 3 RETURN p.vertices` => `vertices:true, edges:false` (the edges should be optimized into Traversal and do not need to be produced)

`FILTER p.vertices[*].weight ALL >= 3 RETURN p.edges` => `vertices:false, edges:true` (same as above)


If they do not work as suggested  I am happy to have them with `true, true` for now, and we have another possible optimization here.

tests/js/server/aql/aql-traversal-paths-optimization.js
```diff
@@ -0,0 +1,216 @@
...
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' COLLECT pl = LENGTH(p) RETURN pl", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' COLLECT key = v._key INTO g RETURN { key, other: g[*].p.edges }", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' FILTER LENGTH(p) > 0 RETURN v", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' SORT p.edges RETURN v", [ false, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' SORT p.vertices RETURN v", [ true, false, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' SORT p.edges, p.vertices RETURN v", [ true, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' SORT LENGTH(p.edges) RETURN v", [ false, true, false ]],
+        [ "FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' SORT LENGTH(p.vertices) RETURN v", [ true, false, false ]],
```
mchacki, line 61: For the sake of completeness we should also include "wierd" ways to write edges / vertices:

```
RETURN p.`vertices`
```
```
RETURN p["vertices"]
```

```
RETURN p[CONCAT("vert","ices")]
```

And also make the access not single attribute:
```
RETURN p.vertices[*].name
```

// Maybe not even array access, this will return null btw.
```
RETURN p.vertices.name
```

I do expect our optimizer to handle all of the weird writing methods before hand, just to be on the save side here.

Also something that comes to my mind is subqueries.
```
"FOR v, e, p IN 1..1 OUTBOUND '" + vn + "/test0' GRAPH '" + gn + "' LET sub = (FOR x IN p.vertices RETURN x.name) RETURN sub"
```

tests/js/server/aql/aql-traversal-paths-optimization.js
```diff
@@ -0,0 +1,216 @@
...
+
+      ].forEach((query) => {
+        let nodes = AQL_EXPLAIN(query[0]).plan.nodes;
+        let traversal = nodes.filter((node) => node.type === 'TraversalNode')[0];
+
+        assertEqual("p", traversal.pathOutVariable.name);
+
+        assertEqual(query[1][0], traversal.options.producePathsVertices);
+        assertEqual(query[1][1], traversal.options.producePathsEdges);
+        assertEqual(query[1][2], traversal.options.producePathsWeights);
```
mchacki, line 90: This is good test-suite.

I would like to make sure that the result does not contain the other values.

I see two options here.
1) AFAIR we have "UnitTests" for TraversalBlocks. There we can inject the expected produce setting and ensure that the resulting AQL value only has the asked for Parts. I think that is the easiest way to implement this.
2) We could take a small sample of the Above values, run them and take a look at the "scannedIndex" value in the resulting Cursor getExtra(). for all queries that do not produce vertices the value should be lower as we can avoid reading those documents all together. We cannot assert weight or edges this way.
3) Maybe? AFAIR we had a way to "create" a plan, expose it as JSON, and execute a JSON plan as AQL. We could take a simple `RETURN p` query (the optimization is NOT applied) then we "plan" it. And overwrite the .produceXXX values in the plan. Then we regenerate the query from the Plan => The returned p should only contain what we asked for. I am not 100% sure if this works.

mchacki (approved):
Technically very good and already a very good test coverage.

I have added some more test ideas, but I would be happy to merge this state already.

mchacki (approved):
LGTM

PR #14247, opened 2021-05-25
Move a hotbackup test to enterprise where it belongs

neunhoef:
This moves the hotbackup-deadlock test to the enterprise repository,
where it belongs, since hotbackup is not available in the community
edition. The test as it stands makes no sense in community and even
sometimes failed.

See corresponding PR in the enterprise repository.


jsteemann (approved):
LGTM

PR #14252, opened 2021-05-25
better URL-encoding in UI

jsteemann:
### Scope & Purpose

Improve URL-encoding in the web UI.
This still needs manual testing.

#### Backports:

- [x] No backports required

### Testing & Verification


js/apps/system/_admin/aardvark/APP/aardvark.js
```diff
@@ -294,6 +294,7 @@ authRouter.get('/query/download/:user', function (req, res) {
     res.throw('not found');
   }
 
+  // TODO: properly escape name here
   res.attachment(`queries-${db._name()}-${user.user}.json`);
```
pluma, line 298: Would doing a simple `.replace(/[^a-z0-9]/gi, "_")` or similar be enough? It's not going to handle umlauts and non-Latin characters gracefully but it should prevent bad things from happening.

js/apps/system/_admin/aardvark/APP/frontend/js/collections/arangoQueries.js
```diff
@@ -13,7 +13,7 @@
     initialize: function (models, options) {
       var self = this;
 
-      $.ajax('whoAmI?_=' + Date.now(), {async: true}).done(
+      $.ajax('whoAmI?_=' + encodeURIComponent(Date.now()), {async: true}).done(
```
pluma, line 16: `Date.now()` returns a number. I don't think this change does anything meaningful.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
```diff
@@ -1217,7 +1217,7 @@
 
       if (window.App.naviView && $('#subNavigationBar .breadcrumb').html() !== undefined) {
         $('#subNavigationBar .breadcrumb').html(
-          'Collection: ' + arangoHelper.escapeHtml(this.collectionName.length > 64 ? this.collectionName.substr(0, 64) + "..." : this.collectionName)
+          'Collection: ' + _.escape(this.collectionName.length > 64 ? this.collectionName.substr(0, 64) + "..." : this.collectionName)
```
pluma, line 1220: @ajurdak @hkernbach FYI We would probably want to handle the truncation via CSS to avoid information loss here. CSS truncation wasn't widely supported when this code was initially written I think.

js/apps/system/_admin/aardvark/APP/frontend/js/views/indicesView.js
```diff
@@ -15,7 +15,7 @@
 
       // rerender
       self.interval = window.setInterval(function () {
-        if (window.location.hash.indexOf('cIndices/' + self.collectionName) !== -1 && window.VISIBLE) {
+        if (window.location.hash.indexOf('cIndices/' + encodeURIComponent(self.collectionName)) !== -1 && window.VISIBLE) {
```
pluma, line 18: As this handles URLs where the hash part looks like this, there's probably a link somewhere in a template that needs to be updated alongside this change.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
jsteemann: I would be all for doing the truncation via CSS, but I don't know how to achieve this. Would need someone else to help me out with this.

pluma: I think it's fine to do this as proposed for now and switch to CSS truncation later. I don't know off the top of my head right now either but maybe @ajurdak or @hkernbach could look into this, possibly in a separate PR.

js/apps/system/_admin/aardvark/APP/aardvark.js
jsteemann: Adjusted

js/apps/system/_admin/aardvark/APP/frontend/js/views/indicesView.js
jsteemann: The `cIndices` routes appears only here, so I had hoped we are covered:
```
$ grep -r "cIndices" js/apps/system/_admin/aardvark/APP/frontend/js/
js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js:          route: '#cIndices/' + encodeURIComponent(collectionName)
js/apps/system/_admin/aardvark/APP/frontend/js/views/indicesView.js:        if (window.location.hash.indexOf('cIndices/' + encodeURIComponent(self.collectionName)) !== -1 && window.VISIBLE) {
js/apps/system/_admin/aardvark/APP/frontend/js/routers/router.js:      'cIndices/:colname': 'cIndices',
js/apps/system/_admin/aardvark/APP/frontend/js/routers/router.js:    cIndices: function (colname) {
```

pluma:
```
js/apps/system/_admin/aardvark/APP/frontend/js/routers/router.js:    cIndices: function (colname) {
```
Note that `colname` will now be the escaped name. The function may need to unescape it before handling it.

jsteemann: I don't think so. I just verified it myself that the parameters that the Backbone router passes into the registered functions are already URL-decoded. So I think it is fine.

js/apps/system/_admin/aardvark/APP/frontend/js/templates/collectionsItemView.ejs
```diff
@@ -12,13 +12,8 @@
...
       </div>
     <% } else if (model.get('status') === "loaded") { %>
-      <% if (!model.get('locked')) { %>
-        <div class="corneredBadge <%= model.get('status') %>">
-          <%= model.get('status') %>
-        </div>
-      <% } else { %>
+      <% if (model.get('locked')) { %>
         <div class="corneredBadge inProgress">
-          <%= model.get('status') %>
```
pluma, line 21: Is this right? Won't this just show an empty badge then?

jsteemann: It will display an unlabeled badge at least.
I don't know how it would look like, and I don't know if we can actually get into this state after reducing the number of collection statuses.
I can try to force this state manually just to see how it looks like, and then I will either remove the badge or leave it there if it still looks differently (the `inProgress` class may still modify the thing visually, but I don't know).

pluma: After checking locally, it does look pretty ugly as the height expands weirdly when the div has no content, so I'd vote for simply not showing a badge at all or keeping the status in there if we still want to show the badge.

pluma: Here's what the `loaded` one looks like without content for comparison:
image

jsteemann: I came to the same conclusion just now. PR is adjusted accordingly.

pluma (approved):
LGTM.

Not sure when a collection would be "locked and loaded" (hah) and if we maybe want to keep the badge for that state for some reason but if you don't know either, I think it's safe to assume there doesn't need to be a badge for it. I'd like @hkernbach to give a +1 on that decision though since he's probably more familiar with this screen.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
hkernbach: Yes, I wouldn't do this in this PR. Separate PR is fine.

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
```diff
@@ -35,7 +35,7 @@
       $.ajax({
         type: 'GET',
         cache: false,
-        url: arangoHelper.databaseUrl('/_api/collection/' + this.get('id') + '/shards?details=true'),
+        url: arangoHelper.databaseUrl('/_api/collection/' + encodeURIComponent(this.get('id')) + '/shards?details=true'),
```
hkernbach, line 38: Just some note: You could use here this.escape('id'), see Backbone Documentation:
- https://backbonejs.org/#Model-escape

```
Similar to get, but returns the HTML-escaped version of a model's attribute. If you're interpolating data from the model into HTML, using escape to retrieve attributes will prevent XSS attacks.
```

hkernbach: But feel free to ignore. This will do the same thing. Your choice.

js/apps/system/_admin/aardvark/APP/frontend/js/routers/router.js
```diff
@@ -28,7 +28,6 @@
       'cSchema/:colname': 'cSchema',
       'cInfo/:colname': 'cInfo',
       'collection/:colid/:docid': 'document',
-      'shell': 'shell',
```
hkernbach, line 31: I do miss the shell :P

hkernbach (approved):
Very good and overdue PR. Thank you very much. Just one minor note where you simply just use the embedded model escape method. Feel free to use the method instead, but also feel free to ignore.

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
jsteemann: We are not HTML-escaping here, but URL-escaping. This is slightly different.

PR #14264, opened 2021-05-28
Soft coordinator shutdown

neunhoef:
This adds the soft shutdown feature for coordinators for the devel
branch.

Here is the design document:

intrusive for clients.

This seems to be complete from my perspective, but has not yet seen any review. Expect changes. Tests seem to be ok.


arangod/GeneralServer/CommTask.cpp
```diff
@@ -528,7 +528,22 @@ bool CommTask::handleRequestAsync(std::shared_ptr<RestHandler> handler,
...
+      // out of memory situation, so we better handle this anyway.
+      jobManager.initAsyncJob(handler);
+    } catch(arangodb::basics::Exception const& exc) {
+      LOG_TOPIC("fee33", INFO, Logger::STARTUP)
+        << "Async job rejected due to soft shutdown, exception:"
+        << exc.what();
+      return false;
+    } catch(std::exception const& exc) {
+      LOG_TOPIC("fee34", INFO, Logger::STARTUP)
+        << "Async job rejected due to out of memory, exception:"
```
jsteemann, line 543:
```suggestion
        << "Async job rejected due to out of memory, exception: "
```

arangod/Pregel/PregelFeature.cpp
```diff
@@ -442,3 +443,16 @@ void PregelFeature::handleWorkerRequest(TRI_vocbase_t& vocbase,
     w->aqlResult(outBuilder, withId);
   }
 }
+
+uint64_t PregelFeature::numberOfActiveConductors() const {
+  MUTEX_LOCKER(guard, _mutex);
+  uint64_t nr{0};
+  for (auto const& p : _conductors) {
+    std::shared_ptr<Conductor> c = p.second.second;
```
jsteemann, line 451: No need to copy the shared_ptr here:
```suggestion
    std::shared_ptr<Conductor> const& c = p.second.second;
```

arangod/Pregel/PregelFeature.cpp
```diff
@@ -442,3 +443,16 @@ void PregelFeature::handleWorkerRequest(TRI_vocbase_t& vocbase,
...
+    if (c->_state == ExecutionState::RUNNING ||
+        c->_state == ExecutionState::STORING) {
```
jsteemann, line 453: I think we also need to count all conductors in state `DEFAULT`. `DEFAULT` is the initial conductor state and will be still set when the conductor is inserted into the PregelFeature's map of conductors.
```suggestion
    if (c->_state == ExecutionState::DEFAULT ||
        c->_state == ExecutionState::RUNNING ||
        c->_state == ExecutionState::STORING) {
```

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -222,6 +222,11 @@ RestStatus RestCursorHandler::registerQueryOrCursor(VPackSlice const& slice) {
     CursorRepository* cursors = _vocbase.cursorRepository();
     TRI_ASSERT(cursors != nullptr);
     _cursor = cursors->createQueryStream(std::move(query), batchSize, ttl);
+    if (_cursor == nullptr) {   // soft shutdown, no longer allowed
+      generateError(Result(TRI_ERROR_SHUTTING_DOWN,
+                           "coordinator soft shutdown, new cursors blocked"));
+      return RestStatus::DONE;
+    }
```
jsteemann, line 229: This creates some coupling between the code here and the code in `createQueryStream`, which I think is unnecessary.
I would prefer if `createQueryStream` simply threw an exception in case of soft shutdown, and return a valid pointer in all other cases.
In case it throws, it can already throw the correct error code in the exception, so we wouldn't need to know anything about soft shutdown here.
All exceptions are handled at least on the outermost layer in all RestHandlers automatically, so we can just let any exception from the CursorRepository propagate from here.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -368,6 +373,11 @@ RestStatus RestCursorHandler::handleQueryResult() {
     TRI_ASSERT(_queryResult.data.get() != nullptr);
     // steal the query result, cursor will take over the ownership
     _cursor = cursors->createFromQueryResult(std::move(_queryResult), batchSize, ttl, count);
+    if (_cursor == nullptr) {
+      generateError(Result(TRI_ERROR_SHUTTING_DOWN,
+                           "coordinator soft shutdown, new cursors blocked"));
+      return RestStatus::DONE;
+    }
```
jsteemann, line 380: Same as directly above.

arangod/RestHandler/RestShutdownHandler.cpp
```diff
@@ -68,6 +70,16 @@ RestStatus RestShutdownHandler::execute() {
     }
   }
 
+  auto const& softShutdownFeature{server().getFeature<SoftShutdownFeature>()};
+  auto& softShutdownTracker{softShutdownFeature.softShutdownTracker()};
+
+  if (_request->requestType() == rest::RequestType::GET) {
+    VPackBuilder builder;
+    softShutdownTracker.toVelocyPack(builder, softShutdownTracker.getStatus());
```
jsteemann, line 78: Calling a method on the soft shutdown tracker and passing it its own internals looked a bit strange.
I figured that `toVelocyPack` is static, so it's how it is. But I would prefer a non-static method `toVelocyPack` on the tracker that dumps it own status into a result. (it may just fetch its own status and call the static method with 2 parameters).

arangod/RestHandler/RestShutdownHandler.cpp
```diff
@@ -68,6 +70,16 @@ RestStatus RestShutdownHandler::execute() {
     }
   }
 
+  auto const& softShutdownFeature{server().getFeature<SoftShutdownFeature>()};
+  auto& softShutdownTracker{softShutdownFeature.softShutdownTracker()};
+
+  if (_request->requestType() == rest::RequestType::GET) {
```
jsteemann, line 76: Here we are not checking if we are a coordinator, but soft shutdown is only available there.
So on non-coordinators, the GET API will work but probably not return the expected results (I haven't checked if this is the case).
Should we be consistent here and return an error if the GET API is called on non-coordinators?

arangod/RestHandler/RestShutdownHandler.cpp
```diff
@@ -91,6 +103,21 @@ RestStatus RestShutdownHandler::execute() {
     clusterFeature.setUnregisterOnShutdown(true);
   }
 
+  bool soft = false;
+  bool softFound = false;
+  std::string const& softString = _request->value("soft", softFound);
+  if (softFound && softString.compare("true") == 0) {
+    soft = true;
+  }
```
jsteemann, line 111:
```suggestion
  bool soft = _request->parsedValue("soft", false);
```

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+      return;   // already stopping, do nothing, and in particular
+                // let's not schedule ourselves again!
+    }
+    if (!this->check()) {   // Initiates shutdown if counters are 0
+      // Rearm ourselves:
+      if (!::queueShutdownChecker(this->_workItemMutex, this->_workItem,
+                                  this->_checkFunc)) {
+        // If this does not work, shut down right away:
+        this->initiateActualShutdown();
+      }
```
jsteemann, line 104: This looked unintuitive to me, although after looking up the functions, I think it works.
My stylistic issues with it: `check()` sounds like it would be side-effects free. It is also declared `const`. But it can trigger the shutdown, which was unexpected to me.
I would prefer if `check()` was really side-effects-free and that the caller would trigger any require side-effects instead.

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+      << "Goal reached for soft shutdown, all ongoing tasks are terminated"
+         ", will now trigger the actual shutdown...";
+  initiateActualShutdown();
+  return true;
+}
+
+void SoftShutdownTracker::initiateActualShutdown() const {
+  Scheduler* scheduler = SchedulerFeature::SCHEDULER;
+  auto self = shared_from_this();
+  bool queued = scheduler->queue(RequestLane::CLUSTER_INTERNAL, [self] {
```
jsteemann, line 162:
```suggestion
  auto self = shared_from_this();
  bool queued = scheduler->queue(RequestLane::CLUSTER_INTERNAL, [self = shared_from_this()] {
```

arangod/RestServer/SoftShutdownFeature.h
```diff
@@ -0,0 +1,114 @@
...
+  struct Status {
+    uint64_t AQLcursors;
+    uint64_t transactions;
+    uint64_t pendingJobs;
+    uint64_t doneJobs;
+    uint64_t pregelConductors;
+    uint64_t lowPrioOngoingRequests;
+    uint64_t lowPrioQueuedRequests;
+    bool allClear;
+    bool softShutdownOngoing;
```
jsteemann, line 86: Would prefer if the counter values would be initialized, `softShutdownOngoing` would be passed on creation, and `allClear()` were a method that analyzes all counters.
That way it cannot be used wrong:
```suggestion
    uint64_t AQLcursors{0};
    uint64_t transactions{0};
    uint64_t pendingJobs{0};
    uint64_t doneJobs{0};
    uint64_t pregelConductors{0};
    uint64_t lowPrioOngoingRequests{0};
    uint64_t lowPrioQueuedRequests{0};

    bool const softShutdownOngoing;
    
    explicit(bool softShutdownOngoing) 
       : softShutdownOngoing(softShutdownOngoing) {}
       
    bool allClear() const noexcept {
      return AQLcursors == 0 &&
             transactions == 0 &&
             pendingJobs == 0 &&
             doneJobs == 0 &&
             lowPrioOngoingRequests == 0 &&
             lowPrioQueuedRequests == 0 &&
             pregelConductors == 0;
    }
```

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+void SoftShutdownFeature::beginShutdown() {
+  _softShutdownTracker->cancelChecker();
+}
+
+void SoftShutdownTracker::cancelChecker() {
+  if (_softShutdownOngoing.load(std::memory_order_relaxed)) {
+    // This is called when an actual shutdown happens. We then want to
+    // delete the WorkItem of the soft shutdown checker, such that the
+    // Scheduler does not have any cron jobs any more:
+    _workItem.reset();
```
jsteemann, line 86:
```suggestion
    std::lock_guard<std::mutex> guard(_workItemMutex);
    _workItem.reset();
```

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+      << "Initiating soft shutdown...";
+
+  // Tell asynchronous job manager:
+  auto& generalServerFeature = _server.getFeature<GeneralServerFeature>();
+  auto& jobManager = generalServerFeature.jobManager();
+  jobManager.initiateSoftShutdown();
+
+  // Tell Pregel subsystem:
+  auto& pregelFeature = _server.getFeature<pregel::PregelFeature>();
+  pregelFeature.initiateSoftShutdown();
```
jsteemann, line 128: I would prefer to have a method `initiateSoftShutdown()` in the ApplicationServer, which would call a new virtual method `initiazeSoftShutdown()` on each ApplicationFeature.
The default implementation of this method would do nothing of course, but any feature could override it is with something useful.

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+  VPackObjectBuilder guard(&builder);
+  builder.add("softShutdownOngoing", VPackValue(status.softShutdownOngoing));
+  builder.add("AQLcursors", VPackValue(status.AQLcursors));
+  builder.add("transactions", VPackValue(status.transactions));
+  builder.add("pendingJobs", VPackValue(status.pendingJobs));
+  builder.add("doneJobs", VPackValue(status.doneJobs));
+  builder.add("pregelConductors", VPackValue(status.pregelConductors));
+  builder.add("lowPrioOngoingRequests", VPackValue(status.lowPrioOngoingRequests));
+  builder.add("lowPrioQueuedRequests", VPackValue(status.lowPrioQueuedRequests));
+  builder.add("allClear", VPackValue(status.allClear));
```
jsteemann, line 185: Ideally, this method would call a (new) `toVelocyPack(VPackBuilder&)` method on the SoftShutdownTracker::Status object.

arangod/RestServer/SoftShutdownFeature.cpp
```diff
@@ -0,0 +1,234 @@
...
+}
+
+SoftShutdownTracker::Status SoftShutdownTracker::getStatus() const {
+  Status status;
+  status.softShutdownOngoing = _softShutdownOngoing.load(std::memory_order_relaxed);
```
jsteemann, line 190: As mentoned below, I would make the `Status` object a bit more intelligent, so it is harder to use it wrong. For example, I would passing `softShutdownOngoing` as a mandatory ctor parameter, so it can't be forgotten. All other counters in the Status object I would initialize to 0 on object creation, and then only add to them.

arangod/V8Server/v8-voccursor.cpp
```diff
@@ -102,6 +102,10 @@ static void JS_CreateCursor(v8::FunctionCallbackInfo<v8::Value> const& args) {
     arangodb::Cursor* cursor =
         cursors->createFromQueryResult(std::move(result),
                                        static_cast<size_t>(batchSize), ttl, true);
+    if (cursor == nullptr) {
+      TRI_V8_THROW_EXCEPTION_FULL(TRI_ERROR_SHUTTING_DOWN,
+           "coordinator soft shutdown, new cursors blocked");
+    }
```
jsteemann, line 108: As mentioned somewhere above, I would remove the check here and make `CursorRepository::createFromQueryResult()` throw an appropriate exception when we are in soft shutdown.

arangod/V8Server/v8-voccursor.cpp
```diff
@@ -339,6 +343,10 @@ struct V8Cursor final {
     
     // specify ID 0 so it uses the external V8 context
     auto cc = cursors->createQueryStream(std::move(q), batchSize, ttl);
+    if (cc == nullptr) {   // soft shutdown, no longer allowed
+      TRI_V8_THROW_EXCEPTION_FULL(TRI_ERROR_SHUTTING_DOWN,
+           "coordinator soft shutdown, new cursors blocked");
+    }
```
jsteemann, line 349: As mentioned somewhere above, I would remove the check here and make `CursorRepository::createQueryStream()` throw an appropriate exception when we are in soft shutdown.

tests/js/client/restart/test-restart-cluster.js
```diff
@@ -86,6 +86,27 @@ function testSuite() {
...
+      let coordinator = coordinators[0];
+      let instanceInfo = global.instanceInfo;
+      let newInstanceInfo = {
+        arangods: [ coordinator ],
+        endpoint: instanceInfo.endpoint,
+      };
+      let options = global.testOptions;
+      let shutdownStatus = pu.shutdownInstance(newInstanceInfo, options, false); 
+      coordinator.pid = null;
+      console.warn("Restarting coordinator without authentication as cleanup...", coordinator);
```
jsteemann, line 101: I found the message hard to understand, and had to think about the "without authentication as cleanup" for a long time :sweat_smile: 
```suggestion
      console.warn("Cleaning up and restarting coordinator without authentication...", coordinator);
```

jsteemann:
General remarks:
* I would make `initateSoftShutdown()` a new method in ApplicationServer, which could call a new virtual method `initiateSoftShutdown()` on all its ApplicationFeatures. The default implementation would do nothing.
* the ApplicationServer should keep track of the boolean flag for whether a soft shutdown was requested or not. 
* there should be a new method to ask the ApplicationServer whether a soft shutdown was requested. 
* this new method should be called from ApplicationFeatures and other places instead of passing a pointer to the atomic around. The assumption is that everything either already has access to the ApplicationServer, or could get access to it pretty easily.
* IMO we should throw "shutting down" exceptions in case an operation is requested from a feature during soft shutdown, but the feature cannot execute it. This IMO is preferrable to adjusting call sites and handling nullptrs/errors there.

arangod/GeneralServer/AsyncJobManager.cpp
```diff
@@ -257,11 +260,36 @@ std::vector<AsyncJobResult::IdType> AsyncJobManager::byStatus(AsyncJobResult::St
...
+}
+
 ////////////////////////////////////////////////////////////////////////////////
-/// @brief initializes an async job
+/// @brief initializes an async job, throws if soft shutdown is already
+/// ongoing.
 ////////////////////////////////////////////////////////////////////////////////
 
 void AsyncJobManager::initAsyncJob(std::shared_ptr<RestHandler> handler) {
+  if (_softShutdownOngoing.load(std::memory_order_relaxed)) {
```
jsteemann, line 289: There is the (rare) possibility for the following sequence of events:

- thread #1 calls `initAsyncJob()` before a soft shutdown is requested, and gets past the initial check for `_softShutdownOngoing` and then goes to sleep.
- thread #2 initiates a soft shutdown, so that `_softShutdownOngoing` now is true.
- thread #2 or any other thread than #1 checks the number of pending operations, which is still zero. this may lead to the conclusion that a soft shutdown is alright, because there is nothing left to execute.
- thread #1 wakes up and inserts the job into the `_jobs` map.

We could prevent this from happening by performing the check on `_softShutdownOngoing` in this method under the write lock.

neunhoef: This is a good idea, will do.

arangod/GeneralServer/CommTask.cpp
```diff
@@ -528,7 +528,22 @@ bool CommTask::handleRequestAsync(std::shared_ptr<RestHandler> handler,
...
     auto& jobManager = _server.server().getFeature<GeneralServerFeature>().jobManager();
-    jobManager.initAsyncJob(handler);
+    try {
+      // This will throw if a soft shutdown is already going on on a
+      // coordinator. But this can also throw if we have an
+      // out of memory situation, so we better handle this anyway.
+      jobManager.initAsyncJob(handler);
+    } catch(arangodb::basics::Exception const& exc) {
+      LOG_TOPIC("fee33", INFO, Logger::STARTUP)
+        << "Async job rejected due to soft shutdown, exception:"
```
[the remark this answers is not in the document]
neunhoef: Fine, will do.

arangod/GeneralServer/CommTask.cpp
neunhoef: Agreed.

arangod/Pregel/PregelFeature.cpp
neunhoef: Agreed.

arangod/Pregel/PregelFeature.cpp
neunhoef: Good catch!

arangod/RestHandler/RestCursorHandler.cpp
neunhoef: Good point, will do.

arangod/RestHandler/RestCursorHandler.cpp
neunhoef: Agreed, will do.

arangod/RestHandler/RestShutdownHandler.cpp
neunhoef: Yes, let's do this, will do.

arangod/RestHandler/RestShutdownHandler.cpp
neunhoef: Fine, will add that, although we are not passing its internals, but a "public" result. Anyway, will do.

arangod/RestHandler/RestShutdownHandler.cpp
neunhoef: Nice.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Good catch!

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Let's rename the method to `checkAndShutdownIfAllClear`, will do.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: I will do that.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Will change.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: OK, will do.

arangod/RestServer/SoftShutdownFeature.h
neunhoef: Good idea.

arangod/V8Server/v8-voccursor.cpp
neunhoef: Will do.

arangod/V8Server/v8-voccursor.cpp
neunhoef: I vaguely remember that at some stage we wanted to get away from exceptions for error handling and rely more on return values, but my memory might not serve me well. Anyway, I do not really care...

arangod/GeneralServer/AsyncJobManager.cpp
neunhoef: Done.

arangod/GeneralServer/CommTask.cpp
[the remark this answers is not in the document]
neunhoef: Done.

arangod/RestHandler/RestCursorHandler.cpp
neunhoef: Done.

arangod/RestHandler/RestCursorHandler.cpp
neunhoef: Done.

arangod/RestHandler/RestShutdownHandler.cpp
neunhoef: Done.

arangod/RestHandler/RestShutdownHandler.cpp
neunhoef: Done.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Done.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Done.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Done.

arangod/RestServer/SoftShutdownFeature.cpp
neunhoef: Done.

arangod/V8Server/v8-voccursor.cpp
neunhoef: Done.

arangod/V8Server/v8-voccursor.cpp
neunhoef: Done.

arangod/V8Server/v8-voccursor.cpp
```diff
@@ -115,6 +110,9 @@ static void JS_CreateCursor(v8::FunctionCallbackInfo<v8::Value> const& args) {
 
     auto result = TRI_V8UInt64String<TRI_voc_tick_t>(isolate, id);
     TRI_V8_RETURN(result);
+  } catch (arangodb::basics::Exception const& exc) {
+    // Probably a coordinator soft shutdown:
+    TRI_V8_THROW_EXCEPTION_FULL(exc.code(), exc.message());
   } catch (...) {
     TRI_V8_THROW_EXCEPTION_MEMORY();
```
jsteemann, line 117: I think we can get rid of the entire catch blocks sequence here, because the function ends with a `TRI_V8_TRY_CATCH_END`, which will expand to an exception handler anyway.

arangod/V8Server/v8-voccursor.cpp
```diff
@@ -342,10 +340,12 @@ struct V8Cursor final {
...
-    if (cc == nullptr) {   // soft shutdown, no longer allowed
-      TRI_V8_THROW_EXCEPTION_FULL(TRI_ERROR_SHUTTING_DOWN,
-           "coordinator soft shutdown, new cursors blocked");
+    Cursor* cc;
+    try {
+      cc = cursors->createQueryStream(std::move(q), batchSize, ttl);
+    } catch(arangodb::basics::Exception const& exc) {
+      // probably a soft shutdown, no longer allowed
+      TRI_V8_THROW_EXCEPTION_FULL(exc.code(), exc.message());
     }
```
jsteemann: same here

arangod/V8Server/v8-voccursor.cpp
neunhoef: OK. Good.

neunhoef: I think here this does not work, since otherwise the `catch (...)` catches it and reports out of memory. I am inclined to leave this catch block in and only remove the one further down in the same file.

jsteemann (approved):
LGTM!

PR #14267, opened 2021-05-28
Parallel multi-shard INSERTs

jsteemann:
### Scope & Purpose

Partly addresses https://arangodb.atlassian.net/browse/BTS-439.

Allow an INSERT operation for a multi-shard collection to execute in
parallel on different DB servers.
Previously INSERTs for a multi-shard collection were executed
sequentially on each DB servers. This could lead to slow execution time,
a large memory usage (DISTRIBUTE has to buffer data for all other DB
servers in memory until the DB servers asks for them) and also query
timeouts.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/Aql/ExecutionBlockImpl.cpp
+++ b/arangod/Aql/ExecutionBlockImpl.cpp
@@ -383,7 +383,8 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
     // store only the first failure we got
     _firstFailure = {TRI_ERROR_INTERNAL, ex.what()};
     LOG_QUERY("2bbd5", DEBUG) << printBlockInfo() << " local statemachine failed with exception: " << ex.what();
-    throw;
+    // Rewire the error, to be consistent with potentially next caller.
+    THROW_ARANGO_EXCEPTION(_firstFailure);
   }
 }
 
--- a/arangod/Aql/ExecutionBlockImpl.cpp
+++ b/arangod/Aql/ExecutionBlockImpl.cpp
@@ -373,18 +373,16 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
     traceExecuteEnd(res);
     return res;
   } catch (basics::Exception const& ex) {
-    if (_firstFailure.ok()) {
-      // store only the first failure we got
-      _firstFailure = {ex.code(), ex.what()};
-      LOG_QUERY("7289a", DEBUG) << printBlockInfo() << " local statemachine failed with exception: " << ex.what();
-    }
+    TRI_ASSERT(_firstFailure.ok());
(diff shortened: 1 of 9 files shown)


neunhoef (approved):
LGTM

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -338,28 +338,55 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
...
+    if (block != nullptr) {
+      block->validateShadowRowConsistency();
+    }
 #endif
-  traceExecuteEnd(res);
-  return res;
+    traceExecuteEnd(res);
+    return res;
+  } catch (basics::Exception const& ex) {
+    if (_firstFailure.ok()) {
```
mpoeter, line 376: Actually we should never get here if `!_firstFailure.ok()`. How about an assertion instead of the if?

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -338,28 +338,55 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
...
+      // store only the first failure we got
+      _firstFailure = {ex.code(), ex.what()};
+      LOG_QUERY("7289a", DEBUG) << printBlockInfo() << " local statemachine failed with exception: " << ex.what();
+    }
+    throw;
+  } catch (std::exception const& ex) {
+    if (_firstFailure.ok()) {
```
mpoeter, line 383: Same as above.

arangod/Aql/OptimizerRules.cpp
```diff
@@ -7423,11 +7427,38 @@ struct ParallelizableFinder final
...
       return true;  // true to abort the whole walking process
     }
+    if (node->getType() == ExecutionNode::DISTRIBUTE) {
+      if (_seenDistribute) {
+        // if we find 2 DISTRIBUTE nodes in the plan, or a DISTRIBUTE
+        // at an unexpected place, we give up
+        _isParallelizable = false;
+        return true;
+      }
+      // node that we have seen a DISTRIBUTE node.
```
mpoeter, line 7441:
```suggestion
      // note that we have seen a DISTRIBUTE node.
```

mpoeter (approved):
A few suggestions, otherwise LGTM!

arangod/Aql/ExecutionBlockImpl.cpp
jsteemann: done

arangod/Aql/ExecutionBlockImpl.cpp
jsteemann: done

PR #14273, opened 2021-05-31
remember exception in ExecutionBlock

jsteemann:
### Scope & Purpose

Remember exception in ExecutionBlock and return it on next invocation.
normally any exception returned by an ExecutionBlock is directly handled
by the block's caller.
However, if the same block is asked semi-concurrently (only protected by
the RestAqlHandler's lease), then there can still be other requests in
flight that wait to enter the block. if an exception escapes the block,
it may be left in an invalid state, so we have to prevent other threads
from entering it afterwards and continue with that invalid state.

This is fixed in devel via https://github.com/arangodb/arangodb/pull/14267

#### Backports:

- [x] No backports required

### Testing & Verification


mchacki (approved):
LGTM.
Thanks for the improvement!

PR #14283, opened 2021-05-31
Improve internal AQL and transaction timeouts

jsteemann:
### Scope & Purpose

Increase internal AQL query and transaction timeout on DB servers from 3 to 5 minutes.
Previously, queries and transactions on DB servers could expire quickly, which led to spurious "query ID not found" or "transaction ID not found" errors on DB servers for multi-server queries/transactions with unbalanced access patterns for the different participating DB servers.

The timeouts on coordinators remain unchanged, so any queries/transactions that are abandoned will be aborted there, which will also be propagated to DB servers. In addition, if a particpating server in an AQL query becomes unavailable, the coordinator is now notified of that and will terminate the query more eagerly.

This PR addresses the timeouts issue described in https://arangodb.atlassian.net/browse/BTS-439

This is a change with potentially large side-effects, so there are no plans to backport it to earlier versions.

This PR is still missing useful tests.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
```diff
@@ -408,7 +427,6 @@ Result EngineInfoContainerDBServerServerBased::buildEngines(
   options.database = _query.vocbase().name();
   options.timeout = network::Timeout(SETUP_TIMEOUT);
   options.skipScheduler = true;  // hack to speed up future.get()
-  options.param("ttl", std::to_string(_query.queryOptions().ttl));
```
jsteemann, line 411: note: this request parameter was set, but not read by RestAqlHandler.

arangod/Aql/ExecutionEngine.cpp
```diff
@@ -490,6 +494,45 @@ struct DistributedQueryInstanciator final
...
+      ClusterInfo& ci = _query.vocbase().server().getFeature<ClusterFeature>().clusterInfo();
+      engine->rebootTrackers().reserve(srvrQryId.size());
+      for (auto const& srvr : srvrQryId) {
+        std::string comment = std::string("AQL query from coordinator ") + ServerState::instance()->getId();
+        // intentional copy, because we may modify the value later
+        auto server = srvr.first;
+        if (server.compare(0, 7, "server:") == 0) {
+          server = server.substr(7);
+        }
+        auto rebootId = ci.rebootTracker().rebootId(server);
```
goedderz, line 515: @jsteemann I might be missing something, but I think taking the RebootId this way is subject to a race. It must either be taken from the answer of the server when creating the snippet, or the expected RebootId (i.e. the one we're using here) must be sent with when creating the snippet for the respective DBServer to check and decline the request if it differs.

Thanks for the PR btw, this is great progress for AQL.

jsteemann: @goedderz: I made this change in the meantime.
If you want to, please re-review. Thanks!

arangod/Cluster/RebootTracker.h
```diff
@@ -74,11 +75,6 @@ class RebootTracker {
 
   void updateServerState(std::unordered_map<ServerID, RebootId> const& state);
 
-
-  /// @brief determine the current reboot id for a server.
-  /// will throw if no reboot id is known for the server
-  RebootId rebootId(ServerID const& server) const;
```
goedderz: :+1:

goedderz (approved):

PR #14294, opened 2021-06-01
fixed BTS-398

jsteemann:
### Scope & Purpose

Fixed issue https://arangodb.atlassian.net/browse/BTS-398:
Cannot force index hint for primary index if FILTER has multiple OR conditions that require different indexes.

#### Backports:

### Testing & Verification


tests/js/server/aql/aql-index-hints.js
```diff
@@ -89,6 +85,61 @@ function ahuacatlSkiplistOverlappingTestSuite () {
...
     },
+    
+    testPrimary : function () {
+      [ 
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: false} FILTER doc._key == 'test' RETURN doc`,
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: true} FILTER doc._key == 'test' RETURN doc`,
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: false} FILTER doc._key == 'test' && doc.testi == 99 RETURN doc`,
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: true} FILTER doc._key == 'test' && doc.testi == 99 RETURN doc`,
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: false} FILTER doc.testi == 99 && doc._key == 'test' RETURN doc`,
+        `FOR doc IN ${cn} OPTIONS {indexHint: 'primary', forceIndexHint: true} FILTER doc.testi == 99 && doc._key == 'test' RETURN doc`,
```
mpoeter, line 96: Shouldn't we use `doc.a` instead of `doc.testi`? As far as I can see there is no index on `testi`, so there is no reason for the optimizer to pick an other index than the primary, or am I missing something?

jsteemann: The idea of this test is to check if index hints cause problems if the hint can only be used for a part of the filter condition.

For example, the filter condition here is `doc.testi == 99 && doc._key == 'test'`. The index hint is `primary`. Only the `doc._key == 'test'` subcondition can make use of the index hint here. The other subcondition `doc.testi == 99` cannot (not only because there is no index on it, but also because the index hint is "primary").

That was basically the whole point of the issue: what happens if there are multiple subconditions, of which only some can make use of a forced index hint. The case in question was using OR-combined conditions though, e.g. `doc.testi == 99 || doc._key == 'test'`.

mpoeter: Got it.

mpoeter (approved):

PR #14300, opened 2021-06-01
Metrics names in snippets fix

neunhoef:
This adds a test to the utils/generateAllMetricsDocumentation.py script
to check if the value in the name attribute of a documentation snippet
matches the file name.

Also corrects a few mistakes which went unnoticed.

Also fixes the averaging time for histogram metrics from 60s to 3m (which is necessary for k8s deployments).

This needs to be backported to 3.8, but does not need automated tests.


fceller (commented):
LGTM

PR #14305, opened 2021-06-01
make test more robust

jsteemann:
### Scope & Purpose

Make assertion about generated string size less strict in a test that relies on random values being inserted.
This is a test-only bugfix, so it does not have a CHANGELOG entry.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtest*.


mchacki (approved):
LGTM

PR #14315, opened 2021-06-02
[3.8] Explain indexes of SHORTEST_PATH, K_SHORTEST_PATHS and K_PATHS

goedderz:
Backport of #14298

http://jenkins01.arangodb.biz:8080/view/PR/job/arangodb-matrix-pr/15734/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.8.0 (XXXX-XX-XX)
 -------------------
 
+* Fix BTS-430: Added missing explain output about indexes for SHORTEST_PATH,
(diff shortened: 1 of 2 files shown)


hkernbach (approved):

KVS85:
Tests blue.

mchacki:
Just FYI:
This issue is only in the rendering of the explain, which is inherently hard to test. All information was available in the correct places

We could allow to test those features if we somehow split the large "explain" method into smaller Parts.
Otherwise we would need super complex and fragile string parsing to make sure everything is at the right place.

PR #14330, opened 2021-06-04
Soft coordinator shutdown

neunhoef:
This adds the soft shutdown feature for coordinators for the 3.8
branch (backport from devel).

Here is the design document:

intrusive for clients.

**To be merged for 3.8.1**, together with https://github.com/arangodb/arangodb/pull/14435


CHANGELOG
```diff
@@ -1,3 +1,18 @@
+v3.8.1 (XXXX-XX-XX)
```
jsteemann, line 1:
```suggestion
v3.8.1 (XXXX-XX-XX)
-------------------
```

arangod/RestServer/SoftShutdownFeature.h
```diff
@@ -0,0 +1,132 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Max Neunhoeffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
```
jsteemann, line 23: We have moved to `#pragma once` only in devel.
I think we should use an old-style include guard in 3.8 and before.

arangod/RestServer/SoftShutdownFeature.h
```diff
@@ -0,0 +1,132 @@
...
+
+#include "ApplicationFeatures/ApplicationFeature.h"
+#include "Basics/debugging.h"
+#include "Scheduler/Scheduler.h"
+#include "RestServer/Metrics.h"
```
jsteemann, line 28: I think `RestServer/Metrics.h` is not needed by this file, so we can remove the include.
```suggestion
```

arangod/RestServer/SoftShutdownFeature.h
neunhoef: Done.

PR #14331, opened 2021-06-04
Soft coordinator shutdown

neunhoef:
This adds the soft shutdown feature for coordinators for the 3.7
branch. This is a backport from devel.

Here is the design document:

intrusive for clients.

This is an unfinished. Since the soft coordinator shutdown relies on the features of >= 3.8 of tracking ongoing low prio queue work, which has not been backported to 3.7, further work is necessary here. It does not even compile.


jsteemann (approved):
LGTM!

PR #14339, opened 2021-06-07
Allow active failover failover

neunhoef:
This is a fix for too long retry behaviour in initial sync to allow an active failover setup to actually fail over quickly.

The basic idea is that we have identified a loop in initial sync on the follower which runs up to 5 minutes even though the leader does not respond any more. This prevents a failover to this follower. The reason why this loop does not terminate is that the HeartbeatThread has aborted the globalReplicationApplier, but not its children DatabaseReplicationAppliers in each database. So they do not know that they should abort. Fortunately, the fix is easy: The children appliers simply ask their parent when asked if they are aborted, too.

This is covered by the release-automation upgrade tests. A local test is difficult to achieve. One could in principle reproduce the problem with the following steps:

 1. Start an active failover setup.
 2. Put some data in.
 3. Stop the follower.
 4. Restart the follower outside the active failover setup.
 5. Erase its data, recreate the collection, put some wrong data in. This will trigger a full resync.
 6. Stop the follower.
 7. Activate a failure point in the leader to crash when asked for initial sync in exactly the right moment.
 8. Restart the follower as active failover.
 9. Observe that it fails over after some 20s to the follower and makes it into a leader.

Since we have another test which already tests this behaviour, it is not worth to create this test, in particular since it would perform a slightly illegal usage pattern.


dothebart (commented):
LGTM.

CHANGELOG
```diff
@@ -1,6 +1,9 @@
 v3.7.12 (XXXX-XX-XX)
 --------------------
 
+* Fixed a problem with active failover, where a failover could take 5 mins
+  because the follower was caught in a bad state during replication.
```
dothebart, line 5: pls add BTS-425

neunhoef: Done.

jsteemann (approved):

dothebart (approved):

PR #14347, opened 2021-06-08
make iterator usage safer after intermediate commits

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/14346

When a transaction triggers an intermediate commit, this may affect rocksdb::Iterator objects used by the transaction. These may become silently invalidated if they point to nodes in the committed transaction's `WriteBatchWithIndex`'s index (i.e. the SkipList). 
This change tries to rebuild any such iterators after an intermediate commit, by tracking the last key an iterator was pointing at when leaving iterator-related code.
When re-entering iterator code, there is a check if an intermediate commit has happened in the meantime, and if so, the iterator will be rebuilt from scratch and will seek back to the last tracked key. This seems to be the only safe way to make sure the previous iterator does not point to any data that is now invalid after the previous transaction committed.

For read-only transactions, this behavior is turned off, i.e. there should be no overhead. For write-transactions, there will be (hopefully) neglible overhead, because the last seen key is copied to already allocated memory only when iterator-related code is left. Tracking keys will only happen for every 1,000 keys or so.
The PR also removes superfluous out-of-bounds checks for read-only transactions, which could even see a small speedup due to the changes.

We should add a lot more tests for these changes and the behavior. In our normal tests, intermediate commits are not happening frequently.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_server_aql, shell_client*.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,10 @@
+v3.8.1 (XXXX-XX-XX)
+-------------------
+
+* Fix internal iterator states after intermediate commits in write
+  transactions.
+
+
 v3.8.0 (XXXX-XX-XX)
 -------------------
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,8 @@ v3.8.1 (XXXX-XX-XX)
 -------------------
 
 * Fix internal iterator states after intermediate commits in write
-  transactions.
+  transactions. Iterators could point to invalid data after an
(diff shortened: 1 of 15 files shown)


jsteemann:
Requires a different fix. Closing this now until a working solution is available in devel, and then will restart work on the backport.

PR #14351, opened 2021-06-10
Introduce callstack splitting to avoid stackoverflows with large AQL queries

mpoeter:
### Scope & Purpose

Fix potential stack overflow when executing large queries. This is achieved by splitting the callstack and moving part of the execution to a separate thread. The number of execution nodes after which such a callstack split should be performed can be configured via the query option `maxNodesPerCallstack` and the command line option `--query.max-nodes-per-callstack`; the default is 250.

#### Backports:

- [ ] No backports required

#### Related Information

- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/697
- [x] Jira ticket number: https://arangodb.atlassian.net/browse/APM-77

### Testing & Verification

### Documentation


arangod/RestServer/QueryRegistryFeature.cpp
```diff
@@ -301,6 +302,12 @@ void QueryRegistryFeature::collectOptions(std::shared_ptr<ProgramOptions> option
   options->addOption("--query.optimizer-max-plans",
                      "maximum number of query plans to create for a query",
                      new UInt64Parameter(&_maxQueryPlans));
+                     
+options->addOption("--query.max-nodes-per-callstack",
+                     "maximum number execution nodes on the callstack before "
+                     "splitting the remaining nodes into a separate thread",
+                     new UInt64Parameter(&_maxNodesPerCallstack),
+                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden));
```
jsteemann, line 310:
```suggestion
                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden))
                     .setIntroducedIn(30900);
```

jsteemann (approved):

arangod/RestServer/QueryRegistryFeature.cpp
```diff
@@ -301,6 +302,13 @@ void QueryRegistryFeature::collectOptions(std::shared_ptr<ProgramOptions> option
   options->addOption("--query.optimizer-max-plans",
                      "maximum number of query plans to create for a query",
                      new UInt64Parameter(&_maxQueryPlans));
+                     
+options->addOption("--query.max-nodes-per-callstack",
+                     "maximum number execution nodes on the callstack before "
+                     "splitting the remaining nodes into a separate thread",
+                     new UInt64Parameter(&_maxNodesPerCallstack),
+                     arangodb::options::makeDefaultFlags(arangodb::options::Flags::Hidden))
+                     .setIntroducedIn(30900);
```
goedderz, line 311: Will this be backported to 3.8 or older?

goedderz (commented):
Very clear PR, thank you. 👍 

I'd like to point out two things. You're surely aware and have discussed it, I just want to make sure it was discussed, and that I've understood it correctly.

First, we'll start roughly #nodes / maxNodesPerCallstack many threads, of which only one is working at any point in time. I assume that is intentional (and I support it) that these work past the scheduler, in order not to clog it. They probably use up a few MB of memory each depending on the system, so we should clearly document that `maxNodesPerCallstack` should be only carefully lowered.

Second, there are other recursive algorithms on the plan, mainly `ExecutionNode::walk` which executes the `WalkerWorker`s. And a handful others, e.g. `ExecutionNode::estimateCost()` and other `ExecutionNode` methods. These surely need an order of magnitude less stack space per node than the `ExecutionBlockImpl::execute` calls do, but will eventually hit a limit as well. If this is high enough that we don't care, that's fine with me. Or it can also be addressed later, it shouldn't be hard to rewrite them in a non-recursive fashion.

On a side note, would it make sense to allow the stack size to be increased via `setrlimit` / `RLIMIT_STACK` and a user-provided parameter, as far as allowed by `ulimit -s`? Or do we rather not touch that?

arangod/RestServer/QueryRegistryFeature.cpp
mpoeter: No, at least not for now.

mpoeter:
@goedderz Yes, the additional threads are an intentional tradeoff for this simple and somewhat hacky fix until we have a proper fix. A documentation PR with more details is on my todo list. The default value of 250 was sufficiently low to work correctly on my windows machine which uses a default stack size of 1MB. AFAIK Linux uses a larger stack size, so theoretically we could use OS dependent defaults. WDYT?

Thanks for pointing out other places with recursive algorithms on the plan. I was aware of `ExecutionNode::walk` which was next on my todo list, but not `ExecutionNode::estimateCost()`.

Regarding increased stack size - I would rather not touch that, but if a user wants to increase the stack size they are free to increase `maxNodesPerCallstack`. Sooner or later we need a proper fix for this anyway.

goedderz:
> @goedderz Yes, the additional threads are an intentional tradeoff for this simple and somewhat hacky fix until we have a proper fix. A documentation PR with more details is on my todo list. The default value of 250 was sufficiently low to work correctly on my windows machine which uses a default stack size of 1MB. AFAIK Linux uses a larger stack size, so theoretically we could use OS dependent defaults. WDYT?

250 sounds like a fine default to me. While Linux has defaults, these can always be changed per system, so if we want to adapt the default to the stack size, we might want to look at the actual stack size. However, that might be overly complicated for a small gain, So I'm fine with letting the default be 250.

> Thanks for pointing out other places with recursive algorithms on the plan. I was aware of `ExecutionNode::walk` which was next on my todo list, but not `ExecutionNode::estimateCost()`.

> Regarding increased stack size - I would rather not touch that, but if a user wants to increase the stack size they are free to increase `maxNodesPerCallstack`. Sooner or later we need a proper fix for this anyway.

Sure, sounds sensible.

goedderz (approved):

arangod/Aql/ExecutionBlockImpl.cpp
```diff
@@ -1969,6 +1980,72 @@ void ExecutionBlockImpl<Executor>::PrefetchTask::execute(ExecutionBlockImpl& blo
...
+  _thread.join();
+}
+
+template <class Executor>
+auto ExecutionBlockImpl<Executor>::CallstackSplit::execute(ExecutionContext& ctx, AqlCallType const& aqlCall)
+    -> UpstreamResult {
+  std::variant<UpstreamResult, std::exception_ptr, std::nullopt_t> result{std::nullopt};
+  Params params{result, ctx, aqlCall};
+  
+  _lock.lock();
```
mchacki, line 2004: I would prefer to use a guard here.
Right now this code is perfectly fine.
But in case something between lock() and unlock() throws the lock will be held, and the Destructor of this class will block as it is trying to get the lock again.

As far as I see everything between lock and unlock is no throw. So for now all is fine, nothing to block this PR for.

tests/js/server/aql/aql-queries-simple.js
```diff
@@ -1335,6 +1335,18 @@ function ahuacatlQuerySimpleTestSuite () {
...
+      let q = "";
+      const cnt = 1000;
+      for (let i = 0; i < cnt; ++i) {
+        q += `LET v${i} = NOOPT(1)\n`;
+      }
+      q += "RETURN v0";
+      for (let i = 1; i < cnt; ++i) {
+        q += ` + v${1}`;
+      }
+      assertEqual([cnt], AQL_EXECUTE(q, {}, {optimizer: {rules: ['-all']}}).json);
```
mchacki, line 1348: Can we add another Assertion here that the resulting plan has >1000 Blocks.
And best if you could assert that the Callstack split has happened.
For some systems 1000 Blocks my be still okay without splitting if your StackSize is high enough (or if for some reason we can reduce the stacksize in AQL later on)

arangod/Aql/OptimizerRules.cpp
```diff
@@ -7889,6 +7905,34 @@ void arangodb::aql::enableAsyncPrefetching(ExecutionPlan& plan) {
...
+    // V8 requires thread local context configuration, so we cannot 
+    // use our thread based split solution...
+    return;
+  }
+  
+  auto const& options = plan.getAst()->query().queryOptions();
+  struct CallstackSplitter : WalkerWorkerBase<ExecutionNode> {
+    explicit CallstackSplitter(size_t maxNodes) : maxNodesPerCallstack(maxNodes) {}
+    bool before(ExecutionNode* n) override {
+      if (n->getType() == EN::REMOTE) {
```
mchacki, line 7919: Can we add an assert here that we do not see a Subquery node?
A subquery node is the non-inlined variant that is still present to make Planning easier.
This will branch off and would not account the Subquery itself for the CalStackSize then.
If the Subquery is later inlined we may still break up on the callStack.

e.g.:
```
LET sub = (
  LET q0 = 1
  [...]
  LET qn = 1
  RETURN qn
)
RETURN sub
```
would have a main call stack size of 3 (Singleton, Subquery, Return) before inlining.
While the subquery could be well over the StackSize.

The way this rule is applied I am sure that we do not need to handle the above case, we need to make sure that we not accidentally get into it with refactoring, so an assert is enough.

tests/js/server/aql/aql-queries-simple.js
```diff
@@ -1335,6 +1335,18 @@ function ahuacatlQuerySimpleTestSuite () {
...
+    }
```
mchacki, line 1349: I would like to ask for a second Test like the one you already have,
but with the CallStackOverflow hidden within a Subquery (see my comment in the Optimizer Rule)

tests/js/server/aql/aql-queries-simple.js
```diff
@@ -1335,6 +1335,18 @@ function ahuacatlQuerySimpleTestSuite () {
...
   };
```
mchacki, line 1350: I would like to ask for a third Test, pretty much again the same as you have.
However this time on a One-Shard Database (EE) with a Single Collection Access included.
The reasoning behind this is:
If we are on OneShard the query will be optimized onto a single DBServer, so the Splitting needs to take place on a different Server, not the Coordinator.

I would love to have such a test on a "normal" collection access, but it will be utterly hard to write an AQL containing 1000 nodes that are pushed into the same DBServer snippet and not be moved/optimized away. You cannot get away with all rules disabled here.

mchacki (changes requested):
Technically and in general I very very good PR I like it a lot 👍 

All my comments are just for some more test cases making it slightly harder for the rule here, all just to be on the save side.

arangod/Aql/OptimizerRules.cpp
jsteemann: @mchacki: I think in maintainer mode we already walk the plan at the end of all optimizations and make sure that there are no subquery nodes anymore:
https://github.com/arangodb/arangodb/blob/devel/arangod/Aql/Optimizer.cpp#L384

Do you think we should add an extra assertion to the .js test for this here?

mchacki: I would like to have an assertion on the ExecutionNode* n input at exactly this point.

This walker is standalone and is not supporting SubqueryNode.
So I agree right now but outer circumstances it is assured that there will never be a SubqueryNode here, and the assertion will not trigger.
But who knows, in a couple of month someone might move rule ordering around and suddenly we get this Walker at a position where the SubqueryNode is injected and it will silently fail (as it does not react to it)

This assertion is just to save our future selves.

mpoeter: Done!

arangod/Aql/ExecutionBlockImpl.cpp
mpoeter: Done!

tests/js/server/aql/aql-queries-simple.js
mpoeter: I intentionally did not add any assertion about callstack splitting. This test case is meant to test that large queries can be executed successfully without caring _how_ this is achieved. Increasing the number of blocks beyond 1000 causes crashes on my windows machine, because `ExecutionNode::walk` is also implemented recursively which results in a stackoverflow. This will be fixed in a separate PR, at which point we can increase the number of blocks in this test.

tests/js/server/aql/aql-queries-simple.js
mpoeter: Done.

tests/js/server/aql/aql-queries-simple.js
mpoeter: Actually this should be relatively simple:
```
let q = "FOR d IN docs LET x = (\n";
const cnt = 900;
for (let i = 0; i < cnt; ++i) {
  q += `LET v${i} = NOOPT(1)\n`;
}
q += "RETURN d.v + v0";
for (let i = 1; i < cnt; ++i) {
  q += ` + v${i}`;
}
q += ")[0]\nSORT x RETURN x";
```
Assuming we have a collection `docs` containing documents with an attribute `v` ranging from 0..x this should return `[cnt..cnt + x]`. In a one-shard setup the whole query should be moved to the DB servers, and since all calculations use `NOOPT` nothing should be optimized away. I will add an according EE test.

tests/js/server/aql/aql-queries-simple.js
mchacki: okay. :+1:

tests/js/server/aql/aql-queries-simple.js
mchacki: I agree on the one-shard setup.
I was looking for a non-one-shard test, but that will be hard to write to get a single snippet, will check the enterprise test, thanks for adding.

mchacki (approved):
LGTM

mchacki (approved):
LGTM Thanks for following my comments :+1:

jsteemann (approved):

PR #14357, opened 2021-06-11
Fix URL request parsing.

mpoeter:
### Scope & Purpose

Fix URL request parsing in case data is handed in in small chunks. Previously the URL could be cut off if the chunk size was smaller than the URL size.

#### Backports:

- [x] Backports required for: 3.6, 3.7, 3.8

### Testing & Verification

  - [x] Added new **integration tests** in shell_client_aql

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Fix URL request parsing in case data is handed in in small chunks.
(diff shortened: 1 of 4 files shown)


jsteemann (approved):

jsteemann:
Tested the fix manually and works for me! :+1:

PR #14358, opened 2021-06-11
Fix URL request parsing.

mpoeter:
### Scope & Purpose

Fix URL request parsing in case data is handed in in small chunks. Previously the URL could be cut off if the chunk size was smaller than the URL size.

#### Backports:

- [x] Backports required for: 3.6, 3.7, devel

### Testing & Verification

  - [x] Added new **integration tests** in shell_client_aql

### DIFF
--- a/tests/js/client/shell/shell-request.js
+++ b/tests/js/client/shell/shell-request.js
@@ -34,14 +34,16 @@ var request = require('@arangodb/request');
 var url = require('url');
 var querystring = require('querystring');
 var qs = require('qs');
+const deriveTestSuite = require('@arangodb/test-helper').deriveTestSuite;
+const internal = require("internal");
 
+'use strict';
 
 ////////////////////////////////////////////////////////////////////////////////
 /// @brief test suite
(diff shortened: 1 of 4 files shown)


tests/js/client/shell/shell-request.js
```diff
@@ -395,12 +397,33 @@ function RequestSuite () {
...
+  deriveTestSuite(BaseRequestSuite(), suite, '_SmallChunks');
+  return suite;
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 /// @brief executes the test suite
 ////////////////////////////////////////////////////////////////////////////////
 
 jsunity.run(RequestSuite);
+jsunity.run(RequestSuiteWithSmallChunks);
```
jsteemann, line 426: I think we can only run this suite if the executable is configured with `ENABLE_FAILURE_TESTS`.
Otherwise the test will try to set failure points, but they are not compiled in.

jsteemann (approved):

PR #13342, opened 2021-01-07
more granular error reporting in Maintenance

jsteemann:
### Scope & Purpose

Add more granular error reporting to `reportInCurrent` maintenance function, and add lots of assertions.
This is not a bugfix, but a vehicle to simplify debugging and bug hunts.

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is a trivial rework / code cleanup without any test coverage.

Link to Jenkins PR run:
http://172.16.10.101:8080/view/PR/job/arangodb-matrix-pr/13495/

PR #12594, opened 2020-09-02
Feature/agency cache database subscription

kvahed:
### Scope & Purpose

*(Please describe the changes in this PR for reviewers)*

- [x] :hankey: Bugfix 
- [ ] :pizza: New feature 
- [ ] :hammer: Refactoring 
- [ ] :book: CHANGELOG entry made
- [ ] :muscle: The behavior in this PR was *manually tested*
- [ ] :computer: The behavior change can be verified via automatic tests

#### Backports:

- [ ] No backports required
- [ ] Backports required for: *(Please specify versions)*

#### Related Information

*(Please reference tickets / specification etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added **Regression Tests**
  - [ ] Added new C++ **Unit Tests**
  - [ ] Added new **integration tests** (i.e. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)

Additionally:

- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

*(Include link to Jenkins run etc)*

### Documentation

> All new Features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the changelog so that
> developers and users have a concise overview. 

- [ ] Added a *Changelog Entry* (referencing the corresponding public or internal issue number)
- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

Please note that for legal reasons we require you to sign the [Contributor Agreement](https://www.arangodb.com/documents/cla.pdf)
before we can accept your pull requests.

PR #14361, opened 2021-06-11
Fixed a velocypack type error in maintenance::reportInCurrent

goedderz:
### Scope & Purpose

In a velocypack type change introduced in #12594, an error-handling branch was overlooked, rendering the error-handling code useless, always resulting in an exception (or assertion in maintainer mode).

Fixes errors like the following:

```
WARNING [ceb1a] {maintenance} caught exception in Maintenance shards error reporting: Expecting Object
ERROR [c9a75] {maintenance} Error reporting in current: Expecting Object
```

It seems there was a previous effort to find the source of this error, or a similar one, in #13342.

#### Backports:

As far as I can tell, the PR introducing the bug was merged into >=3.7, but not into 3.6.

- [X] Backport required for 3.8: #14472
- [X] Backport required for 3.7: #14473

#### Related Information

*(Please reference tickets / specification / other PRs etc)* **TODO**

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

**TODO** needs additional unit tests. This code is currently completely untested!


arangod/Cluster/Maintenance.cpp
```diff
@@ -1466,7 +1466,9 @@ arangodb::Result arangodb::maintenance::reportInCurrent(
 
           std::string c = key.substr(pos + 1, pos2);
           std::string s = key.substr(pos2 + 1);  // shard name
-          auto const pdb = pit->second->slice();
+          TRI_ASSERT(pit->second->slice().isArray());
+          TRI_ASSERT(pit->second->slice().length() == 1);
```
gnusi, line 1470: how can we guarantee that?

goedderz: Only by the way the `plan` argument is created in the first place. Same holds for `current` and `local`. The proper solution IMO would be to replace VelocyPack with static data structures. But that touches the whole maintenance.

So we should introduce tests in this PR for `reportInCurrent`, and the code that writes it input in the first place.

kvahed: @gnusi, yes it is guaranteed. You will not ever get here without a plan object.

kvahed (approved):
LGTM

arangod/Cluster/Maintenance.cpp
gnusi: @kvahed having an array of exactly one element looks really strange. Where is the code building this piece of the plan located?

PR #14368, opened 2021-06-15
Fixes for shard synchronization

jsteemann:
### Scope & Purpose

* Fix partial cleanup of internal write batches for multi-document operations of
  which one or multiple failed. The previous implementation had an unreleased
  performance optimization that wouldn't clean up the write batch completely.
  That could have led to a wrong sequence of events being accumulated in the 
  write batch, which may have confused the WAL tailing API later. This bug was
  only present in 3.8 RCs.

* Fix some occurences in which Merkle trees could silently apply the same change
  multiple times, which led to data drift between the Merkle tree and the
  underlying collection's data. This bug was only present in 3.8 but no earlier
  versions.

  end, because the follower may be responsbile  for _other_ shards as well. 

* Added more context to "dropping follower" messages, so it is easier to analyze
  what exactly went wrong.

* Fixed invalid shard synchronization for documents not added via INSERT with
  `overwriteMode` set to `ignore`. In this case, if a document with the given key
  already exists, it is not changed on the leader (i.e. no write happens on the
  leader). However, a write was replicated to the follower, which was wrong.
  This write is now suppressed, which can only make such insert operations faster.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -2692,6 +2692,22 @@ void RestReplicationHandler::handleCommandHoldReadLockCollection() {
     return;
   }
 
+  // check if we are not the actual leader anymore. this can
+  // happen if there is a leader change after a follower scheduled a
+  // synchronize shard job
+  bool isLeader = col->followers()->getLeader().empty();
+  if (!isLeader) {
+    auto res = cancelBlockingTransaction(id);
```
maierlars, line 2700: Would this transaction block the current server (now follower) from starting a SynchronizeShard with the new leader?

arangod/Transaction/Methods.cpp
```diff
@@ -1164,14 +1175,17 @@ Future<OperationResult> transaction::Methods::insertLocal(std::string const& cna
           prevDocResult.revisionId().isSet()) {
         TRI_ASSERT(didReplace);
 
+        TRI_ASSERT(res.ok());
```
maierlars, line 1178: That assertion is only executed if `res` is `TRI_ERROR_ARANGO_CONFLICT` and thus _not_ ok. Wasn't that triggered by jenkins?

arangod/Transaction/Methods.cpp
```diff
@@ -1825,19 +1862,19 @@ Future<OperationResult> transaction::Methods::truncateLocal(std::string const& c
...
+          return OperationResult(TRI_ERROR_ARANGO_READ_ONLY, options);
+        case FollowerInfo::WriteState::STARTUP:
+          return OperationResult(TRI_ERROR_CLUSTER_BACKEND_UNAVAILABLE, options);
+        default:
+          break;
       }
 
       // fetch followers
+      replicationType = ReplicationType::LEADER;
       followers = followerInfo->get();
```
maierlars, line 1876: What bug was fixed here by setting `replicationType` to `Leader` even when there are no followers?

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+    failed = false;
+    Object.entries(shardInfo).forEach(
+      ([shard, [leader, follower]]) => {
+        const leaderTree = fetchRevisionTree(getServerUrl(leader), shard);
+        leaderTree.computed.nodes = "<reduced>";
+        leaderTree.stored.nodes = "<reduced>";
+        if (!leaderTree.equal) {
+          console.error(`Leader has inconsistent tree for shard ${shard}:`, leaderTree);
+          failed = true;
+        };
```
goedderz, line 104:
```suggestion
        }
```

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+        
+        const followerTree = fetchRevisionTree(getServerUrl(follower), shard);
+        followerTree.computed.nodes = "<reduced>";
+        followerTree.stored.nodes = "<reduced>";
+        if (!followerTree.equal) {
+          console.error(`Follower has inconsistent tree for shard ${shard}:`, followerTree);
+          failed = true;
+        };
```
goedderz, line 112:
```suggestion
        }
```

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+
+        if (!_.isEqual(leaderTree, followerTree)) {
+          console.error(`Leader and follower have different trees for shard ${shard}`);
+          console.error("Leader: ", leaderTree);
+          console.error("Follower: ", followerTree);
+          failed = true;
+        }
```
goedderz, line 119: Both `.computed.nodes` and `.stored.nodes` are deleted, so this compares only the metadata if I'm not mistaken. Is this intended, e.g. because we rely on count & hash being different?

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+}
+/*
+for (let i = 0; i < (1 << params.length); ++i) {
+  let paramValues = [];
+  for (let j = params.length - 1; j >= 0; --j) {
+    paramValues.push(Boolean(i & (1 << j)));
+  }
+  addSuite(paramValues);
+}
+*/
```
goedderz, line 298:
```suggestion
```

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+
+// ATM we only create a single suite with all options except Truncate, because there are still known issues.
+// Later we probably want to have all possible combinations, at least for the nightly builds.
+addSuite([true, true, false, true, false]);
```
goedderz, line 302: Can this be resolved now, i.e. add all suites here or in another `nightly` file?

goedderz (commented):
This is a partial review only.

arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -216,6 +224,9 @@ class RocksDBBatchedWithIndexMethods final : public RocksDBMethods {
   rocksdb::Status RollbackToSavePoint() override {
     return rocksdb::Status::OK();
   }
+  rocksdb::Status RollbackToWriteBatchSavePoint() override {
+    return rocksdb::Status::OK();
+  }
   void PopSavePoint() override {}
```
goedderz, line 230: Is it a good idea to have these methods just do nothing, rather than throwing a _not implemented_ error?

arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -185,6 +190,9 @@ class RocksDBBatchedMethods final : public RocksDBMethods {
...
```
goedderz, line 196: Is it a good idea to have these methods just do nothing, rather than throwing a not implemented error?

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      if (plan[s].followers.length === current[s].followers.length) {
+        ++good;
+      }
+    }
+    if (good === shards.length) {
+      break;
+    }
+    internal.wait(1);
+  }
+}
```
goedderz, line 129: This is essentially the same as `waitForShardsInSync` in the other test, can we move them into `test-helper.js` or so to deduplicate?

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      switchConnectionToLeader(collInfo);
+      let trxsLeader = arango.GET("/_api/transaction");
+      assertTrue(trxsLeader.hasOwnProperty("transactions"));
+      let followerTrxId = String(Number(trxid) + 2);
+      let found = false;
+      for (let t of trxsLeader.transactions) {
+        if (t.id === followerTrxId) {
+          found = true;
+        }
+      }
```
goedderz, line 295:
```suggestion
      let found = trxsLeader.transactions.some(t => t.id === followerTrxId);
```

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      assertFalse(found);
+      switchConnectionToFollower(collInfo);
+      let trxsFollower = arango.GET("/_api/transaction");
+      assertTrue(trxsFollower.hasOwnProperty("transactions"));
+      for (let t of trxsFollower.transactions) {
+        if (t.id === followerTrxId) {
+          found = true;
+        }
+      }
```
goedderz, line 304:
```suggestion
      found = trxsFollower.transactions.some(t => t.id === followerTrxId);
```

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      // coordinator otherwise we do not see the follower information.
+
+      switchConnectionToCoordinator(collInfo);
+      waitForShardsInSyncAgain(cn);
+
+      switchConnectionToLeader(collInfo);
+      let count = 0;
+      for (let s of collInfo.shards) {
+        count += db._collection(s).count();
+      }
```
goedderz, line 227:
```suggestion
      let count = _.sumBy(collInfo.shards, s => db._collection(s).count());
```

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      assertEqual(21, count);
+
+      switchConnectionToFollower(collInfo);
+      count = 0;
+      for (let s of collInfo.shards) {
+        count += db._collection(s).count();
+      }
```
goedderz, line 234:
```suggestion
      count = _.sumBy(collInfo.shards, s => db._collection(s).count());
```

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      // coordinator otherwise we do not see the follower information.
+
+      switchConnectionToCoordinator(collInfo);
+      waitForShardsInSyncAgain(cn);
+
+      switchConnectionToLeader(collInfo);
+      let count = 0;
+      for (let s of collInfo.shards) {
+        count += db._collection(s).count();
+      }
```
goedderz, line 318:
```suggestion
      let count = _.sumBy(collInfo.shards, s => db._collection(s).count());
```

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
```diff
@@ -0,0 +1,339 @@
...
+      assertEqual(40, count);
+
+      switchConnectionToFollower(collInfo);
+      count = 0;
+      for (let s of collInfo.shards) {
+        count += db._collection(s).count();
+      }
```
goedderz, line 325:
```suggestion
      count = _.sumBy(collInfo.shards, s => db._collection(s).count());
```

CHANGELOG
```diff
@@ -1,5 +1,36 @@
...
+* Fix some occurrences in which Merkle trees could silently apply the same change
+  multiple times, which led to data drift between the Merkle tree and the
+  underlying collection's data. This bug was only present in 3.8 but no earlier
+  versions.
+
+* On a failure during synchronous replication, do not remove the failed follower 
+  from the list of known servers in the transaction.
+  If we do, we would not be able to send the commit/abort to the follower later.
+  However, we still need to send the commit/abort to the follower at transaction 
+  end, because the follower may be responsbile  for _other_ shards as well.
```
goedderz, line 20:
```suggestion
  end, because the follower may be responsible  for _other_ shards as well.
```

goedderz (commented):
This is the last part of my review.

tests/js/client/shell/shell-dropped-followers-elcheapo-commit-cluster.js
goedderz: (don't forget to `const _ = require('lodash');`)

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+    },
+
+    tearDown: function () {
+      clearAllFailurePoints();
+      db._drop(cn);
+    },
+    
+    testWorkInParallel: function () {
+      let code = (testOpts) => {
+        const key = () => "testmann" + Math.floor(Math.random() * 100000000);
```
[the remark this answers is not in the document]
neunhoef: No, the idea is here to use the birthday paradox and have a certain amount of collisions. The babies API is supposed to go through and report individual collisions. Same with removes, we intentionally try to remove lots of documents which are not actually there. We even had variants where the number was much smaller to have even more conflicts. We should add a comment here, though.

arangod/Transaction/Methods.cpp
jsteemann: Good catch. This is a leftover. Will remove the wrong assert!

arangod/Transaction/Methods.cpp
jsteemann: No bug is fixed by this. However, truncate did this differently than insert/update/replace/remove:
In truncate, `replicationType` was set to `LEADER` only if there is at least one follower.
In the other write operations, `replicationType` was set to `LEADER` even if there are 0 followers. Now it will be at least consistent.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1064,7 +1060,11 @@ ResultT<TRI_voc_tick_t> SynchronizeShard::catchupWithReadLock(
     if (!res.ok()) {
       auto errorMessage = StringUtils::concatT(
           "SynchronizeShard: error in startReadLockOnLeader (soft):", res.errorMessage());
-      return ResultT<TRI_voc_tick_t>::error(TRI_ERROR_INTERNAL, errorMessage);
+      if (res.is(TRI_ERROR_CLUSTER_NOT_LEADER)) {
+        return ResultT<TRI_voc_tick_t>::error(TRI_ERROR_CLUSTER_NOT_LEADER, std::move(errorMessage));
+      } else {
+        return ResultT<TRI_voc_tick_t>::error(TRI_ERROR_INTERNAL, std::move(errorMessage));
+      }
```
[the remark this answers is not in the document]
jsteemann: Good idea, will fix it.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1152,7 +1152,11 @@ Result SynchronizeShard::catchupWithExclusiveLock(
   if (!res.ok()) {
     auto errorMessage = StringUtils::concatT(
         "SynchronizeShard: error in startReadLockOnLeader (hard):", res.errorMessage());
-    return {TRI_ERROR_INTERNAL, std::move(errorMessage)};
+    if (res.is(TRI_ERROR_CLUSTER_NOT_LEADER)) {
+      return {TRI_ERROR_CLUSTER_NOT_LEADER, std::move(errorMessage)};
+    } else {
+      return {TRI_ERROR_INTERNAL, std::move(errorMessage)};
+    }
```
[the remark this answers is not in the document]
jsteemann: Will fix

arangod/RocksDBEngine/RocksDBMethods.h
jsteemann: "Not implemented" would actually not be right here in all cases, I guess.
The changes I made try to mimic the behavior of the existing `RollbackToSavePoint()` method, which already had some implementations that simply returned `Status::Ok()`. I will make it clear via code comments that returning `Ok` is not an oversight here.

arangod/RocksDBEngine/RocksDBMethods.h
jsteemann: See above.

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+        };
+
+        let c = db._collection(testOpts.collection);
+        const opts = () => {
+          let result = {};
+          if (testOpts.withIntermediateCommits) {
+            const r = Math.random();
+            if (r <= 0.5) {
+              result.intermediateCommitCount = 7 + Math.floor(r * 10);
+            }
```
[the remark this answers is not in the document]
mpoeter: I simply wanted to have some small random number, but you are right, we could just as well roll a new one.

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -193,6 +194,12 @@ RestStatus RestDocumentHandler::insertDocument() {
     return RestStatus::DONE;
   }
 
+  if (ServerState::instance()->isDBServer() &&
+      _activeTrx->state()->collection(cname, AccessMode::Type::WRITE) == nullptr) {
+    generateTransactionError(cname, OperationResult(Result(TRI_ERROR_CLUSTER_SHARD_FOLLOWER_REFUSES_OPERATION), opOptions), "");
+    return RestStatus::DONE;
+  }
```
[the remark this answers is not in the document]
jsteemann: It is possible that this happens on a follower.
The leader will at some point, potentially lazily, send a transaction begin request to a follower as part of a replicated write operation. Included in this transaction begin request are the affected (written-to) shards. In edge cases it is possible that the leader puts together the list of followers for a shard for an operation, and then (a bit later) creates the actual requests to these followers. If in the meantime a follower got dropped by a concurrent operation, there is a small window of time in which a request is generated for an already-dropped follower. In this case, the `buildTransactionBody()` function in ClusterTrxMethods.cpp will not include the target shard in the transaction begin request. 
If such requests arrive on followers, they previously failed with assertion failures because it was attempted to write in a read-only transaction (read-only because no shards were listed in either "write" or "exclusive"). With the change, the follower will instantly reject such requests.

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+        let trx = null;
+        let ctx = db;
+        if (testOpts.withStreamingTransactions && Math.random() < 0.5) {
+          trx = db._createTransaction({ collections: { write: [c.name()] } });
+          ctx = trx;
+          c = trx.collection(testOpts.collection);
+          query = trx.query;
+        }
+
+        const ops = Math.round(Math.random() * 5) + 1;
```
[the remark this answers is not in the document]
mpoeter: Good point, it's not really important, but I will change it nonetheless.

arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -94,11 +95,13 @@ RocksDBMetadata::RocksDBMetadata()
  * @param  seq   The sequence number immediately prior to call
  * @return       May return error if we fail to allocate and place blocker
  */
-Result RocksDBMetadata::placeBlocker(TransactionId trxId, rocksdb::SequenceNumber seq) {
+Result RocksDBMetadata::placeBlocker(TransactionId trxId, rocksdb::SequenceNumber& seq) {
   return basics::catchToResult([&]() -> Result {
     Result res;
     WRITE_LOCKER(locker, _blockerLock);
 
+    seq = std::max(seq, _maxBlockersSequenceNumber);
```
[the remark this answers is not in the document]
jsteemann: For each collection, we keep track of the highest assigned sequence number used for a blocker.

This is necessary because otherwise readers could be reading the list of blockers sequence numbers and determine its lower bound, assuming that no lower seqnos will ever be added later. That is at least an assumption that readers currently make. This can happen in rare cases when two write operations were in flight at the same time but scheduled in adverse order. In this case, the higher seqno could be inserted first, and then the lower one. If in-between a reader would have fetched the lower bound, it was not aware of the in-flight operation which would insert a lower value later.

With this change, this is now prevented.

CHANGELOG
jsteemann: Will fix

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+              c.truncate();
+            } else if (d >= 0.9) {
+              let o = opts();
+              let d = docs();
+              console.warn("RUNNING AQL INSERT WITH " + d.length + " DOCS. OPTIONS: " + JSON.stringify(o));
+              query.call(ctx, "FOR doc IN @docs INSERT doc INTO " + c.name(), {docs: d}, o);
+            } else if (d >= 0.8) {
+              let o = opts();
+              console.warn("RUNNING AQL REMOVE. OPTIONS: " + JSON.stringify(o));
+              query.call(ctx, "FOR doc IN " + c.name() + " LIMIT 138 REMOVE doc IN " + c.name(), {}, o);
```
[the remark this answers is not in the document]
mpoeter: It is arbitrary and I just updated the test so that the limit is now chosen randomly.

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,304 @@
...
+            } else if (d >= 0.7) {
+              let o = opts();
+              console.warn("RUNNING AQL REPLACE. OPTIONS: " + JSON.stringify(o));
+              query.call(ctx, "FOR doc IN " + c.name() + " LIMIT 1338 REPLACE doc WITH { pfihg: 434, fjgjg: 555 } IN " + c.name(), {}, o);
```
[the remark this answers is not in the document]
mpoeter: It is arbitrary and I just updated the test so that the limit is now chosen randomly.

tests/js/client/communication/test-chaos-cluster.js
mpoeter: This code dynamically creates test suites for all parameter combinations. ATM we don't use it since we don't want to include ALL suites in the PR tests, but we do want to have them in the nightlies, but that will be done in a separate PR.

tests/js/client/communication/test-chaos-cluster.js
mpoeter: Yes, this is intentional. We remove the computed and stored nodes since we may want to print the trees, but we don't want to print the 262k buckets! Equality of the trees is checked using the single combined hash and document count.

I've added an according comment.

js/client/modules/@arangodb/test-helper.js
```diff
@@ -173,3 +176,147 @@ exports.getMetric = function (endpoint, name) {
...
+  return result.pid;
+};
+
+const buildCode = function(key, command, cn) {
+  let file = fs.getTempFile() + "-" + key;
+  fs.write(file, `
+(function() {
+let tries = 0;
+while (true) {
+  if (++tries % 3 === 0) {
```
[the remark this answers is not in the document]
mpoeter: We want to check the stop signal every three iterations.

tests/js/client/communication/test-chaos-cluster.js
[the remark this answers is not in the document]
goedderz: Okay good, thanks!

arangod/RocksDBEngine/RocksDBMethods.h
goedderz: Okay, thanks for the clarification!

arangod/RocksDBEngine/RocksDBMetadata.cpp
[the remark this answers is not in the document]
goedderz: So it would in principle suffice to use the minimum sequence number of any blocker, correct? And I suppose you used the maximum instead because it's slightly cheaper to keep up to date?

js/client/modules/@arangodb/test-helper.js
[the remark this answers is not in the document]
goedderz: Ah, I misread. Thanks.

arangod/Transaction/Methods.cpp
maierlars: However, this shows that this code segment has close to zero test coverage. 😟

arangod/RestHandler/RestDocumentHandler.cpp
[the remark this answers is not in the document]
goedderz: Okay, so this code (in this case) is executed on the follower, correct? So in case we got such a request like you described here, we're executing

```cpp
...
  _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions);
...
  Result res = _activeTrx->begin();
// res.ok()
...
  _activeTrx->state()->collection(cname, AccessMode::Type::WRITE) == nullptr
```

. But if this follower got dropped for that shard, I still don't quite understand how `createTransaction` (and `begin()`) can succeed at all in that case?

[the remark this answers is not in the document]
jsteemann: The follower got dropped for the shard, but it does not need to know about it when we are here. It only gets a request to begin a transaction, however, the request does not include its own shard in either "write" or "exclusive". The whole point of this extra check is to detect this, and make the operation fail early.
If we don't do this, it may believe that it is still a follower further inside, and try to carry out the write operation inside the read-only transaction, which will lead to an assertion failure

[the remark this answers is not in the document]
goedderz: Okay thanks.

I think though that the check you added should be moved inside `RestVocbaseBaseHandler::createTransaction`, so that we no longer can successfully order a transaction for a collection that does not contain this collection. This will prevent something similar from happening in other places as well.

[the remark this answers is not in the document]
neunhoef: As far as I see `RestVocbaseBaseHandler::createTransaction` handles three cases:
1. No transaction ID was given as HTTP header: then a new SingleCollectionTransaction is created for the given collection.
2. A trxId was given and that transaction already exists.
3. A trxId was given and no transaction exists and is created lazily here.
In case 1. we do not need to check that the newly created transaction contains the given collection.
Cases 2. and 3. are treated the same here, since we are using `ensureManagedTrx` in these cases we can add the check that the transaction has the given collection with the given write mode, and can throw an exception if this is not the case.
I will propose a patch.

tests/js/client/communication/test-chaos-cluster.js
mpoeter: I am going to add the other suites in a separate PR.

arangod/Aql/QueryContext.cpp
```diff
@@ -51,10 +51,10 @@ using namespace arangodb;
 using namespace arangodb::aql;
 
 /// @brief creates a query
-QueryContext::QueryContext(TRI_vocbase_t& vocbase)
+QueryContext::QueryContext(TRI_vocbase_t& vocbase, QueryId id)
     : _resourceMonitor(GlobalResourceMonitor::instance()),
       _baseOverHeadTracker(_resourceMonitor, baseMemoryUsage),
-      _queryId(TRI_NewServerSpecificTick()),
+      _queryId(id ? id : TRI_NewServerSpecificTick()),
```
mchacki, line 57:
```suggestion
      _queryId(id != 0 ? id : TRI_NewServerSpecificTick()),
```
I like tests on numbers to be a bit more specific, we had a lot of bool casting errors in the past.

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -1245,7 +1245,12 @@ futures::Future<OperationResult> countOnCoordinator(transaction::Methods& trx,
   auto* pool = trx.vocbase().server().getFeature<NetworkFeature>().pool();
   for (auto const& p : *shardIds) {
     network::Headers headers;
-    ClusterTrxMethods::addTransactionHeader(trx, /*leader*/ p.second[0], headers);
+    if (p.second.empty()) {
```
mchacki, line 1248:
```suggestion
    TRI_ASSERT(!p.second.empty());
    if (p.second.empty()) {
```
I agree with the below error handling and like it.
However I think we should never actually end up in this situation.
I would like to get this smashed into my face on Jenkins.
And not by some circumstances silently ignored.

arangod/Cluster/ClusterTrxMethods.cpp
```diff
@@ -225,13 +225,16 @@ Future<Result> commitAbortTransaction(arangodb::TransactionState* state,
...
   fuerte::RestVerb verb;
   if (status == transaction::Status::COMMITTED) {
+    stateString = "commit";
     verb = fuerte::RestVerb::Put;
   } else if (status == transaction::Status::ABORTED) {
+    stateString = "abort";
     verb = fuerte::RestVerb::Delete;
   } else {
-    TRI_ASSERT(false);
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "invalid state for commit/abort operation");
```
mchacki, line 237:
```suggestion
     TRI_ASSERT(false);
    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "invalid state for commit/abort operation");
```

I would like to keep the Assert, and include the THROW for production.
This is illegal state and should not somehow slip through our Jenkins, just because the tests silently ignore errors on APIs they are not testing.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -445,7 +451,7 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
 
       startRecurringBatchExtension();
     }
-
+
```
mchacki, line 454:
```suggestion

```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -410,6 +410,12 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
 
     LOG_TOPIC("0a10d", DEBUG, Logger::REPLICATION)
         << "client: getting leader state to dump " << vocbase().name();
+    
+    auto batchCancelation = scopeGuard([this]() {
```
mchacki, line 414: I like this guard but I think this is off right now.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -473,7 +479,6 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
 
     // all done here, do not try to finish batch if leader is unresponsive
     if (r.isNot(TRI_ERROR_REPLICATION_NO_RESPONSE) && !_config.isChild()) {
```
mchacki, line 481:
```suggestion
```

Sorry git does not allow me a multiline suggestion here ;(

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -473,7 +479,6 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
...
-      batchFinish();
     }
```
mchacki, line 482:
```suggestion
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -473,7 +479,6 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
 
     // all done here, do not try to finish batch if leader is unresponsive
```
mchacki, line 480:
```suggestion
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -488,19 +493,10 @@ Result DatabaseInitialSyncer::runWithInventory(bool incremental, VPackSlice dbIn
```
mchacki, line 493: Reaching this code would not trigger the guard.
In the case we are Child we will finish the batch.
We will also finish the batch in case of fail()
Both have not been possible in original Code.

Is it possible that we need to `cancel()` the guard under some circumstances here?

mchacki (changes requested):
I have added some items I found.
Most are really non-ciritcal but I am unsure about the ScopeGuard usage in DatabaseInitialSyncer. that could be wrong right now

arangod/Replication/DatabaseInitialSyncer.cpp
neunhoef: Should the batch not be finished in any case, exactly as the code currently does?

tests/js/client/communication/test-chaos-cluster.js
```diff
@@ -0,0 +1,290 @@
...
+}
+/*
+for (let i = 0; i < (1 << params.length); ++i) {
+  let paramValues = [];
+  for (let j = params.length - 1; j >= 0; --j) {
+    paramValues.push(Boolean(i & (1 << j)));
+  }
+  addSuite(paramValues);
+}
+*/
```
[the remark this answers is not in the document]
jsteemann: @mpoeter wanted to handle this but apparently didn't.
It should still be commented out but will be activated soon, separately. Let's leave it as it is for now.

arangod/Cluster/ClusterTrxMethods.cpp
```diff
@@ -64,7 +66,13 @@ void buildTransactionBody(TransactionState& state, ServerID const& server,
         return true;
       }
       if (!state.isCoordinator()) {
-        if (col.collection()->followers()->contains(server)) {
+        bool doInclude =  col.collection()->followers()->contains(server);
```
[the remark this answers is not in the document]
jsteemann: This is already outdated code, which doesn't exist like that anymore.

arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
```diff
@@ -636,7 +670,7 @@ std::vector<arangodb::network::FutureRes> EngineInfoContainerDBServerServerBased
   options.database = dbname;
   options.timeout = network::Timeout(10.0);  // Picked arbitrarily
   options.skipScheduler = true;              // hack to speed up future.get()
-
+
```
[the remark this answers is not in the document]
jsteemann: Doesn't matter. Let's move on.

arangod/Aql/EngineInfoContainerDBServerServerBased.h
```diff
@@ -168,7 +168,7 @@ class EngineInfoContainerDBServerServerBased {
    * they may be leftovers from Coordinator.
    * Will also clear the list of queryIds after return.
    *
-   * @param pool The ConnectionPool
+   * @param clusterQueryId cluster-wide query id (used from 3.8 onwards)
```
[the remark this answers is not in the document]
jsteemann: Ok, will fix

mchacki (changes requested):
There is one unprotected access to inherited Variable. (see comments)
If that is fixed I am happy to aprove this PR.
Every other comment is optional.

mchacki (approved):
LGTM.

All my major concerns have been addressed, every other comment is nice to have and can be addressed with follow up PRs.

tests/js/client/communication/test-chaos-cluster.js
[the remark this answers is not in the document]
mpoeter: I added a comment in the code two days ago, not sure why it was not in the diff yesterday. 

This code dynamically creates test suites for all parameter combinations. ATM we don't use it since we don't want to include ALL suites in the PR tests, but we do want to have them in the nightlies. I am currently working on a separate PR for that which will replace this test entirely.

PR #14398, opened 2021-06-24
Fixes for shard synchronization

jsteemann:
### Scope & Purpose

Partial backport of https://github.com/arangodb/arangodb/pull/14368

  end, because the follower may be responsbile  for _other_ shards as well. 

* Added more context to "dropping follower" messages, so it is easier to analyze
  what exactly went wrong.

* Fixed invalid shard synchronization for documents not added via INSERT with
  `overwriteMode` set to `ignore`. In this case, if a document with the given key
  already exists, it is not changed on the leader (i.e. no write happens on the
  leader). However, a write was replicated to the follower, which was wrong.
  This write is now suppressed, which can only make such insert operations faster.

#### Backports:

- [x] No backports required

### Testing & Verification


KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-ANY-matrix/46/

KVS85:
http://jenkins01.arangodb.biz:8080/job/arangodb-ANY-matrix/51/

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -132,6 +132,7 @@ void RocksDBReplicationContext::removeVocbase(TRI_vocbase_t& vocbase) {
     if (it->second->vocbase.id() == vocbase.id()) {
       if (it->second->isUsed()) {
         LOG_TOPIC("543d4", ERR, Logger::REPLICATION) << "trying to delete used context";
+        it++;
```
mchacki, line 135: are we generating leftover here?

The original code did run into an "endless" loop waiting until all contexts are unused (there are parallel threads working on them) the new code simply jumps over the used contexts, and returns from this function (there is no report of this operation was a success or not)

mchacki (changes requested):
99,9% Looks good to me, but I think there is a glitch for Replication Contexts.

Maybe we should zoom about this, it could be no issue, but I am not familiar enough with the used components to decide this.

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
jsteemann: I think the original code was not correct. It would not make any progress and block the thread for a potential endless amount of time.

mchacki (approved):
My concern is sorted out via discussion.

good to go

PR #14405, opened 2021-06-25
Following term id

neunhoef:
Synchronous replication requests (for insert, replace, update, delete
and truncate operations) have been checked on the follower in so far,
that it was checked that they come from the correct leader (the one the
follower believes to be following). However, it can in rare cases happen
that a follower is dropped and a replication request from before the
dropping only hits the follower **after** it has gotten in sync again.
This is very bad since it can lead to inconsistent data on shard
replicas which are deemed to be "in sync" and thus equal.

This PR attacks this problem: Whenever a leader/follower relationship is
established, the leader invents a random number and sends it to the
follower together with the confirmation that it has aqcuired an
exclusive lock on the shard (one relative late step of the get-in-sync
protocol). The leader stores this number and sends it along with each
synchronous replication request. The follower stores this number, too,
and checks that a synchronous replication requests comes not only from
the right leader, but also from the right "term" and not from some
distant past term.

The method described is relatively straightforward, the implementation
follows exactly the plan outlined above. One caveat is that there is
also the case of controlled leadership handover. In this case, we need
to play the same game and establish a random number identifying the
term.

Note that these numbers are not persisted (not necessary!), therefore we
prefer to use random numbers as opposed to an ascending counter, since
then we could have a certain possibility to reach the same counter value
after a restart.

### Scope & Purpose

#### Backports:

- [*] Backports required for: at least 3.8, probably also 3.7

### Testing & Verification

- [*] This change is already covered by existing tests, such as all
      tests bringing shards in sync. The new chaos test which delays
      synchronous replication requests artificially.
  - [*] Added new JS test to test behaviour


arangod/Cluster/Maintenance.cpp
```diff
@@ -320,7 +320,6 @@ void handlePlanShard(StorageEngine& engine, uint64_t planIndex, VPackSlice const
             {DATABASE, dbname},
             {COLLECTION, colname},
             {SHARD, shname},
-            {THE_LEADER, std::string()},
```
neunhoef, line 323: It turned out that this was no longer needed. Cleanup done as an aside. See TakeoverShardLeadership action.

neunhoef: It was always empty anyway.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -416,7 +417,7 @@ arangodb::Result SynchronizeShard::getReadLock(
   // nullptr only happens during controlled shutdown
   if (pool == nullptr) {
     return arangodb::Result(TRI_ERROR_SHUTTING_DOWN,
-                            "cancelReadLockOnLeader: Shutting down");
+                            "getReadLock: Shutting down");
```
neunhoef, line 420: This was a wrong error message, fixed as an aside.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -913,10 +924,6 @@ bool SynchronizeShard::first() {
         return false;
       }
 
-      // This is necessary to accept replications from the leader which can
```
neunhoef, line 916: It turns out that this can and must be set even later, namely, when we have got the exclusive lock (aka "hard read lock"), when we know the new `followingTermId`.

arangod/Cluster/TakeoverShardLeadership.cpp
```diff
@@ -86,11 +86,6 @@ TakeoverShardLeadership::TakeoverShardLeadership(MaintenanceFeature& feature,
     error << "database and shard must be specified. ";
   }
 
-  if (!desc.has(THE_LEADER)) {
```
neunhoef, line 89: This was old case which was not used anymore.

arangod/Cluster/TakeoverShardLeadership.cpp
```diff
@@ -112,23 +107,14 @@ static void sendLeaderChangeRequests(network::ConnectionPool* pool,
...
-                                     ShardID const& shardID, std::string const& oldLeader) {
+                                     LogicalCollection& shard, std::string const& oldLeader) {
   if (pool == nullptr) {
     // nullptr happens only during controlled shutdown
     return;
   }
 
   std::string const& sid = ServerState::instance()->getId();
 
-  VPackBufferUInt8 buffer;
```
neunhoef, line 123: Has been moved into the loop to make it recipient specific.

arangod/Cluster/TakeoverShardLeadership.cpp
```diff
@@ -143,6 +129,17 @@ static void sendLeaderChangeRequests(network::ConnectionPool* pool,
...
     }
+    uint64_t followingTermId = shard.followers()->newFollowingTermId(srv);
+    VPackBufferUInt8 buffer;
+    VPackBuilder bodyBuilder(buffer);
+    {
+      VPackObjectBuilder ob(&bodyBuilder);
+      bodyBuilder.add("leaderId", VPackValue(sid));
+      bodyBuilder.add("oldLeaderId", VPackValue(oldLeader));
+      bodyBuilder.add("shard", VPackValue(shard.name()));
+      bodyBuilder.add(StaticStrings::FollowingTermId, VPackValue(followingTermId));
```
neunhoef, line 140: This is the actual change here.

arangod/Cluster/TakeoverShardLeadership.cpp
```diff
@@ -165,86 +162,61 @@ static void sendLeaderChangeRequests(network::ConnectionPool* pool,
 }
 
 static void handleLeadership(uint64_t planIndex, LogicalCollection& collection,
-                             std::string const& localLeader, std::string const& plannedLeader,
+                             std::string const& localLeader,
                              std::string const& databaseName,
                              MaintenanceFeature& feature) {
   auto& followers = collection.followers();
 
-  if (plannedLeader.empty()) {   // Planned to lead
```
neunhoef, line 173: The corresponding else never happened, so it was removed in this change.

neunhoef (commented):
Some comments for the convenience of the reviewer.

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,30 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
...
+  // We want the random number to be non-zero:
+  do {
+    i = RandomGenerator::interval(UINT64_MAX);
+  } while (i == 0);
+  try {
+    _followingTermId[s] = i;
+  } catch(std::bad_alloc const& exc) {
+    i = 1;   // I assume here that I do not get bad_alloc if the key is
+             // already in the map, since it then only has to overwrite
+             // an integer, if the key is not in the map, we default to 0.
```
jsteemann, line 558: The comment here says that we default to `0`, but we seem to default to `1` in reality.

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,30 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
...
+  }
+  return i;
+}
+
+uint64_t FollowerInfo::getFollowingTermId(ServerID const& s) noexcept {
```
jsteemann, line 563:
```suggestion
uint64_t FollowerInfo::getFollowingTermId(ServerID const& s) const noexcept {
```

arangod/Cluster/FollowerInfo.h
```diff
@@ -54,6 +54,20 @@ class FollowerInfo {
   // soon as we can guarantee at least so many followers locally.
   std::shared_ptr<std::vector<ServerID>> _failoverCandidates;
 
+  // The following map holds a random number for each follower, this
+  // random number is sent given to the follower when it gets in sync
```
jsteemann, line 58:
```suggestion
  // random number is sent to the follower when it gets in sync
```

arangod/Cluster/FollowerInfo.h
```diff
@@ -135,6 +149,28 @@ class FollowerInfo {
...
+  //////////////////////////////////////////////////////////////////////////////
+  /// @brief for each run of the "get-in-sync" protocol we generate a
+  /// random number to identify this "following term". This is created
+  /// when the follower fetches the exclusive lock to finally get in sync
+  /// and is stored in _followingTermId, so that it can be forwarded with
+  /// each synchronous replication request. The follower can then decline
+  /// the replication in case it is not "in the same term".
+  //////////////////////////////////////////////////////////////////////////////
+
+  uint64_t getFollowingTermId(ServerID const& s) noexcept;
```
jsteemann, line 172:
```suggestion
  uint64_t getFollowingTermId(ServerID const& s) const noexcept;
```

arangod/Cluster/CreateCollection.cpp
```diff
@@ -165,7 +165,7 @@ bool CreateCollection::first() {
         std::vector<std::string> noFollowers;
         col->followers()->takeOverLeadership(noFollowers, nullptr);
       } else {
-        col->followers()->setTheLeader(leader);
+        col->followers()->setTheLeader(LEADER_NOT_YET_KNOWN);
```
neunhoef, line 168: It was a very bad idea to create a shard which is not yet in sync yet with the leader already configured.

arangod/Cluster/FollowerInfo.h
neunhoef: Done.

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,29 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
   }
   return newValue;
 }
+
+uint64_t FollowerInfo::newFollowingTermId(ServerID const& s) noexcept {
+  uint64_t i = 0;
+  // We want the random number to be non-zero:
+  do {
+    i = RandomGenerator::interval(UINT64_MAX);
```
jsteemann, line 551: one suggestion: should we compare `i` against the value of `_followingTermId[s]` here to prevent the RNG to create the same random number twice in a row?

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,29 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
...
+    i = 1;   // I assume here that I do not get bad_alloc if the key is
+             // already in the map, since it then only has to overwrite
+             // an integer, if the key is not in the map, we default to 1.
+  }
+  return i;
+}
+
+uint64_t FollowerInfo::getFollowingTermId(ServerID const& s) const noexcept {
+  // Note that we assume that find() does not throw!
+  auto it = _followingTermId.find(s);
```
jsteemann, line 565: I think this function can be called by multiple threads (for multiple followers) concurrently, so `_followingTermId` needs to be access-protected by either a mutex or an RW-lock.

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,29 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
...
 }
+
+uint64_t FollowerInfo::newFollowingTermId(ServerID const& s) noexcept {
+  uint64_t i = 0;
+  // We want the random number to be non-zero:
+  do {
+    i = RandomGenerator::interval(UINT64_MAX);
+  } while (i == 0);
+  try {
+    _followingTermId[s] = i;
```
jsteemann, line 554: I think this method can be called by multiple threads (for multiple followers) concurrently, so `_followingTermId` needs to be access-protected by either a mutex or an RW-lock.

arangod/Cluster/FollowerInfo.cpp
neunhoef: Indeed. will fix.

arangod/Cluster/FollowerInfo.cpp
neunhoef: This is an interesting one. I have gone left and right on this one, without a clear preference. I will add this, and add a mutex to protect this map at the same time. Looking at this again it struck my eye that this is unprotected traffic.

arangod/Cluster/FollowerInfo.cpp
neunhoef: Indeed, will fix.

arangod/Cluster/FollowerInfo.cpp
neunhoef: Done

arangod/Cluster/FollowerInfo.cpp
neunhoef: Done.

arangod/Cluster/FollowerInfo.cpp
neunhoef: Done.

jsteemann (approved):
LGTM

arangod/Cluster/FollowerInfo.cpp
```diff
@@ -542,3 +543,29 @@ VPackBuilder FollowerInfo::newShardEntry(VPackSlice oldValue) const {
...
+  }
+  return i;
+}
+
+uint64_t FollowerInfo::getFollowingTermId(ServerID const& s) const noexcept {
+  // Note that we assume that find() does not throw!
+  auto it = _followingTermId.find(s);
+  if (it == _followingTermId.end()) {
+    // If not found, we use the default from above:
+    return 1;
```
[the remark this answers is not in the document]
neunhoef: This is a fallback case for bad_alloc situations. Note that both `newFollowingTermId` and `getFollowingTermId` are `noexcept`. This particular case here is needed to make `newFollowingTermId` not fail in a bad_alloc situation and then return here a consistent ID, such that things still work (albeit giving worse protection against delayed requests). So this is intentional.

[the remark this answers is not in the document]
mpoeter: Got it. So this should only ever happen if we previously failed to insert the term id due to a bad_alloc situation, in which case `newFollowingTermId` already returned 1, so for consistency we also return 1 here.

PR #14408, opened 2021-06-28
Change update behavior for empty updates

jsteemann:
### Scope & Purpose

Change update behavior for empty updates

* Removed a special case for empty document update operations (i.e. update
  requests in which no attributes were specified to be updated) were handled
  in a special way without performing any writes. The problem was that such
  updates did not update the local state, but could have been replicated to
  followers.
  This special empty update case is now removed and update operations that do
  not update any attributes are treated as normal write operations both
  locally and in the replication.

#### Backports:

- [x] No backports required

### Testing & Verification


neunhoef (approved):
LGTM

PR #14411, opened 2021-06-28
Fix transactions with dropped followers

jsteemann:
### Scope & Purpose

Fix handling of dropped followers with transactions:

  end, because the follower may be responsible  for _other_ shards as well.

#### Backports:

- [x] No backports required

### Testing & Verification


neunhoef (approved):
LGTM

PR #14432, opened 2021-06-30
Feature/hybrid smart graph testsuite extended

hkernbach:
### Scope & Purpose

Modified Mocks for more flexibility in our CPP testing environment.
Mostly EE relevant changes, see https://github.com/arangodb/enterprise/pull/699 for more details.


tests/Mocks/Servers.cpp
```diff
@@ -638,24 +625,20 @@ std::shared_ptr<LogicalCollection> MockClusterServer::createCollection(
...
-  }
-  LogicalCollection dummy(*vocbase, props.slice(), true);
 
-  auto shards = std::make_shared<ShardMap>();
-  for (auto const& [shard, server] : shardNameToServerNamePairs) {
-    shards->emplace(shard, std::vector<ServerID>{server});
+    if (additionalProperties.isObject()) {
+      for (auto it : VPackObjectIterator(additionalProperties)) {
+        props.add(it.key.copyString(), it.value.value());
+      }
```
mchacki, line 632: the injected additionalProperties could have a key that is already in props. (e.g. DataSourceName).
There also is a helper method that can add an ObjectIterator:

```suggestion
      props.add(VPackObjectIterator(additionalProperties));
```
this will throw an error on duplicate key.

I think the original version does work, but generates an invalid JSON.
I also think it will overwrite the original value, but I am not sure about this, also I think we never want a value (new or old) to silently be ignored.

mchacki (approved):
LGTM.

I have a minor suggestion.

PR #14450, opened 2021-07-02
fix locking of write transactions on DB servers

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/701

Fix the locking of AQL query snippets on DB servers.
Previously, the transaction for an AQL write query was locked on DB servers in read mode, which could have led to problems when sending two write AQL queries to the same streaming transaction.
There is also a potential race between the async query shutdown being in progress and a client application sending the next write query to the same transaction.

#### Backports:

- [x] Backports required for: *3.8*, *3.7*

#### Related Information

- [x] Enterprise repository PR: https://github.com/arangodb/enterprise/pull/701

### Testing & Verification


arangod/RocksDBEngine/RocksDBTransactionState.h
```diff
@@ -223,10 +233,29 @@ class RocksDBTransactionState final : public TransactionState {
...
+
   /// @brief if true there key buffers will no longer be shared
   bool _parallel;
 };
 
+/// @brief a struct that makes sure that the same RocksDBTransactionState
+/// is not used by different write operations in parallel. will only do
+/// something in maintainer mode, and do nothing in release mode!
+struct RocksDBTransactionStateGuard {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
```
mchacki, line 248: I like this guard :+1:

mchacki (approved):
LGTM, thanks for this PR

mpoeter (approved):

arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
```diff
@@ -502,16 +518,37 @@ Result EngineInfoContainerDBServerServerBased::buildEngines(
...
       // So we know that all Transactions are aborted.
-      // We do NOT care for the actual result.
-      futures::collectAll(requests).wait();
-      snippetIds.clear();
+      Result res;
+      for (auto& tryRes : requests) {
+        network::Response const& response = tryRes.get(); 
+        if (response.fail()) {
+          // note first error, but continue iterating over all results
+          res.reset(response.combinedResult());
```
mpoeter, line 532: We are effectively loosing information since we only report the last error, so I would like to add some debug/trace logging here.

mpoeter (approved):

arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
jsteemann: Ok, added.

PR #14455, opened 2021-07-03
Feature/hybrid smart graph test setup helper

mchacki:
Intermediate Merge.
This improves the MockServers, to support Indexes on Coordinators.
For some reason activating this causes some tests to get into assertions, so we made it explicit to activate the ClusterIndexes.
This assertion can be solved by "Mocking" the creation on MockStorageEngine as well, and not reuse the Production index Code


arangod/Aql/EngineInfoContainerDBServerServerBased.h
```diff
@@ -56,6 +56,9 @@ class QuerySnippet;
 
 class EngineInfoContainerDBServerServerBased {
  private:
+  // TODO Temporary. Do not check in!!
```
hkernbach, line 59: Open / outdated todo?

arangod/InternalRestHandler/InternalRestTraverserHandler.cpp
```diff
@@ -32,6 +32,7 @@
 #include "Rest/GeneralResponse.h"
 #include "Transaction/StandaloneContext.h"
 
+#include <Logger/LogMacros.h>
```
hkernbach, line 35: Think this can be removed now.

tests/Mocks/Servers.cpp
```diff
@@ -522,6 +530,36 @@ void MockClusterServer::startFeatures() {
...
+  auto query = std::make_unique<arangodb::aql::Query>(
+      arangodb::transaction::StandaloneContext::Create(getSystemDatabase()),
+      fakeQueryString, bindParams, queryOptions.slice());
+  callback(*query);
+  query->prepareQuery(aql::SerializationFormat::SHADOWROWS);
+
+  //  auto engine =
+  //      std::make_unique<aql::ExecutionEngine>(*query, aql::SerializationFormat::SHADOWROWS);
+  //  query->setEngine(std::move(engine));
+
```
hkernbach, line 559:
```suggestion
```

unused code. If this was meant for the future, please ignore.

tests/Mocks/Servers.h
```diff
@@ -58,7 +58,13 @@ namespace mocks {
 /// @brief mock application server with no features added
 class MockServer {
  public:
-  MockServer();
+  // Note, setting injectClusterIndexes causes the "create" methods to fail.
+  // this is all hardly worked around the Cluster engine and needs a proper
+  // clean up. It is highly recommended to not set injectClusterIndexes unless
+  // you want to specificly test something that selects an index, but cannot use
+  // it. Use with care for now.
```
hkernbach, line 65: IDK if possible, but could we somehow print a warning for developers, which might not read that HINT? 
If not (e.g. jenkins is not happy), ignore for now. But could help in case any developer might skip that section.

hkernbach: Ah, I think it can be ignored as it is false by default.

hkernbach (changes requested):

arangod/Aql/EngineInfoContainerDBServerServerBased.h
mchacki: Keep, we need to remove this on BaseBranch -> Devel merge

PR #14458, opened 2021-07-05
Following term id for 3.8

neunhoef:
Synchronous replication requests (for insert, replace, update, delete
and truncate operations) have been checked on the follower in so far,
that it was checked that they come from the correct leader (the one the
follower believes to be following). However, it can in rare cases happen
that a follower is dropped and a replication request from before the
dropping only hits the follower after it has gotten in sync again.
This is very bad since it can lead to inconsistent data on shard
replicas which are deemed to be "in sync" and thus equal.

This PR attacks this problem: Whenever a leader/follower relationship is
established, the leader invents a random number and sends it to the
follower together with the confirmation that it has aqcuired an
exclusive lock on the shard (one relative late step of the get-in-sync
protocol). The leader stores this number and sends it along with each
synchronous replication request. The follower stores this number, too,
and checks that a synchronous replication requests comes not only from
the right leader, but also from the right "term" and not from some
distant past term.

The method described is relatively straightforward, the implementation
follows exactly the plan outlined above. One caveat is that there is
also the case of controlled leadership handover. In this case, we need
to play the same game and establish a random number identifying the
term.

Note that these numbers are not persisted (not necessary!), therefore we
prefer to use random numbers as opposed to an ascending counter, since
then we could have a certain possibility to reach the same counter value
after a restart.
Scope & Purpose

    [*] hankey Bugfix (requires CHANGELOG entry)

Backports:

    No backports required
    [*] This is the backport for 3.8, probably, 3.7 will be done as
    well. Original PR: https://github.com/arangodb/arangodb/pull/14405

Testing & Verification

    [*] This change is already covered by existing tests, such as all
    tests bringing shards in sync. The new chaos test which delays
    synchronous replication requests artificially.
    [*] This PR adds tests that were used to verify the changes:
        `tests/js/client/shell/shell-following-term-id-cluster.js`


jsteemann (approved):
LGTM

Issue #14431, opened 2021-06-29
Query timeouts & coordinator crash


CMogilko:
## My Environment

* __ArangoDB Version__: 3.7.11
* __Storage Engine__: RocksDB
* __Deployment Mode__: Cluster
* __Deployment Strategy__: Manual Start in Docker
* __Configuration__: 5 dbserver, 5 agency on 5 separate server, 8 coordiators on 8 separate servers
* __Infrastructure__: own
* __Operating System__: centos7
* __Total RAM in your machine__: 120gb
* __Disks in use__: ssd nvme
* __Used Package__: Docker - official Docker library

## Component, Query & Data

__Affected feature__:
github.com/arangodb/go-driver v0.0.0-20210621075908-e7a6fa0cbd18 (master)

__AQL query (if applicable)__:
Find group IDs for given entity ID:
```
WITH entities
FOR v, e IN 1..100 INBOUND @id membership OPTIONS { bfs: true, uniqueVertices : 'global' }
    RETURN v
```

__Dataset__:
100k edges: 1 entity to 1 group

__Size of your Dataset on disk__:
300 MB

__Replication Factor & Number of Shards (Cluster only)__:
Entities: 1 shard, 5 replicas, 3 write concern
Membership: 1 shard, 5 replicas, 3 write concern

## Steps to reproduce

1. Run load testing with only 1500 RPS this query
2. Get 30% errors `i/o timeout` in clients
3. Get SIGSEGV in a coordinator

__Problem__:
I was running a load testing with simple graph query which returns user's group ids only in a single collection in the powerful cluster configuration. But on a relatively low rps gets timeouts with very low cpu usage (1% on 16 cores) and disk utilization (1%).
Then coordinator server just crashed segfault.

```
[...]

PR #14459, opened 2021-07-05
Fix potential segfault in tcp/ssl socket handling.

mpoeter:
### Scope & Purpose

#### Backports:

- [x] No backports required (devel and 3.8 already have this fix)

#### Related Information

Potentially fixes this issue: https://github.com/arangodb/arangodb/issues/14431

### Testing & Verification

I first fixed this issued in https://github.com/arangodb/arangodb/pull/14029/commits/da1e761c7443c34204ea7e1e5be66d3babb454aa after TSan reported a data race. From the race, I figured that it would be possible to cause a segfault, but unfortunately I was never able to trigger any. Thus there is not additional test to cover this.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,9 +1,10 @@
 v3.7.13 (XXXX-XX-XX)
 --------------------
 
-* Fix potential segfault in tcp/ssl socket handling.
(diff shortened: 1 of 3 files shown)


jsteemann (approved):
LGTM

PR #14467, opened 2021-07-05
APM-112: prevent invalid use of OPTIONS in AQL queries

jsteemann:
### Scope & Purpose

APM-112: invalid use of OPTIONS in AQL queries will now raise a warning in the query. 
The feature is useful to detect misspelled attribute names in OPTIONS, e.g.
  
    INSERT ... INTO collection 
      OPTIONS { overwrightMode: 'ignore' } /* should be 'overwriteMode' */

It is also useful to detect the usage of valid OPTIONS attribute names that are used for a wrong query part, e.g.
  
    FOR doc IN collection 
      FILTER doc.value == 1234 
      INSERT doc INTO other 
        OPTIONS { indexHint: 'myIndex' } /* should be used above for FOR */

In case a wrong option attribute is used, a warning with code 1575 will be raised. 
By default, warnings are reported but do not lead to the query being aborted. 
This can be toggled by the startup option `--query.fail-on-warnings` or the per-query runtime option `failOnWarnings`.

The addition of warnings for wrong OPTIONS is a slight behavior change (although for good), so it can lead to some incompatibilities.

#### Backports:

TO BE DECIDED!

- [ ] No backports required

#### Related Information

- [x] Docs PR: TO BE DONE
- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/APM-112

### Testing & Verification


lib/Basics/errors.dat
```diff
@@ -251,7 +251,7 @@ ERROR_QUERY_FULLTEXT_INDEX_MISSING,1571,"no suitable fulltext index found for fu
 ERROR_QUERY_INVALID_DATE_VALUE,1572,"invalid date value","Will be raised when a value cannot be converted to a date."
 ERROR_QUERY_MULTI_MODIFY,1573,"multi-modify query","Will be raised when an AQL query contains more than one data-modifying operation."
 ERROR_QUERY_INVALID_AGGREGATE_EXPRESSION,1574,"invalid aggregate expression","Will be raised when an AQL query contains an invalid aggregate expression."
-ERROR_QUERY_COMPILE_TIME_OPTIONS,1575,"query options must be readable at query compile time","Will be raised when an AQL data-modification query contains options that cannot be figured out at query compile time."
+ERROR_QUERY_INVALID_OPTIONS,1575,"query options must be readable at query compile time","Will be raised when an AQL query contains OPTIONS that cannot be figured out at query compile time or when an unknown attribute is used inside OPTIONS."
```
maierlars, line 254: If `"query options must be readable at query compile time"` is what the user sees when the error is created, then we need a better message.

maierlars: Is it okay to just change an existing error code? Maybe we should just allocate a new one?

maierlars (approved):

maierlars (approved):

PR #14475, opened 2021-07-07
Bug fix/non recursive plan handling

mpoeter:
### Scope & Purpose

Adapt various places to make handling of execution plans non-recursive in order to avoid stack overflows.

Before these changes, on my Windows machine (which has a default stack size of 1MB) I could handle queries with up to 1000 nodes. After these changes, I can run queries with up to 7000 nodes. More than that is still not possible, because `ExecutionBlock::initializeCursor` is still recursive and larger queries will again run into a stack overflow.

I made some basic tests that showed that performance of the non-recursive version is slightly better than before. In order to test this, I created some huge queries with thousands of variables, which revealed a performance issue in RegisterPlan where `std::find` was used on a `HashSet` which could be changed to a simple `HashSet::contains`, so this PR also massively improves register planning performance for large queries.

One thing I noted is that the detailed query plan required by `_explain` grows quadratic in the number of variables. This is because each node has the `varInfoList` and `varsUsedLaterStack` lists, both of which grow linear with the number of variables in the query. While this can probably be improved, it is not critical. However, it means that even though we can _execute_ such large queries, they can effectively not be `_explain`ed.

In order to prevent stack overflows and issues with the size explosion in `_explain`, we should limit the number of query nodes, and also the expression depths in calculation nodes. But this should be discussed and implemented in a separate PR.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/APM-77

### Testing & Verification

During development I had test code in place to ensure that the old and new implementation behave identically (e.g., that `walk` visits all nodes in the exact same order, etc.).

I have adapted some tests to use an even larger number of nodes than before (previously these numbers would have resulted in a crash).

### DIFF
--- a/arangod/Aql/ClusterNodes.cpp
+++ b/arangod/Aql/ClusterNodes.cpp
@@ -302,7 +302,7 @@ std::unique_ptr<ExecutionBlock> DistributeNode::createBlock(
                                                                   std::move(infos));
 }
 
-/// @brief doToVelocyPack, for DistributedNode
+/// @brief doToVelocyPack, for DistributeNode
 void DistributeNode::doToVelocyPack(VPackBuilder& builder, unsigned flags) const {
   // call base class method
   ScatterNode::doToVelocyPack(builder, flags);
--- a/arangod/Aql/ClusterNodes.cpp
+++ b/arangod/Aql/ClusterNodes.cpp
@@ -153,18 +153,14 @@ std::unique_ptr<ExecutionBlock> RemoteNode::createBlock(
                                                               queryId());
 }
 
-/// @brief toVelocyPack, for RemoteNode
-void RemoteNode::toVelocyPackHelper(VPackBuilder& nodes, unsigned flags,
-                                    std::unordered_set<ExecutionNode const*>& seen) const {
+/// @brief doToVelocyPack, for RemoteNode
+void RemoteNode::doToVelocyPack(VPackBuilder& nodes, unsigned flags) const {
   // call base class method
-  DistributeConsumerNode::toVelocyPackHelperInternal(nodes, flags, seen);
+  DistributeConsumerNode::doToVelocyPack(nodes, flags);
 
   nodes.add("database", VPackValue(_vocbase->name()));
   nodes.add("server", VPackValue(_server));
   nodes.add("queryId", VPackValue(_queryId));
-
-  // And close it:
-  nodes.close();
 }
 
 /// @brief estimateCost
@@ -202,16 +198,10 @@ std::unique_ptr<ExecutionBlock> ScatterNode::createBlock(
                                                                std::move(executorInfos));
 }
 
-/// @brief toVelocyPack, for ScatterNode
-void ScatterNode::toVelocyPackHelper(VPackBuilder& nodes, unsigned flags,
-                                     std::unordered_set<ExecutionNode const*>& seen) const {
-  // call base class method
-  ExecutionNode::toVelocyPackHelperGeneric(nodes, flags, seen);
-
+/// @brief doToVelocyPack, for ScatterNode
+void ScatterNode::doToVelocyPack(VPackBuilder& nodes, unsigned flags) const {
   // serialize clients
   writeClientsToVelocyPack(nodes);
-  // And close it:
-  nodes.close();
 }
 
 bool ScatterNode::readClientsFromVelocyPack(VPackSlice base) {
@@ -312,23 +302,15 @@ std::unique_ptr<ExecutionBlock> DistributeNode::createBlock(
                                                                   std::move(infos));
 }
 
-/// @brief toVelocyPack, for DistributedNode
-void DistributeNode::toVelocyPackHelper(VPackBuilder& builder, unsigned flags,
-                                        std::unordered_set<ExecutionNode const*>& seen) const {
+/// @brief doToVelocyPack, for DistributedNode
+void DistributeNode::doToVelocyPack(VPackBuilder& builder, unsigned flags) const {
   // call base class method
-  ExecutionNode::toVelocyPackHelperGeneric(builder, flags, seen);
-
+  ScatterNode::doToVelocyPack(builder, flags);
   // add collection information
   CollectionAccessingNode::toVelocyPack(builder, flags);
 
-  // serialize clients
-  writeClientsToVelocyPack(builder);
-
   builder.add(VPackValue("variable"));
-  _variable->toVelocyPack(builder);;
-
-  // And close it:
-  builder.close();
+  _variable->toVelocyPack(builder);
 }
 
 void DistributeNode::replaceVariables(std::unordered_map<VariableId, Variable const*> const& replacements) {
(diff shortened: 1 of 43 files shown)


arangod/Aql/ModificationNodes.cpp
```diff
@@ -55,13 +55,10 @@ ModificationNode::ModificationNode(ExecutionPlan* plan, arangodb::velocypack::Sl
...
 /// @brief toVelocyPack
-void ModificationNode::toVelocyPackHelper(VPackBuilder& builder, unsigned flags,
-                                          std::unordered_set<ExecutionNode const*>& seen) const {
-  // call base class method
-  ExecutionNode::toVelocyPackHelperGeneric(builder, flags, seen);
-
+void ModificationNode::doToVelocyPack(VPackBuilder& builder, unsigned flags) const {
   // add collection information
   CollectionAccessingNode::toVelocyPack(builder, flags);
+  CollectionAccessingNode::toVelocyPackHelperPrimaryIndex(builder);
```
jsteemann, line 61: Adding the `toVelocyPackHelperPrimaryIndex` here is a change compared to the previous plan output.
Previously, the `toVelocyPackHelperPrimaryIndex` method was called when serializing REMOVE, UPDATE, REPLACE and UPSERT, but it was not called for INSERT.
I don't know if it does any harm to add this for INSERT nodes as well, just wondering.

mpoeter: Previously every derived node called `ModificationNode::toVelocyPackHelperPrimaryIndex`. This is now streamlined by moving this call here. This actually allow me to remove the overrides in `UpdateNode` and `ReplaceNode` since these don't add any additional data.

arangod/Aql/ClusterNodes.cpp
```diff
@@ -312,23 +302,15 @@ std::unique_ptr<ExecutionBlock> DistributeNode::createBlock(
                                                                   std::move(infos));
 }
 
-/// @brief toVelocyPack, for DistributedNode
-void DistributeNode::toVelocyPackHelper(VPackBuilder& builder, unsigned flags,
-                                        std::unordered_set<ExecutionNode const*>& seen) const {
+/// @brief doToVelocyPack, for DistributedNode
```
jsteemann, line 305:
```suggestion
/// @brief doToVelocyPack, for DistributeNode
```

arangod/Aql/ExecutionNode.cpp
```diff
@@ -516,20 +516,24 @@ ExecutionNode::ExecutionNode(ExecutionPlan& plan, ExecutionNode const& other)
...
+/// i.e., the upmost dependency will be the first, and this node will be the
+/// last in the array.
+void ExecutionNode::allToVelocyPack(VPackBuilder& builder, unsigned flags) const {
+
+  struct NodeSerializer : WalkerWorker<ExecutionNode, WalkerUniqueness::Unique> {
+    NodeSerializer(VPackBuilder& builder, unsigned flags)
+        : builder(builder), flags(flags) {}
+    virtual void after(ExecutionNode* n) {
+      n->toVelocyPack(builder, flags);
+    };
```
jsteemann, line 530:
```suggestion
    void after(ExecutionNode* n) override {
      n->toVelocyPack(builder, flags);
    }
```

arangod/Aql/ModificationNodes.cpp
mpoeter: I should have read the comment more carefully..! Your are right, the `InsertNode` previously did not call `toVelocyPackHelperPrimaryIndex`, while the new implementation _does_ call it, so this is indeed a change in behavior.
`toVelocyPackHelperPrimaryIndex` adds a field `index` with an array containing information about the primary index. However, I don't even know if or where this information is used. At least loading the nodes from VPack does not read this information.

arangod/Aql/ExecutionNode.h
```diff
@@ -487,6 +486,15 @@ class ExecutionNode {
   auto getRegsToKeepStack() const -> RegIdSetStack;
 
  protected:
+  /// @brief serialize this ExecutionNode to VelocyPack
+  void toVelocyPack(arangodb::velocypack::Builder&, unsigned flags) const;
```
mchacki, line 490: For Logging / Testing / Debugging Purposes we could even make this Method Public, as it will Serialize a single Node standalone (it cannot make sense if its dependencies but everything else should be self-contained)

arangod/Aql/ModificationNodes.cpp
mchacki: i tend to agree @mpoeter here.
The "indexes" seems to be never read by any of the Nodes, also I have not seen any access to "indexes" in the RestHandlers (just Collection Methods)

The GraphNode seems to generate an "indexes" field as well, which is never read. Only the IndexNode seems to read it.

At this Code position it can also not be there for Backwards Compatibility, as this is coordinators, and they are upgrade last, so they do not talk to older servers.

mchacki (changes requested):
I very much like this PR.

i have a couple of minor things in the comments and there is an ongoing discussion still.

arangod/Aql/ExecutionNode.cpp
```diff
@@ -791,167 +811,146 @@ ExecutionNode const* ExecutionNode::getLoop() const {
...
-  // If this assert triggers we have created a loop.
-  // There is a dependency path that leads back to this node
-  TRI_ASSERT(seen.find(this) == seen.end());
-  seen.emplace(this);
-  nodes.openObject();
-  nodes.add("type", VPackValue(getTypeString()));
+/// @brief serialize this ExecutionNode to VelocyPack
+/// Note: The builder has to be an Array Element that is still Open.
+void ExecutionNode::toVelocyPack(arangodb::velocypack::Builder& builder, unsigned flags) const {
+  TRI_ASSERT(builder.isOpenArray());
```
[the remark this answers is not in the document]
mpoeter: You are right, this is not a requirement anymore - will remove the assert and the note.

arangod/Aql/ExecutionNode.h
mpoeter: Yes, we might have ambiguities with `CollectionAccessingNode::toVelocyPack`, `DocumentProducingNode::toVelocyPack` and possibly others, but these can be sorted out. This should also simplify some unit tests. Will do, and also add a dedicated test for `allToVelocyPack`.

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2061,21 +2031,10 @@ SubqueryNode::SubqueryNode(ExecutionPlan* plan, arangodb::velocypack::Slice cons
...
-  _outVariable->toVelocyPack(nodes);
-
-  nodes.add("isConst", VPackValue(const_cast<SubqueryNode*>(this)->isConst()));
-
-  // And add it:
-  nodes.close();
+/// @brief doToVelocyPack, for SubqueryNode
+void SubqueryNode::doToVelocyPack(VPackBuilder& nodes, unsigned flags) const {
+  // since we have spliced subqueries this should never be called!
+  TRI_ASSERT(false);
```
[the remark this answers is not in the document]
mpoeter: Done

arangod/Aql/ModificationNodes.cpp
mpoeter: It looks like these "indexes" are not used anywhere and can be removed entirely. However, we decided to keep the current version for now and address this in a separate PR.

mchacki (approved):
LGTM! Thanks

PR #14484, opened 2021-07-08
Replicated Logs Companion PR

maierlars:
### Scope & Purpose

This contains changes from #14415, that are not directly related to replicated logs. This is done to make the actual PR smaller and ease the review process.

#### Backports:

- [x] No backports required


maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16131/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16132/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16134/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16138/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16140/

arangod/Agency/AgencyComm.cpp
```diff
@@ -177,13 +177,54 @@ AgencyOperation::AgencyOperation(std::string const& key,
   _opType.value = opType;
 }
 
+AgencyOperation::AgencyOperation(std::string const& key, AgencyValueOperationType opType,
+                                 std::shared_ptr<velocypack::Builder> value)
+    : AgencyOperation(std::move(key), opType, value->slice()) {
```
jsteemann, line 182: I don't think that `key` can be moved here, as it is `const`.

jsteemann: I also think this calls the base constructor, which will do a full copy of the VPack slice in `value` into a newly allocated shared_ptr `_holder`, only to overwrite `_holder` later. So `_value` may point into the old `_holder` and it could be invalid.

goedderz: Right, thanks, the move is useless. I think the rest is correct though. This delegates to the constructor `AgencyOperation(std::string const&, AgencyValueOperationType, Slice)`, which does not copy the slice. Only the constructors getting some `T` which is then passed to `VPackValue` do that. I've added assertions to make that clear.

arangod/Agency/AgencyComm.cpp
```diff
@@ -177,13 +177,54 @@ AgencyOperation::AgencyOperation(std::string const& key,
...
+    : _key(path->str()), _opType(), _value(value) {
+  _opType.type = AgencyOperationType::Type::VALUE;
+  _opType.value = opType;
+}
+
+AgencyOperation::AgencyOperation(std::shared_ptr<cluster::paths::Path const> const& path,
+                                 AgencyValueOperationType opType,
+                                 std::shared_ptr<velocypack::Builder> value)
+    : AgencyOperation(path, opType, value->slice()) {
+  _holder = std::move(value);
```
jsteemann, line 216: I don't think this is safe, as `_value` will point into the slice held by the old version of `_holder`, which is overriden here.

goedderz: As above, it should be okay with the delegating constructor. I've also added an assertion here now to clarify that this works as intended.

jsteemann: Ok, thanks. Too many constructors here...

arangod/Agency/AgencyComm.cpp
```diff
@@ -177,13 +177,56 @@ AgencyOperation::AgencyOperation(std::string const& key,
...
+                                 AgencySimpleOperationType opType)
+    : _key(path->str()), _opType() {
+  _opType.type = AgencyOperationType::Type::SIMPLE;
+  _opType.simple = opType;
+}
+
+AgencyOperation::AgencyOperation(std::shared_ptr<cluster::paths::Path const> const& path,
+                                 AgencyValueOperationType opType,
+                                 velocypack::Slice const value)
+    : _key(path->str()), _opType(), _value(value) {
```
jsteemann, line 208: This constructor is potentially unsafe to use. It only works if the lifetime of the `Slice` in value exceeds the lifetime of the AgencyOperation.
Should we make this ctor protected, or at least add a code comment that indicates that?

arangod/Agency/AgencyComm.cpp
```diff
@@ -177,13 +177,56 @@ AgencyOperation::AgencyOperation(std::string const& key,
...
+    : AgencyOperation(path, opType, value->slice()) {
+  _holder = std::move(value);
+  TRI_ASSERT(_holder->slice().start() == _value.start());
+}
+
+AgencyOperation::AgencyOperation(std::shared_ptr<cluster::paths::Path const> const& path,
+                                 AgencyValueOperationType opType,
+                                 velocypack::Slice const newValue,
+                                 velocypack::Slice const oldValue)
+    : _key(path->str()), _opType(), _value(newValue), _value2(oldValue) {
```
jsteemann, line 225: Same here. This is easy to misuse. We should at least add a code comment saying that the lifetime of `newValue` and `oldValue` must exceed the lifetime of the AgencyOperation here.

arangod/Agency/AgencyComm.cpp
goedderz: I agree. Note that these constructors (initializing from a Slice, but not copying it) are the oldest and possibly most used ones, so we can't just protect them without rewriting some code. I've added some comments now. I'll see if I can spare some time to clear all this up, but rather not in this PR if that's fine with you.

lib/Basics/Exceptions.h
```diff
@@ -35,43 +35,89 @@
...
+
+#define ASSERT_OR_THROW_ARANGO_EXCEPTION_MESSAGE(...) \
+  THROW_AND_CATCH_(THROW_ARANGO_EXCEPTION_MESSAGE, __VA_ARGS__)
+
+#define ASSERT_OR_THROW_ARANGO_EXCEPTION_IF_FAIL(...) \
+  THROW_AND_CATCH_(THROW_ARANGO_EXCEPTION_IF_FAIL, __VA_ARGS__)
+
+#else
+
+#define ASSERT_OR_THROW_ARANGO_EXCEPTION(...) \
```
jsteemann, line 107: I am not really a big fan of these macros.
Do I understand correctly that in maintainer mode, they will throw an exception and terminate the program (but not in case no exception occurs...?).
And in release mode they will simply throw, but not end the program?
The macro names indicate that it's going to be either an assertion or a throw here, but I am wondering what the use case is for these macros? IMO we shouldn't add more if not absolutely required. Maybe there is a use case for such macros, but I don't see it yet.

lib/Containers/ImmerMemoryPolicy.h
```diff
@@ -0,0 +1,63 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright $YEAR-2021 ArangoDB GmbH, Cologne, Germany
```
jsteemann, line 4: Let's remove $YEAR here, and insert a copyright header as in the other files.

lib/Basics/Exceptions.h
goedderz: It's something I do often: Add an assertion in maintainer mode, and throw an exception in release builds, e.g
```
if (<condition>) {
  TRI_ASSERT(false);
  THROW_ARANGO_EXCEPTION_MESSAGE(<code>, <message>);
}
```
. Essentially, in many, if not most, places where I add an assertion (in maintainer mode), I also want to throw an exception in release mode. This also makes sure the specific error is logged in maintainer mode as well, which is lost in the example above.

They are not absolutely required, though I do think I would probably use them quite often now if available.

lib/Containers/ImmerMemoryPolicy.h
goedderz: Oops.

arangod/Agency/AgencyComm.cpp
goedderz: See https://github.com/arangodb/arangodb/pull/14484#discussion_r668670529

jsteemann (approved):
LGTM

PR #14492, opened 2021-07-09
Repair GEO_INTERSECTS

neunhoef:
This PR repairs the GEO_INTERSECT AQL function.

It was found that quite a few cases are not implemented and silently
return `false` for an intersection. Other cases return wrong results.
This needs to be fixed. We first add more unit tests to expose the
problem and then fix. At the very least we should return "not yet
implemented" instead of a silent `false`.

### Scope & Purpose

#### Backports:

- [*] Backports required for: *(Please specify versions and link PRs)*
      Probably for all published versions, since this is functionality
      from 3.4.

#### Related Information

- [ ] GitHub issue / Jira ticket number: BTS_475, DEVSUP-818

### Testing & Verification

  - [*] Added new C++ **Unit tests**


lib/Geo/ShapeContainer.cpp
```diff
@@ -571,27 +589,60 @@ bool ShapeContainer::intersects(S2Polyline const* other) const {
...
+  v.reserve(5);
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(0)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(1)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(2)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(3)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(0)).toPoint());
+  std::unique_ptr<S2Loop> loop;
+  loop = std::make_unique<S2Loop>(std::move(v), S2Debug::DISABLE);
+  std::unique_ptr<S2Polygon> rectPoly;
+  rectPoly = std::make_unique<S2Polygon>(std::move(loop), S2Debug::DISABLE);
```
gnusi, line 632: that makes absolutely no sense to allocate `rectPoly` on a heap

lib/Geo/ShapeContainer.cpp
```diff
@@ -600,7 +651,10 @@ bool ShapeContainer::intersects(S2LatLngRect const* other) const {
     }
 
     case ShapeContainer::Type::S2_POLYLINE: {
-      return contains(other);
+      auto rectPoly = ::latLngRectToPolygon(static_cast<S2LatLngRect const*>(other));
```
gnusi, line 654: that makes absolutely no sense to allocate rectPoly on a heap

gnusi (changes requested):

lib/Geo/ShapeContainer.cpp
neunhoef: Right. Not that it would matter much, since the computation which is following is orders of magnitude more expensive than the memory allocation, but I fixed it.

lib/Geo/ShapeContainer.cpp
neunhoef: Fixed.

maierlars (approved):

lib/Geo/ShapeContainer.cpp
```diff
@@ -55,6 +55,24 @@
...
+
+S2Polygon latLngRectToPolygon(S2LatLngRect const* rect) {
+  // Construct polygon from rect:
+  std::vector<S2Point> v;
+  v.reserve(5);
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(0)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(1)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(2)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(3)).toPoint());
+  v.emplace_back(Coordinate::fromLatLng(rect->GetVertex(0)).toPoint());
```
gnusi, line 68:
```suggestion
  v.emplace_back(rect->GetVertex(0).toPoint());
  v.emplace_back(rect->GetVertex(1).toPoint());
  v.emplace_back(rect->GetVertex(2).toPoint());
  v.emplace_back(rect->GetVertex(3).toPoint());
  v.emplace_back(rect->GetVertex(0).toPoint());
```

lib/Geo/ShapeContainer.cpp
```diff
@@ -557,41 +575,67 @@ bool ShapeContainer::equals(ShapeContainer const* cc) const {
...
+  std::unique_ptr<S2Loop> loop;
+  loop = std::make_unique<S2Loop>(std::move(v), S2Debug::DISABLE);
+  S2Polygon rectPoly{std::move(loop), S2Debug::DISABLE};
+  return poly->Intersects(&rectPoly);
```
gnusi, line 626:
```suggestion
  auto rectPoly = ::latLngRectToPolygon(rect)
  return poly->Intersects(&rectPoly)
```

lib/Geo/ShapeContainer.cpp
neunhoef: Actually, it has to be `ToPoint` with a capital T, but then it works. The `S2LatLng` object only has a `ToPoint` method, which I missed. Great!

lib/Geo/ShapeContainer.cpp
neunhoef: Good point, will fix.

gnusi (approved):

PR #14498, opened 2021-07-13
Fix invalid assertion and always remove blocker object on replication failure

jsteemann:
### Scope & Purpose

Slightly extend chaos tests and add some fixes that occurred during chaos, in particular:
* always remove `blocker` object for revision trees in case of replication failure
* fix `hasBlockerUpTo` function
* fix invalid assertion

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *chaos*.


arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -335,9 +343,34 @@ Result RocksDBMetadata::serializeMeta(rocksdb::WriteBatch& batch,
   }
 
   const rocksdb::SequenceNumber maxCommitSeq = committableSeq(appliedSeq);
+
+#ifdef ARANGODB_ENABLE_FAILURE_TESTS
+  // simulate another transaction coming along and trying to commit while
+  // we are serializing
+  TransactionId trxId = TransactionId::none();
+  
+  TRI_IF_FAILURE("TransactionChaos::randomSleep") {
```
mpoeter, line 352: Should we use a different failure point here? This is no "randomSleep".

mpoeter (approved):

arangod/RocksDBEngine/RocksDBMetadata.cpp
jsteemann: Done

KVS85 (approved):
LGTM.

PR #14504, opened 2021-07-15
Reduce the amount of exception macros

goedderz:
### Scope & Purpose

Reduce the amount of macros for exceptions.

_Should be merged after #14486_

#### Backports:

- [X] No backports required

### Testing & Verification

- [X] This change is a trivial rework / code cleanup without any test coverage.


lib/Basics/Exceptions.cpp
```diff
@@ -170,3 +154,17 @@ std::string Exception::FillFormatExceptionString(char const* format, ...) {
       << "Failed to create an error message, giving up. " << errorMessage;
   FATAL_ERROR_EXIT();
 }
+
+[[noreturn]] void ::arangodb::basics::abortOrThrowException(Exception e) {
```
jsteemann, line 158: I guess `e` should be reference here, so we don't copy it.
```suggestion
[[noreturn]] void ::arangodb::basics::abortOrThrowException(Exception const& e) {
```

lib/Basics/Exceptions.cpp
```diff
@@ -170,3 +154,17 @@ std::string Exception::FillFormatExceptionString(char const* format, ...) {
...
+#ifndef ARANGODB_ENABLE_MAINTAINER_MODE
+  throw e;
+#else
+  try {
+    throw std::move(e);
+  } catch (::arangodb::basics::Exception const& ex) {
+    LOG_TOPIC("fa7a1", FATAL, ::arangodb::Logger::CRASH) << ex.message();
+    TRI_ASSERT(false);
+  }
+  TRI_ASSERT(false);
```
jsteemann, line 168: What is the added value of throwing the exception and catching it immediately after?
Can't we just log its message directly and then abort?

lib/Basics/ResultError.cpp
```diff
@@ -55,6 +55,7 @@ auto Error::errorMessage() && noexcept -> std::string {
   if (!_errorMessage.empty()) {
     return std::move(_errorMessage);
   }
+  // Note that this is can throw.
```
jsteemann, line 58:
```suggestion
  // Note that this can throw.
```

jsteemann (approved):

lib/Basics/Exceptions.cpp
goedderz: `throw` copy-initializes the exception. So my train of thought was, if we take a `const&` here, that'd mean we'd have to copy (instead of moving) it when throwing, unconditionally. This way, the caller can at least choose to move the exception to avoid a copy altogether. However, I seem to have forgotten the move in non-maintainer mode; I will fix that.

lib/Basics/Exceptions.cpp
goedderz: True, thanks for noticing, that's a relict of incomplete refactoring. Will fix it.

lib/Basics/Exceptions.cpp
mpoeter: Taking the exception by value might also result in object slicing if we have a subtype of Exception. How about making this a template and the argument a universal reference?

goedderz: Good point!

PR #14508, opened 2021-07-16
simplify blocker handling

jsteemann:
### Scope & Purpose

This introduces RocksDBBlockerGuard so that acquiring and releasing blockers is made simpler, and especially that releasing blockers is not forgotten.
In addition, it makes placeBlocker() throw in case of an error. the method previously returned a Result, but that was picked up nowhere. Throwing in case of an (unlikely) error will force some sort of error handling on the call site or an abort of the operation.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/RocksDBEngine/RocksDBMetadata.cpp
```diff
@@ -805,3 +752,55 @@ void RocksDBMetadata::loadInitialNumberDocuments() {
...
+
+rocksdb::SequenceNumber RocksDBBlockerGuard::placeBlocker(TransactionId trxId) {
+  TRI_ASSERT(!_trxId.isSet());
+
+  auto& engine = _collection->vocbase().server().getFeature<EngineSelectorFeature>().engine<RocksDBEngine>();
+  rocksdb::SequenceNumber blockerSeq = engine.db()->GetLatestSequenceNumber();
+
+  if (!trxId.isSet()) {
+    trxId = TransactionId(transaction::Context::makeTransactionId());
+  }
```
mpoeter, line 789: I would prefer to have a second overload without any parameters that generates the transactionId and then calls this overload.

mpoeter (approved):

arangod/RocksDBEngine/RocksDBMetadata.cpp
jsteemann: Done

PR #14533, opened 2021-07-22
Use lower defaultMaxNodesPerCallstack for Mac.

mpoeter:
### Scope & Purpose

On OSX the default stack size for worker threads (non-main thread) is 512kb which is rather low, so we have to use a lower default for the max. number of nodes per callstack,

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] The behavior in this PR was *manually tested* by @mchacki

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Reduce default value for max-nodes-per-callstack to 200 for OSX, because on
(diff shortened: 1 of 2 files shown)


cpjulia (approved):
LGTM

mchacki (approved):
LGTM!
I think this is a small injection only, and only effects MacOs.

PR #14543, opened 2021-07-23
Bug fix/refactor rocks db transaction state and rocks db methods

mpoeter:
### Scope & Purpose

Attempt to untangle `RocksDBTransactionState` and `RocksDBMethods` & Co. There is still some room for improvement, but it should be better than before.

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/RocksDBEngine/Methods/RocksDBSingleOperationReadOnlyMethods.cpp
```diff
@@ -0,0 +1,73 @@
...
+Result RocksDBSingleOperationReadOnlyMethods::commitTransaction() {
+  return {};
+}
+
+Result RocksDBSingleOperationReadOnlyMethods::abortTransaction() {
+  return {};
+}
+
+rocksdb::ReadOptions RocksDBSingleOperationReadOnlyMethods::iteratorReadOptions() const {
+  return {};
```
jsteemann, line 51: I think this should never be called. Should we make this function throw an exception instead?

arangod/RocksDBEngine/Methods/RocksDBSingleOperationReadOnlyMethods.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+rocksdb::SequenceNumber RocksDBSingleOperationReadOnlyMethods::GetSequenceNumber() const noexcept {
+  return _db->GetLatestSequenceNumber();
+}
+                
+rocksdb::Status RocksDBSingleOperationReadOnlyMethods::Get(rocksdb::ColumnFamilyHandle* cf,
+                                            rocksdb::Slice const& key,
+                                            rocksdb::PinnableSlice* val) {
```
jsteemann, line 60:
```suggestion
rocksdb::Status RocksDBSingleOperationReadOnlyMethods::Get(rocksdb::ColumnFamilyHandle* cf,
                                                           rocksdb::Slice const& key,
                                                           rocksdb::PinnableSlice* val) {
```

arangod/RocksDBEngine/Methods/RocksDBSingleOperationReadOnlyMethods.cpp
```diff
@@ -0,0 +1,73 @@
...
+  ro.prefix_same_as_start = true;  // should always be true
+  ro.fill_cache = _state->options().fillBlockCache;
+  return _db->Get(ro, cf, key, val);
+}
+
+std::unique_ptr<rocksdb::Iterator> RocksDBSingleOperationReadOnlyMethods::NewIterator(
+    rocksdb::ColumnFamilyHandle*, ReadOptionsCallback) {
+  // This should never be called for a single operation transaction.
+  TRI_ASSERT(false);
+  THROW_ARANGO_EXCEPTION(TRI_ERROR_INTERNAL);
```
jsteemann, line 72: Let's add a descriptive message here.
```suggestion
  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "should not call NewIterator for single operation methods");
```

arangod/RocksDBEngine/Methods/RocksDBSingleOperationTrxMethods.cpp
```diff
@@ -0,0 +1,106 @@
...
+RocksDBSingleOperationTrxMethods::RocksDBSingleOperationTrxMethods(RocksDBTransactionState* state, rocksdb::TransactionDB* db)
+    : RocksDBTrxBaseMethods(state, db) {
+  TRI_ASSERT(_state->isSingleOperation());
+  TRI_ASSERT(!_state->hasHint(transaction::Hints::Hint::INTERMEDIATE_COMMITS));
+}
+
+rocksdb::ReadOptions RocksDBSingleOperationTrxMethods::iteratorReadOptions() const {
+  // This should never be called for a single operation transaction.
+  TRI_ASSERT(false);
+  THROW_ARANGO_EXCEPTION(TRI_ERROR_INTERNAL);
```
jsteemann, line 46:
```suggestion
  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "should not call iteratorReadOptions for single operation methods");
```

arangod/RocksDBEngine/Methods/RocksDBSingleOperationTrxMethods.cpp
```diff
@@ -0,0 +1,106 @@
...
+      break; 
+    }
+  }
+}
+
+std::unique_ptr<rocksdb::Iterator> RocksDBSingleOperationTrxMethods::NewIterator(
+    rocksdb::ColumnFamilyHandle*, ReadOptionsCallback) {
+  // This should never be called for a single operation transaction.
+  TRI_ASSERT(false);
+  THROW_ARANGO_EXCEPTION(TRI_ERROR_INTERNAL);
```
jsteemann, line 105:
```suggestion
  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, "should not call NewIterator for single operation methods");
```

arangod/RocksDBEngine/Methods/RocksDBTrxMethods.cpp
```diff
@@ -0,0 +1,214 @@
...
+  // call it ourselves here, because the call in the base class would only execute
+  // the base class implementation, therefore leaking the iteratorReadSnapshot.
+  cleanupTransaction();
+}
+
+Result RocksDBTrxMethods::beginTransaction() {
+  Result result = RocksDBTrxBaseMethods::beginTransaction();
+  
+  TRI_ASSERT(_iteratorReadSnapshot == nullptr);
+  if (hasIntermediateCommitsEnabled()) {
```
jsteemann, line 53: Should we check here is `beginTransaction()` succeeded, i.e.

arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
```diff
@@ -0,0 +1,408 @@
...
+  }
+  return result;
+}
+
+Result RocksDBTrxBaseMethods::abortTransaction() {
+  Result result;
+  if (_rocksTransaction != nullptr) {
+    rocksdb::Status status = _rocksTransaction->Rollback();
+    result = rocksutils::convertStatus(status);
+  }
```
jsteemann, line 92: This looks kind of asymmetrical:
When we commit a transaction, we call `cleanupTransaction()` if committing succeeds.
If committing fails or when we abort a transaction, we don't call `cleanupTransaction()` at all. It is then only called in the class' destructor. 
This may be ok. I was just wondering.

arangod/RocksDBEngine/RocksDBTransactionState.cpp
```diff
@@ -152,113 +144,63 @@ Result RocksDBTransactionState::beginTransaction(transaction::Hints hints) {
...
-                 _options.intermediateCommitSize != UINT64_MAX);
-      _readSnapshot = db->GetSnapshot();  // must call ReleaseSnapshot later
-      TRI_ASSERT(_readSnapshot != nullptr);
+    if (isSingleOperation()) {
+      _rocksMethods = std::make_unique<RocksDBSingleOperationTrxMethods>(this, db);
+    } else {
+      _rocksMethods = std::make_unique<RocksDBTrxMethods>(this, db);
     }
+  }
+  _rocksMethods->beginTransaction();
```
jsteemann, line 161: AFAIK `beginTransaction` returns a `Result`, so we should check here if it has failed or not.

arangod/RocksDBEngine/RocksDBTransactionState.cpp
```diff
@@ -477,12 +272,8 @@ Result RocksDBTransactionState::abortTransaction(transaction::Methods* activeTrx
...
   TRI_ASSERT(activeTrx->isMainTransaction());
   
-  Result result;
+  Result result = _rocksMethods->abortTransaction();
 
-  if (_rocksTransaction != nullptr) {
-    rocksdb::Status status = _rocksTransaction->Rollback();
-    result = rocksutils::convertStatus(status);
-  }
   cleanupTransaction();  // deletes trx
```
jsteemann, line 277: Ah, it seems `cleanupTransaction` is called here (after the abort). How the cleanups are currently done feels a bit awkward.

arangod/RocksDBEngine/RocksDBTransactionState.cpp
```diff
@@ -491,226 +282,67 @@ Result RocksDBTransactionState::abortTransaction(transaction::Methods* activeTrx
...
+      queryCache->invalidate(&_vocbase, tcoll->collection()->guid());
+    }
   }
-
-  createTransaction();
-  _rocksReadOptions.snapshot = _rocksTransaction->GetSnapshot();
-  TRI_ASSERT(_readSnapshot != nullptr);  // snapshots for iterators
-  TRI_ASSERT(_rocksReadOptions.snapshot != nullptr);
-  return TRI_ERROR_NO_ERROR;
+  return {};
```
jsteemann, line 341: The value of `result` should be returned here. Otherwise we may have a failure in `result` but not react on it.
```suggestion
  return result;
```

arangod/RocksDBEngine/RocksDBTransactionState.cpp
mpoeter: Yes, unfortunately it has been like this before and I wanted to limit the changes to avoid unintended consequences. But I agree that this feels akward - I'll try to clean this up a bit more.

PR #14548, opened 2021-07-26
Feature/refactor arangobench

cpjulia:
### Scope & Purpose
* Refactored arangobench:
  - Updated testcases to show description of them when beginning execution
  - Fixed testcase histogram with time measures when batch size > 0
  - Integrated testcases with Velocypack for simplification 
  - Deprecated some testcases
  - Internal changes for performance optimization

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests, such as *arangobench*.


arangosh/Benchmark/testcases/RandomShapesTest.h
```diff
@@ -25,98 +25,78 @@
...
+        Builder b;
+        b.openObject();
+        b.add("_key", Value(std::string("testkey") + std::to_string(keyId)));
+        uint32_t t = _randomValue % static_cast<uint32_t>(globalCounter + threadNumber + 1);
+        for (uint64_t i = 1; i <= n; ++i) {
+          std::string fieldName = "value" + std::to_string(static_cast<uint64_t>(globalCounter + i));
+          if (t % 3 == 0) {
+            b.add(fieldName, Value(true));
+          } else if (t % 3 == 1) {
+            b.add(fieldName, Value(NULL));
```
jsteemann, line 89: IMO `NULL` is equivalent to a numeric value of `0`, so it will be stored as a number. We want a JSON `null` here instead:
```suggestion
            b.add(fieldName, Value(ValueType::Null));
```

arangosh/Benchmark/testcases/StreamCursorTest.h
```diff
@@ -25,83 +25,68 @@
...
+      } else if (mod == 0) {
+        values = "UPDATE { \"_key\": \"1\" } WITH { \"foo\": 1";
+        for (uint64_t i = 1; i <= n; ++i) {
+          values += ", \"value" + std::to_string(i) + "\": true";
+        }
+        values += " } INTO " + _arangobench.collection() + std::string(" OPTIONS { \"ignoreErrors\": true } ");
+        b.add("query", Value(values));
+      } else {
+        values = "FOR doc IN " + _arangobench.collection() 
+          += std::string(" RETURN doc");
```
jsteemann, line 80:
```suggestion
        values = "FOR doc IN " + _arangobench.collection() + std::string(" RETURN doc");
```

arangosh/Benchmark/testcases/StreamCursorTest.h
```diff
@@ -25,83 +25,68 @@
...
+          values += ", \"value" + std::to_string(i) + "\": true";
+        }
+        values += "} INTO " + _arangobench.collection();
+        b.add("query", Value(values));
+      } else if (mod == 0) {
+        values = "UPDATE { \"_key\": \"1\" } WITH { \"foo\": 1";
+        for (uint64_t i = 1; i <= n; ++i) {
+          values += ", \"value" + std::to_string(i) + "\": true";
+        }
+        values += " } INTO " + _arangobench.collection() + std::string(" OPTIONS { \"ignoreErrors\": true } ");
```
jsteemann, line 76: Can be simplified slightly:
```suggestion
        values += " } INTO " + _arangobench.collection() + std::string(" OPTIONS { ignoreErrors: true }");
```

arangosh/Benchmark/BenchmarkThread.h
```diff
@@ -271,357 +245,222 @@ class BenchmarkThread : public arangodb::Thread {
...
+            _operationsCounter->incFailures(errorCount);
+            if (++_warningCount < maxWarnings) {
+              LOG_TOPIC("b1db5", WARN, arangodb::Logger::BENCH)
+                << type << " operation sServer side warning count: " << errorCount;
+            }
+          }
+        }
       }
+
       return;
```
mpoeter, line 371: If `result != nullptr && result->isComplete() && !result->wasHttpError() && !batch` is satisfied we will return here without updating any counters - this is a change to be previous behavior.

mpoeter (changes requested):
I think the handling of failed requests is broken now, otherwise only some minor suggestions.

arangosh/Benchmark/BenchmarkThread.h
jsteemann: I think the change is valid, because counters should only increased when there are any errors (the counters are simply error counters).

For non-batch requests (i.e. `!batch`) an error was previously assumed if `result == nullptr || !result->isComplete() || result->wasHttpError()`.
In the refactored version, we are still counting in this case.

And for batch requests (.e. `batch`) this case should also be ok, because for complete HTTP responses the number of errors is only in some header, which we are still checking.

And in case `result == nullptr || !result->isComplete() || result->wasHttpError()` we are still counting everything, regardless of batch requests or single requests.

lib/Rest/GeneralRequest.cpp
```diff
@@ -39,36 +39,37 @@
...
-      return "PATCH";
+      return {"PATCH", 5};
 
     case RequestType::POST:
-      return "POST";
+      return {"POST", 4};
 
     case RequestType::PUT:
-      return "PUT";
+      return {"PUT", 3};
```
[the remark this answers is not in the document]
jsteemann: Fixed.

lib/Rest/GeneralRequest.cpp
```diff
@@ -39,36 +39,37 @@
...
-          << "illegal http request method encountered in switch";
-      return "UNKNOWN";
+      break;
   }
+      
+  LOG_TOPIC("62a53", WARN, arangodb::Logger::FIXME)
+        << "illegal http request method encountered in switch";
 
-  return "UNKNOWN";  // in order please MSVC
+  return {"UNKNOWN", 7};  // in order to please MSVC
```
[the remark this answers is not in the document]
jsteemann: Fixed.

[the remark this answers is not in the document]
jsteemann: Fixed.

lib/Basics/StringUtils.h
```diff
@@ -156,35 +156,73 @@ std::vector<std::string> wrap(std::string const& sourceStr, size_t size,
...
+        [[fallthrough]];
+      case 3:
+        result += (tolower(lhs[offset + 2]) != tolower(rhs[offset + 2]));
+        [[fallthrough]];
+      case 2:
+        result += (tolower(lhs[offset + 1]) != tolower(rhs[offset + 1]));
+        [[fallthrough]];
+      case 1:
+        result += (tolower(lhs[offset + 0]) != tolower(rhs[offset + 0]));
+    }
```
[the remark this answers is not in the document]
jsteemann: At least the previous boost::iequals calls (which are now replaced with our own case-agnostic string comparison) ranked very high in the list of the most expensive functions during profiling. The comparison is done repeatedly for every request that is sent/received, and there twice in a loop that iterates over all HTTP headers.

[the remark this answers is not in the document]
jsteemann: Will try to adjust the code to raw pointers now.

arangosh/Benchmark/BenchmarkThread.h
mpoeter: Ah, so basically a `batch` request should never report an HTTP error?
I saw that you tried to unify the error handling for the single/batch requests, but unfortunately I still think this function does not behave like the previous code. For example, for batch requests the old code increased `_operationsCounter` by `numOperations` while now we always increment by one. Also, the old code for batch requests had an optional verbose output that is no longer available in the new code.
Can we have batch requests with a single operation? If not, I would replace the `batch` argument with a `numOperations` argument and derive `bool batch = numOperations > 1`.

mpoeter (approved):

PR #14549, opened 2021-07-27
Reduce memory usage of Merkle trees

jsteemann:
### Scope & Purpose

Reduce memory usage for in-memory revision trees. Previously, a revision tree instance for a non-empty collection/shard was using 4 MB of memory when uncompressed. Trees that were unused for a while were compressed on the fly to use less memory, and later uncompressed again when needed.
Now the uncompressed in-memory version of the revision tree will dynamically allocate memory as needed. This allows the initial version of the trees to get away with just 64 KB of memory. Memory usage will grow lazily when more parts of the trees get populated. The compression of unused in-memory tree data is still in place.

#### Backports:

- [x] Backports required for: potentially 3.8 (needs discussion)

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtest, recovery, ...*.


lib/Containers/MerkleTree.cpp
```diff
@@ -750,60 +893,122 @@ void MerkleTree<Hasher, BranchingBits>::serialize(velocypack::Builder& output,
...
+          snappy::Compress(&source, &sink);
+          compressedLength = output.size() - previousLength;
+        }
+        uint32_t value = basics::hostToLittle(static_cast<uint32_t>(compressedLength));
+        char const* p = reinterpret_cast<const char*>(&value);
+
+        // patch offsets table with compressed length
+        for (size_t j = 0; j < sizeof(uint32_t); ++j) {
+          output[tableBase + i * sizeof(uint32_t) + j] = p[j];
+        }
```
mpoeter, line 994:
```suggestion
        uint32_t value = basics::hostToLittle(static_cast<uint32_t>(compressedLength));
        // patch offsets table with compressed length
        memcpy(&output[tableBase + i * sizeof(uint32_t)], &value, sizeof(value));
```

mpoeter (approved):

lib/Containers/MerkleTree.cpp
```diff
@@ -910,36 +1146,95 @@ template <typename Hasher, std::uint64_t const BranchingBits>
...
+  if (_data.shards.size() <= shard) {
+    // allocate a few shards at once
+    _data.shards.resize(std::max(shard + 1, std::uint64_t(4)));
+  }
+  if (_data.shards[shard] == nullptr) {
+    _data.shards[shard] = std::make_unique<uint8_t[]>(ShardSize);
+    Node* p = reinterpret_cast<Node*>(_data.shards[shard].get());
+    
+    std::uint64_t nodesPerShard = shardSize(meta().depth) / NodeSize;
+    for (std::uint64_t i = 0; i < nodesPerShard; ++i) {
```
neunhoef, line 1176: This could potentially also be done by `memset` and could be a performance critical part...

lib/Containers/MerkleTree.cpp
```diff
@@ -910,36 +1146,95 @@ template <typename Hasher, std::uint64_t const BranchingBits>
...
+MerkleTreeBase::Node& MerkleTree<Hasher, BranchingBits>::node(std::uint64_t index) {
   // not thread-safe, lock buffer from outside
+  
+  // determine shard
+  std::uint64_t shard = shardForIndex(index);
+  TRI_ASSERT(shard < numberOfShards());
+
+  // lazily allocate backing memory
+  if (_data.shards.size() <= shard) {
+    // allocate a few shards at once
```
neunhoef, line 1168: The comment says that we allocate a few at once, but the code does not do this, unless the underlying container does it automatically. After 4 shards have been reached, this code just resizes one by one, as far as I see.

lib/Containers/MerkleTree.cpp
```diff
@@ -1002,7 +1297,7 @@ void MerkleTree<Hasher, BranchingBits>::modify(std::vector<std::uint64_t> const&
 
 template <typename Hasher, std::uint64_t const BranchingBits>
 bool MerkleTree<Hasher, BranchingBits>::modifyLocal(
-    Node& node, std::uint64_t count, std::uint64_t value, bool isInsert) noexcept {
+    Node& node, std::uint64_t count, std::uint64_t value, bool isInsert) {
```
neunhoef, line 1300: Why is this no longer noexcept? I do not see any exception which could happen.

neunhoef: I understand the cases below, where `node(i)` is called, but here, it seems `noexcept`.

lib/Containers/MerkleTree.cpp
```diff
@@ -1045,13 +1340,16 @@ void MerkleTree<Hasher, BranchingBits>::leftCombine(bool withShift) noexcept {
...
     // empty -> n
     for (std::uint64_t i = 2; i < n; ++i) {
-      Node& src = this->node(i);
-      Node& dst = this->node((i + 1) / 2);
-
-      TRI_ASSERT(&src != &dst);
-      dst.count += src.count;
-      dst.hash ^= src.hash;
-      src = Node{0, 0};
+      if (!this->empty(i)) {
```
neunhoef, line 1343: I fear this change is not valid. For example, we combine buckets 3 and 4 into 2. If 3 and 4 are both empty (not even allocated), then 2 must be empty after the transformation. However, this if in line 1343 would prevent any change in bucket 2, so that if there was data in bucket 2 before, it would now be wrong!
I fear we must change the loop to do one combination in one iteration and distinguish 4 cases:
  (1) all empty ==> do nothing
  (2) destination empty, at least one of the sources not ==> copy and combine
  (3) destination non-empty, at least one of the sources not ==> copy and combine
  (4) destination non-empty, both sources empty ==> make destination empty
And: Bucket 0 and buckets 1,2 -> 1 are special, because of aliasing!

lib/Containers/MerkleTree.cpp
```diff
@@ -1063,15 +1361,18 @@ void MerkleTree<Hasher, BranchingBits>::leftCombine(bool withShift) noexcept {
     // ...
     // empty     -> n - 1
     for (std::uint64_t i = 1; i < n; ++i) {
-      Node& src = this->node(i);
-      Node& dst = this->node(i / 2);
+      if (!this->empty(i)) {
```
neunhoef, line 1364: Similar thing here.

lib/Containers/MerkleTree.cpp
```diff
@@ -1159,13 +1460,16 @@ void MerkleTree<Hasher, BranchingBits>::rightCombine(bool withShift) noexcept {
...
     // n-1       -> n - 1
     for (std::uint64_t i = n - 3; /* i >= 0 */ ; --i) {
-      Node& src = this->node(i);
-      Node& dst = this->node((n + i - 1) / 2);
-
-      TRI_ASSERT(&src != &dst);
-      dst.count += src.count;
-      dst.hash ^= src.hash;
-      src = Node{0, 0};
+      if (!this->empty(i)) {
```
neunhoef, line 1463: I fear a similar problem hides here.

lib/Containers/MerkleTree.cpp
```diff
@@ -1180,13 +1484,16 @@ void MerkleTree<Hasher, BranchingBits>::rightCombine(bool withShift) noexcept {
...
     // n-2, n-1  -> n - 1
     for (std::uint64_t i = n - 2; /* i >= 0 */; --i) {
-      Node& src = this->node(i);
-      Node& dst = this->node((n + i) / 2);
-
-      TRI_ASSERT(&src != &dst);
-      dst.count += src.count;
-      dst.hash ^= src.hash;
-      src = Node{0, 0};
+      if (!this->empty(i)) {
```
neunhoef, line 1487: And here.

neunhoef (changes requested):
This looks overall good, but I think I found a few things which are actually bugs. See comments. Happy to discuss.

lib/Containers/MerkleTree.cpp
neunhoef: This comment was wrong. Due to the fact that earlier loop iterations had the current `dst` as a `src` and have then erased it, all is well.

neunhoef: I have added a comment to explain this in the code.

lib/Containers/MerkleTree.cpp
neunhoef: Also correct, please forget.

lib/Containers/MerkleTree.cpp
neunhoef: Also correct, please forget.

lib/Containers/MerkleTree.cpp
neunhoef: Also correct, please forget.

lib/Containers/MerkleTree.cpp
```diff
@@ -166,17 +264,85 @@ MerkleTree<Hasher, BranchingBits>::fromSnappyCompressed(std::string_view buffer)
...
+        }
+      }
+
+      size_t length;
+      bool canUncompress = snappy::GetUncompressedLength(compressedData, compressedLength, &length);
+      if (!canUncompress || length != shardLength) {
+        throw std::invalid_argument(
+            "Cannot determine size of Snappy-compressed data.");
+      }
+      if (!snappy::RawUncompress(compressedData, compressedLength, reinterpret_cast<char*>(data.shards[i].get()))) {
```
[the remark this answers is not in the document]
jsteemann: I converted `struct Node` into a POD, so it should be compatible with memory operations on the byte level. I also moved the code away from placement new and instead used plain memset to initialize the memory.

lib/Containers/MerkleTree.cpp
jsteemann: Thanks! Applied this change.

lib/Containers/MerkleTree.cpp
jsteemann: Done

lib/Containers/MerkleTree.cpp
jsteemann: Fixed.

lib/Containers/MerkleTree.cpp
jsteemann: True. Changed it back to be `noexcept`.

3rdParty/snappy/snappy-1.1.9/CMakeLists.txt
```diff
@@ -78,8 +78,8 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions")
 
   # Disable RTTI.
-  string(REGEX REPLACE "-frtti" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
+  #string(REGEX REPLACE "-frtti" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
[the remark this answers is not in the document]
jsteemann: Unfortunately this change was necessary in order to successfully link it against arangod.
I left some comments in the CMakeLists.txt file itself and in README_maintainers.md now.

neunhoef (approved):
LGTM

lib/Containers/MerkleTree.cpp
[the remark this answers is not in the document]
mpoeter: Although this should work as expected, IMO this is still not standard conform. We are allocating a memory block of type `uint8_t[]` but later access it as `Node[]`, and on top of it we cast it to `char*` in for the snappy uncompress.
I'm not saying we should change this code now, but we should be careful with code like this because AFAIK this is unspecified if not undefined behavior.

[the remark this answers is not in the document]
mpoeter: FWIW: if we would allocate a `Node[]` instead of a `uint8_t[]` this would already improve the situation. As it is now it is not guaranteed that the allocation has the proper alignment required by `Node`.

mpoeter (approved):
Would prefer if we could change the shard type from `uint8_t[]` to `Node[]` to avoid at least some of the type aliasing problems. Otherwise LGTM.

lib/Containers/MerkleTree.h
```diff
@@ -49,10 +52,86 @@ class FnvHashProvider : public HashProvider {
...
+  static_assert(sizeof(Meta) == 64, "Meta size assumptions invalid.");
+  static_assert(sizeof(Meta::Padding) == 16, "Meta padding size assumptions invalid.");
+  static constexpr std::uint64_t MetaSize = sizeof(Meta);
+  
+  // size of each shard, in bytes. 
+  // note: trees with a small depth may only have a single shard which is smaller than this value
+  static constexpr std::uint64_t ShardSize = (1 << 16);
+
+  struct Data {
+    using ShardType = std::unique_ptr<std::uint8_t[]>;
```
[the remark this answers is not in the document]
jsteemann: Thanks, fixed that.

mpoeter (approved):

PR #14550, opened 2021-07-27
Remove mutexes from Logging

jsteemann:
### Scope & Purpose

PR that tries to remove mutexes from initialization, shutdown and flushing of the Logger.
The problem with the mutexes is that they are held for the duration of the mentioned functions, and at the same time may be entered by a signal handler. In this case things may deadlock.

The change in this PR is to remove the `_initializeMutex` entirely, and instead use the `_active` and `_loggingThread` variables atomically. This should be safer and reentrant.

#### Backports:

- [x] No backports required

### Testing & Verification


lib/Logger/Logger.cpp
```diff
@@ -704,64 +714,62 @@ void Logger::append(LogGroup& group,
...
                                    "Logger already initialized");
   }
 
   // logging is now active
-  TRI_ASSERT(_active);
-  
   if (threaded) {
-    _loggingThread = std::make_unique<LogThread>(server, ::LogThreadName);
-    if (!_loggingThread->start()) {
+    auto loggingThread = new LogThread(server, ::LogThreadName);
```
jsteemann, line 724: Can we make it a unique_ptr again? Otherwise we may leak it if `loggingThread->start()` fails. It's not highly important because the program will abort then anyway, but just out of habit we should not use raw `new` ops if avoidable.
```suggestion
    auto loggingThread = std::make_unique<LogThread>(server, ::LogThreadName);
```

lib/Logger/Logger.cpp
```diff
@@ -704,64 +714,62 @@ void Logger::append(LogGroup& group,
...
+    if (!loggingThread->start()) {
       LOG_TOPIC("28bd9", FATAL, arangodb::Logger::FIXME)
           << "could not start logging thread";
       FATAL_ERROR_EXIT();
     }
-  }
   
-  // generate threaded logging?
-  _threaded = threaded;
+    _loggingThread.store(loggingThread, std::memory_order_release);
```
jsteemann, line 731:
```suggestion
    _loggingThread.store(loggingThread.release(), std::memory_order_release);
```

lib/Logger/Logger.cpp
```diff
@@ -704,64 +714,62 @@ void Logger::append(LogGroup& group,
...
+  if (!_active.exchange(false, std::memory_order_acquire)) {
+    // if logging not activated or already shut down, then we can abort here
     return;
   }
+  // logging is now inactive
 
-  TRI_ASSERT(!_active);
-
+  // reset the instance variable in Logger, so that others won't see it anymore
+  auto loggingThread = _loggingThread.exchange(nullptr, std::memory_order_relaxed);
```
jsteemann, line 747: I think `loggingThread` will be a raw pointer here, so it will be leaked once the function ends. Can we capture in a unique_ptr, so that whenever we exit this function, we'll destroy it.
```suggestion
  auto loggingThread = std::unique_ptr<LogThread>(_loggingThread.exchange(nullptr, std::memory_order_relaxed));
```

lib/Logger/Logger.h
```diff
@@ -343,6 +338,19 @@ class Logger {
   static std::string _outputPrefix;
   static std::string _hostname;
 
-  static std::unique_ptr<LogThread> _loggingThread;
+  struct ThreadRef {
+    ThreadRef();
```
jsteemann, line 342: Shall we declare this struct as non-copyable in addition? Currently it is never copied nor moved, but explicitly forbidding it feels better.

mpoeter: Done

mpoeter (approved):

cpjulia (approved):
LGTM

PR #14551, opened 2021-07-27
experiment to check if SyncWAL works on Windows

jsteemann:
### Scope & Purpose

Enable RocksDB SyncWAL() calls on Windows as well.
Previously they were disabled for Windows but enabled for all other platforms. For Windows, there was a workaround. The reason for not calling SyncWAL() on Windows previously was that RocksDB reported the operation as unsupported.

Tested manually on Windows that it now works.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
+++ b/arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
@@ -376,7 +376,7 @@ arangodb::Result RocksDBTrxBaseMethods::doCommit() {
   _state->commitCollections(_lastWrittenOperationTick);
   cleanupCollTrx.cancel();
 
-  // wait for sync if required, for all other platforms but Windows
+  // wait for sync if required
   if (_state->waitForSync()) {
     auto& selector = _state->vocbase().server().getFeature<EngineSelectorFeature>();
     auto& engine = selector.engine<RocksDBEngine>();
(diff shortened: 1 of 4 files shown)


arangod/RocksDBEngine/RocksDBSyncThread.cpp
```diff
@@ -75,7 +75,6 @@ Result RocksDBSyncThread::syncWal() {
 }
 
 Result RocksDBSyncThread::sync(rocksdb::DB* db) {
-#ifndef _WIN32
   // if called on Windows, we would get the following error from RocksDB:
   // > Not implemented: SyncWAL() is not supported for this implementation of
   // WAL file
```
mpoeter, line 80: This comment is now also obsolete.

arangod/RocksDBEngine/RocksDBTransactionState.cpp
```diff
@@ -429,7 +418,6 @@ arangodb::Result RocksDBTransactionState::internalCommit() {
   commitCounts();
   cleanupCollTrx.cancel();
 
-#ifndef _WIN32
   // wait for sync if required, for all other platforms but Windows
```
mpoeter, line 421:
```suggestion
  // wait for sync if required
```

mpoeter (approved):
I have tested this on my Windows machine and it now works as intended. I have also checked the RocksDB code and the `IsSyncThreadSafe()` function which is used to determine whether the old error `SyncWAL() is not supported for this implementation of WAL file` should be returned is now implemented as follows:
```

File: arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp [hunk 3]
```diff
@@ -387,7 +376,6 @@ arangodb::Result RocksDBTrxBaseMethods::doCommit() {
   _state->commitCollections(_lastWrittenOperationTick);
   cleanupCollTrx.cancel();
 
-#ifndef _WIN32
   // wait for sync if required, for all other platforms but Windows
```
jsteemann, line 379:
```suggestion
  // wait for sync if required
```

arangod/RocksDBEngine/RocksDBSyncThread.cpp
jsteemann, line 80:
```suggestion
```

PR #14553, opened 2021-07-28
attempt to pacify clang-1100

jsteemann:
### Scope & Purpose

Devel port: https://github.com/arangodb/arangodb/pull/14554

Attempt to pacify specific clang compiler version. This is just a workaround and may not fix the problem.

#### Backports:

- [x] No backports required

### Testing & Verification


mchacki (approved):
LGTM.

I think right now we do not move it at all.
However I do not see a reason why it should not be default movable.
But as this I not required right now I am fine with merging this pr.

KVS85 (commented):
LGTM.

PR #14554, opened 2021-07-28
attempt to pacify clang-1100

jsteemann:
### Scope & Purpose

Devel port of https://github.com/arangodb/arangodb/pull/14553

Attempt to pacify specific clang compiler version. This is just a workaround and may not fix the problem.

#### Backports:

- [x] Backports required for 3.8: https://github.com/arangodb/arangodb/pull/14553

### Testing & Verification


mchacki (approved):
LGTM.

I think right now we do not move it at all.
However I do not see a reason why it should not be default movable.
But as this I not required right now I am fine with merging this pr.

PR #14563, opened 2021-07-29
Cleanup random stuff

mpoeter:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/720

Cleans up some random stuff that I stumbled over in the past weeks (mainly unused code removal and reduction of code duplication).

#### Backports:

- [x] No backports required

### Testing & Verification


arangod/Transaction/Context.cpp
```diff
@@ -109,10 +108,7 @@ void transaction::Context::cleanup() noexcept {
 
   _stringBuffer.reset();
 
-  if (_ownsResolver) {
-    delete _resolver;
-    _resolver = nullptr;
-  }
```
jsteemann, line 115: Awesome. I remember we once had functionality to pass an existing resolver into a transaction::Context, but that seems to have been removed a while ago already. So going with a unique_ptr is exactly the right thing here :+1:

arangod/RocksDBEngine/RocksDBIterators.h
```diff
@@ -82,6 +82,9 @@ class RocksDBAnyIndexIterator final : public IndexIterator {
   void resetImpl() override;
 
  private:
+  template <typename Func>
+  bool doNext(size_t limit, Func func);
```
jsteemann, line 86: Should be pass by const reference to avoid copying here?
```suggestion
  bool doNext(size_t limit, Func const& func);
```

mpoeter: Sure, although the lamdas of the two callers are small enough that this shouldn't make any difference.

arangod/Transaction/Context.cpp
mpoeter: I figured there had to be a historical reason for the code to look like that. 😉

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -141,6 +141,11 @@ class RocksDBCollection final : public RocksDBMetaCollection {
                 RevisionId expectedRev, ManagedDocumentResult& previous,
                 OperationOptions& options);
 
+  Result performUpdateOrReplace(transaction::Methods* trx,
```
cpjulia, line 144: Have you considered marking this function `[[nodiscard]]`?

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -141,6 +141,11 @@ class RocksDBCollection final : public RocksDBMetaCollection {
...
+                                velocypack::Slice newSlice,
```
cpjulia, line 145: I think `newSlice` could be `const` because apparently none of its non-`const` methods are used in this function.

arangod/RocksDBEngine/RocksDBVPackIndex.h
```diff
@@ -147,6 +147,10 @@ class RocksDBVPackIndex : public RocksDBIndex {
   Result checkReplace(transaction::Methods& trx, RocksDBMethods* methods,
                       LocalDocumentId const& documentId, velocypack::Slice doc,
                       OperationOptions const& options) override;
+                      
+  Result checkOperation(transaction::Methods& trx, RocksDBMethods* methods,
```
cpjulia, line 151: Perhaps `[[nodiscard]]` here too? Would the caller ignoring the returned `Result` be acceptable?

cpjulia (approved):
LGTM

PR #14572, opened 2021-07-30
Experimental LogContext implementation.

mpoeter:
### Scope & Purpose

Experimental LogContext implementation for structured logging.
This only adds LogContext infrastructure, so no actual changes to log messages yet, therefore not CHANGELOG entry.

- [x] :pizza: New feature

#### Backports:

- [x] No backports required

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/APM-24

### Testing & Verification


lib/Logger/LogContext.h
```diff
@@ -0,0 +1,112 @@
...
+template <typename Func>
+auto withLogContext(Func func) {
+  return [func = std::move(func), ctx = LogContext::current()](auto&&... args) mutable {
+    LogContext::ScopedContext ctxGuard(std::move(ctx));
+    using result_t = std::invoke_result_t<Func, decltype(args)...>;
+    if constexpr (std::is_same_v<result_t, void>) {
+      func(std::forward<decltype(args)>(args)...);
+    } else {
+      return func(std::forward<decltype(args)>(args)...);
+    }
```
maierlars, line 107:
```suggestion
auto withLogContext(Func&& func) {
  return [func = std::forward<Func>(func), ctx = LogContext::current()](auto&&... args) mutable {
    LogContext::ScopedContext ctxGuard(std::move(ctx));
    return std::forward<Func>(func)(std::forward<decltype(args)>(args)...);
```
One should call the lambda with the same ref-qualifiers as it was passed in, for example if it only callable as r-value. Therefore I added `std::forward<Func>(f)`. Furthermore you can return `void` in C++. You don't need that `if constexpr`. 

Lastly, the `LogContext` is moved out of the lambda which makes this lambda single use. This might not be true for `Func` and thus could lead to confusion and unexpected behavior.

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,122 @@
...
+  TRI_ASSERT(_idx < LogContext::current()._impl->values.size());
+  TRI_ASSERT(LogContext::current()._impl->values[_idx].second == _newValue);
+#endif
+  LogContext::current().restore(_idx, std::move(_oldValue));
+}
+
+LogContext::ScopedMerge::ScopedMerge(LogContext ctx) {
+  _values.reserve(ctx._impl->values.size());
+  for (auto const& v : ctx._impl->values) {
+    _values.emplace_back(v.first, v.second);
```
maierlars, line 59: Don't we want to call `set` instead? This depends on the intended semantics; currently if I set a value on a log context with a key that already exists, the value is replaced. However, when we merge, the key is inserted multiple times. I would be in favor of last or first values wins.

jsteemann: Good point. I would be in favor or letting the last value win, i.e. simply overwrite existing values when they are stored again.

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+LogContext::ScopedContext::~ScopedContext() {
+  if (_oldContext) {
+    LogContext::setCurrent(LogContext(std::move(_oldContext)));
+  }
+}
+  
+LogContext::LogContext() : _impl(std::make_shared<Impl>()) {}
+
+std::pair<std::size_t, std::optional<std::string>> LogContext::set(std::string key, std::string value) {
```
jsteemann, line 78: Why is this returning an std::optional<std::string> as its second result value?
IMO the result string will be populated by this method. There is no code path in which it is not getting populated.

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,122 @@
...
+  
+void LogContext::restore(std::size_t index, std::optional<std::string>&& value) {
+  ensureUniqueOwner();
+  TRI_ASSERT(index < _impl->values.size());
+  if (!value.has_value()) {
+    TRI_ASSERT(index == _impl->values.size() - 1);
+    _impl->values.pop_back();
+  } else {
+    _impl->values[index].second = std::move(value.value());
+  }
```
jsteemann, line 101: This API does 2 things:
- it can overwrite the value of an arbitrary index
- it can remove the last existing item
IMO these are 2 different things, so I would favor having 2 separate functions (with different signatures) for it.

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+void LogContext::ensureUniqueOwner() {
+  if (!_impl.unique()) {
+    // this is a very simple copy-on-write implementation, but we can easily switch
+    // to something more elaborate like `immer` later.
+    _impl = std::make_shared<Impl>(*_impl);
+  }
+}
+
+void LogContext::each(std::function<void(std::string const&, std::string const&)> cb) {
```
jsteemann, line 112: Do we need to copy the function here? Shouldn't it be passed by const reference
```suggestion
void LogContext::each(std::function<void(std::string const&, std::string const&)> const& cb) {
```

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,112 @@
...
+    ~ScopedContext();
+   private:
+    std::shared_ptr<Impl> _oldContext;
+  };
+  
+  LogContext();
+
+  /// @brief iterates through the local LogContext's key/value pairs and calls
+  /// the callback function for each of them.
+  void each(std::function<void(std::string const&, std::string const&)>);
```
jsteemann, line 78:
```suggestion
  void each(std::function<void(std::string const&, std::string const&)> const&);
```

lib/Logger/LogContext.cpp
jsteemann: It also looks like we can mark this method as `const`.

lib/Logger/Logger.cpp
```diff
@@ -529,6 +531,12 @@ void Logger::log(char const* logid, char const* function, char const* file, int
       dumper.appendString(_hostname.data(), _hostname.size());
     }
 
+    // meta data from log context
+    logContext.each([&out](std::string const& key, std::string const& value) {
+      out.append(",\"", 2).append(key).append("\":\"", 3).append(value);
```
jsteemann, line 536: Here we need to ensure that at least the values are properly JSON-encoded, but ideally we also escape all keys here to be ready for unexpected things. See above for the usage of `dumper.appendString()`.

lib/Logger/LogContext.cpp
mpoeter: The second result is the previous value of the key if any existed. If there was no key/value pair with the given key, then the returned optional is empty.

lib/Logger/LogContext.cpp
mpoeter: Here we want to restore the context to the state previous to applying our changes. Thus we have to differentiate between the cases "we have replaced an existing key" and "we have inserted a new key".

If we have replaced an existing key, then the previous value is stored in our `optional<std::string>` and we again overwrite the value at the index where we found the existing key. Otherwise we have inserted a new key/value pair, in which case our index must be the last one in the vector and we simply remove the whole pair.

lib/Logger/LogContext.h
mpoeter: > One should call the lambda with the same ref-qualifiers as it was passed in, for example if it only callable as r-value. Therefore I added std::forward<Func>(f). 

Good point!

> Furthermore you can return void in C++. You don't need that if constexpr.

Unfortunately the MSVC compiler complained about that, that's why I had to add that annoying if.

> Lastly, the LogContext is moved out of the lambda which makes this lambda single use. This might not be true for Func and thus could lead to confusion and unexpected behavior.

Also a good point...

maierlars: According to [this](https://en.cppreference.com/w/cpp/language/return)
> In a function returning (possibly cv-qualified) void, the return statement with expression can be used, if the expression type is (possibly cv-qualified) void.

And [it looks like ](https://godbolt.org/z/PqoT5rWed)v19.28 doesn't complain. Which version do you use?

mpoeter: If have 19.29, and for some reason it now compiled successfully. Weird..,

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,122 @@
...
+#endif
+  std::tie(_idx, _oldValue) = LogContext::current().set(std::move(key), std::move(value));
+}
+
+LogContext::ScopedValue::~ScopedValue() {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+  TRI_ASSERT(_idx < LogContext::current()._impl->values.size());
+  TRI_ASSERT(LogContext::current()._impl->values[_idx].second == _newValue);
+#endif
+  LogContext::current().restore(_idx, std::move(_oldValue));
```
Dronplane, line 53: Maybe we should not rely on LogContext::current()  is the same as it was provided in the ScopedValue constructor?
E.g. One could create a ScopedValue as class member. Set some "context"  at class instance creation. Then that instance could be transferred to another thread  and dtor will be called by some another thread. So we will set values in one LogContext and restore to some another LogContext.

mpoeter: A `ScopedValue` should not be stored in a data member of some data structure that is passed between different threads. `ScopedValue`, `ScopedContext` etc. should only ever be created on the stack. Unfortunately there is not good way to enforce this in C++. I can (and should) make `ScopedValue` non-copyable though.

Dronplane: Maybe at least store thread_id in the ScopedValue ctor and ASSERT it on dtor ?

arangod/Scheduler/Scheduler.h
```diff
@@ -145,8 +146,13 @@ class Scheduler {
 
   template<typename F>
   struct WorkItem final : WorkItemBase, F {
-    explicit WorkItem(F f) : F(std::move(f)) {}
-    void invoke() override { this->operator()(); }
+    explicit WorkItem(F f) : F(std::move(f)), logContext(LogContext::current()) {}
+    void invoke() override {
+      LogContext::ScopedContext ctxGuard(logContext);
```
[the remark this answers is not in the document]
Dronplane: Good point.  Overlooked this.

lib/Logger/LogContext.cpp
mpoeter: Good idea, will do!

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    ScopedValue(std::shared_ptr<Values>&& v) {
+      appendEntry(std::make_shared<EntryImpl<std::shared_ptr<Values>>>(std::move(v)));
+    }
+
+    ~ScopedValue();
+  private:
+    void appendEntry(std::shared_ptr<Entry> e) {
+      _oldTail = e.get();
```
jsteemann, line 130: This cannot be set in non-maintainer mode, because `_oldTail` is not defined then.
Should be wrapped into an `#ifdef`.

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,83 @@
...
+LogContext::ScopedValue::~ScopedValue() {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+  TRI_ASSERT(_oldTail == LogContext::current()._tail.get());
+#endif
+  LogContext::current().popTail();
+}
+
+LogContext::ScopedContext::ScopedContext(LogContext ctx) {
+  if (ctx._tail != LogContext::current()._tail) {
+    _oldTail = ctx._tail;
```
jsteemann, line 47: This cannot be set in non-maintainer mode, so it should be wrapped into an `#ifdef`.

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,214 @@
...
+      LogContext::current().pushEntry(std::move(e));
+    }
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+    Entry* _oldTail;
```
jsteemann, line 134: Looks safer to me if we initialize it to a defined value rather than leaving it undefined.
```suggestion
    Entry* _oldTail = nullptr;
```

lib/Logger/LogContext.cpp
```diff
@@ -0,0 +1,83 @@
...
+  if (entry != nullptr) {
+    doVisit(visitor, entry->_prev);
+    entry->visit(visitor);
+  }
+}
+
+LogContext& LogContext::current() { return localLogContext; }
+
+void LogContext::setCurrent(LogContext ctx) {
+  localLogContext = ctx;
```
jsteemann, line 82:
```suggestion
  localLogContext = std::move(ctx);
```

arangod/GeneralServer/RestHandler.h
```diff
@@ -200,6 +201,13 @@ class RestHandler : public std::enable_shared_from_this<RestHandler> {
   void compressResponse();
 
  protected:
+  // This alias allows the RestHandler and derived classes to add values to the LogContext.
+  // The intention behind RestHandler beeing a friend of LogContext::Accessor and defining
```
jsteemann, line 205:
```suggestion
  // The intention behind RestHandler being a friend of LogContext::Accessor and defining
```

arangod/RestHandler/RestVocbaseBaseHandler.cpp
```diff
@@ -227,6 +228,15 @@ RestVocbaseBaseHandler::RestVocbaseBaseHandler(application_features::Application
 
 RestVocbaseBaseHandler::~RestVocbaseBaseHandler() = default;
 
+void RestVocbaseBaseHandler::prepareExecute(bool isContinue) {
+  //_dbNameContext = LogContext::current().push(LogContext::makeValue().with<logContextKeyDatabaseName>(_vocbase.name()));
```
jsteemann, line 232: I think right now we only want to merge the infrastructure for context logging into devel, but we haven't defined (exactly) what values to log.
So I am fine with commenting this out. However, we should add a code comment here that explains the reason behind nothing being added here (likely "TODO: decide about which values to log in a future step").

arangod/RestHandler/RestVocbaseBaseHandler.h
```diff
@@ -219,6 +222,8 @@ class RestVocbaseBaseHandler : public RestBaseHandler {
 
   /// @brief the vocbase, managed by VocbaseContext
   TRI_vocbase_t& _vocbase;
+  
+  LogContext::Entry* _dbNameContext;
```
jsteemann, line 226: As far as I can see, this variable is neither initialized in the header file nor in the constructor, so it will have an undefined value. We should initialize it to a nullptr!

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,559 @@
...
+ protected:
+  virtual void release(EntryCache& cache) noexcept {
+    this->~Entry();
+    ::operator delete(this);
+  }
+
+ private:
+  void incRefCnt() noexcept {
+    TRI_ASSERT(_refCount.load() > 0);
+    _refCount.fetch_add(1, std::memory_order_acquire);
```
jsteemann, line 280: Instead of two separate load and add operations, we could fuse it to use a single add operation.
For release mode, this won't make a difference though, so feel free to reject the suggestion.
```suggestion
    [[maybe_unused]] auto old = _refCount.fetch_add(1, std::memory_order_acquire);
    TRI_ASSERT(old > 0);
```

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,559 @@
...
+  }
+  [[nodiscard]] std::size_t decRefCnt() noexcept {
+    TRI_ASSERT(_refCount.load() > 0);
+    return _refCount.fetch_sub(1, std::memory_order_relaxed);
```
jsteemann, line 284: Same here:
```suggestion
    auto old = _refCount.fetch_sub(1, std::memory_order_relaxed);
    TRI_ASSERT(old > 0);
    return old;
```

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,559 @@
...
+  }
+
+  friend class LogContext;
+  std::atomic<std::size_t> _refCount{0};
+  Entry* _prev;
```
jsteemann, line 289: `_prev` is only initialized when `pushEntry` is called. But I think it would be reasonable to initialize it to a nullptr here too, so the object is always in a defined state.

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,559 @@
...
+  explicit EntryImpl(Vals&& v) : _values(std::move(v)) {}
+  void visit(Visitor const& visitor) const override { _values->visit(visitor); }
+  void release(EntryCache& cache) noexcept override;
+
+ private:
+  Vals _values;
+};
+
+struct LogContext::EntryCache {
+  struct CachedEntryAlloc {
```
jsteemann, line 303: A `CachedEntryAlloc` instance can be created without specifying a value for `next`, so it will have an undefined value. What about making construction safer by requiring to set the `next` value in the constructor?
```
struct CachedEntryAlloc {
  explicit CachedEntryAlloc(CachedEntryAlloc* n) noexcept { next = n; }
  CachedEntryAlloc* next;
};

File: lib/Logger/LogContext.h [hunk 21]
```diff
@@ -0,0 +1,559 @@
...
+
+template <class Vals>
+inline void LogContext::EntryImpl<Vals>::release(EntryCache& cache) noexcept {
+  if (sizeof(*this) <= LogContext::EntryCache::MinEntryCacheSize && !cache.isFull()) {
+    this->~EntryImpl();
+    cache.addToCache(this);
+  } else {
+    this->~EntryImpl();
+    ::operator delete(this);
+  }
```
jsteemann, line 433: It looks like we are calling `this->~EntryImpl();` unconditionally:
```suggestion
  this->~EntryImpl();
  if (sizeof(*this) <= LogContext::EntryCache::MinEntryCacheSize && !cache.isFull()) {
    cache.addToCache(this);
  } else {
    ::operator delete(this);
  }
```

lib/Logger/LogContext.h
```diff
@@ -0,0 +1,559 @@
...
+inline LogContext::ValueBuilder<> LogContext::makeValue() noexcept {
+  return ValueBuilder<std::tuple<>>({});
+}
+
+inline LogContext& LogContext::current() noexcept {
+  return _threadControlBlock._logContext;
+}
+
+inline LogContext::Entry* LogContext::pushEntry(std::unique_ptr<Entry> entry) noexcept {
+  entry->_prev = _tail;
```
jsteemann, line 523: Shall we add an assertion here that `entry`'s `_refCount` value is currently 0?

lib/Logger/LogContext.h
mpoeter: I prefer to have the assertion before the fetch_add so that the variable remains unchanged in case the assertion fails.

lib/Logger/LogContext.h
mpoeter: I prefer to have the assertion before the fetch_add so that the variable remains unchanged in case the assertion fails.

jsteemann (approved):
LGTM!

PR #14580, opened 2021-08-02
Bug fix 3.8/UI https geo tile loading

hkernbach:
### Scope & Purpose

Web UI: Fixes the loading of map tiles which are being used to display the
  query output based on a world map when using SSL encryption. This lead to
  not displaying some world map tiles correctly (OASIS-590).

#### Backports:

- [x] Backports required for: All supported versions (3.7/3.8/devel)

#### Related Information

- [x] GitHub issue / Jira ticket number: (OASIS-590)

### Testing & Verification


js/apps/system/_admin/aardvark/APP/frontend/js/lib/tile.stamen.js
```diff
@@ -44,7 +44,7 @@ var SUBDOMAINS = "a. b. c. d.".split(" "),
         }
     };
 
-PROVIDERS["terrain-classic"].url = "http://{S}tile.stamen.com/terrain/{Z}/{X}/{Y}.png";
+PROVIDERS["terrain-classic"].url = "//stamen-tiles-{S}a.ssl.fastly.net/terrain/{Z}/{X}/{Y}.png";
```
pluma, line 47: Is there any reason not to always use https rather than protocol-relative? Does the server even serve plaintext HTTP?

hkernbach: That's just how it's officially being used.

pluma: Fine by me, but I think this should probably be https since otherwise it will use http unless aardvark is accessed over https, which could pose minor security implications.

pluma (approved):

ajurdak (commented):
LGTM

PR #14598, opened 2021-08-06
Feature/hybrid smart graph modify graph

mchacki:
Intermediate Commit.

* Adding tests on creation / modification API of hybrid-smart graphs.
* Adding implementation on creation / modification API of hybrid-smart-graphs


arangod/Graph/GraphManager.cpp
```diff
@@ -227,62 +188,88 @@ Result GraphManager::checkForEdgeDefinitionConflicts(std::map<std::string, EdgeD
...
       }
+      existentDocumentCollections.emplace(col);
+    } else if (!res.is(TRI_ERROR_ARANGO_DATA_SOURCE_NOT_FOUND)) {
+      return res;
     } else {
-      auto res = graph.validateCollection(*def.get());
-      if (res.fail()) {
-        return res;
+      if (edgeCollectionsToCreate.find(vertexColl) == edgeCollectionsToCreate.end()) {
+        // We can make this EE Code
```
hkernbach, line 257: Let's create a Kanban card for that before we merge, so this will not be forgotten.

tests/js/common/shell/shell-general-graph.js
```diff
@@ -278,7 +278,7 @@ function GeneralGraphCreationSuite() {
       var g2 = graph._graph(gN2);
       g2._addVertexCollection("mj7");
 
-      graph._renameCollection("mj7", "MarcelJansen");
+      db._collection("mj7").rename("MarcelJansen");
```
hkernbach, line 281: Perfect, this must stay :P

hkernbach (approved):
LGTM .

arangod/Graph/GraphManager.cpp
mchacki:
```suggestion
```

mchacki: Good catch, I was talking about the Method called below here, which is already EE.

PR #14626, opened 2021-08-11
[R2] Replicated Log Compaction

maierlars:
### Scope & Purpose

Add compaction process to the replication 2.0.


arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -204,13 +215,26 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
     }
 
     auto action = std::invoke([&]() noexcept -> DeferredAction {
+      TRI_ASSERT(req.largestCommonIndex >= self->_largestCommonIndex);
+      if (self->_largestCommonIndex > req.largestCommonIndex) {
+        LOG_CTX("fc467", TRACE, self->_follower._loggerContext)
+            << "largest common index went from " << self->_largestCommonIndex
+            << " to " << req.largestCommonIndex << ".";
```
mpoeter, line 222: We first have an assertion that `req.largestCommonIndex >= self->_largestCommonIndex` and then check if `self->_largestCommonIndex > req.largestCommonIndex`, so the inverse of the previous assertion. I take it this situation is unexpected (otherwise we would not have an assertion), but then IMHO the log message should have at least WARN level. Am I missing something?

maierlars: 😆  I think my intention was to log that message and check compaction if the lci became bigger. I just screwed up the condition. It should be the strict version of the assertion.

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -423,14 +448,45 @@ replicated_log::LogFollower::~LogFollower() {
...
   result.commitIndex = _commitIndex;
-  result.spearHead.index = _inMemoryLog.getLastIndex();
-  result.spearHead.term = _inMemoryLog.getLastTerm();
+  result.firstIndex = _inMemoryLog.getFirstIndex();
+  result.spearHead = _inMemoryLog.getLastTermIndexPair();
   return result;
 }
+
+auto LogFollower::GuardedFollowerData::checkCompaction() -> Result {
+  auto const compactionStop = std::min(_largestCommonIndex, _releaseIndex);
```
maierlars, line 473:
```suggestion
  auto const compactionStop = std::min(_largestCommonIndex, _releaseIndex + 1);
```

goedderz (approved):

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -822,6 +832,41 @@ replicated_log::LogLeader::GuardedLeaderData::GuardedLeaderData(replicated_log::
...
+      return {};
+    }
+    self._releaseIndex = doneWithIdx;
+    LOG_CTX("a0c95", TRACE, _logContext) << "new release index set to " << self._releaseIndex;
+    return self.checkCompaction();
+  });
+}
+
+auto replicated_log::LogLeader::GuardedLeaderData::checkCompaction() -> Result {
+  auto const compactionStop = std::min(_largestCommonIndex, _releaseIndex);
```
mpoeter, line 848: Didn't we want to change this to `_releaseIndex + 1`?

mpoeter (approved):

arangod/Replication2/ReplicatedLog/LogLeader.cpp
maierlars: We did on the follower side. I forgot the leader :headdesk:
```suggestion
  auto const compactionStop = std::min(_largestCommonIndex, _releaseIndex+1);
```

PR #14659, opened 2021-08-18
Revert JSUnity improvement

mchacki:
We now need to have unique testCase names whenever we have multiple suites per file again.
Sorry.


jsteemann (approved):
LGTM!

PR #14667, opened 2021-08-19
[R2] Log Multiplexer

maierlars:
### Scope & Purpose

Log Multiplexer implementation.Used to multiplex multiple typed stream over a replicated log.

### DIFF
--- a/arangod/Replication2/Streams/LogMultiplexer.h
+++ b/arangod/Replication2/Streams/LogMultiplexer.h
@@ -34,10 +34,9 @@ struct LogMultiplexerStreamDispatcher : std::enable_shared_from_this<Self>,
     return getStreamByDescriptor<Descriptor>();
   }
 
-  template <StreamId Id, typename Descriptor = stream_descriptor_by_id_t<Id, Spec>,
-            typename Type = stream_descriptor_type_t<Descriptor>>
-  auto getStreamById() -> std::shared_ptr<StreamType<Type>> {
-    return getStreamByDescriptor<Descriptor>();
+  template <StreamId Id>
+  auto getStreamById() -> std::shared_ptr<StreamType<stream_type_by_id_t<Id, Spec>>> {
+    return getStreamByDescriptor<stream_descriptor_by_id_t<Id, Spec>>();
   }
 
   template <typename Descriptor>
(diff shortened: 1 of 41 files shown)


arangod/Replication2/Streams/LogMultiplexer.h
```diff
@@ -0,0 +1,85 @@
...
+ * Common stream dispatcher class for Multiplexer and Demultiplexer. You can
+ * obtain a stream given its id using getStreamById. Alternatively, you can
+ * static_cast the a pointer to StreamBase<Descriptor> for the given stream.
+ * @tparam Self
+ * @tparam Spec
+ * @tparam StreamType
+ */
+template <typename Self, typename Spec, template <typename> typename StreamType>
+struct LogMultiplexerStreamDispatcher : std::enable_shared_from_this<Self>,
+                                        StreamDispatcherBase<Spec, StreamType> {
```
mpoeter, line 30: For the CRTP parameter I usually prefer the name `Derived`, because IMHO it communicates more clearly how this class needs to be used:
```suggestion
template <typename Derived, typename Spec, template <typename> typename StreamType>
struct LogMultiplexerStreamDispatcher : std::enable_shared_from_this<Derived>,
                                        StreamDispatcherBase<Spec, StreamType> {
```

arangod/Replication2/Streams/LogMultiplexer.h
```diff
@@ -0,0 +1,85 @@
...
+  auto getStreamBaseById()
+      -> std::shared_ptr<StreamGenericBase<Descriptor, StreamType>> {
+    return getStreamByDescriptor<Descriptor>();
+  }
+
+  template <StreamId Id, typename Descriptor = stream_descriptor_by_id_t<Id, Spec>,
+            typename Type = stream_descriptor_type_t<Descriptor>>
+  auto getStreamById() -> std::shared_ptr<StreamType<Type>> {
+    return getStreamByDescriptor<Descriptor>();
+  }
```
mpoeter, line 41: It is a bit strange the these functions are named `ById`, but the result actually depends on the `Descriptor`. By default the descriptor is based on the Id, but if the descriptor is passed explicitly, then the `ById` part is confusing/misleading. I would either strip the Descriptor parameter or rename the function.

PR #14674, opened 2021-08-23
Protocol specific metrics

neunhoef:
Add protocol specific metrics.

This adds histograms about the body sizes of requests, distinguishing
HTTP/2, VST and HTTP/1.1 requests.

Also: Adds metrics about total number of accepted HTTP/2 and VST
connections.

Also: Switches metrics tests to /_admin/metrics/v2 in more tests.

### Scope & Purpose

- [*] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)

#### Backports:

- [*] Backports required for: 3.8 and 3.7

### Testing & Verification

### Documentation

- [*] Added a new section in the *Manual*  (metrics, auto-generated)


tests/js/client/shell/shell-check-metrics-move-noncluster.js
```diff
@@ -0,0 +1,112 @@
...
+        assertEqual(vstReqCount, vstReqCount2);
+        assertEqual(vstReqSum, vstReqSum2);
+      }
+      if (arango.protocol() === "http") {
+        assertNotEqual(http1ReqCount, http1ReqCount2);
+        assertNotEqual(http1ReqSum, http1ReqSum2);
+      } else {
+        assertEqual(http1ReqCount, http1ReqCount2);
+        assertEqual(http1ReqSum, http1ReqSum2);
+      }
```
jsteemann, line 81: If the protocol used during the test is "http", we will count our own requests (executed by this test) plus all requests that are sent to the server by other tasks (or internally). I think this only works because the test is named "-noncluster". If this test would run in cluster, too, all the cluster-internal requests would be counted as well and the test wouldn't be deterministic. 
So I guess it works, but we should add a comment to the test that this works on in non-cluster mode because otherwise cluster-internal requests would be counted.

tests/js/client/shell/shell-check-metrics-move-noncluster.js
```diff
@@ -0,0 +1,112 @@
...
+      let http2ConnCount = getMetric("arangodb_http2_connections_total");
+      let vstConnCount = getMetric("arangodb_vst_connections_total");
+      let allConnCount = getMetric("arangodb_client_connection_statistics_connection_time_count");
+      // I have not found a reliable way to trigger a new connection from
+      // `arangosh`. `arango.reconnect` does not work since it is caching
+      // connections and the request timeout is not honoured for HTTP/2
+      // and VST by fuerte. The idle timeout runs into an assertion failure.
+      // Therefore, we must be content here to check that the connection
+      // count is non-zero for the currently underlying protocol:
+      //assertNotEqual(0, allConnCount);
```
jsteemann, line 101: `allConnCount` is not used anymore by the test. Shall we remove it?

jsteemann (approved):
LGTM

tests/js/client/shell/shell-check-metrics-move-noncluster.js
neunhoef: I will fix this. My understanding was that this metric counts all connections, but this does not seem to be true. I will investigate and either delete or fix.

tests/js/client/shell/shell-check-metrics-move-noncluster.js
neunhoef: Right, I will add a comment.

neunhoef: Done.

tests/js/client/shell/shell-check-metrics-move-noncluster.js
neunhoef: Done.

PR #14676, opened 2021-08-23
TRI_ASSERT Log Stream

maierlars:
### Scope & Purpose

Allow to print additional information in case an assertion fails:
```
  TRI_ASSERT(_commitIndex < newIndex)
      << "_commitIndex == " << _commitIndex << ", newIndex == " << newIndex;
```
Will produce the following output:
```
assertion failed in /work/ArangoDB/arangod/Replication2/ReplicatedLogs/LogLeader.cpp:667 [LogLeader::checkCommitIndex]: _commitIndex < newIndex ; _commitIndex = 12, newIndex == 8
```

I made sure that no string is build if the the condition is satisfied or the maintainer mode is disabled. [See more.](https://godbolt.org/z/85cbfjavG)

#### Backports:

- [x] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

I will add tests to ensure, that this also works with types that have a custom `operator<<`.

### DIFF
--- a/lib/Basics/debugging.h
+++ b/lib/Basics/debugging.h
@@ -26,6 +26,7 @@
 
 #include <stdlib.h>
 #include <ostream>
+#include <sstream>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -199,6 +200,33 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
   return o;
 }
 
+namespace debug {
+
+struct NoOpStream {
+  template <typename T>
+  auto operator<<(T const&) -> NoOpStream& {
+    return *this;
+  }
+};
+
+struct AssertionLogger {
+  void operator&(std::stringstream const& str) {
+    std::string message = str.str();
+    arangodb::CrashHandler::assertionFailure(file, line, function, expr,
+                                             message.empty() ? nullptr : message.c_str());
+  }
+  void operator&(std::ostream const& os) {
+    operator&(static_cast<std::stringstream const&>(os));
+  }
+
+  const char* file;
+  int line;
+  const char *function;
+  const char* expr;
+};
+
+}  // namespace debug
+
 }  // namespace arangodb
 
 /// @brief assert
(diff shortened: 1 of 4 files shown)


lib/Basics/debugging.h
```diff
@@ -199,31 +200,51 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
...
-    if (!(ADB_LIKELY(expr))) {                                                           \
-      arangodb::CrashHandler::assertionFailure(__FILE__, __LINE__, __FUNCTION__, #expr); \
-    } else {                                                                             \
-    }                                                                                    \
-  } while (false)
+#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/                                        \
+  (!(ADB_LIKELY(expr)))                                                             \
+      ? (void)nullptr                                                               \
+      : arangodb::debug::AssertionLogger{__FILE__, __LINE__, __FUNCTION__, #expr} & \
+            (std::stringstream{})
```
jsteemann, line 241: Shouldn't this be the other way around, i.e. if 
```suggestion
  (ADB_LIKELY(expr))                                                                \
      ? (void)nullptr                                                               \
      : arangodb::debug::AssertionLogger{__FILE__, __LINE__, __FUNCTION__, #expr} & \
            (std::stringstream{})
```

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16717/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16719/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16721/

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16723/

lib/Basics/debugging.h
```diff
@@ -199,31 +200,54 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
   return o;
 }
 
+namespace debug {
+struct NoOpStream {
+  template <typename T>
+  auto operator<<(T const&) -> NoOpStream& {
```
jsteemann, line 206: shall we explicitly mark this as noexcept?

lib/Basics/debugging.h
```diff
@@ -199,31 +200,54 @@ enable_if_t<is_container<T>::value, std::ostream&> operator<<(std::ostream& o, T
...
 
-#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/ \
-  do {                                       \
-    if (false) {                             \
-      (void)(expr);                          \
-    }                                        \
-  } while (false)
+#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/        \
+  (true) ? ((false) ? (void)(expr) : (void)nullptr) \
+         : ::arangodb::debug::AssertionLogger{} & ::arangodb::debug::NoOpStream {}
```
jsteemann, line 249: Just wondering why we need to create a default-constructed AssertionLogger here? Can't we get away without it in non-maintainer mode?

maierlars: To be clear, we are talking about dead code here. It is there, because we need to cast to void, but a void cast `(void)` has a higher precedence than `<<`. Thus this is done using `operator&`. Exactly as we do in the logger macros.

mpoeter: Do we need `epxr` here? I know that the old code also used it, but is it necessary? The problem is that even though the expression is never evaluated, it must at least compile. I only just had a case where I had a member that was only available in maintainer mode which was used in an assertion. So I had to wrap the assertion in yet another ifdef because the compiler would complain that it does not know the member used in the expression.

mpoeter (approved):

lib/Basics/debugging.h
jsteemann: Yes, we need `expr` here!
The reason is that some variables may only be used in assertions, and the compiler will complain about unused variables if the assertion macro just swallows their (sole) usage.

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16731/

lib/Basics/debugging.h
mpoeter: Makes sense (unfortunately).

maierlars:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/16736/

goedderz (approved):

PR #14701, opened 2021-08-25
prototype for forceOneShardAttributeValue

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/744

(Enterprise Edition only): added query option `forceOneShardAttributeValue` to explicitly set a shard key value that will be used during query snippet distribution to limit the query to a specific server in the cluster.

This query option can be used in complex queries in case the query optimizer cannot automatically detect that the query can be limited to only a single server (e.g. in a disjoint smart graph case).
When the option is set to the correct shard key value, the query will be limited to the target server determined by the shard key value. It thus requires that all collections in the query use the same distribution (i.e. `distributeShardsLike` attribute via disjoint SmartGraphs).
  
Limiting the query to a single DB server is a performance optimization and may make complex queries run a lot faster because of the reduced setup and teardown costs and the reduced cluster-internal traffic during query execution.

If the option is set incorrectly, i.e. to a wrong shard key value, then the query may be shipped to a wrong DB server and may not return results (i.e. empty result set). It is thus the caller's responsibility to set the `forceOneShardAttributeValue` correctly or not use it.

The `forceOneShardAttributeValue` option will only honor string values. All other values as well as the empty string will be ignored and treated as if the option is not set.

If the option is set and the query satisfies the requirements for using the option, the query's execution plan will contain the "cluster-one-shard" optimizer rule.

#### Backports:

#### Related Information

- [x] Docs PR: TO BE DONE
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/744

### Testing & Verification

  - [x] Added new **integration tests** (shell_server, in enterprise repository)


arangod/Aql/ShardLocking.cpp
```diff
@@ -44,6 +45,49 @@ std::unordered_set<ShardID> const ShardLocking::EmptyShardListUnordered{};
...
+    auto errorCode = TRI_ERROR_NO_ERROR;
+    if (col->isDisjoint()) {
+      // if disjoint smart edge collection, we must insert an
+      // artifical key with two colons, to pretend it is a real
+      // smart graph key
+      errorCode = col->getCollection()->getResponsibleShard(
+          forceOneShardAttributeValue + ":test:" + forceOneShardAttributeValue, shardId);
+    } else {
+      auto const& shardKeys = col->getCollection().get()->shardKeys();
+      TRI_ASSERT(shardKeys.size());
```
mchacki, line 67:
```suggestion
      TRI_ASSERT(shardKeys.size() > 0);
```

mchacki:
```suggestion
      TRI_ASSERT(!shardKeys.empty());
```

mchacki (approved):
LGTM.
Just one Assertion.  It is correct, however it could be made easier to understand

PR #14707, opened 2021-08-26
protoype for forceOneShardAttributeValue

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/14701

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/745

(Enterprise Edition only): added query option `forceOneShardAttributeValue` to explicitly set a shard key value that will be used during query snippet distribution to limit the query to a specific server in the cluster.

This query option can be used in complex queries in case the query optimizer cannot automatically detect that the query can be limited to only a single server (e.g. in a disjoint smart graph case).
When the option is set to the correct shard key value, the query will be limited to the target server determined by the shard key value. It thus requires that all collections in the query use the same distribution (i.e. `distributeShardsLike` attribute via disjoint SmartGraphs).
  
Limiting the query to a single DB server is a performance optimization and may make complex queries run a lot faster because of the reduced setup and teardown costs and the reduced cluster-internal traffic during query execution.

If the option is set incorrectly, i.e. to a wrong shard key value, then the query may be shipped to a wrong DB server and may not return results (i.e. empty result set). It is thus the caller's responsibility to set the `forceOneShardAttributeValue` correctly or not use it.

The `forceOneShardAttributeValue` option will only honor string values. All other values as well as the empty string will be ignored and treated as if the option is not set.

If the option is set and the query satisfies the requirements for using the option, the query's execution plan will contain the "cluster-one-shard" optimizer rule.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: TO BE DONE
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/745

### Testing & Verification

  - [x] Added new **integration tests** (shell_server, in enterprise repository)


hkernbach:
Jenkins run green

CHANGELOG
```diff
@@ -1,6 +1,50 @@
...
+  All other values as well as the empty string will be ignored and treated 
+  as if the option is not set.
+
+  If the option is set and the query satisfies the requirements for using
+  the option, the query's execution plan will contain the "cluster-one-shard"
+  optimizer rule.
+
+* Reduce internal priority of AQL execution. This prevents possible deadlocks
+  with modification operations in a cluster and replicationFactor >= 2, and can
+  also improve responsiveness under high load of AQL queries.
```
mchacki, line 46: AFAIR this is a different bugfix. Where we have a separate PR for.
We should cross check to avoid merge-conflicts or differences with the separate Bugfix

arangod/Aql/RestAqlHandler.cpp
```diff
@@ -851,3 +851,11 @@ RestStatus RestAqlHandler::handleFinishQuery(std::string const& idString) {
...
   }));
 }
+
+RequestLane RestAqlHandler::lane() const {
+  if (ServerState::instance()->isCoordinator()) {
+    return RequestLane::CLUSTER_AQL_INTERNAL_COORDINATOR;
+  } else {
+    return RequestLane::CLUSTER_AQL;
+  }
+}
```
mchacki, line 861: This is also part of the other bug-fix I mentioned above

arangod/Aql/RestAqlHandler.h
```diff
@@ -44,7 +44,7 @@ class RestAqlHandler : public RestVocbaseBaseHandler {
 
  public:
   char const* name() const override final { return "RestAqlHandler"; }
-  RequestLane lane() const override final { return RequestLane::CLUSTER_AQL; }
+  RequestLane lane() const override final;
```
mchacki, line 47: Also other Bug-fix

arangod/Aql/SharedQueryState.cpp
```diff
@@ -129,34 +130,37 @@ void SharedQueryState::queueHandler() {
...
+        }
+      } else {
+        return;
+      }
+      lck.unlock();
+    } while (true);
+
+    TRI_ASSERT(lck);
+    self->queueHandler();
+  });
```
mchacki, line 163: This is also the other bug-fix

arangod/GeneralServer/RequestLane.h
```diff
@@ -62,15 +62,16 @@ enum class RequestLane {
...
-  // DBserver to the Coordinator or from the Coordinator to the DBserver.
-  // These have High priority.
-  CLUSTER_AQL_CONTINUATION,
+  // For requests from the DBserver to the Coordinator, and continuations on the
+  // Coordinator.
+  // These have medium priority. Because client requests made against the
+  // RestCursorHandler (with lane CLIENT_AQL) might block and need these to
+  // finish. Ongoing low priority requests can also prevent low priority lanes
+  // from being worked on, having the same effect.
+  CLUSTER_AQL_INTERNAL_COORDINATOR,
```
mchacki, line 74: This file as well

tests/js/client/server_parameters/test-statistics-enabled-false.js
```diff
@@ -0,0 +1,88 @@
...
+if (getOptions === true) {
+  return {
+    'server.statistics': "false",
+    'server.export-metrics-api': "true",
+  };
+}
+
+const jsunity = require('jsunity');
+
+function testSuite() {
```
mchacki, line 40: How is this test related with this PR at all?

mchacki (commented):
I think the Part that is supposed for this forceOneShard is good. (Would approve)

However this PR seems to contain unrelated parts (especially request lanes in AQL)
I just want to avoid merging issues for the other PR, or having something included unter the radar). So maybe we need to make sure to merge in correct order.
If this comment is sorted out count this as approved!

CHANGELOG
```diff
@@ -20,6 +20,50 @@ v3.8.2 (XXXX-XX-XX)
...
```
jsteemann, line 65:
```suggestion
```

mchacki (approved):
LGTM.
My last comments have been sorted out, as the unrelated changes are part of BaseBranch now :+1:

PR #14710, opened 2021-08-27
Proposal to enable golang for integration tests

neunhoef:
This is a proposal. The idea is to be able to formulate integration
tests in golang using the go-driver. This has a few benefits:

 - tests with a high performance client are possible
 - tests with all protocols HTTP/1, HTTP/2 and VST, with and without
   encryption can be formulated, and a single test can even mix
   different protocols (although not with and without encryption)
 - tests with high parallelism are possible and convenient

How does it work?

There is one more testsuite called "gotests", which look for files
under `tests/go` with a file name beginning with `test_` and ending
with `.go`. Each such file is run as a test in this way:

```
  cd tests/go
  go run test_version.go helper.go
```

That is, the file has access to the go driver and to some convenience
routines in `tests/go/helper.go` but must otherwise compile to a
standalone binary with a single `main` function.

The executable is run and the test is successful if and only if the 
exit code is 0.

The test code can write to the console whatever it wants.

This PR includes three example test files:

  - `tests/go/test_version.go`: simple example just checking the version
  - `tests/go/test_protocol_metrics.go`: test if some new metrics
    actually advance if connections and requests with different
    protocols are sent
  - `tests/go/test_parallelism.go`: an example test using 100 go
    routines and very high parallelism, this also shows the performance,
    it can insert 10M documents in <10s in debug mode

Disadvantages:

 - We have one more language in our zoo.
 - We need to install golang in the oskar build and test images.
 - We need to deploy golang in our jenkins Mac and Windows machines.

### DIFF
--- a/js/client/modules/@arangodb/testsuites/gotests.js
+++ b/js/client/modules/@arangodb/testsuites/gotests.js
@@ -125,8 +125,8 @@ function goTests (options) {
     let pos = file.lastIndexOf("/");
     let basename = file.slice(pos+1);
     let path = file.slice(0, pos);
-    args.push(basename)
-    args.push("helpers.go")
+    args.push(basename);
+    args.push("helpers.go");
     if (options.extremeVerbosity) {
       print(process.env);
       print(args);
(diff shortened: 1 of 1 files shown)


neunhoef:
This met enough opposition so that I have lost the appetite to follow up on it.

PR #14722, opened 2021-09-01
Bug fix/search 238 late materialization with constrained heap

Dronplane:
### Scope & Purpose

PR modifies distributeSortToClusterRule optimizer rule. While moving SORT node to the DBServer  insertion point is not right after REMOTE node but rule tries to skip as much CALCULATION nodes as possible (e.g. if calculation results are not used in sort). This is similar to what is done by moveCalculationsDownRule.
That will make it possible to use late materialization. Also SORT could be constrained so less calculations will be executed.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/tests/js/server/aql/aql-optimizer-rule-late-document-materialization-arangosearch.inc
+++ b/tests/js/server/aql/aql-optimizer-rule-late-document-materialization-arangosearch.inc
@@ -380,11 +380,11 @@
         });
         assertEqual(0, expected.size);
       },
-	  testConstrainedSortOnDbServer() {
-	    let query = "FOR d IN " + vn  + " SEARCH d.value > 9 SORT BM25(d) LIMIT 10 RETURN {key: d._key, value: d.value}";
-		let plan = AQL_EXPLAIN(query).plan;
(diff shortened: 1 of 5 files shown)


tests/js/server/aql/aql-optimizer-rule-late-document-materialization-arangosearch.inc
```diff
@@ -380,6 +380,30 @@
...
         assertEqual(0, expected.size);
       },
+      testConstrainedSortOnDbServer() {
+        let query = "FOR d IN " + vn  + " SEARCH d.value > 9 SORT BM25(d) LIMIT 10 RETURN {key: d._key, value: d.value}";
+        let plan = AQL_EXPLAIN(query).plan;
+        assertNotEqual(-1, plan.rules.indexOf(ruleName));
+        let materializeNodeFound = false;
+        let nodeDependency = null;
+        plan.nodes.forEach(function(node) {
+          if( node.type === "MaterializeNode") {
```
gnusi, line 390: maybe check the whole plan?

gnusi (approved):

PR #14723, opened 2021-09-01
BTS-697: Eradicate LngLatRects from Geo code

neunhoef:
This PR addresses some issues with the geo code:

 1. S2LatLngRects are not useful as optimization for general polygons,
    since their boundaries are circles of latitude rather than great
    circles. This PR removes the special detection (whose detection was
    wrongly implemented anyway). This also removes code for "intersects"
    and "contains" for S2LatLngRects, which was wrongly implemented.
    For legacy indexes and -behaviour, see below.

 2. Loops were on principle normalized during the parsing of polygons.
    As a consequence it was impossible to describe regions which cover
    more than half of the earth, although GeoJSON and S2 are able to
    do this. Normalization is switched off in this PR.

 3. Literally lots of bugs found in special situations.

This PR needs: https://github.com/arangodb/enterprise/pull/746 in the enterprise repository.

The fundamental problem with this is that old geoindexes have used the old parsing to index geo data. If we now upgrade to a new version without dropping the indexes, we end up with internal errors, since if somebody removes a document and we want to take its cells out of the index, we get different cells than when we inserted and this is a mess.

Therefore, we also needed to do the following things:

- duplicate the code for geojson parsing (essentially the ShapeContainer class in lib/Geo/ShapeContainer.* , keep one copy with the old parsing behaviour (and the wrong contains and intersects code for it), and a the new version 
 - adjust all the testing, probably by testing the new behaviour but keeping the old unit tests for ShapeContainer
 - create a new flag for geo indexes `legacyPolygons` which, if not set, uses under the hood the new ShapeContainer
 - let the old type of geo index use the legacy ShapeContainer
 - add additional tests for the different geo index flag
 - document this whole business
 - carefully word the release notes for 3.10 w.r.t. this business
 - fix lots of bugs with the old implementation
 - use a different algorithm for pure `GEO_CONTAINS` and `GEO_INTERSECTS` queries.
 - add validation for `GEO_POLYGON` and `GEO_MULTIPOLYGON` in AQL.
 - better define what is valid input for polygons and multipolygons, be more tolerant at times.
 - document rules for polygons thoroughly.


arangod/GeoIndex/Covering.h
```diff
@@ -0,0 +1,124 @@
...
+
+  LocalDocumentId const& getNext() const {
+    TRI_ASSERT(hasNext());
+    return _buffer.front();
+  }
+
+  inline void next() {
+    TRI_ASSERT(hasNext());
+    _buffer.pop_front();
+  }
```
maierlars, line 75: Is there a reason why we use `pop_front` instead of maintaining an index? From what I can tell, it looks like that `_buffer` is filled once with everything that is found in one interval and then consumed completely. One could even use a vector here instead of a deque, right?

maierlars (approved):

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -263,6 +420,205 @@ class RDBNearIterator final : public IndexIterator {
...
+
+      if (seek) {  // try to avoid seeking at all cost
+        _iter->Seek(bds.start());
+      }
+
+      while (_iter->Valid() && cmp->Compare(_iter->key(), bds.end()) <= 0) {
+        _covering.reportFound(RocksDBKey::indexDocumentId(_iter->key()),
+                              RocksDBValue::centroid(_iter->value()));
+        _iter->Next();
+      }
```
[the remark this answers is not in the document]
neunhoef: We limit to 1024 documents being found (and accepted). We could improve this, but it would be a logic change. I suggest we leave this for a later performance optimization PR.

arangod/GeoIndex/Index.cpp
```diff
@@ -187,7 +197,7 @@ S2LatLng Index::parseGeoDistance(aql::AstNode const* args,
     if (json.isArray() && json.length() >= 2) {
       res = shape.parseCoordinates(json, /*GeoJson*/ true);
     } else {
-      res = geo::geojson::parseRegion(json, shape);
+      res = geo::geojson::parseRegion(json, shape, false);
```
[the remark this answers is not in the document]
neunhoef: See https://arangodb.atlassian.net/browse/APM-457. We merge this PR without the change, but PM to decide about this issue.

lib/Geo/GeoJson.cpp
```diff
@@ -452,36 +482,63 @@ Result parseMultiPolygon(velocypack::Slice const& vpack,
...
-                      "Subsequent loop not a hole in polygon");
+        bool bad = false;
+        if (legacy) {
+          bad = true;
+        } else {
+          loops.back()->Invert();
+          if (!loops[outerLoop]->Contains(loops.back().get())) {
+            bad = true;
+          }
+        }
```
MBkkt, line 516: Same nit
```suggestion
        bool bad = legacy;
        if (!bad) {
          loop->Invert();
          if (!loops[outerLoop]->Contains(loop)) {
            bad = true;
          }
        }
```

arangod/GeoIndex/Covering.h
```diff
@@ -0,0 +1,124 @@
...
+  bool _allIntervalsCovered;
+
+  /// Total number of interval calculations
+  size_t _numScans;
+
+  /// buffer of found documents
+  GeoDocumentsQueue _buffer;
+
+  // deduplication filter
+  std::unordered_set<uint64_t> _seenDocs;
```
MBkkt, line 117: What about using `containers::FlatHashSet` or `containers::NodeHashSet` if you really need pointer stability?

arangod/GeoIndex/Covering.h
```diff
@@ -0,0 +1,124 @@
...
+
+#include <queue>
+#include <vector>
+
+#include "Geo/GeoParams.h"
+#include "Geo/Utils.h"
+#include "VocBase/Identifiers/LocalDocumentId.h"
+
+namespace arangodb {
+namespace geo_index {
```
MBkkt, line 39: NIT
```suggestion
namespace arangodb::geo_index {
```

arangod/GeoIndex/Covering.h
```diff
@@ -0,0 +1,124 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <s2/s2cap.h>
+#include <s2/s2cell_id.h>
+#include <s2/s2region.h>
+#include <s2/s2region_coverer.h>
+
+#include <queue>
```
MBkkt, line 31: Should be deque, not queue, because of you use deque
```suggestion
#include <deque>
```

arangod/GeoIndex/Covering.h
```diff
@@ -0,0 +1,124 @@
...
+class CoveringUtils {
+ public:
+  /// @brief Type of documents buffer
+  typedef std::deque<LocalDocumentId> GeoDocumentsQueue;
+
+  explicit CoveringUtils(geo::QueryParams&& params) noexcept;
+  ~CoveringUtils();
+
+ public:
+  inline geo::FilterType filterType() const { return _params.filterType; }
```
MBkkt, line 53: I don't really like practically useless `inline` keyword usage, why do you want to write this boilerplate?
If you really think it should be inlined then write some always/force inline
If not then allow the compiler to decide, why not?

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -263,6 +420,205 @@ class RDBNearIterator final : public IndexIterator {
...
+    // scanned all intervals delivered by _covering:
+    while (limit > 0 &&
+           (!_covering.isDone() || _scanningInterval < _scan.size())) {
+      if (!_covering.hasNext()) {
+        performScan();
+      }
+
+      while (limit > 0 && _covering.hasNext()) {
+        if (std::forward<F>(cb)(_covering.getNext())) {
+          limit--;
```
MBkkt, line 464: NIT
```suggestion
          --limit;
```

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -263,6 +420,205 @@ class RDBNearIterator final : public IndexIterator {
...
+      // we already know that no further results will be returned by the index
+      return false;
+    }
+
+    // We keep going, until either we have reached our limit or we have
+    // scanned all intervals delivered by _covering:
+    while (limit > 0 &&
+           (!_covering.isDone() || _scanningInterval < _scan.size())) {
+      if (!_covering.hasNext()) {
+        performScan();
```
MBkkt, line 459: Why can't we use `cb` with `limit` in `PerformScan` instead of storing value in covering and only then use it?
Why do we scan more than necessary (if I'm correct)?

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -263,6 +420,205 @@ class RDBNearIterator final : public IndexIterator {
...
+        if (std::forward<F>(cb)(_covering.getNext())) {
+          limit--;
+        }
+        _covering.next();
+      }
+    }
+    return !_covering.isDone() || _scanningInterval < _scan.size();
+  }
+
+  bool nextDocumentImpl(DocumentCallback const& cb, size_t limit) override {
```
MBkkt, line 472: I think if you make a separate function and only then pass it to `read` it will be more readable:
I mean these places:
`nextDocumentImpl` ... `read`
`nextImpl` ... `read`

arangod/GeoIndex/Covering.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <s2/s2region_intersection.h>
+
+#include "Basics/Common.h"
+#include "Geo/GeoParams.h"
+#include "Geo/Utils.h"
+#include "Logger/Logger.h"
+#include "Logger/LogMacros.h"
+
+namespace arangodb {
+namespace geo_index {
```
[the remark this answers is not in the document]
neunhoef: I prefer it this way.

arangod/GeoIndex/Covering.h
neunhoef: I prefer it this way.

arangod/GeoIndex/Covering.h
neunhoef: Good catch!

arangod/GeoIndex/Covering.h
neunhoef: Fine with me. Done.

arangod/GeoIndex/Covering.h
neunhoef: We have agreed to do concentrate on correctness for this PR, it was difficult enough to achieve that. We will do performance improvements in a separate PR. I will put it on the list of possible performance improvements.

lib/Geo/ShapeContainer.cpp
```diff
@@ -241,6 +242,13 @@ bool ShapeContainer::mayIntersect(S2CellId cell) const noexcept {
   return _data->MayIntersect(S2Cell(cell));
 }
 
+static bool isExcessiveLngLat(S1Angle lngsmall, S1Angle lngbig,
+                              S1Angle latsmall, S1Angle latbig) {
+  return fabs(lngbig.radians() - lngsmall.radians()) +
+             fabs(latbig.radians() - latsmall.radians()) >=
```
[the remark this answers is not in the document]
neunhoef: It must be `std::fabs`, but happy to do the change.

lib/Geo/GeoParams.h
```diff
@@ -154,6 +154,24 @@ struct QueryParams {
...
+        "minDistance: " + std::to_string(minDistance) +
+        " incl: " + t(minInclusive) +
+        " maxDistance: " + std::to_string(maxDistance) +
+        " incl: " + t(maxInclusive) + " sorted: " + t(sorted) +
+        " ascending: " + t(ascending) +
+        " origin: " + std::to_string(origin.lng().degrees()) + " , " +
+        std::to_string(origin.lat().degrees()) +
+        " pointsOnly: " + t(pointsOnly) + " limit: " + std::to_string(limit) +
+        " filterType: " + std::to_string((int)filterType) +
+        " filterShape: " + std::to_string((int)filterShape.type());
```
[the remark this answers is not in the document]
neunhoef: Done.

lib/Geo/GeoJson.cpp
```diff
@@ -353,40 +355,68 @@ Result parsePolygon(VPackSlice const& vpack, ShapeContainer& region) {
...
-      return Result(TRI_ERROR_BAD_PARAMETER,
-                    "Subsequent loop not a hole in polygon");
+      bool bad = false;
+      if (legacy) {
+        bad = true;
+      } else {
+        loops.back()->Invert();
+        if (!loops.front()->Contains(loops.back().get())) {
+          bad = true;
+        }
```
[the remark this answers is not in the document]
neunhoef: Done.

lib/Geo/GeoJson.cpp
neunhoef: Done.

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
neunhoef: Done.

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
neunhoef: Performance later.

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
neunhoef: You have a certain point here. However, I don't want to touch the logic here unnecessarily. This code is anyway copy/pasted from the corresponding code in `RDBNearIterator`. So let's get this PR done first and do stylistic improvements some time later.

lib/Geo/GeoJson.cpp
```diff
@@ -353,40 +355,68 @@ Result parsePolygon(VPackSlice const& vpack, ShapeContainer& region) {
...
-    // normalization ensures that point orientation does not matter for Polygon
-    // type the RFC recommends this for better compatibility
-    loop->Normalize();
+    if (legacy) {
+      loop->Normalize();
+    }
 
-    // subsequent loops must be holes within first loop
+    // subsequent loops must be holes within first loop:
     if (loops.size() > 1 && !loops.front()->Contains(loop)) {
```
cpjulia, line 379: does this cover the case in which there is more then one hole inside the outer loop? I mean because of checking if the first loop, which is the outer one, contains the last loop in the vector. Like if there's the outer loop followed by two holes, wouldn't it have to check if loops[1] is contained as well?

lib/Geo/GeoJson.cpp
```diff
@@ -353,40 +355,68 @@ Result parsePolygon(VPackSlice const& vpack, ShapeContainer& region) {
...
-      return Result(TRI_ERROR_BAD_PARAMETER,
-                    "Subsequent loop not a hole in polygon");
+      bool bad = false;
+      if (legacy) {
```
cpjulia, line 381:
```suggestion
      if (legacy || !loops.back()->Contains(loop) {
        bad = true;
      }
      if (!legacy) {
        loop->Invert();
      }
```
maybe something like this to make use of the same statement?

lib/Geo/GeoJson.cpp
```diff
@@ -452,36 +482,63 @@ Result parseMultiPolygon(velocypack::Slice const& vpack,
             TRI_ERROR_BAD_PARAMETER,
             std::string("Invalid loop in polygon: ").append(error.text()));
       }
+      // Note that we are using InitNested for S2Polygon below.
+      // Therefore, we are supposed to deliver all our loops in CCW
+      // convention (aka right hand rule, interiour is to the left of
+      // the polyline).
+      // Since we want to want to allow for loops, whose interiour covers
```
cpjulia, line 489:
```suggestion
      // Since we want to allow for loops, whose interiour covers
```

lib/Geo/GeoJson.cpp
cpjulia: and, if so, the same with parseMultiPolygon

lib/Geo/GeoJson.cpp
```diff
@@ -353,40 +355,68 @@ Result parsePolygon(VPackSlice const& vpack, ShapeContainer& region) {
           std::string("Invalid loop in polygon: ").append(error.text()));
     }
 
+    // Note that we are using InitNested for S2Polygon below.
+    // Therefore, we are supposed to deliver all our loops in CCW
+    // convention (aka right hand rule, interiour is to the left of
+    // the polyline).
+    // Since we want to want to allow for loops, whose interiour covers
```
cpjulia, line 362:
```suggestion
    // Since we want to allow for loops, whose interiour covers
```

lib/Geo/GeoJson.cpp
cpjulia: So from my understanding, the scenario in the comment made by me above wouldn't happen, because it would be classified as a multipolygon and hence parsed in the other method?

arangod/Aql/Functions.cpp
```diff
@@ -6244,6 +6244,14 @@ AqlValue Functions::GeoPolygon(ExpressionContext* expressionContext,
   builder->close();  // coordinates
   builder->close();  // object
 
+  // Now actually parse the result with S2:
+  geo::ShapeContainer container;
```
cpjulia, line 6248: Perhaps this part down until line 6254, as is repeated  in line 6330 for the multipolygon constructor, could be put in a lambda or something for being reused?

arangod/GeoIndex/Index.cpp
```diff
@@ -49,8 +50,10 @@ namespace geo_index {
 
 Index::Index(VPackSlice const& info,
              std::vector<std::vector<basics::AttributeName>> const& fields)
-    : _variant(Variant::NONE) {
+    : _variant(Variant::NONE), _legacyPolygons(false) {
```
cpjulia, line 53: does this initialization have to take place in the initializer list as we're doing something with it in the constructor block?

arangod/GeoIndex/Index.cpp
cpjulia, line 200: so the legacy boolean here would be false because there would be no such a case for arrays, but does it have anything to do with _legacyPolygons being set to true or these values are not correlated?

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -263,6 +420,205 @@ class RDBNearIterator final : public IndexIterator {
...
+                       [&](LocalDocumentId const&, VPackSlice doc) {
+                         geo::FilterType const ft = _covering.filterType();
+                         geo::ShapeContainer const& filter =
+                             _covering.filterShape();
+                         TRI_ASSERT(filter.type() !=
+                                    geo::ShapeContainer::Type::EMPTY);
+                         geo::ShapeContainer test;
+                         Result res = _index->shape(doc, test);
+                         TRI_ASSERT(res.ok());  // this should never fail here
+                         if (res.fail() ||
```
cpjulia, line 488: could it be removed if the assert should never fail?

cpjulia (approved):
LGTM

lib/Geo/GeoJson.cpp
neunhoef: The first loop in the polygon is simply taken as is. All following polygons are tested for containment in the first. If they are contained, they are properly nested, if not, they are inverted and we take the only sensible interpretation. This is done for each and any loop after the first one.
S2 tests separately that no two edges of these loops intersect each other. This is true for each loop individually (so each loop is not self-intersecting), and for edges from different loops.
As a consequence it follows, that if all loops are properly contained in the first one, and no two intersect, the set of loops forms a hierarchy of properly nested loops. Yes, the third can, for example, lie within the second and so on.
This is exactly the input which `InitNested` needs.
So as far as I see, all is well.
No, such a polygon is always parsed as a single polygon and never as a multipolygon.
However, if you look beneath the surface, it turns out that polygons and multipolygons are actually processed in exactly the same way within S2, as far as I know.

lib/Geo/GeoJson.cpp
neunhoef: Sorry, I do not understand this transformation and I believe it is wrong. It is critical here that we first check containment in line 379, and (if not legacy), then first try to invert (which changes the loop in place) and then check again for containment. Your code change would simply check the polygon twice in the non-legacy case, without first inverting it.

lib/Geo/GeoJson.cpp
neunhoef: Well spotted! Changed.

lib/Geo/GeoJson.cpp
neunhoef: Same. Done.

arangod/Aql/Functions.cpp
neunhoef: Actually, it is `parseMultiPolygon` there, so it is genuinely different. And the two are not in the same function, so a lambda is not possible. I think I prefer it the way it is.

arangod/GeoIndex/Index.cpp
neunhoef: This is an interesting one. This code is only used when we set up an IndexIterator for a geo index. It is only used for the object `obj` in a statement like

```
FILTER GEO_DISTANCE(obj, d.geo) <= 10000
```

or so. Thus, it does not influence the way geo data in the database is parsed for indexing. That is why I intentionally put `false` here.

However, it does ever so slightly influence the centroid of `obj` which is used for the actual distance calculation when the query is executed. So therefore, I think you have a point, if we want to be fully backwards compatible for our queries, we would have to use `_legacyPolygons` here, and I will change this.
It is slightly painful, since this is a static member function, so we must hand the argument in, but this is doable.

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
neunhoef: It could, if we are right and do not have a but which would trigger the assertion. This is just defensive programming: If ever we have a situation in which the `shape` call fails, then in a release build the assertion is ignored. In this case we prefer in production to simply ignore the result rather than handing out a potentially damaged result.

lib/Geo/GeoJson.cpp
```diff
@@ -452,36 +482,63 @@ Result parseMultiPolygon(velocypack::Slice const& vpack,
...
-                      "Subsequent loop not a hole in polygon");
+        bool bad = false;
+        if (legacy) {
+          bad = true;
+        } else {
+          loop->Invert();
+          if (!loops[outerLoop]->Contains(loop)) {
+            bad = true;
+          }
+        }
```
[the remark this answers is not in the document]
neunhoef: That would be correct, too. It is two lines shorter. Would be fine with me...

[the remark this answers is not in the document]
MBkkt: I just don't like a bit this:
```C++
bool bad = false;
if (legacy) {
  bad = true;
}
```
But I don't insist

arangod/GeoIndex/Index.cpp
neunhoef: Done.

PR #14729, opened 2021-09-03
Bug-fix-3.8/make iterator usage safer after intermediate commits

mpoeter:
### Scope & Purpose

Backport of #14543 and #14346 (which builds upon the former).

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/752

* Fix internal iterator states after intermediate commits in write transactions. Iterators could point to invalid data after an intermediate commit, producing undefined behavior.
  
* Fix read-own-write behavior in different scenarios:
    - in some cases writes performed by an AQL query could be observed within the same query. This was not intended and is fixed now.
    - AQL queries in streaming transactions could observe their own writes in even more cases, which could potentially result in an endless loop when the query iterates over the same collection that it is inserting documents into.
    - UPSERT did not find documents inserted by a previous iteration if the subquery relied on a non-unique secondary index.
    - disabled intermediate commits for queries with UPSERTs, because intermediate commits can invalidate the internal read-own-write iterator required by UPSERT. Previously, UPSERTs that triggered intermediate commits could have produced unexpected results (e.g., previous inserts that have been committed might not be visible) or even crashes.

To achieve the correct read-own-write behavior in streaming transactions, we sometimes have to copy the internal WriteBatch from the underlying RocksDB transaction. In particular, the copy is created whenever an AQL query with
  modification operations (INSERT/REMOVE/UPDATE/UPSERT/REPLACE) is executed in the streaming transaction. If there have not been any other modifications so far (queries/document operations), then the WriteBatch is empty and creating the copy is essentially a no-op. However, if the transaction already contains a lot of modifications, creating the WriteBatch copy might incur some overhead that can now lead to decreased performance.

### Testing & Verification

  - [x] Added new **integration tests** in shell_client


arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
```diff
@@ -82,7 +82,6 @@ EngineInfoContainerDBServerServerBased::TraverserEngineShardLists::TraverserEngi
   auto const& restrictToShards = query.queryOptions().restrictToShards;
   // Extract the local shards for edge collections.
   for (auto const& col : edges) {
-    TRI_ASSERT(col != nullptr);
```
jsteemann, line 85: I think the changes in this file should all be reverted. They have probably slipped into this PR by mistake.

arangod/RocksDBEngine/RocksDBTransactionMethods.cpp
```diff
@@ -0,0 +1,56 @@
...
+    // disable a check that we do not create read-own-write iterators with intermediate commits enabled.
+    // we can safely do this here, because our iterator has a limited lifetime and can therefore not
+    // be invalidated by intermediate commits.
+    opts.checkIntermediateCommits = false;
+  });
+  iter->Seek(bounds.start());
+  auto end = bounds.end();
+  rocksdb::Comparator const* cmp = bounds.columnFamily()->GetComparator();
+
+  // extra check to avoid extra comparisons with isElementInRange later;
```
jsteemann, line 45: the comment is misleading here, because we are not performing any _extra_ check here. I suggest to remove the comment.

jsteemann (changes requested):
All good except some changes in ExecutionBlockImpl.cpp that should all be reverted in this PR.

arangod/Aql/EngineInfoContainerDBServerServerBased.cpp
mpoeter: Good catch - done!

arangod/RocksDBEngine/RocksDBTransactionMethods.cpp
mpoeter:
```suggestion
```

jsteemann (approved):
LGTM!

KVS85:
Go tests seem to be unreliable atm.

neunhoef (commented):
LGTM

PR #14745, opened 2021-09-07
using lookahead in zkd

romanatarango:
### Scope & Purpose

This is an experimental improvement of the ZKD based index iterator. While querying a multidimensional range of values one can now choose (instead of computing the next intersection point of the Z-curve with the range) to try first several documents that are immediately after the current document. If one of them is in the range, the computation of the next intersection point is not necessary, otherwise the point is computed.

A new parameter can be passed from the query: `lookahead` (a non-negative number). It tells how many documents after the current one should be checked for being in the search range before looking for the next point where the Z-curve intersects the search range. The lookahead is passed from the query to, finally, `RocksDBZkdIndexIterator`. However, it possibly should depend on the dimension of the index and on the limits of the query (see function `numNextTries`). The main change is in the function `RocksDBZkdIndexIterator::nextImpl()`.

#### Backports:

- [ ] No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation


arangod/Aql/IndexHint.h
```diff
@@ -56,9 +56,11 @@ class IndexHint {
   std::string typeName() const;
   std::string toString() const;
 
+  size_t getLookahead() const noexcept { return _lookahead; }
  private:
   HintType _type;
   bool _forced;
+  size_t _lookahead = 1;
```
maierlars, line 63: Here we have default value 1.

arangod/Indexes/IndexIterator.h
```diff
@@ -285,6 +285,9 @@ struct IndexIteratorOptions {
   bool evaluateFCalls = true;
   /// @brief enable caching
   bool enableCache = true;
+  /// @brief number of lookahead elements considered before computing the next
+  /// intersection of the Z-curve with the search range
+  size_t lookahead = 2;
```
maierlars, line 290: And here we have a default value of 2. This can be confusing.

maierlars (changes requested):
There are zdk index js tests. we should run another set of those tests with lookahead enabled to check that it does not miss any documents. See `tests/js/server/aql/aql-optimizer-zkdindex-multi.js`

arangod/Indexes/IndexIterator.h
romanatarango: made 1 as well

arangod/Aql/IndexHint.h
romanatarango: Now the default is everywhere 1.

arangod/Indexes/IndexIterator.h
romanatarango: Now the default is everywhere 1.

romanatarango:
> There are zdk index js tests. we should run another set of those tests with lookahead enabled to check that it does not miss any documents. See `tests/js/server/aql/aql-optimizer-zkdindex-multi.js`

Added testing with loolkahead 32.

maierlars (approved):

mpoeter (approved):

PR #14747, opened 2021-09-07
disable RTTI for snappy compilation

jsteemann:
### Scope & Purpose

This restores the compilation of Snappy to how it was done before commit 573f7ebfdd0f5856970b2ea93555175d5ee9743b. That commit disabled the hack in Snappy that turns of RTTI in its CMakeLists.txt.
The reason for enabling RTTI for Snappy compilation in that previous commit was that we were deriving from two Snappy source files, which are supposed to be used as interfaces. However, g++ cannot link files which have base classes compiled without RTTI and derived classes that use RTTI. That's why RTTI was enabled for the Snappy files in that commit.
That required changes to the original Snappy CMakeLists.txt, and may have had an adverse impact on performance (still to be proven by our performance tests). Now, instead of turning RTTI on for Snappy, we turn it back off and in addition we turn off RTTI for the one file that derives from Snappy's interfaces. So we have to use some source-file specific compile hack in CMakeLists.txt to turn RTTI off for one of our own files.
This is hopefully less intrusive than turning RTTI off for Snappy and having to maintain a patch for Snappy's CMakeLists.txt long-term.

This doesn't need to be backported because the commit above was only in devel/3.9.

Testing if this works is basically done by compilers on all platforms. If this has a positive impact on performance we will see soon after merging by observing the trends in our performance tests.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/3rdParty/README_maintainers.md
+++ b/3rdParty/README_maintainers.md
@@ -104,17 +104,6 @@ http://s2geometry.io/
 Compression library
 https://github.com/google/snappy
 
-The original Snappy code disables RTTI. This prevents the code from linking with
-ArangoDB, so we need to comment the section in Snappy's CMakeLists.txt that disables
-RTTI unconditionally.
-
-     # Disable RTTI.
-    -string(REGEX REPLACE "-frtti" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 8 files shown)


mpoeter (approved):

jsteemann:
Merging despite unrelated test failure on Windows (cluster).

PR #14748, opened 2021-09-08
Feature/apm 70

cpjulia:
### Scope & Purpose

Implementation of APM-70: https://arangodb.atlassian.net/browse/APM-70
Docs PR: https://github.com/arangodb/docs/pull/760

Allow the PRUNE statement condition to be stored in a variable, in such a way that it can be used later as a condition for another statement, as in the example below:
```
FOR doc INBOUND a GRAPH graph
  PRUNE result = doc.age > 10 and doc.state == 'active'
  FILTER result and doc.type == 'customer'
  ...
  ```
  the PRUNE condition `doc.age > 10 and doc.state == 'active'` is stored in variable `result`, which is used as part of the condition for the FILTER statement.

- [X] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)

#### Backports:

- [X] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/760
- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/APM-70

### Testing & Verification

  - [x] Added new **integration tests** (in shell_server_aql)

### CODE REVIEW

File: tests/js/server/aql/aql-graph-traverser.js [hunk 1]
```diff
@@ -2412,6 +2412,320 @@ function complexFilteringSuite() {
...
+    },
+
+    testReassignVariableInPruneCondition: function () {
+      let query = `FOR v,e,p IN 1..10 OUTBOUND @start @@eCol  PRUNE p = v.value == 75 RETURN 1`;
+      try {
+        let bindVars = {
+          '@eCol': en,
+          'start': vertex.A
+        };
+       db._query(query, bindVars);
```
jsteemann, line 2526:
```suggestion
        db._query(query, bindVars);
```

tests/js/server/aql/aql-graph-traverser.js
```diff
@@ -2412,6 +2412,320 @@ function complexFilteringSuite() {
...
+        if(node.type === "FilterNode") {
+          assertEqual(node["inVariable"]["name"], "pruneCondition");
+        }
+        if(node.type === "TraversalNode") {
+          assertEqual(node["options"]["order"], "bfs");
+        }
+      });
+      assertTrue(foundExtraLet);
+    },
+
```
mchacki, line 2728: I like the above tests, very good, they over a lot.

However I would like to ask for some more tests, just to increase the confidence here.

1) Reassigning of variables:
  a) Add a variable, then try to reassign it as the "prune" variabel (you did this on `p` however this `p` is part of the node and somewhat special. I am looking more for `FOR other IN 1..2 FOR v IN 1 OUTBOUND ... PRUNE other = v.name == "test"`
    b) Try to resign the prune condition later `PRUNE pruneCondition = v.name == "test" LET pruneCondition = e.weight`
    c) Use a second traversal statement, and try to reassign the prune condition there:
       `FOR v1 IN <...> PRUNE pruneCondition = v1.name == "test"  <...> FOR v2 IN 1 OUTBOUND v1 PRUNE pruneCondition = v2.name`

All of the above tests should fail due to invalid reassignment.
2) Subquery
  Try to add PRUNE pruneConditon = (FOR other IN 1..2 RETURN other)
  This should be disallowed as well. (subquery is disallowed)
3) Use the pruneCondition in more complex places (I expect all of those to work)
  a) Try to use it within a subquery `PRUNE condition = .... LET subQuery = (FOR q IN p.edges FILTER condition RETURN q)`
   b) Try to use it in a calculation LET array = [] FOR v IN 1 OUTBOUND PRUNE condition = <....> RETURN PUSH(array, condition)
  c) Try to use a second Traversal and use the Prune Expression within a path filter:
  ```
   FOR v1 IN 1 OUTBOUND <...> PRUNE pruneCondition = v1.name == "test"
    FOR v2, e2, p2 IN 2 ANY v1 FILTER  p2.edges[*].value ALL== pruneCondition
```

All of the above tests basically target the fact that the calculation is handed around, and we need to make sure the references stay intact.
If I understood the implementation correctly you are not storing the result (Which you cannot get to) but your are storing the way it should be calculated.

File: tests/js/server/aql/aql-graph-traverser.js [hunk 3]
```diff
@@ -2412,6 +2412,320 @@ function complexFilteringSuite() {
...
+        }
+        if(node.type === "FilterNode") {
+          assertEqual(node["inVariable"]["name"], "pruneCondition");
+        }
+        if(node.type === "TraversalNode") {
+          assertEqual(node["options"]["order"], "bfs");
+        }
+      });
+      assertTrue(foundExtraLet);
+    },
```
mchacki, line 2727: This Test I think is very important, so even if you would not do any of the user suggested test I would like to see at least this one here.

Test the Scope of the PRUNE condition.

Try to use the following Query:
```
FOR v IN 1 OUTBOUND @start PRUNE pruneCondition = v.right == true
/* Collect will break the scope and needs to invalidate all variables, this collect however is a "pass through" as _key is always unique */
COLLECT tmp = v._key
/* Here v is not defined anymore*/
/* if the PRUNE condition is simply written down, this can fail because v is undefined */
RETURN pruneCondition
```

And also this query (change only after COLLECT):

```
FOR v IN 1 OUTBOUND @start PRUNE pruneCondition = v.right == true
/* Collect will break the scope and needs to invalidate all variables, this collect however is a "pass through" as _key is always unique */
COLLECT tmp = v._key
/* Here v is not defined anymore*/
/* We redefine v to something unrelated */
LET v = {right: true}
/* Now the pruneCondition has a v to read, however it is totally unrelated to the original v as used by the condition.*/
RETURN pruneCondition
```

Both of the above queries should fail in my opinion as the pruneCondition needs to be invalidated by COLLECT.

mchacki (changes requested):
Technical Change very good, small change, and I really like it :+1:
Also the test coverage is good.

However there are some more strange situations in AQL and I tried to add some more test suggestions to it.

The requested change is the two Tests around COLLECT.
Every other comment can be treated as nice-to-have.

tests/js/server/aql/aql-graph-traverser.js
cpjulia: Added these suggestions in the last commit

tests/js/server/aql/aql-graph-traverser.js
cpjulia: Added these tests in the last commit

tests/js/server/aql/aql-graph-traverser.js
```diff
@@ -2412,6 +2412,320 @@ function complexFilteringSuite() {
...
+
+    testStorePruneConditionPathInVariableExplain: function () {
+      // note: using NOOPT(...) here to disable the optimization that pulls
+      // certain FILTER conditions into the traversal
+      const condition = "NOOPT(p.edges[0].right) == true";
```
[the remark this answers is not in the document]
cpjulia: Added a test that is optimized and checks that the PRUNE variable is non existant in the AQL_EXPLAIN

mchacki (approved):
Perfect nice tests added :+1:

LGTM!
Thank you very much

PR #14754, opened 2021-09-09
Bug fix/fix active failover foxx failover

dothebart:
### Scope & Purpose

properly invoke self heal when active failover happens to actually have a failover. 

#### Backports:

- [x] Backports required for: 3.7, 3.7

#### Related Information
- [x] GitHub issue / Jira ticket number: BTS-558

### Testing & Verification

Run https://github.com/arangodb/release-test-automation/blob/main/release_tester/test.py with `--starter-mode AFO`

- [x] This change is already covered by existing tests, such as RTA triggering failover, and revalidating services.
- [ ] There are tests in an external testing repository: https://github.com/arangodb/release-test-automation


pluma (approved):
LGTM

js/server/leader.js
```diff
@@ -0,0 +1,34 @@
...
+// /
+// / Unless required by applicable law or agreed to in writing, software
+// / distributed under the License is distributed on an "AS IS" BASIS,
+// / WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// / See the License for the specific language governing permissions and
+// / limitations under the License.
+// /
+// / Copyright holder is ArangoDB GmbH, Cologne, Germany
+// /
+// / @author Allen Plum
```
jsteemann, line 27:
```suggestion
// / @author Alan Plum
```

js/server/modules/@arangodb/foxx/queues/manager.js
```diff
@@ -226,11 +226,6 @@ exports.manage = function () {
   }
 
   if (global.ArangoServerState.getFoxxmasterQueueupdate()) {
-    if (!isCluster) {
-      // On a Foxxmaster change FoxxmasterQueueupdate is set to true
-      // we use this to signify a Leader change to this server
-      foxxManager.healAll(true);
-    }
```
jsteemann, line 233: What is the rationale for removing this code here?
If that code is going to be removed, where is the healAll being executed for single servers instead?

dothebart: this route is invoked through the foxx queues, which don't have to be enabled. 
Otoh, the `true` ends up to only unpacking foxx services on the failover host, and not registering their routes.

js/server/leader.js
```diff
@@ -0,0 +1,34 @@
...
+// / @author Copyright 2021, ArangoDB GmbH, Cologne, Germany
+// //////////////////////////////////////////////////////////////////////////////
+
+(function(){
+  "use strict";
+require("@arangodb/foxx/manager").healAll();
```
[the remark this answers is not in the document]
dothebart: it should be run in all, so that the routes become installed in all v8-contexts; else they will 404 and not list in aardvark on first try, which was another facette of the original bug.

js/server/modules/@arangodb/foxx/queues/manager.js
jsteemann: What about single server here? Is there any `healAll(true)` being executed for single servers now?

js/server/leader.js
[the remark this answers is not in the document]
jsteemann: Got it. IMO it would be good to add locks around the self-heal procedure because when executed concurrently, it can lead to undefined behavior. But this has been the case even before this PR so it should be attacked separately.

js/server/modules/@arangodb/foxx/queues/manager.js
dothebart: this was added by Simon especially for active failover according to git blame.

jsteemann: That's fine. But apparently the `if` block was executed in single server as well.
So just wondering if it was needed on single server or not.

pluma: The `true` does less than the default, it opts out of reloading the routing, meaning it doesn't directly take effect for the installed services (this is intended to be used in situations where doing so would result in unnecessary churn and the reloading is triggered explicitly later).

This is useless in a single server as the self-heal is only relevant if the server needs to restore services from the database, which wouldn't happen outside failover for single server deployments. One could argue whether it should happen in the startup though.

js/server/leader.js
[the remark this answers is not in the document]
pluma: No, this only needs to run in one. The reload routing is going to tell all other contexts to purge their routing maps and reload the Foxx services. @jsteemann is right.

pluma (changes requested):
See comments.

js/server/modules/@arangodb/foxx/queues/manager.js
pluma: If we wipe out a single server failover network and then reboot it, we probably want an initial selfHeal too? So we might have to add the new `healAll(false)` (or `healAll()`, which is identical) in `server/server.js` too.

pluma:
https://github.com/arangodb/arangodb/pull/14754#discussion_r705659069

https://github.com/arangodb/arangodb/pull/14754#discussion_r705657761

js/server/leader.js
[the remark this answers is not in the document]
dothebart: @jsteemann in the case of `threadNumber === 0` probably the c++ wraper needs to fetch the special context from the dealer? Is this available with `--console` ?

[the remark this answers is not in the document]
jsteemann: The V8 context used for `--console` will be #1, not #0. So I guess it is safe to restrict the healAll call to `threadNumber === 0`:
```
if (require("internal").threadNumber === 0) { 
  require("@arangodb/foxx/manager").healAll(); 
}
```

tests/js/client/active-failover/basic.js
```diff
@@ -293,10 +296,136 @@ function waitUntilHealthStatusIs(isHealthy, isFailed) {
...
+  assertEqual(reply.code, "200", JSON.stringify(reply));
+
+  print("Foxx: Itzpapalotl summoning Chalchihuitlicue");
+  reply = arango.GET_RAW('/_db/_system/itz/Chalchihuitlicue/summon', onlyJson);
+  assertEqual(reply.code, "200", JSON.stringify(reply));
+  let parsedBody = JSON.parse(reply.body);
+  assertEqual(parsedBody.name, "Chalchihuitlicue");
+  assertTrue(parsedBody.summoned);
+
+ //print("Foxx: crud testing get xxx");
```
jsteemann, line 356: This is a new test, but the code is commented out.
Can the comments be removed or does the test need to be extended?

js/server/modules/@arangodb/foxx/queues/manager.js
dothebart: @pluma do you mean to cope with manual replication? can you push that?

tests/js/client/active-failover/basic.js
dothebart: I copied the complete test from checkdata, which also installs a foxx service that has crud capabilities.
Add service too? Remove? pls decide.

jsteemann: If the decision is about whether to remove the commented out or enable it, I'd say let's enable it. Provided it is no major effort to get the extra test code to work.
If it turns out to be a major issue, I suggest removing the commented out test code.

pluma:
Lgtm :shipit:

js/server/modules/@arangodb/foxx/queues/manager.js
dothebart: tested whether a cold restart with a new leader would also properly restore, as discussed.

tests/js/client/active-failover/basic.js
dothebart: it seems more cleanup was needed, so it took a little longer.

jsteemann (approved):
LGTM

PR #14757, opened 2021-09-10
Fix windows unicode issues

mpoeter:
### Scope & Purpose

This PR adds a manifest that sets the process's codepage to UTF-8. This way all calls to ANSI versions of the Windows API will treat the input data as UTF-8 instead of current system codepage. This feature is available since Windows 10 1903 (see https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page).

In addition, this PR sets the console output codepage to UTF-8 as well so that unicode characters are correctly displayed in the console.

Fixes BTS-531

#### Backports:

- [x] No backports required

#### Related Information

- [x] Jira ticket number: BTS-531

### Testing & Verification


Dronplane:
I suggest to add the same manifest to the tests executables to have unit test environment closer to the "production"

arangod/RestServer/arangod.cpp
```diff
@@ -374,6 +374,10 @@ int main(int argc, char* argv[]) {
 
   TRI_GET_ARGV(argc, argv);
 #if _WIN32
+  // Set the codepage for the console output to UTF-8 so that unicode characters
+  // are displayed correctly.
+  SetConsoleOutputCP(CP_UTF8);
```
Dronplane, line 379: We have some console-related tasks in the ShellColorsFeature::prepareConsole  - maybe move this call to that function?
As a bonus we will enable UTF-8 encodings in all other executables (e.g. arangodump/arangorestore....)

dothebart (approved):

Dronplane (approved):
LGTM

PR #14763, opened 2021-09-13
Adapt scheduler to support bounded and unbounded queue operations.

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/754

The unbounded operation is expected to always succeed and therefore is noexcept and returns void. However, the queue operation must always allocate memory, and a bad_alloc would therefore terminate the process (similar to the case when allocation of the continuation block in the futures library fails). We decided that this is acceptable since this probably a very rare scenario, and a out-of-memory situation where we cannot allocate such a small amount of memory for a single task is most likely unrecoverable anyway.

All internal tasks should be scheduled with the unbounded queue operation; only user requests should be scheduled with the bounded version. That way internal tasks can not be blocked by a large number of user requests. Since we only want to count bounded tasks, we can no longer rely on the lock-free queue's bounded functions, but instead have to maintain our own counter for the number of bounded tasks currently queued.

#### Backports:

- [x] No backports required

### Testing & Verification

- [x] This change is already covered by existing tests


arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -249,20 +251,43 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+  auto& queue = _queues[queueNo];
+  if (bounded) {
+    auto maxSize = _maxFifoSizes[queueNo];
+    auto queueFull = [&](){
+      LOG_TOPIC("98d94", DEBUG, Logger::THREADS)
+        << "unable to push job to scheduler queue: queue is full";
+      logQueueFullEveryNowAndThen(queueNo, maxSize);
+      ++_metricsQueueFull;
+    };
+    if (queue.numCountedItems.load(std::memory_order_relaxed) > maxSize) {
```
neunhoef, line 280: What do we achieve with this load before the `fetch_add`? True, if the queue is full, we save two stores. But we pay in the non-full case with an additional load. I would probably have optimized for the queue not full case and paid with two stores if the queue is full...

neunhoef (approved):
LGTM. See one question/comment about some optimization decision.

arangod/Pregel/Conductor.cpp
```diff
@@ -953,16 +938,12 @@ ErrorCode Conductor::_sendToAllDBServers(std::string const& path, VPackBuilder c
...
-        return TRI_ERROR_QUEUE_FULL;
-      }
+      scheduler->queue(RequestLane::INTERNAL_LOW, [this, path, message,
+                                                   self = shared_from_this()] {
+        TRI_vocbase_t& vocbase = _vocbaseGuard.database();
+        VPackBuilder response;
+        _feature.handleWorkerRequest(vocbase, path, message.slice(), response);
+      });
     }
     return TRI_ERROR_NO_ERROR;
```
jsteemann, line 948: It looks like this method now always returns `TRI_ERROR_NO_ERROR`. Shall we declare it void?

arangod/Scheduler/SupervisedScheduler.cpp
jsteemann: I agree. We should unconditionally execute `fetch_add(1, ...)` without doing an extra check before. That way we may over-count the number of queued items and may even drop a few requests. But this will only happen when we are close to the bound anyway, so I don't think this is a problem if we treat the bounds as "best effort" exact only.

jsteemann (approved):
LGTM!

arangod/Scheduler/SupervisedScheduler.cpp
mpoeter: Fair enough, will change.

arangod/Pregel/Conductor.cpp
mpoeter: Only in the cluster case, in the single server case we also have code paths that can return error codes.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -249,20 +251,36 @@ SupervisedScheduler::SupervisedScheduler(application_features::ApplicationServer
...
+  _queues[1].queue.reserve(fifo1Size);
+  _queues[2].queue.reserve(fifo2Size);
+  _queues[3].queue.reserve(fifo3Size);
   TRI_ASSERT(fifo3Size > 0);
 }
 
 SupervisedScheduler::~SupervisedScheduler() = default;
 
-bool SupervisedScheduler::queueItem(RequestLane lane, std::unique_ptr<WorkItemBase> work) {
+bool SupervisedScheduler::queueItem(RequestLane lane, std::unique_ptr<WorkItemBase> work, bool bounded) {
```
cpjulia, line 263: can this method be a noexcept or still there might be throwable memory exceptions?

cpjulia (approved):
LGTM

arangod/Scheduler/SupervisedScheduler.cpp
jsteemann: Actually that is a good point. I don't vote for making this method noexcept, but what will happen if we increase the queue's counter and then the pushing to the queue throws an exception?
In that case we will have increased the counter but will not have pushed an item to the queue, and the counter will not fix itself anymore.
I think we need to wrap the `queue.push(...)` into a try...catch and decrease the queue's counter again if we catch an exception and the queue was bounded.
@mpoeter: what do you think?

cpjulia: Good point, I agree with this

jsteemann:
Thanks, @mpoeter !

PR #14767, opened 2021-09-13
Remove `readCompleteInput` flag

jsteemann:
### Scope & Purpose

* Remove internal AQL query option `readComplete` that controled if all inputs for a modification operation (UPDATE / REPLACE / REMOVE) were read into memory first. This was a necessity with the MMFiles storage engine in cases when a query read from a collection and wrote into it in the same query. With the RocksDB engine and its snapshotting, we never need to read the entire inputs into memory first.

#### Backports:

- [x] No backports required

### Testing & Verification


CHANGELOG
```diff
@@ -1,6 +1,13 @@
 devel
 -----
 
+* Remove internal AQL query option `readComplete` that controled if all
+  inputs for a modification operation (UPDATE / REPLACE / REMOVE) were
+  read into memory first. This was a necessity with the MMFiles storage
+  engine in cases when a query read from a collection and wrote into it
+  in the same query. With the RocksDB engine and its snapshotting, we
+  never need to read the entire inputs into memory first.
```
jsteemann, line 9:
```suggestion
* Remove internal AQL query option `readCompleteInput` that controled if
  all input for a modification operation (UPDATE / REPLACE / REMOVE) are
  read into memory first. This was a necessity with the MMFiles storage
  engine in cases when a query read from a collection and wrote into it
  in the same query afterwards. With the RocksDB engine and its snapshots, 
  we never need to read the entire input into memory first.
```

PR #14782, opened 2021-09-15
Fix /_admin/cluster/health if one agent gone

neunhoef:
This is a rather trivial bug fix which addreses BTS-584.
It fixes a regression on /_admin/cluster/health if an agent is off line,
since the health endpoint retried for too long to reach the agent before
giving up.

This is a trivial fix which is covered by QA tests.


joerg84:
@KVS85 FYI

dothebart (approved):
revalidated, fixes the incident.

PR #14788, opened 2021-09-16
[APM-73] Change result code for not-fulfilled write-concern

neunhoef:
This implements a PM decision from APM-73:

If there are not enough replicas of a shard on-line for the configured
write concern, then we now return a special error with error code 503,
rather than "collection is read-only" with error code 403. This has the
important consequence that the coordinator can retry for up to 15
minutes.

There are several things missing for this PR. One is documentation and making sure that all drivers and driver tests are fine.
Some additional tests are still needed.
Second is that in all my tries Jenkins had something to complain about. Things seemed to be unrelated, but who knows?
And finally, there is a decision necessary as to if this hits 3.9 or will only appear with 3.10.

### Scope & Purpose

#### Backports:

- [*] No backports required

      This is a breaking API change and should not be backported.

#### Related Information

- [*] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/APM-73

### Testing & Verification

### Documentation


neunhoef:
This has bit rot and is not complete. There is not a lot in it to merit keeping it. Closing.

PR #14803, opened 2021-09-20
dead code removal and cleanup

jsteemann:
### Scope & Purpose

Remove dead code and simplify some other places.

#### Backports:

- [x] No backports required

### Testing & Verification


lib/Containers/Helpers.h
```diff
@@ -0,0 +1,72 @@
...
+static std::size_t nextCapacity(T const& container, std::size_t initialCapacity) {
+  std::size_t capacity;
+  if (container.empty()) {
+    // reserve some initial space
+    capacity = std::max(std::size_t(1), initialCapacity);
+  } else {
+    capacity = container.size() + 1;
+    // allocate with power of 2 growth
+    if (capacity > container.capacity()) {
+      capacity *= 2;
```
goedderz, line 48: Unrelated, but afaik choosing a somewhat lower growth rate than 2 (like 1.5) should improve performance, because memory might be reused, making it more cache-friendly.

goedderz (approved):
LGTM

lib/Containers/Helpers.h
jsteemann: @goedderz: I suggest to merge this PR as is, because it doesn't change existing behavior.
After that, we can create a separate PR for the 1.5x growth factor. We can then check in our daily performance tests if that change has any measurable impact. I can't predict it right now :sweat_smile:

PR #14809, opened 2021-09-21
Stabilize Hybrid Smart Graph Test [EE only]

mchacki:
### Scope & Purpose
 This is just a stabilization of an enterprise test.
 In this part of the PR there is a generic helper method that is used in the test, and could be used by others potentially.
 Relevant change is here:
 https://github.com/arangodb/enterprise/pull/762
 
 As this is only test code, this PR has on purpose no Changelog entry, User-space remains unmodified.


kvahed (approved):
LGTM

PR #14811, opened 2021-09-22
experimental: try a container growth factor of 1.5x

jsteemann:
### Scope & Purpose

As suggest by @goedderz in https://github.com/arangodb/arangodb/pull/14803#discussion_r713709584:

Experimental: try a container growth factor of 1.5x instead of 2x. We will have to see if this has an impact in our regular performance tests.
If this does not have any measurable or a negative effect on performance, we can simply revert the changes in this PR.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/lib/Containers/Helpers.h
+++ b/lib/Containers/Helpers.h
@@ -33,8 +33,8 @@ namespace arangodb::containers {
 struct Helpers {
 /// @brief calculate capacity for the container for at least one more 
 /// element.
-/// if this would exceed the container's capacity, use a factor-of-2
-/// growth strategy to calculate the capacity.
+/// if this would exceed the container's capacity, use a growth factor of
+/// 1.5 to calculate the new capacity.
 template<typename T>
 static std::size_t nextCapacity(T const& container, std::size_t initialCapacity) {
   std::size_t capacity;
@@ -42,10 +42,11 @@ static std::size_t nextCapacity(T const& container, std::size_t initialCapacity)
     // reserve some initial space
     capacity = std::max(std::size_t(1), initialCapacity);
   } else {
+    // minimum is that we have room for at least one more element.
     capacity = container.size() + 1;
-    // allocate with power of 2 growth
     if (capacity > container.capacity()) {
-      capacity *= 2;
+      // grow with a growth factor of 1.5
+      capacity = (container.size() * 3) / 2;
     }
   }
   TRI_ASSERT(capacity > container.size());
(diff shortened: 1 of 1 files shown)


lib/Containers/Helpers.h
```diff
@@ -33,28 +33,29 @@ namespace arangodb::containers {
...
     // reserve some initial space
     capacity = std::max(std::size_t(1), initialCapacity);
   } else {
+    // minimum is that we have room for at least one more element.
     capacity = container.size() + 1;
-    // allocate with power of 2 growth
     if (capacity > container.capacity()) {
-      capacity *= 2;
+      // grow with a growth factor of 1.5
+      capacity = (container.size() * 3) / 2;
```
goedderz, line 49: I think there's a bug here if `container.size() == 1`, as it would set capacity to one.
```suggestion
      capacity = ((container.size() - 1) * 3) / 2 + 1;
```
As `!container().empty()`, this won't underflow. And we get capacity rounded up rather than down. Or we just use
```suggestion
      capacity = (container.size() * 3 + 1) / 2;
```
.

goedderz (commented):
I took the time to quickly search for this. folly does this for its fbvector implementation with 1.5 as well, except for smaller capacities, where 2 is used. And what "smaller capacity" means may depend on jemalloc.

This is to exactly match the block sizes of the allocator if possible for small sizes, and start to try and re-use memory for larger sizes.

https://github.com/facebook/folly/blob/master/folly/FBVector.h#L1151
https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md

lib/Containers/Helpers.h
jsteemann: Ok, I have applied the suggestion.

goedderz (approved):

lib/Containers/Helpers.h
```diff
@@ -33,28 +33,37 @@ namespace arangodb::containers {
...
+      constexpr size_t jemallocMinInPlaceExpandable = 4096;
+      if (container.capacity() < jemallocMinInPlaceExpandable / sizeof(typename T::value_type)) {
+        capacity = container.capacity() * 2;
+      } else {
+        // grow with a growth factor of 1.5
+        capacity = (container.size() * 3 + 1) / 2;
+      }
     }
   }
+  TRI_ASSERT(capacity >= container.capacity());
```
jsteemann, line 59:
```suggestion
```

lib/Containers/Helpers.h
```diff
@@ -33,28 +33,37 @@ namespace arangodb::containers {
...
   } else {
+    TRI_ASSERT(container.capacity() > 0);
+    // minimum requirement is that we have room for at least one more element.
     capacity = container.size() + 1;
-    // allocate with power of 2 growth
     if (capacity > container.capacity()) {
-      capacity *= 2;
+      // inspired by facebook/folly (https://github.com/facebook/folly/blob/master/folly/memory/Malloc.h):
+      constexpr size_t jemallocMinInPlaceExpandable = 4096;
+      if (container.capacity() < jemallocMinInPlaceExpandable / sizeof(typename T::value_type)) {
```
goedderz, line 51: Would this be more accurate?
```suggestion
      if (container.capacity() < jemallocMinInPlaceExpandable / std::max(sizeof(typename T::value_type), alignof(typename T::value_type))) {
```

goedderz (approved):

PR #14824, opened 2021-09-27
Fix crash in optimizer rule remove-collect-variables

goedderz:
### Scope & Purpose

As reported in https://github.com/arangodb/arangodb/issues/14807, certain AQL queries provoke a crash during the optimization phase in the optimizer rule `remove-collect-variables`. 

- [X] :hankey: Bugfix (requires CHANGELOG entry)
- [X] :book: CHANGELOG entry made

#### Backports:

- [X] Backport required for 3.8: https://github.com/arangodb/arangodb/pull/14825
- [X] Backport required for 3.7: https://github.com/arangodb/arangodb/pull/14826

#### Related Information

- [X] GitHub issue / Jira ticket number: https://github.com/arangodb/arangodb/issues/14807

### Testing & Verification

- [X] The behavior in this PR was *manually tested*
- [X] This PR adds tests that were used to verify all changes:
  - [X] Added new **integration tests** (added regression tests)
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

PR #14826, opened 2021-09-27
[3.7] Fix crash in optimizer rule remove-collect-variables

goedderz:
Backport of https://github.com/arangodb/arangodb/pull/14824


jsteemann (approved):
LGTM!

PR #14830, opened 2021-09-27
Feature/apm 173

cpjulia:
### Scope & Purpose

  Now, arangobench, arangorestore and arangodump support multiple coordinators, so the option 
  `--server.endpoint` can be used multiple times, as in the example below:

   ```
  arangobench 
    --server.endpoint tcp://[::1]::8529 
    --server.endpoint tcp://[::1]::8530 
    --server.endpoint tcp://[::1]::8531 
  ``` 

  This does not compromise the use of the other client tools, which preserve
  the behavior of having one coordinator.

#### Backports:

- [x] No backports required

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/777
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/APM-173
- [ ] Design document: 

### Testing & Verification

STILL NOT TESTED AUTOMATICALLY


arangosh/Shell/ClientFeature.h
```diff
@@ -112,7 +112,9 @@ class ClientFeature final : public HttpEndpointProvider {
   void loadJwtSecretFile();
 
   std::string _databaseName;
-  std::string _endpoint;
+  // std::string _endpoint;
```
jsteemann, line 115:
```suggestion
```

arangosh/Shell/ClientFeature.h
```diff
@@ -73,10 +77,9 @@ class ClientFeature final : public HttpEndpointProvider {
   uint64_t sslProtocol() const { return _sslProtocol; }
   bool forceJson() const { return _forceJson; }
   void setForceJson(bool value) { _forceJson = value; }
-  
-  std::unique_ptr<httpclient::GeneralClientConnection> createConnection();
+
   std::unique_ptr<httpclient::GeneralClientConnection> createConnection(std::string const& definition);
-  std::unique_ptr<httpclient::SimpleHttpClient> createHttpClient() const;
+  std::unique_ptr<httpclient::SimpleHttpClient> createHttpClient(int threadNumber = 0) const;
```
jsteemann, line 82: In all other places the thread numbers are unsigned. Here they are signed. For consistency it would be better to use an unsigned type throughout.
```suggestion
  std::unique_ptr<httpclient::SimpleHttpClient> createHttpClient(int threadNumber = 0) const;
```

arangosh/Shell/ClientFeature.h
```diff
@@ -59,8 +59,12 @@ class ClientFeature final : public HttpEndpointProvider {
 
   std::string const& databaseName() const { return _databaseName; }
   bool authentication() const { return _authentication; }
+  /*
   std::string const& endpoint() const { return _endpoint; }
   void setEndpoint(std::string const& value) { _endpoint = value; }
+  */
+  std::string const& endpoint() const { return _endpoints[0]; }
+  void setEndpoint(std::string const& value) { _endpoints[0] = value; }
```
jsteemann, line 67:
```suggestion
  // get single endpoint. used by client tools that can handle only one endpoint
  std::string const& endpoint() const { return _endpoints[0]; }
  // set single endpoint
  void setEndpoint(std::string const& value) { _endpoints[0] = value; }
```

arangosh/Shell/ClientFeature.cpp
```diff
@@ -44,17 +45,21 @@
 #include <iostream>
 #include <thread>
 
+#include <Logger/LogMacros.h>
+
```
jsteemann, line 49: Is this include still needed, or was it just needed for some previous debugging?

arangosh/Shell/ClientFeature.cpp
```diff
@@ -388,7 +389,7 @@ std::unique_ptr<httpclient::SimpleHttpClient> ClientFeature::createHttpClient(
 }
 
 std::vector<std::string> ClientFeature::httpEndpoints() {
```
jsteemann, line 391: It seems that this method only cares about endpoint 0, which may or may not be right.
However, I can't find any caller of this method, so what about circumventing this problem and simply removing the method completely?

CHANGELOG
```diff
@@ -1,6 +1,17 @@
...
+* APM-173: Now, arangobench supports multiple coordinators, so the flag 
+  `--server.endpoint` can be used multiple times, as in the example below:
+
+  arangobench \
+    --server.endpoint tcp://[::1]::8529 \
+    --server.endpoint tcp://[::1]::8530 \
+    --server.endpoint tcp://[::1]::8531 
+
+  This does not compromise the use of the other client tools, which preserve
+  the behavior of having one coordinator.
```
jsteemann, line 13: What about arangodump and arangorestore? They are going to be changed by this PR, too...

arangosh/Shell/ClientFeature.cpp
cpjulia: there's LOG_TOPIC here

arangosh/Shell/ClientFeature.cpp
cpjulia: Addressed in commit

mchacki (approved):
LGTM

PR #14726, opened 2021-09-02
Capture immutable fields during view creation through web UI.

adityamukho:
### Scope & Purpose

This PR extends the Views web UI, letting it capture View properties that are immutable once created. These include:
1. `primarySort`
2. `storedValues`
3. `primarySortCompression`
4. `writebufferIdle`
5. `writebufferActive`
6. `writebufferSizeMax`

Prior to this fix, any view created through the web UI will necessarily use **default** values for the above properties, and there is no way to change them later.

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [x] No backports required
- [ ] Backports required for: *(Please specify versions and link PRs)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket number:
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run: https://jenkins.arangodb.biz:3456/view/PR/job/arangodb-matrix-pr/16861/

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
> support, end users, and other developers have a concise overview. 

- [ ] Added entry to *Release Notes* 
- [ ] Added a new section in the *Manual* 
- [ ] Added a new section in the *HTTP API* 
- [ ] Added *Swagger examples* for the HTTP API  
- [ ] Updated license information in *LICENSES-OTHER-COMPONENTS.md* for 3rd party libraries

### External contributors / CLA Note 

[...]

PR #14848, opened 2021-10-01
Bug fix/views web UI part 2

hkernbach:
### Scope & Purpose

Bugfixes/additions for already opened PR: 
https://github.com/arangodb/arangodb/pull/14726


js/apps/system/_admin/aardvark/APP/frontend/js/templates/modalTable.ejs
```diff
@@ -96,12 +102,13 @@ function createInput(row) {
...
                 <td>
                             <span style="display: inline-flex">
-                                <button class="graphViewer-icon-button gv-icon-small add addAfter"></button>
                                 <% if (idx) {%>
                                 <button style="margin-left: 5px;"
                                         class="graphViewer-icon-button gv_internal_remove_line gv-icon-small delete addDelete">
                                 </button>
                                 <% } %>
+                                <% //<i class="moveRowButton moveRowUp disabled fa fa-arrow-circle-up"></i>
+                                // <i class="moveRowButton moveRowDown disabled fa fa-arrow-circle-down"></i> %>
```
hkernbach, line 111:
```suggestion
```

js/apps/system/_admin/aardvark/APP/frontend/scss/_modals.scss
```diff
@@ -519,6 +519,24 @@
...
+    }
+
+    &.disabled {
+      color: rgba(0,0,0,.3);
+
+      &:hover {
+        cursor: default;
+      }
+    }
+  }
```
hkernbach, line 539:
```suggestion
```

js/apps/system/_admin/aardvark/APP/frontend/js/templates/modalTable.ejs
adityamukho: Is there an alternative way to move rows up/down? Or was this requirement removed?

js/apps/system/_admin/aardvark/APP/frontend/js/views/viewsView.js
```diff
@@ -312,21 +316,24 @@
...
           'newStoredValues',
-          'Stored Values',
-          ['Fields (one per line)', 'Compression'],
+          {
+            noLabel: true // Dummy object. Entry will only be created if this is set as a String.
+          },
+          ['Fields', 'Compression'],
           [[
-            window.modalView.createBlobEntry(
+            window.modalView.createSelect2Entry(
```
adityamukho, line 327: No option to reorder contents except delete/re-insert?

js/apps/system/_admin/aardvark/APP/frontend/scss/_modals.scss
adityamukho: Same as above - we don't need move functionality anymore?

js/apps/system/_admin/aardvark/APP/frontend/js/templates/modalTable.ejs
hkernbach: Due to technical issues with the old framework, this feature should be added in our upcoming refactoring using the react js framework. This feature is planned for 3.10. The required time for implementing this feature in 3.9 (using backbone) only, and then drop it in 3.10 anyway, is not worth it. Additionally, a 3.9 backbone based implementation would have been super ugly, from a technical point of view. Also, the required time is out of proportion compared to the benefits it would deliver here.

js/apps/system/_admin/aardvark/APP/frontend/js/views/viewsView.js
hkernbach: Due to technical issues with the old framework, this feature should be added in our upcoming refactoring using the react js framework. This feature is planned for 3.10. The required time for implementing this feature in 3.9 (using backbone) only, and then drop it in 3.10 anyway, is not worth it. Additionally, a 3.9 backbone based implementation would have been super ugly, from a technical point of view. Also, the required time is out of proportion compared to the benefits it would deliver here.

js/apps/system/_admin/aardvark/APP/frontend/scss/_modals.scss
hkernbach: Due to technical issues with the old framework, this feature should be added in our upcoming refactoring using the react js framework. This feature is planned for 3.10. The required time for implementing this feature in 3.9 (using backbone) only, and then drop it in 3.10 anyway, is not worth it. Additionally, a 3.9 backbone based implementation would have been super ugly, from a technical point of view. Also, the required time is out of proportion compared to the benefits it would deliver here.

adityamukho (approved):

PR #14859, opened 2021-10-04
fix race in RocksDB throttle startup

jsteemann:
### Scope & Purpose

* Fix race in RocksDB throttle listener, when it was getting started lazily during server shutdown.

#### Backports:

- [x] No backports required

### Testing & Verification

### DIFF
--- a/arangod/RocksDBEngine/RocksDBThrottle.cpp
+++ b/arangod/RocksDBEngine/RocksDBThrottle.cpp
@@ -150,7 +150,6 @@ void RocksDBThrottle::stopThread() {
     }
 
     // wait until startup has finished
-    TRI_ASSERT(state == ThrottleState::Starting);
     std::this_thread::sleep_for(std::chrono::milliseconds(5));
   }
 }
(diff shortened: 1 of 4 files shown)


arangod/RocksDBEngine/RocksDBThrottle.cpp
```diff
@@ -99,89 +101,94 @@ thread_local sPriorityInfo gThreadPriority = {false, 0, 0};
...
+
+          _internalRocksDB = nullptr;
+          _delayToken.reset();
+        }
+        break;
+      }
+    }
+
+    // wait until startup has finished
+    TRI_ASSERT(state == ThrottleState::Starting);
```
mpoeter, line 153: I think this assert is not correct. Suppose we reach the `_throttleState.compare_exchange_strong` in line 127, but in the meantime `_throttleState` has been updated to Running. I think this could be fixed by removing the else in line 130, because the `compare_exchange` updates `state` (assuming stopThread cannot be called by multiple threads).

mpoeter (changes requested):

mpoeter (approved):

PR #14913, opened 2021-10-15
Set default module root to cwd

pluma:
### Scope & Purpose

This prevents accidental traversal to the file system root when attempting to resolve module paths while using arangosh. We already set the module root for Foxx services but left it unbounded by default so arangosh wouldn't know when to stop searching. This can have unexpected results when a node_modules folder is present at a higher level of hierarchy.

#### Backports:

- [x] No backports required


pluma:
@jsteemann does this fix the problem you encountered?

pluma:
Test failure seems to be unrelated:

```
* Test "server_permissions"
    [FAILED]  tests/js/client/server_permissions/test-oasis.js

      "testStatisics" failed: Error: at assertion #1: assertTrue: (false) does not evaluate to true
```

CHANGELOG
```diff
@@ -6036,6 +6036,7 @@ devel
 
 * Fixed an agency bug found in Windows tests.
 
+* Limited module resolution in arangosh to the path from which arangosh is invoked.
```
jsteemann, line 6039: Please move this entry to the top of the CHANGELOG, so it becomes the "newest" entry in the "devel" section.

pluma:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/19801/

jsteemann (approved):
LGTM!

PR #14979, opened 2021-10-25
[CINFRA-54] Async Log Compaction

maierlars:
### Scope & Purpose

RemoveFront now add its request to the persistor. It will be part a write batch.

### Testing & Verification

- [x] This change is already covered by existing tests, such as:
     - `tests/Replication2/ReplicatedLog/RocksDBLogTest.cpp`


arangod/RocksDBEngine/RocksDBPersistedLog.h
```diff
@@ -49,17 +50,37 @@ struct RocksDBLogPersistor : std::enable_shared_from_this<RocksDBLogPersistor> {
...
+    struct RemoveFront {
+      replication2::LogIndex stop;
+    };
+
+    auto execute(rocksdb::WriteBatch& wb) -> Result;
+
     std::shared_ptr<arangodb::replication2::replicated_log::PersistedLog> log;
-    std::unique_ptr<arangodb::replication2::PersistedLogIterator> iter;
+    std::variant<InsertEntries, RemoveFront> action;
     futures::Promise<Result> promise;
```
mpoeter, line 79: Can these be const?

arangod/RocksDBEngine/RocksDBPersistedLog.h
```diff
@@ -49,17 +50,37 @@ struct RocksDBLogPersistor : std::enable_shared_from_this<RocksDBLogPersistor> {
...
+                   futures::Promise<Result> promise)
+        : log(std::move(log)), action(RemoveFront{stop}), promise(std::move(promise)) {}
+
+    struct InsertEntries {
+      std::unique_ptr<arangodb::replication2::PersistedLogIterator> iter;
+    };
+
+    struct RemoveFront {
+      replication2::LogIndex stop;
+    };
```
mpoeter, line 73: I would like to define an alias `using Action = std::variant<InsertEntries, RemoveFront>` and have a single `persist` method that takes such an action. (That way we we also need just a single `PersistRequest` constructor that takes that an action). IMHO that would also make the call sites of `persist` clearer.

arangod/Replication2/ReplicatedLog/LogCore.cpp
```diff
@@ -85,7 +85,11 @@ auto replicated_log::LogCore::logId() const noexcept -> LogId {
...
-auto LogCore::removeFront(LogIndex stop) -> Result {
+auto LogCore::removeFront(LogIndex stop) -> futures::Future<Result> {
   std::unique_lock guard(_operationMutex);
-  return _persistedLog->removeFront(stop);
+  return _persistedLog->removeFront(stop).thenValue(
+      [guard = std::move(guard)](Result&& res) mutable {
+        guard.unlock();
+        return std::move(res);
+      });
 }
```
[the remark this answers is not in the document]
maierlars: There is, normally, just a single user of a LogCore. This lock is only there, so that operations started in one term are not interfered by operations that are started from different (new) term.

arangod/RocksDBEngine/RocksDBPersistedLog.h
maierlars: If those member become `const` they have to copied during a move operation. `futures::Promise` is move-only. Additionally `Promise::setValue` is not `const`. `InsertEntries` contains a `unique_ptr` and is therefore move-only, hence the variant becomes move-only as well...

`log` can be const, but then again, it will be copied during move operations.

arangod/RocksDBEngine/RocksDBPersistedLog.h
maierlars: Great idea!

arangod/RocksDBEngine/RocksDBPersistedLog.h
mpoeter: Good point...

mpoeter (approved):

PR #15009, opened 2021-10-28
simplify the usage of SmallVector

jsteemann:
### Scope & Purpose

Simplify the usage of SmallVector, so that the vector and the arena do not have to be declared separately.
In addition, reserve enough space in SmallVector ahead of time in order to use the arena to the fullest.
No backports are planned for this.

### Testing & Verification


lib/Containers/SmallVector.h
```diff
@@ -33,6 +33,129 @@ namespace containers {
...
+  }
+  
+  void push_back(T const& value) {
+    _vector.emplace_back(value);
+  }
+
+  std::size_t size() const noexcept {
+    return _vector.size();
+  }
+
```
mpoeter, line 100: Can we also add `capacity`?
```suggestion
  std::size_t size() const noexcept {
    return _vector.size();
  }

  std::size_t capacity() const noexcept {
    return _vector.capacity();
  }

```

tests/Containers/SmallVectorTest.cpp
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  uint64_t i = 0;
+  auto it = values.rbegin();
+  while (it != values.rend()) {
+    EXPECT_EQ((100 - i - 1) * 3, *it);
+    ++it;
+    ++i;
+  }
+}
```
mpoeter, line 216: I would like to have some tests that check that the initial capacity matches the specified buffer size and that the elements are actually stored inside the arena. I.e., insert capacity number of elements, and for each one check that (pseudocode) `&vec[i] >= &vec && &vec[i] <= &vec + sizeof(vec)`.

jsteemann: Added an extra test.

mpoeter (approved):

PR #15048, opened 2021-11-05
Enable C++20

maierlars:
### Scope & Purpose
This is an experiment to see if we can compile arangodb using c++20 as is. (small modifications needed, because something that was depricated in c++17 has been removed in c++20)

This is a list of changes required and why:
- `char8_t` is now part of C++, it should be used for utf-8 is almost always the same as `unsigned char`
- `std::chrono` has received updates, now there ambiguities with 3rdParty `date`. A follow-up project could be to check if date is still required.
- member relational operator overload resolution has changed. The easiest fix is to make them friend functions (which they should be anyways)
- `std::allocator` removed some typedefs that are depricated in c++17 anyways.
- `requires` now is a keyword. I replaced identifiers with `dependsOn`.
- `std::memory_order::memory_order*` no longer exist. either user `std::memory_order_relaxed` or `std::memory_order::relaxed`.
- Enums no longer decay to their base type. Hence we need a `static_cast<int>` in V8.
- GCC complaint about constructors that look like ClassName<T>. I don't know why they used it in the first place. 
- capturing `this` via `=` is now deprecated.


3rdParty/iresearch/core/search/disjunction.hpp
```diff
@@ -1153,9 +1153,11 @@ class block_disjunction final : public doc_iterator, private score_ctx {
           if constexpr (is_min_match) {
             match_count_ = 1;
           }
-        } else if constexpr (is_min_match) {
-          if (target == value) {
-            ++match_count_;
+        } else {
```
Dronplane, line 1156: Why this change was made? Some compiler bug? constructs like  "else if constexpr" seems to be accepted by gcc, clang and msvc https://godbolt.org/z/fE5Pc9cae

maierlars: The original code caused MSVC to create a warning as indicated:
```c++
      visit_and_purge([this, target, &doc](auto& it) mutable {
        UNUSED(this);
        const auto value = it->seek(target);

        if (doc_limits::eof(value)) {
          // exhausted
          return false;
        }

        // this is to circumvent bug in GCC 10.1 on ARM64
        constexpr bool is_min_match = traits_type::min_match();
        if (value < doc.value) {
          doc.value = value;
          if constexpr (is_min_match) {
            match_count_ = 1;
          }
        } else if constexpr (is_min_match) {
          if (target == value) {
            ++match_count_;
          }
        }

        return true; // warning: not all control paths return a value
      });
```
And it then actually invoked undefined behavior and made the tests fail. This is a compiler bug.

Dronplane: And that warning has appeared only after c++20 was enabled ?

Dronplane: Just tried to build this on my  fresh-updated msvc - no such  warning for this code 😕  Maybe this is because MSVC on Jenkins is outdated?

maierlars: Yes. Either `/std:c++20` or `/std:c++latest`.

maierlars: Interesting, @mpoeter reproduced it locally and in Jenkins. Maybe we just have to upgrade msvc. 🤔

KVS85: Jenkins has the following version of VS 2019: `16.11.5` (was released 12.10.2021). It's not the latest one (which is `16.11.6` from 07.11.2021).

Dronplane: But I've compiled this with "devel" settings e.g. c++17 
I`ll try to compile this branch with  this change reverted and see. Maybe this will change the outcome.

Dronplane:
Also interesting thing  - we have the following in iresearch top level  CMakeLists.txt
`################################################################################
### set required standard version

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)`

So effectively iresearch part was still built with C++17. And that may explain observed misbehaviour.
Actually patching iresearch to  `set(CMAKE_CXX_STANDARD 20)` makes it fail to compile due to use of some deprecated stuff from c++17 (shared_ptr::unique in particular)

Dronplane:
As a temporary solution we could set `_HAS_DEPRECATED_SHARED_PTR_UNIQUE=1` for enabling deprecated stuff

3rdParty/iresearch/core/search/disjunction.hpp
Dronplane: Okay, I could confirm that even the latest MSVC with the c++latest have this warning "not all control paths return a value"  (with the original code)

joerg84 (approved):

PR #15053, opened 2021-11-05

neunhoef:

Problem description:

means of revision trees.

Solution description:

scheduled.

is failed.

The default log level ist now `Info`, we see by default when shard synchronizations
are completed and when maintenance jobs are rescheduled. To not
spam the logs too much, some log messages have been changed from Info to
Debug.

### Scope & Purpose

- [*] :fire: Performance improvement (getting in sync faster)

#### Backports:

- [*] Backport for 3.9: *(Please link PR)*
- [*] Backport for 3.8: *(Please link PR)*

#### Related Information

- [*] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/PRESUPP-413

### Testing & Verification

  - [*] Added new unit test.

### DIFF
--- a/arangod/Cluster/Action.cpp
+++ b/arangod/Cluster/Action.cpp
@@ -217,9 +217,6 @@ void Action::addNewFactoryForTest(std::string const& name,
   factories.emplace(name, std::move(factory));
 }
 
-void addNewFactoryForTest(std::string const& name,
-                          std::function<std::unique_ptr<ActionBase>(MaintenanceFeature&, ActionDescription const&)>&& creator);
-
 namespace std {
 ostream& operator<<(ostream& out, arangodb::maintenance::Action const& d) {
   out << d.toVelocyPack().toJson();
--- a/arangod/Cluster/Action.cpp
+++ b/arangod/Cluster/Action.cpp
@@ -212,10 +212,12 @@ bool Action::operator<(Action const& other) const {
   return false;
 }
 
+#ifdef ARANGODB_USE_GOOGLE_TESTS
 void Action::addNewFactoryForTest(std::string const& name,
       std::function<std::unique_ptr<ActionBase>(MaintenanceFeature&, ActionDescription const&)>&& factory) {
   factories.emplace(name, std::move(factory));
 }
+#endif
 
 namespace std {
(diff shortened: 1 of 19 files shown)


arangod/Cluster/MaintenanceFeature.cpp
```diff
@@ -105,6 +109,39 @@ bool findNotDoneActions(std::shared_ptr<maintenance::Action> const& action) {
 
 }  // namespace
 
+arangodb::Result arangodb::maintenance::collectionCount(arangodb::LogicalCollection const& collection,
```
neunhoef, line 112: This was moved here from `Cluster/SynchronizeShard.cpp` since it is now needed in two places.

arangod/Cluster/MaintenanceWorker.cpp
```diff
@@ -173,7 +178,6 @@ void MaintenanceWorker::nextState(bool actionMore) {
     // finish the current action
     if (_curAction) {
       _lastResult = _curAction->result();
-      _curAction->endStats();
```
neunhoef, line 176: This was duplicate, since it is called further down anyway.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -217,39 +219,6 @@ static arangodb::Result getReadLockId(network::ConnectionPool* pool,
   return res;
 }
 
-static arangodb::Result collectionCount(arangodb::LogicalCollection const& collection,
```
neunhoef, line 220: Has been moved to `Cluster/MaintenanceFeature.cpp` and made public there.

arangod/Cluster/Action.cpp
```diff
@@ -212,6 +212,11 @@ bool Action::operator<(Action const& other) const {
   return false;
 }
 
+void Action::addNewFactoryForTest(std::string const& name,
```
jsteemann, line 215: I think this should be wrapped inside some `#ifdef ARANGODB_USE_GOOGLE_TESTS` ... `#endif` (same for the header file).
```suggestion
void Action::addNewFactoryForTest(std::string const& name,
```

jsteemann (approved):
LGTM

kvahed (commented):
LGTM

PR #15055, opened 2021-11-05

neunhoef:

Problem description:

means of revision trees.

Solution description:

scheduled.

is failed.

The default log level ist now Info, we see by default when
shard synchronizations are completed and when maintenance jobs are
rescheduled. To not spam the logs too much, some log messages have been
changed from Info to Debug.

Scope & Purpose

 -  [*] Bugfix (requires CHANGELOG entry)
 -  [*] Performance improvement (getting in sync faster)

Backports:

  - [*] This is the backport for 3.8 from https://github.com/arangodb/arangodb/pull/15053

Related Information

  - [*] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/PRESUPP-413

Testing & Verification

 - [*] The behavior in this PR was manually tested

        - [*] Added new unit test


arangod/Cluster/EnsureIndex.cpp
```diff
@@ -116,6 +117,27 @@ bool EnsureIndex::first() {
       body.add(VPackObjectIterator(props));
     }
 
+    uint64_t docCount = 0;
+    if (arangodb::maintenance::collectionCount(*col, docCount).fail()) {
+      std::stringstream error;
+      error << "failed to get count of local collection " << shard << " in database " + database;
+      LOG_TOPIC("23561", ERR, Logger::MAINTENANCE) << "EnsureIndex: " << error.str();
+      result(TRI_ERROR_INTERNAL, error.str());
```
jsteemann, line 125: Here I would prefer to not dismiss the actual error, but keep it.
I also suggest to downgrade the log level from ERR to WARN, because the issue may fix itself in a following run.
```suggestion
    if (Result res = arangodb::maintenance::collectionCount(*col, docCount); res.fail()) {
      std::stringstream error;
      error << "failed to get count of local collection " << shard << " in database " << database << ": " << res.errorMessage();
      LOG_TOPIC("23561", WARN, Logger::MAINTENANCE) << "EnsureIndex: " << error.str();
      result(res.errorNumber(), error.str());      
```

arangod/Cluster/MaintenanceFeature.cpp
```diff
@@ -1144,3 +1187,14 @@ MaintenanceFeature::ShardActionMap MaintenanceFeature::getShardLocks() const {
...
 }
+
+Result MaintenanceFeature::requeueAction(std::shared_ptr<maintenance::Action>& action,
+    int newPriority) {
+  TRI_ASSERT(action->getState() == ActionState::COMPLETE ||
+             action->getState() == ActionState::FAILED);
+  auto newAction = std::make_shared<maintenance::Action>(*this, action->describe());
+  newAction->setPriority(newPriority);
+  WRITE_LOCKER(wLock, _actionRegistryLock);
+  registerAction(newAction, false);
```
jsteemann, line 1198: micro optimization
```suggestion
  registerAction(std::move(newAction), false);
```

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -862,15 +866,39 @@ bool SynchronizeShard::first() {
     }
 
     auto ep = clusterInfo.getServerEndpoint(leader);
+    uint64_t docCountOnLeader = 0;
+    if (!collectionCountOnLeader(ep, docCountOnLeader).ok()) {
+      std::stringstream error;
+      error << "failed to get a count on leader " << database << "/" << shard;
+      LOG_TOPIC("1254a", ERR, Logger::MAINTENANCE) << "SynchronizeShard " << error.str();
+      result(TRI_ERROR_INTERNAL, error.str());
+      return false;
```
jsteemann, line 875:
```suggestion
    if (Result res = collectionCountOnLeader(ep, docCountOnLeader); res.fail()) {
      std::stringstream error;
      error << "failed to get a count on leader " << database << "/" << shard << ": " << res.errorMessage();
      LOG_TOPIC("1254a", ERR, Logger::MAINTENANCE) << "SynchronizeShard " << error.str();
      result(res.errorNumber(), error.str());
      return false;
```

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -862,15 +866,39 @@ bool SynchronizeShard::first() {
...
+    }
+
     uint64_t docCount = 0;
     if (!collectionCount(*collection, docCount).ok()) {
       std::stringstream error;
-      error << "failed to get a count on leader " << database << "/" << shard;
+      error << "failed to get a count here " << database << "/" << shard;
```
jsteemann, line 881: I suggest to not dismiss the original error here, so that we have a better error message should this happen, and can also return the proper error number.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1357,9 +1385,18 @@ Result SynchronizeShard::catchupWithExclusiveLock(
 
 void SynchronizeShard::setState(ActionState state) {
   if ((COMPLETE == state || FAILED == state) && _state != state) {
+    bool haveRequeued =
+      result().errorNumber() == TRI_ERROR_ACTION_UNFINISHED;
```
jsteemann, line 1389:
```suggestion
    bool haveRequeued = result().is(TRI_ERROR_ACTION_UNFINISHED);
```

arangod/Cluster/Action.h
```diff
@@ -195,6 +195,22 @@ class Action {
...
 
+  bool requeueRequested() const {
+    return _action->requeueRequested();
+  }
+
+  int requeuePriority() const {
+    return _action->requeuePriority();
+  }
+
+  void pleaseRequeueMe(int requeuePriority) {
```
kvahed, line 206: IMHO `please` could be omitted.

arangod/Cluster/Action.h
```diff
@@ -195,6 +195,22 @@ class Action {
...
+    _action->pleaseRequeueMe(requeuePriority);
```
kvahed, line 207:
```suggestion
  void requeueMe(int requeuePriority) {
    _action->requeueMe(requeuePriority);
```

arangod/Cluster/ActionBase.h
```diff
@@ -193,6 +194,24 @@ class ActionBase {
...
+  bool requeueRequested() const {
+    return _requeueRequested;
+  }
+
+  int requeuePriority() const {
+    TRI_ASSERT(_requeueRequested);
+    return _requeuePriority;
+  }
+
+  void pleaseRequeueMe(int requeuePriority) {
```
kvahed, line 210:
```suggestion
  void requeueMe(int requeuePriority) {
```

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -862,15 +866,39 @@ bool SynchronizeShard::first() {
...
+        docCount != docCountOnLeader &&
+        ((docCount < docCountOnLeader && docCountOnLeader - docCount > 10000) ||
+         (docCount > docCountOnLeader && docCount - docCountOnLeader > 10000))) {
+      // This could be a larger job, let's reschedule ourselves with
+      // priority SLOW_OP_PRIORITY:
+      LOG_TOPIC("25a62", INFO, Logger::MAINTENANCE)
+        << "SynchronizeShard action found that leader's and follower's "
+           "document count differ by more than 10000, will reschedule with "
+           "slow priority, database: " << database << ", shard: " << shard;
+      pleaseRequeueMe(maintenance::SLOW_OP_PRIORITY);
```
kvahed, line 897:
```suggestion
      requeueMe(maintenance::SLOW_OP_PRIORITY);
```

arangod/Cluster/EnsureIndex.cpp
```diff
@@ -116,6 +117,27 @@ bool EnsureIndex::first() {
...
+    }
+
+    if (_priority != maintenance::SLOW_OP_PRIORITY && docCount > 100000) {
+      // This could be a larger job, let's reschedule ourselves with
+      // priority SLOW_OP_PRIORITY:
+      LOG_TOPIC("25a63", INFO, Logger::MAINTENANCE)
+        << "EnsureIndex action found a shard with more than 100000 documents, "
+           "will reschedule with slow priority, database: "
+        << database << ", shard: " << shard;
+      pleaseRequeueMe(maintenance::SLOW_OP_PRIORITY);
```
kvahed, line 136:
```suggestion
      requeueMe(maintenance::SLOW_OP_PRIORITY);
```

arangod/Cluster/MaintenanceFeature.cpp
```diff
@@ -105,6 +109,39 @@ bool findNotDoneActions(std::shared_ptr<maintenance::Action> const& action) {
...
+  std::string collectionName(collection.name());
+  transaction::StandaloneContext ctx(collection.vocbase());
+  SingleCollectionTransaction trx(
+    std::shared_ptr<transaction::Context>(
+      std::shared_ptr<transaction::Context>(), &ctx),
+    collectionName, AccessMode::Type::READ);
+
+  Result res = trx.begin();
+  if (res.fail()) {
+    LOG_TOPIC("5be16", ERR, Logger::MAINTENANCE) << "Failed to start count transaction: " << res;
```
kvahed, line 123: I suggest that this log and the next be `WARN` level

arangod/Cluster/MaintenanceFeature.cpp
```diff
@@ -105,6 +109,39 @@ bool findNotDoneActions(std::shared_ptr<maintenance::Action> const& action) {
...
+
+  if (res.fail()) {
+    LOG_TOPIC("26ed2", ERR, Logger::MAINTENANCE)
+        << "Failed to finish count transaction: " << res;
+    return res;
+  }
+
+  VPackSlice s = opResult.slice();
+  TRI_ASSERT(s.isNumber());
+  c = s.getNumber<uint64_t>();
```
kvahed, line 140:
```suggestion
  c = s.getNumber<int64_t>();
  if (c < 0) {
    LOG_TOPIC(...);
    return ...;
  }
```

kvahed: Just a suggestion. Not sure this would even be a problem.

kvahed (changes requested):

arangod/Cluster/Action.h
neunhoef: Done.

arangod/Cluster/MaintenanceFeature.cpp
neunhoef: Done.

arangod/Cluster/MaintenanceFeature.cpp
neunhoef: No, counts are created to be `uint64_t`, negative results are not possible.

arangod/Cluster/SynchronizeShard.cpp
neunhoef: Done.

neunhoef (commented):
Done requested changes.

jsteemann (approved):
LGTM!

PR #15063, opened 2021-11-08
Bug fix/agency cleanup

jsteemann:
### Scope & Purpose

This PR contains several micro optimizations and refactorings for agency code. More detailed commit messages can be found in the underlying commits for this PR.
A few things that changed:

* creator sites of `trans_ret_t`, `read_ret_t` and `write_ret_t` now move data into these structs if possible
* make `AgentInterface`'s `isCommitted()` method const. It wasn't before
* remove optionality from `query_t` parameter in `Agent::triggerPollsNoLock()`. The parameter was a nullptr only in a special case. This case is now handled at the call site, so that `triggerPollsNoLock()` does not need to handle a special case.
* the `Store` API methods were previously based on `query_t`, which is an `std::shared_ptr<Builder>`. While this works, it requires all callers to create `shared_ptr`s to use the `Store` API. This is now changed so that the `Store` methods accept `Slice`s instead.
* Clean up a few other places in the agency that unnecessarily wrapped data in `std::shared_ptr`s.
* Remove non-implemented `read()` method from `Store` API: there were previously 3 `read` methods in the header file, but only 2 of them were actually implemented. Removed the obsolete one.
* Previously there were 2 overloads of `Store::read(...)`, with different parameters. One of them was used to read multiple values, and the other was used to read a single value. Renamed the multi-method to `readMultiple()` to avoid confusion, and adjust all callers.
* Remove `virtual` method from `Store`. `Store` does not need to be virtual, as there is no class that derives from it.This was tested by temporariliy making the `Store` class `final`.

As a follow-up, one could improve the APIs in `AgentInterface` so that they can be used with `std::shared_ptr<Builder>`s, but with simple `Slices`. This would allow removing a lot of temporary `shared_ptr` creation and overhead from other agency methods (RestHandler etc.).

### Testing & Verification

- [x] This change is already covered by existing tests, such as *agency, gtest*.


arangod/Agency/Agent.h
```diff
@@ -138,7 +138,7 @@ class Agent final : public arangodb::Thread, public AgentInterface {
   void removeTrxsOngoing(Slice trxs) noexcept;
 
   /// @brief Check whether a trx is ongoing.
-  bool isTrxOngoing(std::string& id);
+  bool isTrxOngoing(std::string const& id) const noexcept;
```
maierlars, line 141: can't we use a `std::string_view` instead?

maierlars (approved):

arangod/Agency/Agent.h
jsteemann: We could, but the function internally will convert the `id` to a string anyway, because it does a lookup for the id in an `std::unordered_set<std::string>`. So without heterogenous lookups, we won't save any conversion here.

maierlars: 😭 C++20 is soo close...

neunhoef (approved):
LGTM

PR #15065, opened 2021-11-08
remove duplicate code, arangod/GeneralServer/GeneralServerFeature.cpp already has these.

dothebart:
### Scope & Purpose

cleanup duplicate code. See https://github.com/arangodb/arangodb/blob/devel/arangod/GeneralServer/GeneralServerFeature.cpp#L174 for the remaining place.

### Testing & Verification

```
./bin/arangod --server.allow-method-override true  --server.hide-product-header false --server.keep-alive-timeout 10 --no-server false  /tmp/blarg
```
However, 
```
./bin/arangod --server.default-api-compatibility true /tmp/blarg 
Error while processing command-line options for arangod:
  unknown option '--'

Did you mean one of these?
  --gid     switch to group-id after reading config files
  --hund    make ArangoDB bark on startup
  --log     the global or topic-specific log level
  --uid     switch to user-id after reading config files

Use --help or --help-all to get an overview of available options
```
is probably bad behaviour because of bad error messages, maybe it should be removed alltogether?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
+* remove the not anymore supported --server.default-api-compatibility option
+
 * Fixed SEARCH-261: Fix possible race between file creation and directory
   cleaner (ArangoSearch).
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,7 @@
 devel
 -----
 
-* remove the not anymore supported --server.default-api-compatibility option
+* Remove unsupported `--server.default-api-compatibility` startup option.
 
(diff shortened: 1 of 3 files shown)


jsteemann (approved):
LGTM

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 devel
 -----
 
+* remove the not anymore supported --server.default-api-compatibility option
```
```suggestion
* Remove unsupported `--server.default-api-compatibility` startup option.
```

jsteemann (approved):
LGTM

PR #15070, opened 2021-11-08
Bug fix/avoid cpp17 deprecations

mpoeter:
### Scope & Purpose

Fix various places where STL features have been used that have been deprecated in C++17 and removed in C++20.

### Testing & Verification

### DIFF
--- a/3rdParty/immer/v0.6.2/tools/include/nonius.h++
+++ b/3rdParty/immer/v0.6.2/tools/include/nonius.h++
@@ -658,13 +658,13 @@ namespace nonius {
                 return {};
             }
         };
-        template <typename Sig>
-        using ResultOf = typename std::invoke_result<Sig>::type;
+        template <typename Fun, typename... Args>
+        using ResultOf = typename std::invoke_result<Fun, Args...>::type;
 
         // invoke and not return void :(
         template <typename Fun, typename... Args>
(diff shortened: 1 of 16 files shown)


3rdParty/immer/v0.6.2/tools/include/nonius.h++
```diff
@@ -659,7 +659,7 @@ namespace nonius {
             }
         };
         template <typename Sig>
-        using ResultOf = typename std::result_of<Sig>::type;
+        using ResultOf = typename std::invoke_result<Sig>::type;
```
maierlars, line 662: Is that actually correct?

maierlars (approved):

3rdParty/immer/v0.6.2/tools/include/nonius.h++
mpoeter: Good catch - probably not! To make this correct we would have to change the template signature.

mpoeter: Fixed.

markuspf (commented):
Are the changes to `immer` unnecessary if #15072 is merged?

maierlars:
Yes, immer is a c++11 library, so they can not use invoke_result.

PR #15076, opened 2021-11-09
Bugfix/msvc build optimization

mpoeter:
### Scope & Purpose

Improves several build configuration settings for MSVC:
  - remove `/Ob1` which limited inline function expansion to functions explicitly marked as `inline`.
  - enable SSE42 and ASM_OPTIMIZATIONS
  - increase the reserved stack size from the default 1MB to 4MB. This should allow execution of larger queries without overflowing the stack.
  - enable COMDAT folding and elimination of functions and data that are never referenced. This reduced the size of the arangod.exe for a non-maintainer build from 61MB to 39MB and the PDB from 753MB to 443MB.

#### Backports:

_To be discussed_

### Testing & Verification


CMakeLists.txt
```diff
@@ -974,17 +978,17 @@ if (MSVC)
...
   set(BASE_FLAGS "/wd4996 ${BASE_FLAGS}")
 
-  set(CMAKE_C_FLAGS                "/MTd"                                              CACHE INTERNAL "default C compiler flags")
-  set(CMAKE_C_FLAGS_DEBUG          "/D _DEBUG /MTd /Zi /Ob0 /Od /RTC1 /bigobj"         CACHE INTERNAL "C debug flags")
-  set(CMAKE_C_FLAGS_MINSIZEREL     "/MT /O1 /Ob1"                                      CACHE INTERNAL "C minimal size flags")
-  set(CMAKE_C_FLAGS_RELEASE        "/MT /O2 /Ob2"                                      CACHE INTERNAL "C release flags")
-  set(CMAKE_C_FLAGS_RELWITHDEBINFO "/MT /Zi /O2 /Ob1"                                  CACHE INTERNAL "C release with debug info flags")
-
-  set(CMAKE_CXX_FLAGS                "/MTd"                                            CACHE INTERNAL "default C++ compiler flags")
-  set(CMAKE_CXX_FLAGS_DEBUG          "/D _DEBUG /MTd /Zi /Ob0 /Od /RTC1 /bigobj"       CACHE INTERNAL "C++ debug flags")
```
Dronplane, line 984: Is it intentional to remove /bigobj ? AFAIK it was added on purpose when we have faced  compiler error.

Dronplane (approved):
LGTM.  Please verify the "/bigobj" removal.

CMakeLists.txt
mpoeter: It is added unconditionally for all configurations here: https://github.com/arangodb/arangodb/blob/15f8011b7116367b3c6edd11fc1b339319832202/CMakeLists.txt#L998

Dronplane: Ah, nice. Than it is ok here.

jsteemann (commented):
LGTM!

PR #15078, opened 2021-11-09

neunhoef:

Problem description:

means of revision trees.

Solution description:

scheduled.

is failed.

The default log level ist now Info, we see by default when
shard synchronizations are completed and when maintenance jobs are
rescheduled. To not spam the logs too much, some log messages have been
changed from Info to Debug.

Scope & Purpose

    [*] Performance improvement (getting in sync faster)

Backports:

    [*] This is the backport for 3.8 from 

    Introduce slow track in maintenance #15053

Related Information

    [*] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/PRESUPP-413

Testing & Verification

    This PR adds tests that were used to verify all changes:
    - [*] Added new unit test.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 v3.9.0 (XXXX-XX-XX)
 -------------------
 
+* Fixed BTS-637: Slow SynchronizeShard jobs which need to copy data could
+  block quick SynchronizeShard jobs which have the data and only need to
+  resync.
+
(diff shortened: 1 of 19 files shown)


jsteemann (approved):
LGTM!

jsteemann (changes requested):
Ah, seems one commit is missing here.

jsteemann (approved):
Never mind. I was confused when reviewing this.
The PR is ok and nothing is missing.
LGTM!

PR #15081, opened 2021-11-09
Bug fix/bts 590

ajurdak:
### Scope & Purpose

It was possible to create a database despite the value of write concern was bigger than the value of the replication factor. Now these two values will be compared and in case the write concern is bigger than the replication factor no database will be created and an error message will be displayed.

### Testing & Verification


CHANGELOG
```diff
@@ -1,6 +1,10 @@
 devel
 -----
 
+* When creating a new database the value of the write concern has to be smaller
+  or equal to the replication factor. Otherwise an error message will be displayed
+  and no database will be created.
```
hkernbach, line 6:
```suggestion
* In the shards overview the list of servers to move the leader shard to, now
  also contains the current followers. This means that from now on also active
  follower servers can be nominated as the leading server for that specific
  shard.
```

hkernbach (changes requested):
Code LGTM.

Changelog entry was a bit confusing. Tried to improve there. Please check. Feel free to optimize (e.g. language) even further.

CHANGELOG
hkernbach: Sorry, wrong PR addressed here.

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
```diff
@@ -176,34 +176,39 @@
...
-          self.handleError(err, dbname);
-        },
-        success: function (data) {
-          if (window.location.hash === '#databases') {
-            self.updateDatabases();
+      
+      let replicationFactor = $('#new-replication-factor').val();
+      let writeConcern = $('#new-write-concern').val();
+
+      if(parseInt(writeConcern) <= parseInt(replicationFactor)) {
```
hkernbach, line 183: What happens in case replicationFactor || writeConcern will not be typeof String? 
(Before it was the same - but I think we should validate input here) - Try / catch would help I guess.

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
```diff
@@ -176,34 +176,39 @@
...
           }
-          arangoHelper.arangoNotification('Database ' + _.escape(data.get('name')) + ' created.');
-        }
-      });
+        });
 
-      arangoHelper.arangoNotification('Database creation in progress.');
+        arangoHelper.arangoNotification('Database creation in progress.');
+      } else {
+        arangoHelper.arangoError('DB', 'No database created. Write concerns must be smaller or equal compared to the replication factor.');
```
hkernbach, line 210: Totally fine, but just asking: Why do we need that implementation now? Shouldn't the backend handle it anyway?

hkernbach (approved):
Code looks good. I would vote for better input verification (see comment). But as it has been this way for a longer time - doing it like this also fine. Your decision.

Zyqsempai:
@ajurdak @hkernbach Can we merge this one?

mpoeter:
@Zyqsempai just merged in devel and started another Jenkins. Once it is through this can be merged.

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
mpoeter: Yes, this should definitely be handled in the backend, but as it currently isn't, it is better to at least check it in the frontend.

mpoeter (approved):

Zyqsempai:
@ajurdak Let's push it forward, I mean merge it;)

PR #15083, opened 2021-11-09
[devel] clang-format operations

aMahanna:
### Scope & Purpose

**Objective**
* Introduce a consistent clang-format standard for ArangoDB via Githooks and Actions.

**Implementation**
1.[ Establish a baseline](https://github.com/arangodb/arangodb/pull/15384) by formatting the current codebase via a derivation of the Google style 
(created by @gnusi, see https://github.com/arangodb/arangodb/pull/15131)
2. Introduce a pre-commit githook to automatically format added/modified files locally.
3. Introduce a Github Action workflow to alert on "leftover" clang-format violations

**Baseline example**
* `arangod/**/*` : 78bb583
* `arangosh/**/*`: f4c3b3d
* `lib/**/*`: 8410968
* `tests/**/*`: f0f6424


aMahanna:
Will establish a clang-format baseline of the codebase upon receiving target directory & version confirmation.

scripts/clang-format/clang-format.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cd ${PWD}/scripts/clang-format # assume current dir is arangodb (not error-proof)
+docker build -t clang-format .
+while [[ $PWD != '/' && ${PWD##*/} != 'arangodb' ]]; do cd ..; done # cd up to arangodb
```
jsteemann, line 4: The assumption of having a directory named `arangodb` feels strange and doesn't work for me right now (and probably won't work for a lot of others).

scripts/clang-format/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+cd arangodb
+#find arangod -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
+find arangosh -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
```
jsteemann, line 4: IMO we don't need to care about `.cu` files here. A common file extension for C++ files is also `.cc`, which is not in the list (and also not used by us). We may or may not want to add this to be "future-proof" :sweat_smile:

scripts/clang-format/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+cd arangodb
```
jsteemann, line 2: The assumption of having to have an `arangodb` directory does not feel right.

scripts/clang-format/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
...
+#find arangod -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
+find arangosh -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
+#find lib -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
+#find tests -regex '.*\.\(cpp\|hpp\|cu\|c\|h\)' -exec clang-format -style=file -i {} \;
```
jsteemann, line 6: If this script is supposed to be run locally by developers, we should also add `enterprise` here, if it exists. So we should test for the existence of the `enterprise` subdirectory, and then apply formatting there too.

scripts/clang-format/clang-format.sh
jsteemann: We may get away with something like `pushd ... popd` here. These seem to be supported by most of the relevant shells today.

scripts/clang-format/clang-format.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cd ${PWD}/scripts/clang-format # assume current dir is arangodb (not error-proof)
```
jsteemann, line 2: I think instead of `${PWD}/scripts/clang-format` we should use `"${PWD}/scripts/clang-format"` here (plus of all following occurrences of variables). Otherwise it may not work with directory names that contain spaces.

aMahanna:
note: PR is on hold until further notice (awaiting publish of the `arangodb/clang-format` image based on https://github.com/aMahanna/arangodb-clang-format/tree/master)

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: Ana06/get-changed-files@v2.0.0
+        id: gcf
+        with:
+          filter: |
+            **/*.cpp
+            **/*.hpp
+            **/*.cc
+            **/*.c
+            **/*.h
+      - uses: aMahanna/clang-format-checker@v1.2.2
```
aMahanna, line 17: This will be changed to `arangodb/clang-format-action@v1.0.0` once we publish https://github.com/arangodb/clang-format-action

aMahanna: Note: CI is failing (https://github.com/arangodb/arangodb/runs/4576063938?check_suite_focus=true) because the ArangoDB Clang-Format Action is yet to be published (see https://github.com/arangodb/clang-format-action)

aMahanna:
Note: https://github.com/arangodb/arangodb/pull/15384 and https://github.com/arangodb/enterprise/pull/831 must be merged before Squashing & Merging this PR

scripts/clang-format.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ -d "$adb_path/$ent_dir" ] # assume enterprise directory is within arangodb directory
+then
+   cd "$adb_path/$ent_dir"
+   enterprise_diff=$(git diff HEAD --diff-filter=ACMR --name-only -- '*.cpp' '*.hpp' '*.cc' '*.c' '*.h' | sed "s,^,$ent_dir/,")
+fi
+
+diff="$community_diff $enterprise_diff"
+if ! [[ -z "${diff// }" ]]; then
+  docker run --rm -v "$adb_path":/usr/src/arangodb arangodb/clang-format:1.0 "$diff"
```
neunhoef, line 21: On Linux systems, this will change ownership of all files touched to `root.root`. This would be a nuisance. However, it is easy to provide a second script which runs `clang-format` locally without docker, solving the problem.

neunhoef (commented):
I like the fact that the `clang-format.sh` script only looks at changes in comparison to the HEAD. But the file ownership is a (small) problem on Linux.

scripts/clang-format.sh
mpoeter: Running the container with `--user $(id -u):$(id -g)` should avoid this ownership problem.

aMahanna: I've created a separate [clang-format-linux.sh](https://github.com/arangodb/arangodb/pull/15083/files#diff-fe2859c9145b46a037443925b6569a707229f36e1353d4e227eb06513efa50c6) file to include the changes we discussed.

This script actually works fine on Mac as well (tested it locally). If it happens to work on Windows too, then there will be no need to have 2 different shell scripts

neunhoef: Yes, I would think so. However, changes are probably not great, but I am no Windows expert.

aMahanna:
OS-dependent hooks & scripts (MacOS, Linux, Windows) have been introduced.

Linux & MacOS implementations are currently identical, but may be subject to change in the future (hence why I still decided to split them up).

joerg84 (approved):

PR #15085, opened 2021-11-09
Feature/shard sync improvements

jsteemann:
### Scope & Purpose

* Close a potential gap during shard synchronization when moving from the
  initial sync step to the WAL tailing step. In this small gap the leader 
  could purge some of the WAL files that would be required by the following
  WAL tailing step. This was possible because at the end of the initial sync
  step, the snapshot on the leader is released, and there is a small window
  of time before the follower will issue its first WAL tailing request.

* Improve Shards overview in web UI: the number of currently syncing shards is
  now displayed per collection. Additionally, shards on failed servers are now
  displayed in a different color.

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [x] Backport for 3.9: *https://github.com/arangodb/arangodb/pull/15086*
- [x] Backport for 3.8: *https://github.com/arangodb/arangodb/pull/15087*
- [ ] Backport for 3.7: *(Please link PR)*

### Testing & Verification

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

PR #15087, opened 2021-11-09
Feature 3.8/shard sync improvements

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15085

* Close a potential gap during shard synchronization when moving from the
  initial sync step to the WAL tailing step. In this small gap the leader 
  could purge some of the WAL files that would be required by the following
  WAL tailing step. This was possible because at the end of the initial sync
  step, the snapshot on the leader is released, and there is a small window
  of time before the follower will issue its first WAL tailing request.

* Improve Shards overview in web UI: the number of currently syncing shards is
  now displayed per collection. Additionally, shards on failed servers are now
  displayed in a different color.

#### Backports:

### Testing & Verification

### DIFF
--- a/arangod/Cluster/SynchronizeShard.cpp
+++ b/arangod/Cluster/SynchronizeShard.cpp
@@ -802,62 +802,12 @@ bool SynchronizeShard::first() {
           "/" + resolver.getCollectionName(collection->id());
     }
 
-    // old "shortcut" code for getting in sync. This code takes a shortcut if the
-    // shard in question is supposed to be empty. in this case it will simply try
-    // to add itself as an in-sync follower, without running the full replication
-    // protocol. this shortcut relies on the collection counts being correct, and
-    // as these can be at least temporarily off, we need to disable it.
-#if 0
-    if (docCount == 0) {
-      // We have a short cut:
-      LOG_TOPIC("0932a", DEBUG, Logger::MAINTENANCE)
-          << "synchronizeOneShard: trying short cut to synchronize local shard "
-             "'"
-          << database << "/" << shard << "' for central '" << database << "/"
-          << planId << "'";
-
-      // now do a final sync-to-disk call. note that this can fail
-      auto& engine = vocbase->server().getFeature<EngineSelectorFeature>().engine();
-      Result res = engine.flushWal(/*waitForSync*/ true, /*waitForCollector*/ false);
-      if (res.fail()) {
-        LOG_TOPIC("a49d1", INFO, Logger::MAINTENANCE) << res.errorMessage();
-        result(res);
-        return false;
-      }
-
-      try {
-        NetworkFeature& nf = _feature.server().getFeature<NetworkFeature>();
-        network::ConnectionPool* pool = nf.pool();
-        auto asResult = addShardFollower(pool, ep, database, shard, 0, clientId,
-                                         SyncerId{}, _clientInfoString, 60.0);
-
-        if (asResult.ok()) {
-          if (Logger::isEnabled(LogLevel::DEBUG, Logger::MAINTENANCE)) {
-            std::stringstream msg;
-            msg << "SynchronizeShard: shortcut worked, done, ";
-            AppendShardInformationToMessage(database, shard, planId, startTime, msg);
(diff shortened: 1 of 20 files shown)


arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -879,61 +797,40 @@ bool SynchronizeShard::first() {
...
+      }
+      
+      ReplicationApplierConfiguration configuration =
+          ReplicationApplierConfiguration::fromVelocyPack(feature().server(), builder.slice(), getDatabase());
+      // will throw if invalid
+      configuration.validate();
+  
+      // build DatabaseTailingSyncer object for WAL tailing
+      TRI_ASSERT(tailingSyncer == nullptr);
+      tailingSyncer = DatabaseTailingSyncer::create(guard.database(), configuration, /*lastTick*/ 0, /*useTick*/true);
```
neunhoef, line 825: I suggest to check `tailingSyncer` for `nullptr` here such that we know later on that it is valid. This saves several checks later. And if it is `nullptr` here, we might as well give up right away.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1017,11 +913,12 @@ bool SynchronizeShard::first() {
       auto lastTick =
         arangodb::basics::VelocyPackHelper::getNumericValue<TRI_voc_tick_t>(
           sy, LAST_LOG_TICK, 0);
-      VPackBuilder builder;
 
       ResultT<TRI_voc_tick_t> tickResult =
-        catchupWithReadLock(ep, *collection, clientId, leader, lastTick, builder);
+          catchupWithReadLock(ep, *collection, clientId, leader, lastTick, tailingSyncer);
```
neunhoef, line 918: With the check suggested above, this call is safe, since the called function assumes that `tailingSyncer` is not a nullptr.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1017,11 +913,12 @@ bool SynchronizeShard::first() {
...
+
       if (!tickResult.ok()) {
+        tailingSyncer->unregisterFromLeader();
```
neunhoef, line 921: With the above check this is safe.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1030,13 +927,17 @@ bool SynchronizeShard::first() {
 
       // Now start an exclusive transaction to stop writes:
       Result res = catchupWithExclusiveLock(ep, *collection, clientId,
-                                            leader, syncerId, lastTick, builder);
+                                            leader, syncerId, lastTick, tailingSyncer);
```
neunhoef, line 930: With the above check this is safe.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1030,13 +927,17 @@ bool SynchronizeShard::first() {
...
       if (!res.ok()) {
+        tailingSyncer->unregisterFromLeader();
```
neunhoef, line 932: With the above check this is safe.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1030,13 +927,17 @@ bool SynchronizeShard::first() {
...
         LOG_TOPIC("be85f", INFO, Logger::MAINTENANCE) << res.errorMessage();
         result(res);
         return false;
       }
     } catch (basics::Exception const& e) {
+      if (tailingSyncer != nullptr) {
```
neunhoef, line 938: This check would be unnecessary with the above proposed check.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1049,6 +950,10 @@ bool SynchronizeShard::first() {
       result(e.code(), e.what());
       return false;
     } catch (std::exception const& e) {
+      if (tailingSyncer != nullptr) {
```
neunhoef, line 953: With the above check this is unnecessary.

neunhoef (approved):
LGTM with a tiny exception: I suggest to change the check for nullptr for `tailingSyncer` in `SynchronizeShard`, see comments.

PR #15089, opened 2021-11-10
remove unused code from repository

jsteemann:
### Scope & Purpose

Remove unused code from repository.
Only entire non-virtual methods were removed, so that the C++ compiler will find any errors for us.
An exception is the removal of the handling code for "getSome" / "skipSome" calls in RestAqlHandler.cpp. This code seemed obsolete, so hopefully it can be removed too.

### Testing & Verification


arangod/Aql/RestAqlHandler.cpp
```diff
@@ -651,31 +651,10 @@ auto AqlExecuteCall::fromVelocyPack(VPackSlice const slice) -> ResultT<AqlExecut
 
 // handle for useQuery
 RestStatus RestAqlHandler::handleUseQuery(std::string const& operation,
-                                          VPackSlice const querySlice) {
+                                          VPackSlice querySlice) {
```
cpjulia, line 654: Can't this still be a const?

cpjulia (approved):
LGTM

arangod/Aql/RestAqlHandler.cpp
jsteemann: Yes, it could still be const but as the VPackSlice is passed by value, making it const or non-const won't change the API. It only affects the const-ness of the `querySlice` variable inside `handleUseQuery`. As a `VPackSlice` does not have any mutators, it can always be passed as non-const copy.

PR #15094, opened 2021-11-10
[3.9] MSVC build optimizations

mpoeter:
### Scope & Purpose

Improves several build configuration settings for MSVC:
  - remove `/Ob1` which limited inline function expansion to functions explicitly marked as `inline`.
  - enable SSE42 and ASM_OPTIMIZATIONS
  - increase the reserved stack size from the default 1MB to 4MB. This should allow execution of larger queries without overflowing the stack.
  - enable COMDAT folding and elimination of functions and data that are never referenced. This reduced the size of the arangod.exe for a non-maintainer build from 61MB to 39MB and the PDB from 753MB to 443MB.

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.9.0 (XXXX-XX-XX)
 -------------------
 
+* Increase default stack size on Windows from 1MB to 4MB. This should allow
+  execution of larger queries without overflowing the stack.
+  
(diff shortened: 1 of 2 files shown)


jsteemann (commented):
LGTM!

KVS85 (approved):
LGTM.

PR #15114, opened 2021-11-15
Extend Windows minidump with referenced memory regions.

mpoeter:
### Scope & Purpose

Extend Windows minidumps with memory regions referenced from CPU registers or the stack to provide more contextual information in case of crashes.

#### Backports:

- [x] Backport for 3.8: #15115

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.9.0 (XXXX-XX-XX)
 -------------------
 
+* Extend Windows minidumps with memory regions referenced from CPU registers or
+  the stack to provide more contextual information in case of crashes.
+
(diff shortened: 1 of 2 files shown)


Dronplane:
Should we also add "MiniDumpIgnoreInaccessibleMemory" flag for MiniDumpWriteDump ?  
As we now execute memory reads and IsBadReadPtr  could give us "false positive"

mpoeter:
@Dronplane good point! In my tests I definitely had some invalid addresses in the list and it worked nonetheless. Anyway, I will also add `MiniDumpIgnoreInaccessibleMemory` just in case.

Dronplane (approved):
LGTM

PR #15117, opened 2021-11-15
Bugfixes for old incremental sync protocol

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15120 (plus additional changes)

* Fixed an issue in old incremental sync protocol with document keys that
  contained special characters (`%`). These keys could be send unencoded in
  the incremental sync protocol, leading to wrong key ranges being transfered
  between leader and follower, and thus causing follow-up errors and preventing
  getting in sync.

* Additionally, this PR improves the behavior of savepoints during
  incremental sync: previously, if during a preflight check it was found
  that a document cannot be inserted/updated, some LogData leftovers for
  the document may have remained in the WriteBatchWithIndex. These will
  have made it into WAL when the overall transaction was committed. This
  was probably harmless, but is now cleaned up.

### Testing & Verification

  - [x] Added new **integration tests** (replication_sync, replication_ongoing)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,12 @@
 v3.7.16 (XXXX-XX-XX)
 --------------------
 
+* Fixed an issue in old incremental sync protocol with document keys that 
+  contained special characters (`%`). These keys could be send unencoded in
+  the incremental sync protocol, leading to wrong key ranges being transfered
+  between leader and follower, and thus causing follow-up errors and preventing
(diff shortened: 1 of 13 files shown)


neunhoef (approved):
LGTM. I suggest that we address the `NO_INDEXING` in a separate PR and also change then that we commit the transaction after every chunk to keep the RocksDB transaction smaller.

PR #15131, opened 2021-11-16
Update .clang-format style file

aMahanna:
(in-preparation for https://github.com/arangodb/arangodb/pull/15083)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -123,7 +123,7 @@ RawStringFormats:
     CanonicalDelimiter: ''
     BasedOnStyle:    google
 ReflowComments:  true
-SortIncludes:    true
+SortIncludes:    false
 SortUsingDeclarations: true
 SpaceAfterCStyleCast: false
 SpaceAfterTemplateKeyword: false
--- a/.clang-format
+++ b/.clang-format
@@ -140,7 +140,7 @@ SpacesInContainerLiterals: true
 SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
-Standard: Cpp11
+Standard: Latest
 StatementMacros:
   - Q_UNUSED
   - QT_REQUIRE_VERSION
--- a/.clang-format
+++ b/.clang-format
@@ -16,9 +16,10 @@ AllowShortLoopsOnASingleLine: true
 AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: true
 BinPackParameters: true
-BraceWrapping:   
+BraceWrapping:
   AfterClass:      false
   AfterControlStatement: false
   AfterEnum:       false
@@ -37,6 +38,7 @@ BraceWrapping:
 BreakBeforeBinaryOperators: None
 BreakBeforeBraces: Attach
 BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
 BreakBeforeTernaryOperators: true
 BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
@@ -49,16 +51,16 @@ ConstructorInitializerAllOnOneLineOrOnePerLine: true
 ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4
 Cpp11BracedListStyle: true
-DerivePointerAlignment: false
+DerivePointerAlignment: true
 DisableFormat:   false
 ExperimentalAutoDetectBinPacking: false
 FixNamespaceComments: true
-ForEachMacros:   
+ForEachMacros:
   - foreach
   - Q_FOREACH
   - BOOST_FOREACH
-IncludeBlocks:   Preserve
-IncludeCategories: 
+IncludeBlocks:   Regroup
+IncludeCategories:
   - Regex:           '^<ext/.*\.h>'
     Priority:        2
   - Regex:           '^<.*\.h>'
@@ -79,22 +81,58 @@ MacroBlockBegin: ''
 MacroBlockEnd:   ''
 MaxEmptyLinesToKeep: 1
 NamespaceIndentation: None
+ObjCBinPackProtocolList: Never
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
 PenaltyBreakAssignment: 2
-PenaltyBreakBeforeFirstCallParameter: 100
+PenaltyBreakBeforeFirstCallParameter: 1
 PenaltyBreakComment: 300
 PenaltyBreakFirstLessLess: 120
 PenaltyBreakString: 1000
-PenaltyExcessCharacter: 1
+PenaltyBreakTemplateDeclaration: 10
+PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 200
 PointerAlignment: Left
-BasedOnStyle:    google
+RawStringFormats:
+  - Language:        Cpp
+    Delimiters:
+      - cc
+      - CC
+      - cpp
+      - Cpp
+      - CPP
+      - 'c++'
+      - 'C++'
+    CanonicalDelimiter: ''
+    BasedOnStyle:    google
+  - Language:        TextProto
+    Delimiters:
+      - pb
+      - PB
+      - proto
+      - PROTO
+    EnclosingFunctions:
+      - EqualsProto
+      - EquivToProto
+      - PARSE_PARTIAL_TEXT_PROTO
+      - PARSE_TEST_PROTO
+      - PARSE_TEXT_PROTO
+      - ParseTextOrDie
+      - ParseTextProtoOrDie
+    CanonicalDelimiter: ''
+    BasedOnStyle:    google
 ReflowComments:  true
 SortIncludes:    true
 SortUsingDeclarations: true
 SpaceAfterCStyleCast: false
-SpaceAfterTemplateKeyword: true
+SpaceAfterTemplateKeyword: false
 SpaceBeforeAssignmentOperators: true
+SpaceBeforeCpp11BracedList: false
+SpaceBeforeCtorInitializerColon: true
+SpaceBeforeInheritanceColon: true
 SpaceBeforeParens: ControlStatements
+SpaceBeforeRangeBasedForLoopColon: true
 SpaceInEmptyParentheses: false
 SpacesBeforeTrailingComments: 2
 SpacesInAngles:  false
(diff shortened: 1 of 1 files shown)


maierlars (approved):

aMahanna:
It appears we are encountering a clang-format warning when using this style file in `arangodb/lib`, more specifically when we want to format the header (`.h`) files within that directory.

Running `find lib -regex '.*\.\(h\)' -exec clang-format -style=file -i {} \;` results in the following warning:
```
Configuration file(s) do(es) not support Objective-C: /usr/src/arangodb/.clang-format, /usr/src/.
```

.clang-format
```diff
@@ -79,31 +81,68 @@ MacroBlockBegin: ''
...
+SpaceBeforeRangeBasedForLoopColon: true
 SpaceInEmptyParentheses: false
 SpacesBeforeTrailingComments: 2
 SpacesInAngles:  false
 SpacesInContainerLiterals: true
 SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
-Standard:        Cpp11
+Standard: Cpp11
```
MBkkt, line 143: `c++20`?
What version of the clang-format are we using? 13?

.clang-format
```diff
@@ -79,31 +81,68 @@ MacroBlockBegin: ''
...
+      - PARSE_PARTIAL_TEXT_PROTO
+      - PARSE_TEST_PROTO
+      - PARSE_TEXT_PROTO
+      - ParseTextOrDie
+      - ParseTextProtoOrDie
+    CanonicalDelimiter: ''
+    BasedOnStyle:    google
 ReflowComments:  true
-SortIncludes:    true
+SortIncludes:    false
```
MBkkt, line 126: I think we still need to sort the includes and get rid of the fragile code that depends on the order of the includes

MBkkt (approved):

.clang-format
aMahanna: we are using version clang-format `10.0.0`, via the following docker image: https://github.com/arangodb/clang-format-docker

aMahanna: @gnusi, can you confirm that we want to stick with `Cpp11`?

MBkkt: Why run a docker image for a clang-format, why not just use ubuntu or another linux machine?

MBkkt: https://packages.ubuntu.com/bionic/clang-format-10

aMahanna: We want consistent clang-format usage across all developer machines. @joerg84 and I established that this would be a good start, since we can enforce a clang-format version through a simple distribution like Alpine. It also mirrors the clang-format version we've established for our Github Action: https://github.com/aMahanna/clang-format-checker

aMahanna: The pre-commit then uses that docker image here: https://github.com/arangodb/arangodb/pull/15083/files#diff-87320095d7c4f39eed5d8f3866b512eb910e4eec8e7926faaeef6a53ab786fcbR9

.clang-format
aMahanna: good to know, I'm just not sure what the course of action would be if we set `SortIncludes` back to `true`. Do we hold off any clang-format feature until the fragile code is resolved? Who would we reach out to to solve our fragile code issue? How do we prevent having more fragile code like this in the future?

MBkkt: > good to know, I'm just not sure what the course of action would be if we set SortIncludes back to true. Who would we reach out to solve our fragile code issue?

Have you tried reformatting and building the entire repository? It seems the only expected problem is `velocypack-aliases` (the solution is to delete this file or write a [workaround](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#IncludeCategories) for this file in clang-format). And in general, I think such problems are solved quite realistically if you want, I can try on the weekend.

> Do we hold off any clang-format feature until the fragile code is resolved? 

I think we should merge PR like that, then do a big reformat and add validation to the CI and script for developers.
Then we should try to fix the problem with the order of includes and do a big reformat for them too.

> How do we prevent having more fragile code like this in the future?

Regarding the non-occurrence of such code in the future, this is usually done by reviewing and building the code on jenkins. And as with any other code.
Of course, in theory, there is a small chance that someone wrote code that depends on the order of inclusion, and then someone else ran into a problem, but it seems too small?

aMahanna: My previous reformat & build is outdated (78bb583, f4c3b3d, 8410968, f0f6424), so I decided to do it again today locally. I was able to successfully build, as long as the following conditions are met after the reformat:
1. Revert the `#include` sorting in `arangod/Replication2/ReplicatedLog/ILogParticipant.cpp` (related to `velocypack-aliases`)
2. Revert the `#include` sorting in `arangod/V8Server/v8-user-functions.cpp` (related to `velocypack-aliases`)
3. Revert the `#include` sorting in `arangod/V8Server/v8-replicated-logs.cpp` (related to `velocypack-aliases`)
4. Add `#include <string>` to `lib/Utilities/IsArangoExecutable.h`

As for the solution to `velocypack-aliases`, I am not sure if removing the file would be easy, since it is currently being referenced in a significant amount of files (let me know what you think).

As for the next steps for clang-format, I am thinking of doing the following in order:
1. Figure out what to do with the `velocypack-aliases` issue
2. Set `SortIncludes` back to `true`
3. Merge the new `.clang-format` file into `devel` (this PR)
4. Open & merge new PR to perform the big reformat of entire codebase (will most likely be done on Dec 11)
5. Merge https://github.com/arangodb/arangodb/pull/15083 to introduce CI validation & pre-commit githook for developers. 

I want to figure out the `velocypack-aliases` issue as Step 1 since it prevents a successful build. 

Let me know what you think about all this,

CCing @joerg84 and @gnusi

goedderz: I'd have nothing against removing `velocypack-aliases`, but as you said, it'd be a lot of work.

But I think it should be usable with sorting. One problem I can think of is that it might still be used in some *header* files, in which it can break in other files including those headers. If that's the case, removing it from those headers should solve the problem. Or is the reason here a different one?

aMahanna: `arangod/Replication2/ReplicatedLog/ILogParticipant.cpp`

Before sort:
```
#include "Replication2/LoggerContext.h"
#include "Replication2/ReplicatedLog/LogCore.h"
#include "Replication2/ReplicatedLog/LogStatus.h"
#include "Replication2/ReplicatedLog/ReplicatedLogMetrics.h"
#include "Metrics/Gauge.h"

#include <Basics/Exceptions.h>
#include <Basics/StaticStrings.h>
#include <Basics/application-exit.h>
#include <velocypack/velocypack-aliases.h>
```

After sort:
```
#include <Basics/Exceptions.h>
#include <Basics/StaticStrings.h>
#include <Basics/application-exit.h>
#include <velocypack/velocypack-aliases.h>

#include "Metrics/Gauge.h"
#include "Replication2/LoggerContext.h"
#include "Replication2/ReplicatedLog/LogCore.h"
#include "Replication2/ReplicatedLog/LogStatus.h"
#include "Replication2/ReplicatedLog/ReplicatedLogMetrics.h"
```

None of these header files are directly referencing `velocypack-aliases`, but `LogStatus.h` is referencing the following velocypack elements:
```
#include <velocypack/Builder.h>
#include <velocypack/Slice.h>
```
From what I've seen the only thing in common that `Build.h`, `Slice.h` and `velocypack-aliases` have is that they all include `velocypack/velocypack-common.h`. But I don't see anything alarming with that. 

@goedderz, do you notice anything unusual from this diff?

MBkkt: @aMahanna I did mean you can use a workaround like this
```
IncludeCategories:
  - Regex:           '<velocypack/velocypack-aliases.h>'
    Priority:        4
  - Regex:           '^<ext/.*\.h>'
    Priority:        2
  - Regex:           '^<.*\.h>'
    Priority:        1
  - Regex:           '^<.*'
    Priority:        2
  - Regex:           '.*'
    Priority:        3
```

aMahanna: I reformated & rebuilt ArangoDB with that new `.clang-format` style change, but encountered new errors, regarding `arangod/V8Server/v8-user-functions.cpp`. The re-ordering of the `#include` lines shifted `#include "v8-user-functions.h"` to be **before** others:
image

This caused multiple errors of the `use of undeclared identifier` nature. 

The developer of this file must have intentionally placed the corresponding header file of `v8-user-functions.cpp` to be after all of these required elements to avoid any `undeclared identifier` errors.

Even if we fixed this one via another `IncludeCategories` addition, there is no guarantee that we won't see the issue again for another file. This could fill up our `IncludeCategories` with multiple regex matches to various header file names, which I am not a fan of having, as it would become difficult to maintain.

MBkkt: Just use -priotlrity, like
1 => -5
2 => -4
3 => -3
4 => -2

Then main category, that have 0 priority by default will be last

MBkkt: Btw I thought in your case these includes (V8/*) should be in header file, no?

aMahanna: > Btw I thought in your case these includes (V8/*) should be in header file, no?
This definitely should be the case here, but I think this change is beyond the scope of what we want to have here

aMahanna: > Just use negative priority, like
1 => -4
2 => -3
3 => -2
4 => -1

Sorry could you give an example of what that would look like in `.clang-format`? I’m guessing what you mean is to modify the priorities of the existing `IncludeCategories`, but not 100% certain

MBkkt:
```
IncludeCategories:
  - Regex:           '<velocypack/velocypack-aliases.h>'
    Priority:        -1
  - Regex:           '^<ext/.*\.h>'
    Priority:        -3
  - Regex:           '^<.*\.h>'
    Priority:        -4
  - Regex:           '^<.*'
    Priority:        -3
  - Regex:           '.*'
    Priority:        -2
```

MBkkt (commented):
@aMahanna Do you have any questions? Or why working on this PR was stopped?

aMahanna:
> @aMahanna Do you have any questions? Or why working on this PR was stopped?

@MBkkt we are just waiting for a review by @gnusi 

The solution you proposed [here](https://github.com/arangodb/arangodb/pull/15131#discussion_r765235263) works specifically for the `velocypack-aliases.h` sort issue, but we then encounter other non-velocypack-aliases sort issues that are affected by the displacement of different header files. 

As per our slack conversation we will be disabling `SortIncludes` (for now, but we can come back to it later) and await @gnusi's second review of the `.clang-format` file, to resolve any deprecated values (for example `Standard: Cpp11`).

goedderz:
Sorry for the belated reply.

> @goedderz, do you notice anything unusual from this diff?

Yes. `velocypack-aliases.h` must be included *after all other* velocypack headers. Including `Builder.h` and `Slice.h` via `LogStatus.h` after `velocypack-aliases.h` can change things. In particular, the `VPackBuilder` and `VPackSlice` aliases will no longer be available.

If `ILogParticipant.cpp` uses any `VPack*` alias, it must include the corresponding file(s) for itself. This should solve the problem.

I.e. I think following if the following rules are adhered, everything should be fine:

* `velocypack-aliases` is only included in `.cpp` files, never in header files
* all other (directly included) velocypack headers are ordered before `velocypack-aliases`
* all velocypack headers needed in that file are included directly, not just indirectly (iwyu)

P.S.: I've just seen a [recent commit (1cd26fd)](https://github.com/arangodb/arangodb/commit/1cd26fdf93bb144c4040cfabb14d5885a4d5397c) removed the aliases from `velocypack-aliases.h`. So if you pull in devel, the problem will probably just go away, and you can forget everything I said above.

.clang-format
joerg84: Changed SortIncludes to false

joerg84 (approved):

PR #15132, opened 2021-11-16
[CINFRA-146] Implement multi-insert into coordinator

apetenchea:
### Scope & Purpose

The multi-insert method, available at `_api/log/<log-id>/multi-insert` can be used to add multiple replicated log entries at once. For now, only DB servers have proper support for it. This pull requests adds the method to Coordinators, similar to how the regular insert was implemented. The coordinator forwards the request to the log leader.

### Testing & Verification


arangod/Replication2/Methods.cpp
```diff
@@ -320,7 +320,48 @@ struct ReplicatedLogMethodsCoordinator final
...
+    VPackBuilder builder{};
+    {
+      VPackArrayBuilder arrayBuilder{&builder};
+      while (const auto payload = iter.next()) {
+        if (payload.has_value()) {
+          builder.add(payload->slice());
+        } else {
+          builder.add(VPackSlice::emptyObjectSlice());
+        }
+        ++payloadSize;
```
maierlars, line 335: Here is a nice trick: `std::optional<T>` is convertible to `bool` and it converts to `true` if and only if `has_value()` returns true. That means that your `else` branch is never executed. The iterator signals that it has reached the end by returning a `std::nullopt`. Thus we can write
```
while (auto const payload = iter.next()) {
    builder.add(payload->slice()); // safe because otherwise we would have left the loop
    payloadSize++;
}
```

arangod/Replication2/Methods.cpp
```diff
@@ -320,7 +320,48 @@ struct ReplicatedLogMethodsCoordinator final
...
+          auto result = resp.slice().get("result");
+          auto waitResult = result.get("result");
+
+          auto quorum = std::make_shared<replication2::replicated_log::QuorumData const>(
+              waitResult.get("quorum"));
+          auto commitIndex = waitResult.get("commitIndex").extract<LogIndex>();
+
+          std::vector<LogIndex> indexes;
+          indexes.reserve(payloadSize);
+          auto indexIter = velocypack::ArrayIterator(result.get("indexes"));
```
maierlars, line 356: I just noticed that the http-api returns something different than the js api does (this applies to the single insert function as well). In JS on the server (i.e. using V8) we have only the `index` or `indexes` field, in Http and therefore client-js there are information about the quorum. We should include the quorum information in V8 as well.

arangod/V8Server/v8-replicated-logs.cpp
```diff
@@ -205,6 +206,52 @@ static void JS_Insert(v8::FunctionCallbackInfo<v8::Value> const& args) {
...
+    VPackObjectBuilder ob(&response);
+    {
+      VPackArrayBuilder ab{&response, "indexes"};
+      for (auto const logIndex : result.first) {
+        response.add(VPackValue(logIndex));
+      }
+    }
+    response.add(VPackValue("result"));
+    result.second.toVelocyPack(response);
+  }
```
maierlars, line 249: Add information about quorum (`result.second`) here as well.

arangod/Replication2/ReplicatedLog/ReplicatedLogIterator.h
```diff
@@ -37,6 +37,9 @@
 #pragma warning(pop)
 #endif
 
+#include <velocypack/Iterator.h>
+#include <velocypack/velocypack-aliases.h>
```
maierlars, line 41: We should try to not include the `velocypack-aliases.h` in a header file. Additionally we can simply forward declare `arangodb::velocypack::ArrayIterator` in the header and get rid of the include as well.

maierlars (changes requested):
I am very happy with the code overall! :)

arangod/Replication2/ReplicatedLog/ReplicatedLogIterator.h
```diff
@@ -125,4 +128,20 @@ class InMemoryLogIterator : public TypedLogIterator<InMemoryLogEntry> {
...
+    }
+
+    return std::nullopt;
+  }
+
+  explicit LogPayloadIterator(VPackSlice slice) : _iter(slice) {}
+
+ private:
+  VPackArrayIterator _iter;
+};
```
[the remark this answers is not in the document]
apetenchea: What do you say about adding a static method `createIteratorFromVPackArray` to the `LogPayload` struct? Would you find a separate `Utilities` header to be more useful?

arangod/Replication2/ReplicatedLog/ReplicatedLogIterator.h
apetenchea: Solving this by moving the new class to `Utilities.h`.

tests/js/common/shell/shell-replicated-logs-cluster.js
```diff
@@ -97,6 +97,21 @@ function ReplicatedLogsWriteSuite () {
...
+    testMultiInsert : function() {
+      let log = db._replicatedLog(logId);
+      let index = 0;
+      for (let i = 0; i < 100; i += 3) {
+        let indexes = log.multiInsert([{foo0: i}, {foo1: i + 1}, {foo2: i + 2}]).indexes;
+        assertTrue(Array.isArray(indexes));
+        assertTrue(indexes.length === 3);
+        let next = indexes[indexes.length - 1];
+        assertTrue(next > index);
+        index = next;
```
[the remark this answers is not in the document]
apetenchea: Added a few more checks.

maierlars (approved):

PR #15138, opened 2021-11-17
improvements for getting in sync

jsteemann:
### Scope & Purpose

Improvements for getting in sync
* Reuse the same DatabaseTailingSyncer instance for both the catchup (soft lock) and the exclusive (hard lock) phase.
  This allows the syncer to use the same syncer id with which it registers on the leader.
* Also make the leader return its current sequence number in "lastLogTick" attribute when the exclusive lock is acquired. This allows the follower to limit its WAL tailing efforts to only up to the sequence number that was returned by the leader.

The changes in this PR have already been made in 3.8 and above.

### Testing & Verification

- [x] This change is already covered by existing tests, such as *replication_sync, replication_ongoing, chaos*.

### DIFF
--- a/arangod/Cluster/SynchronizeShard.cpp
+++ b/arangod/Cluster/SynchronizeShard.cpp
@@ -40,7 +40,6 @@
 #include "Replication/DatabaseInitialSyncer.h"
 #include "Replication/DatabaseReplicationApplier.h"
 #include "Replication/DatabaseTailingSyncer.h"
-#include "Replication/GlobalInitialSyncer.h"
 #include "Replication/GlobalReplicationApplier.h"
 #include "Replication/ReplicationApplierConfiguration.h"
 #include "Replication/ReplicationFeature.h"
@@ -94,7 +93,8 @@ using namespace std::chrono;
 // first:
 //  - wait until leader has created shard
 //  - lookup local shard
-//  - call `replicationSynchronize`
+//  - call `replicationSynchronizeCatchup` (WAL tailing, configure leaderId
+//  - do WAL tailing with read-lock (configure leaderId
 //  - call `catchupWithReadLock`
 //  - call `catchupWithExclusiveLock`
 // replicationSynchronize:
@@ -118,7 +118,8 @@ using namespace std::chrono;
 
 SynchronizeShard::SynchronizeShard(MaintenanceFeature& feature, ActionDescription const& desc)
   : ActionBase(feature, desc),
-    _followingTermId(0) {
+    _followingTermId(0),
+    _tailingUpperBoundTick(0) {
   std::stringstream error;
 
   if (!desc.has(COLLECTION)) {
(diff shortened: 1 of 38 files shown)


arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1176,11 +1032,21 @@ ResultT<TRI_voc_tick_t> SynchronizeShard::catchupWithReadLock(
...
+  TRI_ASSERT(tailingSyncer != nullptr);
+
   uint64_t lockJobId = 0;
   LOG_TOPIC("da129", DEBUG, Logger::MAINTENANCE)
       << "synchronizeOneShard: startReadLockOnLeader: " << ep << ":" << database
       << ":" << collection.name();
+  
+  // we should not yet have an upper bound for WAL tailing.
+  // the next call to startReadLockOnLeader may set it if the leader already implements
+  // it (ArangoDB 3.8.3 and higher)
```
neunhoef, line 1047: This should say "ArangoDB 3.7.16 and higher" for 3.7.

arangod/Replication/DatabaseTailingSyncer.cpp
```diff
@@ -81,21 +83,175 @@ Result DatabaseTailingSyncer::saveApplierState() {
...
+  TRI_ASSERT(!_unregisteredFromLeader);
+
+  try {
+    // always start from _initialTick
+    _initialTick = fromTick;
+    Result res = syncCollectionCatchupInternal(collectionName, timeout, false, until, didTimeout, context);
+    if (res.fail()) {
+      // if we failed, we can already unregister ourselves on the leader, so that
+      // we don't block WAL pruning
+      unregisterFromLeader();
```
neunhoef, line 100: This call and the one below in line 106 make the corresponding calls in `SynchronizeShard.cpp` unnecessary. On the other hand, they also do not hurt, since we have `_unregisteredFromLeader`, so we might as well leave it as it is.

neunhoef (approved):
LGTM. One comment is to slightly change a comment (optional!), the other is just a comment and requires no action.

arangod/Cluster/SynchronizeShard.cpp
KVS85: Rephrased as "ArangoDB 3.7.16, 3.8.3 and higher".

jsteemann: Ok, comment updated.

arangod/Replication/DatabaseTailingSyncer.cpp
jsteemann: Yes, it won't hurt because of `_unregisteredFromLeader`. I suggest to leave it as is as well.

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -874,59 +749,45 @@ bool SynchronizeShard::first() {
...
+        builder.add(COLLECTION, VPackValue(shard));
+        builder.add(LEADER_ID, VPackValue(leader));
+        builder.add("requestTimeout", VPackValue(600.0));
+        builder.add("connectTimeout", VPackValue(30.0));
       }
+      
+      ReplicationApplierConfiguration configuration =
+          ReplicationApplierConfiguration::fromVelocyPack(feature().server(), builder.slice(), database);
+      // will throw if invalid
+      configuration.validate();
```
mpoeter, line 774: Can we please simply initialize the `ReplicationApplierConfiguration` directly instead of using the velocypack in between? Then we can probably also get rid of this `builder` variable (a few lines down we create yet another `builder` anyway which would shadow this one).
```suggestion
    {
      // build configuration for WAL tailing
      ReplicationApplierConfiguration configuration(feature().server());
      configuration._endpoint = ep;
      configuration._database = database;
      configuration._requestTimeout = 600.0;
      configuration._connectTimeout= 30.0;
      // will throw if invalid
      configuration.validate();
```
As far as I can see `ReplicationApplierConfiguration::fromVelocyPack` does not handle the `COLLECTION` and `LEADER_ID` fields. Is this an oversight?

arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -1020,12 +873,12 @@ bool SynchronizeShard::first() {
...
           sy, LAST_LOG_TICK, 0);
-      VPackBuilder builder;
 
       ResultT<TRI_voc_tick_t> tickResult =
-        catchupWithReadLock(ep, database, *collection, clientId, shard,
-                            leader, lastTick, builder);
+          catchupWithReadLock(ep, database, *collection, clientId, shard,
+                              leader, lastTick, tailingSyncer);
       if (!tickResult.ok()) {
+        tailingSyncer->unregisterFromLeader();
```
mpoeter, line 881: Perhaps this should be done using a scopeGuard since there are two other places where we call `unregisterFromLeader`?

arangod/Cluster/SynchronizeShard.h
```diff
@@ -31,12 +31,15 @@
 #include "VocBase/voc-types.h"
 
 #include <chrono>
+#include <functional>
```
mpoeter, line 34: I see no reason why we need to include this here.
```suggestion
```

mpoeter (changes requested):

arangod/Cluster/SynchronizeShard.cpp
jsteemann: To be honest: I don't like doing this in this PR right now. This is the 3.7 version of a series of changes that we merged for devel, 3.9, 3.8 already. Changing it only in this last open PR wouldn't be ideal, because changes would be lost from 3.8 onwards. So we could open yet more PRs to do these changes in the 3 other branches, which will just delay things further. 
I think code style improvements should be made in devel first (and probably only there) and are not necessarily be done in stable branches.
The collection and leader id attributes will be set somewhere else when calling `syncCollectionCatchup` and `syncCollectionFinalize`. Probably not ideal as it is now, but has been like this for years, and I am not happy with making this change in 3.7 now and then having to re-test everything in all versions.

arangod/Cluster/SynchronizeShard.cpp
jsteemann: I agree this would work too and will look "nicer". But IMHO it is a "nice-to-have" change which will not affect the behavior.
So I suggest doing all the cleanup work you suggested just for devel, in a separate PR.

mpoeter (approved):

PR #15145, opened 2021-11-17
fix non-deterministic dropping of followers during tests

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15144

Fix non-deterministic dropping of followers during tests

This change increases the supervision grace period from 10 seconds to a higher value when running the `shell_client` tests. it seems that 10 seconds is not enough on overloaded servers, so that we randomly get failedLeader jobs during the tests, which can lead to spurious dropping of followers. this is a problem because some tests assume that during the tests no followers are getting dropped.

### Testing & Verification


neunhoef (approved):
LGTM

PR #15159, opened 2021-11-18
Add fmt-8.0.1 lib

markuspf:
### Scope & Purpose

Import `fmt-8.0.1` as this is the only way to get `std::format` c++20 functionality at the moment (yes, really).

@maierlars and the CINFRA team have an immediate use-case coming up.


maierlars:
Possible improvements:
- remove the docs directory to reduce repo size.
- maybe we can remove the tests as well?
- add the license to the `LINCESES-OTHER-COMPONENTS.md`
- we should add the library to `3rdParty/CMakeLists.txt`

maierlars (approved):

markuspf:
New PR will require introducing a submodule of some kind.

PR #15167, opened 2021-11-18
[CINFRA-141] Report commit failure participant

apetenchea:
### Scope & Purpose

Sometimes during commit index calculation, a quorum cannot be reached. This could happen if there were not enough eligible participants, or some forced participant(s) could not be included. The work we did here adds a little hint to `who` might be the one responsible for failing to reach a quorum. If there are multiple participants responsible for this, only one is reported.
We added serialization/de-serialization tests and `calculateCommitIndex` tests to check if the correct participant is reported.

### Testing & Verification

- [x] This change is already covered by existing tests, such as `CalcCommitIndexTest.cpp`.


tests/Replication2/ReplicatedLog/CalcCommitIndexTest.cpp
```diff
@@ -449,3 +449,51 @@ TEST_F(CalcCommitIndexTest, DISABLED_more_forced_than_quorum_size) {
...
+
+  auto [index, reason, quorum] =
+      algorithms::calculateCommitIndex(participants,
+                                       CalculateCommitIndexOptions{1, 1, 2},
+                                       LogIndex{1}, LogIndex{50});
+
+  EXPECT_TRUE(
+      std::holds_alternative<CommitFailReason::QuorumSizeNotReached>(reason.value));
+  EXPECT_EQ(std::get<CommitFailReason::QuorumSizeNotReached>(reason.value).who, "A");
+}
```
markuspf, line 499: Bonus challenge: can you devise a test like the above that fails on MSVC but passes on clang/gcc?

I think this is possible because of the different implementation of `std::nth_element` between them.

 This test passes by virtue of the `EXPECT_EQ` checking *the only possible* option for `who`.

markuspf (changes requested):
Other than my comment about the. tests this looks good to me

 Why wouldn't it, I came up with it together with you; probably best if @maierlars has a look as well ;)

arangod/Replication2/ReplicatedLog/LogCommon.h
```diff
@@ -354,16 +354,18 @@ struct CommitFailReason {
...
     void toVelocyPack(velocypack::Builder& builder) const;
+    ParticipantId who;
   };
   std::variant<NothingToCommit, QuorumSizeNotReached, ForcedParticipantNotInQuorum> value;
 
   static auto withNothingToCommit() noexcept -> CommitFailReason;
-  static auto withQuorumSizeNotReached() noexcept -> CommitFailReason;
-  static auto withForcedParticipantNotInQuorum() noexcept -> CommitFailReason;
+  static auto withQuorumSizeNotReached(ParticipantId const& who) noexcept -> CommitFailReason;
+  static auto withForcedParticipantNotInQuorum(ParticipantId const& who) noexcept -> CommitFailReason;
```
maierlars, line 367: I am completely fine with that method, just small thing: Those functions will certainly make a copy of `who`. Making a copy is not `noexcept`. (So this `noexcept` here is actually wrong) You could instead pass the string value. In that case the caller will make the copy for you and you just move your parameter into your member, which is noexcept. 

This has another advantage: if the caller doesn't need that string anymore, it is free to move the string into your function. With that signature above, there will always be a copy, even if the caller could move the string.

The third option would be to provide additional overloads that take a `&&` reference. Then only the overloads that accept `&&` would be noexcept. But this is very often a overkill.

arangod/Replication2/ReplicatedLog/LogCommon.cpp
```diff
@@ -500,6 +511,7 @@ auto replicated_log::to_string(CommitFailReason const& r) -> std::string {
 
 auto replicated_log::operator==(CommitFailReason const& left,
                                 CommitFailReason const& right) noexcept -> bool {
+  // Ignoring the `who` field so that we only compare the reason of failure.
   return left.value.index() == right.value.index();
```
maierlars, line 496: What is the implication of that? I don't remember what the `operator==` is used for, but it might be confusing it something is not updated, because the index did not change, although the `who` changed. I imagen a situation like: `Can not commit because of X` - you fix `X` - now you can not commit of `Y` but the message still say `X`.

tests/Replication2/ReplicatedLog/CalcCommitIndexTest.cpp
maierlars: If `"B"` is also a valid return value, then we should check for both.

maierlars (approved):

arangod/Replication2/ReplicatedLog/LogCommon.cpp
markuspf: We discussed this and my reasoning (which I don't claim is sound) was: 

The `who` is only informational, the error is that quorum size wasn't reached; (and I don't want to deal with `std::variant` right now)

It is certainly possible (and in view of what you say desirable) to compare `who`.

maierlars: Let me just mention, that `std::variant` forward comparison [operators](https://en.cppreference.com/w/cpp/utility/variant/operator_cmp). You only have to define an `operator==` for each type of the variant.

And then I think it could even be a space shit.

apetenchea: Declared `operator==` as `default` for each type of the variant and for `CommitFailReason`. The `who` member will be taken into account.

arangod/Replication2/ReplicatedLog/LogCommon.h
apetenchea: ParticipantId is passed by value.

tests/Replication2/ReplicatedLog/CalcCommitIndexTest.cpp
apetenchea: @markuspf added `who_quorum_size_not_reached_multiple` test.

apetenchea: @maierlars now we check for all valid participants.

markuspf (approved):

markuspf:
We will have to insist on jenkins passing though ;)

PR #15170, opened 2021-11-19
Fix missing rename of _thresholdNetworkBatchSize.

maierlars:
### Scope & Purpose

Fix missing rename of _thresholdNetworkBatchSize.


apetenchea (approved):
LGTM

PR #15176, opened 2021-11-22
Feature/simplify read own writes tagging

jsteemann:
### Scope & Purpose

* Simplify the tagging of EnumerateCollectionNodes and IndexNodes with the
  "read-own-writes" flag. Previously the tagging only happened after all query
  optimizations were completed, making the tag unavailable to the optimizer.
  Now the tag is set early on, so it is accessible by the query optimizer.

This change is required to complete https://arangodb.atlassian.net/browse/BTS-624

#### Backports:

- [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/15178
- [x] Backport for 3.8: https://github.com/arangodb/arangodb/pull/15181

### Testing & Verification

### DIFF
--- a/arangod/Aql/ConditionFinder.cpp
+++ b/arangod/Aql/ConditionFinder.cpp
@@ -139,7 +139,7 @@ bool ConditionFinder::before(ExecutionNode* en) {
           // index cannot be used for filtering, but only for sorting
           // remove the condition now
           TRI_ASSERT(sorting);
-          condition.reset(new Condition(_plan->getAst()));
+          condition = std::make_unique<Condition>(_plan->getAst());
           condition->normalize(_plan);
         }
 
@@ -162,6 +162,10 @@ bool ConditionFinder::before(ExecutionNode* en) {
               // if the enumerate collection node had the counting flag
               // set, we can copy it over to the index node as well
               idx->copyCountFlag(node);
+              if (node->canReadOwnWrites() == ReadOwnWrites::yes) {
+                // copy over the read-own-writes flag from EnumerateCollectionNode to IndexNode
+                idx->setCanReadOwnWrites(ReadOwnWrites::yes);
+              }
               return idx;
             })
         );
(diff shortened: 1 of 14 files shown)


arangod/Aql/ConditionFinder.cpp
```diff
@@ -162,6 +162,10 @@ bool ConditionFinder::before(ExecutionNode* en) {
               // if the enumerate collection node had the counting flag
               // set, we can copy it over to the index node as well
               idx->copyCountFlag(node);
+              if (node->canReadOwnWrites() == ReadOwnWrites::yes) {
+                // copy over the read-own-writes flag from EnumerateCollectionNode to IndexNode
+                idx->setCanReadOwnWrites(ReadOwnWrites::yes);
+              }
```
mpoeter, line 168: Can't we simplify like this?
```suggestion
              // copy over the read-own-writes flag from EnumerateCollectionNode to IndexNode
              idx->setCanReadOwnWrites(node->canReadOwnWrites());
```

tests/js/server/aql/aql-read-own-writes.js
```diff
@@ -0,0 +1,170 @@
...
+      db._create(cn);
+    },
+
+    tearDown : function () {
+      db._drop(cn);
+    },
+    
+    testNonCollectionForLoop : function () {
+      const query = `FOR i IN 1..3 RETURN i`;
+      let nodes = AQL_EXPLAIN(query).plan.nodes;
```
cpjulia, line 52: could these nodes variables be const because you initialize them, but don't edit them later on?

cpjulia (approved):
LGTM

tests/js/server/aql/aql-read-own-writes.js
jsteemann: Yes, they could be const. But it won't make a difference here.

mpoeter (approved):
LGTM, but perhaps we should add another test case with an UPSERT inside a subquery.

arangod/Aql/ConditionFinder.cpp
jsteemann: Yes!

tests/js/server/aql/aql-read-own-writes.js
jsteemann: Changed anyway.

jsteemann:
> LGTM, but perhaps we should add another test case with an UPSERT inside a subquery.

done

PR #15206, opened 2021-11-24
avoid duplication of arangod process options

jsteemann:
### Scope & Purpose

Avoid duplication of arangod process startup options in `scripts/startLocalCluster.sh`. The options were previously specified twice for agents, and twice for DB servers/coordinators. Now they are specified only once for agents and once for DB servers/coordinators.

In addition, also log an annoying maintenance log message later during the startup process, so that it actually is logged to the configured logfile and not stdout.

### Testing & Verification


neunhoef (approved):
LGTM

PR #15207, opened 2021-11-24
Reduce memory usage of Merkle trees

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/14549

Reduce memory usage for in-memory revision trees. Previously, a revision tree instance for a non-empty collection/shard was using 4 MB of memory when uncompressed. Trees that were unused for a while were compressed on the fly to use less memory, and later uncompressed again when needed.
Now the uncompressed in-memory version of the revision tree will dynamically allocate memory as needed. This allows the initial version of the trees to get away with just 64 KB of memory. Memory usage will grow lazily when more parts of the trees get populated. The compression of unused in-memory tree data is still in place.

### Testing & Verification

- [x] This change is already covered by existing tests, such as *gtest, recovery, ...*.


neunhoef (approved):
LGTM. This is a relatively straightforward backport after all. It does not change the on disk format, so no downgrade problems within the 3.8 series are to be expected. The code as presented has been for a while in 3.9 and `devel`. I validated the fact that there is no more memory spike on restart or during `SynchronizeShard` operations in the presence of many shards. This is good and should solve the problem.

mpoeter (approved):

PR #15208, opened 2021-11-25
[BTS-565] 3.8 Timeout without deadlock

maierlars:
### Scope & Purpose

This increases the test timeout to 30s to detect a deadlock.


jsteemann (approved):
LGTM

PR #15210, opened 2021-11-25
[BTS-565] 3.9 Timeout without Deadlock

maierlars:
### Scope & Purpose

This increases the test timeout to 30s to detect a deadlock.


KVS85 (approved):
LGTM.

PR #15212, opened 2021-11-25
[3.9] Disable COMDAT folding.

mpoeter:
### Scope & Purpose

No CHANGELOG entry since this is not user facing and does not affect a released version.

- [x] :hankey: Bugfix

### Testing & Verification


jsteemann (approved):
LGTM!

Dronplane:
MSDN says `By default, /OPT:ICF is enabled by the linker unless /OPT:NOICF or /DEBUG is specified. To override this default and prevent COMDATs from being folded in the program, specify /OPT:NOICF.` here https://docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-160
So I suggest instead of simply removing `/OPT:ICF` add `/OPT:NOICF`

jsteemann:
> MSDN says `By default, /OPT:ICF is enabled by the linker unless /OPT:NOICF or /DEBUG is specified. To override this default and prevent COMDATs from being folded in the program, specify /OPT:NOICF.` here [docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-160](https://docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-160) So I suggest instead of simply removing `/OPT:ICF` add `/OPT:NOICF`

According to @mpoeter the explicit _enabling_ of ICF reduced program size greatly, so it seems to have been opt-in rather than opt-out. I am not sure however it that statement referred to a debug build or a release build.

Dronplane:
For debug builds ICF is diabled (to make stack traces more readable). For non debug builds ICF is implicitly enabled.

mpoeter:
@Dronplane The `/DEBUG` linker option enables creation of debugging information for the executable. So this is not only for debug builds, but also release builds, since we want to have debug symbols for our release as well. But you are right that it would be better to disable it explicitly, and I also add a comment explaining the reasons for it.

PR #15213, opened 2021-11-25
Disable COMDAT folding.

mpoeter:
### Scope & Purpose

No CHANGELOG entry since this is not user facing and does not affect a released version.

- [x] :hankey: Bugfix

#### Backports:

- [x] Backport for 3.9: #15212

### Testing & Verification


jsteemann (approved):
LGTM!

PR #15215, opened 2021-11-25
Feature/debugger breakin on assertion

Dronplane:
### Scope & Purpose

Current crash handler implementation  calls just exit  on Windows in case of intentional crash procedure (such as assertion failure in the maintainer  mode). This leaves us without proper crash dump (as exit even with SIGABRT is  not recognized as abnormal program termination). To fix this this PR adds explicit triggering of the minidump creation.
Another option could be replacing `exit` call with `abort` call but that may trigger unexpected dialog windows shown by OS and  I suggest to leave `exit` for now.

Another point was to add `DebugBreak` call inside failed assertion handler, but that was dropped due to using process monitor in our CI runs. Process monitor is detected as debugger and just blocks the execution on worker node (as there is no one to actually do the debug there).

#### Backports:

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation


lib/Basics/CrashHandler.cpp
```diff
@@ -103,6 +105,8 @@ std::atomic<bool> killHard(false);
     // TerminateProcess is async, alright wait here for selfdestruct (we will never exit wait)
     WaitForSingleObject(hSelf, INFINITE);
   } else {
+    // exit will not trigger dump creation. So do this manually.
+    createMiniDump(nullptr);
```
mpoeter, line 109: This way we will not get a good callstack for the calling thread, but that is actually the callstack we are most interested in.
The problem is that the current thead (which contains the failed assertion) is also writing the minidump, so at the time we take the process snapshot for the dump the thread is in the middle of the dump routine and the callstack for that thread will simply contain a single frame for `NtGetThreadContext`.
To solve this we need to pass in an EXCEPTION_POINTER struct. This could be achieved like this:
```suggestion
    auto filter = [](EXCEPTION_POINTERS* pointers) {
      std::cout << pointers->ContextRecord->Rip << std::endl;
      return EXCEPTION_EXECUTE_HANDLER;
    };
    __try {
      __debugbreak();
    } __except(filter(GetExceptionInformation())) {}
```
This would have the nice side effect that during debugging we stop at the breakpoint and do not create a minidump since the debugger "swallows" the int3 exception.

mpoeter (changes requested):

lib/Basics/CrashHandler.cpp
Dronplane: Using a DebugBreak is not an option - this hangs our CI in case of assertion failure.
Regarding `NtGetThreadContext`  - test run with intentional assertion failure showed just normal stacks 🤔

KVS85 (approved):
LGTM.

mpoeter (approved):

PR #15219, opened 2021-11-25
Disable ICF explicitly.

mpoeter:
### Scope & Purpose

Follow up PR for #15213 to disable ICF explicitly and add a comment explaining why.

- [x] :hankey: Bugfix


Dronplane (approved):
LGTM

PR #15221, opened 2021-11-25
[3.8] Disable COMDAT folding.

mpoeter:
### Scope & Purpose

Backport for #15213 and #15219


jsteemann (approved):
LGTM!

PR #15229, opened 2021-11-26
fixed smart graph creation for pregel smart graph example

hkernbach:
### Scope & Purpose

This PR fixed the test: `tests/js/client/shell/shell-custom-pregel-timecritical.js`
which is executable via:
```
scripts/unittest shell_client --cluster true --test tests/js/client/shell/shell-custom-pregel-timecritical.js
```

- [x] :hankey: Bugfix (does not require CHANGELOG entry - as issue has not been released yet)

#### Backports:

- No backports needed. Devel related only.

### Testing & Verification


jsteemann (approved):
I can confirm that this PR fixes the problem!

PR #15240, opened 2021-11-29
Feature/delay intermediate commits during multi operations

jsteemann:
### Scope & Purpose

Do not execute intermediate commits in the middle of array/multi-document operations.
This change will delay any intermediate commit until the end of the array operation. This is necessary so that the synchronous replication can be carried out while the transaction's document keys are still fully locked on the leader and dictate the sequence of replicated events for any given key.

#### Backports:

- [x] Backport for 3.9: TO DO

### Testing & Verification

- [x] This change is already covered by existing tests, such as *shell_client*, *shell_server*.


arangod/Transaction/Methods.cpp
```diff
@@ -2827,6 +2861,12 @@ futures::Future<OperationResult> Methods::countInternal(std::string const& colle
   return futures::makeFuture(countLocal(collectionName, type, options));
 }
 
+// perform a (deferred) intermediate commit if required
+Result Methods::performIntermediateCommitIfRequired(DataSourceId collectionId) {
+  [[maybe_unused]] bool hasPerformedIntermediateCommit = false;
+  return _state->performIntermediateCommitIfRequired(collectionId, hasPerformedIntermediateCommit);
```
mpoeter, line 2867: I think I found no caller that actually uses the value stored in `hasPerformedIntermediateCommit`. Can we get rid for that parameter?

tests/js/client/shell/shell-transaction-intermediate-commit-babies-cluster.js
```diff
@@ -0,0 +1,306 @@
...
+      let didWork = false;
+      try {
+        db._query('FOR i IN 1..10000 INSERT { _key: CONCAT("test", i), value: ASSERT(i < 10000, "peng!") } IN ' + cn, {}, {intermediateCommitCount: 100});
+        didWork = true;
+      } catch (err) {
+      }
+      assertFalse(didWork);
+      
+      let droppedFollowersAfter = getMetric(leader, "arangodb_dropped_followers_total");
+      assertEqual(droppedFollowersBefore + 1, droppedFollowersAfter);
```
mpoeter, line 179: Why do we expect a dropped follower in this case?

tests/js/client/shell/shell-transaction-intermediate-commit-babies-cluster.js
```diff
@@ -0,0 +1,306 @@
...
+      assertEqual(intermediateCommitsBefore, intermediateCommitsAfter);
+        
+      assertEqual(1000, db._collection(cn).count());
+    },
+    
+    testDocumentsTransactionIntermediateCommitsSuppressed: function () {
+      let c = db._create(cn, { numberOfShards: 1, replicationFactor: 1 });
+      let shards = db._collection(cn).shards(true);
+      let shardId = Object.keys(shards)[0];
+      let leader = getEndpointById(shards[shardId][0]);
```
mpoeter, line 272: This initialization code is almost identical for all testcases, though some use rf=2 and some rf=1. I would prefer if we could split these tests into two separate testsuites and move the common initialization code into the respective setup functions.

mpoeter (approved):

tests/js/client/shell/shell-transaction-intermediate-commit-babies-cluster.js
jsteemann: Added a comment to the PR.

tests/js/client/shell/shell-transaction-intermediate-commit-babies-cluster.js
jsteemann: Actually the setups are slightly different. I thus created 2 separate testsuites out of it so it becomes a bit clearer.

arangod/Transaction/Methods.cpp
jsteemann: Good catch. Fixed!

mpoeter (approved):

PR #15249, opened 2021-11-30
Metrics refactoring

MBkkt:
### Scope & Purpose

In their current form, metrics are almost impossible to change. 
Since this leads to the recompilation of a very large amount of code.
This change is intended to correct this issue.

Also, for the convenience of naming, metrics have been moved to their own namespace.

So I almost didn't change the code of the metrics themselves.

But I want (In next PRs):
* remove `std::optional<std::reference_wrapper<SomeMetricType>>`,
which is no different from `T*` since the `operator->` that is used in our code is [`noexcept`](https://en.cppreference.com/w/cpp/utility/optional/operator*)
* remove the `shared_ptr` on metrics, which isn't really needed.
* use the help string from the YAML doc and remove it from DECLARE. Now the help string that we display may not equal the documentation help string
* add tags to HTTP metrics endpoint, see [design document](https://arangodb.atlassian.net/wiki/spaces/REED/pages/1641250933/ArangoSearch+metrics)

#### Backports:

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/816
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation


arangod/Cluster/ClusterFeature.h
```diff
@@ -98,11 +98,11 @@ class ClusterFeature : public application_features::ApplicationFeature {
...
-  
-  Counter& followersDroppedCounter() { return _followersDroppedCounter->get(); }
-  Counter& followersRefusedCounter() { return _followersRefusedCounter->get(); }
-  Counter& followersWrongChecksumCounter() { return _followersWrongChecksumCounter->get(); }
-  Counter& followersTotalRebuildCounter() { return _followersTotalRebuildCounter->get(); }
+
+  metric::Counter& followersDroppedCounter() { return _followersDroppedCounter->get(); }
+  metric::Counter& followersRefusedCounter() { return _followersRefusedCounter->get(); }
+  metric::Counter& followersWrongChecksumCounter() { return _followersWrongChecksumCounter->get(); }
+  metric::Counter& followersTotalRebuildCounter() { return _followersTotalRebuildCounter->get(); }
```
jsteemann, line 105: The PR description of this PR states:
> remove std::optional<std::reference_wrapper<SomeMetricType>>, which is no different from T* since the operator-> that is used in our code is noexcept

Is it intentional that options with refererence_wrappers are still used here, or should they be changed here as well?

There are also additional places in other files where reference wrappers are still used. I will not mention them separately.

MBkkt: > But I want:

What is written in these paragraphs isn't what I did in this PR.
This is what I will want to do. This is some motivation for this PR.

MBkkt: I don’t do it in this PR because it seems like it might cause discussion.
But this PR is big and difficult to maintain and review.
So I decided not to make it bigger.

MBkkt: I have updated the description to make it clearer.

jsteemann: Thanks, got it.

jsteemann:
Some suggestions if we are doing a large-scale refactoring of metrics anyway:
* I would prefer to rename the new namespace to `metrics` and not `metric`, but it is just a matter of taste
* we should move all metrics code out of the directory `arangod/RestServer/Metric` but into a new directory `arangod/Metrics`. The metrics have nothing to do with the RestServer, so I am wondering why they are in there right now.
* include directives are normally ordered alphabetically. The includes for metrics are now just "somewhere". This doesn't break anything, but is somewhat inconsistent.

arangod/RestServer/Metric/Batch.h
```diff
@@ -0,0 +1,64 @@
...
+template <typename T>
+class Batch final : public Metric {
+ public:
+  Batch(T&& metric, std::string_view name, std::string_view help, std::string_view labels)
+      : Metric{name, help, labels}, _metric{std::move(metric)} {}
+
+  [[nodiscard]] std::string_view type() const noexcept final {
+    return "untyped";
+  }
+  void toPrometheus(std::string& result, std::string_view globals, std::string_view) const final {
```
jsteemann, line 41: The `final` is not necessary here, as the class is already marked `final`. I suggest we add `override` here because otherwise compilers may warn about it being missing.
```suggestion
  void toPrometheus(std::string& result, std::string_view globals, std::string_view) const override {
```

arangod/RestServer/Metric/Batch.h
```diff
@@ -0,0 +1,64 @@
...
+    load().toPrometheus(result, globals, labels());
+  }
+  void toPrometheusBegin(std::string& result, std::string_view name) const final {
```
jsteemann, line 44:
```suggestion
  void toPrometheusBegin(std::string& result, std::string_view name) const override {
```

arangod/RestServer/Metric/Batch.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
```
jsteemann, line 1: By convention, we place the `#pragma once` below the copyright header. I would favor doing it in this PR too.

arangod/RestServer/Metric/BatchBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+#include <RestServer/Metric/Builder.h>
+
+namespace arangodb::metric {
+
+template <typename T>
+class BatchBuilder : public GenericBuilder<BatchBuilder<T>> {
+ public:
+  using MetricT = Batch<T>;
+
+  [[nodiscard]] std::string_view type() const noexcept final {
```
jsteemann, line 36: Should be marked as `override`
```suggestion
  [[nodiscard]] std::string_view type() const noexcept final {
```

arangod/RestServer/Metric/BatchBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+    return "untyped";
+  }
+  [[nodiscard]] std::shared_ptr<Metric> build() const final {
```
jsteemann, line 39: Should be marked as `override`.
```suggestion
  [[nodiscard]] std::shared_ptr<Metric> build() const final {
```

arangod/RestServer/Metric/Batch.h
MBkkt: What compiler says that `override` is needed when there is `final`? Can you give an example from [godbolt](https://godbolt.org/)?

arangod/RestServer/Metric/Batch.h
MBkkt: Ok

MBkkt:
@jsteemann 

> I would prefer to rename the new namespace to metrics and not metric, but it is just a matter of taste

I named it according to my taste, if there are any agreements, or if it is of principle, or for example, we want to rename the folder to Metrics. Then I can rename

> we should move all metrics code out of the directory arangod/RestServer/Metric but into a new directory arangod/Metrics. The metrics have nothing to do with the RestServer, so I am wondering why they are in there right now.

Ok

> include directives are normally ordered alphabetically. The includes for metrics are now just "somewhere". This doesn't break anything, but is somewhat inconsistent.

I think this should be done by clang-format, not a human. 
Unfortunately, I cannot apply clang-format to the files that I changed.

arangod/RestServer/Metric/Counter.cpp
```diff
@@ -0,0 +1,94 @@
...
+Counter& Counter::operator++() noexcept {
+  count();
+  return *this;
+}
+
+//Counter& Counter::operator++(int) noexcept {
+//  count(1);
+//  return *this;  // TODO(MBkkt) remove this function
+//}
+
```
jsteemann, line 85: Let's remove this.
```suggestion
//Counter& Counter::operator++(int) noexcept {
//  count(1);
//  return *this;  // TODO(MBkkt) remove this function
//}

```

arangod/RestServer/Metric/Metric.h
```diff
@@ -0,0 +1,61 @@
...
+  [[nodiscard]] std::string_view labels() const noexcept;
+
+  virtual std::string_view type() const noexcept = 0;
+  virtual void toPrometheus(std::string& result, std::string_view globals,
+                            std::string_view alternativeName) const = 0;
+  virtual void toPrometheusBegin(std::string& result, std::string_view name) const;
+  virtual ~Metric();
+
+ private:
+  std::string_view _name;
```
jsteemann, line 48: Do we currently have any metrics with dynamic names? If yes, then we will likely need an `std::string` here to store the name. At least I guess that dynamic names are generated at call sites that generate the `Metric` objects and probably go out of scope while the `Metric` object lives on afterwards.

arangod/RestServer/Metric/Batch.h
MBkkt: Motivation to use `final` instead of `override` then possible (from a semantic point of view, they mean the same thing, except for the case of the most basic `virtual` function, you can write `final` there, but you cannot `override`, but if you never write `final` together with `virtual` (this is pointless), this problem doesn't exist):
More information for the compiler, you are right that in this case, it is not so, but if at some time our `class` becomes not `final`, the method that was `final` in its meaning will cease to be so

arangod/RestServer/Metric/MetricKey.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace arangodb::metric {
+
+struct MetricKey {
+  std::string_view name;
+  std::string_view labels;
+
+  MetricKey();
+  MetricKey(std::string_view name);
+  MetricKey(std::string_view name, std::string_view labels);
```
jsteemann, line 37: All of these can be marked `noexcept`

arangod/RestServer/Metric/Metric.h
MBkkt: > Do we currently have any metrics with dynamic names?
No
Their name is `#name` in macros

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+  result.reserve(32768);
+
+  // QueryRegistryFeature
+  auto& q = server().getFeature<QueryRegistryFeature>();
+  q.updateMetrics();
+  {
+    bool changed = false;
+
+    std::lock_guard guard(_lock);
+    if (_globalLabels.find("shortname") == _globalLabels.end()) {
```
jsteemann, line 431: with c++20 it can be simplified to
```suggestion
    if (!_globalLabels.contains("shortname")) {
```

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+      std::string shortName = ServerState::instance()->getShortName();
+      // Very early after a server start it is possible that the
+      // short name is not yet known. This check here is to prevent
+      // that the label is permanently empty if metrics are requested
+      // too early.
+      if (!shortName.empty()) {
+        _globalLabels.try_emplace("shortname", shortName);
```
jsteemann, line 438:
```suggestion
        _globalLabels.try_emplace("shortname", std::move(shortName));
```

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+        changed = true;
+      }
+    }
+    if (_globalLabels.find("role") == _globalLabels.end() &&
```
jsteemann, line 442:
```suggestion
    if (!_globalLabels.contains("role") &&
```

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+        _globalLabelsStr += i.first + "=\"" + i.second + "\"";
+      }
+    }
+
+    std::string lastType{};
+    for (auto const& i : _registry) {
+      std::string name{i.second->name()};
+      std::string alternativeName;
+      if (!v2) {
+        if (v1suppressions.find(name) != v1suppressions.end()) {
```
jsteemann, line 466:
```suggestion
        if (v1suppressions.contains(name)) {
```

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+          } else if (labels == POOL_CLUSTERCOMM) {
+            name += "_ClusterComm";
+          } else {
+            // Avoid someone sneaking in an other connection
+            // pool without dedicated metric for v1
+            TRI_ASSERT(false);
+          }
+        }
+        alternativeName = name;
+      } else if (auto iter = v2suppressions.find(name); iter != v2suppressions.end()) {
```
jsteemann, line 492:
```suggestion
      } else if (v2suppressions.contains(name)) {
```

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+        // In v1 we do a name conversion. Note that we set
+        // alternativeName == name in the end, in v2 though,
+        // alternativeName is empty and no conversion happens.
+        auto it = nameVersionTable.find(name);
+        if (it != nameVersionTable.end()) {
+          name = it->second;
+        }
+        static std::string const ARANGODB_CONNECTION = "arangodb_connection_";
+        static std::string const POOL_AGENCYCOMM = "pool=\"AgencyComm\"";
+        static std::string const POOL_CLUSTERCOMM = "pool=\"ClusterComm\"";
```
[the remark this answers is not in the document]
MBkkt: I agree with your comments about this file. But are you sure that I should do it in this pr? 
Now I just renamed it, removed the unused code.

From the point of view of refactoring
For example here these constants can be `constexpr std::string_view`.
Or, for example, it seems rather strange to me that each metric builder (DECLARE ..) is a separate class, why not `using` with a compile-time name "string"? Because help "string" is dynamic (
Or, for example, MetricsKey isn't really needed, we can simply write a comparator for the set.
Well, other points that I wrote in descriptions

I don't know if it is worth changing in this PR? 
As I wrote, I tried to make a minimum of changes.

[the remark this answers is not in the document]
jsteemann: It's fine for me to not do the changes in this PR.
But if you are planning larger follow-up refactorings for the metrics after this PR is merged, I would appreciate if you could include the above suggestions there too. Thanks!

arangod/RestServer/Metric/Batch.h
jsteemann: Ok, it seems you are right. There doesn't need to be an extra `override` specifier for methods that are marked `final`.
So it should work fine as it is in the PR right now. I prefer the other variant (declaring just the class `final` but not all methods, and instead mark overriden methods as `override`), but that is just a matter of taste.
So let's leave it as it is.

arangod/RestServer/Metric/Metric.h
jsteemann: I was rather thinking about metrics that will be created only at runtime, e.g. metrics for certain views or collections (which could come and go while the server is running). If we don't have such metrics right now and all names are known at compile-time, then it is fine as it is.

apetenchea (commented):
I'm on the same page with Jan regarding his `#pragma once` remark, since all files begin with the heading comment it's better to keep it that way.
I tried to compiling your branch with `-DUSE_ENTERPRISE=On` and it seems there will be some changes needed in there as well.
Other than that, on behalf of @arangodb/team-cinfra, this looks fine.

MBkkt:
@apetenchea

> I tried to compiling your branch with -DUSE_ENTERPRISE=On and it seems there will be some changes needed in there as well.

In description:
> Enterprise PR: [Metrics refactoring enterprise#816](https://github.com/arangodb/enterprise/pull/816)

arangod/RestServer/Metric/LinScale.h
```diff
@@ -0,0 +1,72 @@
...
+template <typename T>
+struct LinScale final : public Scale<T> {
+ public:
+  using Value = T;
+  static constexpr ScaleType kScaleType = ScaleType::Linear;
+
+  LinScale(T const& low, T const& high, size_t n) : Scale<T>{low, high, n} {
+    this->_delim.resize(n - 1);
+    _div = (high - low) / (T)n;
+    if (_div <= 0) {
```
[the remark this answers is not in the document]
MBkkt: It's not my code, it's clone old metrics code. But you're right I will remove this

arangod/RestServer/Metric/Metric.h
MBkkt: > If we don't have such metrics right now and all names are known at compile-time, then it is fine as it is.

We have, arango search metrics, but they have only dynamic labels, not names. You cant create metric with dynamic name now

arangod/RestServer/Metric/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+        "arangodb_scheduler_jobs_submitted_total",
+        "arangodb_scheduler_jobs_done_total",
+    };
+  } catch (std::exception const& e) {
+    LOG_TOPIC("efd51", ERR, Logger::MEMORY)
+        << "failed to allocate and populate the metrics v1/v2 translation "
+           "table "
+        << e.what();
+    FATAL_ERROR_EXIT();
+  }
```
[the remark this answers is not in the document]
MBkkt: Let me do it separately later. 
I'm just afraid that some test depends on this code.

mpoeter (approved):
I really like these changes and the clean one-class-per-file approach!

apetenchea (approved):

arangod/Agency/AgencyComm.cpp
```diff
@@ -44,7 +44,8 @@
 #include "Logger/Logger.h"
 #include "Random/RandomGenerator.h"
 #include "Rest/GeneralRequest.h"
-#include "RestServer/MetricsFeature.h"
+#include <Metrics/Histogram.h>
+#include <Metrics/LogScale.h>
```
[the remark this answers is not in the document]
MBkkt: Can you explain why?
I see we use `#include " ... "` as well as `#include <...>`. 
We are finding in the paths passed with `-I` options. Then why would we find elsewhere before that?

Probably consistency is more important, and I will change it to `#include "..."`, but I would like to understand why such a choice was made.

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+        "arangodb_scheduler_jobs_submitted",
+        "arangodb_scheduler_jobs_done",
+    };
+
+    v1suppressions = {
+        "arangodb_scheduler_jobs_dequeued_total",
+        "arangodb_scheduler_jobs_submitted_total",
+        "arangodb_scheduler_jobs_done_total",
+    };
+  } catch (std::exception const& e) {
```
jsteemann, line 354: Still thinking the whole try..catch block is unnecessary here.
If any of the `emplace` operations fails, startup will be aborted properly.

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -0,0 +1,524 @@
...
+std::shared_ptr<Metric> MetricsFeature::doAdd(Builder& builder) {
+  auto metric = builder.build();
+  MetricKey key{metric->name(), metric->labels()};
+  bool success = false;
+  {
+    std::lock_guard guard(_lock);
+    success = _registry.try_emplace(key, metric).second;
+  }
+  if (!success) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL, builder.type().data() +
```
jsteemann, line 390: I don't think this is safe.
From https://en.cppreference.com/w/cpp/string/basic_string_view/data:
> Unlike std::basic_string::data() and string literals, data() may return a pointer to a buffer that is not null-terminated. Therefore it is typically a mistake to pass data() to a routine that takes just a const CharT* and expects a null-terminated string.

arangod/Metrics/Metric.h
```diff
@@ -0,0 +1,60 @@
...
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Kaveh Vahedipour
+////////////////////////////////////////////////////////////////////////////////
+#pragma once
+
+#include <RestServer/counter.h>
+
+#include <ostream>
```
jsteemann, line 27: I think we can even get away with the more lightweight iosfwd header here
```suggestion
#include <iosfwd>
```

tests/Metrics/MetricsFeatureTest.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
jsteemann, line 1: Let's add the pragma after the copyright info.

jsteemann (changes requested):

arangod/Metrics/MetricsFeature.cpp
MBkkt: But we know our `type()` string, its string literal, so everything fine

arangod/Metrics/MetricsFeature.cpp
MBkkt: I'm ready to change that, but in the next PR
https://github.com/arangodb/arangodb/pull/15249#discussion_r761027800

arangod/RestServer/Metric/MetricsFeature.cpp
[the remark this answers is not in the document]
jsteemann: I don't think that any test will depend on this code.

arangod/Metrics/MetricsFeature.cpp
jsteemann: Ok!

arangod/Metrics/MetricsFeature.cpp
jsteemann: Ok, it may work for now, but it is not necessarily future proof.
Can't we explicitly convert it into an `std::string(builder.type().data(), builder.type().size())` here? That will enable it to work with arbitrary types that will be added in the future.

MBkkt: call string ctor now

jsteemann (approved):
LGTM!

Issue #15084, opened 2021-11-09
ArangoDB 3.8 crashing on simple queries


chrishedwards:
## My Environment

* __ArangoDB Version__:        3.8.1 and 3.8.2
* __Storage Engine__:             RocksDB
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:   ArangoDB Starter
* __Configuration__:               
* __Infrastructure__:               own set up on laptop and desktop computer
* __Operating System__:        Windows 10 
* __Total RAM in your machine__:        16Gb laptop, and 32GB desktop
* __Disks in use__:        NVMe on both laptop and desktop machines
* __Used Package__:              just used through browser interface

## Component, Query & Data

__Affected feature__:
AQL query using web interface
I have avidly used ArangoDB since 2017 and have developed a model that works well for exploring longitudinal analysis of student study. However, since moving to RocksDB and 3.8, I have started to experience crashes and cannot work out why. I'm using the Community edition on my Windows 10 laptop with 16GB RAM. I don't think the computer is the problem as I also experience it on my desktop machine.

Simple queries that used to run without issue with no limits, now crash ArangoDB if I do not limit the number of results to around 10. For example, this query selects a number of students with LIMIT, collects their study attempts (vertices), just keeping minimal data to RETURN. The objects retained are pi, an 8 character student id; pres5, a year plus letter, e.g. 2012J; module, a string of between 4 and 7 characters. This query would run without issue on previous versions in MMfiles.

I hope you can recommend a solution.
Best regards
Chris

__AQL query (if applicable)__:
```
FOR id in temp2  //list of user ids 
  LIMIT 2000,100 //have tried various combinations of values within LIMIT 
  FOR v,e,p IN 1..10 OUTBOUND CONCAT("StartEnd/",id._key,"-start") Path  
    OPTIONS {"bfs":true, uniqueVertices :"global", ignoreErrors:true}    
    FILTER !IS_NULL(v.module) //removes the final vertex for each user, as always null    
    LET summary=KEEP(v,"pi","pres5", "module")   //retain 3 objects, id, presentation, module 
    RETURN summary

```
__AQL explain and/or profile (if applicable)__:
Query String (455 chars, cacheable: true):
 FOR id in temp2  //list of user ids 
   LIMIT 2000,100 //have tried combinations of values within LIMIT 
   FOR v,e,p IN 1..10 OUTBOUND CONCAT("StartEnd/",id._key,"-start") Path  
[...]

PR #15265, opened 2021-12-02
[3.8] Fixed a potential use-after-free on Windows.

mpoeter:
### Scope & Purpose

Fix GitHub issue #15084.

Enterprise companion: https://github.com/arangodb/enterprise/pull/820

Fixed a potential use-after-free on Windows for queries that used the NeighborsEnumerator (though other PathEnumerators might have been affected as well).

The reason was that in some cases the MSVC STL unordered_map/unordered_set calculates hashes in `clear`, but the stringHeap that stored the strings which were reference by the stringRefs in the unordered_set was cleared before the set.

#### Backports:

- [x] devel port: *#15266*
- [x] Backport for 3.9: *#15267*
- [x] Backport for 3.7: *#15268*

#### Related Information

- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/820
- [x] GitHub issue: #15084

### Testing & Verification


hkernbach (approved):

jsteemann (approved):
LGTM, but seems like some changes must be carried out for test classes derived from PathEnumerator that are used in `tests`.

arangod/Graph/WeightedEnumerator.cpp
```diff
@@ -57,14 +57,15 @@ WeightedEnumerator::WeightedEnumerator(Traverser* traverser, TraverserOptions* o
...
-
+void WeightedEnumerator::clear() {
   _schreier.clear();
   _schreierIndex = 0;
   _lastReturned = 0;
   _queue.clear();
+}
 
+void WeightedEnumerator::setStartVertex(arangodb::velocypack::StringRef startVertex) {
+  PathEnumerator::setStartVertex(startVertex);
```
Dronplane, line 68: As I see before - `setStartVertex` implied "clear" but here I don't see `clear` call like in other enumerators. Is this intended?

jsteemann, line 68:
```suggestion
  PathEnumerator::setStartVertex(startVertex);
  clear();
```

jsteemann: Yes, it's probably better to call `clear()` here as well!

PR #15270, opened 2021-12-02
do not copy generated log message string

jsteemann:
### Scope & Purpose

Do not copy generated log message string

When a log message is put together in an output stream, earlier versions of C++ had no means of accessing the data from the ostream in a read-only fashion without copying it. With c++20 it is now possible to get a read-only view on the ostream's contents, without having to copy it.
We can use this to avoid making a copy of the ostream content's for every log message 😊

### Testing & Verification


jsteemann:
Requires proper c++20 support from compilers (currently not available).

jsteemann:
Closing for now because of compiler support still lacking.

PR #15271, opened 2021-12-03
[ORCO-13] Fix potential read inconsistency for single document operations

mpoeter:
### Scope & Purpose

Fix potential read inconsistency for single document operations.

Wen reading a single document that is concurrently being updated or replaced, the read operation could erroneously return a "document not found" error although the document actually existed. This only happened for single document operations, i.e., no transactions or AQL queries.

#### Backports:

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/ORCO-13

### Testing & Verification


tests/js/common/shell/aql-index-usage.js
```diff
@@ -122,6 +124,54 @@ function IndexUsageSuite () {
       assertTrue(success > 0, success);
     },
 
+    
+    testPrimaryIndexLookupConsistency : function () {
+      debugSetFailAt("RocksDBCollection::read-delay");
```
jsteemann, line 129: Just wondering if this will work in the cluster case...
IMO the failure point will only be set on the coordinator here, but have no effect on a DB server. 
It should be fine in single server though.

PR #15287, opened 2021-12-06
[Search-183] Fix strange test failed

MBkkt:
### Scope & Purpose

This assertion checked that the `сleanup` after the second removes, which removes 100 times fewer documents than the first, works faster or the same. Apparently, sometimes this isn't true, locally it isn't reproduced for me. So I just removed this statement

### Testing & Verification


jsteemann (approved):
LGTM

PR #15288, opened 2021-12-06
Fix gcc11 compatibility issue.

mpoeter:
### Scope & Purpose

Fixes a compatibility issue when building with gcc11.

- [x] :hankey: Bugfix

### Testing & Verification


jsteemann (approved):
I can confirm this fixes the compilation issue with g++11.

PR #15291, opened 2021-12-06
[Search-183] Add tests

MBkkt:
### Testing & Verification

### DIFF
--- a/tests/IResearch/IResearchLinkTest.cpp
+++ b/tests/IResearch/IResearchLinkTest.cpp
@@ -2224,7 +2224,7 @@ class IResearchLinkMetricsTest : public IResearchLinkTest {
     return NAN;
   }
 
-  void remove(uint64_t begin, uint64_t end) {
+  double remove(uint64_t begin, uint64_t end) {
     auto* l = getLink();
     arangodb::transaction::Methods trx(arangodb::transaction::StandaloneContext::Create(_vocbase),
                                        kEmpty, kEmpty, kEmpty,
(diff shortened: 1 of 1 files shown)


tests/IResearch/IResearchLinkTest.cpp
```diff
@@ -2581,7 +2589,7 @@ TEST_F(IResearchLinkMetricsTest, LinkAndMetics) {
     std::string actual;
     getPrometheusStr(actual);
 
-    EXPECT_EQ(actual, expected);
+    EXPECT_TRUE(actual == expected);
```
Dronplane: Why not leave EXPEC_EQ  here ?

tests/IResearch/IResearchLinkTest.cpp
```diff
@@ -2537,7 +2545,7 @@ TEST_F(IResearchLinkMetricsTest, LinkAndMetics) {
...
```
Dronplane: Why not leave EXPEC_EQ  here ?

Dronplane (changes requested):
LGTM in general, but couple of places looks suspicious

tests/IResearch/IResearchLinkTest.cpp
MBkkt: Just typo

tests/IResearch/IResearchLinkTest.cpp
MBkkt: same here

Dronplane (approved):
LGTM

PR #15296, opened 2021-12-06
Remove unnecessary optional

MBkkt:
### Scope & Purpose

Remove `std::optional<std::reference_wrapper<SomeMetricType>>`, which is no different from `SomeMetricType*` since the `operator->` and `operator*` that is used in our code is [`noexcept`](https://en.cppreference.com/w/cpp/utility/optional/operator*)

### Testing & Verification


arangod/Cluster/ClusterFeature.h
```diff
@@ -183,10 +195,10 @@ class ClusterFeature : public application_features::ApplicationFeature {
   ServerState::RoleEnum _requestedRole = ServerState::RoleEnum::ROLE_UNDEFINED;
   metrics::Histogram<metrics::LogScale<uint64_t>>& _agency_comm_request_time_ms;
   std::unique_ptr<network::ConnectionPool> _asyncAgencyCommPool;
-  std::optional<std::reference_wrapper<metrics::Counter>> _followersDroppedCounter;
-  std::optional<std::reference_wrapper<metrics::Counter>> _followersRefusedCounter;
-  std::optional<std::reference_wrapper<metrics::Counter>> _followersWrongChecksumCounter;
-  std::optional<std::reference_wrapper<metrics::Counter>> _followersTotalRebuildCounter;
+  metrics::Counter* _followersDroppedCounter = nullptr;
```
MBkkt, line 198: Maybe we need an abstraction like `RawPtr<T>`? 
A class that encapsulates access to a pointer and adds an assertion that it isn't `nullptr` and adds default ctor.

mpoeter:
I would prefer to make them real references (like all the other metrics) and instead move the metrics into dedicated classes.

alexbakharew:
We can implement some "DummyMetric" which is inherited from "Metric" class. It can replace usage of pointers . Instead of it, we will store reference to metric. Also it avoids checks and dereferences of nullptr metrics

MBkkt:
> We can implement some "DummyMetric" which is inherited from "Metric" class. It can replace usage of pointers . Instead of it, we will store reference to metric. Also it avoids checks and dereferences of nullptr metrics

But in this case we should make all metrics member functions are virtual or we can't empty call for dummy (not an empty call even worse)

So I thought if or assert is better solution

In general, I think we can solve this problem another way: we can store metrics in their classes(opposite now we store all metrics in MetricsFeature) and register them(or some abstraction for them) in the MetricsFeature

If I understood correctly, this is exactly the approach that @mpoeter suggests, and I like it more

Dronplane:
I like the @mpoeter  idea  to move metrics to the  dedicated classes. But what about moving the metrics to the "respective" classes? E.g. each object stores internally shared_ptr to the metric, created by some "Factory". And then registers the "Metric" in the metrics feature. The Metrics Feature stores only weak_ptrs. So after the "owner" instance is deleted  - weak_ptr is expired and MetricFeature eventually removes it.

mpoeter:
@Dronplane I like the idea of moving ownership to the respective classes, but I wouldn't store weak_ptrs in the MetricFeatures. Instead, I think the Metric instance should just deregister itself from the MetricFeature upon destruction.

Dronplane:
@mpoeter weak_ptr is exactly the robust way of "deregistration".   But of course some "deregister" noexcept call in dtor is also fine. But than we must somehow ensure that at the moment of destructing  the last "Metric" the Metric Feature is still alive.

arangod/Cluster/ClusterFeature.h
jsteemann: Sounds good!

jsteemann (approved):
LGTM!
I think your idea of adding a `RawPtr<T>` wrapper is really useful, and would be good to simplify the access to optional metrics even more. This can be done in a separate PR however.

goedderz:
> I think your idea of adding a `RawPtr<T>` wrapper is really useful, and would be good to simplify the access to optional metrics even more. This can be done in a separate PR however.

Seconded. For the record, that sounds a lot like [gsl::not_null](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-nullptr).

Dronplane (approved):
LGTM

alexbakharew (approved):

mpoeter:
TBH I am not convinced that a `RawPtr<T>` brings much to the table. If I have a pointer that must not be null, why not use a reference? The `gsl::not_null` referenced by @goedderz is a bit different since it should also work for `unique_ptr` and `shared_ptr`, which often cannot be replaced by references.
Instead of making each individual metric optional, I would prefer to group them together in a class and use an optional member of that class.

> @mpoeter weak_ptr is exactly the robust way of "deregistration". [...] we must somehow ensure that at the moment of destructing the last "Metric" the Metric Feature is still alive.
@Dronplane I would argue that his depends. IMHO in this scenario `weak_ptr` is not a good solution, partly exactly because of the ownership/lifetime management problem. It would work with the current metrics, but I have some ideas on how to extend the current metrics implementation (I have briefly discussed this with @MBkkt already), which would make this more complicated, which is why I would prefer an explicit deregistration that complete removes the metric. But these are details for a separate discussion...

MBkkt:
@goedderz 

I thought `gsl::not_null` just `std::reference_wrapper`?

@mpoeter 

> If I have a pointer that must not be null, why not use a reference?

For lazy initialization

For example kotlin have not null types, but sometimes, you want not null but lazy initialization type, they have special keyword for that.

In C++ I suppose RawPtr that initializing by nullptr, but not null in any other expression most convenient way for that.

Another example sometimes we want optional pointer, and we can use nullptr as nullopt to avoid extra "overhead" for optional

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -55,308 +55,459 @@ MetricsFeature::MetricsFeature(application_features::ApplicationServer& server)
   setOptional(false);
   startsAfter<LoggerFeature>();
   startsBefore<application_features::GreetingsFeaturePhase>();
+  nameVersionTable = {
```
gnusi, line 58: @MBkkt I suggest we use `frozen::unordered_map` instead of `std::unordered_map`. I'd also move `nameVersionTable` into `cpp` completely

MBkkt: Done

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -143,71 +143,59 @@ struct TimeTracker {
...
   }
 
   ~TimeTracker() {
-    if (statistics._exportReadWriteMetrics) {
-      // metrics collection is not free. only do it if metrics are enabled
-      // unit is seconds here
-      TRI_ASSERT(histogram.has_value());
-      histogram->get().count(std::chrono::duration<float>(getTime() - start).count());
+    if (metrics) {
+      // metrics collection is not free. only do it if metrics are enabled unit is seconds here
```
jsteemann, line 159: I know the code comment was there before, but I had some trouble understanding it when reading it right now. So I suggest to slightly change the wording:
```suggestion
      // metrics gathering is not free. only do it if metrics are enabled. Unit is seconds here
```

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -143,71 +143,59 @@ struct TimeTracker {
   TimeTracker(TimeTracker const&) = delete;
   TimeTracker& operator=(TimeTracker const&) = delete;
 
-  explicit TimeTracker(arangodb::TransactionStatistics const& statistics, 
-                       std::optional<std::reference_wrapper<arangodb::metrics::Histogram<arangodb::metrics::LogScale<float>>>>& histogram) noexcept
-      : statistics(statistics),
-        histogram(histogram) {
-    if (statistics._exportReadWriteMetrics) {
+  using Metrics = arangodb::TransactionStatistics::ReadWriteMetrics;
+  using Count = void (*)(Metrics& metrics, float time);
```
jsteemann, line 147: As we are calling the `Count` function from destructors, should we change the function signature to include a `noexcept` specifier?

jsteemann (approved):
LGTM!

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: Done

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: You're right. Done

mpoeter:
>>    If I have a pointer that must not be null, why not use a reference?

> For lazy initialization

I understand that. But since we have _multiple_ optional metrics that are either _all null_ or _all non-null_ it would IMHO be cleaner to put these metrics in a class and store an (optional) pointer to that class. E.g., instead of
```
  metrics::Histogram<metrics::LogScale<uint64_t>>* _phase1_runtime_msec = nullptr;
  metrics::Histogram<metrics::LogScale<uint64_t>>* _phase2_runtime_msec = nullptr;
  metrics::Histogram<metrics::LogScale<uint64_t>>* _agency_sync_total_runtime_msec = nullptr;

  metrics::Counter* _phase1_accum_runtime_msec = nullptr;
  metrics::Counter* _phase2_accum_runtime_msec = nullptr;
  metrics::Counter* _agency_sync_total_accum_runtime_msec = nullptr;

  metrics::Counter* _action_duplicated_counter = nullptr;
  metrics::Counter* _action_registered_counter = nullptr;
  metrics::Counter* _action_done_counter = nullptr;
  ```
  I would prefer
  ```
  struct Metrics {  
    metrics::Histogram<metrics::LogScale<uint64_t>>& _phase1_runtime_msec;
    metrics::Histogram<metrics::LogScale<uint64_t>>& _phase2_runtime_msec;
    metrics::Histogram<metrics::LogScale<uint64_t>>& _agency_sync_total_runtime_msec;

    metrics::Counter& _phase1_accum_runtime_msec;
    metrics::Counter& _phase2_accum_runtime_msec;
    metrics::Counter& _agency_sync_total_accum_runtime_msec;

    metrics::Counter& _action_duplicated_counter;
    metrics::Counter& _action_registered_counter;
    metrics::Counter& _action_done_counter;
  };
  std::unique_ptr<Metrics> _metrics;
  ```

MBkkt:
@mpoeter What do you think about merge it? In my opinion, `std::unique_ptr` is overkill (unnecessary allocations, etc.). We can use `std::optional` for the same purpose, but I don't like it too much. 
I know for sure that the arangodb code is full of other pointers to metrics. And this PR is intended primarily to get rid of `std::optional<std::reference_wrapper>`, which are strange and don't allow to include only forward declaration of metrics.

gnusi:
@mpoeter I tend to agree that `std::unique_ptr<...>` is over engineering in this particular case. Main purpose of this PR was to eliminate obviously strange things like `std::optional<std::reference_wrapper<...>>` and such. IMO the overall metrics design is strange and ownership logic in particular.

mpoeter:
True, you could also use a `std::optional` instead of a `std::unique_ptr`. I don't think it's over engineering, but clean design (-> Single Responsibility Principle). But you are right that this doesn't have to be done in this PR, especially if we want to further cleanup the metrics design. So I am ok with merging it as it is.

PR #15304, opened 2021-12-08
[BTS-611] [3.9] Bugfixes in constrained sort and late materialize

goedderz:
Backport of #15263


markuspf (approved):
LGTM as per review of #15263

PR #15311, opened 2021-12-09
[CINFRA-12] Adding test for lci communication.

maierlars:
### Scope & Purpose

Adding test for lci communication.


apetenchea (approved):
Looks good! Regarding code, the only suggestion is to use the method already provided by `FakeFollower.currentRequest`, at lines 271, 289, 312, 330 and 338. I double checked the test, it is correct.

PR #15315, opened 2021-12-10
added various startup options for throttle configuration

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15317
Docs PR: https://github.com/arangodb/docs/pull/841

Added startup options to adjust previously hard-coded parameters for the RocksDB throttle:
- `--rocksdb.throttle-frequency`: frequency for write-throttle calculations (in milliseconds, default value is 60000, i.e. 60 seconds).
- `--rocksdb.throttle-slots`: number of historic measures to use for throttle value calculation (default value is 63).
- `--rocksdb.throttle-scaling-factor`: adaptiveness scaling factor for write-throttle calculations (default value is 17). There is normally no need to change this value.
- `--rocksdb.throttle-slow-down-writes-trigger`: number of level 0 files whose payload is not considered in throttle calculations when penalizing the presence of L0 files. There is normally no need to change this value.
- `--rocksdb.throttle-max-write-rate`: maximum write rate enforced by the throttle (in bytes per second, default value is 0, meaning "unlimited"). The actual write rate will be the minimum of this value and the value the throttle calculation produces.

All these options will only have an effect if `--rocksdb.throttle` is enabled (which is the default). The configuration options introduced here use the previously hard-coded settings as their default values, so there should not be a change in behavior if the options are not adjusted.

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/841

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,23 @@
 v3.8.5 (XXXX-XX-XX)
 -------------------
 
+* Added startup options to adjust previously hard-coded parameters for the 
+  RocksDB throttle:
+  - `--rocksdb.throttle-frequency`: frequency for write-throttle calculations
+    (in milliseconds, default value is 60000, i.e. 60 seconds)
+  - `--rocksdb.throttle-slots`: number of historic measures to use for throttle
+    value calculation (default value is 63)
+  - `--rocksdb.throttle-scaling-factor`: adaptiveness scaling factor for write-
+    throttle calculations (default value is 17)
+  - `--rocksdb.throttle-max-write-rate`: maximum write rate enforced by the
+    throttle (in bytes per second, default value is 0, meaning "unlimited").
+    The actual write rate will be the minimum of this value and the value the
+    throttle calculation produces.
+  All these options will only have an effect if `--rocksdb.throttle` is enabled
+  (which is the default). The configuration options introduced here use the
+  previously hard-coded settings as their default values, so there should not be
+  a change in behavior if the options are not adjusted.
+
 * Improve visibility in case of potential data corruption between primary index
   and actual document store in documents column family.
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,7 +16,7 @@ v3.8.5 (XXXX-XX-XX)
     throttle calculation produces.
   - `--rocksdb.throttle-slow-down-writes-trigger`: number of level 0 files whose
     payload is not considered in throttle calculations when penalizing the 
-    presence of L0 files.
+    presence of L0 files. There is normally no need to change this value.
   All these options will only have an effect if `--rocksdb.throttle` is enabled
   (which is the default). The configuration options introduced here use the
   previously hard-coded settings as their default values, so there should not be
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,15 +4,19 @@ v3.8.5 (XXXX-XX-XX)
 * Added startup options to adjust previously hard-coded parameters for the 
   RocksDB throttle:
   - `--rocksdb.throttle-frequency`: frequency for write-throttle calculations
-    (in milliseconds, default value is 60000, i.e. 60 seconds)
+    (in milliseconds, default value is 60000, i.e. 60 seconds).
   - `--rocksdb.throttle-slots`: number of historic measures to use for throttle
-    value calculation (default value is 63)
+    value calculation (default value is 63).
   - `--rocksdb.throttle-scaling-factor`: adaptiveness scaling factor for write-
-    throttle calculations (default value is 17)
+    throttle calculations (default value is 17). There is normally no need to
+    change this value.
   - `--rocksdb.throttle-max-write-rate`: maximum write rate enforced by the
     throttle (in bytes per second, default value is 0, meaning "unlimited").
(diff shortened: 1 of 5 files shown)


arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -577,6 +577,23 @@ class RocksDBEngine final : public StorageEngine {
   std::deque<RocksDBKeyBounds> _pendingCompactions;
   /// @brief number of currently running compaction jobs
   size_t _runningCompactions;
+
+  // frequency for throttle in milliseconds
+  uint64_t _throttleFrequency = 60 * 1000;
```
kvahed, line 582:
```suggestion
  uint64_t _throttleFrequency = 5 * 1000; 
```

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -577,6 +577,23 @@ class RocksDBEngine final : public StorageEngine {
...
+
+  // frequency for throttle in milliseconds
+  uint64_t _throttleFrequency = 60 * 1000; 
+
+  // number of historic data slots to keep around for throttle
+  uint64_t _throttleSlots = 63;
+  // adaptiveness factor for throttle
+  // following is a heuristic value, determined by trial and error.
+  //  its job is slow down the rate of change in the current throttle.
+  //  do not want sudden changes in one or two intervals to swing
```
kvahed, line 589:
```suggestion
  //  we do not want sudden changes in one or two intervals to swing
```

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -577,6 +577,23 @@ class RocksDBEngine final : public StorageEngine {
...
+  //  the throttle value wildly.  Goal is a nice, even throttle value.
```
kvahed, line 590:
```suggestion
  //  the throttle value wildly.  the goal is a nice, even throttle value.
```

kvahed (changes requested):
Maybe take over the proven values?

CHANGELOG
```diff
@@ -1,6 +1,23 @@
 v3.8.5 (XXXX-XX-XX)
 -------------------
 
+* Added startup options to adjust previously hard-coded parameters for the 
+  RocksDB throttle:
+  - `--rocksdb.throttle-frequency`: frequency for write-throttle calculations
+    (in milliseconds, default value is 60000, i.e. 60 seconds)
```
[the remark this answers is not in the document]
neunhoef: Do we really want to change this always? I think for local SSDs the current values are probably sufficient. For cloud disk volumes, the 5 seconds are probably better...

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -376,6 +376,31 @@ void RocksDBEngine::collectOptions(std::shared_ptr<options::ProgramOptions> opti
...
+                     new UInt64Parameter(&_throttleScalingFactor),
+                     arangodb::options::makeFlags(arangodb::options::Flags::DefaultNoComponents, arangodb::options::Flags::OnDBServer, arangodb::options::Flags::OnSingle, arangodb::options::Flags::Hidden))
+                     .setIntroducedIn(30805);
+  
+  options->addOption("--rocksdb.throttle-max-write-rate", "maximum write rate enforced by throttle (in bytes per second, 0 = unlimited)",
+                     new UInt64Parameter(&_throttleMaxWriteRate),
+                     arangodb::options::makeFlags(arangodb::options::Flags::DefaultNoComponents, arangodb::options::Flags::OnDBServer, arangodb::options::Flags::OnSingle, arangodb::options::Flags::Hidden))
+                     .setIntroducedIn(30805);
+  
+  options->addOption("--rocksdb.throttle-slow-down-writes-tigger", "number of level 0 files that are considered as 'too many pending'",
```
neunhoef, line 400:
```suggestion
  options->addOption("--rocksdb.throttle-slow-down-writes-trigger", "number of level 0 files that are considered as 'too many pending'",
```

neunhoef (approved):
LGTM. One typo in option name.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Fixed

CHANGELOG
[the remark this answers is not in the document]
jsteemann: I don't object to changing the default value, we should do it in a separate PR after this one.

arangod/RocksDBEngine/RocksDBEngine.h
jsteemann: I don't object to changing the default value, we should do it in a separate PR after this one.

arangod/RocksDBEngine/RocksDBEngine.h
jsteemann: Fixed

arangod/RocksDBEngine/RocksDBEngine.h
jsteemann: Fixed

kvahed (approved):
LGTM

PR #15321, opened 2021-12-10
added various startup options for throttle configuration

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15317
Docs PR: https://github.com/arangodb/docs/pull/841

Added startup options to adjust previously hard-coded parameters for the RocksDB throttle:
- `--rocksdb.throttle-frequency`: frequency for write-throttle calculations (in milliseconds, default value is 60000, i.e. 60 seconds).
- `--rocksdb.throttle-slots`: number of historic measures to use for throttle value calculation (default value is 63).
- `--rocksdb.throttle-scaling-factor`: adaptiveness scaling factor for write-throttle calculations (default value is 17). There is normally no need to change this value.
- `--rocksdb.throttle-slow-down-writes-trigger`: number of level 0 files whose payload is not considered in throttle calculations when penalizing the presence of L0 files. There is normally no need to change this value.
- `--rocksdb.throttle-max-write-rate`: maximum write rate enforced by the throttle (in bytes per second, default value is 0, meaning "unlimited"). The actual write rate will be the minimum of this value and the value the throttle calculation produces.

All these options will only have an effect if `--rocksdb.throttle` is enabled (which is the default). The configuration options introduced here use the previously hard-coded settings as their default values, so there should not be a change in behavior if the options are not adjusted.

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/841

### Testing & Verification

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,29 @@
+v3.8.4.1 (XXXX-XX-XX)
+---------------------
+
+* Added startup options to adjust previously hard-coded parameters for the 
+  RocksDB throttle:
+  - `--rocksdb.throttle-frequency`: frequency for write-throttle calculations
+    (in milliseconds, default value is 60000, i.e. 60 seconds).
+  - `--rocksdb.throttle-slots`: number of historic measures to use for throttle
+    value calculation (default value is 63).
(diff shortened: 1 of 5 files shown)


neunhoef (approved):
LgTM

PR #15324, opened 2021-12-10
[ORCO-17] Feature/arangoexport add query bindvars

cpjulia:
### Scope & Purpose

Added flag `--custom-query-bindvars` to arangoexport and changed the flag names `--query` to `--custom-query` and `--query-max-runtime` to `--custom-query-max-runtime` to be like in the other client-tools.

#### Backports:

No backports required.

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/842
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/ORCO-17 Obs.: don't know if it goes here
- [ ] Design document: 

### Testing & Verification


client-tools/Export/ExportFeature.cpp
```diff
@@ -98,11 +98,23 @@ void ExportFeature::collectOptions(std::shared_ptr<options::ProgramOptions> opti
...
+  options
+      ->addOption("--custom-query", "AQL query to run", new StringParameter(&_customQuery))
+      .setIntroducedIn(31000);
+  options->addOldOption("--query-max-runtime", "custom-query-max-runtime");
+  options
+      ->addOption(
+          "--custom-query-max-runtime",
+          "runtime threshold for AQL queries (in seconds, 0 = no limit)",
+          new DoubleParameter(&_customQueryMaxRuntime))
+      .setIntroducedIn(31000);
```
jsteemann, line 111: I am undecided whether we should the change the version number here. It is true that the option didn't exist before under this name, but it existed before under the old name. Not sure what our convention for this is...
@Simran-B @ansoboleva : any idea?

js/client/modules/@arangodb/testsuites/export.js
```diff
@@ -209,324 +191,432 @@ function exportTest (options) {
...
-      DOMParser.parseFromString(filesContent);
-      results[testName] = {
-        failed: 0,
-        status: true
-      };
-
-      if (xmlErrors !== null) {
+    {
+      let args = Object.assign({}, commonArgValues);
+      args['compress-output'] = 'true';
```
jsteemann, line 334: This is now missing
```
 args['type'] = 'xgmml';
```

js/client/modules/@arangodb/testsuites/export.js
```diff
@@ -121,82 +121,64 @@ function exportTest (options) {
...
+    }
+
+    const commonArgValues = {
       'configuration': fs.join(pu.CONFIG_DIR, 'arangoexport.conf'),
       'server.username': options.username,
       'server.password': options.password,
       'server.endpoint': instanceInfo.endpoint,
       'server.database': name,
-      'collection': 'UnitTestsExport',
       'type': 'json',
```
jsteemann, line 133: It would be good if we didn't set the `type` globally, but rather set it individually for each test case.

client-tools/Export/ExportFeature.cpp
ansoboleva: @jsteemann Since the options existed before and this here is only a name change, I would lean towards keeping the version number as it was. Otherwise, it might give a false assumption.

jsteemann: @ansoboleva : fine for me.
@cpjulia : can you please change the version number back?

cpjulia: Addressed

js/client/modules/@arangodb/testsuites/export.js
cpjulia: Will address in latest commit with suggestion below of type not being a commong arg for all tests

client-tools/Export/ExportFeature.cpp
```diff
@@ -98,11 +98,23 @@ void ExportFeature::collectOptions(std::shared_ptr<options::ProgramOptions> opti
...
 
-  options->addOption("--query", "AQL query to run", new StringParameter(&_query));
-  
-  options->addOption("--query-max-runtime", "runtime threshold for AQL queries (in seconds, 0 = no limit)", 
-                     new DoubleParameter(&_queryMaxRuntime))
-                     .setIntroducedIn(30800);
+  options->addOldOption("--query", "custom-query");
+  options
+      ->addOption("--custom-query", "AQL query to run", new StringParameter(&_customQuery))
+      .setIntroducedIn(30800);
```
jsteemann, line 104: I don't think that `--query` (or `--custom-query`) was introduced with 3.8.0. The previous version of the code didn't have a `.setIntroducedIn` attribute for this startup option.

cpjulia: Addressed

jsteemann (approved):
LGTM!

PR #15331, opened 2021-12-13
cancel ongoing RocksDB compactions on shutdown

jsteemann:
### Scope & Purpose

Cancel ongoing RocksDB compactions on server shutdown.
This is an attempt to fix hangs in threads that are waiting in RocksDB compactions during the server shutdown.

#### Backports:

- [x] Backport for 3.9: TO BE DEFINED

### Testing & Verification

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -139,7 +139,7 @@ std::string const RocksDBEngine::EngineName("rocksdb");
 std::string const RocksDBEngine::FeatureName("RocksDBEngine");
 
 // global flag to cancel all compactions. will be flipped to true on shutdown
-std::atomic<bool> cancelCompactions{false};
+static std::atomic<bool> cancelCompactions{false};
 
 // minimum value for --rocksdb.sync-interval (in ms)
 // a value of 0 however means turning off the syncing altogether!
@@ -1023,8 +1023,10 @@ void RocksDBEngine::beginShutdown() {
     _replicationManager->beginShutdown();
   }
 
-  // from now on, all started compactions will be canceled
-  ::cancelCompactions.store(true, std::memory_order_relaxed);
+  // from now on, all started compactions can be canceled.
+  // note that this is only a best-effort hint to RocksDB and
+  // may not be followed immediately.
+  ::cancelCompactions.store(true, std::memory_order_release);
 }
 
 void RocksDBEngine::stop() {
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -138,6 +138,9 @@ DECLARE_COUNTER(arangodb_revision_tree_resurrections_total, "Number of revision 
 std::string const RocksDBEngine::EngineName("rocksdb");
 std::string const RocksDBEngine::FeatureName("RocksDBEngine");
 
+// global flag to cancel all compactions. will be flipped to true on shutdown
+std::atomic<bool> cancelCompactions{false};
+
 // minimum value for --rocksdb.sync-interval (in ms)
 // a value of 0 however means turning off the syncing altogether!
 static constexpr uint64_t minSyncInterval = 5;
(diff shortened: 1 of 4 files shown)


arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -1020,6 +1023,8 @@ void RocksDBEngine::beginShutdown() {
     _replicationManager->beginShutdown();
   }
 
+  // from now on, all started compactions will be canceled
+  ::cancelCompactions.store(true, std::memory_order_relaxed);
```
MBkkt, line 1027: I think we want `release` because for read we use `acquire`

MBkkt, line 1027: Is it ok if other threads see some effects of `beginShutdown` early than they see `cancelCompactions` is `true`?
If not I think u should use `release`, overwise please add a comment.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -138,6 +138,9 @@ DECLARE_COUNTER(arangodb_revision_tree_resurrections_total, "Number of revision
 std::string const RocksDBEngine::EngineName("rocksdb");
 std::string const RocksDBEngine::FeatureName("RocksDBEngine");
 
+// global flag to cancel all compactions. will be flipped to true on shutdown
+std::atomic<bool> cancelCompactions{false};
```
MBkkt, line 142: Can be static

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: I think this `canceled` flag is just a hint that allows RocksDB to abort a running compaction somewhere in the middle. It is not really time-critical, because it is only a hint and is not even guaranteed to be honored.
But I will change the memory order here.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Yes. But it is already in the anynomous namespace and in the .cpp file... I think declaring it static in addition does not provide any benefit here, or does it?

arangod/RocksDBEngine/RocksDBEngine.cpp
MBkkt: I make mistake, I mean `later` not `early`. Question not about _speed_, but about the correctness

arangod/RocksDBEngine/RocksDBEngine.cpp
MBkkt: > anynomous namespace

should be I suppose

But I don't see this

MBkkt: > I think declaring it static in addition does not provide any benefit here, or does it?

If we already have anynomous namespace I suppose no benefits

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Once we are in shutdown mode, it is fine for this flag to be picked up by anything in RocksDB. We basically want to cancel pending compactions eventually, so they don't block shutdown.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Awww, my bad. The definition was in `namespace arangodb`, not the anonymous namespace.
But anyway, I changed it to be static in my last commit, so we should be fine now.

MBkkt (approved):
LGTM!

mpoeter (commented):
What about this compaction code?
https://github.com/arangodb/arangodb/blob/efd40cffde0b9e126ba99c39748e36c918f5f338/arangod/RocksDBEngine/RocksDBCommon.cpp#L244-L249

arangod/RocksDBEngine/RocksDBEngine.cpp
mpoeter: Since RocksDB actually uses `memory_order_acquire` to read the value, it probably makes sense to use `memory_order_release` for the store, although I doubt that we need to establish any happens-before relation via this variable...

mpoeter (approved):

PR #15336, opened 2021-12-13
Add Precision to ROUND AQL method

solisoft:
### Scope & Purpose

Add Precision to AQL Method

```
RETURN ROUND(1.12345) // 1
RETURN ROUND(1.12345,1) // 1.1
RETURN ROUND(1.12345,2) // 1.12
RETURN ROUND(1.12345,3) // 1.123
RETURN ROUND(1.12345,4) // 1.1235
RETURN ROUND(1.12345,5) // 1.12345
```

#### Backports:

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document: 

### Testing & Verification

Link to Jenkins PR run:

### Documentation

### DIFF
--- a/arangod/Aql/AqlFunctionFeature.cpp
+++ b/arangod/Aql/AqlFunctionFeature.cpp
@@ -216,7 +216,7 @@ void AqlFunctionFeature::addNumericFunctions() {
   // numeric functions
   add({"FLOOR", ".", flags, &Functions::Floor});
   add({"CEIL", ".", flags, &Functions::Ceil});
-  add({"ROUND", ".", flags, &Functions::Round});
+  add({"ROUND", ".|.,.", flags, &Functions::Round});
   add({"ABS", ".", flags, &Functions::Abs});
   add({"SQRT", ".", flags, &Functions::Sqrt});
   add({"POW", ".,.", flags, &Functions::Pow});
--- a/arangod/Aql/Functions.cpp
+++ b/arangod/Aql/Functions.cpp
@@ -6770,11 +6770,13 @@ AqlValue Functions::Matches(ExpressionContext* expressionContext, AstNode const&
 AqlValue Functions::Round(ExpressionContext*, AstNode const&,
                           VPackFunctionParameters const& parameters) {
   AqlValue const& value = extractFunctionParameterValue(parameters, 0);
+  AqlValue const& precision = extractFunctionParameterValue(parameters, 1);
 
   double input = value.toDouble();
+  double exp = std::pow(10, precision.toDouble() >= 0 ? precision.toDouble() : 0);
 
   // Rounds down for < x.4999 and up for > x.50000
-  return ::numberValue(std::floor(input + 0.5), true);
+  return ::numberValue(std::floor((input * exp) + 0.5) / exp, true);
 }
 
 /// @brief function ABS
--- a/tests/js/server/aql/aql-functions-numeric.js
+++ b/tests/js/server/aql/aql-functions-numeric.js
@@ -3953,6 +3953,19 @@ function ahuacatlNumericFunctionsTestSuite () {
 /// @brief test round function
 ////////////////////////////////////////////////////////////////////////////////
 
+    testRoundWithPrecision : function () {
+      var expected = [ -100.1, -3.1, -2.1, 1.1, 2.1, 100.1 ];
+      var actual = getQueryResults("FOR r IN [ -100.123, -3.123, -2.123, 1.123, 2.123, 100.123 ] return ROUND(r, 1)");
+      assertEqual(expected, actual);
+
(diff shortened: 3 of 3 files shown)


jsteemann:
@solisoft : thanks a lot for your contribution!
I'd like to work on integrating your patch, but for legal reasons we need you to fill out and sign our Contributor License Agreement (CLA). We are using a Apache 2 CLA for ArangoDB, which can be found here: https://www.arangodb.com/documents/cla.pdf. You can scan and email the CLA PDF file to cla@arangodb.com or send it via fax to +49-221-2722999-88.
Sorry for the trouble, but I am not allowed to accept contributions without having the CLA in place.
Thanks!

goedderz (commented):
Hi @solisoft,

thank you for taking the time to open the PR!

However, I am apprehensive of this change, due to numerical problems.

While rounding a floating point value to an integer works fine, a floating point value cannot in general be rounded to a given _decimal_ precision.

Which means, rounding `12.9` to `13` works fine. But rounding `1.29` to `1.3` does not, because `1.3` is not representable as a finite binary fraction. While this may seem to work, it's only because the value is usually rounded again to a certain decimal fraction precision when it is being printed. A small example: https://godbolt.org/z/TGqcYqdve

This is of course not due to your implementation, but due to the fact that `double`s are used to represent fractions.

I fear that providing such a function will lead to more problems for users than it solves.

solisoft:
The main purpose of this change is to allow rounding values for displaying it ... By exemple I have a lot of charts to create and datasets are numbers like 13.19182726 ... and I only need the rounded value for display ... I understand the issue and I'll think about a fix.

jsteemann:
@goedderz : while it is true that in our case we cannot represent certain numbers exactly in a fixed-width floating-point value, rounding a value to a _near_ representable value may still be useful.
For example, rounding the value of `1.29` to the nearest representable value for `1.3` (e.g. `1.3000000000000000444089209850062616169452667236328125`) may be more desirable than going with 1.29. Because the `1.3xxx` value may well be truncated for displaying later, so it will likely come out as `1.3`.

If the issue is explained in the docs, I think that some rounding functionality is still a useful thing to add.

However, we will only be able to integrate the PR with a signed CLA.

solisoft:
just sent the CLA doc to cla@arangodb.com

jsteemann:
@solisoft : I will move this forward during the next few days, ensuring review, testing and such. As I will not be working tomorrow and the day after, it may be delayed until some point next week. 
But I will take all the blame if this does not make progress until mid next week :sweat_smile:.

arangod/Aql/AqlFunctionFeature.cpp
```diff
@@ -216,7 +216,7 @@ void AqlFunctionFeature::addNumericFunctions() {
   // numeric functions
   add({"FLOOR", ".", flags, &Functions::Floor});
   add({"CEIL", ".", flags, &Functions::Ceil});
-  add({"ROUND", ".", flags, &Functions::Round});
+  add({"ROUND", ".|.,.", flags, &Functions::Round});
```
Simran-B, line 219: AFAIK, `|` isn't a logical OR but a separator like `,` that also indicates the beginning of optional arguments.

```suggestion
  add({"ROUND", ".|.", flags, &Functions::Round});
```

Simran-B (commented):
Might be worth to check what values JavaScript's toFixed() method produces.

arangod/Aql/AqlFunctionFeature.cpp
jsteemann: @Simran-B : true, the signature should be changed according to your suggestion.

jsteemann:
@solisoft: after reading the suggestion from @goedderz, I think it could make sense to turn this function from `ROUND(number, precision) -> number` into `TO_FIXED(number, precision) -> string`. WDYT?

solisoft:
@jsteemann good idea !

solisoft:
Oh ok good to know

arangod/Aql/AqlFunctionFeature.cpp
solisoft: Ok good to know... Make sense now 😃

jsteemann:
Functionality continued via PR https://github.com/arangodb/arangodb/pull/15414.
That is a separate PR because unfortunately our CI infrastructure doesn't support pulling from different origins.

jsteemann:
@Fruchtgummi : I moved the changes over to another PR (https://github.com/arangodb/arangodb/pull/15414) a while ago and wanted to land them in our devel.
Unfortunately it turned out to be more complicated than I thought, because the serialization of numeric values to strings using the C++ standard library turned out to be not portable (i.e. different results on different platforms).
There is hope that `std::fmt` will help here, but I haven't tried it yet due to lack of time. 
I will revisit the issue in a few weeks when I will have more time. We can leave this PR here (and mine) open so it is not forgotten. I am still in favor of adding that functionality once there is a reliable and portable solution.

PR #15345, opened 2021-12-14
[CINFRA-148] Expand the Flags update to allow adding and removing followers

goedderz:
### Scope & Purpose

Allows to add and/or remove followers during a term while updating participant flags.

- [X] :pizza: New feature

#### Backports:

None.

#### Related Information

- [X] Jira ticket number: https://arangodb.atlassian.net/browse/CINFRA-148

### Testing & Verification

  - [X] Added new C++ **Unit tests**


arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -317,9 +317,10 @@ auto replicated_log::LogFollower::GuardedFollowerData::checkCommitIndex(
     }
   };
 
-  TRI_ASSERT(newLCI >= _largestCommonIndex)
-      << "req.lci = " << newLCI << ", this.lci = " << _largestCommonIndex;
-  if (_largestCommonIndex < newLCI) {
+  // This assertion is no longer true, as followers can now be added.
+  // TRI_ASSERT(newLCI >= _largestCommonIndex)
+  //     << "req.lci = " << newLCI << ", this.lci = " << _largestCommonIndex;
+  if (_largestCommonIndex != newLCI) {
```
maierlars, line 323: I see why this can happen. But I think we should prevent this from happening, because the LCI is essentially the index you can use for compaction and you might have removed entries that are then before the LCI.

In particular only the following can happen:
1. The lci is a X.
2. A follower F1 is added, the lci falls to F1.spearhead.index
3. The leader might have already compacted, hence it will override the followers log entirely, starting at X.
4. The LCI will rise again to X.

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -1250,20 +1260,86 @@ auto replicated_log::LogLeader::waitForLeadership()
...
+    }
+    if (data.activeParticipantsConfig->generation != previousGeneration) {
+      // This is to make sure the `additionalFollowers` list is really the
+      // (asymmetric) difference between the current and new configuration.
+      THROW_ARANGO_EXCEPTION_FORMAT(
+          TRI_ERROR_BAD_PARAMETER,
+          "assumed participant config generation does not match the current "
+          "generation - refusing to update; previous = %zu, current = %zu",
+          previousGeneration, data.activeParticipantsConfig->generation);
+    }
```
maierlars, line 1288: If there is not test for this, we should make sure to add one. This small piece of code is very important! 🙂

maierlars (approved):

arangod/Replication2/ReplicatedLog/LogFollower.cpp
goedderz: Right. I'll change the condition back to
```suggestion
  // This assertion is no longer true, as followers can now be added.
  // TRI_ASSERT(newLCI >= _largestCommonIndex)
  //     << "req.lci = " << newLCI << ", this.lci = " << _largestCommonIndex;
  if (_largestCommonIndex < newLCI) {
```
to prevent `_largestCommonIndex` from decreasing.

arangod/Replication2/ReplicatedLog/LogLeader.cpp
goedderz: Agreed, I'll add tests.

maierlars (approved):

PR #15353, opened 2021-12-15
WebUI node development module update (ansi-html)

hkernbach:
### Scope & Purpose

First of all: This is only a development environment related change.
`ansi-html` is a dependency which is being used by `webpack-dev-server`. 
As upgrading the development server needs a bit of time (Major upgrade, API changes), 
we'll use a resolution change for now targeting from:
`ansi-html 0.0.7` to `ansi-html-community 0.0.8`. 

- [x] :hammer: Node module upgrade

#### Backports:

- [x] Backport for 3.9: *https://github.com/arangodb/arangodb/pull/15355*

### Testing & Verification


js/apps/system/_admin/aardvark/APP/react/package.json
```diff
@@ -169,5 +170,8 @@
     "presets": [
       "react-app"
     ]
+  },
+  "resolutions": {
+    "ansi-html": "https://registry.yarnpkg.com/ansi-html-community/-/ansi-html-community-0.0.8.tgz#69fbc4d6ccbe383f9736934ae34c3f8290f1bf41"
```
adityamukho, line 175: Not sure if the full URL is really needed. Since the package exists at https://www.npmjs.com/package/ansi-html-community, it should suffice to just map to that package by name and version.

See https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md#mapping-version-specifications-as-well-as-packages-name

hkernbach: Thanks, let me check.

hkernbach: I tried and I failed, not working with that much cleaner and simpler solution. So unfortunately, we've to stick to this.

adityamukho: Ok, since this is the only option that works.

adityamukho (approved):
LGTM

ajurdak (approved):
LGTM.

PR #15354, opened 2021-12-15
Unify db._version for arangosh and server console

mpoeter:
### Scope & Purpose

Unify the result structure of `db._version(true)` calls for arangosh and server console. Previously such a call in the server console would return a different structure that only consisted of the `details` subobject. This is now unified so that the result structure in the server console is consistent with arangosh, but strictly speaking this is a breaking change.

Enterprise companion https://github.com/arangodb/enterprise/pull/828

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/848
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/828

### Testing & Verification

- [x] This change is already covered by existing tests, such as shell_server.
  - [x] Added new **integration tests** in shell_server


arangod/RestHandler/RestVersionHandler.cpp
```diff
@@ -35,6 +35,24 @@ using namespace arangodb;
 using namespace arangodb::basics;
 using namespace arangodb::rest;
 
+static void addVersionDetails(application_features::ApplicationServer& server, VPackBuilder& result) {
+  result.add("details", VPackValue(VPackValueType::Object));
+      Version::getVPack(result);
```
jsteemann, line 40: Indentation changes here unexpectedly.

arangod/RestHandler/RestVersionHandler.cpp
```diff
@@ -61,27 +73,24 @@ RestStatus RestVersionHandler::execute() {
...
-      result.close();
-    }  // found
-  }    // allowInfo
-  result.close();
+  ServerSecurityFeature& security = server().getFeature<ServerSecurityFeature>();
+
+  bool const allowInfo = security.canAccessHardenedApi();
+  bool found;
+  std::string const& detailsStr = _request->value("details", found);
+  bool const includeDetails = found && StringUtils::boolean(detailsStr);
```
jsteemann, line 91: While we are at this anyway, let's simplify it to
```suggestion
  bool const includeDetails = _request->parsedValue("details", false);
```

tests/js/client/replication2/replication2-cluster.js
```diff
@@ -39,8 +40,8 @@ const {
 } = jsunity.jsUnity.assertions;
 
 // The RestLogHandler is only available in maintainer mode.
-const isInMaintainerMode = require('internal').db._version(true)['maintainer-mode'] === 'true';
-const replication2Enabled = require('internal').db._version(true)['replication2-enabled'] === 'true';
+const isInMaintainerMode = require('internal').db._version(true).details['maintainer-mode'] === 'true';
```
jsteemann, line 43: Note: this also changes the access to the result of `db._version(true)` on the client side, i.e. in arangosh, but this PR did not modify the result structure of client-side version info.
I guess this change is made here is still ok because the previous access here was just wrong.

arangod/RestHandler/RestVersionHandler.cpp
mpoeter: Nice!

tests/js/client/replication2/replication2-cluster.js
mpoeter: Correct! The reason I found this inconsistency in the first place was because I copied this code to use somewhere else, but it did not work as expected. Due to this error these tests were always skipped before.

jsteemann (approved):
LGTM!

PR #15393, opened 2021-12-21
[CINFRA-33] Cheaper getStatus alternative

apetenchea:
### Scope & Purpose

`getStatus` returns a lot of information about the inner state of the log. This is sometimes inefficient, since just a small subset of this information is ever used. This pull request introduces `getQuickStats` and the corresponding `QuickLogStatus` object, meant to serve as a minimalist version of `LogStatus`.
Most of the replacements were done in *Maintenance*, by adding the `getReplicatedLogsQuickStatus` method, equivalent to `getReplicatedLogs`. No changes were made in places were the full log stats are useful, such as in the `RestLogHandler`.
Please note that unit tests were left alone, although we could use replace `getStatus` in there as well.
Consistency between `LogStatus` and `QuickLogStatus` is checked in `LogStatusTest.cpp`.

### Testing & Verification

- [x] This change is already covered by existing tests, such as `ReplicatedLog/MaintenaceTests.cpp`.


arangod/Cluster/Maintenance.cpp
```diff
@@ -1233,20 +1233,20 @@ static auto reportCurrentReplicatedLogLocal(replication2::replicated_log::LogSta
   return std::nullopt;
 }
 
-static auto reportCurrentReplicatedLogLeader(replication2::replicated_log::LeaderStatus const& status,
+static auto reportCurrentReplicatedLogLeader(replication2::replicated_log::QuickLogStatus const& status,
                                              replication2::agency::LogCurrent::Leader const* currentLeader)
     -> std::optional<replication2::agency::LogCurrent::Leader> {
   if (status.leadershipEstablished) {
```
maierlars, line 1239: We could assert here, that `status.role == kLeader`?

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -319,6 +319,18 @@ auto replicated_log::LogFollower::getStatus() const -> LogStatus {
...
 
+auto replicated_log::LogFollower::getQuickStatus() const -> QuickLogStatus {
+  return _guardedFollowerData.doUnderLock([this](auto const& followerData) {
+    if (followerData._logCore == nullptr) {
+      THROW_ARANGO_EXCEPTION(
+          TRI_ERROR_REPLICATION_REPLICATED_LOG_FOLLOWER_RESIGNED);
+    }
+    return QuickLogStatus{.role = ParticipantRole::kFollower,
+                          .term = _currentTerm,
+                          .local = followerData.getLocalStatistics()};
```
maierlars, line 330: shall we add leader established here as well? It would be set to true if the commit index is non-zero.

maierlars (approved):

arangod/Cluster/Maintenance.cpp
apetenchea: That is a good idea!

arangod/Replication2/ReplicatedLog/LogFollower.cpp
apetenchea: That used to be part only of the `LeaderStatus` before, that's why I have ignored if for followers. Since it's going to take up that space in memory anyway, it would be better indeed to put some proper value in it.

maierlars (approved):

PR #15396, opened 2021-12-21
[CINFRA-181] Rest API default limits

apetenchea:
### Scope & Purpose

This pull requests sets the following default values for some of the methods:
- head?limit=10
- tail?limit=10
- slice?start=0&end=11
- poll?first=0&limit=10

Constants are used instead of hardcoded values.
Some tests were modified to ensure consistency between the V8 implementation and the HTTP client.

### Testing & Verification

- [x] This change is already covered by existing tests, such as replication2 tests.

### DIFF
--- a/js/client/modules/@arangodb/replicated-logs.js
+++ b/js/client/modules/@arangodb/replicated-logs.js
@@ -33,6 +33,8 @@ function ArangoReplicatedLog(database, id) {
 
 exports.ArangoReplicatedLog = ArangoReplicatedLog;
 
+ArangoReplicatedLog.prototype.defaultLimit = 10;
+
 ArangoReplicatedLog.prototype.id = function () {
   return this._id;
 };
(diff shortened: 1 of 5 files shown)


js/client/modules/@arangodb/replicated-logs.js
```diff
@@ -69,25 +71,25 @@ ArangoReplicatedLog.prototype.status = function() {
   return requestResult.result;
 };
 
-ArangoReplicatedLog.prototype.head = function(limit = 100) {
+ArangoReplicatedLog.prototype.head = function(limit = this.defaultLimit) {
```
maierlars, line 74: Instead of doing that you could check for `=== undefined` and then simply omit the query parameter entirely. Then it will use the default value from the methods class and we can remove the duplicate value in line 36.

maierlars (approved):

apetenchea:
Some inconsistencies might arise due to the way `args` is handled in `v8-replicated-logs.cpp`. It is indexed like an array, therefore arguments are expected in the same order as they are in the docs. For example the `JS_Slice` method expects `start` on `args[0]` and `stop` on `args[1]`. On the other hand, the HTTP API does not have this limitation. For now it is not a problem, but this felt like a good moment to point it out.

maierlars (approved):

PR #15400, opened 2021-12-22
updated indirect dev dependency ansi-html to latest

mchacki:
Backport of #15353


js/apps/system/_admin/aardvark/APP/react/package.json
```diff
@@ -17,6 +17,7 @@
     "@types/styled-components": "^4.4.3",
     "@types/underscore": "^1.10.1",
     "axios": "^0.18.0",
+    "ansi-html-community": "0.0.8",
```
adityamukho, line 20: `ansi-html-community@0.0.8` is intended to serve as a drop-in replacement for `ansi-html@0.0.7`, since the latter has known security issues.

Since the package names are different, and also since `ansi-html` is a transitive dependency of `webpack-dev-server`, we need a different mechanism that would tell `npm` to apply this substitution in `package-lock.json` and also inside the `node_modules` folder.

This mechanism is available in `npm` from version 8.3.0 onwards, and is defined in the RFC linked below:
https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md

The documentation is available in the official `npm` docs here: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides.

In particular, the following syntax snippet ought to work, although this is based more on my interpretation of the docs, so it needs to be tested:

`package.json`
```
{
 ...
  "overrides": {
    "ansi-html@0.0.7": "npm:ansi-html-community@0.0.8"
  }
...
}
```

The above snippet should replace all (version-specific) mentions of `ansi-html` by `ansi-html-community`. This could be perhaps generalized to target any encountered version of `ansi-html`, not just a specific one.

Also, this requires `npm@>=8.3.0` in order to work.

PR #15406, opened 2021-12-23
micro-optimize connection pool function

jsteemann:
### Scope & Purpose

Micro-optimize method in ConnectionPool, by moving time-gathering outside of a mutexed section, and by moving a shard_ptr instead of copying it.
The performance impact of these changes will be negligible, so no backports should be done.

### Testing & Verification


arangod/Network/ConnectionPool.cpp
```diff
@@ -234,13 +234,14 @@ ConnectionPtr ConnectionPool::selectConnection(std::string const& endpoint,
...
+  milliseconds const ttl(_config.idleConnectionMilli);
   
-  std::lock_guard<std::mutex> guard(bucket.mutex);
-
   auto start = steady_clock::now();
-  
   isFromPool = true;  // Will revert for new collections
+
+  // exclusively lock the bucket
+  std::unique_lock<std::mutex> guard(bucket.mutex);
```
MBkkt, line 247: NIT: You don't need write <std::mutex>

arangod/Network/ConnectionPool.cpp
```diff
@@ -234,13 +234,14 @@ ConnectionPtr ConnectionPool::selectConnection(std::string const& endpoint,
                                                ConnectionPool::Bucket& bucket,
                                                bool& isFromPool) {
   using namespace std::chrono;
-  const milliseconds ttl(_config.idleConnectionMilli);
+  milliseconds const ttl(_config.idleConnectionMilli);
   
-  std::lock_guard<std::mutex> guard(bucket.mutex);
-
   auto start = steady_clock::now();
```
MBkkt, line 243: Do you really want measure time with lock time?

arangod/Network/ConnectionPool.cpp
```diff
@@ -295,17 +295,22 @@ ConnectionPtr ConnectionPool::selectConnection(std::string const& endpoint,
   auto c = std::make_shared<Context>(createConnection(builder),
                                      now, 1 /* leases*/);
   bucket.list.push_back(c);
+
+  guard.unlock();
+  // continue without the bucket lock
+
+  isFromPool = false;
```
MBkkt, line 307: How we use this bool&? Is change it without mutex ok?

jsteemann: Yes, it is used as a return value

arangod/Network/ConnectionPool.cpp
jsteemann: Yes, I think measuring the time including locking is even better, because it more correctly indicates how long things took.

MBkkt: Ok

mpoeter (approved):

PR #15412, opened 2021-12-23
[CINFRA-193] Speed up Maintenance

maierlars:
### Scope & Purpose

Speed up the maintenance by making databases dirty after an async operation on the replicated log completed.

### DIFF
--- a/lib/Basics/ResultT.h
+++ b/lib/Basics/ResultT.h
@@ -172,7 +172,7 @@ class ResultT {
 
   T const& operator*() const& { return get(); }
 
-  T&& operator*() && { return get(); }
+  T&& operator*() && { return std::move(get()); }
 
   T const&& operator*() const&& { return get(); }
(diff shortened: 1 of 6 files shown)


lib/Basics/ResultT.h
```diff
@@ -172,7 +172,7 @@ class ResultT {
 
   T const& operator*() const& { return get(); }
 
-  T&& operator*() && { return get(); }
+  T&& operator*() && { return std::move(get()); }
```
goedderz, line 175: Does this really make a difference?

goedderz (approved):

lib/Basics/ResultT.h
maierlars: The difference of compiling and not compiling, yes. `get` returns an l-value. Without `std::move` you can not return an r-value reference.

PR #15416, opened 2021-12-24
add optional "storedValues" attribute for persistent indexes

jsteemann:
### Scope & Purpose

The feature is described in the docs PR in detail, so there is no separate PR description here.
Docs PR: https://github.com/arangodb/docs/pull/858

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/858

### Testing & Verification

  - [x] Added new **integration tests** (shell_client, shell_server)


gnusi:
In ArangoSearch we have similar concept called "storedValues", would be nice to use similar term

jsteemann:
> In ArangoSearch we have similar concept called "storedValues", would be nice to use similar term

The index lookup values (a.k.a. "fields") are also stored in the index.
I don't think "storedValues" is a good term, because it someone implies that the other values (in "fields") are _not_ stored. The same is true for arangosearch by the way.
Maybe there is even a better term... Anyway, it is just naming, and changing it later (before merging) will be easy.

gnusi:
@jsteemann naming is also important for `inverted` index, currently we use term `storedValues` similar to views. So we have to agree on naming to be consistent.

Dronplane:
As a suggestion: we have a PR https://github.com/arangodb/arangodb/pull/14240 I hope to be merged soon. This PR introduces a `CoveringData` abstraction for the index callbacks. Could we wait for #14240 to be merged and use the same abstraction here? So callback will not "know" about index projections  storage schemes.

jsteemann:
Setting this to draft state until merge of #14240.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -874,12 +934,24 @@ Result RocksDBVPackIndex::insert(transaction::Methods& trx,
...
   if (_unique) {
-    // unique indexes have a different key structure
+    // build index value (storedValues array will be stored in value if
+    // storedValues are used)
     RocksDBValue value = RocksDBValue::UniqueVPackIndexValue(documentId);
+    if (!_storedValuesPaths.empty()) {
+      transaction::BuilderLeaser leased(&trx);
+      leased->openArray(true);
+      for (auto const& it : _storedValuesPaths) {
+        leased->add(doc.get(it));
```
Dronplane, line 944: What do we want to in case of missing attribute? For the `fields` we either stop indexing document (sparse index)  or forcefully put `null` slice. As I see here we may put `none`. Maybe always force `null` for consistency with `fields`?

Dronplane (approved):
LGTM. Left couple comments regarding missing `storedValues` attributes in  the indexed document

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -930,14 +1003,29 @@ Result RocksDBVPackIndex::insert(transaction::Methods& trx,
...
 
+    // build index value (storedValues array will be stored in value if
+    // storedValues are used)
     RocksDBValue value = RocksDBValue::VPackIndexValue();
-    for (RocksDBKey& key : elements) {
+    if (!_storedValuesPaths.empty()) {
+      transaction::BuilderLeaser leased(&trx);
+      leased->openArray(true);
+      for (auto const& it : _storedValuesPaths) {
+        leased->add(doc.get(it));
```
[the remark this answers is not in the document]
jsteemann: Addressed.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
jsteemann: Addressed.

Dronplane (approved):
LGTM

PR #15421, opened 2021-12-28
[3.9] clang-format operations

aMahanna:


aMahanna:
OS-dependent hooks & scripts (MacOS, Linux, Windows) have been introduced.

Linux & MacOS implementations are currently identical, but may be subject to change in the future (hence why I still decided to split them up).

joerg84 (approved):

PR #15438, opened 2021-12-29
[devel] patch clang-format sha value

aMahanna:
Addresses false-negatives of [this nature](https://github.com/arangodb/arangodb/runs/4659813949?check_suite_focus=true)


markuspf (approved):
LGTM

PR #15440, opened 2021-12-29
[CINFRA-168] Report Supervision Election result in coordinator

apetenchea:
### Scope & Purpose

The coordinator asks both the leader and the agency (using the agency cache) for the log status and election results. This way we make sure to get a clear picture of what is currently going on, even without a leader elected.

### Testing & Verification

- [ ] This change is already covered by existing tests, such as `shell_replicated_logs_cluster.js`.


tests/js/server/replication2/replication2-supervision-replicated-log-target-cluster.js
```diff
@@ -48,7 +49,7 @@ const waitForReplicatedLogAvailable = function (id) {
     while (true) {
         try {
             let status = db._replicatedLog(id).status();
-            if (status.role === "leader") {
+            if ("logStatus" in status && status.logStatus.role === "leader") {
```
maierlars, line 52: This looks like python to me, but it is [actually](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Operators/in) valid js 😆

arangod/Replication2/ReplicatedLog/LogStatus.h
```diff
@@ -134,4 +135,16 @@ struct LogStatus {
   VariantType _variant;
 };
 
+/**
+* @brief Provides a more general view of what's currently going on, without
+* completely relying on the leader.
+ */
+struct GlobalStatus {
+  agency::LogCurrentSupervision supervision;
+  std::optional<LogStatus> logStatus;
```
maierlars, line 144: I would have expected this to be a `unordered_map<ParticipantId, LogStatus>`. Hence you would get something like this:
```{
   "supervision": {...},
   "participants": { // only leader
     "PRMR-xxxx": { /* local status */ },
     // in future on could add the other participants here as well
   }
   "leaderId": "PRMR-xxxx",
}
```

arangod/Replication2/Methods.h
```diff
@@ -54,6 +58,10 @@ struct WaitForResult;
 struct ReplicatedLogMethods {
   static constexpr auto kDefaultLimit = std::size_t{10};
 
+  typedef std::variant<replication2::replicated_log::LogStatus,
+                       replication2::replicated_log::GlobalStatus>
+      GenericLogStatus;
```
maierlars, line 63:
```suggestion
  using GenericLogStatus = std::variant<replication2::replicated_log::LogStatus,
                       replication2::replicated_log::GlobalStatus>;
```
we usually don't use typedef because using is more generic and we want to keep consistent.

arangod/Replication2/Methods.cpp
```diff
@@ -202,23 +213,56 @@ struct ReplicatedLogMethodsCoordinator final
...
+      leaderId = getLogLeader(id);
+    } catch (basics::Exception const& ex) {
+      if (ex.code() == TRI_ERROR_REPLICATION_REPLICATED_LOG_LEADER_RESIGNED) {
+        return GlobalStatus{
+            replication2::agency::methods::getCurrentSupervision(vocbase, id),
+            {},
+            std::nullopt};
+      }
+      throw;
+    }
```
maierlars, line 238: instead of dealing with this try..catch here you could use `clusterInfo.getReplicatedLogLeader(vocbase.name(), id)` directly, which returns a `ResultT<ParticipantId>`.

arangod/Replication2/Methods.cpp
```diff
@@ -202,23 +213,56 @@ struct ReplicatedLogMethodsCoordinator final
...
+          auto supervision =
+              replication2::agency::methods::getCurrentSupervision(vocbase, id);
+          auto leaderStatus =
+              replication2::replicated_log::LogStatus::fromVelocyPack(
+                  resp.slice().get("result"));
+          auto participants = std::unordered_map<ParticipantId, LogStatus>{
+              {leaderId, std::move(leaderStatus)}};
+          return GlobalStatus{.supervision = std::move(supervision),
+                              .participants = std::move(participants),
+                              .leaderId = leaderId};
```
maierlars, line 256: you can move `leaderId` here.

arangod/V8Server/v8-replicated-logs.cpp
```diff
@@ -278,7 +278,31 @@ static void JS_Status(v8::FunctionCallbackInfo<v8::Value> const& args) {
   }
 
   auto result =
-      ReplicatedLogMethods::createInstance(vocbase)->getLogStatus(id).get();
+      ReplicatedLogMethods::createInstance(vocbase)->getStatus(id).get();
+  auto response = std::visit(
+      [](auto&& r) {
+        VPackBuilder response;
+        r.toVelocyPack(response);
+        return std::move(response);
```
maierlars, line 286: This `std::move` prevents copy elision.

maierlars (changes requested):

maierlars (approved):

PR #15443, opened 2021-12-30
[3.9] fix: clang-format git diff filter

aMahanna:
Copy of #15442 for `3.9`

* adds the [T] value to the `--diff-filter` flag, to stay consistent with the [current clang-format action implementation](https://github.com/arangodb/arangodb/blob/3.9/.github/workflows/clang-format.yml#L19)

[Link]( https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203) to `--diff-filter` documentation


KVS85 (approved):
LGTM.

Issue #14630, opened 2021-08-12
arangodump fails db backup - exit code 141


duduz:
## My Environment

* __ArangoDB Version__:        
* __Storage Engine__:             
* __Deployment Mode__:       
* __Deployment Strategy__:   
* __Configuration__:               
* __Infrastructure__:               
* __Operating System__:        
* __Total RAM in your machine__:        
* __Disks in use__:        
* __Used Package__:              

## Component, Query & Data

__Affected feature__:

__Size of your Dataset on disk__:

__Replication Factor & Number of Shards (Cluster only)__:

## Steps to reproduce

1. arangodump --configuration none --progress true --server.endpoint tcp://internal:8529 --server.authentication true --output-directory arangodb-dump/dbname --server.endpoint=ssl://arango.svc:8529/ --server.database=dbname --server.password=PW --dump-data true --include-system-collections true --compress-output --log.output file:///var/log/arangodump.log --overwrite true --threads 1
2. executing this with more threads, same behavior
3. using arangodump 3.7.9, however, 3.8.0 provides the same results.
4. it doesn't look like a storage problem / memory since there's way more than enough of that on the machine
5. I'm able to replicate it quite consistently, this happens only with one or two database (probably the largest).

__Problem__:
After finishing the export of the database, arangodump exits with exit code 141.
However, it actually finishes to export all the collections in the dataset, no errors in the log.
when finishing to export the last collection, it hangs for a few seconds and skips the typical "completion" message:
> Processed 43 collection(s), wrote 408173500 byte(s) into datafiles, sent 88 batch(es)
and exits with exit code 141.

__Expected result__:
1. Understand what exit code 141 means
2. Finish backups with exit code 0 and get the typical "completion" message :)

duduz:
Still an issue..
Dataset is much bigger now.

calebjacobson:
I am having this same issue, both with arangodump and arangorestore (v3.6.4). My dataset is much smaller at ~1Gb dump (directory with compressed data files and structure files). issues started happening after I added a collection with a lot of data (~800Mb compressed dump of data file) compared to the db size. We do nightly dumps of our db and use some of those dumps in our test environment (using arangorestore). This issues is happening whenever I run arangodump, and every night in our automated process (although at first it only happened every few nights - however, the largest collection is only ever written to with new data, data is never erased or changed after being written).

dothebart:
Hi, 
Please note that ArangoDB 3.6.x is EOL.
[...]

PR #15445, opened 2022-01-03
BTS-710: fix undefined behaviour by initializing the exit code variable

dothebart:
### Scope & Purpose

In case of success Arangodump would choose a random exit value

#### Backports:

- [x] Backport for 3.8: *(Please link PR)*

#### Related Information
- [x] GitHub issue https://github.com/arangodb/arangodb/issues/14630

### Testing & Verification

Run arangodump more often, to find whether it exits zero.


client-tools/Dump/arangodump.cpp
```diff
@@ -67,7 +67,7 @@ int main(int argc, char* argv[]) {
             argv[0], "Usage: arangodump [<options>]",
             "For more information use:", BIN_DIRECTORY));
     ApplicationServer server(options, BIN_DIRECTORY);
-    int ret;
+    int ret = EXIT_SUCCESS;
```
jsteemann, line 70: Initialization is always good.

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Fixed undefined behaviour for exit codes of arangodump
```
```suggestion
* Fixed potentially undefined behavior for exit code of arangodump.
```

lib/Basics/threads-posix.cpp
```diff
@@ -65,7 +65,7 @@ struct thread_data_t {
 ////////////////////////////////////////////////////////////////////////////////
 
 static void* ThreadStarter(void* data) {
-  arangodb::signals::maskAllSignals();
+  arangodb::signals::maskAllSignalsServer();
```
jsteemann, line 68: I dislike calling the server-side variant here unconditionally, as the code in `lib` is used by the client-tools as well.
I have no doubt it will work for _now_, but it doesn't look very fool-proof.
We should **at least** leave a code comment here why we are unconditionally calling the server variant here. 
I would even prefer to have 2 different implementations for `maskAllSignals()` which are compiled separately in different TUs (one only for arangod, the other one for client-tools). That sounds like the cleanest solution to me.

CHANGELOG
```diff
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Fixed potentially undefined behavior for exit code of arangodump.
+
+* ignore SIG_HUP etc. on client tools.
```
```suggestion
* Ignore signals such as SIGPIPE in client tools.
```

client-tools/Backup/arangobackup.cpp
```diff
@@ -50,6 +51,7 @@ int main(int argc, char* argv[]) {
   TRI_GET_ARGV(argc, argv);
   return ClientFeature::runMain(argc, argv, [&](int argc, char* argv[]) -> int {
     ArangoGlobalContext context(argc, argv, BIN_DIRECTORY);
+    arangodb::signals::maskAllSignalsClient();
     context.installHup();
```
jsteemann, line 55: Do we really need a handler for SIGHUP in the client tools? I know it was there before already, but I have my doubts this is useful.

client-tools/Dump/arangodump.cpp
jsteemann: But... why do we only initialize in arangodump and not the other client tools? If we fix it here, we should fix it in the other client tools as well.

jsteemann (approved):
LGTM

PR #15446, opened 2022-01-03
add more indexHints: `disableIndex` and `maxProjections`

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/864

* Added `disableIndex` index hint for AQL FOR loops. This index hint disables
  the usage of any index (except geo and fulltext indexes) and will always cause 
  a full scan over the collection.
  In some circumstances a full scan can be more efficient than an index scan,
  for example if the index scan produces many matches (close to the number of
  documents in the collection) and the index is not fully covering the query.
  The `disableIndex` hint can be given per FOR loop in the query, e.g.:

      FOR doc IN collection OPTIONS { disableIndex: true } 
        RETURN doc.value

  The default value of `disableIndex` is `false`.
  In case a different index hint is provided, `disableIndex: true` takes
  precendence and produces a warning about the ambiguous settings.

* Added `maxProjections` hint for AQL FOR loops. This hint can be used to set
  the maximum number of document attributes that are taken into account for
  using projections.

  For example, in the following query, no projections will be used because the
  number of potential projection attributes (`value1`, value2`, `value3`) is 
  higher than the maximum number of projection attributes set via the
  `maxProjections` option:

      FOR doc IN collection OPTIONS { maxProjections: 2 } 
        RETURN [ doc.value1, doc.value2, doc.value3 ]

  The default value for `maxProjections` is `5`, which is compatible with the
  previous hard-coded default value.

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/864

### Testing & Verification


arangod/Aql/IndexHint.cpp
```diff
@@ -176,6 +219,7 @@ void IndexHint::toVelocyPack(VPackBuilder& builder) const {
   TRI_ASSERT(builder.isOpenObject());
   VPackObjectBuilder guard(&builder, ::FieldContainer);
   builder.add(::FieldForced, VPackValue(_forced));
+  builder.add(::FieldLookahead, VPackValue(_lookahead));
```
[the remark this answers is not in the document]
jsteemann: Same as above. The `maxProjections` attribute is actually not owned/managed by IndexHint, but by DocumentProducingNode.

tests/js/server/aql/aql-projections.js
```diff
@@ -657,7 +655,88 @@ function projectionsExtractionTestSuite () {
...
+        } else {
+          assertEqual(-1, plan.rules.indexOf(ruleName));
+        }
+      });
+    },
+    
+    testAdjustedMaxProjections : function () {
+      let queries = [];
+      [ 1, 2, 3, 5, 6, 10, 12 ].forEach((maxProjections) => {
+        for (let i = 1; i < 11; ++i) {
```
mchacki, line 708: This test will never exceed the 10 yet the 12 limit.
```suggestion
        for (let i = 1; i < 14; ++i) {
```

mchacki: Alternative stop at 10 max value.
Both variants are good enough.
(This test is by no means wrong, it just ends at unexpected position to me)

mchacki (approved):
From my point of view this PR looks good.

I have a minor suggestion on a test, but I can certainly live with it being not applied.

Regarding the discussion on extractMaxProjections in indexHint.

I see points on both sides.
The parsing does not belong into IndexHint.
Pulling that out is more effort for quite little gain.
In ideal world I would imagine something like:

one IndexHint covering everything related to selecting an index. Owned by the EnumerateCollection (actually not even there, as "Enumerate" is the decision to not take the index, but that is way too much work for virtually no gain)
and another "LookupHint" covering Projections, and everything that has to be applied on the Document lookup, not on the IndexLookup.
But for me we could do this cleanup later.

maierlars (approved):

PR #15451, opened 2022-01-04
[CINFRA-184] Replicated Log + Supervision Integration Test

maierlars:
### Scope & Purpose

Add integration tests for replicated logs and supervision code. This is just a skeleton and more tests will be added soon in different PRs.


apetenchea (approved):
Runs smoothly

PR #15457, opened 2022-01-05
[CINFRA-190] Move participant list out of term

goedderz:
### Scope & Purpose

Currently, there are two participant lists in the Agency's Plan. One in `currentTerm` of a replicated log, the other in the `participantConfig`. While the latter can hold additional flags, but which entries are optional as long as the flags don't differ from the default.

The participant list in `currentTerm` should be removed, at least because the participant list can now be changed during a term. Plus, it is both undesirable and unnecessary to maintain two lists that must somewhat match each other.

Thus this PR removes `currentTerm.participants`, and instead uses `participantConfig.participants`, while the latter now always holds all participants, not just those with differing flags.

- [X] :pizza: New feature

#### Backports:

None.

### Testing & Verification

  - [X] Added new C++ **Unit tests**


arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -348,7 +393,8 @@ auto algorithms::updateReplicatedLog(
       // Note that newParticipants contains the leader, while oldFollowers does
       // not.
       auto const& oldFollowers = leaderStatus->follower;
-      auto const& newParticipants = spec->currentTerm->participants;
+      auto const& newParticipants = spec->participantsConfig.participants;
+      // TODO move this calculation into updateParticipantsConfig()
```
maierlars, line 397: This TODO was introduced here, would you like to create a ticket for that?

maierlars (approved):

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -80,7 +80,7 @@ auto LogFollower::appendEntriesPreFlightChecks(GuardedFollowerData const& data,
     // hard to find out. Thus we increase the log level for this message to make
     // this more visible.
 #ifdef ARANGODB_USE_GOOGLE_TESTS
-#define LOST_LOG_CORE_LOGLEVEL WARN
+#define LOST_LOG_CORE_LOGLEVEL INFO
```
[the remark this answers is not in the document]
goedderz: I think so. There are tests where this does happen, so I wanted to reduce the noise for working test runs, while not making it invisible when something went wrong. But I'm fine either way.

arangod/Replication2/ReplicatedLog/Algorithms.cpp
goedderz: Done: [CINFRA-213](https://arangodb.atlassian.net/browse/CINFRA-213)

maierlars (approved):

PR #15467, opened 2022-01-06
fix pitfall for comdat folding

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/848

we have had 2 static methods with identical code, which returned the
addresses of the distinct method-local static objects. when enabling
comdat folding, the compiler decided to merge the two functions into
one, which gave both objects the same in-memory address.
this lead to follow-up errors because the object addresses were used to
tell two types of object (collections and views) apart.
this was not only error-prone, but also complicated. this PR replaces
all this with a simple enum class.

### Testing & Verification


arangod/VocBase/LogicalCollection.h
```diff
@@ -105,7 +105,7 @@ class LogicalCollection : public LogicalDataSource {
   };
 
   /// @brief the category representing a logical collection
-  static Category const& category() noexcept;
```
gnusi, line 108: it has to remain `static` as member is already present in `LogicalDataSource`

arangod/VocBase/LogicalDataSource.h
```diff
@@ -124,7 +104,7 @@ class LogicalDataSource {
 
   virtual ~LogicalDataSource() = default;
 
-  Category const& category() const noexcept { return _category; }
+  virtual Category category() const noexcept = 0;
```
gnusi, line 107: let's just store Category as a member to avoid unnecessary virtual calls

gnusi (changes requested):

arangod/VocBase/LogicalCollection.h
jsteemann: I don't think so. It is declared as pure virtual in `LogicalDataSource`.

arangod/VocBase/LogicalDataSource.h
jsteemann: That would also work, but I dislike that approach for the reason that then in theory it would be possible to create a LogicalView with category `kCollection`. I prefer the category to be part of the derived classes, so it's error-proof to use (and one less data member to carry around and think about). Of course this approach has the virtual calls as a downside, but I wouldn't expect any actual performance regression in our tests. Normally the category is determined once or maybe twice per LogicalDataSource for a high-level operation. E.g. an AQL query iterating over the docs in a collection will determine the category once or twice at start, but the actual query execution will not do it. 
So it shouldn't really matter IMO.

gnusi: @jsteemann we can add an assertion in `LogicalCollection`/`LogicalView` to ensure that category is correct

gnusi (approved):
LGTM, but I believe we need to re-enable comdat folding on MSVC

mpoeter (approved):

PR #15473, opened 2022-01-07
[bts-590] Check writeConcern and replicationFactor in WebUI

ajurdak:
### Scope & Purpose

It was possible to create a database despite the value of write concern was bigger than the value of the replication factor. Now these two values will be compared and in case the write concern is bigger than the replication factor no database will be created and an error message will be displayed.

#### Backports:

- [x] Backport for devel: #15081
- [x] Backport for 3.9: #15473 
- [x] Backport for 3.8: #15475
- [x] Backport for 3.7: #15477

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/BTS-590

### Testing & Verification


js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
```diff
@@ -176,34 +176,39 @@
...
-      this.collection.create(options, {
-        error: function (data, err) {
-          self.handleError(err, dbname);
-        },
-        success: function (data) {
-          if (window.location.hash === '#databases') {
-            self.updateDatabases();
+      
+      let replicationFactor = $('#new-replication-factor').val();
+      let writeConcern = $('#new-write-concern').val();
```
hkernbach, line 181: replicationFactor and writeConcern could be string types if a user enters characters, right? In that case, I would add a try-catch block around the whole next block:
```
if(parseInt(writeConcern) <= parseInt(replicationFactor)) {
```

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
```diff
@@ -176,34 +176,39 @@
...
+
+      if(parseInt(writeConcern) <= parseInt(replicationFactor)) {
+        var options = {
+          name: dbname,
+          "options" : {
+            "sharding" : sharding,
+            "replicationFactor" : Number(replicationFactor),
+            "writeConcern" : Number(writeConcern),
```
hkernbach, line 189: As we're parsing here to a numeric value again, I would create a small try-catch block before we reach the if statement where we can parse the value (and report and error in case it fails) and then execute the if block afterwards without the parse statement like:
```
if(writeConcern <= replicationFactor) {
```

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
hkernbach: See updated comment below.

hkernbach (changes requested):

hkernbach (approved):

PR #15485, opened 2022-01-09
SEARCH-184  Cluster metrics

MBkkt:
### Scope & Purpose

Add cache for fast text objects

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: SEARCH-184
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/ClusterMethods.cpp
+++ b/arangod/Cluster/ClusterMethods.cpp
@@ -56,11 +56,11 @@
 #include "Transaction/Manager.h"
 #include "Transaction/ManagerFeature.h"
 #include "Transaction/Methods.h"
+#include "Utilities/NameValidator.h"
 #include "Utils/CollectionNameResolver.h"
 #include "Utils/Events.h"
 #include "Utils/ExecContext.h"
 #include "Utils/OperationOptions.h"
-#include "Utilities/NameValidator.h"
 #include "V8Server/FoxxFeature.h"
 #include "VocBase/KeyGenerator.h"
 #include "VocBase/LogicalCollection.h"
@@ -71,10 +71,6 @@
 #ifdef USE_ENTERPRISE
 #include "Enterprise/RocksDBEngine/RocksDBHotBackup.h"
 #endif
-#include "StorageEngine/EngineSelectorFeature.h"
-#include "ClusterEngine/Common.h"
-#include "ClusterEngine/ClusterEngine.h"
-
 #include <velocypack/Buffer.h>
 #include <velocypack/Builder.h>
 #include <velocypack/Collection.h>
@@ -82,14 +78,17 @@
 #include <velocypack/Iterator.h>
 #include <velocypack/Slice.h>
 
+#include <algorithm>
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
-
-#include <algorithm>
 #include <numeric>
-#include <vector>
 #include <random>
+#include <vector>
+
+#include "ClusterEngine/ClusterEngine.h"
+#include "ClusterEngine/Common.h"
+#include "StorageEngine/EngineSelectorFeature.h"
 
 using namespace arangodb;
 using namespace arangodb::basics;
@@ -1352,13 +1351,62 @@ futures::Future<OperationResult> countOnCoordinator(
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// @brief gets the metrics from DBServers
+////////////////////////////////////////////////////////////////////////////////
+
+futures::Future<metrics::RawDBServers> metricsOnCoordinator(
+    NetworkFeature& network, ClusterFeature& cluster) {
+  auto* pool = network.pool();
+  auto serverIds = cluster.clusterInfo().getCurrentDBServers();
+
+  std::vector<Future<network::Response>> futures;
+  futures.reserve(serverIds.size());
+  for (auto const& id : serverIds) {
+    network::Headers headers;
+    headers.emplace(StaticStrings::Accept,
+                    StaticStrings::MimeTypeJsonNoEncoding);
+    futures.push_back(network::sendRequest(
+        pool, "server:" + id, fuerte::RestVerb::Get, "/_admin/metrics", {},
+        network::RequestOptions{}.param("type", "json"), std::move(headers)));
+  }
+  return collectAll(futures).thenValue(
+      [](std::vector<Try<network::Response>>&& responses) {
+        metrics::RawDBServers metrics;
+        metrics.reserve(responses.size());
+        for (auto& response : responses) {
+          if (!response.hasValue() || response->fail() ||
+              !response->hasResponse()) {
+            continue;  // Shit happens, just ignore it
+          }
+          auto payload = response->response().stealPayload();
+          if (!payload) {
+            TRI_ASSERT(false);
+            continue;
+          }
+          velocypack::Slice slice{payload->data()};
+          if (!slice.isArray()) {
+            TRI_ASSERT(false);
+            continue;
+          }
+          if (auto const size = slice.length(); size % 3 != 0) {
+            TRI_ASSERT(false);
+            continue;
+          }
+          metrics.push_back(std::move(payload));
+        }
+        return metrics;
+      });
+}
+
+////////////////////////////////////////////////////////////////////////////////
 /// @brief gets the selectivity estimates from DBservers
 ////////////////////////////////////////////////////////////////////////////////
 
-Result selectivityEstimatesOnCoordinator(
-    ClusterFeature& feature, std::string const& dbname,
-    std::string const& collname,
-    std::unordered_map<std::string, double>& result, TransactionId tid) {
+Result selectivityEstimatesOnCoordinator(ClusterFeature& feature,
+                                         std::string const& dbname,
+                                         std::string const& collname,
+                                         IndexEstMap& result,
+                                         TransactionId tid) {
   // Set a few variables needed for our work:
   ClusterInfo& ci = feature.clusterInfo();
 
@@ -1414,7 +1462,7 @@ Result selectivityEstimatesOnCoordinator(
   //            }
   // }
 
-  std::map<std::string, std::vector<double>> indexEstimates;
+  containers::FlatHashMap<std::string, std::vector<double>> indexEstimates;
   for (Future<network::Response>& f : futures) {
     network::Response const& r = f.get();
 
--- a/arangod/ClusterEngine/ClusterSelectivityEstimates.cpp
+++ b/arangod/ClusterEngine/ClusterSelectivityEstimates.cpp
@@ -30,33 +30,24 @@
 #include "Indexes/Index.h"
 #include "VocBase/LogicalCollection.h"
 
-using namespace arangodb;
+namespace arangodb {
 
 ClusterSelectivityEstimates::ClusterSelectivityEstimates(
     LogicalCollection& collection)
-    : _collection(collection), _updating(false) {}
+    : _collection(collection) {}
 
 void ClusterSelectivityEstimates::flush() {
-  // wait until we ourselves are able to set the _updating flag
-  while (_updating.load(std::memory_order_relaxed) ||
-         _updating.exchange(true, std::memory_order_acquire)) {
-    std::this_thread::yield();
-  }
-
-  auto guard = scopeGuard(
-      [this]() noexcept { _updating.store(false, std::memory_order_release); });
-
+  std::lock_guard guard{_update};
   std::atomic_store(&_data, std::shared_ptr<InternalData>());
 }
 
-IndexEstMap ClusterSelectivityEstimates::get(bool allowUpdating,
+IndexEstMap ClusterSelectivityEstimates::get(bool allowUpdate,
                                              TransactionId tid) {
   auto data =
       std::atomic_load<ClusterSelectivityEstimates::InternalData>(&_data);
-
-  if (allowUpdating) {
+  std::unique_lock guard{_update, std::defer_lock};
+  if (allowUpdate) {
     double const now = TRI_microtime();
-
     bool useExpired = false;
     int tries = 0;
     do {
@@ -69,51 +60,38 @@ IndexEstMap ClusterSelectivityEstimates::get(bool allowUpdating,
           return estimates;
         }
       }
-
-      // only one thread is allowed to fetch the estimates from the DB servers
-      // at any given time
-      if (_updating.load(std::memory_order_relaxed) ||
-          _updating.exchange(true, std::memory_order_acquire)) {
+      if (!guard.try_lock()) {
+        // only one thread is allowed to fetch the estimates from the DB servers
+        // at any given time
         useExpired = true;
       } else {
-        auto guard = scopeGuard([this]() noexcept {
-          _updating.store(false, std::memory_order_release);
-        });
-
-        // must fetch estimates from coordinator
-        IndexEstMap estimates;
-        Result res = selectivityEstimatesOnCoordinator(
-            _collection.vocbase().server().getFeature<ClusterFeature>(),
-            _collection.vocbase().name(), _collection.name(), estimates, tid);
-
-        if (res.ok()) {
-          // store the updated estimates and return them
+        IndexEstMap estimates;  // must fetch estimates from coordinator
+        if (selectivityEstimatesOnCoordinator(
+                _collection.vocbase().server().getFeature<ClusterFeature>(),
+                _collection.vocbase().name(), _collection.name(), estimates,
+                tid)
+                .ok()) {  // store the updated estimates and return them
           set(estimates);
           return estimates;
         }
+        guard.unlock();
       }
-
-      data =
-          std::atomic_load<ClusterSelectivityEstimates::InternalData>(&_data);
-    } while (++tries <= 3);
+      data = std::atomic_load<InternalData>(&_data);
+    } while (++tries <= 3);  // give up!
   }
 
-  // give up!
   if (data) {
-    // we have got some estimates before
-    return data->estimates;
+    return data->estimates;  // we have got some estimates before
   }
-  // return an empty map!
-  return IndexEstMap();
+  return {};  // return an empty map!
 }
 
-void ClusterSelectivityEstimates::set(IndexEstMap const& estimates) {
+void ClusterSelectivityEstimates::set(IndexEstMap estimates) {
   // push new selectivity values into indexes' cache
   auto indexes = _collection.getIndexes();
 
-  for (std::shared_ptr<Index>& idx : indexes) {
+  for (auto& idx : indexes) {
     auto it = estimates.find(std::to_string(idx->id().id()));
-
     if (it != estimates.end()) {
       idx->updateClusterSelectivityEstimate(it->second);
     }
@@ -126,7 +104,8 @@ void ClusterSelectivityEstimates::set(IndexEstMap const& estimates) {
   }
 
   // finally update the cache
-  std::atomic_store(&_data,
-                    std::make_shared<ClusterSelectivityEstimates::InternalData>(
-                        estimates, TRI_microtime() + ttl));
+  std::atomic_store(&_data, std::make_shared<InternalData>(
+                                std::move(estimates), TRI_microtime() + ttl));
 }
+
+}  // namespace arangodb
--- a/arangod/IResearch/IResearchLink.cpp
+++ b/arangod/IResearch/IResearchLink.cpp
@@ -21,18 +21,18 @@
 /// @author Andrey Abramov
 /// @author Vasiliy Nabatchikov
 ////////////////////////////////////////////////////////////////////////////////
+#include "IResearchLink.h"
+
 #include <index/column_info.hpp>
 #include <utils/singleton.hpp>
 
-#include "IResearchDocument.h"
-#include "IResearchLink.h"
-
 #include "ApplicationFeatures/ApplicationServer.h"
 #include "Aql/QueryCache.h"
-#include "Basics/StaticStrings.h"
 #include "Basics/DownCast.h"
+#include "Basics/StaticStrings.h"
 #include "Cluster/ClusterFeature.h"
 #include "Cluster/ClusterInfo.h"
+#include "IResearchDocument.h"
 #ifdef USE_ENTERPRISE
 #include "Cluster/ClusterMethods.h"
 #endif
@@ -40,13 +40,14 @@
 #include "IResearch/IResearchCompression.h"
 #include "IResearch/IResearchFeature.h"
 #include "IResearch/IResearchLinkHelper.h"
+#include "IResearch/IResearchMetricStats.h"
 #include "IResearch/IResearchPrimaryKeyFilter.h"
 #include "IResearch/IResearchView.h"
 #include "IResearch/IResearchViewCoordinator.h"
 #include "IResearch/VelocyPackHelper.h"
 #include "Metrics/Batch.h"
-#include "Metrics/Guard.h"
 #include "Metrics/GaugeBuilder.h"
+#include "Metrics/Guard.h"
 #include "Metrics/MetricsFeature.h"
 #include "RestServer/DatabaseFeature.h"
 #include "RestServer/DatabasePathFeature.h"
--- a/arangod/IResearch/IResearchLink.cpp
+++ b/arangod/IResearch/IResearchLink.cpp
@@ -115,7 +115,7 @@ DECLARE_GAUGE(arangodb_search_consolidation_time, uint64_t,
   checkFieldFeatures(meta, checkFieldFeatures);
 }
 
-constexpr std::string_view kSearchStats = "search_stats";
+constexpr std::string_view kSearchStats = "arangodb_search_link_stats";
 
 template<typename T>
 T getMetric(IResearchLink const& link) {
--- a/arangod/IResearch/IResearchLink.h
+++ b/arangod/IResearch/IResearchLink.h
@@ -35,53 +35,12 @@
 #include "IResearch/IResearchViewMeta.h"
 #include "Indexes/Index.h"
 #include "Metrics/Fwd.h"
-#include "Metrics/Guard.h"
 #include "RestServer/DatabasePathFeature.h"
 #include "Transaction/Status.h"
 #include "Utils/OperationOptions.h"
 #include "VocBase/Identifiers/IndexId.h"
 
 namespace arangodb::iresearch {
-
-struct MetricStats : public metrics::Guard<IResearchDataStore::Stats> {
-  static constexpr size_t kSize = 5;
-  static constexpr std::array<std::string_view, kSize> kName = {
-      "arangodb_search_num_docs",       //
-      "arangodb_search_num_live_docs",  //
-      "arangodb_search_num_segments",   //
-      "arangodb_search_num_files",      //
-      "arangodb_search_index_size",     //
-  };
-  using DataToString = std::string (*)(Data const&);
-  static constexpr std::array<DataToString, kSize> kToString = {
-      [](IResearchDataStore::Stats const& stats) {
-        return std::to_string(stats.numDocs);
-      },
-      [](IResearchDataStore::Stats const& stats) {
-        return std::to_string(stats.numLiveDocs);
-      },
-      [](IResearchDataStore::Stats const& stats) {
-        return std::to_string(stats.numSegments);
-      },
-      [](IResearchDataStore::Stats const& stats) {
-        return std::to_string(stats.numFiles);
-      },
-      [](IResearchDataStore::Stats const& stats) {
-        return std::to_string(stats.indexSize);
-      },
-  };
-  // TODO(MBkkt) Remove these arrays when we make generation maps from docs
-  static constexpr std::array<std::string_view, kSize> kHelp = {
-      "Number of documents",         //
-      "Number of live documents",    //
-      "Number of segments",          //
-      "Number of files",             //
-      "Size of the index in bytes",  //
-  };
-  static constexpr std::array<std::string_view, kSize> kType = {
-      "gauge", "gauge", "gauge", "gauge", "gauge",
-  };
-};
 
 ////////////////////////////////////////////////////////////////////////////////
 /// @brief common base class for functionality required to link an ArangoDB
@@ -174,7 +133,8 @@ class IResearchLink : public IResearchDataStore {
   /// @brief initialize from the specified definition used in make(...)
   /// @return success
   ////////////////////////////////////////////////////////////////////////////////
-  Result init(velocypack::Slice definition, InitCallback const& init = {});
+  virtual Result init(velocypack::Slice definition,
+                      InitCallback const& init = {});
 
   ////////////////////////////////////////////////////////////////////////////////
   /// @return arangosearch internal format identifier
--- a/arangod/IResearch/IResearchLinkCoordinator.cpp
+++ b/arangod/IResearch/IResearchLinkCoordinator.cpp
@@ -27,19 +27,21 @@
 #include <velocypack/Builder.h>
 #include <velocypack/Slice.h>
 
+#include <memory>
+
 #include "ApplicationFeatures/ApplicationServer.h"
 #include "Basics/StringUtils.h"
-#include "Cluster/ClusterInfo.h"
 #include "ClusterEngine/ClusterEngine.h"
-#include "IResearchCommon.h"
-#include "IResearchFeature.h"
-#include "IResearchLinkHelper.h"
-#include "IResearchViewCoordinator.h"
+#include "IResearch/IResearchCommon.h"
+#include "IResearch/IResearchFeature.h"
+#include "IResearch/IResearchLinkHelper.h"
+#include "IResearch/IResearchMetricStats.h"
 #include "Indexes/IndexFactory.h"
 #include "Logger/Logger.h"
+#include "Metrics/ClusterMetricsFeature.h"
+#include "Metrics/Metric.h"
 #include "RocksDBEngine/RocksDBIndex.h"
 #include "StorageEngine/EngineSelectorFeature.h"
-#include "VelocyPackHelper.h"
 #include "VocBase/LogicalCollection.h"
 
 namespace {
@@ -67,8 +69,7 @@ ClusterEngineType getEngineType(ArangodServer& server) {
 
 }  // namespace
 
-namespace arangodb {
-namespace iresearch {
+namespace arangodb::iresearch {
 
 IResearchLinkCoordinator::IResearchLinkCoordinator(
     IndexId id, LogicalCollection& collection)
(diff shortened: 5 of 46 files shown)


arangod/Metrics/ClusterMetricsFeature.h
```diff
@@ -0,0 +1,99 @@
...
+#include <variant>
+
+namespace arangodb::metrics {
+
+class ClusterMetricsFeature final
+    : public application_features::ApplicationFeature {
+ public:
+  using MetricValue = std::variant<uint64_t>;
+  using MetricKey = std::pair<std::string, std::string>;
+  using CoordinatorMetrics = std::map<MetricKey, MetricValue>;
```
MBkkt, line 47: Maybe replace to `FlatHashMap<std::string, FlatHashMap<std::string, MetricValue>>`?

arangod/ClusterEngine/ClusterSelectivityEstimates.cpp
```diff
@@ -30,33 +30,24 @@
...
-
-  auto guard = scopeGuard(
-      [this]() noexcept { _updating.store(false, std::memory_order_release); });
-
+  std::lock_guard guard{_update};
   std::atomic_store(&_data, std::shared_ptr<InternalData>());
 }
 
-IndexEstMap ClusterSelectivityEstimates::get(bool allowUpdating,
+IndexEstMap ClusterSelectivityEstimates::get(bool allowUpdate,
```
gnusi, line 44: `IndexEstMap` can be made `FlatHashMap<std::string, double>` :-)

gnusi (changes requested):

arangod/IResearch/IResearchLink.h
```diff
@@ -278,8 +278,8 @@ class IResearchLink {
   /// @brief initialize from the specified definition used in make(...)
   /// @return success
   ////////////////////////////////////////////////////////////////////////////////
-  Result init(velocypack::Slice const& definition,
-              InitCallback const& initCallback = {});
+  virtual Result init(velocypack::Slice definition,
```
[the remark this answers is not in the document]
MBkkt: Because I want if we have Coordinator link, we call its init

arangod/Metrics/Batch.h
```diff
@@ -42,23 +42,27 @@ class Batch final : public Metric {
...
+    load().toVPack(server, builder, labels());
+  }
   void toPrometheusBegin(std::string& result,
                          std::string_view name) const final {
     std::lock_guard guard{_m};
     _metric.needName();
   }
 
-  void store(T&& metric) {
+  void store(T metric) {
```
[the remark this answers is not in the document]
MBkkt: Yeah, you right, expect inline compiler optimization.

I think I just add two overloads: && and const&

arangod/Metrics/ClusterMetricsFeature.h
```diff
@@ -0,0 +1,99 @@
...
+#include <shared_mutex>
+#include <variant>
+
+namespace arangodb::metrics {
+
+class ClusterMetricsFeature final
+    : public application_features::ApplicationFeature {
+ public:
+  using MetricValue = std::variant<uint64_t>;
+  using MetricKey = std::pair<std::string, std::string>;
```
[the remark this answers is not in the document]
MBkkt: struct doesn't have compare (

arangod/Metrics/ClusterMetricsFeature.h
```diff
@@ -0,0 +1,99 @@
...
+  using CoordinatorMetrics = std::map<MetricKey, MetricValue>;
+  // We want map because promtool stupid format
```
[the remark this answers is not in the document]
MBkkt: Yeah you right

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+          if (!needRetry) {
+            _lastUpdate = std::chrono::steady_clock::now();
+            // We want more time than kTimeout should expire after last cross
+            // cluster communication
+            set(std::move(*metrics));
+            needRetry = _count.fetch_sub(1, std::memory_order_acq_rel) != 1;
+            // if true then was call asyncUpdate() after _count.store(1)
+          }
+          if (needRetry) {
+            forceAsyncUpdate();
```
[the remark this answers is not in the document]
MBkkt: Not recursion!

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -580,46 +581,20 @@ void MetricsFeature::validateOptions(std::shared_ptr<options::ProgramOptions>) {
...
   // minimize reallocs
   result.reserve(32768);
 
   // QueryRegistryFeature
   auto& q = server().getFeature<QueryRegistryFeature>();
   q.updateMetrics();
   {
-    bool changed = false;
-
     std::lock_guard guard(_lock);
```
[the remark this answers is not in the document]
MBkkt: I think you right, for example what if two different coordinator goes to metrics feature in the same time

Its old code, but we should fix it

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -1351,6 +1351,41 @@ futures::Future<OperationResult> countOnCoordinator(
...
+
+  std::vector<Future<network::Response>> futures;
+  futures.reserve(serverIds.size());
+  for (auto const& id : serverIds) {
+    network::Headers headers;
+    headers.emplace(StaticStrings::Accept,
+                    StaticStrings::MimeTypeJsonNoEncoding);
+    futures.push_back(network::sendRequest(
+        pool, "server:" + id, fuerte::RestVerb::Get, "/_admin/metrics/v2", {},
+        network::RequestOptions{}.param("json", "y").param("coordinator", "y"),
```
jsteemann, line 1371: Can we turn the "y" into "true" here? That would be consistent with our other APIs that handle boolean inputs.

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -1351,6 +1351,41 @@ futures::Future<OperationResult> countOnCoordinator(
...
+        headers));
```
jsteemann, line 1372:
```suggestion
        std::move(headers)));
```

arangod/IResearch/IResearchLink.cpp
```diff
@@ -2379,6 +2353,35 @@ void IResearchLink::LinkStats::toPrometheus(std::string& result,       //
...
+  std::string /*TODO(MBkkt) Fix cluster info interface*/ shardId{
+      labels.substr(start + 8, labels.size() - (start + 8) - 1)};
+  auto r =
+      server.getFeature<ClusterFeature>().clusterInfo().getResponsibleServer(
+          shardId);
+  if (r->empty()) {
+    return;  // TODO(MBkkt) We should fix cluster info :(
+  }
+  if ((*r)[0] != ServerState::instance()->getId()) {
+    return;  // We not in leader shard, so we just return
```
jsteemann, line 2369:
```suggestion
    return;  // We are not in leader shard, so we just return
```

arangod/IResearch/IResearchLink.cpp
```diff
@@ -2379,6 +2353,35 @@ void IResearchLink::LinkStats::toPrometheus(std::string& result,       //
   _needName = false;
 }
 
+void IResearchLink::LinkStats::toVPack(
+    application_features::ApplicationServer& server,
+    velocypack::Builder& builder, std::string_view labels) const {
+  auto const start = labels.find(",shard=\"");
```
jsteemann, line 2359: If for whatever reason the `find(",shard=\"");` returns `std::string_view::npos`, then the following code can fail.
We should add a check here that `start != std::string_view::npos`!

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +83,77 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+        auto& cell = std::get<uint64_t>(
+            metrics[metrics::ClusterMetricsFeature::MetricKey{
+                name, labels.stringView()}]);
+        cell += value.getUIntUnchecked();
+      };
+  metrics::ClusterMetricsFeature::ToPrometheus batchToPrometheus{
+      [](std::string& result, std::string_view name) {
+        result.append("# HELP ").append(name);
+        result.push_back(' ');
+        result.append("NO HELP");  // TODO(MBkkt) Should be help string
```
jsteemann, line 107: Will this be addressed in a follow-up PR?

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +83,77 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+  metric.add(arangosearch_index_size::kName, batchToCoordinator,
+             batchToPrometheus);
+  std::function gaugeToCoordinator =
+      [](metrics::ClusterMetricsFeature::CoordinatorMetrics& metrics,
+         std::string_view name, velocypack::Slice labels,
+         velocypack::Slice value) {
+        auto labelsStr = labels.stringView();
+        auto& cell = std::get<uint64_t>(
+            metrics[metrics::ClusterMetricsFeature::MetricKey{
+                name, labelsStr.substr(0, labelsStr.find(",shard=\""))}]);
```
jsteemann, line 144: If `labelsStr` does not contain `",shard=\""`, then this will fail. We should guard against that.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+          }
+        });  // detach() here
+  };
+  auto now = std::chrono::steady_clock::now();
+  if (now < _lastUpdate + kTimeout) {
```
jsteemann, line 83: As far as I can see, `forceAsyncUpdate()` is private, but it can be called at any time by the public method `asyncUpdate()`. So it seems that there can be concurrent calls to `forceAsyncUpdate()`.
If that's true, then `_lastUpdate` and `_handle` should be protected by some synchronization, because they are otherwise not thread-safe.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+  auto data = std::make_shared<Data>(std::move(metrics));
+  std::atomic_store_explicit(&_data, data, std::memory_order_release);
+  // TODO(MBkkt) serialize to velocypack array
+  // TODO(MBkkt) set to agency
+}
+
+void ClusterMetricsFeature::add(std::string_view metric,
+                                ToCoordinator const& toCoordinator) {
```
jsteemann, line 124: Does it make sense to accept the `ToCoordinator` parameter by value here, and move it into `_toCoordinator[metric]` below?

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+  std::lock_guard lock{_m};
+  _toCoordinator[metric] = toCoordinator;
+}
+
+void ClusterMetricsFeature::add(std::string_view metric,
+                                ToCoordinator const& toCoordinator,
+                                ToPrometheus const& toPrometheus) {
```
jsteemann, line 131: Same here: should these parameters be accepted by value and moved to the target location below?

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -677,10 +652,73 @@ void MetricsFeature::toPrometheus(std::string& result, bool v2) const {
...
+        "arangosearch_num_failed_consolidations",
+        "arangosearch_commit_time",
+        "arangosearch_cleanup_time",
+        "arangosearch_consolidation_time",
+    });
+
+void MetricsFeature::toVPack(VPackBuilder& builder, bool coordinator) const {
+  builder.openArray(true);
+  {
+    std::lock_guard guard(_lock);
```
jsteemann, line 678: As far as I can see we need the exclusive lock only during the call to `unsafeInitTo()`, but could then go on with a shared lock to serialize the metrics.

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +83,77 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
   _sparse = true;   // always sparse
 }
 
+Result IResearchLinkCoordinator::init(
+    velocypack::Slice definition,
+    IResearchLink::InitCallback const& initCallback) {
+  auto r = IResearchLink::init(definition, initCallback);
```
Dronplane, line 89: Do we really need to continue initialization in case of  `IResearchLink::init` failure ?

arangod/Metrics/MetricsFeature.cpp
```diff
@@ -677,10 +652,73 @@ void MetricsFeature::toPrometheus(std::string& result, bool v2) const {
...
+          (coordinator && !kCoordinatorMetrics.count(name))) {
+        continue;
+      }
+      i.second->toVPack(server(), builder);
+    }
+  }
+  builder.close();
+}
+
+void MetricsFeature::unsafeInitTo() const {
```
jsteemann, line 692: This function seems to be some sort of an init function, which will initialize some internal state only once after server startup. Maybe we can call this once in `MetricsFeature::start()` or so, to avoid having to use the exclusive `_lock` every time we are serializing the metrics.

arangod/Metrics/Parse.h
```diff
@@ -0,0 +1,37 @@
...
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Valery Mironov
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <string>
```
jsteemann, line 26: nit: this include seems unnecessary.

arangod/RestHandler/RestMetricsHandler.cpp
```diff
@@ -148,23 +131,32 @@ RestStatus RestMetricsHandler::execute() {
...
   }
 
-  std::vector<std::string> const& suffixes = _request->suffixes();
-
-  bool v2 = false;
-  if (suffixes.size() > 0 && suffixes[0] == "v2") {
-    v2 = true;
+  auto const& suffixes = _request->suffixes();
+  auto const& values = _request->values();
+  bool const v2 = !suffixes.empty() && suffixes[0] == "v2";
```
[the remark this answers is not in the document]
jsteemann: I think in 3.10 the primary endpoint should be `/_admin/metrics`. 
`/_admin/metrics/v2` should also be supported and return the same result as `/_admin/metrics`. But the v2 endpoint should be deprecated.
The very old (pre-3.8) metrics format should not be supported anymore by 3.10 IMHO.

arangod/RestHandler/RestMetricsHandler.cpp
```diff
@@ -148,23 +131,32 @@ RestStatus RestMetricsHandler::execute() {
...
+  if (values.contains("json")) {
+    TRI_ASSERT(v2);
+    bool const coordinator = values.contains("coordinator");
```
jsteemann, line 144: This only checks if `json` / `coordinator` are used as URL parameters, but does not check their values.
I think we should here check their boolean values, so that one can call the API with `json=true` or `json=false`. We can use `basics::StringUtils::boolean(...)` for parsing the parameter value. That would be consistent with most other APIs then.

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+/// @brief test suite
+////////////////////////////////////////////////////////////////////////////////
+
+function CoordinatorMetricsTestSuite() {
+    return {
+        setUpAll: function () {
+            db._useDatabase("_system");
+            db._create("foo1", {"replication": 3, "numberOfShards": 9});
+            db._create("foo2", {"replication": 2, "numberOfShards": 12});
+            db._create("foo3", {"replication": 1, "numberOfShards": 1});
```
jsteemann, line 45:
```suggestion
            db._create("foo1", {"replicationFactor": 3, "numberOfShards": 9});
            db._create("foo2", {"replicationFactor": 2, "numberOfShards": 12});
            db._create("foo3", {"replicationFactor": 1, "numberOfShards": 1});
```

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+const db = arangodb.db;
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief test suite
+////////////////////////////////////////////////////////////////////////////////
+
+function CoordinatorMetricsTestSuite() {
+    return {
+        setUpAll: function () {
+            db._useDatabase("_system");
```
jsteemann, line 42: It shouldn't be necessary to explicitly move into the `_system` database here.
```suggestion
```

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+        testIResearchLinkStats: function () {
+            let coordinators = getEndpointsByType("coordinator");
+            let c = coordinators[0];
+            {// make metrics actual
+                getMetricRaw(c);
+                require('internal').sleep(2);
+            }
+            print();
```
jsteemann, line 85: JS tests normally shouldn't print.
```suggestion
```

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,129 @@
...
+/// @brief CoordinatorMetricsTestSuite tests
+////////////////////////////////////////////////////////////////////////////////
+
+        testIResearchLinkStats: function () {
+            let coordinators = getEndpointsByType("coordinator");
+            let c = coordinators[0];
+            {// make metrics actual
+                getMetricRaw(c);
+                require('internal').sleep(2);
+            }
```
jsteemann, line 84: Is this necessary? A few lines below the code calls `getMetricRaw(c)` anyway.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+  if (isEnabled() && _count.fetch_add(1, std::memory_order_acq_rel) == 0) {
+    forceAsyncUpdate();
+  }
+}
+
+void ClusterMetricsFeature::forceAsyncUpdate() {
+  auto internal = [this] {
+    _count.store(1, std::memory_order_relaxed);
+    // TODO try get from agency
+    metricsOnCoordinator(server().getFeature<NetworkFeature>(),
```
Dronplane, line 65: `metricsOnCoordinator` is not noexcept.  We may miss fetch_sub call and no other `asyncUpdate`  call will schedule new metrics update.  We could use some "scopeGuard" to ensure proper restart in the worst case.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+                         server().getFeature<ClusterFeature>())
+        .then([this](futures::Try<RawDBServers>&& metrics) mutable {
+          bool needRetry = !metrics.hasValue();
+          if (!needRetry) {
+            _lastUpdate = std::chrono::steady_clock::now();
+            // We want more time than kTimeout should expire after last cross
+            // cluster communication
+            set(std::move(*metrics));
```
Dronplane, line 73: Also `set` here may throw VPack error in case on inconsistent slice received. And we will miss fetch_sub

tests/js/client/http/coordinator-metrics-cluster.js
MBkkt: May be faster, but yes.
getMetricRaw don't blocking wait answer for cluster metrics, it returns last valid metric and can async call to collect new values

arangod/RestHandler/RestMetricsHandler.cpp
MBkkt: Ok

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: I think not?

arangod/Metrics/MetricsFeature.cpp
MBkkt: Maybe, I just move some code to function

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: No, this is std::function, maybe i should use fu2::unique_function?

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: forceAsyncUpdate cannot running parallel

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: We know that it contains...

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: Yes, of course

Its about pr with one source of true for metrics

arangod/IResearch/IResearchLink.cpp
MBkkt: I can add tri_assert

arangod/Cluster/ClusterMethods.cpp
MBkkt: Ok, for consistency

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: I think if we have same format on coordinator and dbserver slice should be valid?
But i will check it

Dronplane: I mean in general slice "should" be always  valid.  But for some buggy reason this assumption could be violated. And we must be able to gracefully handle that.

arangod/IResearch/IResearchLinkCoordinator.cpp
Dronplane: But why? We will return failure anyway.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
+
+void ClusterMetricsFeature::asyncUpdate() {
+  if (isEnabled() && _count.fetch_add(1, std::memory_order_acq_rel) == 0) {
+    forceAsyncUpdate();
+  }
+}
+
+void ClusterMetricsFeature::forceAsyncUpdate() {
```
Dronplane, line 61: NIT: name `forceAsyncUpdate` is a bit misleading. It does not "forces"  immediate metrics update, but just schedules an update. Suggest to rename to something like `scheduleAsyncUpdate`

Dronplane:
As the ClusterMetricsFeature schedules delayed execution of metrics update, it should somewhere cancel the queued task in case of shutdown. The good place would be:
```
// notify the feature about a shutdown request
 virtual void beginShutdown();
```

arangod/RestHandler/RestMetricsHandler.cpp
MBkkt: image

```C++
  if (str.empty()) {
    return false;
  }
  std::string lower = trim(str);
  tolowerInPlace(lower);

  if (lower == "true" || lower == "yes" || lower == "on" || lower == "y" ||
      lower == "1" || lower == "✓") {
    return true;
  }
  return false;
```

arangod/IResearch/IResearchLink.cpp
```diff
@@ -2379,6 +2353,35 @@ void IResearchLink::LinkStats::toPrometheus(std::string& result,       //
...
+  std::string /*TODO(MBkkt) Fix cluster info interface*/ shardId{
```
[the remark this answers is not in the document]
MBkkt: Yes

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -1351,6 +1351,41 @@ futures::Future<OperationResult> countOnCoordinator(
...
+  return collectAll(futures).thenValue(
+      [](std::vector<Try<network::Response>>&& responses) {
+        metrics::RawDBServers metrics;
+        metrics.reserve(responses.size());
+        for (auto& response : responses) {
+          if (!response.hasValue() || response->fail() ||
+              !response->hasResponse()) {
+            continue;  // Shit happens, just ignore it
+          }
+          metrics.push_back(response->response().stealPayload());
```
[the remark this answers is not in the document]
MBkkt: Will be called same ctors, so no reason use emplace back

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "Cluster/ClusterMethods.h"
+#include "Metrics/MetricsFeature.h"
+#include "Network/NetworkFeature.h"
+#include "Scheduler/SchedulerFeature.h"
+
+namespace arangodb::metrics {
+namespace {
+
+using namespace std::chrono_literals;
+constexpr std::chrono::steady_clock::duration kTimeout = 15s;
```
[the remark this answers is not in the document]
MBkkt: Done?

arangod/ClusterEngine/ClusterSelectivityEstimates.cpp
MBkkt: Done

arangod/Cluster/ClusterMethods.cpp
MBkkt: Done

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: tri_assert was added

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: Fixed

tests/js/client/http/coordinator-metrics-cluster.js
MBkkt: Done, thanks!

tests/js/client/http/coordinator-metrics-cluster.js
MBkkt: Done

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: Done

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: Maybe done, `scheduleUpdate`

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: Fixed

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+void ClusterMetricsFeature::set(RawDBServers&& raw) const {
+  Metrics metrics;
+  std::shared_lock lock{_m};
+  for (auto const& payload : raw) {
+    TRI_ASSERT(payload);
+    velocypack::Slice slice{payload->data()};
+    auto const size = slice.length();
+    TRI_ASSERT(size % 3 == 0);
```
gnusi, line 136: * Check slice type
* Turn assertion into real check of array size

arangod/RestHandler/RestMetricsHandler.cpp
```diff
@@ -87,32 +81,19 @@ RestStatus RestMetricsHandler::execute() {
...
-      for (auto const& srv : ci.getServers()) {
-        // validate if server id exists
-        if (srv.first == serverId) {
-          found = true;
-          break;
-        }
-      }
-
-      if (!found) {
+      if (auto servers = ci.getServers(); !servers.contains(serverId)) {
```
gnusi, line 90: maybe add helper method to ClusterInfo to avoid copying a map?

arangod/RestHandler/RestMetricsHandler.cpp
```diff
@@ -134,6 +115,7 @@ RestStatus RestMetricsHandler::execute() {
             } else {
               // the response will not contain any velocypack.
               // we need to forward the request with content-type text/plain.
+              TRI_ASSERT(r.hasResponse());
```
gnusi, line 118: can we turn it into the real check?

arangod/RestHandler/RestMetricsHandler.cpp
```diff
@@ -148,16 +130,27 @@ RestStatus RestMetricsHandler::execute() {
...
   }
 
-  std::string result;
-  metrics.toPrometheus(result);
-  _response->setResponseCode(rest::ResponseCode::OK);
-  _response->setContentType(rest::ContentType::TEXT);
-  _response->addRawPayload(result);
-
+  auto const& values = _request->values();
+  auto it = values.find("json");
```
gnusi, line 139: TBH I'd be in favor of using `type`instead of just a marker so that we can easily extend it in future

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,128 @@
...
+        foo1_values.push({"name": i});
+        foo2_values.push({"name": i + 100000});
+        foo3_values.push({"name": i - 100000});
+      }
+      db.foo1.save(foo1_values);
+      db.foo2.save(foo2_values);
+      db.foo3.save(foo3_values);
+      db._createView("foov", "arangosearch", {
+        "links": {
+          "foo1": {"includeAllFields": true},
```
gnusi, line 58: I suggest we turn consolidation off so that results are reliable

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,192 @@
...
+  startsAfter<NetworkFeature>();
+  startsAfter<SchedulerFeature>();
+}
+
+void ClusterMetricsFeature::collectOptions(
+    std::shared_ptr<options::ProgramOptions> options) {
+  options
+      ->addOption(
+          "--server.cluster-metrics-timeout",
+          "Seconds that will be between cross cluster calls that collect "
```
gnusi, line 60: maybe something like `Cluster metrics polling timeout in seconds`?

tests/js/client/http/coordinator-metrics-cluster.js
```diff
@@ -0,0 +1,128 @@
...
+      assertEqual(metrics["arangodb_search_num_files"]["foo3"], 6);
+
+      assertTrue(metrics["arangodb_search_index_size"]["foo1"] > 0);
+      assertTrue(metrics["arangodb_search_index_size"]["foo2"] > 0);
+      assertTrue(metrics["arangodb_search_index_size"]["foo3"] > 0);
+    }
+  };
+}
+
+////////////////////////////////////////////////////////////////////////////////
```
gnusi, line 122: can we also add a test checking json format?

gnusi (changes requested):

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +84,61 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+  auto& metric = IResearchLink::collection()
+                     .vocbase()
+                     .server()
+                     .getFeature<ClusterMetricsFeature>();
+
+  auto batchToCoordinator = [](ClusterMetricsFeature::Metrics& metrics,
+                               std::string_view name, velocypack::Slice labels,
+                               velocypack::Slice value) {
+    ClusterMetricsFeature::MetricKey key{name, labels.stringView()};
+    std::get<uint64_t>(metrics[key]) += value.getNumber<uint64_t>();
```
jsteemann, line 104: Just wondering: can there ever be any metrics with non-unsigned integer values, e.g. signed negative values, real values etc...? Or is that not possible by definition?
I know that at least in some parts of the code we have metrics that are floats internally and something like `metrics::Histogram<metrics::LogScale<float>>&`.

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +84,61 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+  };
+  auto batchToPrometheus = [](std::string& result, std::string_view globals,
+                              ClusterMetricsFeature::MetricKey const& key,
+                              ClusterMetricsFeature::MetricValue const& value) {
+    Metric::addMark(result, key.first, globals, key.second);
+    result.append(std::to_string(std::get<uint64_t>(value))) += '\n';
```
jsteemann, line 110: Same here

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: > Turn assertion into real check of array size

Do we really want this? I don't like unreachable code

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +84,61 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+  metric.add("arangodb_search_num_files", batchToCoordinator,
+             batchToPrometheus);
+  metric.add("arangodb_search_index_size", batchToCoordinator,
+             batchToPrometheus);
+  auto gaugeToCoordinator = [](ClusterMetricsFeature::Metrics& metrics,
+                               std::string_view name, velocypack::Slice labels,
+                               velocypack::Slice value) {
+    auto labelsStr = labels.stringView();
+    auto end = labelsStr.find(",shard=\"");
+    TRI_ASSERT(end != std::string_view::npos);
```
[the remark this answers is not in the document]
MBkkt: > Turn assertion into real check of array size

Do we really want this? I don't like unreachable code

arangod/Metrics/ClusterMetricsFeature.cpp
MBkkt: Ok

arangod/IResearch/IResearchLinkCoordinator.cpp
```diff
@@ -83,6 +84,61 @@ IResearchLinkCoordinator::IResearchLinkCoordinator(
...
+    labelsStr = labelsStr.substr(0, end);
+    metrics::ClusterMetricsFeature::MetricKey key{name, labelsStr};
+    std::get<uint64_t>(metrics[key]) += value.getNumber<uint64_t>();
+  };
+  metric.add("arangodb_search_num_failed_commits", gaugeToCoordinator);
```
jsteemann, line 131: `gaugeToCoordinator` sounds as if the metrics used here and in the following lines would all be gauges. 
But I think most of them are actually counters, not gauges.
I know it is just a name of an internal lambda, but I found the name "gauge" somewhat misleading here.

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: This code works for uint64_t on the db server-side.
Also now ClusterMetrics don't works with histogram

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -0,0 +1,192 @@
...
+void ClusterMetricsFeature::scheduleUpdate() noexcept {
+  auto now = std::chrono::steady_clock::now();
+  auto next = _lastUpdate + std::chrono::seconds{_timeout};
+  _handle = SchedulerFeature::SCHEDULER->queueDelayed(
+      RequestLane::CLUSTER_INTERNAL, (now < next ? next - now : 0ns),
+      [this](bool canceled) {
+        if (canceled) {
+          return;
+        }
+        _count.store(1, std::memory_order_relaxed);
```
jsteemann, line 93: Does this store have to be synced with the fetch_add() in line 79?

arangod/IResearch/IResearchMetricStats.h
```diff
@@ -0,0 +1,115 @@
...
+      "arangodb_search_num_files",      //
+      "arangodb_search_index_size",     //
+  };
+
+  // toVPack
+  static bool skip(ArangodServer& server, std::string_view labels) {
+    auto const start = labels.find(kShard);
+    TRI_ASSERT(start != std::string_view::npos);
+    std::string /*TODO(MBkkt) Fix cluster info interface*/ shardId{
+        labels.substr(start + 8, labels.size() - (start + 8) - 1)};
```
jsteemann, line 52: The +8 here look really magic. Is that `kShard.size()`? If so, we should write it as such, to improve understandability.

arangod/Metrics/ClusterMetricsFeature.cpp
Dronplane: Not sure. But looks like this line could cancel the effect of `beginShutdown` if timing is unlucky.

arangod/IResearch/IResearchLinkCoordinator.cpp
MBkkt: But I suppose not in this PR (These metrics already exist as gauge)

arangod/IResearch/IResearchLink.cpp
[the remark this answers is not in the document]
MBkkt: https://github.com/arangodb/arangodb/pull/15782

tests/js/client/http/coordinator-metrics-cluster.js
MBkkt: - [ ] Add json test

arangod/Metrics/ClusterMetricsFeature.h
MBkkt: Maybe later

MBkkt:
https://vpna.arangodb.biz:3456/view/PR/job/arangodb-matrix-pr/20119/

gnusi (approved):

PR #15487, opened 2022-01-10
BTS-718: potential fix for assertion failure

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15488

Potential fix for assertion failure during testing in https://arangodb.atlassian.net/browse/BTS-718.
This is a test-only bugfix, thus it intentionally does not contain a CHANGELOG entry.

### Testing & Verification


mpoeter (approved):
LGTM! Just one minor suggestions.

arangod/RocksDBEngine/RocksDBMetadata.h
```diff
@@ -74,6 +77,28 @@ struct RocksDBMetadata final {
...
+    return _tainted;
+#else
+    // if we don't use failure tests, the document counts are never tainted.
+    return false;
+#endif
+  }
+#else
+  // non-maintainer mode...
+  static constexpr bool tainted() const noexcept { return false; }
+#endif
```
mpoeter, line 100: This could be simplified:
```suggestion
#if defined(ARANGODB_ENABLE_FAILURE_TESTS) && defined(ARANGODB_ENABLE_MAINTAINER_MODE)
  // if we use failure tests, the document counts may have been intentionally
  // corrupted. the tainted state is set by the failure points that corrupt
  // the counters.
  bool tainted() const noexcept { return _tainted; }
#else
  // non-maintainer mode or failure tests not enabled.
  static constexpr bool tainted() const noexcept { return false; }
#endif
```

mpoeter (approved):
LGTM! Just one minor suggestion.

PR #15489, opened 2022-01-10
[3.9] [BTS-705] Add more decay tests

mpoeter:
### Scope & Purpose

Add more tests for the DECAY_* functions.

#### Backports:

- [x] Backport for devel: #15490

#### Related Information

- [X] Jira ticket number: https://arangodb.atlassian.net/browse/BTS-705

### Testing & Verification

### DIFF
--- a/tests/js/server/aql/aql-functions-numeric.js
+++ b/tests/js/server/aql/aql-functions-numeric.js
@@ -28,6 +28,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 var internal = require("internal");
+var arangodb = require('@arangodb');
 var errors = internal.errors;
 var jsunity = require("jsunity");
 var helper = require("@arangodb/aql-helper");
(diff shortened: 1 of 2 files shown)


tests/js/server/aql/aql-functions-numeric.js
```diff
@@ -5767,26 +5768,18 @@ function ahuacatlNumericFunctionsTestSuite () {
...
 
     testDecayGauss : function () {
-        var data = [
+        [
           [1, 0, 10, 0, 0.2, 0.9840344433634576],
           [-10, 40, 5, 0, 0.1, 1.0000000000000458e-100],
           [41, 40, 5, 5, 0.7, 1]
-        ];
+        ].forEach(function (value) {
+          const [arg, origin, scale, offset, decay] = value;
```
alexbakharew, line 5776: May be also unpack expected value? Therefore we can avoid using indexes in "value" variable while comparing real and expected values.

alexbakharew (changes requested):

tests/js/server/aql/aql-functions-numeric.js
mpoeter: Good point!

PR #15518, opened 2022-01-13
Added unittest for complex queries

cpjulia:
### Scope & Purpose

This is a PR just for the unittest to run for QA-208, so only the added js file is important to observe here.


KVS85 (approved):
LGTM.

PR #15519, opened 2022-01-13
Added test for complex queries

cpjulia:
### Scope & Purpose

This PR is only for adding a test for QA-208, so other files can be ignored.


jsteemann (approved):
LGTM!

PR #15531, opened 2022-01-14
Refuse to concat a char.

maierlars:
### Scope & Purpose

If you write something like
```cpp
basics::StringUtils::concatT(curServersKnown, '/', serverId);
```
it **does** not what you think it does. The reason is, that for a `char` type, `concatT` will (like for all numerical data types) call `std::to_string`. Hence the above results in something like
```cpp
"/Current/ServersKnown45PRMR-XXXXX"
```
I had to debug that and I never want to do it again. After this change `concatT` refuses to compile using a `char` as parameter.
Same applies to `joinT`.

Instead use
```cpp
basics::StringUtils::concatT(curServersKnown, "/", serverId);
```
or
```cpp
basics::StringUtils::concatT("The value is: ", static_cast<int>(myCharVariable));
```

### DIFF
--- a/arangod/Replication2/ReplicatedLog/LogCommon.cpp
+++ b/arangod/Replication2/ReplicatedLog/LogCommon.cpp
@@ -332,7 +332,7 @@ auto replication2::intersect(LogRange a, LogRange b) noexcept -> LogRange {
 }
 
 auto replication2::to_string(LogRange const& r) -> std::string {
-  return basics::StringUtils::concatT('[', r.from, ", ", r.to, ')');
+  return basics::StringUtils::concatT("[", r.from, ", ", r.to, ")");
 }
 
 auto LogRange::end() const noexcept -> LogRange::Iterator {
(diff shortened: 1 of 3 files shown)


markuspf (approved):
LGTM, modulo comment.

arangod/Replication2/ReplicatedLog/LogCommon.cpp
```diff
@@ -332,7 +332,7 @@ auto replication2::intersect(LogRange a, LogRange b) noexcept -> LogRange {
 }
 
 auto replication2::to_string(LogRange const& r) -> std::string {
-  return basics::StringUtils::concatT('[', r.from, ", ", r.to, ')');
+  return basics::StringUtils::concatT("[", r.from, ", ", r.to, ")");
```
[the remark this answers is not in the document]
maierlars: yes, it's a half-open interval. like in math, you know?

tests/Basics/StringUtilsTest.cpp
```diff
@@ -436,6 +436,7 @@ TEST_F(StringUtilsTest, concatT) {
   EXPECT_EQ("42", StringUtils::concatT(42));
   EXPECT_EQ("2.500000", StringUtils::concatT(2.5));
   EXPECT_EQ("hello, world", StringUtils::concatT("hello", ", ", "world"));
+  EXPECT_EQ("45", StringUtils::concatT(static_cast<int>('/')));
```
apetenchea, line 439: The value of '/' is 47. This test is going to fail :(

apetenchea, line 439:
```suggestion
  EXPECT_EQ("47", StringUtils::concatT(static_cast<int>('/')));
```

apetenchea (changes requested):

PR #15548, opened 2022-01-18
[CINFRA-220] No-Op state machine

maierlars:
### Scope & Purpose

This adds a black-hole (noop) state machine. This is only for testing and will be replaced by a more complex state machine later. This allows independent parts of the replicated state framework to be tested, without waiting for complex state machine.


apetenchea (approved):

maierlars:
dump_no_envelope is completely unreleated.

PR #15550, opened 2022-01-18
Fixed python imports to work with 3.10 (#15502)

dothebart:
backport of https://github.com/arangodb/arangodb/pull/15502 and https://github.com/arangodb/arangodb/pull/15552

PR #15552, opened 2022-01-18
which fuckup once more in V8-gyp files

dothebart:
### Scope & Purpose
After https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1814372.html  https://www.python.org/dev/peps/pep-0632/ bites us in the same place. https://stackoverflow.com/questions/9877462/is-there-a-python-equivalent-to-the-which-command sugests, that `shutil` is the way to go.

Backports:
- 3.7 https://github.com/arangodb/arangodb/pull/15551
- 3.8 https://github.com/arangodb/arangodb/pull/15550
- 3.9 https://github.com/arangodb/arangodb/pull/15549


goedderz (approved):

MBkkt (approved):
Works for me, LGTM)

jsteemann (approved):

PR #15555, opened 2022-01-18
Various refactorings

jsteemann:
### Scope & Purpose

This PR refactors a few components:
* remove unused `_keyOptions` member in LogicalCollection
* clean up ExpressionContext a bit
* remove basics::StringBuffer from several places and use std::string instead, e.g.
  * transaction::Context::leaseStringBuffer()
  * transaction::Context::returnStringBuffer()
  * several RestHandlers
  * Aql/Functions.cpp
* adding operator* to transaction::BuilderLeaser and transaction::StringLeaser
* remove `isDataFromCollection()` function from ExpressionContext. the previous `isDataFromCollection` flag in Aql/Variable.cpp is now renamed to `isFullDocumentFromCollection` and set for a few more ExecutionNode types than before
* rename a few constructors for AqlValue to use more appropriate names
* add Function::canUseInAnalyzer enum value, so that functions can be marked as usable/unusable inside arangosearch analyzers. this allows removal an extra list of forbidden AQL functions from the analyzer code
* make methods in Expression static so that they only rely on their input values and cannot modify the Expression object. this is a refactoring for making Expression more thread-safe to use in later use cases
* move standalone Expression execution from analyzer code (arangod/IResearch/IResearchAqlAnalyzer.cpp) to arangod/Aql/StandaloneCalculation.cpp. the moved code will likely be called from other components too and will not be iresearch-specific anymore.

This will not be backported.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**


arangod/Aql/FixedVarExpressionContext.cpp
```diff
@@ -31,22 +31,28 @@
...
+  // first check temporary variables in QueryExpressionContext.
+  if (!_variables.empty()) {
+    auto it = _variables.find(variable);
+
+    if (it != _variables.end()) {
+      // copy the slice we found
+      mustDestroy = true;
+      return AqlValue((*it).second);
+    }
+  }
```
gnusi, line 46: that's 4th occurrence of this code :-) better to move it to a separate protected member of QueryExpressionContext. That will allow us to make `_variables` member private.

arangod/Aql/FixedVarExpressionContext.cpp
```diff
@@ -82,36 +89,13 @@ FixedVarExpressionContext::FixedVarExpressionContext(
...
+    : QueryExpressionContext(trx, context, cache) {}
 
-void SingleVarExpressionContext::setVariableValue(Variable* variable,
-                                                  AqlValue& value) {
-  _variable = variable;
-  _value = value;
+AqlValue SingleVarExpressionContext::getVariableValue(
+    Variable const* /*variable*/, bool /*doCopy*/,
+    bool& /*mustDestroy*/) const {
+  return AqlValue(AqlValueHintNull());
```
gnusi, line 100: why always `Null` for `SingleVarExpressionContext`?

arangod/Aql/FixedVarExpressionContext.h
```diff
@@ -57,13 +55,13 @@ class FixedVarExpressionContext final : public QueryExpressionContext {
...
-  void clearVariableValue(Variable const*);
+  void clearVariableValue(Variable const*) noexcept;
 
   void serializeAllVariables(velocypack::Options const& opts,
                              arangodb::velocypack::Builder&) const;
 
  private:
-  /// @brief temporary storage for expression data context
+  // temporary storage for expression data context
   std::unordered_map<Variable const*, AqlValue> _vars;
```
gnusi, line 65: `std::unordered_map<Variable const*, AqlValue>` -> `FlatHashMap<Variable const*, AqlValue>`

arangod/Aql/Function.h
```diff
@@ -76,14 +76,17 @@ struct Function {
     /// @brief whether or not the function may read documents from the database
     CanReadDocuments = 16,
 
+    /// @brief whether or not the function is usable in an analyzer
+    CanUseInAnalyzer = 32,
```
gnusi, line 80: I hope we don't persist this enum value

arangod/Aql/InAndOutRowExpressionContext.cpp
```diff
@@ -69,32 +69,27 @@ void InAndOutRowExpressionContext::invalidateInputRow() {
...
+  if (!_variables.empty()) {
+    TRI_ASSERT(_variables.empty());
+    auto it = _variables.find(variable);
+
+    if (it != _variables.end()) {
+      // copy the slice we found
+      mustDestroy = true;
+      return AqlValue((*it).second);
+    }
+  }
```
gnusi, line 85: 5th occurence, see comments about moving it to protected `QueryExpressionContext` member above

arangod/Aql/OptimizerRules.cpp
```diff
@@ -2655,55 +2648,50 @@ void arangodb::aql::removeRedundantCalculationsRule(
 
       if (current->getType() == EN::CALCULATION) {
         try {
+          buffer.clear();
           // ExecutionNode::castTo<CalculationNode*>(current)->expression()->node()->dump(0);
```
gnusi, line 2652: remove?

arangod/Aql/SortNode.cpp
```diff
@@ -123,8 +122,8 @@ bool SortNode::simplify(ExecutionPlan* plan) {
 }
 
 /// @brief returns all sort information
-SortInformation SortNode::getSortInformation(
-    ExecutionPlan* plan, arangodb::basics::StringBuffer* buffer) const {
+SortInformation SortNode::getSortInformation(ExecutionPlan* plan,
+                                             std::string& buffer) const {
```
gnusi, line 126: `std::string&` -> `std::string&&`?

arangod/Aql/SortNode.cpp
```diff
@@ -159,10 +158,10 @@ SortInformation SortNode::getSortInformation(
...
         return result;
       }
-      result.criteria.emplace_back(std::make_tuple(
-          const_cast<ExecutionNode const*>(setter),
-          std::string(buffer->c_str(), buffer->length()), (*it).ascending));
-      buffer->reset();
+      result.criteria.emplace_back(
+          std::make_tuple(const_cast<ExecutionNode const*>(setter),
+                          std::move(buffer), (*it).ascending));
+      buffer.clear();
```
gnusi, line 164: seems to be redundant given `std::move(buffer)` above

gnusi (changes requested):

arangod/Aql/Function.h
jsteemann: No, these enum bits are only available at runtime and never persisted.

arangod/Aql/FixedVarExpressionContext.cpp
jsteemann: As far as I can see, the previous code here was like this:
``` 
AqlValue SingleVarExpressionContext::getVariableValue(Variable const* var, bool,
                                                      bool&) const {
  if (var == _variable) {
    return _value;
  } else {
    return AqlValue(AqlValueHintNull());
  }
}
```
So `_value` was return if `var == _variable`. 
`_variable` however is only initialized in the ctor of `SingleVariableExpressionContext` and in its `setVariableValue()` method.
There were 2 ctors for `SingleVariableExpressionContext` before, and only one of them was used. That ctor initialized `_variable` to nullptr. And there were no callers to `setVariableValue`.
So basically the previous `SingleVariableExpressionContext` was effectively a `NoVariableExpressionContext`.
It is only used from AqlAnalyzers though. I can't tell if this is a bug, or was just missing a cleanup.

gnusi: @jsteemann Ok, that looks like a bug, but let's ask the author himself :-)

gnusi: @alexbakharew any thoughts ^^^?

arangod/Aql/ExecutionPlan.cpp
```diff
@@ -2623,19 +2623,33 @@ bool ExecutionPlan::fullCount() const noexcept {
...
+    } else if (node->getType() == ExecutionNode::TRAVERSAL) {
+      TraversalNode const* tn =
+          ExecutionNode::castTo<TraversalNode const*>(node);
+      if (variable == tn->vertexOutVariable() ||
+          variable == tn->edgeOutVariable()) {
+        // traversal vertices and edges are always returned as full documents
+        variable->isFullDocumentFromCollection = true;
       }
     }
+    // TODO: maybe add more node types here, e.g. arangosearch nodes or
```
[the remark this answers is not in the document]
jsteemann: Happy to add ENUMERATE_VIEW here, but have a few questions:
does an ENUMERATE_VIEW node always produce full documents, i.e. unmodified documents as returned by the storage engine, without any rewrites/projections? If so, ENUMERATE_VIEW can be added here.
If an ENUMERATE_VIEW only produces certain attributes or somehow rewrites the documents after retrieving them from RocksDB, we cannot add it here. Any idea?

arangod/Aql/AqlValue.cpp
```diff
@@ -1544,15 +1545,13 @@ AqlValue::AqlValue(arangodb::velocypack::Buffer<uint8_t>&& buffer) {
...
+  setPointer<true>(v.slice.start());
   TRI_ASSERT(!VPackSlice(_data.pointerMeta.pointer).isExternal());
 }
 
-AqlValue::AqlValue(AqlValueHintCopy const& v) {
-  TRI_ASSERT(v.ptr != nullptr);
-  VPackSlice slice(v.ptr);
-  initFromSlice(slice, slice.byteSize());
+AqlValue::AqlValue(AqlValueHintSliceCopy v) {
+  initFromSlice(v.slice, v.slice.byteSize());
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/AqlValue.cpp
```diff
@@ -1512,12 +1513,12 @@ AqlValue::AqlValue(char const* value, size_t length) {
 AqlValue::AqlValue(std::string const& value)
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/AstNode.cpp
```diff
@@ -1924,81 +1917,83 @@ AstNode* AstNode::clone(Ast* ast) const { return ast->clone(this); }
 /// (only for objects that do not contain dynamic attributes)
 /// note that this may throw and that the caller is responsible for
 /// catching the error
-void AstNode::stringify(arangodb::basics::StringBuffer* buffer, bool verbose,
-                        bool failIfLong) const {
+void AstNode::stringify(std::string& buffer, bool failIfLong) const {
   // any arrays/objects with more values than this will not be stringified if
   // failIfLong is set to true!
-  static size_t const TooLongThreshold = 80;
+  constexpr size_t tooLongThreshold = 80;
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/DocumentExpressionContext.cpp
```diff
@@ -33,13 +34,21 @@ DocumentExpressionContext::DocumentExpressionContext(
...
                                                      bool& mustDestroy) const {
+  if (!_variables.empty()) {
+    auto it = _variables.find(variable);
+
+    if (it != _variables.end()) {
+      // copy the slice we found
+      mustDestroy = true;
+      return AqlValue((*it).second);
+    }
+  }
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/DocumentIndexExpressionContext.cpp
```diff
@@ -37,6 +38,15 @@ DocumentIndexExpressionContext::DocumentIndexExpressionContext(
...
     Variable const* variable, bool doCopy, bool& mustDestroy) const {
+  if (!_variables.empty()) {
+    auto it = _variables.find(variable);
+
+    if (it != _variables.end()) {
+      // copy the slice we found
+      mustDestroy = true;
+      return AqlValue((*it).second);
+    }
+  }
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/ExecutionPlan.cpp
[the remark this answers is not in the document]
jsteemann: Added ENUMERATE_IRESEARCH_VIEW for now, under the assumption that views always produce full documents as returned by the storage engine, and no projections etc.

arangod/Aql/ExecutorExpressionContext.cpp
```diff
@@ -41,12 +41,21 @@ ExecutorExpressionContext::ExecutorExpressionContext(
...
-  mustDestroy = false;
+  if (!_variables.empty()) {
+    auto it = _variables.find(variable);
+
+    if (it != _variables.end()) {
+      // copy the slice we found
+      mustDestroy = true;
+      return AqlValue((*it).second);
+    }
+  }
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/Expression.cpp
```diff
@@ -666,46 +664,43 @@ AqlValue Expression::executeSimpleExpressionObject(AstNode const* node,
...
 
+      buffer->clear();
       Functions::Stringify(&vopts, adapter, slice);
 
       if (mustCheckUniqueness) {
-        std::string key(buffer->begin(), buffer->length());
-
         // prevent duplicate keys from being used
-        auto it = keys.find(key);
+        auto it = keys.find(*buffer);
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Aql/FixedVarExpressionContext.h
jsteemann: Done

arangod/Aql/InAndOutRowExpressionContext.cpp
jsteemann: Done

arangod/Aql/SortNode.cpp
jsteemann: I simplified the API of this method even more. Turned out `plan` wasn't even used by the method. And it feels clumsy to fix in an std::string for temporary string buildup, so I removed that from the API and instead used a local std::string value inside the method.

arangod/Aql/SortNode.cpp
jsteemann: Also changed.

arangod/Aql/OptimizerRules.cpp
jsteemann: I think the `clear()` should stay in place here, because otherwise `buffer` is not guaranteed to be empty. It is populated one or two lines below by the call to `Expression::stringifyIfNotTooLong(...)`, but that only appends to the buffer and doesn't clear it initially.
And this stuff is executed in a loop. My conclusion is: clear() must still be there to keep the same behavior.

arangod/Aql/FixedVarExpressionContext.cpp
jsteemann: Done

arangod/Aql/OptimizerRules.cpp
gnusi: sorry I meant the comment 1 line below :-)
```
// ExecutionNode::castTo<CalculationNode*>(current)->expression()->node()->dump(0);
```

arangod/Aql/FixedVarExpressionContext.cpp
alexbakharew: Hi @jsteemann 

Everything is OK with AqlAnalyzers. We had discussion with @Dronplane about SingleVarExpressionContext. And its purpose is to store only one variable against few in FixedVarExpressionContext. 

In process of implementing optimizer rules, we understood that we actually don't want to store value in SingleVarExpressionContext, that's why it is nullptr at the moment. And now it is clear that your suggestion about adding NoVariableExpressionContext is very useful.

alexbakharew: But I think that we should replace SingleVarExpressionContext with NoVariableExpressionContext or don't modify getVariableValue() method. Because I think that it's very unclear for other developers why we always return nullptr

arangod/Aql/ExecutionPlan.cpp
```diff
@@ -2632,19 +2632,35 @@ bool ExecutionPlan::fullCount() const noexcept {
...
+    } else if (node->getType() == ExecutionNode::TRAVERSAL) {
+      TraversalNode const* tn =
+          ExecutionNode::castTo<TraversalNode const*>(node);
+      if (variable == tn->vertexOutVariable() ||
+          variable == tn->edgeOutVariable()) {
+        // traversal vertices and edges are always returned as full documents
+        variable->isFullDocumentFromCollection = true;
       }
     }
+    // TODO: maybe add more node types here, e.g. materialize nodes?
```
Dronplane, line 2663: MATERIALIZE node I guess should always have `variable->isFullDocumentFromCollection = true`

arangod/Aql/FixedVarExpressionContext.cpp
jsteemann: Renamed as suggested.

arangod/Aql/ExecutionPlan.cpp
jsteemann: Added MATERIALIZE nodes now as well.

arangod/Aql/ExecutionPlan.cpp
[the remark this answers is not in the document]
jsteemann: Also added MATERIALIZE nodes here.

arangod/Aql/ExecutionPlan.cpp
```diff
@@ -2632,19 +2632,38 @@ bool ExecutionPlan::fullCount() const noexcept {
...
```
jsteemann, line 2666:
```suggestion
    // TODO: maybe add more node types here?
```

PR #15560, opened 2022-01-19
Build/add zlib dependency

dothebart:
### Scope & Purpose

the integration test `api-http-spec.rb` uses zlib, but our bundle doesn't demand it so far.


KVS85 (approved):
LGTM.

PR #15561, opened 2022-01-19
[BTS-441] Ensure permission requests do not have null users

ajurdak:
### Scope & Purpose

- [x] :hankey: Bugfix

### Checklist

- [x] Manually tested
- [ ] Backports
  - [ ] Backport for 3.9: *(ToDo)*
  - [ ] Backport for 3.8: *(ToDo)*

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-441

### DIFF
--- a/js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
+++ b/js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
@@ -1184,7 +1184,7 @@
 
     checkCollectionPermissions: function (collectionID, roCallback) {
       var url = arangoHelper.databaseUrl('/_api/user/' +
-        encodeURIComponent(window.App.userCollection.activeUser) +
+        encodeURIComponent(window.App.userCollection.activeUser || "root") +
         '/database/' + encodeURIComponent(frontendConfig.db) + '/' + encodeURIComponent(collectionID));
 
       // FETCH COMPLETE DB LIST
@@ -1206,7 +1206,7 @@
 
     checkDatabasePermissions: function (roCallback, rwCallback) {
       var url = arangoHelper.databaseUrl('/_api/user/' +
-        encodeURIComponent(window.App.userCollection.activeUser) +
+        encodeURIComponent(window.App.userCollection.activeUser || "root") +
         '/database/' + encodeURIComponent(frontendConfig.db));
 
       // FETCH COMPLETE DB LIST
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 devel
 -----
+* Fixed BTS-441: Honour read only mode with disabled authentication
+
 * Upgraded boost to 1.78.0
 
 * Fixed BTS-712: Collation analyzer now always produces valid UTF-8 sequence.


js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
```diff
@@ -1206,7 +1206,7 @@
 
     checkDatabasePermissions: function (roCallback, rwCallback) {
       var url = arangoHelper.databaseUrl('/_api/user/' +
-        encodeURIComponent(window.App.userCollection.activeUser) +
+        encodeURIComponent(window.App.userCollection.activeUser || "root") +
```
joerg84, line 1209: How easy is it to add tests for such things?

ajurdak: To implement unit tests for the complete current UI is kind of an impossible task as the code is kind of not prepared for that. There are some automated unit tests I have to have a look into. We could add a test for this read-only-scenario with "disabled authentication". But this would test the scenario and not the added lines of code.

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
```diff
@@ -1206,7 +1206,7 @@
 
     checkDatabasePermissions: function (roCallback, rwCallback) {
       var url = arangoHelper.databaseUrl('/_api/user/' +
```
hkernbach, line 1208:
```suggestion
    checkDatabasePermissions: function (roCallback, rwCallback) {
      if (!window.frontendConfig.authenticationEnabled) {
        // We don't need to check permissions, as permission is disabled and no user is therefore set/logged in
        return;
      }
      var url = arangoHelper.databaseUrl('/_api/user/' +
```

hkernbach (changes requested):
Please check if my thoughts are correct, means:
No auth enabled -> no user permission check required. 

If that is true, then my fix should be clearer instead of sending just root in an non authenticated environment. I've not tested this manually - please check and verify before pushing back to devel.

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
```diff
@@ -1184,7 +1184,7 @@
 
     checkCollectionPermissions: function (collectionID, roCallback) {
       var url = arangoHelper.databaseUrl('/_api/user/' +
-        encodeURIComponent(window.App.userCollection.activeUser) +
+        encodeURIComponent(window.App.userCollection.activeUser || "root") +
```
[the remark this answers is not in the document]
mpoeter: Correct, but we can have authentication disabled but the server being in read-only mode. The problem is that when authentication is disabled we have no user in the frontend. But we still need to make the request to the server because this is the only way we can determine whether we are in read-only mode or not. I also don't like the hack to simply use "root" in case authentication is disabled, but I could not think of a better approach.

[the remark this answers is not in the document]
hkernbach: @mpoeter @ajurdak Ok in that case, my comments / code adjustments can be removed. But please add a TODO / Comment / Jira Card (please decide what suites best) to keep track of that issue. Otherwise we will forget why "root" is being used here. Will approve the PR after the issue is being tracked.

jsteemann:
@hkernbach @ajurdak @mpoeter : please sort out this PR.
Thanks!

PR #15570, opened 2022-01-20
delete test since  test for gzip transfer encoding /_api/version

dothebart:
### Scope & Purpose
via ruby since we've got the more thorough shell-foxx-api-spec.js doing that now.


KVS85 (approved):
LGTM.

PR #15587, opened 2022-01-20
[CINFRA-206] Fix checkCommitIndex returning without a commit fail reason

apetenchea:
### Scope & Purpose

First, `checkCommitIndex` has to always fill in a `commitFailReason`.
Second, because `collectEligibleFollowerIndex` used to filter out all followers which are not in the same term as the leader, the `calculateCommitIndex` function is unable to check if any of these filtered followers were forced. This leads to the possibility of a quorum being formed successfully, without including a forced follower.

This PR addresses both issues described above. One major change in the data structures is that ParticipantStateTuple now has a `TermIndexPair` instead of `LogIndex`. Tests that cover these changes were added to `CalcCommitIndexTest.cpp`.

Additional changes that were not initially planned for this PR:
- ~~Renamed `CommitFailReason` and `NothingToCommit` to something less depressing.~~
- `replicationFactor` removed from `CalculateCommitOptions`
- UpdateParticipantsConfig test suite (among others) required some updates after removing an assertion from collectFollowerIndexes`
- Small JS test update

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports


arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -489,63 +489,87 @@ auto ParticipantStateTuple::isFailed() const noexcept -> bool {
...
+auto ParticipantStateTuple::lastIndex() const noexcept -> LogIndex {
+  return lastAckedEntry.index;
+}
+
 auto operator<=>(ParticipantStateTuple const& left,
                  ParticipantStateTuple const& right) noexcept {
   // return std::tie(left.index, left.id) <=> std::tie(right.index, right.id);
   // -- not supported by apple clang
-  if (auto c = left.index <=> right.index; c != 0) {
+  if (auto c = left.lastAckedEntry <=> right.lastAckedEntry; c != nullptr) {
```
maierlars, line 504: I think we still want to sort by index only. Not sure if this is still correct when we first sort by term and then by index. 🤔

arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -489,63 +489,87 @@ auto ParticipantStateTuple::isFailed() const noexcept -> bool {
...
-
-  // number of failed participants
-  auto nrFailed = std::count_if(std::begin(indexes), std::end(indexes),
-                                [](auto& p) { return p.isFailed(); });
-  auto actualWriteConcern = std::max(
-      opt._writeConcern,
-      std::min(opt._replicationFactor - nrFailed, opt._softWriteConcern));
-  // vector of participants that are neither excluded nor
-  // have failed
+    TermIndexPair lastTermIndex)
```
maierlars, line 517: `spearhead` seems to be the right for that.

arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -489,63 +489,87 @@ auto ParticipantStateTuple::isFailed() const noexcept -> bool {
...
-  // the minimal commit index caused by forced participants
-  // if there are no forced participants, this component is just
-  // the spearhead (the furthest we could commit to)
+  if (actualWriteConcern > indexes.size() &&
+      indexes.size() == eligible.size()) {
+    // With WC greater than the number of participants, we cannot commit
+    // anything, even if all participants are eligible.
+    TRI_ASSERT(!indexes.empty());
+    auto quorum = std::vector<ParticipantId>{};
+    std::transform(std::begin(eligible), std::end(eligible),
```
maierlars, line 540: One could reserve `quorum` here as well.

arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -489,63 +489,87 @@ auto ParticipantStateTuple::isFailed() const noexcept -> bool {
...
+                   std::back_inserter(quorum), [](auto& p) { return p.id; });
+    auto const& who = std::min(std::begin(eligible), std::end(eligible));
+    return {currentCommitIndex,
+            CommitDetails::withQuorumSizeNotReached(who->id), quorum};
+  }
+
+  auto const spearhead = lastTermIndex.index;
```
maierlars, line 547: Now I see why you renamed the parameter.

arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -589,21 +618,23 @@ auto algorithms::calculateCommitIndex(
...
-          p.id, CommitFailReason::NonEligibleServerRequiredForQuorum::kFailed);
+          p.id, CommitDetails::NonEligibleServerRequiredForQuorum::kFailed);
     } else if (p.isExcluded()) {
       candidates.emplace(
-          p.id,
-          CommitFailReason::NonEligibleServerRequiredForQuorum::kExcluded);
+          p.id, CommitDetails::NonEligibleServerRequiredForQuorum::kExcluded);
+    } else if (p.lastTerm() != lastTermIndex.term) {
+      candidates.emplace(
+          p.id, CommitDetails::NonEligibleServerRequiredForQuorum::kWrongTerm);
```
maierlars, line 631: `WrongTerm` might be misleading, because the follower might be in the right term, it just never has acked an entry of the current term. We can leave it as it is now and I create a ticket to sort this out later.

arangod/Replication2/ReplicatedLog/LogCommon.h
```diff
@@ -399,31 +398,31 @@ struct CommitFailReason {
...
+      -> bool = default;
 
  private:
   template<typename... Args>
-  explicit CommitFailReason(std::in_place_t, Args&&... args) noexcept;
+  explicit CommitDetails(std::in_place_t, Args&&... args) noexcept;
 };
 
-auto to_string(CommitFailReason const&) -> std::string;
+auto to_string(CommitDetails const&) -> std::string;
```
maierlars, line 425: The name `CommitDetails` implicates that this contains information about the last commit that happened. This confusion additionally manifests itself in renaming `NothingToCommit` to `SuccessfulQuorum` which is a statement about the last commit operation. 

The purpose of this data structure was always to explain to the customer why the system **currently does not commit** more entries. i.e. it is the answer to the question why the commit index is not higher than it currently is. So if there are no log entries pending for commit, it should say in a way, that there is nothing to commit. On the other hand, declaring that the commit index is not higher, because there was a `SuccessfulQuorum` is ... wrong?

You might as well just set it always to `SuccessfulQuorum` without lying.

I have no idea how to call that thing to prevent this confusion again.

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -879,38 +885,16 @@ auto replicated_log::LogLeader::GuardedLeaderData::
...
-    }
+    TRI_ASSERT(follower->lastAckedEntry.term <= this->_self._currentTerm);
+
+    auto flags = activeParticipantsConfig->participants.find(pid);
+    TRI_ASSERT(flags != std::end(activeParticipantsConfig->participants));
+
+    indexes.emplace_back(algorithms::ParticipantStateTuple{
+        .lastAckedEntry = follower->lastAckedEntry,
+        .id = follower->_impl->getParticipantId(),
+        .failed = false,
```
maierlars, line 896: Just noticed, that we always put `false` here. I created a ticket for that. It's not urgent.

tests/Replication2/ReplicatedLog/CalcCommitIndexTest.cpp
```diff
@@ -54,21 +54,34 @@ struct CalcCommitIndexTest : ::testing::Test {
     return buf.str();
   }
 
+  auto createDefaultTermIndexPair(decltype(LogIndex::value) value)
+      -> TermIndexPair {
+    return {LogTerm{1}, LogIndex{value}};
+  }
+
```
maierlars, line 61: Either put this function (and while you're at it, the other helper functions as well) in an anonymous namespace. Otherwise they are publicly visible in the global namespace. This might cause strange bugs, if it results in an ODR violation.

maierlars (changes requested):

maierlars (approved):

PR #15588, opened 2022-01-20
Program startup options: Rename options::Flags::Hidden to options::Flags::Uncommon

Simran-B:
Superseded by #15590

### DIFF
--- a/lib/ProgramOptions/Option.h
+++ b/lib/ProgramOptions/Option.h
@@ -35,8 +35,8 @@ namespace options {
 /// @brief option flags. these can be bit-ORed to combine multiple flags
 enum class Flags : uint16_t {
   None = 0,        // nothing special here
-  Hidden = 1,      // the option is hidden by default, only made visible by
-                   // --help-all or --help-.
+  Uncommon = 1,    // the option is not listed by --help, only made visible by
+                   // --help-all, --help-. or --help-uncommon
   Obsolete = 2,    // the option is obsolete. setting it does not influence the
                    // program behavior
   Enterprise = 4,  // the option is only available in the Enterprise Edition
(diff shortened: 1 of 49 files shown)


lib/ProgramOptions/Option.h
```diff
@@ -46,6 +46,11 @@ enum class Flags : uint16_t {
                    // target host configuration
   FlushOnFirst = 32,  // when we first see this parameter, we will flush the
                       // contents of its default value before setting it.
+  // also see addOldOption() for renamed and addObsoleteOption() for deprecated
+  // options. for internal, "hidden" options use:
+  //   #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
```
Simran-B, line 51: Should this rather be `#if defined(ARANGODB_ENABLE_MAINTAINER_MODE)`?
https://github.com/arangodb/arangodb/blob/0cbb9ea895e746892e1c31d5a650b13b68c69632/arangod/Replication2/ReplicatedLog/ReplicatedLogFeature.cpp#L75

jsteemann: `#ifdef` and `#if defined` are equivalent, so it can stay as is.
Anyway, the new code comment's position here is a bit misleading, because it is currently located between `FlushOnFist` and `OsLinux`, and both flags don't have anything to do with this comment. Shouldn't it be put somewhere more appropriate?

jsteemann (approved):
LGTM

lib/ProgramOptions/Option.h
Simran-B: Right, how about placing it after Obsolete or above the enum?

jsteemann: near Obsolete seems fine, yes.

Simran-B:
The base branch for PRs can be changed but not the head branch, thus opening a new PR: #15590

PR #15589, opened 2022-01-20
APM-237: computed values

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/958

Implementation of computed values for APM-237.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [x] Docs PR: to be done!
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/958
- [x] Design document: https://arangodb.atlassian.net/wiki/spaces/REED/pages/1773207555/DD+Persistent+computed+attributes


arangod/Aql/Query.cpp
```diff
@@ -950,7 +950,7 @@ QueryResult Query::explain() {
     }
 
     enterState(QueryExecutionState::ValueType::LOADING_COLLECTIONS);
-    _ast->validateAndOptimize(*_trx);
+    _ast->validateAndOptimize(*_trx, /*optimizeNonCacheable*/ true);
```
mpoeter, line 953: I prefer to use structs instead of bools as arguments, then this would not only be easily extensible, but we could also use aggregate initialization instead of comments: `_ast->validateAndOptimize(*_trx, { .optimizeNonCacheable = true });`
But this is just a remark, no need to change to code!

arangod/ClusterEngine/ClusterCollection.cpp
```diff
@@ -156,13 +161,14 @@ PhysicalCollection* ClusterCollection::clone(LogicalCollection& logical) const {
...
 void ClusterCollection::getPropertiesVPack(velocypack::Builder& result) const {
-  // objectId might be undefined on the coordinator
   TRI_ASSERT(result.isOpenObject());
 
   if (_engineType == ClusterEngineType::RocksDBEngine) {
     result.add(StaticStrings::CacheEnabled,
                VPackValue(Helper::getBooleanValue(
                    _info.slice(), StaticStrings::CacheEnabled, false)));
+
+    // TODO: computedValues?
```
mpoeter, line 171: What about this TODO?

arangod/Aql/Expression.cpp
```diff
@@ -358,24 +358,24 @@ void Expression::initAccessor(ExpressionContext& ctx) {
...
         // generate a constant value
-        transaction::BuilderLeaser builder(&ctx.trx());
-        _node->toVelocyPackValue(*builder.get());
+        velocypack::Builder builder;
+        _node->toVelocyPackValue(builder);
 
-        _data = new uint8_t[static_cast<size_t>(builder->size())];
-        memcpy(_data, builder->data(), static_cast<size_t>(builder->size()));
+        _data = new uint8_t[static_cast<size_t>(builder.size())];
+        memcpy(_data, builder.data(), static_cast<size_t>(builder.size()));
```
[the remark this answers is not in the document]
jsteemann: I made the method "prepareForExecution" public and removed the "trx" parameter from its signature.
I wanted to avoid that this function would store any data permanently inside something owned by the transaction, because the expressions now can be used across multiple transactions. That's why no BuilderLeaser is used here anymore, as it depends on the transaction's context.
I tried to reduce the overhead of the function a bit after your comment, by at least avoiding the shared_ptr reference count increase when creating a new Builder. Now it is creating a Builder from an existing Buffer, which is slightly more efficient.

arangod/Aql/Query.cpp
jsteemann: Adjusted the code according to your suggestion.

arangod/ClusterEngine/ClusterCollection.cpp
jsteemann: It was obsolete, and is removed now.

arangod/VocBase/ComputedValues.cpp
```diff
@@ -0,0 +1,543 @@
...
+  return _trx.vocbase();
+}
+
+transaction::Methods& ComputedValuesExpressionContext::trx() const {
+  return _trx;
+}
+
+void ComputedValuesExpressionContext::registerWarning(ErrorCode errorCode,
+                                                      char const* msg) {
+  if (_failOnWarning) {
```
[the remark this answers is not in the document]
jsteemann: Added logging of warnings & errors that happen during computed values computation.

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
```diff
@@ -334,22 +334,38 @@
...
+    
+    changeComputedValues: function (computedValues, callback) {
+      if (!computedValues) {
+        computedValues = null;
+      }
+
+      $.ajax({
+        cache: false,
+        type: 'PUT',
+        url: arangoHelper.databaseUrl('/_api/collection/' + encodeURIComponent(this.get('id')) + '/properties'),
```
adityamukho, line 346: It seems accessing collection names by their numeric ID has been deprecated since 3.4.0. Please see https://www.arangodb.com/docs/stable/http/collection-modifying.html#change-properties-of-a-collection.

adityamukho (commented):
Only reviewed files under the `js` folder. LGTM except one comment that might be relevant.

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
jsteemann: Fixed

adityamukho (approved):
LGTM (Only Web UI reviewed).

arangod/Aql/Expression.cpp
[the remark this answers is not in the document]
mpoeter: How about we have two `prepareForExecution` overloads, one that takes a builder and one that doesn't. The one without arguments can create a local builder (just like you do right now) and call the second overload passing that builder. However, `execute` could than pass a builder from the context's BuilderLeaser.

arangod/Aql/StandaloneCalculation.cpp
```diff
@@ -268,22 +271,37 @@ Result StandaloneCalculation::validateQuery(TRI_vocbase_t& vocbase,
...
+            case arangodb::aql::NODE_TYPE_SUBQUERY:
+            case arangodb::aql::NODE_TYPE_FOR:
+            case arangodb::aql::NODE_TYPE_LET: {
+              // these nodes are only ok for analyzer expressions, but
+              // not for computed values
+              if (isComputedValue) {
+                errorMessage = "Node type '";
+                errorMessage.append(node->getTypeString())
+                    .append("' is forbidden")
+                    .append(errorContext);
```
MBkkt, line 294:
```suggestion
                errorMessage = absl::StrCat("Node type '", node->getTypeString(), "' is forbidden", errorContext);
```

arangod/VocBase/ComputedValues.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+  for (auto it : VPackArrayIterator(params)) {
+    VPackSlice name = it.get("name");
+    if (!name.isString() || name.getStringLength() == 0) {
+      return res.reset(
+          TRI_ERROR_BAD_PARAMETER,
+          "invalid 'computedValues' entry: invalid attribute name");
+    }
+
+    std::string_view n = name.stringView();
```
MBkkt, line 451: auto?

arangod/Transaction/Methods.cpp
```diff
@@ -1918,20 +2161,30 @@ Future<OperationResult> transaction::Methods::removeLocal(
       resultBuilder.add(VPackSlice::emptyObjectSlice());
     }
   }
+  replicationData->close();
+
+  // LOG_DEVEL << "REMOVE. VALUE: " << value.toJson()
+  //          << ", RESULTBUILDER: " << resultBuilder.slice().toJson()
+  //          << ", REPLICATIONDATA: " << replicationData->slice().toJson()
+  //          << ", SILENT: " << options.silent;
```
MBkkt, line 2169: Remove it?

arangod/Transaction/Methods.cpp
```diff
@@ -1751,162 +2028,128 @@ Future<OperationResult> transaction::Methods::removeLocal(
...
+        (res.ok() && !options.silent)) {
+      TRI_ASSERT(oldRevisionId.isSet());
+      buildDocumentIdentity(
+          *collection, resultBuilder, cid, key, oldRevisionId,
+          RevisionId::none(),
+          options.returnOld ? previousDocumentBuilder.get() : nullptr, nullptr);
+    }
+
+    if (res.ok() && !excludeFromReplication) {
+      replicationData->openObject(true);
```
MBkkt, line 2124: Just for me: When I should to write true for openObject or openArray

arangod/Transaction/Methods.cpp
```diff
@@ -1751,162 +2028,128 @@ Future<OperationResult> transaction::Methods::removeLocal(
...
       key = keySlice.stringView();
-    } else {
-      return Result(TRI_ERROR_ARANGO_DOCUMENT_HANDLE_BAD);
     }
 
-    // Primary keys must not be empty
+    // primary key must not be empty
     if (key.empty()) {
-      return Result(TRI_ERROR_ARANGO_DOCUMENT_HANDLE_BAD);
+      return res.reset(TRI_ERROR_ARANGO_DOCUMENT_HANDLE_BAD);
```
MBkkt, line 2084: Can you explain why you suggest to write 
`return ....reset(...)` instead of `return {...};`

I think it can be better only if we already have an error and replace it with another
But know we don't make optimization in this case, and this case is very rare

arangod/Transaction/Methods.cpp
```diff
@@ -1751,162 +2028,128 @@ Future<OperationResult> transaction::Methods::removeLocal(
...
+  if (res.fail()) {
+    return futures::makeFuture(OperationResult(std::move(res), options));
+  }
 
-  TRI_ASSERT((replicationType == ReplicationType::LEADER) ==
-             (followers != nullptr));
-  TRI_ASSERT(!options.silent || replicationType != ReplicationType::LEADER ||
-             followers->empty());
+  bool excludeAllFromReplication =
+      (replicationType != ReplicationType::LEADER ||
```
MBkkt, line 2046: unnecessary parentheses

arangod/Transaction/Methods.cpp
```diff
@@ -1302,11 +1445,21 @@ Future<OperationResult> transaction::Methods::insertLocal(
       resultBuilder.add(VPackSlice::emptyObjectSlice());
     }
   }
+  replicationData->close();
+
+  // LOG_DEVEL << "INSERT. VALUE: " << value.toJson()
+  //          << ", RESULTBUILDER: " << resultBuilder.slice().toJson()
```
MBkkt, line 1451: remove it?

arangod/Transaction/Methods.cpp
```diff
@@ -79,7 +82,77 @@ using Future = futures::Future<T>;
...
-enum class ReplicationType { NONE, LEADER, FOLLOWER };
+/// @brief check if a list of attributes have the same values in two vpack
+/// documents
+bool shardKeysChanged(LogicalCollection const& collection,
+                      velocypack::Slice oldValue, velocypack::Slice newValue,
+                      bool isPatch) {
+  TRI_ASSERT(oldValue.isObject());
+  TRI_ASSERT(newValue.isObject());
+
+  std::vector<std::string> const& shardKeys = collection.shardKeys();
```
MBkkt, line 93: auto?

arangod/Transaction/Helpers.cpp
```diff
@@ -433,6 +443,469 @@ std::string transaction::helpers::makeIdFromParts(
...
+  VPackSlice fromSlice;
+  VPackSlice toSlice;
+
+  containers::FlatHashMap<std::string_view, VPackSlice> newValues;
+  {
+    VPackObjectIterator it(newValue, true);
+    while (it.valid()) {
+      auto current = *it;
+      std::string_view key(current.key.stringView());
+      if (key.size() >= 3 && key[0] == '_' &&
```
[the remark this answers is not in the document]
MBkkt: Not for this pr

arangod/RocksDBEngine/RocksDBIncrementalSync.cpp
```diff
@@ -315,10 +328,27 @@ Result syncChunkRocksDB(DatabaseInitialSyncer& syncer,
...
+        std::pair<LocalDocumentId, RevisionId> lookupResult;
+        Result r = physical->lookupKey(trx, localKey, lookupResult,
+                                       ReadOwnWrites::yes);
 
-        auto r = physical->remove(*trx, keyBuilder->slice(), mdr, options);
+        if (r.ok()) {
+          TRI_ASSERT(lookupResult.first.isSet());
+          TRI_ASSERT(lookupResult.second.isSet());
+          LocalDocumentId documentId = lookupResult.first;
+          RevisionId revisionId = lookupResult.second;
```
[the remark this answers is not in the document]
jsteemann: Will adjust.

arangod/VocBase/ComputedValues.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+using namespace arangodb;
+using namespace arangodb::aql;
+
+namespace {
+// name of bind parameter variable that contains the current document
+constexpr std::string_view docParameter("doc");
+
+// helper function to turn an enum class value into its underlying type
+std::underlying_type<arangodb::ComputeValuesOn>::type mustComputeOnValue(
```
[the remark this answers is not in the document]
jsteemann: Will adjust.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -168,7 +230,7 @@ struct TimeTracker {
   }
 
  private:
-  std::optional<arangodb::TransactionStatistics::ReadWriteMetrics>& metrics;
+  std::optional<TransactionStatistics::ReadWriteMetrics>& metrics;
```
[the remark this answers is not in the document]
jsteemann: Ok

arangod/Transaction/Methods.cpp
```diff
@@ -1148,151 +1273,169 @@ Future<OperationResult> transaction::Methods::insertLocal(
...
+        if (res.ok()) {
+          TRI_ASSERT(previousDocumentBuilder->slice().isObject());
+
+          res = modifyLocalHelper(
+              *collection, value, oldDocumentId, oldRevisionId,
+              previousDocumentBuilder->slice(), newRevisionId,
+              *newDocumentBuilder, options, batchOptions, /*isUpdate*/ false);
+        }
+
+        TRI_ASSERT(res.fail() || newDocumentBuilder->slice().isObject());
```
mpoeter, line 1353: The two code paths look almost identical except for the `/*isUpdate* true` vs `/*isUpdate* false` and the handling of `excludeFromReplication`. Can we streamline this and remove duplicate code?

arangod/Transaction/Methods.cpp
```diff
@@ -1148,151 +1273,169 @@ Future<OperationResult> transaction::Methods::insertLocal(
...
       }
-      res = workForOneDocument(s, true, excludeFromReplication);
       if (res.fail()) {
         createBabiesError(replicationType == ReplicationType::FOLLOWER
                               ? nullptr
                               : &resultBuilder,
                           errorCounter, res);
-      } else if (excludeFromReplication) {
-        excludePositions.insert(it.index());
+        res.reset();
```
mpoeter, line 1432: This `res.ok()` check seems superfluous since in the error case we always reset `res`.

arangod/Transaction/Methods.cpp
```diff
@@ -1415,229 +1620,198 @@ Future<OperationResult> transaction::Methods::modifyCoordinator(
...
+    }
+    if (options.validate) {
+      batchOptions.schema = collection->schema();
+    }
+    std::shared_ptr<ComputedValues> cv = collection->computedValues();
+    if (cv != nullptr && ((isUpdate && cv->mustComputeValuesOnUpdate()) ||
+                          (!isUpdate && cv->mustComputeValuesOnReplace()))) {
+      batchOptions.computedValues = std::move(cv);
+    }
+  }
```
mpoeter, line 1683: `insertLocal` has a very similar code block. Can we pull out a helper function to reduce code duplication?

arangod/Transaction/Methods.cpp
```diff
@@ -1647,28 +1821,27 @@ Future<OperationResult> transaction::Methods::modifyLocal(
       resultBuilder.add(VPackSlice::emptyObjectSlice());
     }
   }
+  replicationData->close();
+
+  // LOG_DEVEL << "MODIFY: RESULTBUILDER: " << resultBuilder.slice().toJson()
+  //          << ", REPLICATIONDATA: " << replicationData->slice().toJson();
```
mpoeter, line 1827:
```suggestion
```

mpoeter (approved):

arangod/VocBase/LogicalCollection.h
```diff
@@ -492,6 +484,10 @@ class LogicalCollection : public LogicalDataSource {
   /// @brief sharding information
   std::unique_ptr<ShardingInfo> _sharding;
 
+  // `_computedValues` must be used with atomic accessors only!!
+  // We use relaxed access (load/store) as we only care about atomicity.
```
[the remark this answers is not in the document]
mpoeter: I think relaxed is not correct. I suggested to use acquire/release in the according places but missed this comment which obviously should be adapted as well.

arangod/VocBase/ComputedValues.cpp
```diff
@@ -0,0 +1,594 @@
...
+  THROW_ARANGO_EXCEPTION_MESSAGE(errorCode, error);
+}
+
+std::string ComputedValuesExpressionContext::buildLogMessage(
+    std::string_view type, char const* msg) const {
+  std::string error;
+
+  // note: on DB servers, the error message will contain the shard name
+  // rather than the collection name.
+  error.append("computed values expression evaluation produced a runtime ")
```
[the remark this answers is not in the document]
MBkkt: Maybe true, but StrCat definitely simply than current code

arangod/VocBase/LogicalCollection.h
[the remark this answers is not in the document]
MBkkt: @jsteemann Can you also change/fix another ptr (it has same comment) usage

arangod/Transaction/Helpers.cpp
```diff
@@ -433,6 +443,469 @@ std::string transaction::helpers::makeIdFromParts(
...
+
+  // Find the attributes in the newValue object:
+  VPackSlice fromSlice;
+  VPackSlice toSlice;
+
+  containers::FlatHashMap<std::string_view, VPackSlice> newValues;
+  {
+    VPackObjectIterator it(newValue, true);
+    while (it.valid()) {
+      auto current = *it;
```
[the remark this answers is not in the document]
MBkkt: @jsteemann Why not?

arangod/Transaction/Helpers.cpp
```diff
@@ -433,6 +443,469 @@ std::string transaction::helpers::makeIdFromParts(
...
+    revisionId = collection.newRevisionId();
+    b->add(StaticStrings::RevString, revisionId.toValuePair(&ridBuffer[0]));
+  }
+
+  containers::FlatHashSet<std::string_view> keysWritten;
+
+  // add other attributes after the system attributes
+  VPackObjectIterator it(newValue, true);
+  while (it.valid()) {
+    std::string_view key(it.key().stringView());
```
MBkkt, line 862:
```suggestion
    auto key = it.key().stringView();
```
suggest grep for that

arangod/Transaction/Helpers.cpp
```diff
@@ -433,6 +443,469 @@ std::string transaction::helpers::makeIdFromParts(
...
+    encoding::storeNumber<uint64_t>(p, collection.planId().id(),
+                                    sizeof(uint64_t));
+  } else {
+    // local server
+    encoding::storeNumber<uint64_t>(p, collection.id().id(), sizeof(uint64_t));
+  }
+
+  // _from and _to
+  if (collection.type() == TRI_COL_TYPE_EDGE) {
+    bool extendedNames = trx.vocbase().server().hasFeature<DatabaseFeature>() &&
```
MBkkt, line 708: NIT
Is `trx.vocbase().server().hasFeature<DatabaseFeature>()` always true?
But still think get server once will be look better

arangod/Transaction/Methods.cpp
```diff
@@ -79,7 +82,122 @@ using Future = futures::Future<T>;
...
+          opType != TRI_VOC_DOCUMENT_OPERATION_INSERT &&
+          (collection.shardKeys().size() > 1 ||
+           collection.shardKeys()[0] != StaticStrings::KeyString);
+      batchOptions.validateSmartJoinAttribute =
+          collection.hasSmartJoinAttribute();
+    }
+    if (options.validate) {
+      batchOptions.schema = collection.schema();
+    }
+    std::shared_ptr<ComputedValues> cv = collection.computedValues();
```
MBkkt, line 103: NIT
auto

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -219,47 +224,98 @@ LogicalCollection::LogicalCollection(TRI_vocbase_t& vocbase, VPackSlice info,
...
-      return {TRI_ERROR_VALIDATION_BAD_PARAMETER,
-              "Error when building schema: "s + ex.what()};
+    // schema will be removed if empty object is given
+    if (!schema.isEmptyObject()) {
+      try {
+        newSchema = std::make_shared<ValidatorJsonSchema>(schema);
+      } catch (std::exception const& ex) {
+        using namespace std::literals::string_literals;
+        return {TRI_ERROR_VALIDATION_BAD_PARAMETER,
+                "Error when building schema: "s + ex.what()};
```
MBkkt, line 264:
```suggestion
                absl::StrCat("Error when building schema: ", ex.what())};
```

arangod/VocBase/ComputedValues.cpp
```diff
@@ -0,0 +1,575 @@
...
+  // note: on DB servers, the error message will contain the shard name
+  // rather than the collection name.
+  std::string error =
+      absl::StrCat("computed values expression evaluation produced a runtime ",
+                   type, " for attribute '", _name, "' of collection '",
+                   _collection.vocbase().name(), "/", _collection.name(), "'");
+
+  if (msg != nullptr) {
+    error.append(": ");
+    error.append(msg);
```
MBkkt, line 115:
```suggestion
    absl::StrAppend(&error, ": ", msg);
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -110,39 +110,47 @@ arangodb::Result removeRevisions(
...
-  }
+  if (!toRemove.empty()) {
+    PhysicalCollection* physical = collection.getPhysical();
 
-  PhysicalCollection* physical = collection.getPhysical();
+    arangodb::OperationOptions options;
+    options.silent = true;
+    options.ignoreRevs = true;
+    options.isRestore = true;
+    options.waitForSync = false;
```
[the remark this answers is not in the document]
MBkkt: next pr?

arangod/Transaction/Helpers.cpp
jsteemann: > Why not?

A VPackSlice only contains one member, which is a pointer. It can always be passed by value.
There is no need to hand it around or access it by const reference.

MBkkt (approved):

arangod/VocBase/ComputedValues.h
```diff
@@ -0,0 +1,210 @@
...
+
+  void registerWarning(ErrorCode errorCode, char const* msg) override;
+
+  void registerError(ErrorCode errorCode, char const* msg) override;
+
+  void failOnWarning(bool value) noexcept { _failOnWarning = value; }
+
+  void setName(std::string_view name) noexcept { _name = name; }
+
+  icu::RegexMatcher* buildRegexMatcher(char const* ptr, size_t length,
```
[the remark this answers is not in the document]
jsteemann: This is an already existing API, which I just inherit/override here.
We can change that API at any point, independently, but I don't think it needs to be done as part of this PR.

arangod/VocBase/ComputedValues.h
```diff
@@ -0,0 +1,210 @@
...
+
+// expression context used for calculating computed values inside
+class ComputedValuesExpressionContext final : public aql::ExpressionContext {
+ public:
+  explicit ComputedValuesExpressionContext(transaction::Methods& trx,
+                                           LogicalCollection& collection);
+
+  void registerWarning(ErrorCode errorCode, char const* msg) override;
+
+  void registerError(ErrorCode errorCode, char const* msg) override;
```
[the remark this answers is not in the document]
jsteemann: This is an already existing API, which I just inherit/override here.
We can change that API at any point, independently, but I don't think it needs to be done as part of this PR.

arangod/Transaction/Methods.cpp
jsteemann: `VPackBuilder::openObject(/*unindexed*/ = false)`

_unindexed_ here means we will write a VPack type Object _without_ an index table. This can save space and processing time for building the index table. At this code location, we know that we will only write two keys into the object, so it is a small optimization I recycled from previous code.

arangod/VocBase/ComputedValues.h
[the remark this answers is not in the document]
MBkkt: Understood

arangod/VocBase/ComputedValues.h
```diff
@@ -0,0 +1,210 @@
...
+  kReplace = 4,
+};
+
+// expression context used for calculating computed values inside
+class ComputedValuesExpressionContext final : public aql::ExpressionContext {
+ public:
+  explicit ComputedValuesExpressionContext(transaction::Methods& trx,
+                                           LogicalCollection& collection);
+
+  void registerWarning(ErrorCode errorCode, char const* msg) override;
```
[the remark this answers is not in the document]
jsteemann: IMO this is a valid point of view.
But my point of view is this:
As the class is currently `final` there is no immediate benefit in marking all methods as `final` again. There would only be benefit if someone removes the `final` specifier from the class, which right now seems unlikely. And when the `final` specifier gets removed from the class, IMO it is even better that not every method is marked as final individually, because a `final` specifier gets removed in order to override things. 

I think both view points are reasonable. So what shall we do?

arangod/Transaction/Methods.cpp
MBkkt: Thanks, so true can be useful as optimization when I know I always have only few values, or doesn't have access to them via get/at?

arangod/VocBase/ComputedValues.h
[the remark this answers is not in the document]
MBkkt: I think for this PR do nothing (for faster merge and lesser code changes).

But in general maybe we can make poll, choose one option, put it in codestyle, and try to follow it.

Btw as another argument to suggested option:
it more consistent, final methods always marked as final, don't need to check it in class, more easy to find, etc

arangod/Transaction/Methods.cpp
jsteemann: Yes, exactly!

arangod/VocBase/ComputedValues.h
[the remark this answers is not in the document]
jsteemann: Running a poll is a good idea. I can live with both outcomes, but I also think we should agree on one variant for consistency.

PR #15515, opened 2022-01-13
[BTS-714] Fix various TSan issues. 

mpoeter:
### Scope & Purpose

This PR fixes various issues found during 3.9 release testing. I also found most of them independently during local testing and fixed them in #15514 for devel, so this is basically a backport of that PR.

- [x] :hankey: Bugfix 

#### Backports:

- [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/BTS-714

### Testing & Verification

- [x] The behavior in this PR was *manually tested*
- [x] This change is already covered by existing tests
- [x] I ensured this code runs with TSan

PR #15514, opened 2022-01-13
[ORCO-26] Fix various TSan issues.

mpoeter:
### Scope & Purpose

This PR fixes various issues I found while running integration tests locally with a TSan build.

- [x] :hankey: Bugfix 

#### Backports:

- [x] Backport for 3.9: #15515

#### Related Information

- [x] Jira ticket number: https://arangodb.atlassian.net/browse/ORCO-26

### Testing & Verification

- [x] The behavior in this PR was *manually tested*
- [x] This change is already covered by existing tests
- [x] I ensured this code runs with TSan

PR #15605, opened 2022-01-25
Fix a potential deadlock in QueryList introduced in #15515

mpoeter:
### Scope & Purpose

devel fix will be part of #15514; other backports are not necessary because the faulty change was not backported to 3.8.

The problem was that when the QueryList acquires the shared_ptr, it can happen that once it drops that reference, it has to call the destructor (because this was the _last_ reference), which would then try to remove the query from the QueryList. For that we need to acquire the write lock, but since we are calling from some QueryList function that already holds the read lock, this results in a dead lock.

- [x] :hankey: Bugfix

### DIFF
--- a/arangod/Aql/QueryList.cpp
+++ b/arangod/Aql/QueryList.cpp
@@ -130,8 +130,7 @@ bool QueryList::insert(Query& query) {
     }
 
     // return whether or not insertion worked
-    bool inserted =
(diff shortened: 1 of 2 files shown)


KVS85:
- [ ] https://jenkins.arangodb.biz/job/arangodb-ANY-matrix/88/

jsteemann (approved):
LGTM!

PR #15606, opened 2022-01-25
Reduce overhead for checking failure points when none are set.

mpoeter:
### Scope & Purpose

Several profiling runs showed high contention on the failure points lock when running with a maintainer build.

### DIFF
--- a/lib/Basics/debugging.cpp
+++ b/lib/Basics/debugging.cpp
@@ -52,6 +52,8 @@ using namespace arangodb;
 #ifdef ARANGODB_ENABLE_FAILURE_TESTS
 
 namespace {
+std::atomic<bool> hasFailurePoints{false};
+
 /// @brief a read-write lock for thread-safe access to the failure points set
 arangodb::basics::ReadWriteLock failurePointsLock;
 
@@ -108,8 +110,11 @@ void TRI_TerminateDebugging(std::string_view message) {
 
 /// @brief check whether we should fail at a specific failure point
 bool TRI_ShouldFailDebugging(std::string_view value) noexcept {
-  READ_LOCKER(readLocker, ::failurePointsLock);
-  return ::failurePoints.find(value) != ::failurePoints.end();
+  if (::hasFailurePoints.load(std::memory_order_relaxed)) {
+    READ_LOCKER(readLocker, ::failurePointsLock);
+    return ::failurePoints.find(value) != ::failurePoints.end();
+  }
+  return false;
 }
 
 /// @brief add a failure point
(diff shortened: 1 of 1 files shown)


lib/Basics/debugging.cpp
```diff
@@ -121,6 +126,7 @@ void TRI_AddFailurePointDebugging(std::string_view value) {
   }
 
   if (added) {
+    ::hasFailurePoints.store(true, std::memory_order_relaxed);
```
jsteemann, line 129: Updating the list of failure points and the new atomic variable does not happen atomically.
So this will introduce a potential race in a multi-threaded setup, because we are updating the atomic variable while not holding the lock.
This is likely not a problem for our current tests which set a failure point via this API and wait until the API call returns. But for other multi-threaded tests it may cause issues (in the future).

lib/Basics/debugging.cpp
```diff
@@ -149,6 +158,7 @@ void TRI_ClearFailurePointsDebugging() noexcept {
     numExisting = ::failurePoints.size();
     ::failurePoints.clear();
   }
+  ::hasFailurePoints.store(false, std::memory_order_relaxed);
```
jsteemann, line 161: This could also cause problems because we are not updating the atomic variable while under the lock.
Imagine we are removing all failure points here and then are about to store false in the atomic variable.
Another thread could slip in here and a new a failure point before we are setting the atomic variable to `false`. Then this API will unconditionally overwrite the atomic's `true` value with a `false`, and we will have an inconsistent state.

mpoeter:
@jsteemann You are absolutely right, these were of course supposed to be _inside_ the locks. 🤦 However, if we ever had tests that would trigger these errors that would mean that we have racing add/remove/clear calls, so the active failure points would actually be nondeterministic.

lib/Basics/debugging.cpp
mpoeter: Fixed!

lib/Basics/debugging.cpp
mpoeter: Fixed!

lib/Basics/debugging.cpp
```diff
@@ -118,6 +123,9 @@ void TRI_AddFailurePointDebugging(std::string_view value) {
   {
     WRITE_LOCKER(writeLocker, ::failurePointsLock);
     added = ::failurePoints.emplace(value).second;
+    if (added) {
```
jsteemann, line 126: I think this `if` condition can be removed here.
If we emplace and the failure point is already present in `::failurePoints`, we could still set the atomic variable to `true`. This would still be correct. And probably the case of adding an already set failure point is unlikely.

jsteemann (approved):
LGTM!

PR #15623, opened 2022-01-27
BTS-621 fix

Dronplane:
### Scope & Purpose

Fix for BTS-621.
During recovery DBServer could be in the same time in Upgrade mode with cluster feature disabled. This lead to initializing a link with internal datastore set to  nullptr, During recovery insert call lead to segfault. Now even if cluster feature is disabled  - at least datastore is properly initialized so link could do insert/removes. In the rare case of "old" link definition without the collection name stored - init throws an exception asking to finish the recovery first.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports


arangod/IResearch/IResearchLink.cpp
```diff
@@ -321,18 +324,39 @@ Result IResearchLink::init(velocypack::Slice definition,
...
       }
+    }
 
-      if (!clusterWideLink) {
-        // prepare data-store which can then update options
-        // via the IResearchView::link(...) call
-        auto res = initDataStore(initCallback, meta._version, sorted,
-                                 storedValuesColumns, primarySortCompression);
+    if (!clusterWideLink) {
+      if (meta._collectionName.empty() && !clusterIsEnabled &&
```
gnusi, line 330: we can reduce risk even more in case if we can check if `_id` attribute is indexed.

gnusi (approved):

PR #15627, opened 2022-01-27
Change some default values for RocksDB memory usage

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/888

**This PR is a proposal to adjust some default values for RocksDB memory accounting**. This will not be backported but targets only 3.10/devel.

Changed default value of startup option `--rocksdb.cache-index-and-filter-blocks` from `false` to `true`.  

This makes RocksDB track all loaded index and filter blocks in the block cache, so they are accounted for in RocksDB's block cache. Also the default value for the startup option `--rocksdb.enforce-block-cache-size-limit` was flipped from `false` to `true` to make the RocksDB block cache not temporarily exceed the configured memory limit (`--rocksdb.block-cache-size`).

These default value changes will make RocksDB adhere much better to the configured memory limit. This is a trade-off between memory usage stability and performance. These change may have a small negative impact on performance because if the block cache is not large enough to hold the data plus the index and filter blocks, additional disk I/O may be performed compared to previous versions. In case there is still unused RAM capacity available, it may be sensible to increase the total size of the RocksDB block cache.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/888
- [ ] Enterprise PR:
- [ ] Design document:


kvahed (approved):
LGTM

PR #15638, opened 2022-01-30
reduce peak memory usage of some AQL queries

jsteemann:
### Scope & Purpose

Reduce memory usage of inner joins if they were performed by the
IndexExecutor with dynamic index lookup expressions that needed to be
recomputed for input from the outer loop.

For example, in the query
```
FOR i IN 1..1000
  FOR doc IN collection
    FILTER doc.indexAttribute == i
    RETURN doc
```
the inner loop will be executed 1000 times. The IndexExecutor in the inner
loop needed to rebuild the index lookup attribute from the value of `i` 1000
times as well. The memory for index lookup attributes came from the Ast's
memory allocator and was not freed until the end of the query. In this
query, it would mean that up to 1000 lookup values were held in memory. With
larger inputs even more memory would be used.

Now the memory for index lookup values is freed when a new lookup value is
computed, i.e. only a single lookup value is held in memory.
This drastically reduces peak memory usage for queries that use index lookups
in inner loops and that get lots of different inputs from outer loops.

In a real world example query, I have observed a peak memory usage change
from ~2GB without the change down to ~10MB with the change, i.e. 99.5%
reduction.

Backport versions unclear yet. Will be discussed.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/15645

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/AstResources.cpp
```diff
@@ -44,15 +44,38 @@ AstResources::AstResources(arangodb::ResourceMonitor& resourceMonitor)
...
+  _nodes.clearMost();
+  clearStrings();
+  _resourceMonitor.decreaseMemoryUsage(memoryUsage);
+
+  _shortStringStorage.clearMost();
+}
+
+// clears dynamic string data
+void AstResources::clearStrings() noexcept {
   for (auto& it : _strings) {
```
neunhoef, line 74: I like the fact that there is now a `clearStrings`. However, memory usage must be updated by **the called** of `clearStrings`, this is unintuitive and error-prone. I know that we do not want to add another atomic operation, so we should at least add a clearly visible comment to `clearStrings` that the **caller** has to adjust the memory usage **before** calling `clearStrings` by `_stringsLength`. Both current call sites do this, so there is no bug, but this is dangereous,.

neunhoef (changes requested):
This is overall a very good idea and well done. See two tiny comments where I would like to see a small adjustment. Other than that: LGTM.

arangod/Aql/AstResources.cpp
```diff
@@ -65,12 +88,13 @@ size_t AstResources::newCapacity(T const& container,
 
   size_t capacity = container.size() + 1;
   if (capacity > container.capacity()) {
-    capacity *= 2;
+    capacity = container.size() * 2;
```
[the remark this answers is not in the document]
jsteemann: No, in line 91, `size()` can newer be 0. This is because there is a check for container emptiness in line 84 (directly before). So `size()` is at least one here, so that doubling it will always have an effect.

arangod/Aql/AstResources.cpp
jsteemann: Ok, will do!

jsteemann: Made `clearStrings()` private and added code comments.

arangod/Aql/AstResources.cpp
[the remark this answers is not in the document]
neunhoef: Good. I had not seen this.

neunhoef (approved):
LGTM. I second backporting this, since it is an important improvement for the memory usage of AQL.

mpoeter (approved):

PR #15647, opened 2022-01-31
[CINFRA-232] Server Failure Oracle

apetenchea:
### Scope & Purpose

This PR introduces a cache that keeps track of which server is considered failed by the supervision. This information is crucial for the implementation of soft write concern that relies on the number of failed participants and temporarily allows a smaller write concern. This is done using an agency callback together with a periodic flush of the cache.

https://arangodb.atlassian.net/browse/CINFRA-232

Rest API:
1. *_admin/cluster/failureOracle/status* - get cache status
2. *_admin/cluster/failureOracle/flush* - clear and rebuild cache on a particular server (behaves differently on coordinators)

Design Document: https://arangodb.atlassian.net/wiki/spaces/CInfra/pages/1811808257/Concept+Code+FailureOracle

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**


arangod/Replication2/ReplicatedLog/ParticipantsCacheFeature.cpp
```diff
@@ -0,0 +1,114 @@
...
+  try {
+    auto x = server().getFeature<ClusterFeature>().agencyCallbackRegistry();
+    x->unregisterCallback(agencyCallback);
+  } catch (std::exception const& ex) {
+    LOG_TOPIC("42af2", WARN, Logger::REPLICATION2)
+        << "caught unexpected exception while unregistering agency callback in "
+           "ParticipantsCacheFeature: "
+        << ex.what();
+  }
+}
```
maierlars, line 85: I suggest to deregister the call back in the features `stop` function. You have no control over when the destructor is executed with respect to other features destructor or the server destructor(afaik)

arangod/Replication2/ReplicatedLog/FailureOracle.h
```diff
@@ -0,0 +1,37 @@
...
+/// @author Manuel Pöter
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <string_view>
+
+namespace arangodb::replication2 {
+
+struct FailureOracle {
```
maierlars, line 31: Similar to `ILogFollower` and `ILogLeader` this should be named starting with a captial i.

arangod/Cluster/ParticipantsCacheFeature.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "RestServer/arangod.h"
+#include "Scheduler/Scheduler.h"
+
+#include <memory>
+#include <unordered_map>
+
+namespace arangodb::cluster {
+struct IFailureOracle;
+class ParticipantsCache;
```
maierlars, line 35: This forward declaration is not needed.

arangod/Cluster/ParticipantsCacheFeature.h
```diff
@@ -0,0 +1,63 @@
...
+
+class ParticipantsCacheFeature final
+    : public ArangodFeature,
```
maierlars, line 38: I dislike the name `ParticipantsCache`. Is is very board could be anything and the thing it does is the last thing I think about given the name, `Participant` is a term, that is exclusively used in the replicated logs business and in all other places these are referred to as servers. And it's not the participants that are cached. 

`ServerFailureCache` would be better. Or `FailureOracleFeature` and than `ParticipantsCache` becomes simply `FailureOracleImpl` 🤷🏽 names are the hardest but most important part ...

arangod/Cluster/ParticipantsCacheFeature.cpp
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  _flushJob = scheduler->queueDelayed(RequestLane::AGENCY_CLUSTER, 50s,
+                                      [this](bool canceled) {
+                                        if (!canceled) {
+                                          this->flush();
+                                          this->scheduleFlush();
+                                        }
+                                      });
+}
```
maierlars, line 249: What happens if `flush` or `scheduleFlush` throws an exception? What if `queueDelayed` throws an exception? Are there other exception possible than just `std::bad_alloc` and some `std::system_error` from the mutexes?

arangod/Cluster/ParticipantsCacheFeature.cpp
```diff
@@ -0,0 +1,251 @@
...
+  if (!scheduler) {
+    return;
+  }
+
+  _flushJob = scheduler->queueDelayed(RequestLane::AGENCY_CLUSTER, 50s,
+                                      [this](bool canceled) {
+                                        if (!canceled) {
+                                          this->flush();
+                                          this->scheduleFlush();
+                                        }
```
maierlars, line 247: Is this code actually ok? I think there can be a race between calling `cancel` and this lambda being executed. The feature might also be deleted (I am not sure about that part). But I am pretty sure that the lambda can still be executed after `stop` returned. (which `cancel == false`)

arangod/Cluster/ParticipantsCacheFeature.cpp
```diff
@@ -0,0 +1,251 @@
...
+        << " expected object in agency at " << kSupervisionHealthPath
+        << " but got " << result.toString();
+    FailureMap newMap;
+    for (auto const& [key, value] : VPackObjectIterator(result)) {
+      auto serverId = key.copyString();
+      auto isServerGood =
+          value.get(kHealthyServerKey).isEqualString(HealthyServerValue);
+      newMap[serverId] = !isServerGood;
+    }
+    _cache->reset(std::move(newMap));
```
maierlars, line 218: This looks similar to the code in the agency callback. Can we make one function out of that?

maierlars (changes requested):
Let's find a better name for the feature and the implementation class.

arangod/Cluster/ParticipantsCacheFeature.h
apetenchea: There is a `shared_ptr<ParticipantsCache>` member.

arangod/Cluster/ParticipantsCacheFeature.h
apetenchea: Indeed, the name was a generic one. I was thinking it's going to have some other functionality in the future, but it's worth renaming it later and having a more specific name right now. I'm using `FailureOracleFeature` and `FailureOracleImpl`.

arangod/Cluster/ParticipantsCacheFeature.cpp
apetenchea: That's a good observation, thanks!

arangod/RestHandler/RestAdminClusterHandler.cpp
```diff
@@ -2123,3 +2135,115 @@ RestStatus RestAdminClusterHandler::handleRebalanceShards() {
...
+            {
+              VPackObjectBuilder participantsFlushStatus(&response);
+              std::size_t idx = 0;
+              for (auto const& [pid, pf] : status) {
+                auto& result = statusCodes.at(idx++);
+                response.add(pid, VPackValue(std::move(result.get())));
+              }
+              response.add(id, VPackValue(fuerte::StatusAccepted));
+            }
+            return std::move(response);
```
maierlars, line 2234: This move prevents copy elision.

arangod/RestServer/arangod.cpp
```diff
@@ -56,6 +56,7 @@
 #include "Cluster/ClusterFeature.h"
 #include "Cluster/ClusterUpgradeFeature.h"
 #include "Cluster/MaintenanceFeature.h"
+#include "Cluster/FailureOracleFeature.h"
```
neunhoef, line 59: Why is this included here? It does not seem to be used in the file...

neunhoef (approved):
LGTM. See one comment, but this is a tiny thing only.

arangod/RestServer/arangod.cpp
apetenchea: It needs to be there for some static assertions in ApplicationServer, fails to compile otherwise.

maierlars (approved):

PR #15657, opened 2022-02-01
Move dev deps into devDependencies

pluma:
### Scope & Purpose

This moves undocumented JS dependencies we only use during ArangoDB development (e.g. testing, build) into the `devDependencies` section of our `package.json` file to allow easier filtering of these during releases.

This has no effect on its own but allows us to adjust the release process to not rely on hardcoded lists and instead look at the actual JSON.

### Checklist

N/A

#### Related Information

- [x] Jira ticket: APM-277


mpoeter:
What about mocha?

pluma:
Replaced by #16542, which includes these changes.

PR #15664, opened 2022-02-01
Fix sanitizer issues in LogContext

mpoeter:
### Scope & Purpose

Fixes a potential memory leak and some data races.

- [x] :hankey: Bugfix

### DIFF
--- a/lib/Logger/LogContext.h
+++ b/lib/Logger/LogContext.h
@@ -402,11 +402,14 @@ struct LogContext::Entry {
  private:
   void incRefCnt() noexcept {
     TRI_ASSERT(_refCount.load() > 0);
+    // (1) - this acquire-fetch-add synchronizes-with the release-fetch-sub (2)
     _refCount.fetch_add(1, std::memory_order_acquire);
   }
   [[nodiscard]] std::size_t decRefCnt() noexcept {
     TRI_ASSERT(_refCount.load() > 0);
-    return _refCount.fetch_sub(1, std::memory_order_relaxed);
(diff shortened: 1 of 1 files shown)


jsteemann:
Last Jenkins run had a failure in LogContext destructor.

lib/Logger/LogContext.h
```diff
@@ -641,17 +646,14 @@ inline LogContext::~LogContext() {
   auto* t = _tail;
   while (t != nullptr) {
     auto prev = t->_prev;
-    if (t->_refCount == 1) {
-      // we have the only reference to this Entry, so we can "reuse" t's
+    if (t->_refCount.load(std::memory_order_relaxed) == 1 || t->decRefCnt() == 1) {
```
MBkkt, line 649: Always expect the last call we call decRefCnt, that we try to optimize?

mpoeter: Not sure if I understand the question. If `refCount == 1` (which is not unlikely) we hold the last reference to this entry. In this case we can omit the atomic decrement completely and just release the entry. Otherwise we perform the decrement but still need to check if the other references have been dropped in the meantime.

MBkkt: > which is not unlikely

My question is mostly about why this is true.
It's only once time, but "otherwise case" is more likely?
So for me this looks like we optimize unlikely case, and pessimize likely

mpoeter: Because there are probably many cases where we do not copy the LogContext and therefore have no additional references to that entry. If this assumption turns out to be incorrect, we can remove this optimization attempt later.

PR #15669, opened 2022-02-02
Improved performance of nested loop joins for vpack index

jsteemann:
### Scope & Purpose

Improved performance of inner joins with dynamic lookup conditions being injected from an outer loop, for indexes of type "persistent", "hash" and "skiplist" (internally all of these use the VPackIndex). Performance improvements can be expected if the inner join is invoked a lot of times with many different values fed in by the outer loop. The performance improvements are due to some improved handling of index lookup conditions in the internals of the VelocyPack-based index.

Example use case:
```js
unique = false; // optimization performed for both true and false
prefix = "test-der-hund-der-mag-es-bunt-";
n = 1 * 1000 * 1000;

db._create("v"); 
docs = []; 
for (let i = 0; i < n; ++i) { 
  docs.push({ value: prefix + i }); 
  if (docs.length === 5000) { 
    db.v.insert(docs); 
    docs = []; 
  } 
} 
db.v.ensureIndex({ type: "hash", fields: ["value"], unique}); 
db._profileQuery(`
  FOR i IN 0..${n - 1} 
    LET l = CONCAT('${prefix}', i) 
    FOR doc IN v 
      FILTER doc.value == l 
      RETURN doc.value
`);
```

Profile output with this PR:
```
Execution plan:
 Id   NodeType            Calls     Items   Filtered   Runtime [s]   Comment
  1   SingletonNode           1         1          0       0.00008   * ROOT
  2   CalculationNode         1         1          0       0.00008     - LET #3 = 0 .. 999999   /* range */   /* simple expression */
  3   EnumerateListNode    1000   1000000          0       0.19994     - FOR i IN #3   /* list iteration */
 10   IndexNode            1000   1000000          0       7.97476       - FOR doc IN v   /* hash index scan, index only, projections: `value` */    
  8   CalculationNode      1000   1000000          0       0.40689         - LET #7 = doc.`value`   /* attribute expression */   /* collections used: doc : v */
  9   ReturnNode           1000   1000000          0       0.03017         - RETURN #7

Query Statistics:
 Writes Exec   Writes Ign   Scan Full   Scan Index   Filtered   Peak Mem [b]   Exec Time [s]
           0            0           0      1000000          0          98304         8.64211
```

Profile output with regular devel (without this PR):
```
Execution plan:
 Id   NodeType            Calls     Items   Filtered   Runtime [s]   Comment
  1   SingletonNode           1         1          0       0.00006   * ROOT
  2   CalculationNode         1         1          0       0.00007     - LET #3 = 0 .. 999999   /* range */   /* simple expression */
  3   EnumerateListNode    1000   1000000          0       0.20470     - FOR i IN #3   /* list iteration */
 10   IndexNode            1000   1000000          0      11.97263       - FOR doc IN v   /* hash index scan, index only, projections: `value` */    
  8   CalculationNode      1000   1000000          0       0.41033         - LET #7 = doc.`value`   /* attribute expression */   /* collections used: doc : v */
  9   ReturnNode           1000   1000000          0       0.03028         - RETURN #7

Query Statistics:
 Writes Exec   Writes Ign   Scan Full   Scan Index   Filtered   Peak Mem [b]   Exec Time [s]
           0            0           0      1000000          0          98304        12.64905
```

This PR only changes some internals of the VPackIndex and IndexExecutor code. There is no end-user visible change except improved performance.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


tests/js/server/aql/aql-vpack-index-rearming.js
```diff
@@ -0,0 +1,205 @@
...
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Jan Steemann
+////////////////////////////////////////////////////////////////////////////////
+
+const db = require('@arangodb').db;
+const jsunity = require("jsunity");
+const { deriveTestSuite } = require('@arangodb/test-helper');
+
+function VPackIndexRearmingSuite (unique) {
```
mchacki, line 32: This test suite is good.
However I think It will also pass if the rearm optimization for some reason is not applied.
Can we add an additional quick test with a failure point inside the Rearm() implementation, such that we run the query (same as in existing tests) and expect the failure point to throw an error. Just to make sure no-one accidentally breaks the rearm logic in future versions.

CHANGELOG
```diff
@@ -1,10 +1,18 @@
...
+  "skiplist". Performance improvements can be expected if the inner join is
+  invoked a lot of times with many different values fed in by the outer loop.
+  The performance improvements are due to some improved handling of index
+  lookup conditions in the internals of the VelocyPack-based index.
+
+* Limited module resolution in arangosh to the path from which arangosh is 
+  invoked.
+
+* Fixed BTS-621 Fixed rare case of segfault in cluster during database recovery
+  if DBServer is in upgrade mode in the same time.
```
mchacki, line 15: These seem to be non-related Comments right?

Don't bother if this is some kind of merge conflict right now, we should just make sure we do not get something into the CHANGELOG that actually is not merged.

mchacki (approved):
I am asking for an additional test, to reduce risk in accidentally removing this improvement as a side-effect in future versions, which is a nice to have addition for me
Everything else LGTM.

CHANGELOG
jsteemann: It was an intentional change, because the 80 chars line length limit was not followed in these entries, but I removed this change from my PR for improved clarity.

tests/js/server/aql/aql-vpack-index-rearming.js
jsteemann: Added a way to track the number of cursor creations and rearmings via tests.

tests/js/server/aql/aql-vpack-index-rearming.js
```diff
@@ -0,0 +1,230 @@
...
+
+      for (let i = 0; i < n - 1; ++i) {
+        let doc = results[i];
+        assertEqual(prefix + String(i).padStart(5, "0"), doc.value1);
+        assertEqual(i, doc.value2);
+      }
+    },
+    
+    testLookupBySingleAttributeUsingIn : function () {
+      const q = `FOR i IN 0..${n - 2} FOR doc IN ${cn} FILTER doc.value1 IN [${pad}, 'piff'] RETURN doc`;
```
mpoeter, line 186: How about a test that combines `IN` on `value1` with some condition on `value2`?
What about queries like `FILTER (doc.value1 >= A AND doc.value1 < B) OR (doc.value1 >= X AND doc.value1 < Y)`?

jsteemann: Tests added.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -420,9 +531,10 @@ class RocksDBVPackIndexIterator final : public IndexIterator {
   arangodb::RocksDBVPackIndex const* _index;
   RocksDBVPackComparator const* _cmp;
   std::unique_ptr<rocksdb::Iterator> _iterator;
-  RocksDBKeyBounds const _bounds;
+  RocksDBKeyBounds _bounds;
   // used for iterate_upper_bound iterate_lower_bound
   rocksdb::Slice _rangeBound;
+  RocksDBVPackIndexSearchValueFormat _format;
```
mpoeter, line 537: This can probably be const.
```suggestion
  RocksDBVPackIndexSearchValueFormat const _format;
```

mpoeter (approved):
LGTM

PR #15685, opened 2022-02-03
Fix nullptr access with WINDOW operation.

mpoeter:
### Scope & Purpose

Fix null pointer access when using WINDOW operation with a COUNT/LENGTH aggregate function without any arguments.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**
- [x] Backports
  - [x] Backport for 3.9: *#15686*
  - [x] Backport for 3.8: *#15687*


arangod/Aql/WindowNode.cpp
```diff
@@ -358,7 +357,10 @@ void WindowNode::doToVelocyPack(VPackBuilder& nodes, unsigned flags) const {
       nodes.add(VPackValue("outVariable"));
       aggregateVariable.outVar->toVelocyPack(nodes);
       nodes.add(VPackValue("inVariable"));
```
jsteemann, line 359: Must be removed.
```suggestion
```

PR #15687, opened 2022-02-03
Fix nullptr access with WINDOW operation.

mpoeter:
### Scope & Purpose

Fix null pointer access when using WINDOW operation with a COUNT/LENGTH aggregate function without any arguments.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 v3.8.6 (XXXX-XX-XX)
 -------------------
-
+* Fix null pointer access when using WINDOW operation with a COUNT/LENGTH
+  aggregate function without any arguments.
+  
 * Fix potential access to dangling reference in cancellation of shard
   synchronization.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG
```diff
@@ -1,6 +1,8 @@
 v3.8.6 (XXXX-XX-XX)
 -------------------
-
+* Fix null pointer access when using WINDOW operation with a COUNT/LENGTH
```
mpoeter, line 3:
```suggestion

* Fix null pointer access when using WINDOW operation with a COUNT/LENGTH
```

arangod/Aql/WindowNode.cpp
```diff
@@ -349,7 +349,10 @@ void WindowNode::toVelocyPackHelper(VPackBuilder& nodes, unsigned flags,
       nodes.add(VPackValue("outVariable"));
       aggregateVariable.outVar->toVelocyPack(nodes);
       nodes.add(VPackValue("inVariable"));
```
mpoeter, line 351:
```suggestion
```

jsteemann (approved):
LGTM!

PR #15693, opened 2022-02-04
Bug fix/window executor in subqueries api hardening

mchacki:
### Scope & Purpose

This is a follow-up PR to https://github.com/arangodb/arangodb/pull/15690/
This PR does **not fix anything** or changes the behavior in anyway.
It is a counter-measurement to have the compiler help us detect unintentional use of the API as happened in the item fixed with linked PR.
No further cases were detected.

Side-Note: This PR also removed some functions that have not been in use anymore

Technical Description:
We now have two different States:
a) for the Executor (HASMORE, DONE) which is scoped to the context of a single subquery execution, (if one loop through the Subquery is over DONE is reached, if the next loop starts the state is reset to HASMORE).
b) For the Query in total (HASMORE, DONE) which scoped for the mainQuery to indicate that there is really not any data left from upstream anymore, there are also no further subquery iterations available from upstream. Hence this state will always be HASMORE, until we reach the end of the Stream, where it will switch to DONE, and can never be reset to HASMORE.

Now the API calls all return one of the above states matching their intentions, and by that avoid getting mixed up.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

No additional Tests! We did only change a Type here, no logic was touched.
There is no significant effect on the release (we have one enum more), no effect on user, so no CHANGELOG entry is made.
No Backports will be made, to avoid unnecessary risk on stable releases.

#### Related Information

No ticket for this.


arangod/Aql/WindowExecutor.cpp
```diff
@@ -242,7 +242,7 @@ ExecutorState WindowExecutor::consumeInputRange(
       return state;
     }
   }
-  return inputRange.finalState();
+  return inputRange.upstreamState();
```
goedderz, line 245: This is just here because the bug-fix PR #15695 isn't merged yet, right?

goedderz (approved):
LGTM, though I think this should be merged strictly after #15695!

arangod/Aql/WindowExecutor.cpp
mchacki: Yes correct

PR #15704, opened 2022-02-08
[CINFRA-249] Replicated Maintenance Actions Fast Track

maierlars:
### Scope & Purpose
This PR tries to fix a timeout in Jenkins:
1. Put the Replicated Log/State actions on the fast track. Otherwise it can happen that they are executed after SynchronizeShard actions. This leads to timeouts in the test.
2. Increase the test timeout to two minutes. Jenkins is not a realtime system.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


markuspf (approved):
LGTM

PR #15717, opened 2022-02-10
[CINFRA-255] Unconfigured Log Replicated State Fix

maierlars:
### Scope & Purpose
Rename `largestCommonIndex` to `lowestIndexToKeep`. The largest common index can actually fall if you add a new follower. We don't want this index to fall, hence `lowestIndexToKeep` fits better.

Furthermore, we don't create a replicated state on top of a replicated log, if the replicated log is still unconfigured. This is a temporary workaround and will later be replaced by a unconfigured state in the replicated state.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**

PR #15718, opened 2022-02-10
[CINFRA-253] Added an UnconfiguredState

goedderz:
### Scope & Purpose

This allows replicated states to be built while the log is still unconfigured. It replaces the provisional fix from #15717, in particular [this change in `vocbase.cpp`](https://github.com/arangodb/arangodb/pull/15717/files#diff-f545699f3f9f9bc59ffccf0c4d67b1e9240d7f62f3eca6bb2f16339546c4972d), which throws an exception when that happens.

 Instead, this PR introduces an `UnconfiguredStateManager` for `ReplicatedState`, corresponding to the `LogUnconfiguredParticipant` of its underlying `ReplicatedLog`. In order to notice when the participant actually does get configured, the method `ILogParticipant::waitForResign()` is introduced for all participants.

The following changes are done additionally:

- The `LeaderStateManager` previously used `waitFor(LogIndex{std::numeric_limits<std::uint64_t>::max()})` to wait for its `LogLeader` to resign. `waitForResign()` is now used instead.
- The `LogUnconfiguredParticipant` is moved into its own file
- A bugfix regarding `Future<Unit>`, where `setValue()` would set an uninitialized (i.e. invalid) `Try<Unit>`, instead of one holding a (`Unit`-) value.
- Bugfix in `ReplicatedState` `rebuild()`, `flush()` and `start()`, where a `DeferredAction` was executed before the lock was released.

### Checklist

- [X] Tests
  - [X] C++ **Unit tests**

#### Related Information

- [X] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-253


arangod/Replication2/ReplicatedState/UnconfiguredStateManager.tpp
```diff
@@ -0,0 +1,89 @@
...
+    : _parent(parent),
+      _unconfiguredParticipant(std::move(unconfiguredParticipant)),
+      _core(std::move(core)),
+      _token(std::move(token)) {}
+
+template<typename S>
+void UnconfiguredStateManager<S>::run() {
+  // We need to drop the lock before calling `thenFinal`, because it might
+  // immediately be executed and try to acquire the lock again.
+  // Thus execute it in the DeferredAction.
```
maierlars, line 46: This comment is pretty confusing here. I know what it means because I saw it yesterday before you moved the code. But future-me will just be confused.

arangod/Replication2/ReplicatedLog/LogLeader.h
```diff
@@ -146,6 +147,8 @@ class LogLeader : public std::enable_shared_from_this<LogLeader>,
 
   auto waitForLeadership() -> WaitForFuture override;
 
+  auto waitForResign() -> futures::Future<futures::Unit> override;
```
maierlars, line 150: I think we should annotate all the `waitFor` functions, that return a future, with ``[[nodiscard]]`. Otherwise one might assume that the call is blocking. (Has happened before)

maierlars (approved):

arangod/Replication2/ReplicatedLog/LogLeader.h
goedderz: True!

arangod/Replication2/ReplicatedState/UnconfiguredStateManager.tpp
goedderz: Agreed, will remove it.

goedderz:
https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr/20028/

PR #15739, opened 2022-02-15
Remove mocha, chalk

pluma:
### Scope & Purpose

This PR removes mocha and chalk. These are internal (undocumented) dependencies that cause unnecessary noise during audits and are no longer used as of 3.8 (see #14501, #14505).

- [x] :hankey: Bugfix

### Checklist

- [x] :book: CHANGELOG entry made
- [x] Backports
  - [x] Backport for 3.9: *(Please link PR)*

#### Related Information

- [x] GitHub issue / Jira ticket: ES-1015

PR #15740, opened 2022-02-15
[3.9] Remove mocha, chalk

pluma:
### Scope & Purpose

This is a backport of #15739 to 3.9 removing mocha and chalk. Since we no longer use these modules in 3.8 the upgrade should not be broken by this even though 3.7 and earlier depended on mocha and chalk to function.

- [x] :hankey: Bugfix

### Checklist

#### Related Information

- [x] GitHub issue / Jira ticket: ES-1015

@KVS85 we should probably make sure an upgrade from 3.8 or 3.9.0 to this PR will work as we saw breakage when attempting this before by upgrading from a version that still depended on mocha. This should no longer be the case, but better safe than sorry.

I'm tagging @dothebart and @jsteemann as they were involved in the original dependency breakage discussion and Jan applied the change that should prevent further issues when upgrading from 3.9 forward.

I'm additionally tagging @joerg84 and @Zyqsempai as this change might need PM approval: we would be removing a dependency in a bugfix version, but it is internal, undocumented and should not impact users.


dothebart (approved):
nice, tnx!

ajurdak (approved):
LGTM.

pluma:
Marking this as DO NOT MERGE for now, while we wait for a PM decision.

pluma:
Rerunning jenkins after merge.

Zyqsempai (approved):
LGTM

KVS85 (approved):
LGTM.

PR #15753, opened 2022-02-17
refactoring of index iterator API

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/885

Internal refactoring of IndexIterator API.
Simplifies several cases and removes the following APIs:
* `IndexIterator::hasExtra()`, `IndexIterator::nextExtra()`, `IndexIterator::allExtra()`: these were just special cases of covering index accesses on the edge index. The functionality is still available via `IndexIterator::nextCovering()`.
* `IndexIterator::hasCovering(): this is kind of a mismatching API, because an IndexIterator can cover certain fields, but not others. Removed for clarity reasons.
* `Index::hasCoveringIterator()`: now accessible via `Index::covers(Projections&)`.

Other things addressed:
* removed all support for extra fields from IResearchInvertedIndex. Extra fields were just there to make IResearchInvertedIndex support `nextExtra()`. But as it already provides `nextCovering()` and a `covers(Projections&)` implementation, everything should be there to support covering index scans.
* make Pregel edge loader reuse existing cursors when loading edges for different vertices. This saves lots of repeated cursor creations (one cursor per vertex) and should speed up edge loading when there are many vertices.
* make PrimaryIndex-based iterators support covering index scans on both `_key` and `_id` without ever fetching the full documents. This was only implemented partially before.
* slightly improve performance of covering edge index lookups: previously for each edge a velocypack array with _from and _to was created. This is now saved by using a custom IndexIteratorCoveringData implementation.
* slighty improve performance of rearming vpack index iterators: the new key bounds are now created inside the existing `_bounds` member, saving the repeated creation and destruction of key bounds.

This will not be backported.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/885
- [ ] Design document:


arangod/Aql/KShortestPathsNode.cpp
```diff
@@ -520,101 +520,44 @@ void KShortestPathsNode::getVariablesUsedHere(VarSet& vars) const {
...
+    bool res = aql::utils::getBestIndexHandleForFilterCondition(
+        *_edgeColls[i], clonedCondition, options()->tmpVar(), 1000,
+        aql::IndexHint(), indexToUse, onlyEdgeIndexes);
+    if (!res) {
+      THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL,
+                                     "expected edge index not found");
     }
+
+    indexAccessors.emplace_back(std::move(indexToUse),
+                                condition->clone(_plan->getAst()), 0, nullptr,
```
Dronplane, line 557: Why not use here `clonedCondition`? instead of another clone ?

jsteemann: That is a question that I also asked to myself, as currently we are copying twice.
However, that was already the case in the old code, and I am not sure if there are any correctness issues if we don't clone here.
I can check that in more detail...

jsteemann: Talked to @mchacki about it and pushed a change...

Dronplane (approved):
LGTM

arangod/Graph/BaseOptions.cpp
```diff
@@ -56,22 +56,39 @@ using namespace arangodb::aql;
...
 
 BaseOptions::LookupInfo::LookupInfo(arangodb::aql::QueryContext& query,
-                                    VPackSlice const& info,
-                                    VPackSlice const& shards) {
+                                    VPackSlice info, VPackSlice shards) {
   TRI_ASSERT(shards.isArray());
   idxHandles.reserve(shards.length());
 
+  TRI_edge_direction_e dir = TRI_EDGE_ANY;
+  VPackSlice dirSlice = info.get("direction");
```
hkernbach, line 77: I would prefer to use StaticStrings here instead.

arangod/Graph/BaseOptions.cpp
```diff
@@ -56,22 +56,39 @@ using namespace arangodb::aql;
...
+  if (dirSlice.isEqualString("inbound")) {
```
hkernbach, line 78: I would prefer to use StaticStrings here instead.

arangod/Graph/BaseOptions.cpp
```diff
@@ -56,22 +56,39 @@ using namespace arangodb::aql;
...
+    dir = TRI_EDGE_IN;
+  } else if (dirSlice.isEqualString("outbound")) {
```
hkernbach, line 80: I would prefer to use StaticStrings here instead.

arangod/Graph/BaseOptions.cpp
```diff
@@ -129,16 +146,28 @@ BaseOptions::LookupInfo::LookupInfo(arangodb::aql::QueryContext& query,
...
+  TRI_ASSERT(direction == TRI_EDGE_IN || direction == TRI_EDGE_OUT);
 }
 
 void BaseOptions::LookupInfo::buildEngineInfo(VPackBuilder& result) const {
   result.openObject();
+
+  // direction
+  TRI_ASSERT(direction == TRI_EDGE_IN || direction == TRI_EDGE_OUT);
+  if (direction == TRI_EDGE_IN) {
+    result.add("direction", VPackValue("inbound"));
```
hkernbach, line 166: I would prefer to use StaticStrings here instead.

arangod/Graph/BaseOptions.cpp
```diff
@@ -129,16 +146,28 @@ BaseOptions::LookupInfo::LookupInfo(arangodb::aql::QueryContext& query,
...
+  } else {
+    result.add("direction", VPackValue("outbound"));
```
hkernbach, line 168: I would prefer to use StaticStrings here instead.

arangod/Graph/SingleServerEdgeCursor.cpp
```diff
@@ -175,53 +180,66 @@ bool SingleServerEdgeCursor::next(EdgeCursor::Callback const& callback) {
...
-  // because the cursor expect's it to be filled.
+  // because the cursor expects it to be filled.
   do {
     if (cursorSet->empty() || !cursor->hasMore()) {
       if (!advanceCursor(cursor, cursorSet)) {
         return false;
       }
     } else {
-      if (cursor->hasExtra()) {
+      if (coveringPosition != aql::Projections::kNoCoveringIndexPosition) {
```
hkernbach, line 196: Moving that statement into a seperate function would help to maintain readability here. (From a technical point it is not required, but IMHO it would help improve the readability. Just by reading that comparison (without further knowledge) it is hard to understand (e.g. new developers).

arangod/Graph/Cursors/RefactoredSingleServerEdgeCursor.cpp
```diff
@@ -269,12 +306,20 @@ void RefactoredSingleServerEdgeCursor<Step>::readAll(
...
     LogicalCollection* collection = cursor.collection();
     auto cid = collection->id();
-    bool hasExtra = !_requiresFullDocument && cursor.hasExtra();
     auto* expression = lookupInfo.getExpression();
+    uint16_t coveringPosition = lookupInfo.coveringIndexPosition();
 
-    if (hasExtra) {
-      cursor.allExtra([&](LocalDocumentId const& token, VPackSlice edge) {
+    if (!_requiresFullDocument &&
+        coveringPosition != aql::Projections::kNoCoveringIndexPosition) {
```
hkernbach, line 313: Moving that statement into a seperate function would help to maintain readability here. (From a technical point it is not required, but IMHO it would help improve the readability. Just by reading that comparison (without further knowledge) it is hard to understand (e.g. new developers).

hkernbach (commented):
Graph related components review. Only minor comments added. Please feel free to either process them, or ignore in case they are too minor.

hkernbach (approved):
LGTM! Thanks for applying my suggestions 👍🏼

kvahed (approved):
LGTM

PR #15790, opened 2022-02-21
[BTS-800] Cluster not enabled exception

apetenchea:
### Scope & Purpose

This aims to fix the following error: `caught exception during start of feature 'FailureOracle': feature 'Cluster' is not enabled (exception location: /work/ArangoDB/lib/ApplicationFeatures/ApplicationServer.h:466)`.

It is suspected that during `UpgradeFeature` the `ClusterFeature` gets disabled, and thus other features which depend on it should too.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/FailureOracleFeature.cpp
+++ b/arangod/Cluster/FailureOracleFeature.cpp
@@ -215,7 +215,7 @@ void FailureOracleFeature::prepare() {
 void FailureOracleFeature::start() {
   TRI_ASSERT(_cache == nullptr);
   _cache = std::make_shared<FailureOracleImpl>(
-      server().getFeature<ClusterFeature>());
+      server().getEnabledFeature<ClusterFeature>());
   _cache->createAgencyCallback(server());
   _cache->start();
 
--- a/arangod/Cluster/FailureOracleFeature.cpp
+++ b/arangod/Cluster/FailureOracleFeature.cpp
@@ -202,6 +202,8 @@ FailureOracleFeature::FailureOracleFeature(Server& server)
   setOptional(true);
   startsAfter<SchedulerFeature>();
   startsAfter<ClusterFeature>();
+  onlyEnabledWith<SchedulerFeature>();
+  onlyEnabledWith<ClusterFeature>();
 }
 
 void FailureOracleFeature::prepare() {
@@ -213,7 +215,7 @@ void FailureOracleFeature::prepare() {
 void FailureOracleFeature::start() {
   TRI_ASSERT(_cache == nullptr);
   _cache = std::make_shared<FailureOracleImpl>(
-      server().getEnabledFeature<ClusterFeature>());
+      server().getFeature<ClusterFeature>());
   _cache->createAgencyCallback(server());
   _cache->start();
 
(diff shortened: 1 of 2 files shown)


dothebart (approved):

arangod/RestServer/UpgradeFeature.cpp
```diff
@@ -148,6 +149,10 @@ void UpgradeFeature::validateOptions(std::shared_ptr<ProgramOptions> options) {
       server().getFeature<ReplicationFeature>();
   replicationFeature.disableReplicationApplier();
 
+  cluster::FailureOracleFeature& failureOracleFeature =
```
mchacki, line 152: Can you please move this into the list of `_nonServerFeatures` that is handed in from the outside.This way we have it at least in the same position as the other force-disabled features.

arangod/Cluster/FailureOracleFeature.cpp
```diff
@@ -213,7 +215,7 @@ void FailureOracleFeature::prepare() {
 void FailureOracleFeature::start() {
   TRI_ASSERT(_cache == nullptr);
   _cache = std::make_shared<FailureOracleImpl>(
-      server().getEnabledFeature<ClusterFeature>());
+      server().getFeature<ClusterFeature>());
```
mchacki, line 218: This change should not be necessary with any of the other two changes.

arangod/Cluster/FailureOracleFeature.cpp
```diff
@@ -202,6 +202,8 @@ FailureOracleFeature::FailureOracleFeature(Server& server)
   setOptional(true);
   startsAfter<SchedulerFeature>();
   startsAfter<ClusterFeature>();
+  onlyEnabledWith<SchedulerFeature>();
+  onlyEnabledWith<ClusterFeature>();
```
mchacki, line 206: In my opinion this is the desired fix and should be the only one in this PR.

arangod/RestServer/UpgradeFeature.cpp
mchacki: I would actually prefer to not have this, but only have the "onlyEnabledWith<>".

mchacki (changes requested):
In this PR we have 3 additions to work around the same problem.

One is a workaround by just using "getFeature" instead of "getEnabledFeature" I would like to not have this workaround, as the other two should each fix the situation properly.

The onlyEnabledWith<> is my preferred solution, as it tells our dependency tree that the features should belong together and only work together. So this is future proof even if someone disables the cluster (or scheduler) somewhere else in the future.

Force disabling the feature will work as of today, but looks like a rather manual patch on one place, I would like to not include this.

arangod/RestServer/UpgradeFeature.cpp
apetenchea: Reverted this file back to devel version

arangod/Cluster/FailureOracleFeature.cpp
apetenchea: Reverted

mchacki (approved):
LGTM

PR #15804, opened 2022-02-23
[CINFRA-262] The FailureOracle should not regard BAD servers as FAILED

goedderz:
### Scope & Purpose

The FailureOracle should report only servers as failed which are marked as FAILED. Previously, it reported all non-GOOD (including BAD) servers as failed, which is inconsistent with their intended usage.

In particular, a ReplicatedLog could lower its effective writeConcern (from softWriteConcern), before the Supervision takes actions to replace a server.

- [X] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*
  - [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-262
- [ ] Design document:

PR #15805, opened 2022-02-23
[CINFRA-263] Don't regard failed servers as non-eligible

goedderz:
### Scope & Purpose

Currently, the eligible servers in `checkCommitIndex` exclude servers that are either

* excluded,
* don’t have a recent term, or
* failed.

Failed servers must not be excluded.

Reasoning:

First, why a failed server may be included: They will count only up to the LogIndex they reported anyway. And the ReplicatedLog should always assume that those entries are persisted. Only the Supervision should take action when that’s no longer safe to assume (by design) because a server is FAILED.

Second, if failed servers aren’t eligible, this may lead to stalls. These are the reason for the test failure in [CINFRA-261](https://arangodb.atlassian.net/browse/CINFRA-261) (together with the FailureOracle being over-eager and also regarding BAD as failed, see https://github.com/arangodb/arangodb/pull/15804 / [CINFRA-262](https://arangodb.atlassian.net/browse/CINFRA-262)). This happens if the following steps happen in this order:

1) A follower of a ReplicatedLog is marked as FAILED (or before #15804, also BAD)
2) The leader processes a response of that same follower
3) The follower is marked as GOOD again

In step 2, the commitIndex might not be increased in checkCommitIndex, because the follower is failed. However, if this was the last message in flight, and as long as no other inserts happen, it will stay this way. In this particular example, leadership never got established.

An alternative solution would be to introduce on-change callbacks in the FailureOracle, so the leader can react upon that change. However, that seems undesirable to me.

- [X] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [X] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-263
- [ ] Design document:


markuspf (approved):
LGTM

apetenchea (approved):

Issue #317, opened 2012-12-09
Feature Request (from Google Groups): DATE handling 


frankmayer:
I took the liberty to create a feature request out of this discussion: https://groups.google.com/forum/?fromgroups=#!topic/arangodb/3Wt4c44mc0I

Excerpt:
2) ArangoDB Server provides date handling functionality.

a) As date is a special type, I believe it could be handled that way. This can be provided by defining attributes in a collection as of special type 'date'. This could look like this:
     
```
db.myCollection.defineAttributeType('my.attribute', 'date');  
// This would for example set the nested attribute 'attribute' as a date field.
```

... for a nested document attribute inside a list, it could be:

```
db.myCollection.defineAttributeType('my[*].attribute', 'date');
  // This would for example set the nested attribute 'attribute' as a date field. ```
``` 
... or for a list that would contain dates:

```
db.myCollection.defineAttributeType('my[*]', 'date');
  // This would for example set the contents of a list as a date field
```
... mixing both could also be possible. So one could have a list of direct values and nested ones:

```
 {"my":[{"attribute":"2012-11-30T22:24:32.43Z"}, "2012-11-30T22:24:32.43Z"]}
```

b) The user would pass datetime strings like { "my":{"attribute":"2012-11-30T22:24:32.43Z"}} or maybe other formats like  { "my":{"attribute":"2012-11-30"}}... etc

      Internally, ArangoDB would convert and store those as timestamps.
      Knowing that it's a date field, it would automatically convert the datetime strings to timestamps for storing.

      That way, no special JSON handling is required by either the server or the client.
      Error handling: if the datetime string would have invalid data, the operation should be rejected.

c) With those would of course come some datetime functions to query (AQL)
      Simple queries  would probably not need anything special, as they are limited anyway.
      AQL queries:

```
DATE_MATH ("{mathcommand like add || subtract}", {date field to operate on, or a value}, {the modifying value , or a string like INTERVAL 1 WEEK'} );
 // Function to do math on a field or value
```
```
DATE_CURRENT (); Gets current date on server.
```

...maybe  more, that I haven't thought of just yet. But you get the idea :)
  

What do you think?

neandrake:
Would supporting BSON as storage type be beneficial here? That format supports defining date fields as well as binary (Issue #107) in addition to other types which might be beneficial such as 32/64bit integers, 64bit floating point (Issue #555) and regular expressions.

[...]

PR #15809, opened 2022-02-23
Speed up initial sync by prefetching data from leader

neunhoef:
This is a no longer extremely small performance tuning for the case of initial sync provided
the follower has already got some (or most) of the data. Rather than
synchronously fetching data from the leader and then applying it
locally, we prefetch some chunks and thus can write as fast as possible
with one thread and one RocksDB transaction. Unfortunately, this made further changes necessary
on the leader side. Namely, the API there was incapable of delivering revisions out of order.
This has been changed and hopefully further performance improvements have been achieved by using
next on the RocksDB iterator more and seek less.

- Tune initial sync by prefetching data from leader.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **integration tests** (covered by existing tests)
- [*] Backports: questionable


arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,138 @@ arangodb::Result fetchRevisions(
...
-    std::string request = requestBuilder->slice().toJson();
-
-    double t = TRI_microtime();
-    config.connection.lease(
-        [&](arangodb::httpclient::SimpleHttpClient* client) {
-          response.reset(client->retryRequest(arangodb::rest::RequestType::PUT,
-                                              url, request.data(),
-                                              request.size(), headers));
-        });
-    stats.waitedForDocs += TRI_microtime() - t;
```
jsteemann, line 232: `stats.waitedForDocs` will now always be 0. Can we populate it somehow, to get some insights into how long the document fetching takes?

neunhoef: Good point, I will think of something. Maybe only take the actual waiting time for futures.

neunhoef: Done.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
-      VPackArrayBuilder list(requestBuilder.get());
-      for (std::size_t i = 0; i < 5000 && current + i < toFetch.size(); ++i) {
-        requestBuilder->add(toFetch[current + i].toValuePair(ridBuffer));
+  std::deque<arangodb::network::FutureRes> futures;
+  std::deque<std::unordered_set<arangodb::RevisionId>> shoppingLists;
+
+  arangodb::network::ConnectionPool* pool = netFeature.pool();
+
+  std::size_t queueSize = 10;
+  if ((config.leader.majorVersion == 3 && config.leader.minorVersion < 9) ||
```
jsteemann, line 220: nit: what about `config.leader.majorVersion < 3` here?
It's not that anything would work then, but in theory this case is not covered here.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+      reqOptions.timeout = arangodb::network::Timeout(25.0);
+      auto buffer = requestBuilder.steal();
+      auto f = arangodb::network::sendRequestRetry(
+          pool, config.leader.endpoint, arangodb::fuerte::RestVerb::Put, path,
+          std::move(*buffer), reqOptions);
+      futures.emplace_back(std::move(f));
+      shoppingLists.emplace_back(std::move(shoppingList));
+      ++stats.numDocsRequests;
+      LOG_TOPIC("eda42", DEBUG, arangodb::Logger::REPLICATION)
+          << "Have requested a chunk of 5000 revisions from "
```
jsteemann, line 254: nit: the chunk can have a size < 5000 as well. In this case the log message will be (slightly) misleading. It's not important, just mentioning it here.
```suggestion
          << "Have requested a chunk of 5000 revisions from "
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+          for (auto it = newList.begin(); it != newList.end(); ++it) {
+            requestBuilder.add(it->toValuePair(ridBuffer));
+          }
+        }
+
+        arangodb::network::RequestOptions reqOptions;
+        reqOptions.param("collection", leader)
+            .param("serverId", state.localServerIdString)
+            .param("batchId", std::to_string(config.batch.id));
+        reqOptions.timeout = arangodb::network::Timeout(25.0);
```
jsteemann, line 383: the request should be restricted to the correct database, i.e.
```suggestion
        reqOptions.timeout = arangodb::network::Timeout(25.0);
        reqOptions.database = config.vocbase.name();      
```

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+        auto buffer = requestBuilder.steal();
+        auto f = arangodb::network::sendRequestRetry(
+            pool, config.leader.endpoint, arangodb::fuerte::RestVerb::Put, path,
+            std::move(*buffer), reqOptions);
+        futures.emplace_back(std::move(f));
+        shoppingLists.emplace_back(std::move(newList));
+        ++stats.numDocsRequests;
+        LOG_TOPIC("eda45", DEBUG, arangodb::Logger::REPLICATION)
+            << "Have re-requested a chunk of " << newList.size()
```
jsteemann, line 392: `newList` was moved before this line, so we should not read its size.
```suggestion
            << "Have re-requested a chunk of " << shoppingLists.back().size()
```

arangod/Replication/DatabaseInitialSyncer.cpp
neunhoef: Good point, done.

arangod/Replication/DatabaseInitialSyncer.cpp
neunhoef: Introduced a local counter. Good point.

arangod/Replication/DatabaseInitialSyncer.cpp
neunhoef: Ooops. This is important.

neunhoef: Done.

arangod/Replication/DatabaseInitialSyncer.cpp
neunhoef: Oopsie. Done.

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -3319,11 +3319,39 @@ void RestReplicationHandler::handleCommandRevisionDocuments() {
...
+        it.next();
+        if (!it.hasMore() || it.revision() != rev) {
+          // If we did not find what we want with it.next(), then we try
+          // an it.seek(rev) next and that is our last try:
+          it.seek(rev);
+        }
+      } else if (it.revision() > rev) {
+        // This is the case that the iterator is not exhausted, but is
+        // already further than the client. This happens, if the client
+        // "jumps back". Note that his can happen, since the client
```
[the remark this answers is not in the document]
neunhoef: Unfortunately, with the new prefetch, the iterator has to sometimes go back, here is why: We ask for 5000 documents but due to oversized documents, the leader only sends the first 316 documents. In the meantime, the follower has already asked for the next 5000 documents further back. But then it has to go back to the first chunk and ask again for documents #317...#5000. This means that one of the next requests requires the iterator to go back. This has actually happened in tests and broke some. Therefore, if we use prefetch, then the iterator must be able to jump back.

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -3319,11 +3319,39 @@ void RestReplicationHandler::handleCommandRevisionDocuments() {
...
+        // might not have got all ordered documeuts due to the chunkSize
```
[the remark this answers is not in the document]
neunhoef: Done.

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -3319,11 +3319,39 @@ void RestReplicationHandler::handleCommandRevisionDocuments() {
 
     for (VPackSlice entry : VPackArrayIterator(body)) {
       RevisionId rev = RevisionId::fromSlice(entry);
-      if (it.hasMore() && it.revision() < rev) {
+      if (!it.hasMore()) {
+        // If the iterator is exhausted, let's try to seek, this can happen
+        // if the client has jumped back. Normally, the iterator will never
+        // be exhausted, but we must check, since we are not even allowed
+        // to call it.revision() if the iterator is exhausted!
         it.seek(rev);
```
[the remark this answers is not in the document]
neunhoef: We come in here with a rev. We assume that this rev is in the collection. In that case, the seek will always lead to the fact that the iterator moves to that very document and then `hasMore` is true and `isValid`, too. The provision with the empty document is only for the unlikely and under normal circumstances impossible case that the follower has requested a revision which is not actually present in the snapshot. I think in this case it is OK to deliver an empty document instead.

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -3319,11 +3319,39 @@ void RestReplicationHandler::handleCommandRevisionDocuments() {
...
-      }
-      VPackSlice res =
-          it.hasMore() ? it.document() : velocypack::Slice::emptyObjectSlice();
+      } else if (it.revision() < rev) {
+        // This is the normal case, in most cases, the client will ask for
+        // a whole range of adjacent revisions, so we optimize the fast
+        // path for this, so we try it.next() first, since this is cheaper
+        // than a seek:
+        it.next();
+        if (!it.hasMore() || it.revision() != rev) {
```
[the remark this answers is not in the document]
neunhoef: The API's purpose is to deliver a list of revisions for which the follower has asked. I think the API should work with whatever list of revisions it gets. The normal case is that the follower fetches missing revisisions in ascending order and we can usually assume that if it needs a revision, it will probably need the next one next, but this is not clear. The follower might have holes in its list of revisions. So we optimize performance by:
 - not moving the iterator if we happen to be in the right place, since this is a cheap check
 - using `next` for the common case that the next rev the follower asks for is the next one
 - if this is not the case, then we must do an additional seek to get the right thing.
As far as I see, this works always and has a good fast path (next vs. seek) for the most common case.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+            << " revisions from " << config.leader.serverId << " at "
```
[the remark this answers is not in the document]
neunhoef: Done.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+      ++stats.numDocsRequests;
+      LOG_TOPIC("eda42", DEBUG, arangodb::Logger::REPLICATION)
+          << "Have requested a chunk of 5000 revisions from "
+          << config.leader.serverId << " at " << config.leader.endpoint
+          << " for collection " << leader
+          << " length of queue: " << futures.size();
+    }
+
+    if (!futures.empty()) {
+      auto& f = futures.front();
```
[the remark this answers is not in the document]
neunhoef: Done.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -210,130 +209,203 @@ arangodb::Result fetchRevisions(
...
+            << " queue length: " << futures.size();
+      }
+      LOG_TOPIC("eda44", DEBUG, arangodb::Logger::REPLICATION)
+          << "Have applied a chunk of " << docs.length() << " documents from "
+          << config.leader.serverId << " at " << config.leader.endpoint
+          << " for collection " << leader
+          << " queue length: " << futures.size();
+      futures.pop_front();
+      shoppingLists.pop_front();
+      res = trx.state()->performIntermediateCommitIfRequired(collection.id());
```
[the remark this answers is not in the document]
neunhoef: Done.

kvahed (approved):
LGTM

jsteemann (approved):
LGTM!

PR #15776, opened 2022-02-18
Bug fixdevel/only admin see license info

ajurdak:
### Scope & Purpose

Non-root users has no access to the api endpoint "/_admin/time" which we use to calculate and display license info below the navigation bar and therefore returns a 403 error. Meaning those users got the info "Error: Failed to fetch license information" all the time what is bad user experience. As this license information is just an information for the user we removed error handling at fetching the license information to not show errors in the Web UI. In this specific case we improve the user experience in not showing errors. 

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **manually tested**
- [x] :book: CHANGELOG entry made
- [x] Backports
  - [x] Backport for 3.9: #15813

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-48

PR #15813, opened 2022-02-24
Remove error handling fetching license info to improve user experience

ajurdak:
### Scope & Purpose

Backport for #15776

Non-root users has no access to the api endpoint "/_admin/time" which we use to calculate and display license info below the navigation bar and therefore returns a 403 error. Meaning those users got the info "Error: Failed to fetch license information" all the time what is bad user experience. As this license information is just an information for the user we removed error handling at fetching the license information to not show errors in the Web UI. In this specific case we improve the user experience in not showing errors. 

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **manually tested**

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-48


Zyqsempai (approved):
LGTM

js/apps/system/_admin/aardvark/APP/frontend/js/views/navigationView.js
```diff
@@ -160,13 +160,11 @@
           if (!timeData.error && timeData.code === 200 && timeData.time) {
             self.fetchLicenseInfo(timeData.time);
           } else {
-            self.showGetTimeError();  
-          }
-        },
-        error: function () {
-          self.showGetTimeError();
+            self.showGetTimeError();
```
hkernbach, line 163: `/_admin/time` can only be accessed through `_system` database, correct?
Therefore why do that request only then if a user is logged into the system database? 
You can check that through the global accessible variable `frontendConfig`.

hkernbach: Not we're sending that request every time - and we know that for most users it will fail.

joerg84: @ajurdak @Zyqsempai what do we need to proceed here?

ajurdak: Check for _system database was added in a separate PR.

KVS85 (approved):
LGTM.

PR #15816, opened 2022-02-24
Bug fix: Deadlock under heavy write load, with getting a new follower in sync

mchacki:
### Scope & Purpose

Note to reviewer: Please take extra care while reviewing this PR. It does not entail many code changes but they do have side-effect impacts.

This PR mainly has one purpose, but introduces some "side-fixes as well":
*Main Purpose:*
While under very high load, on a shard, a distributed deadlock could be reached:
1) For the final syncing of two shards, we get a "hardLock " (exclusive) on the leader shard
2) Now we flood the queue with write requests to that shard (they will be started to work on and block on the lock, waiting for the lock to be received)
3) After the hardLock is acquired the follower needs to do some further requests (e.g. syncing the remainder of documents, adding ourselves to the follower list)
=> Before this PR some operations in 3) have an identical or lower priority than 2) (e.g.: the way tailing was on LOW). Hence given a high enough load, they had to be queued AFTER the writes are completed.

This was fixed by:
1) Making sure that all requests to the leader following a HARD lock in replication are placed on HIGH priority lane.
2) Making sure on the follower that the response gathering for request send with or after the HARD lock either bypass the scheduler or are continued on HIGH priority (to avoid the follower being a leader to a different shard and stuck in the same situation not working on getting in sync again.)

What was added on the side:
1) A new DEBUG log line on requests, that will print on which Lane / Priority a request to a given URL is posted, which could help for further debugging of distributed Deadlock situations caused by suboptimal settings of priorities.
`P DEBUG [ecd0a] {requests} Handling request 0x7fc20100a9a8 on path /_api/collection/s10014/count on lane CLIENT_SLOW with priority: LOW
`
2) Raise the Priority of `_admin/debug` to HIGH (this is test and internal only, and required to e.g. clear failurePoints) this is required for the test to complete (and unblock itself)
3) Let arangosh send the "x-arango-frontend" header on requests to db._collections(). This request is performed on startup of the arangosh, if the system is under high load (or blocked by above scenario) would cause the arangosh to queue behind the write operations without the header. (This is again required for the test to switch between blocked servers, but I think it is actually good in general, we want to always be able to connect with arangosh in emergency situations.)

*How the attached test works*:
1) We create a new collection, insert some documents and then increase the replicationFactor to trigger the "getting in sync" procudure.
a) For security trying to get into sync using only a softLock, will always cause a CHECKUM missmatch, so we are forced into the hardLock case, even if we can find a shortCut under softLock.
2) We simulate the system being super busy by blocking the scheduler to pull tasks from MEDIUM (or lower) priority queues (This simulates a MEDIUM lane being filled with unrelated tasks, or even blocked tasks, s.t. our replication cannot continue if it would be bound to one of those Priorities)
  a) The Follower is blocking MEDIUM right before it sends out the hardLock request (to make sure the response can be picked up under high load)
  b) The leader is blocking MEDIUM as soon as it is asked to set the hardLock.
3) Even with blocked Medium lanes the Shards now need to get into Sync under the hardLock

- [X] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket:
  (https://arangodb.atlassian.net/browse/BTS-799)
  (https://arangodb.atlassian.net/browse/MDS-930)
- [ ] Design document:


tests/js/client/shell/shell-replication-deadlock-regression-cluster.js
```diff
@@ -0,0 +1,110 @@
...
+      // 3.) On the leader, as soon as we respond to the hardLock request, we block MED and LOW lanes and simulate high load
+      // 4.) Then we wait until the shards are back into sync
+
+      const col = db._collection(collectionName);
+      const shardName = col.shards()[0];
+      // Insert some random documents, to avoid getting away with a fast-path sync
+      col.save([{}, {}, {}, {}]);
+
+      for (const server of getDBServers()) {
+        // TODO activate Failure Points
```
mchacki, line 83:
```suggestion
```

kvahed (approved):

PR #15827, opened 2022-02-28
[CINFRA-247] Replicated State Status API

maierlars:
### Scope & Purpose

This PR adds an HTTP API to query the status of a replicated state.


arangod/Replication2/Methods.cpp
```diff
@@ -591,6 +593,70 @@ struct ReplicatedLogMethodsCoordinator final
...
+                       ResultT<uint64_t>&& res) -> futures::Future<Result> {
+          if (res.fail()) {
+            return futures::Future<Result>{std::in_place, res.result()};
+          }
+
+          return self->clusterInfo.waitForPlan(res.get());
+        });
+  }
+
+  auto deleteReplicatedLog(LogId id) const -> futures::Future<Result> override {
```
apetenchea, line 645: This is not important for the current functionality, but just for my understanding, I assume the intention will be to delete the replicated state.

tests/js/server/replication2/replication2-supervision-replicated-state-target-cluster.js
```diff
@@ -93,14 +91,45 @@ const replicatedStateSuite = function () {
...
+            }
+          };
+        });
+
+    lh.waitFor(spreds.replicatedStateIsReady(database, stateId, servers));
+    const leader = lh.getReplicatedLogLeaderPlan(database, stateId).leader;
+    const followers = _.difference(servers, [leader]);
+    return {stateId, servers, leader, followers};
+  };
+
```
apetenchea, line 122: This is the same as in replicated-state-helper. It's good, but I'm making a note to replace it in my upcoming PR.

apetenchea (approved):

PR #15839, opened 2022-03-02
Fix SEARCH-294: remove DataStore when not valid initialization

MBkkt:
### Scope & Purpose

Fix SEARCH-294: remove DataStore when not valid initialization
Also, fix possible leak physical DataStore directory

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -874,41 +874,29 @@ Result IResearchDataStore::consolidateUnsafeImpl(
   return {};
 }
 
-Result IResearchDataStore::shutdownDataStore() {
-  std::atomic_store(&_flushSubscription, {});
-  // reset together with '_asyncSelf'
-  _asyncSelf->reset();
-  // the data-store is being deallocated, link use is no longer valid
-  // (wait for all the view users to finish)
-
+void IResearchDataStore::shutdownDataStore() noexcept {
+  std::atomic_store(&_flushSubscription, {});  // reset together with _asyncSelf
   try {
+    // the data-store is being deallocated, link use is no longer valid
+    _asyncSelf->reset();  // wait for all the view users to finish
     if (_dataStore) {
-      removeMetrics();
-      _dataStore.resetDataStore();
+      removeMetrics();  // TODO(MBkkt) Should be noexcept?
     }
-  } catch (basics::Exception const& e) {
-    return {e.code(),
-            "caught exception while unloading arangosearch data store '" +
-                std::to_string(id().id()) + "': " + e.what()};
   } catch (std::exception const& e) {
-    return {TRI_ERROR_INTERNAL,
-            "caught exception while unloading arangosearch data store '" +
-                std::to_string(id().id()) + "': " + e.what()};
+    LOG_TOPIC("bad00", ERR, TOPIC)
+        << "caught exception while waiting reset arangosearch data store '"
+        << std::to_string(id().id()) << "': " << e.what();
   } catch (...) {
-    return {TRI_ERROR_INTERNAL,
-            "caught exception while unloading arangosearch data store '" +
-                std::to_string(id().id()) + "'"};
+    LOG_TOPIC("bad01", ERR, TOPIC)
+        << "caught something while waiting reset arangosearch data store '"
+        << std::to_string(id().id()) << "'";
   }
-  return {};
+  _dataStore.resetDataStore();
 }
 
-Result IResearchDataStore::deleteDataStore() {
-  auto res = shutdownDataStore();
-  if (res.fail()) {
-    return res;
-  }
+Result IResearchDataStore::deleteDataStore() noexcept {
+  shutdownDataStore();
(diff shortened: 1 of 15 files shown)


arangod/IResearch/IResearchLink.cpp
```diff
@@ -359,6 +358,7 @@ Result IResearchLink::init(velocypack::Slice definition,
             "error finding view for link '" + std::to_string(_id.id()) + "'"};
   }
   TRI_ASSERT(_meta._sortCompression);
+  auto failedInit = scopeGuard([&]() noexcept { _viewGuid = {}; });
```
gnusi, line 361: can't we explicitly call `deleteDataStore()` here?

MBkkt: I don't want try catch

arangod/IResearch/IResearchLink.cpp
```diff
@@ -459,18 +461,13 @@ bool IResearchLink::setCollectionName(irs::string_ref name) noexcept {
...
 Result IResearchLink::unload() {
-  // this code is used by the MMFilesEngine
-  // if the collection is in the process of being removed then drop it from the
-  // view
-  // FIXME TODO remove once LogicalCollection::drop(...) will drop its indexes
-  // explicitly
-  if (_collection.deleted()  // collection deleted
-      || TRI_vocbase_col_status_e::TRI_VOC_COL_STATUS_DELETED ==
-             _collection.status()) {
+  if (_viewGuid.empty()  // link isn't linked in view and init wasn't successful
```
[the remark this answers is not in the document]
MBkkt: Fixed

tests/IResearch/IResearchInvertedIndexIteratorTest.cpp
```diff
@@ -135,7 +135,8 @@ class IResearchInvertedIndexIteratorTestBase
     _index = std::make_shared<arangodb::iresearch::IResearchInvertedIndex>(
         id, *_collection, std::move(meta));
     EXPECT_TRUE(_index);
-    EXPECT_TRUE(_index->init().ok());
+    bool pathExist = false;
```
Dronplane, line 138: Suggest to verify `pathExist` value

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -919,7 +907,8 @@ Result IResearchDataStore::deleteDataStore() {
 }
 
 Result IResearchDataStore::initDataStore(
-    InitCallback const& initCallback, uint32_t version, bool sorted,
+    bool& pathExist, InitCallback const& initCallback, uint32_t version,
```
Dronplane, line 910: NIT: pathExists looks like more correct name.
```suggestion
    bool& pathExists, InitCallback const& initCallback, uint32_t version,
```

tests/IResearch/IResearchInvertedIndexIteratorTest.cpp
MBkkt: Why?

arangod/IResearch/IResearchDataStore.cpp
MBkkt: Yep, but clang format make code very ugly in some places

tests/IResearch/IResearchInvertedIndexIteratorTest.cpp
Dronplane: Just to verify it works. Or  is there a test for that already added that I`ve missed?

MBkkt: I think we don't need check something here. We use pathExist flag only init return failed result or throw exception

arangod/IResearch/IResearchDataStore.cpp
MBkkt: Done

tests/IResearch/IResearchInvertedIndexIteratorTest.cpp
Dronplane: But how then we make sure it works?

MBkkt: Verify what we just take internal local variable, and make it output variable

We don't change any behaviour in these lines

MBkkt: Maybe default initialization instead of `false` will be look better

Dronplane: I mean just to check that it is true when it should be true and false otherwise. But indeed maybe in separate test.

Dronplane (approved):
LGTM

PR #15841, opened 2022-03-02
less copying of repeatedly used bind parameter values

jsteemann:
### Scope & Purpose

Less copying of repeatedly used bind parameter values in AQL queries.
This reduces memory overhead for queries that access large bind parameter values (large arrays or objects) via repeatedly refering to them by name (e.g. `FOR doc IN collection FILTER doc.value == CONCAT(@largeBind['foo'], @largeBind['bar'], @largeBind[...], ...`).
The bind parameter values are only once converted into an AstNode per query, and from the on only shallow-cloned when accessed.
In some example query this decreased memory usage by 99%.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Ast.cpp
+++ b/arangod/Aql/Ast.cpp
@@ -1842,6 +1842,14 @@ void Ast::injectBindParameters(
               // finally note that the node was created from a bind parameter
               node->setFlag(FLAG_BIND_PARAMETER);
 
+              // register the AstNode for this bind parameter, so when the query
+              // string refers to the same bind parameter multiple times, we
+              // don't have to regenerate an AstNode for it. in case a bind
+              // parameter is used multiple times in the query string, upon any
+              // following occurrences the AstNode registered here will be
+              // found, and only a shallow copy of the existing AstNode will be
+              // created. This helps to save memory and processing time for
+              // large bind parameter values.
               parameters.registerNode(param, node);
             }
           }
@@ -1910,9 +1918,17 @@ void Ast::injectBindParameters(
               }
             }
           }
+
+          TRI_ASSERT(newNode != nullptr);
           node = newNode;
 
           if (cachedNode == nullptr) {
+            // store the just created AstNode for this bind parameter, to mark
+            // the bind parameter as being "used". It does not matter which
+            // AstNode we store for the bind parameter, but we have to store one
+            // if none was yet registered. Otherwise we'll run into an error
+            // "unused bind parameter
+            // '@...' later.
             parameters.registerNode(param, node);
           }
         }
--- a/arangod/Aql/Ast.cpp
+++ b/arangod/Aql/Ast.cpp
@@ -46,6 +46,7 @@
 #include "Basics/tryEmplaceHelper.h"
 #include "Cluster/ClusterFeature.h"
 #include "Cluster/ClusterInfo.h"
+#include "Containers/FlatHashSet.h"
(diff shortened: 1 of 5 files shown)


arangod/Aql/BindParameters.cpp
```diff
@@ -65,28 +66,42 @@ uint64_t BindParameters::hash() const {
   return _builder->slice().hash();
 }
 
-/// @brief mark a bind parameter as "used", and return its value.
+/// @brief return a bind parameter value by name
 /// will return VPackSlice::noneSlice() if the bind parameter does not exist!
-VPackSlice BindParameters::markUsed(std::string const& name) noexcept {
+std::pair<VPackSlice, AstNode*> BindParameters::get(
```
mpoeter, line 71: It no longer returns only a value, so the comment should be updated accordingly.

arangod/Aql/BindParameters.cpp
```diff
@@ -65,28 +66,42 @@ uint64_t BindParameters::hash() const {
...
   if (it == _parameters.end()) {
-    return VPackSlice::noneSlice();
+    return std::make_pair(VPackSlice::noneSlice(), nullptr);
   }
 
-  // mark the bind parameter as being used
-  (*it).second.second = true;
+  // return parameter value and AstNode
+  TRI_ASSERT(!(*it).second.first.isNone());
+  return std::make_pair((*it).second.first, (*it).second.second);
```
mpoeter, line 82: Isn't this just the same as
```suggestion
  return it->second;
```

arangod/Aql/Ast.cpp
```diff
@@ -1888,6 +1911,10 @@ void Ast::injectBindParameters(
             }
           }
           node = newNode;
+
+          if (cachedNode == nullptr) {
+            parameters.registerNode(param, node);
+          }
```
mpoeter, line 1917: What if `cacheNode != nullptr`? Couldn't we also just lookup the param and use a previously registered datasource node instead?

mpoeter (approved):
A few small comments, but overall LGTM.

arangod/Aql/BindParameters.cpp
jsteemann: Done

arangod/Aql/BindParameters.cpp
jsteemann: Absolutely... :facepalm:

arangod/Aql/Ast.cpp
jsteemann: That _may_ be true, but I am unaware of the side effects for this type of node.
At least it does no harm to always create a new AstNode in this case, as the AstNode will always have a small size (in contrast to arbitrarily-sized bind parameter values from JSON/VPACK).
I added more comments to the code to make the intentions more clear.

mpoeter: Ok... but the newly created nodes should be equivalent to any cached ones. Should we at least add some assertions to check that?

jsteemann: Added assertions in latest commit.

PR #15856, opened 2022-03-07
[CINFRA-268] Add additional information for CommitFailReason

goedderz:
### Scope & Purpose

Give more information in `CommitFailReason` to better understand why a replicated log might be lagging behind.

- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [X] C++ **Unit tests**

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-268


lib/Containers/SmallVector.h
```diff
@@ -23,15 +23,22 @@
...
 
-namespace arangodb {
-namespace containers {
+namespace arangodb::containers {
+
+#ifndef _DEBUG
+constexpr static auto minAlignment = std::size_t{1};
+#else
+// Make Debug-MSVC happy
+constexpr static auto minAlignment = alignof(std::_Container_proxy);
```
jsteemann, line 37: Is `std::_Container_proxy` available on all platforms when `_DEBUG` is defined?
Should it be wrapped into another ifdef that checks for MSVC?

goedderz: I assume that it's only set for MSVC, as it's a flag used in MSVC's STL specifically, afaik. However, it still seems prudent to check for both MSVC and _DEBUG; I will add it.

maierlars (approved):

PR #15857, opened 2022-03-07
[CINFRA-223] Added timestamp strings to state snapshot.

maierlars:
### Scope & Purpose

Adds timestamp s to the Snapshot and StateStatus serialization functions.


maierlars:
Failures unrelated

PR #15860, opened 2022-03-08
Bug-Fix /_api/user/<user>/config

hkernbach:
### Scope & Purpose

This PR fixes the PUT API of URL: `/_api/user/${user}/config`.
This has been broken since 2018.

The API logic wrote into the wrong data structure. Data has been written into `userData` instead of `userConfig`. 
As there were no tests, I've added a new suite which tests that API now for the root user, non-root user, system db and non-system db (PUT: write new, write additional, overwrite, merge and DELETE).

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :books: documentation not needed
- [x] Backports
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/15861
  - [x] Backport for 3.8: https://github.com/arangodb/arangodb/pull/15862
  - [x] Backport for 3.7: https://github.com/arangodb/arangodb/pull/15863
#### Related Information

- [x] Docs PR: no docs PR required
- [x] Enterprise PR: no enterprise PR required
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/ES-1078


tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,216 @@ describe('Grants', function() {
...
+    c: 1337,
+    someNewParam: "MJ7"
+  };
+
+  const mergeConfigObject = {
+    "thisMustStay": true
+  };
+
+  const verifyEmptyConfig = (result) => {
+    console.warn(result);
```
jsteemann, line 169: Ideally tests shouldn't log requests/responses unless they are flaky.
```suggestion
```

tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,216 @@ describe('Grants', function() {
...
+    return `/_db/${database}/_api/user/${user}/config`;
+  };
+
+  const generateWriteConfigUrl = (user, database, attribute) => {
+    if (!attribute) {
+      attribute = configPropertyAttribute;
+    }
+    if (database === rootDB) {
+      return `/_api/user/${user}/config/${attribute}`;
+    }
```
jsteemann, line 218: Isn't this covered by the general case already?
```suggestion
```

tests/js/common/http/grants-spec.js
hkernbach: Good! jslint did not complain :(

tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,216 @@ describe('Grants', function() {
...
+    expect(result).to.have.property('e', "aString");
+    expect(result.f).to.be.an('array');
+    expect(result.g).to.be.an('object');
+  };
+
+  const verifyOverwriteConfigObject = (result) => {
+    expect(result.error).to.be.false;
+    expect(result).to.have.property('code', 200);
+
+    result = result.result[configPropertyAttribute];
```
jsteemann, line 192: Should this also verify the "old", non-overwritten attributes (e.g. 'a', 'b', 'c', 'd', 'e', 'f', 'g')?
I understand that overwriting data is just _merging_ values together, so the previous data in non-touched attributes should remain unchanged.

hkernbach: No - this might be a bit misleading. Maybe I should add another comment to this. 
First we initialize "graphs" object, then we overwrite the "graphs" object completely. Then we "merge/ add another attribute to the top level next to graphs".

jsteemann: Ok. I think adding a code comment could help indeed.

tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,209 @@ describe('Grants', function() {
...
+    });
+    verifyEmptyConfig(JSON.parse(response.body));
+  };
+
+  it('should store/patch/delete root config data in _system database', function () {
+    testSuiteHelper(rootUser, rootDB);
+  });
+
+  it('should store/patch/delete root config data in non-system database', function () {
+    testSuiteHelper(rootUser, nonRootDB);
```
jsteemann, line 328: Side node: there is no cleanup for the root user's data.
The test relies on that the cleanup in line 310 works. If it ever fails, the second test that is testing the root user's data will see old values and may return false results.

tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,216 @@ describe('Grants', function() {
...
+    expect(result.a).to.be.true;
+    expect(result.b).to.be.false;
+    expect(result).to.have.property('c', 1);
+    expect(result).to.have.property('d', 1.1);
+    expect(result).to.have.property('e', "aString");
+    expect(result.f).to.be.an('array');
+    expect(result.g).to.be.an('object');
+  };
+
+  const verifyOverwriteConfigObject = (result) => {
```
hkernbach, line 188:
```suggestion
  const verifyOverwriteConfigObject = (result) => {
  // We do have one "global" property we can modify. If we write to that same top-level attribute again, it will be
  // fully replaced by the supplied object given. In case we write to another top-level attribute, it will be merged with
  // the other already available ones (This is tested in verifyMergedConfigObject).
```

tests/js/common/http/grants-spec.js
hkernbach: Added

tests/js/common/http/grants-spec.js
hkernbach: Thanks. Also added in: b481f7a013da7b072ec87d13740ed41d1dfb4469

tests/js/common/http/grants-spec.js
```diff
@@ -134,3 +135,219 @@ describe('Grants', function() {
...
+  });
+
+  afterEach(function () {
+    try {
+      users.remove(nonRootUser);
+    } catch (ignore) {
+    }
+
+    // cleanup root users config we might have been modified in any of our tests
+    const cleanupRootUrl = generateWriteConfigUrl(rootUser, rootDB);
```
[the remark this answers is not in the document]
hkernbach: Yes, that is true. And also this should not be a problem at all. As we only have the _users collection in _system. 
The non-root users will be re-created from scratch every time. The root user's data will be cleaned up before each test run - this is true beyond a single database, but is true for all databases.

PR #15876, opened 2022-03-10
Feature/preg 62 integrate wasm vm

jvolmer:
This PR fixes the integration of the wasm vm and refactors the wasm common files.
The code for a wasm function can now be provided via a byte array or a base64 encoded string.


arangod/WasmServer/WasmCommon.cpp
```diff
@@ -1,108 +1,150 @@
...
+    }
+    auto decodedString =
+        arangodb::basics::StringUtils::decodeBase64(slice.copyString());
+    std::vector<uint8_t> vec(decodedString.begin(), decodedString.end());
+    return Code{vec};
   } else {
-    return ResultT<Slice>::error(TRI_ERROR_BAD_PARAMETER,
-                                 "Field " + fieldname + " should be a string");
+    return ResultT<Code>::error(TRI_ERROR_BAD_PARAMETER,
+                                "Sould be a byte array or base64 string");
```
markuspf, line 136:
```suggestion
                                "Should be a byte array or base64 string");
```

arangod/WasmServer/WasmCommon.cpp
```diff
@@ -1,108 +1,150 @@
...
+      } else {
+        return ResultT<Code>::error(TRI_ERROR_BAD_PARAMETER,
+                                    "Array should include only bytes");
+      }
+    }
+    return Code{code};
+  } else if (slice.isString()) {
+    auto string = slice.copyString();
+    if (!regex_match(string, std::regex("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/"
+                                        "]{3}=|[A-Za-z0-9+/]{2}==)?$"))) {
```
markuspf, line 126: Where does this regex come from? maybe it should be a constexpr somewhere in a header file?

arangod/WasmServer/WasmCommon.h
```diff
@@ -32,33 +32,29 @@
 
 namespace arangodb::wasm {
 
-struct WasmContainer {
-  uint8_t* code;
-  size_t length;
+struct Code {
+  std::vector<uint8_t> bytes;
```
markuspf, line 36: Did you check that the data in vector is guaranteed to be a continous chunk of memory that you can hand to `parse_module`?

arangod/WasmServer/WasmServerFeature.cpp
```diff
@@ -68,6 +68,7 @@ auto WasmServerFeature::executeFunction(std::string const& name, uint64_t a,
     return function.call<uint64_t>(a, b);
```
markuspf, line 68: What happens if the function does not exist in the runtime?

markuspf (changes requested):
* Some people might be sad about the use of `2` instead of `To` in the velocypack functions
* Some minor comments left

arangod/WasmServer/WasmCommon.cpp
jvolmer: I guess the best fit is to add this constexpr to arangodb::baseics::StringUtils.
I just googled a bit, found this regex and also checked it against the description on wikipedia on how base64 works. It makes sense: A base64 string consists of letters A-Z, a-z, 0-9, + or /. At the end there can be a padding of one or two = symbols. This padding results from the fact that base64 converts 8-bit symbols into 6-bit symbols. This can be done without a rest when 3 8-bits are converted into 4 6-bits (therefore the first block in the regex checks for blocks of 4). If the full number of input bits is not a multiple of 3, the rest is dealt with via = symbols: 2 8-bits are converted into 3 6-bits and one = (the second block in the regex) and 1 8-bit is converted into 2 6-bits and two = (the third block).

arangod/WasmServer/WasmCommon.h
jvolmer: > The elements are stored contiguously, which means that elements can be accessed not only through iterators, but also using offsets to regular pointers to elements. This means that a pointer to an element of a vector may be passed to any function that expects a pointer to an element of an array. 
(https://en.cppreference.com/w/cpp/container/vector)
Additionally, same site says that it meets the requirements for ContiguousContainer which is a container that stores objects in contiguous memory locations.
(As far as I can see, only the specialization std::vector<bool> does no need to be contiuous for improved space efficiency (https://en.cppreference.com/w/cpp/container/vector_bool))

markuspf: :+1:

arangod/WasmServer/WasmServerFeature.cpp
jvolmer: I modified the Wasm3cpp.h file such that now `runtime.find_function(name.c_str())` will return a result instead of throwing an error. Then `call` is only called on a valid function.

arangod/WasmServer/WasmCommon.cpp
jvolmer: TODO: Write own decodeBase64 function that returns a vector<uint8_t>. The current StringUtils::decodeBase64 can only be used for encoded strings because: If the encoded value includes a nullbyte, the current decode function does not return the stuff after this nullbyte.

jvolmer: My previous comment was not true: I added tests that verify that base64 de- and encoding works also for strings that include nullbytes.

PR #15881, opened 2022-03-10
[CINFRA-259] refactor participant flags

maierlars:
### Scope & Purpose
This replaced `excluded` flag with two new flags that are more specific: `allowedAsLeader` and `allowedInQuorum`.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


apetenchea (approved):
LGTM

PR #15883, opened 2022-03-10
BTS-820 | Clarify effect of `--server.session-timeout`

Simran-B:
See https://github.com/arangodb/docs/pull/936


ansoboleva (approved):
reviewed

PR #15904, opened 2022-03-14
BTS-818: get rid of ruby dependency for integration tests

dothebart:
### Scope & Purpose

Replace rspec tests with javascript tests, so we may get rid of ruby. This is the first sample semi automatic converted.

- [x] :hankey: Bugfix

### Checklist

This replaces a test with a similar one


fceller (approved):
The principle LGTM. I have not check the completness.

PR #15905, opened 2022-03-14
SEARCH-184 Add agency cache for cluster metrics

MBkkt:
### Scope & Purpose

Add agency cache for cluster metrics:
* Only one coordinator in the same time going to dbservers
* If coordinator read something new it don't go to dbservers
* TODO(MBkkt) What if coordinator die when get lock (need some timeout with reset callback)
* TODO(MBkkt) Option for metrics endpoint that make blocking update all metrics before make response

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/ClusterInfo.cpp
+++ b/arangod/Cluster/ClusterInfo.cpp
@@ -4232,6 +4232,115 @@ Result ClusterInfo::finishModifyingAnalyzerCoordinator(
   return Result(TRI_ERROR_NO_ERROR);
 }
 
+static std::string const kMetricsFlagPath = "Plan/Metrics/Flag";
+static std::string const kMetricsVersionPath = "Plan/Metrics/Version";
+static std::string const kMetricsDataPath = "Plan/Metrics/Data";
+
+void ClusterInfo::initMetrics() {
+  AgencyComm ac{_server};
+
+  VPackBuilder builder;
+  builder.add(VPackValue{uint64_t{0}});
+
+  AgencyWriteTransaction const write{
+      {kMetricsVersionPath, AgencyValueOperationType::SET, builder.slice()},
+      {kMetricsVersionPath, AgencyPrecondition::Type::EMPTY, true}};
+  do {
+    auto const r = ac.sendTransactionWithFailover(write);
+    if (r.successful() ||
+        r.httpCode() == rest::ResponseCode::PRECONDITION_FAILED) {
+      break;
+    }
+  } while (true);
+}
+
+ClusterInfo::MetricsResult ClusterInfo::startCollectMetrics(
+    std::shared_ptr<metrics::ClusterMetricsFeature::Data>& data) {
+  auto& agencyCache = _server.getFeature<ClusterFeature>().agencyCache();
+  auto [builder, index] =
+      agencyCache.read({AgencyCommHelper::path(kMetricsVersionPath)});
+  auto version =
+      builder->slice()
+          .at(0)
+          .get(std::initializer_list<std::string_view>{
+              AgencyCommHelper::path(), "Plan", "Metrics", "Version"})
+          .getUInt();
+  if ((data ? data->version : 0) < version) {
+    std::tie(builder, index) =
+        agencyCache.read({AgencyCommHelper::path(kMetricsVersionPath),
+                          AgencyCommHelper::path(kMetricsDataPath)});
+    data = metrics::ClusterMetricsFeature::Data::fromVPack(
+        builder->slice().at(0).get(std::initializer_list<std::string_view>{
+            AgencyCommHelper::path(), "Plan", "Metrics"}));
+    return MetricsResult::Repeat;
+  } else {
+    data = nullptr;
+  }
+  AgencyComm ac{_server};
+  // TODO(MBkkt) ReadWrite transaction? Write before Read?
+  AgencyWriteTransaction const write{
+      {kMetricsFlagPath, AgencyValueOperationType::SET,
+       VPackSlice::booleanSlice(true)},
+      {kMetricsFlagPath, AgencyPrecondition::Type::EMPTY, true}};
+  do {
+    auto const r = ac.sendTransactionWithFailover(write);
+    if (r.successful()) {
+      return MetricsResult::Update;
+    } else if (r.httpCode() == rest::ResponseCode::PRECONDITION_FAILED) {
+      return MetricsResult::Reschedule;
+    }
+  } while (true);
+}
+
+uint64_t ClusterInfo::endCollectMetrics(velocypack::Slice data) {
+  AgencyComm ac{_server};
+  if (data.isNone()) {
+    AgencyWriteTransaction const write{
+        {kMetricsFlagPath, AgencySimpleOperationType::DELETE_OP},
+        {kMetricsFlagPath, AgencyPrecondition::Type::EMPTY, false},
+    };
+    do {
+      auto const r = ac.sendTransactionWithFailover(write);
+      if (r.successful()) {
+        return 0;
+      }
+      TRI_ASSERT(r.httpCode() != rest::ResponseCode::PRECONDITION_FAILED);
+    } while (true);
+  }
+  // TODO AgencyCache optimization
+  AgencyReadTransaction const read{AgencyCommHelper::path(kMetricsVersionPath)};
+  uint64_t version = 0;
+  do {
+    auto const r = ac.sendTransactionWithFailover(read);
+    if (r.successful()) {
+      version = r.slice()
+                    .at(0)
+                    .get(std::initializer_list<std::string_view>{
+                        AgencyCommHelper::path(), "Plan", "Metrics", "Version"})
+                    .getUInt();
+      break;
+    }
+  } while (true);
+  VPackBuilder builder;
+  builder.add(VPackValue{version});
+  AgencyWriteTransaction const write{
+      {{kMetricsFlagPath, AgencySimpleOperationType::DELETE_OP},
+       {kMetricsVersionPath, AgencySimpleOperationType::INCREMENT_OP},
+       {kMetricsDataPath, AgencyValueOperationType::SET, data}},
+      {{kMetricsFlagPath, AgencyPrecondition::Type::EMPTY, false},
+       {kMetricsVersionPath, AgencyPrecondition::Type::VALUE,
+        builder.slice()}}};
+  do {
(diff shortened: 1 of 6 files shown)


arangod/Cluster/ClusterInfo.cpp
```diff
@@ -4232,6 +4232,115 @@ Result ClusterInfo::finishModifyingAnalyzerCoordinator(
...
 
+static std::string const kMetricsFlagPath = "Plan/Metrics/Flag";
+static std::string const kMetricsVersionPath = "Plan/Metrics/Version";
+static std::string const kMetricsDataPath = "Plan/Metrics/Data";
+
+void ClusterInfo::initMetrics() {
+  AgencyComm ac{_server};
+
+  VPackBuilder builder;
+  builder.add(VPackValue{uint64_t{0}});
```
Dronplane, line 4243: As a suggestion - we could just treat "absence" of the version key as value 0.

kvahed:
I have one major question, please. How offen is an agency write planned?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -4232,6 +4232,115 @@ Result ClusterInfo::finishModifyingAnalyzerCoordinator(
...
+    std::shared_ptr<metrics::ClusterMetricsFeature::Data>& data) {
+  auto& agencyCache = _server.getFeature<ClusterFeature>().agencyCache();
+  auto [builder, index] =
+      agencyCache.read({AgencyCommHelper::path(kMetricsVersionPath)});
+  auto version =
+      builder->slice()
+          .at(0)
+          .get(std::initializer_list<std::string_view>{
+              AgencyCommHelper::path(), "Plan", "Metrics", "Version"})
+          .getUInt();
```
gnusi, line 4267: use `getNumber`

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -4232,6 +4232,115 @@ Result ClusterInfo::finishModifyingAnalyzerCoordinator(
...
+  } while (true);
+}
+
+ClusterInfo::MetricsResult ClusterInfo::startCollectMetrics(
+    std::shared_ptr<metrics::ClusterMetricsFeature::Data>& data) {
+  auto& agencyCache = _server.getFeature<ClusterFeature>().agencyCache();
+  auto [builder, index] =
+      agencyCache.read({AgencyCommHelper::path(kMetricsVersionPath)});
+  auto version =
+      builder->slice()
```
gnusi, line 4263: let's check slice type

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -69,66 +69,112 @@ void ClusterMetricsFeature::validateOptions(
...
+  ci.initMetrics();
+  asyncUpdate();
+}
+
 void ClusterMetricsFeature::beginShutdown() {
+  if (!isEnabled()) {
+    return;
+  }
+  disable();
   _count.store(std::numeric_limits<int32_t>::min() + 1,
```
gnusi, line 87: let's use meaningful constants for values for `_count`

MBkkt:
@kvahed No often than the time it takes to collect metrics from all database servers

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -69,66 +69,112 @@ void ClusterMetricsFeature::validateOptions(
...
-void ClusterMetricsFeature::scheduleUpdate() noexcept {
+void ClusterMetricsFeature::rescheduleUpdate(uint32_t timeout) noexcept {
   auto now = std::chrono::steady_clock::now();
-  auto next = _lastUpdate + std::chrono::seconds{_timeout};
+  auto next = _lastUpdate + std::chrono::seconds{timeout};
   _handle = SchedulerFeature::SCHEDULER->queueDelayed(
       RequestLane::CLUSTER_INTERNAL, (now < next ? next - now : 0ns),
       [this](bool canceled) {
-        if (canceled) {
+        if (canceled || _count.exchange(1, std::memory_order_relaxed) < 1) {
```
gnusi, line 104: let's use a meaningful constant for `1`

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -189,4 +240,45 @@ void ClusterMetricsFeature::toPrometheus(std::string& result,
   }
 }
 
+std::shared_ptr<ClusterMetricsFeature::Data>
+ClusterMetricsFeature::Data::fromVPack(VPackSlice slice) {
+  auto metrics = slice.get("Data");
+  auto version = slice.get("Version").getUInt();
```
gnusi, line 246: `getUInt` -> `getNumber<>`

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -189,4 +240,45 @@ void ClusterMetricsFeature::toPrometheus(std::string& result,
...
+    TRI_ASSERT(false);
+    return {};
+  }
+  auto data = std::make_shared<Data>(version);
+  for (size_t i = 0; i != size; i += 3) {
+    auto name = metrics.at(i).stringView();
+    auto labels = metrics.at(i + 1).stringView();
+    auto type = metrics.at(i + 2).type();
+    MetricValue value;
+    if (type == VPackValueType::UInt || type == VPackValueType::SmallInt) {
```
gnusi, line 262: `if (type == VPackValueType::UInt || type == VPackValueType::SmallInt) {` -> ` metrics.at(i + 2).isNumber()`

kvahed:
@MBkkt What I mean is, is this a regular process or only, when something changes? There is a problem, that you have to consider: We use the agency dumps for debugging customer problems. 50000 entries are kept on the clusters, the rest is compacted, if this process is recurring like once every second, 14 hours later the agency dump is useless for debugging.

arangod/Metrics/ClusterMetricsFeature.cpp
```diff
@@ -189,4 +240,45 @@ void ClusterMetricsFeature::toPrometheus(std::string& result,
...
+      value = metrics.at(i + 2).getUInt();
```
gnusi, line 263: `getUInt` -> `getNumber`

kvahed:
Dear all, I have also chatted a bit with @MBkkt in slack about this. I'm afraid that we need a discussion about this PR before it can be really reviewed. The first hurdle is mainly to figure out, what the consequences are for all stakeholders, support, cluster, graph, .... Changes to clusterinfo/agency must come with a design document, that should be acked by all teams. 
As you all know, agency dumps are at least 50% of our debug business. We cannot keep an infinite amount of the past, and are in the process of evicting other spammers, like arangosync and arangodb starter from the persisted part of the agency.
There is a non-persisted part, that is erased on every leader change for example, that wouldn't have as much side-effects.

gnusi:
@kvahed these changes were discussed with @joerg84 and @jsteemann. If you have another proposal/idea on how to proceed with cluter-wide global metrics, I believe everyone is happy to discuss. ATM your comments look like there is a severe design flow in the Agency but hopefully I'm wrong.

kvahed:
@gnusi There is no design flaw in the agency. It was never meant to store volatile data. The agency's role is clearly defined as a global configuration store. Much like `etcd` in kubernetes clusters. And again, the stake holders, MetricsFeature (myself) and ClusterFeature, Agency (@neunhoef) were not even consulted. Apart from that, we don't have a design document.

neunhoef:
Can somebody maybe describe what it is that should be achieved in the first place? I do not understand "persistent cache between coordinators". That is why we have a procedure for this:
  1. Write a design document.
  2. Involve all stake holders.
  3. Once everybody has signed off, plan development.
  4. Do the implementation.
  5. Get the reviews in and do the testing.
  6. Merge.
  7. Publish.
As Kaveh says, if the agency is involved, Team Clifton has to be involved from the start. If metrics are involved, then Kaveh as the owner of metrics has to be involved.
I understand that this process not been followed here (as it is so often the case). Now let's try to mop up the mess and please somebody start a short document in whatever format describing the target which should be achieved here.
I am pretty sure that regularly writing into the persistent part of the agency is not the solution to whatever it is should be achieved here.

MBkkt:
@neunhoef 

> Can somebody maybe describe what it is that should be achieved in the first place?

Can somebody write some documentation to agency? At least doxygen for methods and enum values?

```
enum class AgencyValueOperationType {
  ERASE,
  SET,
  OBSERVE,
  UNOBSERVE,
  PUSH,
  PREPEND,
  REPLACE
};
```

What is the difference between `REPLACE` and `SET`?
What exactly does mean `OBSERVE` and `UNOBSERVE`?
What does mean `PREPEND`?

And many other questions, why I should write to Kaveh or another devs to get this knowledge?

Also we have some discussion about this changes here https://arangodb.atlassian.net/browse/APM-184

>  I do not understand "persistent cache between coordinators". 

What exactly do you not understand? I try to explain

> As Kaveh says, if the agency is involved, Team Clifton has to be involved from the start. If metrics are involved, then Kaveh as the owner of metrics has to be involved.

I think all these reviewers have been added to this PR.

> describing the target which should be achieved here.

Currently all coordinators don't know about other coordinators when try to collect metrics.
Metrics collect triggered by prometheus scrape.
In this PR I want to achieve these goals:
* Only one coordinator in the same time going to dbservers
* When coordinator collect metrics, it store them on agency (persistent cache between coordinators)
* If coordinator read some new data from agency (persistent cache between coordinators) it don't go to dbservers at this moment (current prometheus scrape)

> I am pretty sure that regularly writing into the persistent part of the agency is not the solution to whatever it is should be achieved here.

So in arangodb cluster we can't reliably store some data that regularly changes.
Do I understand your claim correctly? 
As reliable, I mean that everything is fine as long as there are fewer machine failures than majority of agents at the same moment.

Maybe make more sense to use agency only for distributed locking and version, and leader election for store some data.
I think in that case we can use collection (as Kaveh mention) for store data(with replication factor same as count of agency majority?) and use agency for leader election?

But is it really necessary? If I understand Kaveh correctly main problem not in RSM, problem in that all (write?) transaction write to `agencyDump()` result? But describing above solution doesn't solve this problem.

PR #15828, opened 2022-03-01
[ORCO-30] Feature/http request fuzzer

cpjulia:
### Scope & Purpose

Obs.: Will be updated with more ways of fuzzing and improvement of performance when there's around 20 iterations for the fuzzer

This PR is a fuzzer for http requests. For now, it fuzzes the header and sends to the server, as a normal http request, invoked in arangosh via command `arango.fuzzRequests(<optional parameters>)`.
The optional parameters that can be supplied are: `fuzzRequests(<numRequests>, <numIterations>, <seed>)` 
The parameter numIterations is the amount of times the fuzzer is going to perform it's random actions on the header, and seed is for the seed that is used for randomizing. 
Example: 
```
arango.fuzzRequests(10000, 10) 
```
means the fuzzer would run for 10000 requests, each one having 10 iterations on the fuzzer for creating the header.
Operations are to add ASCII string, add unsigned it, create, copy or remove existing line.

Obs. More info to be added here 

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist
Obs.: tests might be added
- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
 No backports required (for the moment).

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/jira/software/c/projects/ORCO/boards/69?modal=detail&selectedIssue=ORCO-30
- [ ] Design document:

PR #15907, opened 2022-03-14
Fix assertion failure and crash on HTTP header error, BTS-792 (#15728)

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15728/

This PR strives to solve two different problems which can occur if an error is detected in the HTTP header. One is an assertion failure in maintainer mode, the other is a proper crash. The latter is particularly dangerous, since it can be used to crash a server in production with a carefully crafted HTTP header.

First the assertion failure:

The `_response` member of the HttpCommTask object was reset in the `on_message_began` callback. This is not reached if an error is detected in the header before actually reading the message. In this case an assertion failure can be triggered when sending the 505 response. The fix is to additionall reset `_response`  in the `on_header_complete` method.

Now the proper crash:

Since we use the llhttp library for parsing HTTP requests, this problem seems to be present. The problem is that in the case of a bad header, we have returned `HPE_OK` which lead to the fact that both an error response was sent out via `sendSImpleResponse` (actually happening in an IO thread asynchronously), and we continued to work on the request and in the end sent out another, potentially positive response.

The fix is to return `HPE_USER` to indicate a header parsing error. But then we also have to react to the fact that llhttp does not hand this error through, but returns HPE_CB_HEADERS_COMPLETE instead. So in this case, we must not close the connection right away, but must first let the response be sent, and then let the connection be closed due to the error.

Fix was validated using HTTP fuzzing functionality soon to be added to devel (https://github.com/arangodb/arangodb/pull/15828).

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


apetenchea (approved):

arangod/GeneralServer/HttpCommTask.cpp
```diff
@@ -150,12 +151,12 @@ int HttpCommTask<T>::on_header_complete(llhttp_t* p) {
       (p->http_major != 1 || p->http_minor != 1)) {
     me->sendSimpleResponse(rest::ResponseCode::HTTP_VERSION_NOT_SUPPORTED,
                            rest::ContentType::UNSET, 1, VPackBuffer<uint8_t>());
-    return HPE_OK;
+    return HPE_USER;
```
markuspf, line 154: What is the error code `HPE_USER` used for in arangodb? (what is the significance of this change?)

jsteemann: `HPE_USER` is not directly used by arangod, but there are several places that compare the return code of the http processing callbacks with `HPE_OK` (e.g. `if (err != HPE_OK) ...`.
Previously we returned `HPE_OK` from here, but the `sendSimpleResponse` already triggered the sending of a response in the background asynchronously. The `HPE_OK` made us continue reading data from the connection and continuing the parsing. Thus we had a race here, and that could actually lead to a crash. 
Returning `HPE_USER` here means: ok, this is not the same as `HPE_OK`, and it is an error that will abort further processing. But it is not any of the predefined errors provided by llhttp. It is basically a catch-all "custom error" happened (in this case: "unsupported HTTP version" or "request entity too large", for which llhttp doesn't have its own error enum values). The fix was validated with an HTTP fuzzer against 3.9.

neunhoef (approved):
LGTM

markuspf (approved):
LGTM

arangod/GeneralServer/HttpCommTask.cpp
markuspf: :+1:

PR #15918, opened 2022-03-15
only set failurepoints, if the environment supports that

dothebart:
### Scope & Purpose

this part of the test should only be allowed, if the server supports failure points in first place.

- [x] :hankey: Bugfix


dothebart:
only unrelated tests failed.

PR #15924, opened 2022-03-16
Do not audit log replication writes by default [BTS-826]

neunhoef:
We found that replication document writes are audit logged on the DEBUG
log level of the topic "audit-document". Seemingly accidentally the
default log level for this topic is currently DEBUG. This leads to
extreme delays for shards trying to get in sync, since there there is
an audit log call for every single document being replicated. Since
a synchronously written audit log can delay writing a document by up
to a factor of 1000 and since the document write has already been audit
logged on the leader, we propose to set the default log level for
"audit-document" to INFO, which was probably intended in the first
place.

Similarly, the default log level for "audit-authentication" is also
DEBUG. This leads that non-authentication events for low priority things
like credentials missing for internal UI requests and /_open/auth are
currently audit logged by default. Again, we propose to change the 
default log level for "audit-authentication" to INFO, which was probably
intended in the first place.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/15929
  - [*] Backport for 3.8: https://github.com/arangodb/arangodb/pull/15930
  - [*] Backport for 3.7: https://github.com/arangodb/arangodb/pull/15931

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,9 @@ devel
 -----
 
 * No longer put document writes from replication into the audit log by
-  default. Same with low priority authentication.
+  default. Same with low priority authentication like internal UI requests
+  to .html files for the UI. This solves a performance problem for
+  shards getting in sync with audit log switched on.
 
 * Added an HTTP fuzzer that will fuzz requests (the header for the actual
   moment) and send to the server. The amount of requests sent is provided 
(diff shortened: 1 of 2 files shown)


jsteemann (approved):
LGTM

PR #15929, opened 2022-03-16
Do not audit log replication writes by default [BTS_826]

neunhoef:
We found that replication document writes are audit logged on the DEBUG
log level of the topic "audit-document". Seemingly accidentally the
default log level for this topic is currently DEBUG. This leads to
extreme delays for shards trying to get in sync, since there there is
an audit log call for every single document being replicated. Since
a synchronously written audit log can delay writing a document by up
to a factor of 1000 and since the document write has already been audit
logged on the leader, we propose to set the default log level for
"audit-document" to INFO, which was probably intended in the first
place.

Similarly, the default log level for "audit-authentication" is also
DEBUG. This leads that non-authentication events for low priority things
like credentials missing for internal UI requests and /_open/auth are
currently audit logged by default. Again, we propose to change the 
default log level for "audit-authentication" to INFO, which was probably
intended in the first place.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] This is the backport for 3.9. Original: https://github.com/arangodb/arangodb/pull/15924


neunhoef:
This is awaiting release-owners decisions, therefore the DO_NOT_MERGE label.

jsteemann (approved):
LGTM!

neunhoef:
Decision by silence to backport.

kvahed (approved):
LGTM

PR #15942, opened 2022-03-17
[CINFRA-304] Add interim api to replace the participant of a replicated state

goedderz:
### Scope & Purpose

Transitional API to replace a participant of a replicated state.

- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [X] **integration tests**

#### Related Information

- [X] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-304


maierlars (approved):
I think, what is missing, it so force a server to become leader. However, for that to work, the replicated state supervision has to learn that first. We should do this in a separate ticket.

PR #15946, opened 2022-03-17
[CINFRA-315] prototype concurrency test

maierlars:
### Scope & Purpose
This PR adds a test that exposes a bug in the prototype state machine. And it will include a fix eventually.

EDIT:
A fix has been added using `waitForApplied`, a function which got moved from `PrototypeCore` to `PrototypeLeaderState`. `waitForApplied` is called on the previous index, before applying anything to the state machine.
Additionally, `stream->release` calls have been un-commented.

The code is still dirty :hankey:

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mpoeter (commented):
IMHO there are still some races that could result in the changes to be applied in the wrong order. I think this approach reduces the risk for that, but it is still possible.

For example, when we call `waitForApplied`, we use `thenValue` to attach our callback to the future. But what if the thread is interrupted before the call to `thenValue`? The futures are resolved in the correct order, but if the handler has not yet been attached, then it cannot be called. Once the thread finishes the call to `thenValue` on the already resolved future, the function will be executed immediately, but this might be too late and out of order.

PR #15960, opened 2022-03-21
reduce memory usage at startup

jsteemann:
### Scope & Purpose

- for loading icudtl.dat. previously a string buffer was built up
  incrementally while reading the binary input file, while triggered
  some resize operations and some final overallocation along the way. as
  the size of the input file is precisely known when loading the file,
  we can get away with reserving a buffer of exactly the required size
  instead.
- when creating initial indexes for initial collections: previously, a
  (temporary) memory allocation of 32MB was done per index creation.
  this is now reduced to smaller amounts of memory if the indexes are
  expected to contain few documents

No backports are planned for these changes

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
+++ b/arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
@@ -79,11 +79,12 @@ struct BuilderCookie : public arangodb::TransactionState::Cookie {
 }  // namespace
 
 RocksDBBuilderIndex::RocksDBBuilderIndex(
-    std::shared_ptr<arangodb::RocksDBIndex> const& wp)
+    std::shared_ptr<arangodb::RocksDBIndex> wp, uint64_t numDocsHint)
     : RocksDBIndex(wp->id(), wp->collection(), wp->name(), wp->fields(),
                    wp->unique(), wp->sparse(), wp->columnFamily(),
                    wp->objectId(), /*useCache*/ false),
-      _wrapped(wp),
+      _wrapped(std::move(wp)),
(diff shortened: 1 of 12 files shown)


arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -303,6 +304,15 @@ arangodb::Result RocksDBBuilderIndex::fillIndexForeground() {
     _docsProcessed.fetch_add(docsProcessed, std::memory_order_relaxed);
   };
 
+  // reserve some space in WriteBatch
+  size_t batchSize = 1024 * 1024;
```
cpjulia, line 308: maybe this could be a constant  cariable in the .h?

cpjulia (approved):
LGTM

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
jsteemann: I don't think it should be exposed in the .h file. It is internal to the index builder code, and the value does not need to be accessed from other places.

PR #15962, opened 2022-03-22
[BTS-811]

cpjulia:
### Scope & Purpose

This PR is related to ticket https://arangodb.atlassian.net/browse/BTS-811. It is intended to fix the checksum to `.sst` files being made before data is encrytped and written to .sst files, so the data that is checksummed is the same as the encrypted data.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
 No backports needed

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/923
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-811
- [ ] Design document:


arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+        it != _sstFileNamesToHashes.end()) {
+      TRI_ASSERT(fileName.size() > 4);
+      shaFileName.append(fileName, 0,
+                         fileName.size() - 4);  // append without .sst
+      TRI_ASSERT(!isFileNameSst(shaFileName));
+      shaFileName += ".sha." + (*it).second + ".hash";
+      _sstFileNamesToHashes.erase(it);
+    }
+    if (!shaFileName.empty()) {
+      auto res = TRI_UnlinkFile(shaFileName.data());
```
neunhoef, line 210: Maybe we want to use the downward env's `DeleteFile` here rather than messing with the file in the filesystem ourselves?

neunhoef (changes requested):
Basically this looks all right. See my two comments for changes.
And we probably want to add a test, at least for single server, to protect against bit rot.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
+///
```
jsteemann, line 1: First line of standard header is missing
```suggestion
////////////////////////////////////////////////////////////////////////////////
///
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+  EVP_MD_CTX_destroy(_context);
+#endif
+}
+
+void ChecksumCalculator::update(char const* buffer, size_t n) {
+  TRI_ASSERT(_context != nullptr);
+  EVP_DigestUpdate(_context, static_cast<void const*>(buffer), n);
+}
+
+bool ChecksumHelper::isFileNameSst(std::string const& fileName) noexcept {
```
jsteemann, line 58: This cannot be `noexcept`, because `TRI_Basename(...)` returns a new std::string by value. That can throw.
```suggestion
bool ChecksumHelper::isFileNameSst(std::string const& fileName) {
```

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,98 @@
...
+  void update(char const* buffer, size_t n);
+  ~ChecksumCalculator();
+
+ private:
+  EVP_MD_CTX* _context;
+};
+
+class ChecksumHelper {
+ public:
+  ChecksumHelper(std::string const& rootPath) : _rootPath{rootPath} {}
```
jsteemann, line 46:
```suggestion
  explicit ChecksumHelper(std::string const& rootPath) : _rootPath{rootPath} {}
```

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,98 @@
...
+  [[nodiscard]] static bool isFileNameSst(std::string const& fileName) noexcept;
+  bool writeShaFile(std::string const& fileName,
+                    std::string const& checksum);  // perhaps should be void
+  [[nodiscard]] rocksdb::Status DeleteFile(std::string const& fileName);
+  [[nodiscard]] std::string computeChecksum();
+  void checkMissingShaFiles();
+
+ private:
+  std::string _rootPath;
+  EVP_MD_CTX* _context;
```
jsteemann, line 56: Seems to be unnecessary in ChecksumHelper.
```suggestion
```

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,98 @@
...
+ public:
+  explicit ChecksumWritableFile(rocksdb::WritableFile* t,
+                                std::string const& fileName,
+                                std::shared_ptr<ChecksumHelper> helper)
+      : rocksdb::WritableFileWrapper(t),
+        _sstFileName(fileName),
+        _helper(helper) {}
+  rocksdb::Status Append(const rocksdb::Slice& data) override {
+    return rocksdb::WritableFileWrapper::Append(data);
+  }
```
jsteemann, line 71: This override seems unnecessary, as it unconditionally forwards to the base class.

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,98 @@
...
+  Mutex _calculatedHashesMutex;
+};
+
+class ChecksumWritableFile : public rocksdb::WritableFileWrapper {
+ public:
+  explicit ChecksumWritableFile(rocksdb::WritableFile* t,
+                                std::string const& fileName,
+                                std::shared_ptr<ChecksumHelper> helper)
+      : rocksdb::WritableFileWrapper(t),
+        _sstFileName(fileName),
```
jsteemann, line 67: I think this member should be renamed to `_filename`, because it will handle not just .sst files, but every file that RocksDB writes to.
```suggestion
        _sstFileName(fileName),
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return true;
+  }
+
+  LOG_TOPIC("8f7ef", WARN, arangodb::Logger::ENGINES)
+      << "shaCalcFile: TRI_WriteFile failed with " << res << " for "
+      << shaFileName;
+  return false;
+}
+
+std::string ChecksumCalculator::computeChecksum() {
```
jsteemann, line 86: Can the methods in this file be ordered by class? It is hard to get an overview as it is right now, because methods of different classes are interleaved in the file.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+  if (!_helper->isFileNameSst(_sstFileName)) {
+    return rocksdb::WritableFileWrapper::Close();
+  }
+  TRI_ASSERT(_helper != nullptr);
+  auto checksumCalc = ChecksumCalculator();
+  if (TRI_ProcessFile(_sstFileName.c_str(),
+                      [&checksumCalc](char const* buffer, size_t n) {
+                        checksumCalc.update(buffer, n);
+                        return true;
+                      })) {
```
jsteemann, line 180: Mmmmmh, this calculates the entire checksum upon closing the file only, which is inefficient.
Shouldn't we have already calculated the checksum incrementally when we get here?

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return rocksdb::EnvWrapper::DeleteFile(fileName);
+  }
+  return _helper->DeleteFile(fileName);
+}
+
+rocksdb::Status ChecksumHelper::DeleteFile(const std::string& fileName) {
+  std::string shaFileName;
+  {
+    MUTEX_LOCKER(mutexLock, _calculatedHashesMutex);
+    if (auto it = _sstFileNamesToHashes.find(TRI_Basename(fileName));
```
jsteemann, line 200: The basename of the file should be calculated before acquiring the mutex, just to make the critical section shorter.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+        LOG_TOPIC("acb34", WARN, arangodb::Logger::ENGINES)
+            << "deleteCalcFile:  TRI_UnlinkFile failed with " << res << " for "
+            << shaFileName;
+      }
+    }
+  }
+  auto res = TRI_UnlinkFile(fileName.data());
+  if (res == TRI_ERROR_NO_ERROR) {
+    LOG_TOPIC("77a2a", DEBUG, arangodb::Logger::ENGINES)
+        << "deleteCalcFile:  TRI_UnlinkFile succeeded for " << shaFileName;
```
jsteemann, line 224:
```suggestion
        << "deleteCalcFile:  removing file succeeded for " << shaFileName;
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return rocksdb::Status::OK();
+  } else {
+    LOG_TOPIC("ce937", WARN, arangodb::Logger::ENGINES)
+        << "deleteCalcFile:  TRI_UnlinkFile failed with " << res << " for "
```
jsteemann, line 228:
```suggestion
        << "deleteCalcFile:  removing file failed with " << res << " for "
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -66,6 +66,7 @@
 #include "RocksDBEngine/Listeners/RocksDBThrottle.h"
 #include "RocksDBEngine/ReplicatedRocksDBTransactionState.h"
 #include "RocksDBEngine/RocksDBBackgroundThread.h"
+#include "../../3rdParty/rocksdb/6.29/env/composite_env_wrapper.h"  // TODO: GET RID OF THIS WEIRD INCLUDE
```
jsteemann, line 69:
```suggestion
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -122,6 +124,14 @@ using namespace arangodb;
 using namespace arangodb::application_features;
 using namespace arangodb::options;
 
+namespace {
+using Env = rocksdb::Env;
+std::unique_ptr<rocksdb::Env> checksumEnv;
```
jsteemann, line 129: Can we store the pointer to the env inside the RocksDBEngine? This would be much better than storing it in a static variable.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,269 @@
...
+    TRI_ASSERT(false);
+  }
+  _checksum = basics::StringUtils::encodeHex(
+      reinterpret_cast<char const*>(&hash[0]), lengthOfHash);
+}
+
+void ChecksumCalculator::updateIncrementalChecksum(char const* buffer) {
+  TRI_ASSERT(_context != nullptr);
+  updateEVPWithContent(
+      buffer, sizeof(buffer));  // based on the current call to Append()
```
jsteemann, line 59: This won't work! `sizeof(buffer)` will be 8!

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if (_helper->isFileNameSst(_fileName)) {
+    _checksumCalc.updateIncrementalChecksum(data.ToString().data());
+  }
+  return rocksdb::WritableFileWrapper::Append(data);
+}
+
+rocksdb::Status ChecksumWritableFile::Close() {
+  if (_helper->isFileNameSst(_fileName)) {
+    TRI_ASSERT(_helper != nullptr);
+    if (!_helper->writeShaFile(_fileName, _checksumCalc.getChecksum())) {
```
jsteemann, line 205: When this is called, is the checksum already finalized? I don't see any call to computeFinalChecksum() around here.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+  }
+  if (EVP_DigestInit_ex(_context, EVP_sha256(), nullptr) == 0) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL,
+                                   "unable to initialize SHA256 processor");
+  }
+}
+
+void ChecksumCalculator::computeFinalChecksum() {
+  TRI_ASSERT(_context != nullptr);
+  unsigned char hash[EVP_MAX_MD_SIZE];
```
jsteemann, line 52: Let's also add the following assert here: `TRI_ASSERT(_checksum.empty());` so the API cannot be misused.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+void ChecksumCalculator::updateEVPWithContent(const char* buffer, size_t n) {
+  TRI_ASSERT(_context != nullptr);
+  if (EVP_DigestUpdate(_context, static_cast<void const*>(buffer), n) == 0) {
+    TRI_ASSERT(false);
+  }
+}
+
+bool ChecksumHelper::isFileNameSst(std::string const& fileName) {
+  return TRI_Basename(fileName).size() > 4 &&
```
jsteemann, line 83: Let's simplify this to `return `fileName.ends_with(".sst");`.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  LOG_TOPIC("80257", DEBUG, arangodb::Logger::ENGINES)
+      << "shaCalcFile: done " << fileName << " result: " << shaFileName;
+  auto res = TRI_WriteFile(shaFileName.c_str(), "", 0);
+  if (res == TRI_ERROR_NO_ERROR) {
+    {
+      MUTEX_LOCKER(mutexLock, _calculatedHashesMutex);
+      _fileNamesToHashes.try_emplace(TRI_Basename(fileName), checksum);
+    }
+    return true;
```
jsteemann, line 101:
```suggestion
    std::string basename = TRI_Basename(fileName);
    MUTEX_LOCKER(mutexLock, _calculatedHashesMutex);
    _fileNamesToHashes.try_emplace(std::move(basename), checksum);
    return true;
```

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,101 @@
...
+
+namespace arangodb::checksum {
+
+class ChecksumCalculator {
+ public:
+  ChecksumCalculator();
+  void computeFinalChecksum();
+  void updateIncrementalChecksum(char const* buffer, size_t n);
+  void updateEVPWithContent(char const* buffer, size_t n);
+  [[nodiscard]] std::string getChecksum() { return _checksum; }
```
jsteemann, line 39:
```suggestion
  [[nodiscard]] std::string getChecksum() const { return _checksum; }
```

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -646,6 +647,16 @@ class RocksDBEngine final : public StorageEngine {
...
   std::shared_ptr<RocksDBLogPersistor> _logPersistor;
+
+  // Checksum env for when creation of sha files is enabled
+  // this is for when encryption is enabled, sha files will be created
+  // after the encryption of the .sst files
+  std::unique_ptr<rocksdb::Env> _checksumEnv;
+
+  rocksdb::Env* NewChecksumEnv(rocksdb::Env* base_env,
+                               std::string const& path) {
+    return new arangodb::checksum::ChecksumEnv(base_env, path);
```
jsteemann, line 658: Let's move the implementation of this function to the .cpp file. If it's in the header, we need to rely on including `RocksDBChecksumEnv.h` in the header file.

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -646,6 +647,16 @@ class RocksDBEngine final : public StorageEngine {
 
   // @brief persistor for replicated logs
   std::shared_ptr<RocksDBLogPersistor> _logPersistor;
+
+  // Checksum env for when creation of sha files is enabled
+  // this is for when encryption is enabled, sha files will be created
+  // after the encryption of the .sst files
+  std::unique_ptr<rocksdb::Env> _checksumEnv;
+
+  rocksdb::Env* NewChecksumEnv(rocksdb::Env* base_env,
```
jsteemann, line 656: I also suggest that we make this function return a unique_ptr, and not a raw ptr.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -795,8 +795,19 @@ void RocksDBEngine::start() {
...
+  if (_createShaFiles) {
+    _checksumEnv.reset(NewChecksumEnv(rocksdb::Env::Default(), _path));
+    _options.env = _checksumEnv.get();
+    static_cast<checksum::ChecksumEnv*>(_checksumEnv.get())
+        ->getHelper()
+        ->checkMissingShaFiles();
+  } else {
+    _options.env = rocksdb::Env::Default();
+  }
+
```
jsteemann, line 807: Is it safe to do it here? It looks like the encryption setup is done directly afterwards, in ` configureEnterpriseRocksDBOptions(...)`. 
I think we need to move the call to `checkMissingShaFiles()` directly after the `configureEnterpriseRocksDBOptions()` call.

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,101 @@
...
+  EVP_MD_CTX* _context;
+  std::string _checksum;
+};
+
+class ChecksumHelper {
+ public:
+  explicit ChecksumHelper(std::string const& rootPath) : _rootPath{rootPath} {}
+  [[nodiscard]] static bool isFileNameSst(std::string const& fileName);
+  bool writeShaFile(std::string const& fileName, std::string const& checksum);
+  [[nodiscard]] std::string buildShaFileNameFromSst(
```
jsteemann, line 52: AFAIK the method does not refer to any instance variables, so it could be made static.
```suggestion
  [[nodiscard]] static std::string buildShaFileNameFromSst(
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+                 " Computing checksum for "
+              << tempPath;
+          auto checksumCalc = ChecksumCalculator();
+          if (TRI_ProcessFile(tempPath.c_str(),
+                              [&checksumCalc](char const* buffer, size_t n) {
+                                checksumCalc.updateEVPWithContent(buffer, n);
+                                return true;
+                              })) {
+            checksumCalc.computeFinalChecksum();
+            writeShaFile(tempPath, checksumCalc.getChecksum());
```
jsteemann, line 163: Please add a comment that `writeShaFile` also inserts the checksum into the hash table. This is not obvious when looking just at the code here.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+      if (shaIndex != std::string::npos) {
+        sstFileName = it->substr(0, shaIndex) + ".sst";
+        TRI_ASSERT(sstFileName == TRI_Basename(sstFileName));
+        auto nextIt = it + 1;
+        if (nextIt != fileList.end() && nextIt->compare(sstFileName) == 0) {
+          TRI_ASSERT(it->size() >= shaIndex + 64);
+          std::string hash = it->substr(shaIndex + /*.sha.*/ 5, 64);
+          it = nextIt;
+          MUTEX_LOCKER(mutexLock, _calculatedHashesMutex);
+          _fileNamesToHashes.try_emplace(sstFileName, std::move(hash));
```
jsteemann, line 131: AFAIK `sstFileName` it is not used afterwards, so it could be moved.
```suggestion
          _fileNamesToHashes.try_emplace(std::move(sstFileName), std::move(hash));
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+rocksdb::Status ChecksumWritableFile::Append(const rocksdb::Slice& data) {
+  _checksumCalc.updateIncrementalChecksum(data.data(), data.size());
+  return rocksdb::WritableFileWrapper::Append(data);
+}
+
+rocksdb::Status ChecksumWritableFile::Close() {
+  TRI_ASSERT(_helper != nullptr);
+  _checksumCalc.computeFinalChecksum();
+  if (!_helper->writeShaFile(_fileName, _checksumCalc.getChecksum())) {
+    return rocksdb::Status::Corruption("File writing was unsuccessful");
```
jsteemann, line 205: Exiting early here will lead to the underlying file left open.
We need to call Close() regardless of whether writing the .sha file succeeded or not.

jsteemann, line 205: We need to fix the error message, because when this error occurs, it is totally unclear _which_ file caused a problem. I also don't think that a `Corruption` error is justified here. What about `IOError`?

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+  std::unique_ptr<rocksdb::WritableFile> writableFile;
+  rocksdb::Status s =
+      rocksdb::EnvWrapper::NewWritableFile(fileName, &writableFile, options);
+  if (!s.ok()) {
+    return s;
+  }
+  try {
+    if (_helper->isFileNameSst(fileName)) {
+      result->reset(
+          new ChecksumWritableFile(std::move(writableFile), fileName, _helper));
```
jsteemann, line 222: We should avoid using `new` and dealing with raw pointer. Can you use `std::make_unique` here?

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+    } else {
+      *result = std::move(writableFile);
+    }
+  } catch (std::exception& e) {
```
jsteemann, line 226:
```suggestion
  } catch (std::exception const& e) {
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+  if (_helper->isFileNameSst(fileName) ||
+      fileName.find(".sha") != std::string::npos) {
+    std::string checksum = _helper->removeFromTable(fileName);
+    std::string shaFileName =
+        _helper->buildShaFileNameFromSst(fileName, checksum);
+    if (!shaFileName.empty()) {
+      if (rocksdb::EnvWrapper::DeleteFile(shaFileName) ==
+          rocksdb::Status::OK()) {
+        LOG_TOPIC("e0a0d", DEBUG, arangodb::Logger::ENGINES)
+            << "deleteCalcFile:  delete file succeeded for " << shaFileName;
```
jsteemann, line 245:
```suggestion
            << "deleteCalcFile: delete file succeeded for " << shaFileName;
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+        << "WritableFile: exception caught when allocating " << e.what();
+    return rocksdb::Status::MemoryLimit(
+        "WritableFile: exception caught when allocating");
+  }
+  return s;
+}
+
+rocksdb::Status ChecksumEnv::DeleteFile(const std::string& fileName) {
+  if (_helper->isFileNameSst(fileName) ||
+      fileName.find(".sha") != std::string::npos) {
```
jsteemann, line 237: Will RocksDB ever call us to delete a `.sha` file? I guess not, because it is not aware of these files.

neunhoef: I agree. And the code following this will also not work for `.sha` files, as far as I see. I think RocksDB is not aware of the `.sha` files and thus will not call us to delete any.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+bool ChecksumHelper::writeShaFile(std::string const& fileName,
+                                  std::string const& checksum) {
+  TRI_ASSERT(isFileNameSst(fileName));
+
+  std::string shaFileName = buildShaFileNameFromSst(fileName, checksum);
+
+  LOG_TOPIC("80257", DEBUG, arangodb::Logger::ENGINES)
+      << "shaCalcFile: done " << fileName << " result: " << shaFileName;
+  auto res = TRI_WriteFile(shaFileName.c_str(), "", 0);
```
neunhoef, line 95: We are writing the file here directly via the OS. Maybe we should be good citizens and use the underlying ENV to create the sha file. Imagine there were another underlying env which does yet some other magic (like the proposed logging env) beneath, this underlying env would then not be aware of the additional file.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+    return true;
+  }
+
+  LOG_TOPIC("8f7ef", WARN, arangodb::Logger::ENGINES)
+      << "shaCalcFile: writing file failed with " << res << " for "
+      << shaFileName;
+  return false;
+}
+
+void ChecksumHelper::checkMissingShaFiles() {
```
neunhoef, line 110: As far as I see this function is used in two places: 

(1) At startup, when the main db directory is checked. It also updates the member `_fileNamesToHashes` after a restart.
(2) It is called on a hotbackup when that is created.

I wonder whether this second call would confuse the global member `_fileNamesToHashes`?

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+          MUTEX_LOCKER(mutexLock, _calculatedHashesMutex);
+          _fileNamesToHashes.try_emplace(sstFileName, std::move(hash));
+        } else {
+          std::string tempPath =
+              basics::FileUtils::buildFilename(_rootPath, *it);
+          LOG_TOPIC("4eac9", DEBUG, arangodb::Logger::ENGINES)
+              << "checkMissingShaFiles:"
+                 " Deleting file "
+              << tempPath;
+          TRI_UnlinkFile(tempPath.data());
```
neunhoef, line 139: We directly use the OS here. Maybe we should also delegate to the underlying ENV?

arangod/RocksDBEngine/RocksDBEngine.cpp
neunhoef: No, I think this is correct. Without encryption, we need this delegations:

  ChecksumEnv -> DefaultEnv     (if sha files should be generated)

and

  DefaultEnv    (if sha files should not be generated)

If encryption is on, we need

  EncryptingEnv -> ChecksumEnv -> DefaultEnv

That is, because we want the checksums of the **encrypted** data. Therefore, we must first set up the ChecksumEnv and **later** wrap the current env with the EncryptingEnv.

neunhoef (changes requested):
See comments. But this is overall going in the right direction.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
neunhoef: I think my comment is rubbish. I finally understood that a shaFilesManager object is created especially for this purpose in (2) and so all seems to be good.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
jsteemann: I tend to disagree here: I think we should not use the env to write the sha files, simply because they are not RocksDB's own files. They are managed completely outside of RocksDB.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
jsteemann: I tend to disagree here: I think we should not use the env to write the sha files, simply because they are not RocksDB's own files. They are managed completely outside of RocksDB.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
cpjulia: Will remove sha from the if

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
cpjulia: For RocksDB's own files, we are forwarding the call to the base method. When rocksdb doesn't care about the files, like .sha., we use our own OS methods, like TRI_Unlink now.

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
cpjulia: Using ENV for rocksdb files and our own OS methods for the sha files now

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+rocksdb::Status ChecksumWritableFile::Append(const rocksdb::Slice& data) {
+  _checksumCalc.updateIncrementalChecksum(data.data(), data.size());
+  return rocksdb::WritableFileWrapper::Append(data);
+}
+
+rocksdb::Status ChecksumWritableFile::Close() {
+  TRI_ASSERT(_helper != nullptr);
+  _checksumCalc.computeFinalChecksum();
+  if (!_helper->writeShaFile(_fileName, _checksumCalc.getChecksum())) {
+    LOG_TOPIC("0b00e", ERR, arangodb::Logger::ENGINES)
```
jsteemann, line 206:
```suggestion
    LOG_TOPIC("0b00e", WARN, arangodb::Logger::ENGINES)
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+        LOG_TOPIC("e0a0d", DEBUG, arangodb::Logger::ENGINES)
+            << "deleteCalcFile: delete file succeeded for " << shaFileName;
+      } else {
+        LOG_TOPIC("acb34", WARN, arangodb::Logger::ENGINES)
+            << "deleteCalcFile: delete file failed for " << shaFileName;
+      }
+    }
+  }
+  rocksdb::Status res = rocksdb::EnvWrapper::DeleteFile(fileName);
+  if (res == rocksdb::Status::OK()) {
```
jsteemann, line 254: Can be simplified to just
```suggestion
  if (res.ok()) {
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+    LOG_TOPIC("77a2a", DEBUG, arangodb::Logger::ENGINES)
+        << "deleteCalcFile:  delete file succeeded for " << fileName;
```
jsteemann, line 256:
```suggestion
        << "deleteCalcFile: delete file succeeded for " << fileName;
```

arangod/RocksDBEngine/RocksDBChecksumEnv.cpp
```diff
@@ -0,0 +1,264 @@
...
+  } else {
+    LOG_TOPIC("ce937", WARN, arangodb::Logger::ENGINES)
+        << "deleteCalcFile:  delete file failed for " << fileName;
```
jsteemann, line 259:
```suggestion
        << "deleteCalcFile: delete file failed for " << fileName;
```

arangod/RocksDBEngine/RocksDBChecksumEnv.h
```diff
@@ -0,0 +1,104 @@
...
+      const std::string& fileName,
+      std::unique_ptr<rocksdb::WritableFile>* result,
+      const rocksdb::EnvOptions& options) override;
+
+  rocksdb::Status DeleteFile(const std::string& fileName) override;
+
+  std::shared_ptr<ChecksumHelper> getHelper() const { return _helper; }
+
+ private:
+  std::shared_ptr<ChecksumHelper> _helper;
```
jsteemann, line 101: Question: we does the helper need to be a shared_ptr? The env is exclusively owning it, and it doesn't need to be created on the heap AFAIK.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -795,8 +800,19 @@ void RocksDBEngine::start() {
   _options.compaction_readahead_size =
       static_cast<size_t>(opts._compactionReadaheadSize);
 
+  if (_createShaFiles) {
+    _checksumEnv = NewChecksumEnv(rocksdb::Env::Default(), _path);
+    _options.env = _checksumEnv.get();
+    static_cast<checksum::ChecksumEnv*>(_checksumEnv.get())
+        ->getHelper()
+        ->checkMissingShaFiles();
```
jsteemann, line 808: Please leave a comment that this works despite not having set up the encryption yet (encryption setup follows in line 814 via the call to `configureEnterpriseRocksDBOptions()`).
This is because the `checkMissingShaFiles()` does not depend on the env or the encryption properly working, as the method will only look for existing .sst files and calculate file checksums from the file contents, unaware of any encryption etc.

arangod/RocksDBEngine/RocksDBChecksumEnv.h
cpjulia: it shares the ownership with ChecksumWritableFile so they can edit the same hash table that stores the sst files and their shas. I pass with std::move() in the initializer list of ChecksumWritableFile, should it be done differently so Env and ChecksumWritableFile point to the same _helper and both count as references to it?

jsteemann (approved):
LGTM

neunhoef (approved):
LGTM

neunhoef (approved):
LGTM

PR #15964, opened 2022-03-22
ensure users are cleaned up after testcases

dothebart:
### Scope & Purpose

make sure tests don't influence each others by creating and not flushing users

- [x] :hankey: Bugfix


dothebart:
fuck git

PR #15967, opened 2022-03-22
[CINFRA-320] Move LogTarget to new type inspection library, add tests

markuspf:
### Scope & Purpose

Move LogTarget to new type inspection library, add tests


arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
```diff
@@ -215,8 +210,27 @@ struct LogTarget {
...
+auto inspect(Inspector& f, LogTarget::Supervision& x) {
+  return f.object(x).fields(
+      f.field("maxActionsTraceLength", x.maxActionsTraceLength));
+}
+
+template<class Inspector>
+auto inspect(Inspector& f, LogTarget& x) {
+  return f.object(x).fields(
+      f.field(StaticStrings::Id, x.id),
+      f.field(StaticStrings::Participants, x.participants),
```
maierlars, line 228: This field need a fallback to an empty map if it is not provided.

PR #15971, opened 2022-03-22
Feature/enforce users cleanup2

dothebart:
### Scope & Purpose

check whether tests cleanup users, and if not abort.
Exception: Authentication-suite. 

- [x] :hankey: Bugfix

### DIFF
--- a/js/client/modules/@arangodb/testutils/test-utils.js
+++ b/js/client/modules/@arangodb/testutils/test-utils.js
@@ -365,7 +365,16 @@ function performTests (options, testList, testname, runFn, serverOptions, startS
                 ' ] - Original test status: ' +
                 JSON.stringify(results[te])
             };
-
+          }
+          if (db._users.toArray().length !== 1) {
+            continueTesting = false;
+            results[te] = {
+              status: false,
+              message: 'Cleanup of failure points missing - found users left over: [ ' +
+                JSON.stringify(db._users.toArray()) +
+                ' ] - Original test status: ' +
+                JSON.stringify(results[te])
+            };
           }
         } else {
           serverDead = true;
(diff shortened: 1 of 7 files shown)


dothebart:
testrun green except unrelated test.

js/client/modules/@arangodb/testutils/test-utils.js
```diff
@@ -173,7 +175,10 @@ function performTests (options, testList, testname, runFn, serverOptions, startS
   let count = 0;
   let forceTerminate = false;
   let graphCount = 0;
-
+  let usersCount = 0;
+  if (testname !== 'agency') {
+    usersCount = userManager.all().length;
+  };
```
jsteemann, line 181:
```suggestion
  }
```

js/client/modules/@arangodb/testutils/test-utils.js
```diff
@@ -173,7 +175,10 @@ function performTests (options, testList, testname, runFn, serverOptions, startS
   let count = 0;
   let forceTerminate = false;
   let graphCount = 0;
-
+  let usersCount = 0;
+  if (testname !== 'agency') {
```
jsteemann, line 179: Why are we excluding `agency` here, and `agency` and `authentication` further down?

tests/js/client/shell/shell-api-database.js
```diff
@@ -174,6 +174,12 @@ function dealing_with_database_manipulation_methodsSuite () {
 
     tearDown: function() {
       arango.DELETE(api + `/${name}`);
+      db._flushCache();
+      db._users.toArray().forEach(user => {
+        if (user.user !== "root") {
+          arango.DELETE_RAW("/_api/user/" + user.user);
```
jsteemann, line 180: Not sure what's in the user names here, but URL-encoding them sounds like a good idea in general.
```suggestion
          arango.DELETE_RAW("/_api/user/" + encodeURIComponent(user.user));
```

js/client/modules/@arangodb/testutils/test-utils.js
dothebart: agency has no authentification at all, so this will throw. `authentication` was excluded feature wise, but could be added here as well.

PR #15988, opened 2022-03-25
Fix assertion failure and crash on HTTP header error, BTS-792 (#15728)

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/15728

This PR strives to solve two different problems which can occur if an error is detected in the HTTP header. One is an assertion failure in maintainer mode, the other is a proper crash. The latter is particularly dangerous, since it can be used to crash a server in production with a carefully crafted HTTP header.

First the assertion failure:

The `_response` member of the HttpCommTask object was reset in the `on_message_began` callback. This is not reached if an error is detected in the header before actually reading the message. In this case an assertion failure can be triggered when sending the 505 response. The fix is to additionall reset `_response`  in the `on_header_complete` method.

Now the proper crash:

Since we use the llhttp library for parsing HTTP requests, this problem seems to be present. The problem is that in the case of a bad header, we have returned `HPE_OK` which lead to the fact that both an error response was sent out via `sendSImpleResponse` (actually happening in an IO thread asynchronously), and we continued to work on the request and in the end sent out another, potentially positive response.

The fix is to return `HPE_USER` to indicate a header parsing error. But then we also have to react to the fact that llhttp does not hand this error through, but returns HPE_CB_HEADERS_COMPLETE instead. So in this case, we must not close the connection right away, but must first let the response be sent, and then let the connection be closed due to the error.

Bugfix tested manually via fuzz-testing functionality that is available in devel arangosh.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/15907
  - [x] Backport for 3.8: https://github.com/arangodb/arangodb/pull/15987
  - [x] Backport for 3.7: this PR

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

maierlars (approved):

CHANGELOG
```diff
@@ -1,6 +1,12 @@
 v3.7.18 (XXXX-XX-XX)
 --------------------
 
+* Fixed an assertion failure which could occur when there was an error in
+  the HTTP header, so that the message body was not actually read.
+
+* Fixed a crash which could occur when there was an error in the HTTP
+  header parsing).
```
cpjulia, line 8: Unbalanced `)`

cpjulia (approved):
LGTM

PR #15992, opened 2022-03-25
[PREG-103] Make wasm functions run in cluster - DBServer and coordinator

jvolmer:
This PR includes
- PREG-101 Save wasm modules in satellite collection instead of in memory
- PREG-102 Enable wasm functions to run on coordinator
- PREG-103 Enable wasm function to run on DB Server

This PR does still use the old velocypack functions, I'll do a separate PR to fix all of the velocypack usages in the Wasm directory.


markuspf:
I merged this with two rebases which changed the commit hash, so github did not detect it.

PR #15997, opened 2022-03-25
[CINFRA-331] Locking in LeaderManager

maierlars:
### Scope & Purpose
Adds locking to the leader manager. Was missing before. This could fix the performance tests as well.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mpoeter (approved):
LGTM

PR #16001, opened 2022-03-28
Docs: Add unit to lockTimeout (seconds)

Simran-B:
### Scope & Purpose

The JS Transaction endpoint `_api/transaction` and the Stream Transaction endpoint `_api/transaction/begin` accept a parameter `lockTimeout`. We fail to mention the unit (https://github.com/arangodb/docs/issues/938). It is **seconds** according to @jsteemann.

TODO: What is the default value that is mentioned?

> If not specified, a default value will be used.

Does it refer to the startup option `--transaction.streaming-lock-timeout` in case of Stream Transactions?

What about JS Transactions? There is another startup option `--rocksdb.transaction-lock-timeout` but the unit is milliseconds and it seems unrelated to me (storage engine).

- [x] 📚 Documentation

### Checklist

- [x] Backports
  - [x] Backport for 3.8: *(Please link PR)*
  
### Related information
  
  - [x] Docs PR: https://github.com/arangodb/docs/pull/959

### DIFF
--- a/Documentation/DocuBlocks/Rest/Transactions/post_api_transaction.md
+++ b/Documentation/DocuBlocks/Rest/Transactions/post_api_transaction.md
@@ -33,8 +33,9 @@ Allow reading from undeclared collections.
 
 @RESTBODYPARAM{lockTimeout,integer,optional,int64}
 an optional numeric value that can be used to set a
-timeout in seconds for waiting on collection locks. If not specified, a default
-value will be used. Setting *lockTimeout* to *0* will make ArangoDB
+timeout in seconds for waiting on collection locks. This option is only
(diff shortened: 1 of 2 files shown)


jsteemann (approved):
LGTM

Simran-B:
The default JS Transaction timeout is 900 seconds (hardcoded, not controlled by a startup option).

`--transaction.streaming-lock-timeout` can be used to limit the timeout for 2 operations to access the same streaming transaction. This is different to what the `lockTimeout` option does in a transaction. `lockTimeout` is the maximum lock timeout when writing to the same collection/shard concurrently (and one op uses exclusive locks), and `--transaction.streaming-lock-timeout` is the maximum lock timeout for accessing the same streaming transaction concurrently.

PR #15996, opened 2022-03-25
[CINFRA-235] Convergence Indicator for Replicated Log

markuspf:
### Scope & Purpose

Supervision will report to Current if it thinks there is nothing else to do.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*
  - [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #16024, opened 2022-03-29
[CINFRA-330] Speedup log creation

maierlars:
### Scope & Purpose
Reduced number of required supervision roundtrips. This is done by updating the snapshot after the state supervision was executed. The log supervision will see the changes.


arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -501,7 +507,7 @@ auto checkReplicatedLog(LogTarget const& target,
   }
 
   // Here we are converged and can hence signal so
-  return ConvergedToTargetAction{};
+  return EmptyAction{};
```
markuspf, line 510: This is going to conflict with the changes in #15996

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -388,7 +388,13 @@ auto checkReplicatedLog(LogTarget const& target,
                         ParticipantsHealth const& health) -> Action {
   if (!maybePlan) {
     // The log is not planned right now, so we create it
-    return AddLogToPlanAction(target.id, target.participants);
+    LogPlanTermSpecification term;
+    term.config = target.config;
+    term.term = LogTerm{1};
+    auto leader = target.participants.begin()->first;
+    auto rebootId = health._health.at(leader).rebootId;
```
markuspf, line 395: You have to check whether the log has been created with any participants for this to not fail rather ungracefully; we could of course just `ErrorAction` if one tries creating a log without any participants (or, not enough participants for the config) ?

markuspf: I thought I could remove `CreateIntialTermAction`, but while doing that the above showed up; we can probably still remove `CreateInitialTermAction` because we can just create the log with `LogTerm{1}` regardless

markuspf (changes requested):
There are some problems with this approach.

arangod/Replication2/ReplicatedLog/Supervision.cpp
maierlars: Yes, in the end I don't care, how its done. Important to me is, that we safe round trips when the result of those round trips is already known.

We can close this PR. Just copy the part from Agency/Supervision to make writes of ReplicatedState supervision visible to ReplicatedLog supervision.

arangod/Replication2/ReplicatedLog/AgencyLogSpecification.cpp
```diff
@@ -163,6 +163,8 @@ auto agency::to_string(LogCurrentSupervisionError error) noexcept
       return "the leader selected in target is invalid";
     case LogCurrentSupervisionError::TARGET_LEADER_EXCLUDED:
       return "the leader selected in target is excluded";
+    case LogCurrentSupervisionError::TARGET_NOT_ENOUGH_PARTICIPANTS:
+      return "the leader selected in target is excluded";
```
maierlars, line 167: Typo.

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -382,38 +382,44 @@ auto getRemovedParticipant(ParticipantsFlagsMap const &targetParticipants,
...
-    term.leader.emplace(std::move(leader), rebootId);
-    return AddLogToPlanAction(target.id, target.participants, term);
+    // provided we have enough participants
+    if (target.participants.size() + target.config.writeConcern <
+        target.config.replicationFactor + 1) {
+      return ErrorAction(
+          LogCurrentSupervisionError::TARGET_NOT_ENOUGH_PARTICIPANTS);
+    } else {
+      auto leader = target.participants.begin()->first;
+      auto rebootId = health._health.at(leader).rebootId;
```
maierlars, line 398: I know it was me who came up with that. But maybe we want to select the leader randomly? What about the case that someone specified a leader explicitly in Target? Right now, we would select the first and then do a controlled failover. 

Would that be hard to add?

PR #16025, opened 2022-03-29
make sure fuzz requests are 100% deterministic between runs

jsteemann:
### Scope & Purpose

Make sure fuzz requests are 100% deterministic between runs.
Fuzzing may create requests that force the server to close the connection. So before a fuzzing run starts, we need to make sure we have a fresh, valid connection.
This fixes an issue during testing only. There is no need to backport this change, as fuzzing was added for 3.10 tests only.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


dothebart:
how many connections will be built by the tests in this case?

jsteemann:
> how many connections will be built by the tests in this case?

35 for the entire fuzz test.

cpjulia (approved):
LGTM

PR #16035, opened 2022-03-30
[CINFRA-335] Log Follower Dead lock

maierlars:
### Scope & Purpose
Added in flight flag to prevent multiple concurrent append entries. If the leader sends new append entries requests because of a timeout, it is possible, that the log follower code deadlocks.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/LogFollower.cpp
+++ b/arangod/Replication2/ReplicatedLog/LogFollower.cpp
@@ -245,7 +245,6 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
     // This will cause deadlocks.
     decltype(inFlightGuard) inFlightGuardLocal = std::move(inFlightGuard);
     auto data = self->_guardedFollowerData.getLockedGuard();
-    // _appendEntriesInFlight = false;
 
     auto const& res = tryRes.get();
     {
@@ -288,7 +287,8 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
   // Action
   auto f = core->insertAsync(std::move(iter), req.waitForSync);
   // Release mutex here, otherwise we might deadlock in
-  // checkResultAndCommitIndex
+  // checkResultAndCommitIndex if another request arrives before the previous
+  // one was processed.
   dataGuard.unlock();
   return std::move(f)
       .then(std::move(checkResultAndCommitIndex))
--- a/arangod/Replication2/ReplicatedLog/LogFollower.cpp
+++ b/arangod/Replication2/ReplicatedLog/LogFollower.cpp
@@ -148,7 +148,7 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
   MeasureTimeGuard measureTimeGuard{
       _logMetrics->replicatedLogFollowerAppendEntriesRtUs};
 
-  auto dataGuard = _guardedFollowerData.getLockedGuard(ADB_HERE);
+  auto dataGuard = _guardedFollowerData.getLockedGuard();
 
   {
     // Preflight checks - does the leader, log and other stuff match?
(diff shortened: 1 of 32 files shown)


arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -204,65 +221,72 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
...
       -> std::pair<AppendEntriesResult, DeferredAction> {
     // We have to release the guard after this lambda is finished.
     // Otherwise it would be released when the lambda is destroyed, which
     // happens *after* the following thenValue calls have been executed. In
     // particular the lock is held until the end of the future chain is reached.
     // This will cause deadlocks.
-    decltype(selfGuard) self = std::move(selfGuard);
+    decltype(inFlightGuard) inFlightGuardLocal = std::move(inFlightGuard);
+    auto data = self->_guardedFollowerData.getLockedGuard(ADB_HERE);
+    // _appendEntriesInFlight = false;
```
goedderz, line 244:
```suggestion
```

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -449,16 +473,17 @@ replicated_log::LogFollower::LogFollower(
...
   if (self->_commitIndex >= idx) {
     return futures::Future<WaitForResult>{
         std::in_place, self->_commitIndex,
         std::make_shared<QuorumData>(idx, _currentTerm)};
   }
   // emplace might throw a std::bad_alloc but the remainder is noexcept
   // so either you inserted it and or nothing happens
-  auto it =
-      self->_waitForQueue.getLockedGuard()->emplace(idx, WaitForPromise{});
+  // TODO locking ok? Iterator stored but lock guard is temporary
```
goedderz, line 484: Let's add a ticket for that, so we don't forget.

lib/Basics/Guarded.h
```diff
@@ -174,8 +178,10 @@ class Guarded {
   using mutex_guard_type = MutexGuard<value_type, lock_type>;
   using const_mutex_guard_type = MutexGuard<value_type const, lock_type>;
 
-  auto getLockedGuard() -> mutex_guard_type;
-  auto getLockedGuard() const -> const_mutex_guard_type;
+  auto getLockedGuard(std::optional<basics::SourceLocation> = {})
+      -> mutex_guard_type;
+  auto getLockedGuard(std::optional<basics::SourceLocation> = {}) const
+      -> const_mutex_guard_type;
```
goedderz, line 184: I think the `SourceLocation` debug information here and in the UnshackledMutex should be removed. Or at least made available only in maintainer mode in a way that doesn't add costs without it, but I don't see an easy way to get rid of the parameter in that case.

arangod/RocksDBEngine/RocksDBPersistedLog.cpp
```diff
@@ -238,6 +238,7 @@ void RocksDBLogPersistor::runPersistorWorker(Lane& lane) noexcept {
         }
 
         // resolve all promises in [nextReqToResolve, nextReqToWrite)
+        // TODO resolve those promises async using _executor
```
goedderz, line 241: Let's add a ticket for that.

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -140,19 +148,26 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
...
-    self->_lastRecvMessageId = req.messageId;
+    dataGuard->_lastRecvMessageId = req.messageId;
+    _appendEntriesInFlight = true;
   }
 
+  // In case of an exception, this scope guard sets the in flight flag to false.
+  // _appendEntriesInFlight is an atomic variable, hence we are allowed to set
+  // it without acquiring the mutex.
+  auto inFlightScopeGuard =
+      ScopeGuard([&flag = _appendEntriesInFlight]() noexcept { flag = false; });
```
goedderz, line 169: Alternatively, you could acquire the lock inside the ScopeGuard; then in `checkResultAndCommitIndex` acquire the lock, cancel the `inFlightScopeGuard` afterwards, and then set the flag to false.

arangod/Replication2/ReplicatedLog/LogFollower.cpp
```diff
@@ -204,65 +221,72 @@ auto replicated_log::LogFollower::appendEntries(AppendEntriesRequest req)
...
                           std::move(action));
   };
   static_assert(std::is_nothrow_move_constructible_v<
                 decltype(checkResultAndCommitIndex)>);
 
   // Action
-  return core->insertAsync(std::move(iter), req.waitForSync)
+  auto f = core->insertAsync(std::move(iter), req.waitForSync);
+  // Release mutex here, otherwise we might deadlock in
+  // checkResultAndCommitIndex
```
goedderz, line 287:
```suggestion
  // Release mutex here, otherwise we might deadlock in
  // checkResultAndCommitIndex if another request arrives before the previous one was processed.
```

arangod/Replication2/ReplicatedLog/LogFollower.cpp
maierlars: https://arangodb.atlassian.net/browse/CINFRA-338

goedderz (approved):

PR #16037, opened 2022-03-30
general code cleanup for today

jsteemann:
### Scope & Purpose

Small code cleanup.

* removes `memory()` method from PhysicalCollection, as all implementations only returned a value of 0.
* removes C-style string functions `TRI_EqualString(...)` and `TRI_IsPrefixString(...)` and replaces call sites with calls to std functions.
* cleans up nonce features and removes unused code from it.

This PR does not need to be backported.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/V8Server/v8-collection.cpp
```diff
@@ -2570,8 +2570,6 @@ static void JS_CompletionsVocbase(
       .FromMaybe(false);
   result->Set(context, j++, TRI_V8_ASCII_STRING(isolate, "_dropView()"))
       .FromMaybe(false);
-  result->Set(context, j++, TRI_V8_ASCII_STRING(isolate, "_engine()"))
```
jsteemann, line 2573: If this isn't removed here, the console auto-completion shows `db._engine()` twice.
I don't know why exactly, but removing this line here fixes it.

arangod/RestServer/NonceFeature.cpp
```diff
@@ -44,11 +44,4 @@ void NonceFeature::collectOptions(std::shared_ptr<ProgramOptions> options) {
                              "the size of the hash array for nonces", true);
 }
 
-void NonceFeature::prepare() {
-  constexpr uint64_t initialSize = 2 * 1024 * 1024;
```
jsteemann, line 48: Initialization of nonces now happens in `lib/Basics/Nonce.cpp` lazily, with the same `initialSize` as was used here before.

arangod/RestServer/NonceFeature.cpp
```diff
@@ -44,11 +44,4 @@ void NonceFeature::collectOptions(std::shared_ptr<ProgramOptions> options) {
...
-  Nonce::setInitialSize(static_cast<size_t>(initialSize));
-}
-
-void NonceFeature::unprepare() { Nonce::destroy(); }
```
jsteemann, line 52: No need anymore to call `destroy()` on nonces, because they now acquire memory using a unique ptr and not via a raw pointer.

lib/Basics/Nonce.cpp
```diff
@@ -180,66 +143,24 @@ bool checkAndMark(uint32_t timestamp, uint64_t random) {
...
-  if (timestamp > TimestampNonces[random % (SizeNonces - 63)]) {
-    TimestampNonces[random % (SizeNonces - 63)] = timestamp;
+  if (timestamp > allNonces[random % (numNonces - 63)]) {
+    allNonces[random % (numNonces - 63)] = timestamp;
   }
 
   return 0 < proofs;
 }
 
-std::vector<Statistics> statistics() {
```
jsteemann, line 205: `statistics()` function was never called in our code, so it can be removed.

arangod/Transaction/Methods.cpp
```diff
@@ -2790,11 +2790,12 @@ Future<Result> Methods::commitInternal(MethodsApi api) {
     }
   }
 
+  auto f = futures::makeFuture(Result());
```
cpjulia, line 2793: can this throw?

arangod/V8Server/v8-collection.cpp
cpjulia: don't know if has something to do with this happening, but there is one  
`TRI_AddMethodVocbase(isolate, ArangoNS,
                       TRI_V8_ASCII_STRING(isolate, "_engine"), JS_Engine);` 
in `arangod/V8Server/v8-vocbase.cpp` 
maybe not because this is not setting anything to the result

cpjulia (approved):
LGTM

arangod/Transaction/Methods.cpp
jsteemann: Yes, it can throw at least `std::bad_alloc`.

arangod/V8Server/v8-collection.cpp
jsteemann: Yes, I saw this. But for other method names occurring in JS_CompletionsVocbase there is also an equivalent call to `TRI_AddMethodVocbase(...)`, for example for `_engineStats`. But `_engineStats` does not occur twice during auto-completion. That's why I mentioned I don't know why.

PR #16043, opened 2022-03-31
Rework internal queues for connection and request statistics

jsteemann:
### Scope & Purpose

* Rework internal queues for connection and request statistics. The previous implementation allocated a lot of memory at program start for initializing fixed-sized queues for the statistics objects.
  The problem with using fixed-sized queues is that they will mostly require too much memory for almost all cases, but still do not protect from the queues becoming full and not being able to hold more items.
  Now we go with a variable length queue instead, which only requires a small amount of memory initially, and allocate more memory only when needed.
  Freelists for reusing statistics items are still presents to avoid lots of reallocations.
  The change also reduces the size of the executable's .bss section by more than 10MB.

No backports are needed for this change.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Statistics/ConnectionStatistics.cpp
```diff
@@ -33,13 +38,50 @@ using namespace arangodb;
 // --SECTION--                                                  global variables
 // -----------------------------------------------------------------------------
 
-static size_t const QUEUE_SIZE = 64 * 1024 - 2;  // current (1.62) boost maximum
+namespace {
+// initial amount of empty statistics items to be created in statisticsItems
+constexpr size_t kInitialQueueSize = 32;
```
cpjulia, line 43: maybe this can be static?
```suggestion
static constexpr size_t kInitialQueueSize = 32;
```

cpjulia (approved):
LGTM

arangod/Statistics/ConnectionStatistics.cpp
jsteemann: It doesn't need to be static, because it is already in the anonymous namespace of the translation unit.

PR #16053, opened 2022-04-01
Add IO heartbeat

neunhoef:
This adds an IO heartbeat.

This is a test which is run every 15s on instances which have a RocksDB
instance (DBServer, Agent, Single server, but not Coordinator). A small
file in the database directory is written, synced, read and removed
again. If any of these operations fails or takes longer than 1s, a log
line is written on the INFO level to the ENGINES log topic. Furthermore,
a metrics count when these operations fail or are delayed. A histogram
for all such executions is updated.

This is to ensure that we can write to the underlying volume at all
times with a reasonable performance.

- Introduce an IO heartbeat check for the local volume.
- Add metrics, add feature switch.
- Add metrics Documentation.
- Fix metrics documentation.

Furthermore, some cleanup in the metrics documentation is done.

The whole feature is switched on by default, but can be deactivated with
a configuration option.

### Scope & Purpose

- [*] :pizza: New feature
- [*] Improved visibility in case of external trouble.

### Checklist

- [*] Backports
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/16060
  - [*] Backport for 3.8: https://github.com/arangodb/arangodb/pull/16061


Documentation/Metrics/arangodb_ioheartbeat_delays_read_total.yaml
```diff
@@ -0,0 +1,20 @@
...
+category: Health
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This counter is increased whenever the io heartbeat encounters a delay
+  of at least 1s when reading a small file in the database directory.
+  This test is done every 15s to ensure that the underlying volume is
```
jsteemann, line 16: While 15s may be correct now, I can imagine someone changing the frequency in the code later without updating the metrics description. Shall we change it to "periodically"?
```suggestion
  This test is done periodically to ensure that the underlying volume is
```

Documentation/Metrics/arangodb_ioheartbeat_delays_remove_total.yaml
```diff
@@ -0,0 +1,20 @@
...
+category: Health
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This counter is increased whenever the io heartbeat encounters a
+  delay of at least a second when removing a small file in the database
+  directory. This test is done every 15s to ensure that the underlying
```
jsteemann, line 16: same

Documentation/Metrics/arangodb_ioheartbeat_delays_write_total.yaml
```diff
@@ -0,0 +1,20 @@
...
+category: Health
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This counter is increased whenever the io heartbeat encounters a delay
+  of at least 1s when writing a small file to the database directory.
+  This test is done every 15s to ensure that the underlying volume is
```
jsteemann, line 16: same

Documentation/Metrics/arangodb_ioheartbeat_duration.yaml
```diff
@@ -0,0 +1,19 @@
...
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This histogram is updated whenever the io heartbeat runs its test in
+  the database directory. It writes a small file, syncs it to durable
+  storage, reads it, and then unlinks the file again. This test is done
+  every 15s to ensure that the underlying volume is usable and performs
```
jsteemann, line 17: same

Documentation/Metrics/arangodb_ioheartbeat_failures_read_total.yaml
```diff
@@ -0,0 +1,19 @@
...
+category: Health
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This counter is increased whenever the io heartbeat encounters a problem
+  when reading a small file in the database directory. This test is done
+  every 15s to ensure that the underlying volume is readable. The test can
```
jsteemann, line 16: same

Documentation/Metrics/arangodb_ioheartbeat_failures_write_total.yaml
```diff
@@ -0,0 +1,19 @@
...
+category: Health
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  This counter is increased whenever the io heartbeat encounters a problem
+  when writing a small file to the database directory. This test is done
+  every 15s to ensure that the underlying volume is writable. The test can
```
jsteemann, line 16: same

Documentation/Metrics/arangodb_ioheartbeat_failures_remove_total.yaml
```diff
@@ -0,0 +1,19 @@
+name: arangodb_ioheartbeat_failures_remove_total
+introducedIn: "3.8.7"
+help: |
+  Number of failures in the io heartbeat remove test.
```
jsteemann, line 4: General question: do we need different metrics for writes, removes and reads? Wouldn't it be sufficient to see if _one_ of the operations fails?

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+      _readDelays(metricsFeature.add(arangodb_ioheartbeat_delays_read_total{})),
+      _removeDelays(
+          metricsFeature.add(arangodb_ioheartbeat_delays_remove_total{})) {}
+
+IOHeartbeatThread::~IOHeartbeatThread() { shutdown(); }
+
+void IOHeartbeatThread::run() {
+  auto& databasePathFeature = server().getFeature<DatabasePathFeature>();
+  std::string testFilePath = FileUtils::buildFilename(
+      databasePathFeature.directory().c_str(), "TestBalloonFileIOHeartbeat");
```
jsteemann, line 343:
```suggestion
      databasePathFeature.directory(), "TestBalloonFileIOHeartbeat");
```

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+          trouble = true;
+        }
+        auto finish = std::chrono::steady_clock::now();
+        std::chrono::duration<double> dur = finish - start1;
+        bool delayed = dur > std::chrono::seconds(1);
+        if (trouble || delayed) {
+          if (delayed) {
+            ++_writeDelays;
+          }
+          LOG_TOPIC("66662", INFO, Logger::ENGINES)
```
jsteemann, line 383: This message is not 100% accurate if `trouble` is `true`. Then writing the test file failed, and we probably shouldn't log this message at all.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+              << std::chrono::duration_cast<std::chrono::microseconds>(dur)
+                     .count()
+              << " milliseconds.";
+        }
+
+        // Read the file if we can reasonably assume it is there:
+        if (!trouble) {
+          auto start = std::chrono::steady_clock::now();
+          try {
+            std::string content = FileUtils::slurp(testFilePath);
```
jsteemann, line 394: As `content` is not used later...
```suggestion
            FileUtils::slurp(testFilePath);
```

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+            trouble = true;
+          }
+          auto finish = std::chrono::steady_clock::now();
+          std::chrono::duration<double> dur = finish - start;
+          bool delayed = dur > std::chrono::seconds(1);
+          if (trouble || delayed) {
+            if (delayed) {
+              ++_readDelays;
+            }
+            LOG_TOPIC("66669", INFO, Logger::ENGINES)
```
jsteemann, line 408: Again, this message may be inaccurate if `trouble` is `true`.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+                << "IOHeartbeat: trying to read test file took "
+                << std::chrono::duration_cast<std::chrono::microseconds>(dur)
+                       .count()
+                << " milliseconds.";
+          }
+        }
+
+        // And remove it again:
```
jsteemann, line 416: This will be executed even if the initial writing of the file failed. The whole remove attempt should probably be moved into an if clause and only be executed if the initial write operation did not set `trouble` to `true`.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+        _exeTimeHistogram.count(
+            std::chrono::duration_cast<std::chrono::milliseconds>(dur).count());
+      }
+
+      std::this_thread::sleep_for(std::chrono::milliseconds(waitTime()));
+    } catch (...) {
+    }
+    // next iteration
+  }
+  LOG_TOPIC("66664", INFO, Logger::ENGINES) << "IOHeartbeatThread: stopped.";
```
jsteemann, line 453:
```suggestion
  LOG_TOPIC("66664", DEBUG, Logger::ENGINES) << "IOHeartbeatThread: stopped.";
```

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+  unsigned int counter = checkInterval();
+
+  while (true) {
+    try {  // protect thread against any exceptions
+      if (isStopping()) {
+        // done
+        break;
+      }
+
+      if (--counter == 0) {
```
jsteemann, line 359: Instead of using a `counter` variable here and repeatedly waking up the thread without performing actual work, we should use a condition variable. The thread could sleep on the condition variable for 15s then.
On server shutdown, we can signal the condition variable so the thread can wake up timely. All we will need to do for this it to implement a `beginShutdown()` method for the thread that signals, and that calls the parent Thread `beginShutdown()` method. I think this is both cleaner and more efficient than the counter-based approach.

Documentation/Metrics/arangodb_ioheartbeat_delays_read_total.yaml
neunhoef: Done.

Documentation/Metrics/arangodb_ioheartbeat_failures_remove_total.yaml
neunhoef: This is probably a matter of debate. I wanted to have visibility if only certain aspects of the underlying volume's functionality stopped working. For example, if writing would still be working but reading no longer or incorrectly. This is maybe a bit paranoid and rather unlikely. I think I will reduce it to one failure counter and one delay counter as reaction to this comment, but will then leave the different log messages at INFO level in, so that they would be visible in the logs at the default logging level. How does that sound?

neunhoef: Done.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: Done.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: Well, I had intentionally written "trying to write test file took", so it would be accurate even if it failed. And I definitely would like to have a message including the time it took when it has failed. This could for example be a timeout failure or it could be a very quick failure, having the duration in any case is valuable debugging information. I would rather leave it as it is. Let's discuss online.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: I have added a check for the right content. This was forgotten.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,165 @@ void DatabaseManagerThread::run() {
...
+          trouble = true;
+        }
+        finish = std::chrono::steady_clock::now();
+        dur = finish - start;
+        delayed = dur > std::chrono::seconds(1);
+        if (trouble || delayed) {
+          if (delayed) {
+            ++_removeDelays;
+          }
+          LOG_TOPIC("66671", INFO, Logger::ENGINES)
```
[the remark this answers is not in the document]
neunhoef: See above.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: Done.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: One failure case is that the file could not be written because of permissions (since it is already there, but unwritable). Sometimes then the removal would work, that is why I had originally executed this in any case. But I have moved the removal into the if statement and only remove it if it has worked to create it.

arangod/RestServer/DatabaseFeature.cpp
neunhoef: Done.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,153 @@ void DatabaseManagerThread::run() {
...
+      bool delayed = dur > std::chrono::seconds(1);
+      if (trouble || delayed) {
+        if (delayed) {
+          ++_delays;
+        }
+        LOG_TOPIC("66662", INFO, Logger::ENGINES)
+            << "IOHeartbeat: trying to write test file took "
+            << std::chrono::duration_cast<std::chrono::microseconds>(dur)
+                   .count()
+            << " milliseconds.";
```
jsteemann, line 363: The value reported is in microseconds, but the log messages says "milliseconds".

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,153 @@ void DatabaseManagerThread::run() {
...
+        bool delayed = dur > std::chrono::seconds(1);
+        if (trouble || delayed) {
+          if (delayed) {
+            ++_delays;
+          }
+          LOG_TOPIC("66669", INFO, Logger::ENGINES)
+              << "IOHeartbeat: trying to read test file took "
+              << std::chrono::duration_cast<std::chrono::microseconds>(dur)
+                     .count()
+              << " milliseconds.";
```
jsteemann, line 397: The value reported is in microseconds, but the log messages says "milliseconds".

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -291,6 +294,153 @@ void DatabaseManagerThread::run() {
...
+        delayed = dur > std::chrono::seconds(1);
+        if (trouble || delayed) {
+          if (delayed) {
+            ++_delays;
+          }
+          LOG_TOPIC("66671", INFO, Logger::ENGINES)
+              << "IOHeartbeat: trying to remove test file took "
+              << std::chrono::duration_cast<std::chrono::microseconds>(dur)
+                     .count()
+              << " milliseconds.";
```
jsteemann, line 420: The value reported is in microseconds, but the log messages says "milliseconds".

jsteemann (approved):
LGTM!

arangod/RestServer/DatabaseFeature.cpp
neunhoef: LOL. Will fix.

neunhoef:
Well, it can be switched off if that is an issue. Furthermore, for now my priority is to find issues of customers, which we currently cannot explain in any other way than to assume that the IO subsystem is just not working for extended periods of time. We need visibilty here. If this does not find a problem in half a year, I am happy to change the default to off.

goedderz:
I understand and approve the objective. But is it really adequate to strain everyone's hardware and power consumption, so we might have more visibility in some error cases? From a user's perspective, I find that incredibly impolite of any software to do.

neunhoef:
I sympathize with this view, however, I would have thought that very few people are running with actual spinning rust these days, at least in production. And if so, they probably do it in cloud environments and mount their volumes remotely. For SSDs, the argument with the spindown is probably less strong, as far as I know.
But I agree that my arguments here are rather "ArangoDB-egoistic", in the sense that more power consumption for a few select customers weighs less than our ability to see low level problems across all customers by default.
I suggest that we start with the default "on" and if we see in the next few months that we get few or no reports and use out of this, we can easily change the default later. How about that?

kvahed (approved):
LGTM

jsteemann (approved):
LGTM!

PR #16055, opened 2022-04-01
refactor internals of key generators

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/919

Refactor some internals of key generators:

* move some static methods for validation from `KeyGenerator` to already existing `KeyGeneratorHelper` struct.
* change "Cluster" name particle in key generators to "Coordinator", e.g. "PaddedKeyGeneratorCluster" was renamed to "PaddedKeyGeneratorCoordinator". This is necessary to make it more clear where the key generator will be executed.
* use more StaticStrings for frequently occurring strings, e.g. "allowUserKeys" and "keyOptions".
* rename `KeyGenerator::factory(...)` to `KeyGenerator::create(...)` and make it return an `std::unique_ptr` instead of a raw pointer.

This change is internal and does not need to be backported.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/919
- [ ] Design document:


arangod/Aql/OptimizerRules.cpp
```diff
@@ -822,6 +822,15 @@ std::string getSingleShardId(
 
   // all shard keys found!!
 
+  if (node->getType() == EN::INSERT && collection->numberOfShards() != 1 &&
```
neunhoef, line 825: I was not quickly able to decide why this if block has to be here. If `shardKeys` contains at least one element which does not contain a dot, then `toFind` is not empty, and so we never reach this if. Empty shardKeys is not allowed (as far as I know), and so the only case in which this would be triggered would be if shardKeys has one or more entries containing a dot, which is also not allowed (according to some comment above).

neunhoef (approved):
LGTM, but I have one question in a comment where I think some code is not actually used. Let's discuss this, maybe this is a misunderstanding on my side.

tests/js/client/shell/shell-keygen.js
```diff
@@ -0,0 +1,1362 @@
...
+      try {
+        db._create(cn, { keyOptions: { type: "der-fuchs" } });
+        fail();
+      } catch (err) {
+        assertEqual(ERRORS.ERROR_ARANGO_INVALID_KEY_GENERATOR.code, err.errorNum);
+      }
+    },
+
+    testAutoincrementGeneratorInCluster : function () {
+      if (!cluster) {
```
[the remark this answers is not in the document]
jsteemann: Currently (and previously) the autoincrement key generator is completely disabled in the cluster.
We can make it available for single-sharded collections after this PR is merged!

arangod/Aql/OptimizerRules.cpp
jsteemann: `toFind` is initially populated, but some code between its population and line 825 can remove items from `toFind`. 
If `toFind` is empty, it actually means that the lookup conditions of the query covered all the shard keys of the collection, so in case we get past the `if (!toFind.empty()) { return ... }` it simply means that we know the lookup values for the collection's shard keys.

neunhoef: OK, understood, thanks for the explanation!

PR #16056, opened 2022-04-01
Start StatisticsFeature after NetworkFeature

jsteemann:
### Scope & Purpose

Make the StatisticsFeature start after the NetworkFeature, so that any network request issues by cluster statistics gathering can rely on the networking functionality being available until shutdown.

It is still unclear if this change should be backported.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/ServerFeature.cpp
```diff
@@ -58,8 +59,6 @@ ServerFeature::ServerFeature(Server& server, int* res)
 {
   setOptional(true);
   startsAfter<AqlFeaturePhase>();
-
-  startsAfter<StatisticsFeature>();
```
neunhoef, line 62: Now that I see this line removed, I ask myself: The server messes with these Statistics objects for each request. If we remove this dependency, is it clear that this will all still work? I think yes, since the data structures in the `StatisticsFeature` are initialized in `prepare`, so `start` in the `ServerFeature` can already use them, or indeed if the server begins service, we should be good to go.
Note that if the `RestServer` is deactivated by command line option, then this is found during `validateOptions` and then the `StatisticsFeature` is deactivated, which should still be in time before `prepare`. So I think, we are good.

neunhoef (approved):
LGTM. There are some thoughts in a comment, but they come to the conclusion that this is all OK.

arangod/RestServer/ServerFeature.cpp
jsteemann: Yes. I worried about the same thing. I also recently modified the setup of the connection statistics/request statistics so that a call to `RequestStatistics::initialize` or `ConnectionStatistics::initialize` is just an optimization (reserves some initial memory) but not required anymore. When there are no items on the statistics' freelist anymore, the statistics will allocate a new item on the heap now and are guaranteed to work (except in case of OOM).

PR #16061, opened 2022-04-04
Add IO heartbeat

neunhoef:
This adds an IO heartbeat.

This is a test which is run every 15s on instances which have a RocksDB
instance (DBServer, Agent, Single server, but not Coordinator). A small
file in the database directory is written, synced, read and removed
again. If any of these operations fails or takes longer than 1s, a log
line is written on the INFO level to the ENGINES log topic. Furthermore,
a metrics count when these operations fail or are delayed. A histogram
for all such executions is updated.

This is to ensure that we can write to the underlying volume at all
times with a reasonable performance.

- Introduce an IO heartbeat check for the local volume.
- Add metrics, add feature switch.
- Add metrics Documentation.
- Fix metrics documentation.

Furthermore, some cleanup in the metrics documentation is done.

The whole feature is switched on by default, but can be deactivated with
a configuration option.

### Scope & Purpose

- [*] :pizza: New feature
- [*] Improved visibility in case of external trouble.

### Checklist

- [*] Backports: This is the backport to 3.8. Original PR: https://github.com/arangodb/arangodb/pull/16053

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 v3.8.7 (XXXX-XX-XX)
 -------------------
 
+* Added an IO heartbeat which checks that the underlying volume is writable
+  with reasonable performance. The test is done every 15 seconds and can
+  be explicitly switched off. New metrics to give visibility if the test
+  fails.
+
 * Updated ArangoDB Starter to 0.15.4.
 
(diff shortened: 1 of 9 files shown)


jsteemann (approved):
LGTM!

kvahed (approved):
LGTM

PR #16072, opened 2022-04-05
[CINFRA-150] Read replicated log on followers

maierlars:
### Scope & Purpose
For debugging it is sometimes helpful to read a replicated log on a follower. There was no technical problem to be solved here. The API simply forbid to access a non-leader replicated log. This PR lifts that restriction.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature


markuspf (approved):
LGTM, except I have no idea how these error codes work :shrug:

PR #16077, opened 2022-04-05
[CINFRA-345] proper replicated log creation

maierlars:
### Scope & Purpose

Better log creation api. The old api expected a log id to be specified. The new api selects it randomly but uniquely. You can also specify if you want to wait for the log to be finished. (which is now default)

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**

### DIFF
--- a/arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
+++ b/arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
@@ -276,6 +276,7 @@ struct LogCurrentSupervision {
   std::optional<uint64_t> targetVersion;
 
   auto toVelocyPack(VPackBuilder&) const -> void;
+  static auto fromVelocyPack(velocypack::Slice) -> LogCurrentSupervision;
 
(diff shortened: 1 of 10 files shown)


arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
```diff
@@ -276,6 +276,7 @@ struct LogCurrentSupervision {
   std::optional<uint64_t> targetVersion;
 
   auto toVelocyPack(VPackBuilder&) const -> void;
+  static auto fromVelocyPack(velocypack::Slice) -> LogCurrentSupervision;
```
markuspf, line 279: When I first saw this I thought "huh?",  then I checked; this function exists already, and ultimately ends up being a `deserialize` call; what happens if you just try and remove it?

markuspf (approved):

PR #16092, opened 2022-04-08
Improve shard resync

neunhoef:

I have done some tests. I think this PR is good, but it will not be enough to let a follower catch up in good time, if we continue to write to the leader with multiple threads. This is due to the fact that catchup is inherently single-threaded. But this PR will already help a bit: For example, a collection can resync, even if we still write to a **different** collection, which was not the case before.

- All requests needed for replication catchup at least medium prio.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [ ] Backports (questionable)
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/16104
  - [ ] Backport for 3.8: currently being discussed


arangod/Cluster/SynchronizeShard.cpp
```diff
@@ -392,11 +392,13 @@ arangodb::Result SynchronizeShard::collectionCountOnLeader(
...
   options.timeout = network::Timeout(60);
   options.skipScheduler = true;  // hack to speed up future.get()
+  network::Headers headers;
+  headers.insert_or_assign("X-Arango-Frontend", "true");
 
   auto response =
       network::sendRequest(pool, leaderEndpoint, fuerte::RestVerb::Get,
                            "/_api/collection/" + getShard() + "/count",
-                           VPackBuffer<uint8_t>(), options)
+                           VPackBuffer<uint8_t>(), options, headers)
```
jsteemann, line 401: micro optimization:
```suggestion
                           VPackBuffer<uint8_t>(), options, std::move(headers))
```

neunhoef: Done.

PR #16093, opened 2022-04-08
Slightly simplify explain output

jsteemann:
### Scope & Purpose

* Only show slowest optimizer rules in explain output for optimizer rules
  that took a considerable amount of time (>= 0.0002 seconds). Previously
  the slowest 5 optimizer rules were shown, regardless of how long they
  took to execute and even if they executed sufficiently fast.

This is only a simplification for the explain output and does not need to be backported.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/js/common/modules/@arangodb/aql/explainer.js
+++ b/js/common/modules/@arangodb/aql/explainer.js
@@ -257,6 +257,9 @@ function printRules(rules, stats) {
     }
     return { name: key, time: stats.rules[key] };
   });
+  // filter out everything that was reasonably fast
+  times = times.filter((item) => item.time >= 0.0002);
+
   times.sort(function(l, r) {
     // highest cost first
     return r.time - l.time;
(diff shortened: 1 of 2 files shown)


js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -257,20 +257,26 @@ function printRules(rules, stats) {
...
   });
   // top few only
   times = times.slice(0, 5);
 
-  stringBuilder.appendLine(section('Optimization rules with highest execution times:'));
-  stringBuilder.appendLine(' ' + header('RuleName') + '   ' + pad(maxNameLength - 'RuleName'.length) + header('Duration [s]'));
-  times.forEach(function(rule) {
-    stringBuilder.appendLine(' ' + keyword(rule.name) + '   ' + pad(12 + maxNameLength - rule.name.length - rule.time.toFixed(5).length) + value(rule.time.toFixed(5)));
-  });
+  if (times.length > 0) {
```
cpjulia, line 270: wouldn't this if fit above the filtering of time duration line 260?

cpjulia (approved):
LGTM

js/common/modules/@arangodb/aql/explainer.js
jsteemann: I don't think so, because the filtering can modify the size/length of the `times` array.

cpjulia: I was thinking it would be a starting condition before the filtering, but you're right. You don't want to know that you have time entries before filtering, as, after filtering, you might end up with none. So agree with the `if` remaining where it is.

PR #16103, opened 2022-04-11
Add more collection info to web UI

jsteemann:
### Scope & Purpose

* Make web UI show the following information for collections:
  * key generator type
  * whether or not the document and primary index cache is enabled
  * if cache is enabled, show cache usage and allocation size in figures
  The `cacheEnabled` property of collections is now also changable via the UI for existing collections.

The PR also removes a lot of dead code from one of the views, which doesn't seem to be called anymore after some previous UI changes.

Tried manually for both single server and cluster.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
```diff
@@ -289,6 +302,15 @@
...
       };
 
+      if (cacheEnabled === 'true') {
+        cacheEnabled = true;
+      } else if (cacheEnabled === 'false') {
+        cacheEnabled = false;
+      }
+      if (cacheEnabled !== null && cacheEnabled !== undefined) {
+        data.cacheEnabled = cacheEnabled;
+      }
```
adityamukho, line 312: If `cacheEnabled` is any string other than `'true'` or `'false'`, it still gets set in `data.cacheEnabled`, as a string. Either that is not a problem or we must be sure `cacheEnabled` is never any value other than `true|false|null|undefined`.

Personally, I'd prefer a if `cacheEnabled` defaulted to `false` for any _falsey_ value, and `true` otherwise. So Something like
`data.cacheEnabled = !!cacheEnabled;`

jsteemann: I don't think that `!!cachEnabled` would be correct, given the following experiment:
```
[true, false, "true", "false", null, undefined, 1, 0, "piff"].forEach((value) => { print("value: ", value, "bool: ", !!value); });
value:  true bool:  true
value:  false bool:  false
value:  true bool:  true
value:  false bool:  true
value:  null bool:  false
value:  undefined bool:  false
value:  1 bool:  true
value:  0 bool:  false
value:  piff bool:  true
```

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
```diff
@@ -289,6 +302,15 @@
         waitForSync: wfs
       };
 
+      if (cacheEnabled === 'true') {
+        cacheEnabled = true;
+      } else if (cacheEnabled === 'false') {
+        cacheEnabled = false;
+      }
```
jsteemann, line 309:
```suggestion
      if (cacheEnabled === 'true' || cacheEnabled === true) {
        cacheEnabled = true;
      } else {
        cacheEnabled = false;
      }
```

js/apps/system/_admin/aardvark/APP/frontend/js/models/arangoCollectionModel.js
adityamukho: Good catch, thanks!

adityamukho (approved):
LGTM!

ajurdak (approved):
LGTM.

PR #16104, opened 2022-04-12
Improve shard resync

neunhoef:
### Scope & Purpose

I have done some tests. I think this PR is good, but it will not be enough to let a follower catch up in good time, if we continue to write to the leader with multiple threads. This is due to the fact that catchup is inherently single-threaded. But this PR will already help a bit: For example, a collection can resync, even if we still write to a different collection, which was not the case before.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.9: This is the 3.9 backport

Link to original PR for devel: https://github.com/arangodb/arangodb/pull/16092/


neunhoef:
We will only merge this once devel has shown some time without problems with this change.

arangod/RestHandler/RestReplicationHandler.h
```diff
@@ -61,7 +61,9 @@ class RestReplicationHandler : public RestVocbaseBaseHandler {
     if (len >= 1) {
       std::string const& command = suffixes[0];
       if (command == AddFollower || command == HoldReadLockCollection ||
-          command == RemoveFollower || command == LoggerFollow) {
+          command == RemoveFollower || command == LoggerFollow ||
+          command == Batch || command == Inventory || command == Revisions ||
+          command == Dump) {
         return RequestLane::SERVER_REPLICATION_CATCHUP;
```
cpjulia, line 67: when command is "make follower", does it not have to catchup? Is it different behavior from adding a follower?

cpjulia (approved):
LGTM

arangod/RestHandler/RestReplicationHandler.h
neunhoef: To be honest, I do not know what this `make-follower` API is used for. It is definitely not used in the cluster to catch up a follower. Therefore it is OK to leave it out here with the special prioritities. In any case, I could not find any usage of this API or any documentation of it. It might be useful for some active/failover deployment mode or something like this, or even single server asynchronous replication. I do not know.

jsteemann (approved):
LGTM!

PR #16111, opened 2022-04-13
Try to fix some of data races that tsan reported

MBkkt:
### Scope & Purpose

Try to fix some of data races that tsan reported

https://jenkins.arangodb.biz:3456/job/arangodb-ANY-linux-san.x86-64/1078/EDITION=enterprise,SAN_MODE=TSan,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=linux&&test&&x86-64/

Everything except unsafe calls, velocypack function initilization, and agency lock-order-inversion 

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

TSAN build: https://jenkins.arangodb.biz:3456/job/arangodb-ANY-linux-san.x86-64/1082/

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Query.cpp
+++ b/arangod/Aql/Query.cpp
@@ -266,20 +266,16 @@ void Query::setLockTimeout(double timeout) noexcept {
 }
 
 bool Query::killed() const {
-  if (_queryOptions.maxRuntime > std::numeric_limits<double>::epsilon() &&
-      elapsedSince(_startTime) > _queryOptions.maxRuntime) {
-    return true;
-  }
-  return _queryKilled;
+  return (std::numeric_limits<double>::epsilon() < _queryOptions.maxRuntime &&
+          _queryOptions.maxRuntime < elapsedSince(_startTime)) ||
+         _queryKilled.load(std::memory_order_acquire);
 }
 
 /// @brief set the query to killed
 void Query::kill() {
-  if (ServerState::instance()->isCoordinator() && !_queryKilled) {
-    _queryKilled = true;
-    this->cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
-  } else {
-    _queryKilled = true;
+  auto const wasKilled = _queryKilled.exchange(true, std::memory_order_acq_rel);
+  if (ServerState::instance()->isCoordinator() && !wasKilled) {
+    cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
   }
 }
 
@@ -1269,7 +1265,7 @@ void Query::enterState(QueryExecutionState::ValueType state) {
 
 /// @brief cleanup plan and engine for current query
 ExecutionState Query::cleanupPlanAndEngine(ErrorCode errorCode, bool sync) {
-  if (!_resultCode.has_value()) {
+  if (!_resultCode.has_value()) {  // TODO possible data race here
     // result code not yet set.
     _resultCode = errorCode;
   }
--- a/arangod/Aql/QueryList.cpp
+++ b/arangod/Aql/QueryList.cpp
@@ -44,6 +44,8 @@
 #include <velocypack/Sink.h>
 #include <velocypack/Value.h>
 
+#include <absl/container/inlined_vector.h>
+
 using namespace arangodb;
 using namespace arangodb::aql;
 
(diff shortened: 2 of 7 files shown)


jsteemann (approved):
LGTM!

mpoeter (changes requested):
I previously had exactly the same idea and changed the `QueryList` to use `weak_ptr` instead of raw pointer, but unfortunately this does not work and can lead to deadlocks. The reason is that when we lock the `weak_ptr` in the QueryList we have a full `shared_ptr`. But then it can happen that while we hold this `shared_ptr` all other references to the query are dropped, so we have the last reference. When we then drop that reference, the query gets destroyed and the destructor needs to remove the query from the QueryList. But since we are already in the QueryList and the lock is not reentrant, this results in a deadlock.

jsteemann:
> I previously had exactly the same idea and changed the `QueryList` to use `weak_ptr` instead of raw pointer, but unfortunately this does not work and can lead to deadlocks. The reason is that when we lock the `weak_ptr` in the QueryList we have a full `shared_ptr`. But then it can happen that while we hold this `shared_ptr` all other references to the query are dropped, so we have the last reference. When we then drop that reference, the query gets destroyed and the destructor needs to remove the query from the QueryList. But since we are already in the QueryList and the lock is not reentrant, this results in a deadlock.

Darn, you are right. I asked @MBkkt to get in touch with you regarding any changes here, because I knew there was _some_ issue. I just couldn't remember which exactly. Good that you did! :+1:

jsteemann (approved):
AQL parts look good to me.
I can't say much about the changes to the NetworkMethodsTest.

MBkkt:
Revert NetworkTests change

arangod/Aql/QueryList.cpp
```diff
@@ -262,85 +265,93 @@ void QueryList::remove(Query& query) {
...
-  READ_LOCKER(readLocker, _lock);
-
-  for (auto& it : _current) {
-    Query& query = *(it.second);
-
-    if (!filter(query)) {
-      continue;
+  auto const maxLength = _maxQueryStringLength.load(std::memory_order_relaxed);
+  for (auto const& queryPtr : queries) {
+    if (auto& query = *queryPtr; filter(query)) {
```
mpoeter, line 305: NIT: we could already filter in the first loop

arangod/Aql/QueryList.cpp
```diff
@@ -262,85 +265,93 @@ void QueryList::remove(Query& query) {
...
+  }
+  return {};
 }
 
 /// @brief kills all currently running queries that match the filter function
 /// (i.e. the filter should return true for a queries to be killed)
 uint64_t QueryList::kill(std::function<bool(Query&)> const& filter,
                          bool silent) {
+  // We should call dtor std::shared_ptr<Query> not under a lock,
+  // otherwise a deadlock will occur. Query dtor make a WRITE_LOCKER(..., _lock)
```
mpoeter, line 291:
```suggestion
  // We must not call the std::shared_ptr<Query> dtor under the lock,
  // because this can result in a deadlock since the Query dtor tries to
  // remove itself from this list which acquires the `_lock` (again).
```

arangod/Aql/QueryList.cpp
```diff
@@ -262,85 +265,93 @@ void QueryList::remove(Query& query) {
...
-    ++killed;
   }
-
   return killed;
 }
 
 /// @brief get the list of currently running queries
 std::vector<QueryEntryCopy> QueryList::listCurrent() {
+  // We should call dtor std::shared_ptr<Query> not under a lock,
+  // otherwise a deadlock will occur. Query dtor make a WRITE_LOCKER(..., _lock)
```
mpoeter, line 316:
```suggestion
  // We must not call the std::shared_ptr<Query> dtor under the lock,
  // because this can result in a deadlock since the Query dtor tries to
  // remove itself from this list which acquires the `_lock` (again).
```

mpoeter (approved):
Some suggestions, otherwise LGTM.

arangod/Aql/Query.cpp
```diff
@@ -1269,7 +1265,7 @@ void Query::enterState(QueryExecutionState::ValueType state) {
 
 /// @brief cleanup plan and engine for current query
 ExecutionState Query::cleanupPlanAndEngine(ErrorCode errorCode, bool sync) {
-  if (!_resultCode.has_value()) {
+  if (!_resultCode.has_value()) {  // TODO possible data race here
```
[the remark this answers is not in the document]
MBkkt: In devel yes? In kill we call this method if load() is false, then we store(true), if someone else load false before we store store then this code will be called twice.

Now im not sure, this method called from many places

arangod/Aql/QueryList.cpp
MBkkt: I think we don't want filter under lock?

arangod/Aql/QueryList.cpp
MBkkt: I agree, this sounds better, but I think the important point is that in dtor we are trying to acquire a write lock.

arangod/Aql/QueryList.cpp
```diff
@@ -262,85 +265,93 @@ void QueryList::remove(Query& query) {
...
+    queryPtr = it->second.lock();
   }
-
-  killQuery(*it->second, maxLength, false);
-
-  return Result();
+  if (queryPtr) {
+    auto const length = _maxQueryStringLength.load(std::memory_order_relaxed);
+    killQuery(*queryPtr, length, false);
+  }
```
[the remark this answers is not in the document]
MBkkt: Dunno I just keep the old behavior

arangod/Aql/Query.cpp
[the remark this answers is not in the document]
MBkkt: I wrote about these lines
```C++
void Query::kill() {
  if (ServerState::instance()->isCoordinator() && !_queryKilled) {
    _queryKilled = true;
    this->cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
```

arangod/Aql/QueryList.cpp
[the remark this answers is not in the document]
MBkkt: Yes, I think you're right

arangod/Aql/QueryList.cpp
MBkkt: I don't think we want filter under lock?

arangod/Aql/QueryList.cpp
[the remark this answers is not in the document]
MBkkt: Maybe not.
Now `killQuery` just write some log and call doing nothing method if query already killed

arangod/Aql/QueryList.cpp
mpoeter: I don't think it matters since previously we also filtered under lock, but this is just a suggestion, so I am perfectly fine with leaving it as it is.

arangod/Aql/QueryList.cpp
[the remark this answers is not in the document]
mpoeter: True, but strictly speaking this is a change in behavior, since previously this situation simply could not occur. I think both versions are fine, I just think that it might be more consistent to also trigger the error in case we cannot lock the weak_ptr.

PR #16113, opened 2022-04-13
Bugfix/query kill data race

MBkkt:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Query.cpp
+++ b/arangod/Aql/Query.cpp
@@ -266,20 +266,20 @@ void Query::setLockTimeout(double timeout) noexcept {
 }
 
 bool Query::killed() const {
-  if (_queryOptions.maxRuntime > std::numeric_limits<double>::epsilon() &&
-      elapsedSince(_startTime) > _queryOptions.maxRuntime) {
-    return true;
-  }
-  return _queryKilled;
+  return (std::numeric_limits<double>::epsilon() < _queryOptions.maxRuntime &&
+          _queryOptions.maxRuntime < elapsedSince(_startTime)) ||
+         _queryKilled.load(std::memory_order_acquire);
 }
 
 /// @brief set the query to killed
 void Query::kill() {
-  if (ServerState::instance()->isCoordinator() && !_queryKilled) {
-    _queryKilled = true;
-    this->cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
+  if (ServerState::instance()->isCoordinator()) {
+    if (!_queryKilled.load(std::memory_order_acquire) &&
+        !_queryKilled.exchange(true, std::memory_order_acq_rel)) {
+      cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
+    }
   } else {
-    _queryKilled = true;
+    _queryKilled.store(true, std::memory_order_release);
   }
 }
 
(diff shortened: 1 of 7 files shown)


arangod/Aql/Query.cpp
```diff
@@ -266,20 +266,20 @@ void Query::setLockTimeout(double timeout) noexcept {
...
 }
 
 /// @brief set the query to killed
 void Query::kill() {
-  if (ServerState::instance()->isCoordinator() && !_queryKilled) {
-    _queryKilled = true;
-    this->cleanupPlanAndEngine(TRI_ERROR_QUERY_KILLED, /*sync*/ false);
+  if (ServerState::instance()->isCoordinator()) {
+    if (!_queryKilled.load(std::memory_order_acquire) &&
+        !_queryKilled.exchange(true, std::memory_order_acq_rel)) {
```
jsteemann, line 278: No doubt this works, but I think this is a performance optimization that will only have an effect when a query is killed, which should happen very seldomly. At least I think the use case "killing a query" doesn't need to be performance-optimized.

arangod/Aql/QueryList.cpp
```diff
@@ -309,38 +314,40 @@ std::vector<QueryEntryCopy> QueryList::listCurrent() {
   // so we reduce the possibility of having to reserve more room
   // later
   result.reserve(16);
-
-  size_t const maxLength =
-      _maxQueryStringLength.load(std::memory_order_relaxed);
+  absl::InlinedVector<std::shared_ptr<Query>, 16> queries;
```
jsteemann, line 317: What is the reason to populate this vector in this method?
As far as I can see, it is only populated when iterating over all queries, but the vector contents are never used.
Instead of pushing the shared_ptr into this vector, we could as well continue with the shared_ptr itself, i.e.
```
      auto queryPtr = it.second.lock();
      if (!queryPtr) {
        continue;
      }
      auto& query = *queryPtr;
...
```

MBkkt: https://github.com/arangodb/arangodb/pull/16111#pullrequestreview-940874099

Also its draft about this line https://github.com/arangodb/arangodb/pull/16113/files#diff-233b9a3cd7f4156d982e0803152e356f9d8dac03ddef8044172ebbe1c32f3343R1274

arangod/Aql/Query.cpp
MBkkt: I think you are right and only exchange will be look better

arangod/Aql/QueryList.cpp
MBkkt: Yes, main idea is move call dtor from under lock to without lock.
And you right, I should add comment

MBkkt: https://github.com/arangodb/arangodb/pull/16113/commits/881e1b4f8639827b46787bf4f182684189aed20a

MBkkt: https://github.com/arangodb/arangodb/pull/16111/commits/29f2268742ce10cab2e260f18dfefebedda467f5

MBkkt:
Should be fixed in another PR

PR #16116, opened 2022-04-14
Improve shard resync

neunhoef:
### Scope & Purpose

I have done some tests. I think this PR is good, but it will not be enough to let a follower catch up in good time, if we continue to write to the leader with multiple threads. This is due to the fact that catchup is inherently single-threaded. But this PR will already help a bit: For example, a collection can resync, even if we still write to a different collection, which was not the case before.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.8: This is the 3.8 backport

Link to original PR for devel: https://github.com/arangodb/arangodb/pull/16092/


neunhoef:
We have decided to merge this backport only when no problems have been seen in `devel` for some time.

cpjulia (approved):
LGTM

jsteemann (approved):
LGTM!

PR #16147, opened 2022-04-20
Ensure snappy does not use BMI2 instructions.

mpoeter:
### Scope & Purpose

For the Windows build, the new Snappy version, which was introduced in 3.9, generated code that contained BMI2 instructions which where introduced with the Intel Haswell architecture. However, our target architecture for 3.9 is actually Sandy Bridge, which predates Haswell. Running the build on these older CPUs thus resulted in illegal instruction exceptions.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] manually tests


Dronplane (approved):
LGTM

KVS85 (approved):
LGTM.

PR #16148, opened 2022-04-20
Ensure snappy does not use BMI2 instructions.

mpoeter:
### Scope & Purpose

For the Windows build, the new Snappy version, which was introduced in 3.9, generated code that contained BMI2 instructions which where introduced with the Intel Haswell architecture. However, our target architecture for 3.9 is actually Sandy Bridge, which predates Haswell. Running the build on these older CPUs thus resulted in illegal instruction exceptions.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] manually tests
- [x] Backports
  - [x] Backport for 3.9: #16147


jsteemann (approved):
Code changes LGTM, but should be verified by someone with access to a Windows OS.

Dronplane:
Changes are fine by itself. But that will disable BMI2 for Windows  even if we will switch to the next arch. Suggest to check what will be the arch for next major release and maybe we don't need changes for devel.

Dronplane (approved):
LGTM. See minor comment.

KVS85:
> Changes are fine by itself. But that will disable BMI2 for Windows even if we will switch to the next arch. Suggest to check what will be the arch for next major release and maybe we don't need changes for devel.

3.10 stays on Sandy Bridge atm. @mpoeter will try to add `TARGET_ARCHITECTURE` proper support for Windows (including fixing `OptimizeForArchitecture.cmake` module). Then we probably would change this to rely on this rather then own Snappy anlysis.

PR #16157, opened 2022-04-22
Reduce assert overhead.

mpoeter:
### Scope & Purpose

Previously every `TRI_ASSERT` created a temporary `std::ostringstream` instance in case the assertion fails. But unfortunately the compiler always reserves the necessary stack space (376 bytes) and has to generate some additional code for destroying the stream. This PR replaces the temporary `std::ostringstream` instance with thread-local instances. Therefore we no longer have to reserve any stack space and also don't need to clean up. 

This change only affects maintainer mode.

- [x] :hankey: Bugfix


mpoeter:
Closed in favor of #16308

PR #16169, opened 2022-04-26
Adjust importing test structure to be more similar to dumping

dothebart:
### Scope & Purpose

- make imporiting binary config and handling similar to dumping.
- move client tool invocation function up
-  fix exception handling in the importing testsuites so it actually produces errors in the result. 

### Checklist

- [x] Tests


adityamukho (approved):
For some reason, the diff was misaligned causing identical code to show up in the diff. Otherwise, LGTM.

PR #16174, opened 2022-04-26
[CINFRA-364] Supervision Model Checker

maierlars:
### Scope & Purpose
This PR tries to implement a model checker for the replicated log and state supervision. It allows to specify an initial condition and then all possible states are explored. Using assertions one check that certain invariants are never violated and goals are eventually reached.


tests/Replication2/ModelChecker/ActorModel.cpp
```diff
@@ -0,0 +1,23 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Lars Maier
+////////////////////////////////////////////////////////////////////////////////
+
+#include "ActorModel.h"
```
markuspf, line 23: Empty file

markuspf (approved):

PR #16189, opened 2022-04-28
add testsuite missing in backport

dothebart:
### Scope & Purpose

in https://github.com/arangodb/arangodb/pull/16084 we missed adding the converted http_replication testsuite from the devel PR. This commit brings it back, and flushes the testsuites from the blackists.

- [x] :hankey: Bugfix

PR #16200, opened 2022-04-29
as sugested by @adityamukho use Array.includes()

dothebart:
### Scope & Purpose

this test was converted from ruby to js and in the midst of https://github.com/arangodb/arangodb/pull/16189 we discovered that `Array.includes()` must be used.


adityamukho (approved):
LGTM.

PR #16201, opened 2022-04-29
Improve UnshackledMutex via absl::Mutex

MBkkt:
### Scope & Purpose

One futex instead of two

Props:
* Simpler
* Have a kind of fast path (fast lock with cond, unlock without waiters). 
  So it better when there is no contention on mutex
* `noexcept`

Cons:
* Every unlock is possible notify. So we make unnecessary condition check in `absl::Mutex::unlock` in `UnshackledMutex::lock`

About notification counts.
I think it is smaller with absl, because:
* with std we have always have `notify_one`, and two `std::mutex::unlock`
* with absl only two `absl::Mutex::unlock`

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/lib/Basics/UnshackledMutex.cpp
+++ b/lib/Basics/UnshackledMutex.cpp
@@ -25,28 +25,29 @@
 
 #include "Basics/debugging.h"
 
-void arangodb::basics::UnshackledMutex::lock() {
-  auto guard = std::unique_lock(_mutex);
-  _cv.wait(guard, [&] { return !_locked; });
-  TRI_ASSERT(!_locked);
+namespace arangodb::basics {
+
+void UnshackledMutex::lock() noexcept {
+  absl::MutexLock guard{
+      &_mutex,
+      absl::Condition{+[](bool const* locked) noexcept { return !*locked; },
+                      &std::as_const(_locked)}};
   _locked = true;
 }
 
-void arangodb::basics::UnshackledMutex::unlock() {
-  {
-    auto guard = std::unique_lock(_mutex);
-    TRI_ASSERT(_locked);
-    _locked = false;
-  }
-  _cv.notify_one();
+void UnshackledMutex::unlock() noexcept {
+  absl::MutexLock guard{&_mutex};
+  TRI_ASSERT(_locked);
+  _locked = false;
 }
 
-auto arangodb::basics::UnshackledMutex::try_lock() -> bool {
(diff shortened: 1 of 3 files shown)


goedderz (approved):
LGTM, thanks!

lib/Basics/UnshackledMutex.cpp
```diff
@@ -25,28 +25,29 @@
...
-  auto guard = std::unique_lock(_mutex);
-  _cv.wait(guard, [&] { return !_locked; });
-  TRI_ASSERT(!_locked);
+namespace arangodb::basics {
+
+void UnshackledMutex::lock() noexcept {
+  absl::MutexLock guard{
+      &_mutex,
+      absl::Condition{+[](bool const* locked) noexcept { return !*locked; },
+                      &std::as_const(_locked)}};
```
mpoeter, line 34: What is this `+` doing here? `absl::Condition{  +  [](bool const* locked)`

mpoeter (approved):
One minor question, otherwise LGTM!

lib/Basics/UnshackledMutex.cpp
MBkkt: Needs to get function pointer instead of lambda type

MBkkt: absl::Condition have explicit ctor if I remember correctly

MBkkt: Hmm, no. I think it needed for template deduction
https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/mutex.h#L671

MBkkt: >Needs to get function pointer instead of lambda type

Its valid C++ way to get it
https://stackoverflow.com/questions/18889028/a-positive-lambda-what-sorcery-is-this

mpoeter: Thanks, I have never seen this, but I think I could have used this already on occasion. :)

PR #16217, opened 2022-05-03
[CINFRA-325] Prototype State API update

maierlars:
### Scope & Purpose
This PR adds a few features to the prototype state api. For example `readFrom` that allows reading on followers.


apetenchea (approved):
LGTM

PR #16227, opened 2022-05-05
Make rocksdb a submodule and update it to v7.2

mpoeter:
### Scope & Purpose

- [x] :pizza: New feature

### Checklist


3rdParty/README_maintainers.md
```diff
@@ -126,72 +126,15 @@ https://github.com/lz4/lz4
 
 ## rocksdb
 
-(6.29.0, upstream commit e8f116deabc601a5eae011f2fd8a38d02d87be8d)
+v6.29.5
```
jsteemann, line 129: Isn't this PR upgrading the RocksDB version to 7.2?

mpoeter: Fixed!

gnusi:
@jsteemann @mpoeter Do you really think that's a good idea to reference rocksdb repo? That will prevent us from making changes in rocksdb codebase if necessary, i.e. in case of urgent bug-fix.
Wouldn't be better to fork rocksdb in a separate repo under arangodb organization and reference it instead?

jsteemann:
> @jsteemann @mpoeter Do you really think that's a good idea to reference rocksdb repo? That will prevent us from making changes in rocksdb codebase if necessary, i.e. in case of urgent bug-fix. Wouldn't be better to fork rocksdb in a separate repo under arangodb organization and reference it instead?

This PR references a fork of rocksdb inside the `arangodb` organization. So I think it is fine.

jsteemann:
CMake fails with
```
-- Could NOT find Snappy (missing: Snappy_DIR)
CMake Error at cmake/FindPackageHandleStandardArgs.cmake:21 (message):
  Could NOT find Snappy (missing: Snappy_LIBRARIES Snappy_INCLUDE_DIRS)
Call Stack (most recent call first):
  cmake/FindPackageHandleStandardArgs.cmake:261 (_FPHSA_FAILURE_MESSAGE)
  3rdParty/rocksdb/cmake/modules/FindSnappy.cmake:17 (find_package_handle_standard_args)
  3rdParty/rocksdb/CMakeLists.txt:140 (find_package)
```

PR #16250, opened 2022-05-11
Storage Engine Performance Prediction (SEPP)

mpoeter:
### Scope & Purpose

Introduce SEPP for dedicated storage engine performance tests.

The executable can be run as follows: 
`sepp <config-file> [--report=<report-file>] [-- <some.attribute.path>=<value> ...]`

The config file is a json file which specifies how to setup the database an which workload to run (a sample config file is checked in).
When a report file is specified, it will write some statistics in json format to that file. If the file already exists (e.g. from a previous run), it will attempt to load that file and simply append the result from the current run as a new entry in the JSON array. That way one can run multiple rounds of the same test and produce a single report file with all results to perform some statistical analysis.

The additional `attribute=value` pairs are merged into the config from the config-file so one can easily run tests with varying parameter values.

- [x] :pizza: New feature


arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -810,26 +719,20 @@ void RocksDBEngine::start() {
...
 
-  // intentionally set the RocksDB logger to warning because it will
-  // log lots of things otherwise
   if (_debugLogging) {
     _options.info_log_level = rocksdb::InfoLogLevel::DEBUG_LEVEL;
-  } else {
-    if (!opts._useFileLogging) {
-      // if we don't use file logging but log into ArangoDB's logfile,
-      // we only want real errors
-      _options.info_log_level = rocksdb::InfoLogLevel::ERROR_LEVEL;
```
jsteemann, line 826: We are not explicitly setting the log level to `ERROR_LEVEL` for normal operations.
Can this have a negative impact on performance, thanks to the way that log messages are evaluated by RocksDB? (just a speculation, haven't checked)

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <filesystem>
+#include <memory>
+
+#include "Workload.h"
+
+namespace {
+std::size_t getFolderSize(std::string_view path) {
+  return std::accumulate(std::filesystem::recursive_directory_iterator(path),
+                         std::filesystem::recursive_directory_iterator(), 0ull,
+                         [](auto size, auto& path) {
```
jsteemann, line 36:
```suggestion
                         [](auto size, auto const& path) {
```

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,144 @@
...
+  waitUntilAllThreadsAre(ThreadState::finished);
+
+  std::chrono::duration<double, std::milli> runtime =
+      std::chrono::high_resolution_clock::now() - start;
+  return buildReport(runtime.count());
+}
+
+Report Execution::buildReport(double runtime) {
+  for (auto& thread : _threads) {
+    thread->_thread.join();
```
jsteemann, line 114: The join method used here is different to what's used in the destructor:
https://github.com/arangodb/arangodb/pull/16250/files?w=1#diff-81785ac87c72ccdc01f12636ec26ad54a3eeab43c8da68e233250ed9b39a4492R53-R57

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  std::vector<ThreadReport> threadReports;
+  threadReports.reserve(_threads.size());
+  for (auto& thread : _threads) {
+    threadReports.push_back(thread->report());
+  }
+  return {.threads = threadReports,
```
jsteemann, line 122:
```suggestion
  return {.threads = std::move(threadReports),
```

tests/sepp/Execution.h
```diff
@@ -0,0 +1,75 @@
...
+#include "Options.h"
+#include "Report.h"
+
+namespace arangodb::sepp {
+
+struct Workload;
+struct Server;
+
+enum class ExecutionState {
+  starting,
```
jsteemann, line 38: These enum values should be named `kStarting`, `kPreparing`, ... according to our coding guidelines.

tests/sepp/Options.h
```diff
@@ -0,0 +1,107 @@
...
+
+  WorkloadVariants workload;
+
+  RocksDBOptions rocksdb;
+};
+
+template<class Inspector>
+auto inspect(Inspector& f, Options& o) {
+  return f.object(o).fields(
+      f.field("databaseDirectory", o.databaseDirectory).fallback("/tmp/sepp"),
```
jsteemann, line 101: Nit: hard-coding the path here will not be thread-safe. We obviously won't run the performance tests in parallel, but I think we should still use some functionality here to create a unique filename. Probably something like `TRI_GetTempName()`.
```suggestion
      f.field("databaseDirectory", o.databaseDirectory).fallback("/tmp/sepp"),
```

tests/sepp/Report.h
```diff
@@ -0,0 +1,80 @@
...
+  [[nodiscard]] std::uint64_t operations() const {
+    std::uint64_t result = 0;
+    for (const auto& thread : threads) {
+      result += thread.operations;
+    }
+    return result;
+  }
+
+  [[nodiscard]] double throughput() const {
+    return static_cast<double>(operations()) / runtime;
```
jsteemann, line 66: nit: this should guard against `runtime` potentially being `0`.

tests/sepp/Workloads/InsertDocuments.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+void InsertDocuments::Thread::run() {
+  // TODO - make more configurable
+  const unsigned numOps = 10;
+  for (unsigned i = 0; i < numOps; ++i) {
+    auto trx = std::make_unique<SingleCollectionTransaction>(
+        transaction::StandaloneContext::Create(*_server.vocbase()),
+        _options.collection, AccessMode::Type::WRITE);
+
+    trx->begin();
```
jsteemann, line 96: does not handle errors right now.

tests/sepp/Workloads/InsertDocuments.cpp
```diff
@@ -0,0 +1,111 @@
...
+    trx->insert("testcol", _options.object->slice(), {});
```
jsteemann, line 97: Name "testcol" is hard-coded here. It should be dynamic.
```suggestion
    trx->insert("testcol", _options.object->slice(), {});
```

tests/sepp/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+struct Options {
+  std::string_view configFile;
+  std::string_view report;
+  std::vector<std::string_view> params;
+};
+
+std::pair<std::string_view, std::string_view> splitKeyValue(
+    std::string_view s) {
+  auto pos = s.find("=");
```
jsteemann, line 59:
```suggestion
  auto pos = s.find('=');
```

tests/sepp/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+  if (pos == std::string::npos) {
+    throw InvalidArgumentException(s);
+  }
+  return {s.substr(0, pos), s.substr(pos + 1)};
+}
+
+void addConfig(arangodb::velocypack::Builder& config, std::string_view param) {
+  auto [path, v] = splitKeyValue(param);
+  auto value = arangodb::velocypack::Parser::fromJson(std::string(v));
```
jsteemann, line 68:
```suggestion
  auto value = arangodb::velocypack::Parser::fromJson(v.data(), v.size());
```

tests/sepp/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  arangodb::velocypack::Builder builder;
+  builder.openObject();
+  auto fieldPos = path.find(".");
+  for (;;) {
+    builder.add(VPackValue(path.substr(0, fieldPos)));
+    if (fieldPos != std::string::npos) {
+      builder.openObject();
+      path = path.substr(fieldPos + 1);
+      fieldPos = path.find(".");
```
jsteemann, line 78:
```suggestion
      fieldPos = path.find('.');
```

tests/sepp/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  try {
+    Options options;
+    parseOptions(argc, argv, options);
+
+    auto config = parseConfig(options);
+    arangodb::sepp::Runner runner(argv[0], options.report, config->slice());
+    runner.run();
+
+  } catch (const InvalidArgumentException& e) {
```
jsteemann, line 151:
```suggestion
  } catch (InvalidArgumentException const& e) {
```

tests/sepp/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+    std::cerr << "Invalid argument: " << e.what() << std::endl;
+    printUsage();
+    return 1;
+  } catch (const std::exception& e) {
```
jsteemann, line 155:
```suggestion
  } catch (std::exception const& e) {
```

arangod/RocksDBEngine/RocksDBEngine.cpp
mpoeter: This logic has been moved here:
https://github.com/arangodb/arangodb/blob/d6f0bf93597be53b27db6de6d8252016ee52a5d1/arangod/RocksDBEngine/RocksDBOptionFeature.cpp#L802-L806

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+void Execution::waitUntilThreadStateIs(ExecutionThread const& thread,
+                                       ThreadState expected) {
+  auto state = thread._state.load(std::memory_order_relaxed);
+  while (state != expected) {
+    if (state == ThreadState::kFinished) {
+      throw std::runtime_error("worker thread finished prematurely");
+    }
+    state = thread._state.load(std::memory_order_relaxed);
```
jsteemann, line 142: Just realized that this will be a busy-wait loop.
That doesn't sound good in general, but I guess we want it anyway for the following call to this function:
```
  _state.store(ExecutionState::kStopped);

  waitUntilAllThreadsAre(ThreadState::kFinished);

  std::chrono::duration<double, std::milli> runtime =
      std::chrono::high_resolution_clock::now() - start;
```
If we would go to sleep inside here, it would influence the timing results.

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+void Execution::waitUntilAllThreadsAre(ThreadState state) {
+  for (auto& thread : _threads) {
+    waitUntilThreadStateIs(*thread, state);
+  }
+}
+
+void Execution::waitUntilThreadStateIs(ExecutionThread const& thread,
+                                       ThreadState expected) {
```
jsteemann, line 136:
```suggestion
                                       ThreadState expected) const {
```

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,146 @@
...
+  for (auto& thread : _threads) {
+    threadReports.push_back(thread->report());
+  }
+  return {.threads = std::move(threadReports),
+          .runtime = runtime,
+          .databaseSize = getFolderSize(_options.databaseDirectory)};
+}
+
+void Execution::waitUntilAllThreadsAre(ThreadState state) {
+  for (auto& thread : _threads) {
```
jsteemann, line 130:
```suggestion
void Execution::waitUntilAllThreadsAre(ThreadState state) const {
  for (auto const& thread : _threads) {
```

tests/sepp/Execution.cpp
```diff
@@ -0,0 +1,146 @@
...
+      thread->_thread.join();
+    }
+  }
+}
+
+void Execution::createThreads(Server& server) {
+  _threads = _workload->createThreads(*this, server);
+}
+
+ExecutionState Execution::state(std::memory_order order) const {
```
jsteemann, line 64:
```suggestion
ExecutionState Execution::state(std::memory_order order) const noexcept {
```

tests/sepp/ExecutionThread.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  while (_execution.state() == ExecutionState::kRunning && !shouldStop()) {
+    run();
+  }
+
+  _runtime = std::chrono::high_resolution_clock::now() - start;
+  _execution.signalFinishedThread();
+}
+
+void ExecutionThread::waitUntilAllThreadsAreStarted() {
```
jsteemann, line 75:
```suggestion
void ExecutionThread::waitUntilAllThreadsAreStarted() const {
```

tests/sepp/ExecutionThread.cpp
```diff
@@ -0,0 +1,94 @@
...
+  while (_execution.state(std::memory_order_acquire) ==
+         ExecutionState::kStarting) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+  }
+}
+
+void ExecutionThread::waitUntilInitialization() {
```
jsteemann, line 82:
```suggestion
void ExecutionThread::waitUntilInitialization() const {
```

tests/sepp/ExecutionThread.cpp
```diff
@@ -0,0 +1,94 @@
...
+  while (_execution.state() == ExecutionState::kPreparing) {
+    ;
```
jsteemann, line 84: busy waiting... Intentional?

tests/sepp/ExecutionThread.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+void ExecutionThread::waitUntilBenchmarkStarts() {
```
jsteemann, line 88:
```suggestion
void ExecutionThread::waitUntilBenchmarkStarts() const {
```

tests/sepp/ExecutionThread.cpp
```diff
@@ -0,0 +1,94 @@
...
+  while (_execution.state() == ExecutionState::kInitializing) {
+    ;
```
jsteemann, line 90: busy waiting... intentional?

tests/sepp/Report.h
```diff
@@ -0,0 +1,83 @@
...
+  // TODO - rocksdb statistics
+  // velocypack::SliceContainer config;
+
+  std::vector<ThreadReport> threads;
+  double runtime;  // runtime in milliseconds
+  std::uint64_t databaseSize;
+
+  [[nodiscard]] std::uint64_t operations() const {
+    std::uint64_t result = 0;
+    for (const auto& thread : threads) {
```
jsteemann, line 59:
```suggestion
    for (auto const& thread : threads) {
```

tests/CMakeLists.txt
```diff
@@ -433,6 +433,36 @@ target_link_libraries(arangodbtests
   fmt
 )
 
+set(ARANGODB_SEPP_SOURCES
```
jsteemann, line 436: I am happy if the sepp executable gets built whenever arangodbtests is built, as it is currently implemented.
I don't think this will cause issues for release packages, because there we don't build the tests. 
The only potential issue I see here is that the extra executable will take up disk space and increase link times, and as it is statically linked, potentially about 1 GB with debug symbols.
I don't expect this to be a problem, but we need to check.

jsteemann (approved):
LGTM

tests/sepp/Execution.cpp
mpoeter: Yes, this as well as the other loops are intentionally busy loops since they should only wait for a rather small time and we want all threads to start at the same time (alltough with a longer runtime small delays during start should not matter much). I can put in a `pause` to reduce the CPU load though...

I would have liked to use a std::barrier instead of the busy loop, but I think some compiler is still lacking support. So a busy loop was the simplest solution (didn't want to implement my own barrier).

tests/CMakeLists.txt
mpoeter: Sepp can now only be built if the cmake variable `BUILD_SEPP` is enabled.

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace {
+constexpr std::string_view charset =
+    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.,";
+
+}
+
+namespace arangodb::sepp::generators {
+
+RandomStringGenerator::RandomStringGenerator(std::uint32_t size) : _size(size) {
+  _prng.seed(0, 0xdeadbeefdeadbeefULL);
```
jsteemann, line 43: Seeding like a boss.

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  constexpr std::uint32_t charsetSize = 64;
+  static_assert(charset.size() == charsetSize);
+  constexpr std::uint32_t bitsPerChar = 6;
+  constexpr std::uint32_t charsPerRound = 64u / bitsPerChar;
+
+  auto rounds = _size / charsPerRound;
+  for (unsigned i = 0; i < rounds; ++i) {
+    auto v = _prng.next();
+    for (unsigned j = 0; j < charsPerRound; ++j) {
```
jsteemann, line 58: As `charsPerRound` is always `10`, do we always make 10 iterations here?
What if `v` is a low value (by chance) and all its bits have been shifted out in the previous rounds? The character generator will then favor the character `a` over others.
Not really a problem, because we don't rely on proper randomness here, just mentioning it.

jsteemann (approved):
LGTM!

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
mpoeter: Actually there should be a TODO to seed properly so that different threads use different (but deterministic seeds). 🤔

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
mpoeter: Yes, I considered this as well, but since we don't rely on proper randomness I figured it would be good enough.

tests/sepp/Runner.cpp
```diff
@@ -0,0 +1,194 @@
...
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Manuel Pöter
+////////////////////////////////////////////////////////////////////////////////
+
+#include "Runner.h"
+
+#include <fstream>
```
jsteemann, line 26: I think when using `std::cout` and such, you are supposed to included `<iostream>`. But it seems to work without as well.

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
jsteemann: Ok, will you add the TODO here, so we don't forget about it?

tests/sepp/ValueGenerators/RandomStringGenerator.cpp
jsteemann: Ok, it is probably good enough for now.

tests/sepp/ValueGenerators/RandomStringGenerator.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "Basics/xoroshiro128plus.h"
+
+namespace arangodb::sepp::generators {
+
+struct RandomStringGenerator : ValueGenerator {
+  RandomStringGenerator(std::uint32_t size);
+
+  /// @brief Generates a value and writes it to the given Builder.
+  void apply(velocypack::Builder&) override;
```
jsteemann, line 37: `Builder` is undeclared in this header, and only transitively made available.
Let's follow IWYU (include-what-you-use) and either include `velocypack/Builder.h` or forward-declare it.

jsteemann (approved):
LGTM!

tests/sepp/ValueGenerators/RandomStringGenerator.h
mpoeter: The forward declaration is in `ValueGenerator.h` which we rely on since we are deriving from `ValueGenerator` and this is an override of a pure-virtual function from `ValueGenerator`. So IMHO we do not need any additional includes or forward declarations in this file.

PR #16253, opened 2022-05-11
[APM-135] Parallelism in non-unique index creation 

cpjulia:
### Scope & Purpose

https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr/
This PR is related to https://arangodb.atlassian.net/browse/APM-135. Added multithreading to assigning non-unique indexes to documents, in foreground or background mode. The number of threads is hardcoded to 2 because, at least for the moment, there was not much improvement with more threads, and the  threshold amount of documents a thread will process before enqueuing the rest of the documents in the range of document ids is hardcoded to 100000. The threshold of number of documents to start using 2 threads instead of 1 in non unique index creation is > 120000.
As it's an internal change i.e. only apparent to the user in terms of small performance improvement, one can ormally create an index like:
``` 
db.collectionTest.ensureIndex({ type: "persistent", fields: [ "foo" ], unique: false, name: "foo", sparse: true});
```
Multithreading for unique index is to be added.
Also, after rocksdb improves the file ingestion, the code with USE_SST_INGESTION will be used.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  No backports required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/APM-135
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,12 @@
 devel
 -----
 
+* APM-135: Added multithreading to assigning non-unique indexes to documents,
+  in foreground or background mode. The batch size for the thread to process 
+  before splitting its partitions is given by the  parameter `threadBatchSize` 
+  on index creation. The number of threads is hardcoded to 2 and the thread 
+  batch size is hardcoded to 100000 for now.
```
jsteemann, line 8:
```suggestion
  in foreground or background mode. The number of index creation threads 
  is hardcoded to 2 for now. Improvements for higher parallelism are expected
  for future versions.
```

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -194,6 +194,8 @@ class RocksDBEngine final : public StorageEngine {
   std::string databasePath(TRI_vocbase_t const* /*vocbase*/) const override {
     return _basePath;
   }
+  std::string idxPath() { return _idxPath; }
```
jsteemann, line 197: Can be made const.
```suggestion
  std::string idxPath() const { return _idxPath; }
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -118,10 +118,23 @@
...
 using namespace arangodb;
 using namespace arangodb::application_features;
 using namespace arangodb::options;
 
+#ifdef USE_SST_INGESTION
+namespace {
+void cleanUpExtFiles(std::string_view path) {
+  for (auto const& fileName : TRI_FullTreeDirectory(path.data())) {
+    TRI_UnlinkFile(
+        basics::FileUtils::buildFilename(path.data(), fileName.data()).data());
```
jsteemann, line 132:
```suggestion
        basics::FileUtils::buildFilename(path, fileName).data());
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -118,10 +118,23 @@
...
 #include <limits>
 
+#undef USE_SST_INGESTION
+
 using namespace arangodb;
 using namespace arangodb::application_features;
 using namespace arangodb::options;
 
+#ifdef USE_SST_INGESTION
+namespace {
```
jsteemann, line 128: Readers of this code may wonder about the `#ifdef` here and try to find an according `#define`. As they will not find one, they may wonder if this code can be deleted.
Please leave a code comment here that explains the motivation for the ifdef and that we want to keep the code around for later usage, once the issue is fixed in RocksDB that is keeping us from using SST file ingestion.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -23,11 +23,19 @@
 
 #include "RocksDBBuilderIndex.h"
 
+#undef USE_SST_INGESTION
```
jsteemann, line 26: Readers of this code may wonder about the `#ifdef`, `#undef` here and try to find an according `#define`. As they will not find one, they may wonder if this code can be deleted.
Please leave a code comment here that explains the motivation for all of this, and that we want to keep the code around for later usage, once the issue is fixed in RocksDB that is keeping us from using SST file ingestion.

arangod/RocksDBEngine/RocksDBBuilderIndex.h
```diff
@@ -23,15 +23,88 @@
 
 #pragma once
 
+#include "Basics/ConditionLocker.h"
```
jsteemann, line 26: I don't think the `ConditionLocker` is required anymore, as you have switched to `std::condition_variable`.
```suggestion
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -23,11 +23,19 @@
...
+
 #include "ApplicationFeatures/ApplicationServer.h"
+#include "ApplicationFeatures/TempFeature.h"
```
jsteemann, line 29: Not needed here?
```suggestion
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -55,30 +64,296 @@ using namespace arangodb;
...
+    _trx.addHint(transaction::Hints::Hint::LOCK_NEVER);
+  }
+  _trx.addHint(transaction::Hints::Hint::INDEX_CREATION);
+  Result res = _trx.begin();
+  if (!res.ok()) {
+    THROW_ARANGO_EXCEPTION(res);
+  }
+  _trxColl = _trx.resolveTrxCollection();
+  if (_isUniqueIndex) {
+    // for later
```
jsteemann, line 143: Let's put an assertion here, so that this cannot be called with a unique index accidentially.

arangod/RocksDBEngine/Methods/RocksDBSstFileMethods.cpp
```diff
@@ -0,0 +1,180 @@
...
+        uint64_t seq = _rootDB->GetLatestSequenceNumber();
+        // since cuckoo estimator uses a map with seq as key we need to
+        estimator->bufferUpdates(seq, std::move(it->second.inserts),
+                                 std::move(it->second.removals));
+      }
+    }
+  }
+}
+
+rocksdb::Status RocksDBSstFileMethods::writeToFile() {
```
jsteemann, line 73: Can we make this method return an `arangodb::Result` instead of a `rocksdb::Status`?
There is not much need for it to return a `rocksdb::Status`, because it is not an overridden method.

arangod/RocksDBEngine/Methods/RocksDBSstFileMethods.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (!res.ok()) {
+      cleanUpFiles();
+    } else {
+      insertEstimators();
+    }
+  }
+  return res;
+}
+
+rocksdb::Status RocksDBSstFileMethods::stealFileNames(
```
jsteemann, line 112: Can we make this method return an `arangodb::Result` instead of a `rocksdb::Status`?
There is not much need for it to return a `rocksdb::Status`, because it is not an overridden method.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -167,24 +442,80 @@ Result RocksDBBuilderIndex::remove(transaction::Methods& trx,
...
+  for (size_t i = 0; i < numThreads; ++i) {
+    auto newThread = std::make_unique<IndexCreatorThread>(
+        false, isForeground, threadBatchSize, docsProcessed, sharedWorkEnv,
+        rcoll, rootDB, ridx, snap, dbOptions, idxPath);
+    idxCreatorThreads.emplace_back(std::move(newThread));
+  }
+
+  try {
+    for (auto& idxCreatorThread : idxCreatorThreads) {
+      idxCreatorThread->start();
```
jsteemann, line 464: `Thread::start()` returns a boolean result, which is not handled here.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -167,24 +442,80 @@ Result RocksDBBuilderIndex::remove(transaction::Methods& trx,
...
+  sharedWorkEnv->waitUntilAllThreadsTerminate();
+
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+  uint64_t seekCounter = 2;
+  uint64_t nextCounter = 0;
+  for (auto const& threadStats : sharedWorkEnv->getThreadStatistics()) {
+    seekCounter += threadStats.numSeeks;
+    nextCounter += threadStats.numNexts;
+  }
+  LOG_DEVEL << "Total seeks: " << seekCounter << ", next: " << nextCounter;
```
jsteemann, line 481: We shouldn't LOG_DEVEL the statistics here.
Instead, let's remove all the `#ifdef ARANGODB_ENABLE_MAINTAINER_MODE` things for the statistics, so they are recorded unconditionally.
And here, we can log the statistics using a proper `LOG_TOPIC`, but give it a `DEBUG` log severity. That means the statistics will normally not be contained in the log files, but if one turns on debug logging, they will be logged. We also need to adjust the log message so it gets a bit more context (for example, it doesn't say that it is related to an index creation etc.).

arangod/RocksDBEngine/RocksDBBuilderIndex.h
```diff
@@ -122,8 +195,118 @@ class RocksDBBuilderIndex final : public arangodb::RocksDBIndex {
...
+  }
+
+  RocksDBKeyBounds const& getBounds() const { return _bounds; }
+
+  rocksdb::Slice const getUpperBound() const {
+    return rocksdb::Slice(_bounds.end());
+  }
+
+ private:
+  static constexpr size_t kNumThreads = 2;
```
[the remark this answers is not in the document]
cpjulia: This belongs to another class, then I'd have to forward it from the other one by the constructor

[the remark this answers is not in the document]
jsteemann: Yes, it is another class, I see it. But I don't see a reason to have it twice.
If you make the constexpr public in the other class, you access it from anywhere.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
jsteemann: Comment not addressed.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Comment not addressed.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -118,10 +118,23 @@
...
+void cleanUpExtFiles(std::string_view path) {
```
jsteemann, line 129: Quick question: what does `Ext` stand for here? It is not obvious to the reader. Should this be "temp" or "temporary"?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -675,6 +689,21 @@ void RocksDBEngine::start() {
     }
   }
 
+#ifdef USE_SST_INGESTION
+  _idxPath = basics::FileUtils::buildFilename(_path, "tmp-idx-creation");
+  if (basics::FileUtils::isDirectory(_idxPath.data())) {
```
jsteemann, line 694:
```suggestion
  if (basics::FileUtils::isDirectory(_idxPath)) {
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -675,6 +689,21 @@ void RocksDBEngine::start() {
...
+    ::cleanUpExtFiles(_idxPath.data());
```
jsteemann, line 695:
```suggestion
    ::cleanUpExtFiles(_idxPath);
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
jsteemann: Comment not addressed.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -167,24 +437,79 @@ Result RocksDBBuilderIndex::remove(transaction::Methods& trx,
...
+  sharedWorkEnv->waitUntilAllThreadsTerminate();
+
+  uint64_t seekCounter = 2;
+  uint64_t nextCounter = 0;
+  for (auto const& threadStats : sharedWorkEnv->getThreadStatistics()) {
+    seekCounter += threadStats.numSeeks;
+    nextCounter += threadStats.numNexts;
+  }
+  LOG_TOPIC("d9bf2", DEBUG, Logger::ENGINES)
+      << "Total seeks: " << seekCounter << ", next: " << nextCounter;
```
jsteemann, line 476: Can we add some context to this log message. Otherwise it is unclear that the seeks and next calls are originating from an index creation. Any context here would improve the log message.

arangod/RocksDBEngine/Methods/RocksDBSstFileMethods.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBEngine.cpp
cpjulia: Stands for external, as, before rocksdb moved them, they would be in an external non-standard directory, but changed name

arangod/RocksDBEngine/RocksDBEngine.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.h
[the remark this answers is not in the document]
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
cpjulia: done

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -167,24 +437,87 @@ Result RocksDBBuilderIndex::remove(transaction::Methods& trx,
...
+  sharedWorkEnv->waitUntilAllThreadsTerminate();
+
+  uint64_t seekCounter = 2;
+  uint64_t nextCounter = 0;
+  for (auto const& threadStats : sharedWorkEnv->getThreadStatistics()) {
+    seekCounter += threadStats.numSeeks;
+    nextCounter += threadStats.numNexts;
+  }
+  LOG_TOPIC("d9bf2", DEBUG, Logger::ENGINES)
+      << "Parallel index creation status Total seeks: " << seekCounter
```
jsteemann, line 483:
```suggestion
      << "Parallel index creation status. Total seeks: " << seekCounter
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -167,24 +437,87 @@ Result RocksDBBuilderIndex::remove(transaction::Methods& trx,
...
+      << ", next: " << nextCounter;
```
jsteemann, line 484:
```suggestion
      << ", number of next calls: " << nextCounter;
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -118,10 +118,25 @@
...
 using namespace arangodb;
 using namespace arangodb::application_features;
 using namespace arangodb::options;
 
+#ifdef USE_SST_INGESTION
+namespace {
+void cleanUpTempFiles(std::string_view path) {
+  for (auto const& fileName : TRI_FullTreeDirectory(path.data())) {
+    TRI_UnlinkFile(basics::FileUtils::buildFilename(path, fileName).data());
+  }
```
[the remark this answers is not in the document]
jsteemann: @mpoeter: let's not do it right now, as part of this PR. The code part is not active anyway right now (it's only used when using the SstFileWriter). So let's try to get this PR ready here with the "old" filesystem functions.

mpoeter (approved):
I think this code could use a bit more cleanup, but since we want this feature in 3.10 this should be done in a separate PR.

PR #16286, opened 2022-05-17
[CINFRA-374] Delete from Target -> Drop from Plan

maierlars:
### Scope & Purpose
When deleting a replicated log, the coordinator deletes the log from Target, Plan and Current. The supervision checks if a replicated log in Plan has the `owner` attribute set to `target`. If it finds such a log, it checks if a log with the same id is found in Target. If that is not the case, it deletes that replicated log.

Furthermore the supervision now adds a precondition to all modifications of plan, ensuring that target is still in place.


markuspf (approved):
Looks good to me;

PR #16303, opened 2022-05-18
Improve calculation of block cache shard bits

jsteemann:
### Scope & Purpose

* Change default value of `--rocksdb.block-cache-shard-bits` to an automatic
  default value that allows data blocks of at least 128MiB to be stored in each
  cache shard if the block cache's strict capacity limit is used. The strict
  capacity limit for the block cache is enabled by default in 3.10, but can be
  turned off by setting the option `--rocksdb.enforce-block-cache-size-limit`
  to `false`. Also log a startup warning if the resulting cache shard size
  would be smaller than is potentially safe when the strict capacity limit is
  set.
  Enforcing the block cache's capacity limit has the consequence that data
  reads by RocksDB must fit into the block cache or the read operation will
  fail with an "Incomplete" error.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBOptionFeature.cpp
```diff
@@ -634,6 +641,42 @@ void RocksDBOptionFeature::validateOptions(
...
+    // the block cache hasn't been set, we set it dynamically:
+    // we would like that each block cache shard can hold data blocks of
+    // at least a common size. Rationale: data blocks can be quite large. if
+    // they don't fit into the block cache upon reading, the block cache will
+    // return Status::Incomplete() when the block cache's strict capacity limit
+    // is set. then we cannot read any data anymore.
+    _blockCacheShardBits = std::clamp(
+        int64_t(std::floor(
+            std::log2(static_cast<double>(_blockCacheSize) / ::minShardSize))),
+        int64_t(1), int64_t(6));
```
mpoeter, line 657: Why limit to 6 here?

arangod/RocksDBEngine/RocksDBOptionFeature.cpp
```diff
@@ -634,6 +641,42 @@ void RocksDBOptionFeature::validateOptions(
...
+    // we may run into problems when trying to put a large data block into the
+    // cache. in this case the block cache may return a Status::Incomplete
+    // error and fail the entire read. warn the user about it!
+    if (shardSize < ::minShardSize) {
+      LOG_TOPIC("31d7c", WARN, Logger::ROCKSDB)
+          << "size of RocksDB block cache shards seems to be too low. "
+          << "block cache size: " << _blockCacheSize
+          << ", shard bits: " << _blockCacheShardBits
+          << ", shard size: " << shardSize
+          << ", shard bits: " << _blockCacheShardBits
```
mpoeter, line 675: Duplicate line
```suggestion
          << ", shard bits: " << _blockCacheShardBits
          << ", shard size: " << shardSize
```

mpoeter (approved):

PR #16308, opened 2022-05-19
Production Assert

maierlars:
### Scope & Purpose
This PR adds the `ADB_PROD_ASSERT` macro. In usage it is similar to `TRI_ASSERT`, however, it behaves differently.
1. The assertion is not disabled in production. If the assertion fails, the server will crash.
2. The debug message is always evaluated **in** Maintainer Mode. This is to make sure, that in production we do not crash while composing the assertion message.

Additionally this PR allocates the `std::ostringstream` in a thread local variable. This reduces the required stack size by a lot.


jsteemann (approved):
LGTM!

mchacki (approved):
LGTM,
good PR!

apetenchea (approved):

gnusi (approved):

PR #16313, opened 2022-05-20
[CINFRA-355] Only consider participants in Plan

maierlars:
### Scope & Purpose
There was a bug in the supervision, that would pick a server from current, without checking if it is still in plan. This was fixed. This PR adds a test for that.


markuspf (approved):
LGTM

PR #16316, opened 2022-05-20
[CINFRA] Log Supervision Spam Reduction

maierlars:
### Scope & Purpose
Tihs PR reduces the spam created by the log supervision in the agency log. It implements a similar mechanism compared to what the replicated state supervision does.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/Supervision.cpp
+++ b/arangod/Replication2/ReplicatedLog/Supervision.cpp
@@ -748,6 +748,9 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
     -> arangodb::agency::envelope {
   SupervisionContext sctx;
 
+  auto const hasStatusReport = log.current && log.current->supervision &&
+                               log.current->supervision->statusReport;
+
   sctx.enableErrorReporting();
 
   auto maxActionsTraceLength = std::invoke([&log]() {
@@ -763,15 +766,22 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
   auto actionCtx = arangodb::replication2::replicated_log::executeAction(
       std::move(log), sctx.getAction());
 
-  auto rep = sctx.getReport();
-  actionCtx.modifyOrCreate<LogCurrentSupervision>(
-      [&rep](LogCurrentSupervision& currentSupervision) {
-        if (rep.empty()) {
-          currentSupervision.statusReport.reset();
-        } else {
-          currentSupervision.statusReport = rep;
-        }
+  if (sctx.isErrorReportingEnabled()) {
+    if (sctx.getReport().empty()) {
+      if (hasStatusReport) {
+        actionCtx.modify<LogCurrentSupervision>(
+            [&](auto& supervision) { supervision.statusReport.reset(); });
+      }
+    } else {
+      actionCtx.modify<LogCurrentSupervision>([&](auto& supervision) {
+        supervision.statusReport = std::move(sctx.getReport());
       });
+    }
+  }
+
+  if (!actionCtx.hasModification()) {
+    return envelope;
+  }
 
   return buildAgencyTransaction(dbName, log.target.id, sctx, actionCtx,
                                 maxActionsTraceLength, std::move(envelope));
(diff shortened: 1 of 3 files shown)


arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -760,20 +771,54 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
...
+      // now compare the new status with the old status
+      if (log.current && log.current->supervision) {
+        if (log.current->supervision->statusReport == sctx.getReport()) {
+          // report did not change, do not create a transaction
+          return envelope;
+        }
+      }
+    }
+  }
+
```
markuspf, line 790: I am not sure all tthis clutter is worth avoiding just calling executeAction which then does nothing.

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -760,20 +771,54 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
...
+  if (sctx.isErrorReportingEnabled()) {
+    if (sctx.getReport().empty()) {
+      if (hasStatusReport) {
+        actionCtx.modify<LogCurrentSupervision>(
+            [&](auto& supervision) { supervision.statusReport.reset(); });
+      }
+    } else {
+      actionCtx.modify<LogCurrentSupervision>([&](auto& supervision) {
+        supervision.statusReport = std::move(sctx.getReport());
       });
```
markuspf, line 803: If one gives the `statusReport` to `executeAction`, too, this also goes away.

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -760,20 +771,54 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
...
+                 sctx.getAction())) {
+    actionCtx.modify<LogCurrentSupervision>(
+        [&](auto& supervision) { supervision.statusReport.reset(); });
+  }
+
+  // update last time modified
+  if (!hasNoExecutableAction) {
+    actionCtx.modify<LogCurrentSupervision>(
+        [&](auto& supervision) { supervision.lastTimeModified = now; });
+  }
```
markuspf, line 815: `executeAction` should do this; in particular actionCtx will know whether there have been modifications.

markuspf (commented):
I don't really like how convoluted `executeCheckReplicatedLog` has become; I think this can be significantly improved; see my suggestions;

We also merge this as is, and I do the changes as suggested later.

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -747,8 +747,19 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
...
+                               log.current->supervision->statusReport;
+
+  // check if error reporting is enabled
+  if (log.current && log.current->supervision &&
+      log.current->supervision->lastTimeModified.has_value()) {
+    auto const lastMod = *log.current->supervision->lastTimeModified;
+    if ((now - lastMod) > std::chrono::seconds{15}) {
+      sctx.enableErrorReporting();
+    }
+  }
```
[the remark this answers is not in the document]
maierlars: So is the `ok_threshold` and the `grace_period`, same holds true for the timeout on the log leader before it reports the commit fail reason. They are all arbitrary. We should first specify which problem we actually want to solve here. I tried to solve the following problem: we do not always want to collect reports. We want to collect them, whenever progress is not possible. However, detecting that progress is not possible is a distributed problem, hence we just guess locally that after some time without progress, something might got stuck and reporting something then might not hurt much. Now roll a dice and pick 15s. Could be 30s as well...

arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -747,8 +747,19 @@ auto executeCheckReplicatedLog(DatabaseID const& dbName,
                                arangodb::agency::envelope envelope) noexcept
     -> arangodb::agency::envelope {
   SupervisionContext sctx;
-
-  sctx.enableErrorReporting();
+  auto const now = std::chrono::system_clock::now();
+  auto const logId = log.target.id;
+  auto const hasStatusReport = log.current && log.current->supervision &&
+                               log.current->supervision->statusReport;
```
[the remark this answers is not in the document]
maierlars: Just another idea; we could rely on exceptions and just do something like
```log.current.value().supervision.value().statusReport```

maierlars:
I agree with your review and I think we can clearly improve these functions. [Here](https://arangodb.atlassian.net/browse/CINFRA-384) is a ticket that shall remind us.

I will proceed to merge it.

PR #16317, opened 2022-05-20
[CINFRA-385] Incorrect ongoing state in prototype

apetenchea:
### Scope & Purpose

The ongoing state of the prototype leader works like a deque: newer states, containing operations that were only applied locally, are always appended at the back. When state machine operations (log entries) have been replicated and committed, we look at the last log iterator to be applied and pop from the front of the deque until we reach that index.

The problem is that sometimes we could get entries that were not inserted by a user (meta entries), which means the prototype leader doesn’t know about the existence of these entries. These entries still add up to the log index, which could cause our leader to pop from the deque when it’s not supposed to do so.

Consider the following entries: 101 (operation), 102 (meta), 103 (operation). The deque will contain the states at 101, followed by the state at index 103. If log entries 101 and 102 come together, we would pop from the deque and move the state to 103, which is wrong, because we are supposed to stay at 101.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Replication2/StateMachines/Prototype/PrototypeCore.h
+++ b/arangod/Replication2/StateMachines/Prototype/PrototypeCore.h
@@ -142,12 +142,12 @@ void PrototypeCore::applyEntries(std::unique_ptr<EntryIterator> ptr) {
  */
 template<typename EntryIterator>
 void PrototypeCore::update(std::unique_ptr<EntryIterator> ptr) {
-  auto lastAppliedIndex = ptr->range().to.saturatedDecrement();
-  while (!_ongoingStates.empty() &&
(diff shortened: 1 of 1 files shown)


maierlars (approved):
LGTM 

maybe add a comment because, although the logic is perfectly correct, it is not obvious why we have to look at index 1.

PR #16322, opened 2022-05-23
[CINFRA-380] Remove replicationFactor from LogConfig

markuspf:
Removes the member `replicationFactor` from `LogConfig`, fixes tests.

`replicationFactor` was largely superfluous as it can be determined from the number of participants configured for a replicated log.


arangod/Replication2/Methods.cpp
```diff
@@ -283,15 +283,10 @@ struct ReplicatedLogMethodsCoordinator final
     auto dbservers = clusterInfo.getCurrentDBServers();
     std::size_t expectedNumberOfServers =
         std::min(dbservers.size(), std::size_t{3});
-    if (options.config.has_value()) {
-      expectedNumberOfServers = options.config->replicationFactor;
-    } else if (!options.servers.empty()) {
-      expectedNumberOfServers = options.servers.size();
-    }
+    expectedNumberOfServers = options.servers.size();
```
maierlars, line 286: Maybe we want to keep this feature of just specifying a number of servers for a replicated log or a prototype state by having a separate struct for the `CreateOptions` that has an optional `replicationFactor` in there?

arangod/Replication2/StateMachines/Prototype/PrototypeStateMethods.cpp
```diff
@@ -426,15 +426,12 @@ struct PrototypeStateMethodsCoordinator final
     auto dbservers = _clusterInfo.getCurrentDBServers();
     std::size_t expectedNumberOfServers =
         std::min(dbservers.size(), std::size_t{3});
-    if (options.config.has_value()) {
-      expectedNumberOfServers = options.config->replicationFactor;
-    } else if (!options.servers.empty()) {
+    if (!options.servers.empty()) {
```
maierlars, line 429: Same as above for replicated logs?

arangod/Replication2/Methods.cpp
markuspf: Maybe, but that is then a different user story

maierlars (approved):

PR #16327, opened 2022-05-24
[CINFRA-317] Drop Replicated States

maierlars:
### Scope & Purpose
When a replicated state is dropped from target, all its resources should be cleaned up (Log/Target) and its entries in Plan and Current should be removed as well.

### DIFF
--- a/tests/js/server/replication2/replication2-maintenance-replicated-state-cluster.js
+++ b/tests/js/server/replication2/replication2-maintenance-replicated-state-cluster.js
@@ -32,84 +32,65 @@ const SH = require("@arangodb/testutils/replicated-state-helper");
 const spreds = require("@arangodb/testutils/replicated-state-predicates");
 const lpreds = require("@arangodb/testutils/replicated-logs-predicates");
 const {dbservers} = require("@arangodb/testutils/replicated-logs-helper");
+const request = require('@arangodb/request');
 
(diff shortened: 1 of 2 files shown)


tests/js/server/replication2/replication2-maintenance-replicated-state-cluster.js
```diff
@@ -210,12 +191,61 @@ const replicatedStateSuite = function (stateType) {
...
+        for (const server of servers) {
+          const response = getLocalStatus(server, database, logId);
+          if (response.code !== 200) {
+            return Error(`Server ${server} returned code ${response.code}, expected ${200}`);
+          }
+        }
+        return true;
+      });
+
+      LH.replicatedLogDeletePlan(database, logId);
```
apetenchea, line 222: I find name of this test is a bit confusing. Did you intend to drop the replicated log in here? I think this line should not have any effect on the outcome of the test.

apetenchea (changes requested):
Looks good, but there's a problem which causes the agency to crash. 
To reproduce it, try creating a replicated state, delete it, and create it again using the same id. I attached the full log file to the Jira ticket.
After deletion of a replicated state (doesn't matter which type), some participants still report their snapshot status to current, which I find odd.

maierlars:
Alex created follow up tickets for this PR.

PR #16330, opened 2022-05-24
[CINFRA-363] Create replicated state for each shard

apetenchea:
### Scope & Purpose

Create replicated state for each shard, only on databases configured with replication 2.
What has been added:
- upon collection creation, we create a corresponding replicated state for every shard, having the same participants
- during a collection drop, we delete the corresponding replicated states
- during a database drop, we delete all replicated states for all database collections
All of the above happens only if the database has been configures to use replication 2.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cluster/ClusterInfo.cpp
```diff
@@ -513,6 +515,124 @@ void ClusterInfo::triggerBackgroundGetIds() {
...
+  return futures::collectAll(std::move(futureStates))
+      .thenValue(
+          [&clusterInfo = _server.getFeature<ClusterFeature>().clusterInfo()](
+              auto&& raftIndices) {
+            consensus::index_t highestRaftIndex = std::transform_reduce(
+                raftIndices.begin(), raftIndices.end(), (consensus::index_t)0,
+                [](consensus::index_t l, consensus::index_t r) {
+                  return std::max(l, r);
+                },
+                [](auto&& value) { return value.get().get(); });
```
mpoeter, line 567: I know I was the one who wrote this `transform_reduce`, but maybe a simple loop would be easier to read.
```suggestion
            consensus::index_t maxIndex = 0;
            for (auto& v : raftIndices) {
              maxIndex = std::max(maxIndex, v.get().get());
            }
```
Also, what if one of the `raftIndices` entries contains an error?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -513,6 +515,124 @@ void ClusterInfo::triggerBackgroundGetIds() {
...
+              [&] { return std::move(shardResult.get()); });
+          if (r.fail()) {
+            return result.mapError([](result::Error error) {
+              error.appendErrorMessage(
+                  "Failed to delete replicated states corresponding to "
+                  "shards!");
+              return error;
+            });
+          }
+        }
```
mpoeter, line 630: We can probably reduce some code duplication here.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3054,6 +3197,24 @@ Result ClusterInfo::createCollectionsCoordinator(
     opers.emplace_back(CreateCollectionOrder(databaseName, info.collectionID,
                                              info.isBuildingSlice()));
 
+    if (replicationVersion == replication::Version::TWO) {
+      // Create a replicated state for each shard.
+      replicatedStates.reserve(replicatedStates.size() + shardServers.size());
```
mpoeter, line 3202: Why `replicatedStates.size() + shardServers.size()`? Shouldn't this just be `shardServers.size()`?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3301,7 +3493,22 @@ Result ClusterInfo::createCollectionsCoordinator(
...
+         replicatedStatesWait.isReady())) {
+      if (replicationVersion == replication::Version::TWO) {
+        auto result = replicatedStatesWait.get();
+        if (result.fail()) {
+          LOG_TOPIC("ce2be", WARN, Logger::CLUSTER)
+              << "Failed createCollectionsCoordinator for " << infos.size()
+              << " collections in database " << databaseName
+              << " isNewDatabase: " << isNewDatabase
+              << " first collection name: " << infos[0].name
+              << " result: " << result.errorNumber();
```
mpoeter, line 3507: We should also log the error message here.

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -1212,6 +1217,14 @@ void LogicalCollection::decorateWithInternalValidators() {
   decorateWithInternalEEValidators();
 }
 
+replication2::LogId LogicalCollection::shardIdToLogId(ShardID const& shardId) {
+  auto stateId = std::string_view(shardId).substr(1, shardId.size() - 1);
+  auto logId = replication2::LogId::fromString(stateId);
+  ADB_PROD_ASSERT(logId.has_value())
+      << " converting " << shardId << " to LogId failed";
+  return logId.value();
+}
```
mpoeter, line 1226: Maybe we should move this into a static `fromShardId` function in `LogId` (similar to `fromString`).

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3628,9 +3835,33 @@ Result ClusterInfo::dropCollectionCoordinator(  // drop collection
...
+  if (coll->replicationVersion() == replication::Version::TWO) {
+    std::vector<replication2::LogId> stateIds;
+    for (auto pair : VPackObjectIterator(shardsSlice)) {
+      auto shardId = pair.key.copyString();
+      auto stateId = std::string_view{shardId.substr(1, shardId.size() - 1)};
+      auto logId = replication2::LogId::fromString(stateId);
+      ADB_PROD_ASSERT(logId.has_value())
+          << " converting " << stateId
+          << " to LogId failed, during replicated state deletion for shard "
+          << shardId;
```
mpoeter, line 3849: Couldn't we use `LogicalCollection::shardIdToLogId` here?

arangod/Cluster/ClusterInfo.cpp
apetenchea: Then we're going to handle the error inside the `.then()`, but that would cause the function to return a `Result` containing a failure. We could choose to ignore a failed `raftIndex`, but it is a delicate situation, because we might discard the maximum.

arangod/Cluster/ClusterInfo.cpp
apetenchea: Write the `r.fail()` part in a lambda and use the lambda twice?

arangod/Cluster/ClusterInfo.cpp
apetenchea: Unfortunately we are inside a huge for loop starting at line 3028, `for (auto& info : infos)`, hence we might end up with multiple `reserve` calls, depending on the number of collections. Not great, but didn't think of something else at that point.

arangod/VocBase/LogicalCollection.cpp
apetenchea: I did it like that initially, but on second thought I considered it would be too much to add knowledge about shards inside the `LogId` class. I figured that a `LogicalCollection` is the place where such a connection between the two takes place. This is debatable and I understand your point, it could make the method a bit more apparent to everyone looking for it.

arangod/Cluster/ClusterInfo.cpp
apetenchea: Yes, that would be nice.

arangod/VocBase/LogicalCollection.cpp
apetenchea: What do you think?

mpoeter: Makes sense!

arangod/Cluster/ClusterInfo.cpp
mpoeter: Yes, exactly...

PR #16343, opened 2022-05-27
[CINFRA-381] Move Log config from Term to ParticipantsConfig

markuspf:
### Scope & Purpose

Move Log config from Term to ParticipantsConfig; this means the log config is under the `generation` revision mechanism.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/Algorithms.cpp
+++ b/arangod/Replication2/ReplicatedLog/Algorithms.cpp
@@ -150,7 +150,7 @@ auto algorithms::updateReplicatedLog(
 
       TRI_ASSERT(spec->participantsConfig.generation > 0);
       auto newLeader = log->becomeLeader(
-          spec->currentTerm->config, myServerId, spec->currentTerm->term,
+          spec->participantsConfig.config, myServerId, spec->currentTerm->term,
           followers,
           std::make_shared<ParticipantsConfig>(spec->participantsConfig),
           std::move(failureOracle));
--- a/arangod/Replication2/ReplicatedLog/LogEntries.h
+++ b/arangod/Replication2/ReplicatedLog/LogEntries.h
@@ -27,7 +27,10 @@
 #include <velocypack/Buffer.h>
 #include <velocypack/Slice.h>
 
+#include "Replication2/ReplicatedLog/AgencyLogSpecification.h"
 #include "Replication2/ReplicatedLog/LogCommon.h"
+
+using namespace arangodb::replication2::agency;
 
 namespace arangodb::replication2 {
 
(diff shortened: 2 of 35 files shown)


arangod/Replication2/ReplicatedLog/Algorithms.cpp
```diff
@@ -150,7 +150,7 @@ auto algorithms::updateReplicatedLog(
 
       TRI_ASSERT(spec->participantsConfig.generation > 0);
       auto newLeader = log->becomeLeader(
-          spec->currentTerm->config, myServerId, spec->currentTerm->term,
+          spec->participantsConfig.config, myServerId, spec->currentTerm->term,
```
maierlars, line 153: This argument is now superfluous because the participantsConfig is also a parameter for the leader. But we can leave it as is and do the cleanup later.

arangod/Replication2/ReplicatedLog/LogEntries.h
```diff
@@ -27,8 +27,11 @@
 #include <velocypack/Buffer.h>
 #include <velocypack/Slice.h>
 
+#include "Replication2/ReplicatedLog/AgencyLogSpecification.h"
 #include "Replication2/ReplicatedLog/LogCommon.h"
 
+using namespace arangodb::replication2::agency;
```
maierlars, line 33: `using` in a header is unfriendly, because it pollutes the global namespace of all includers.  😞 I know ... I would like to have that as well.

arangod/Replication2/ReplicatedLog/LogStatus.h
```diff
@@ -35,6 +35,8 @@
 #include "Replication2/ReplicatedLog/types.h"
 #include "Replication2/ReplicatedLog/AgencyLogSpecification.h"
 
+using namespace arangodb::replication2::agency;
```
maierlars, line 38: `using` in header.

tests/Replication2/Helper/ModelChecker/HashValues.h
```diff
@@ -90,15 +92,15 @@ static inline std::size_t hash_value(ParticipantFlags const& s) {
   boost::hash_combine(seed, s.forced);
   return seed;
 }
+}  // namespace arangodb::replication2
+namespace arangodb::replication2::agency {
 static inline std::size_t hash_value(ParticipantsConfig const& s) {
   std::size_t seed = 0;
   boost::hash_combine(seed, s.generation);
   boost::hash_combine(seed, s.participants);
+  //  boost::hash_combine(seed, s.config);
```
maierlars, line 101: `config` was never part of the hash for LogPlan? 😟

maierlars (approved):

arangod/Replication2/ReplicatedLog/Algorithms.cpp
markuspf: I will do this in a separate PR, because it trickles down quite a bit.

PR #16354, opened 2022-05-31
[CINFRA-411] Mac Test Issues

maierlars:
### Scope & Purpose
Fixed some tests that depended on a specific timing or a specific order in which things happen. However, the supervision is free to take other paths that are valid. And the test should accept those as well.


markuspf (approved):
LGTM

PR #16363, opened 2022-06-02
[PREGEL] Spill dynamic WCC vertex data into MMAP buffer

markuspf:
### Scope & Purpose

This PR changes the Weakly Connected Component Pregel algorithm to use memory mapped storage for the (dynamically allocated) incoming edges;

It is inherent in the current Pregel implementation that on a given DB Server, loading a shard of vertices, we can only also load outgoing edges.

Since for WCC we need incoming edges as well, we cache sender ids on the first global superstep; the memory use of this step is somewhere in the order of the number of edges times the size of vertex-ids; a fairly significant number.

Hopefully this is only a temporary measure while we implement more robust memory management for pregel.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Pregel/Algos/WCC.cpp
+++ b/arangod/Pregel/Algos/WCC.cpp
@@ -50,6 +50,12 @@ struct WCCComputation
     // We need to propagate on first step
     TRI_ASSERT(globalSuperstep() != 0 || shouldPropagate);
 
+    // On the first step we cache all incoming neighbors
+    // and effectively convert the graph into an undirected one.
+    if (globalSuperstep() == 1) {
+      collectInNeighbors(messages);
+    }
+
     if (shouldPropagate) {
       propagate();
     }
@@ -62,6 +68,27 @@ struct WCCComputation
   }
 
  private:
+  void collectInNeighbors(
+      MessageIterator<SenderMessage<uint64_t>> const& messages) {
+    auto& myData = *mutableVertexData();
+    auto inNeighbors = std::vector<PregelID>();
+    auto inNeighborsByteSize = std::size_t{0};
+
+    for (const SenderMessage<uint64_t>* msg : messages) {
+      // In the first step, we need to retain all inbound connections
+      // for propagation
+      inNeighbors.emplace_back(msg->senderId);
+      inNeighborsByteSize +=
+          sizeof(WCCValue::InNeighbor) + msg->senderId.key.length();
+    }
+
+    auto range = this->allocateAuxiliaryStorage(inNeighborsByteSize);
+    myData.inNeighbors.init(range);
+
+    for (auto const& inNeighbor : inNeighbors) {
+      myData.inNeighbors.emplace(inNeighbor);
+    }
+  }
   /**
    * @brief Scan the input, compare it pairwise with our current value.
    * We store the minimum into our current value.
@@ -80,18 +107,9 @@ struct WCCComputation
     // Either the sender has a wrong component or we have.
     bool shouldPropagate = globalSuperstep() == 0;
 
-    auto inNeighbors = std::vector<PregelID>();
-    auto inNeighborsByteSize = std::size_t{0};
-
     auto& myData = *mutableVertexData();
+
     for (const SenderMessage<uint64_t>* msg : messages) {
-      if (globalSuperstep() == 1) {
-        // In the first step, we need to retain all inbound connections
-        // for propagation
-        inNeighbors.emplace_back(msg->senderId);
-        inNeighborsByteSize +=
-            sizeof(WCCValue::InNeighbor) + msg->senderId.key.length();
-      }
       if (msg->value != myData.component) {
         // we have a difference. Send updates
         shouldPropagate = true;
@@ -103,15 +121,6 @@ struct WCCComputation
       }
     }
 
-    if (globalSuperstep() == 1) {
-      auto range = this->allocateAuxiliaryStorage(inNeighborsByteSize);
-      myData.inNeighbors.init(range, inNeighbors.size());
-
-      for (auto const& inNeighbor : inNeighbors) {
-        myData.inNeighbors.emplace(inNeighbor);
-      }
-      inNeighbors.clear();
-    }
     return shouldPropagate;
   }
 
@@ -141,7 +150,7 @@ struct WCCComputation
       sendMessage(edge, message);
     }
 
-    for (auto pid : myData.inNeighbors) {
+    for (auto const& pid : myData.inNeighbors) {
       sendMessage(pid, message);
     }
   }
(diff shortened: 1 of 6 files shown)


arangod/Pregel/Algos/WCC.cpp
```diff
@@ -96,6 +102,16 @@ struct WCCComputation
...
     }
+
+    if (globalSuperstep() == 1) {
+      auto range = this->allocateAuxiliaryStorage(inNeighborsByteSize);
+      myData.inNeighbors.init(range, inNeighbors.size());
+
+      for (auto const& inNeighbor : inNeighbors) {
+        myData.inNeighbors.emplace(inNeighbor);
+      }
+      inNeighbors.clear();
```
jsteemann, line 113: What is the purpose of clearing `inNeighbors` here?
`clear()` doesn't free the underlying memory, and it is only done conditionally (superstep == 1). Plus, `inNeighbors` is a local variable, which would go out of scope at the end of the function (next line) anyway. I think there is no benefit of clearing the vector here, unless I am missing something. (that side, clearing here also does no harm).

arangod/Pregel/Algos/WCC.cpp
```diff
@@ -124,9 +140,9 @@ struct WCCComputation
 
       sendMessage(edge, message);
     }
-    // Also send to INBOUND neighbors
-    for (auto const& target : myData.inboundNeighbors) {
-      sendMessage(target, message);
+
+    for (auto pid : myData.inNeighbors) {
```
jsteemann, line 144: The value_type of `inNeighbors` is `PregelID`, which is a struct consisting of an std::string and a uint16_t. We should not copy the struct values while iterating over the container.
```suggestion
    for (auto const& pid : myData.inNeighbors) {
```

arangod/Pregel/CommonFormats.h
```diff
@@ -113,8 +114,88 @@ struct SCCValue {
...
+    uint8_t* current = nullptr;
+    uint32_t count = 0;
+
+    auto init(MemRange inStorage, uint32_t inCount) -> void {
+      storage = inStorage;
+      current = const_cast<uint8_t*>(storage.from);
+      count = inCount;
+    }
+
+    auto emplace(PregelID pid) -> void {
```
jsteemann, line 181: `PregelID` is copied here. I think it should rather be passed by const reference.

arangod/Pregel/CommonFormats.h
```diff
@@ -113,8 +114,88 @@ struct SCCValue {
...
+      auto storedInNeighbor = reinterpret_cast<InNeighbor*>(current);
+      storedInNeighbor->fromShard = pid.shard;
+      storedInNeighbor->keylen = pid.key.length();
+      memmove(storedInNeighbor->fromKey, pid.key.c_str(),
```
jsteemann, line 185: I think this should be a `memcpy` and not a `memmove`. Otherwise we will mess with the std::string's underlying data, which may leave the std::string in some strange state.

arangod/Pregel/CommonFormats.h
```diff
@@ -113,8 +114,88 @@ struct SCCValue {
...
+              storedInNeighbor->keylen);
+
+      current += storedInNeighbor->size();
```
jsteemann, line 188: We are increasing `current` here, but there is no check whether `current` points beyond the end of the memory region (in `init` we have an `inCount` value which could be used here to do bounds checking...)

arangod/Pregel/GraphStore.h
```diff
@@ -131,6 +163,22 @@ class GraphStore final {
...
 
+  //
+  // FIXME: This is a hack;
+  //
+  // For some algorithms, for example WCC, we need to store dynamic
+  // data; if we just allow heap allocations, we lose control over
+  // state being spilled to memory mapped files.
+  //
+  // This buffer is currently here to address this problem; the solution
+  // is not ideal, and will hopefulle be improved in future iterations
```
jsteemann, line 174:
```suggestion
  // is not ideal, and will hopefully be improved in future iterations
```

arangod/Pregel/VertexComputation.h
```diff
@@ -92,14 +92,20 @@ class VertexContext {
...
+  /*
+void setVertexData(V const& val) {
+  _graphStore->replaceVertexData(_vertexEntry, (void*)(&val), sizeof(V));
+}
+
+/// store data, will potentially move the data around
+void setVertexData(void const* ptr, size_t size) {
+  _graphStore->replaceVertexData(_vertexEntry, (void*)ptr, size);
+}
+*/
```
jsteemann, line 108: RIP
```suggestion
```

arangod/Pregel/GraphStore.h
```diff
@@ -90,8 +112,18 @@ class GraphStore final {
...
+    ADB_PROD_ASSERT(_config != nullptr);
+    return _config;
+  }
+  auto getFeature() const -> PregelFeature& { return _feature; }
+
   ReportManager* _reports;
 
+  auto allocateAuxiliaryStorage(size_t bytes) -> MemRange {
+    return _auxiliaryStorage.allocate(bytes);
+  };
```
jsteemann, line 125:
```suggestion
  }
```

arangod/Pregel/CommonFormats.h
```diff
@@ -113,8 +114,88 @@ struct SCCValue {
...
+    };
+
+    MemRange storage = MemRange{.from = nullptr, .to = nullptr};
+    // FIXME: output iterator?
+    uint8_t* current = nullptr;
+    uint32_t count = 0;
+
+    auto init(MemRange inStorage, uint32_t inCount) -> void {
+      storage = inStorage;
+      current = const_cast<uint8_t*>(storage.from);
```
jsteemann, line 177: AFAIK `storage.from` is a `uint8_t*` already and not a `uint8_t const*`, so the const_cast seems unnecessary.
```suggestion
      current = storage.from;
```

arangod/Pregel/Algos/WCC.cpp
markuspf: I moved the collection of `inNeighbors` to a separate function; as a bonus we save the dance of allocating a local variable on *every* computation to not use it most of the time.

arangod/Pregel/CommonFormats.h
```diff
@@ -113,8 +114,88 @@ struct SCCValue {
...
+      count = inCount;
```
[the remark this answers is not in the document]
markuspf: removed this variable because it turns out it is indeed not needed.

arangod/Pregel/CommonFormats.h
markuspf: added a n assertion that makes sure we don't write beyond the end of our allocation; if we do, we crash with `ADB_PROD_ASSERT`.

romanatarango (approved):

PR #16369, opened 2022-06-02
Fix initialization of _maxSubcompactions.

mpoeter:
### Scope & Purpose

We deliberately limit `_maxSubcompactions` by the number of threads for low priority operations ( `--rocksdb.num-threads-priority-low`). This number of threads is initialized to zero and if it is not specified, it is dynamically calculated.

However, the code to limit `_maxSubcompactions` to `_numThreadsLow` was executed _before_ that calculation was done, so unless the user explicitly set  `--rocksdb.num-threads-priority-low`, `_maxSubcompactions` was always set to zero.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] manually tested


jsteemann (approved):
LGTM!

PR #16370, opened 2022-06-02
Added autoincrement key generation for cluster with one shard and test

cpjulia:
### Scope & Purpose

This PR adds key generation of type "autoincrement" for cluster mode, but only if the collection is single shard. 
Obs.: To be edited
- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  No backports required.

#### Related Information

- [ ] Docs PR: to be added here
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: related to https://arangodb.atlassian.net/browse/ES-1092?atlOrigin=eyJpIjoiYjEzZTY2ZGExNzhhNDhhMmE3YjJmYTYzZDNhYTQ0YTQiLCJwIjoiaiJ9
- [ ] Design document:

### DIFF
--- a/tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
+++ b/tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
@@ -50,13 +50,13 @@ function getCoordinators() {
 
   const instanceInfo = JSON.parse(require('internal').env.INSTANCEINFO);
   return instanceInfo.arangods.filter(isCoordinator)
-                              .map(toEndpoint)
-                              .map(endpointToURL);
+    .map(toEndpoint)
+    .map(endpointToURL);
 }
 
 const servers = getCoordinators();
 
-function KeyGeneratorSuite () {
+function KeyGeneratorSuite() {
   'use strict';
   const cn = 'UnitTestsCollection';
   let coordinators = [];
@@ -75,7 +75,7 @@ function KeyGeneratorSuite () {
         envelope.body = body;
       }
       res = request(envelope);
-    } catch(err) {
+    } catch (err) {
       console.error(`Exception processing ${method} ${endpoint}`, err.stack);
       return {};
     }
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
     return res;
   }
 
+  function generateCollectionDocs(name) {
+    // check that the generated keys are sequential when we send the requests
+    // via multiple coordinators.
+    db._create(name, {keyOptions: {type: "autoincrement"}});
+    assertNotEqual("", db[name].properties().distributeShardsLike);
+    let lastKey = null;
+    let url = "/_db/" + cn + "/_api/document/" + name;
+    // send documents to both coordinators
+    let i = 0;
+    let startingKey = null;
+    if (Number(name[name.length - 1]) === 1) {
+      startingKey = "abc123";
+    } else if (Number(name[name.length - 1]) === 2) {
+      startingKey = "150000";
+    }
+    for (i; i < 100000; ++i) {
+      let doc = {};
+      if (startingKey === null || i > 0) {
+        doc = {"x": 0, "c": 1};
+      } else {
+        doc = {"_key": `${startingKey}`, "x": 0, "c": 1};
+      }
+      let result = sendRequest('POST', url, /*payload*/ {"x": 0, "c": 1}, {}, i % 2 === 0);
+      assertEqual(result.status, 202);
+      let key = result.body._key;
+      assertTrue(Number(key) === Number(lastKey) + 1 || lastKey === null, {key, lastKey});
+      lastKey = key;
+    }
+    let res = db._query(`FOR doc IN ${name} FILTER doc.x == 0 RETURN doc`).toArray();
+    assertTrue(res.length === 100000);
+    lastKey = null;
+    for (i = 0; i < res.length; ++i) {
+      let key = res[i].key;
+      assertTrue(Number(key) === Number(lastKey) + 1 || lastKey === null, {key, lastKey});
+      assertTrue(res[i].hasOwnProperty("x"));
+      assertTrue(res[i].x === 0);
+      assertTrue(res[i].hasOwnProperty("c"));
+      assertTrue(res[i].c === 1);
+      lastKey = key;
+    }
+  }
+
   return {
-    setUpAll: function() {
+    setUpAll: function () {
       coordinators = getCoordinators();
       if (coordinators.length < 2) {
         throw new Error('Expecting at least two coordinators');
       }
     },
 
-    testPadded: function() {
+    testPadded: function () {
       // check that the generated keys are sequential when we send the requests
       // via multiple coordinators.
-      db._create(cn, { numberOfShards: 1, keyOptions: { type: "padded" } });
+      db._create(cn, {numberOfShards: 1, keyOptions: {type: "padded"}});
 
       try {
         let lastKey = null;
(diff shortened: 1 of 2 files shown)


tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -140,15 +182,40 @@ function KeyGeneratorSuite () {
...
     },
-    
+
+    testAutoincrementOnMultipleShards: function () {
+      // check that the generated keys are sequential when we send the requests
+      // via multiple coordinators.
+      try {
+        db._create(cn, {numberOfShards: 5, keyOptions: {type: "autoincrement"}});
+      } catch (error) {
+        assertTrue(error.message === "the specified key generator is not supported for collections with more than one shard");
```
jsteemann, line 200: Instead of checking for a specific error message, we should rather check for a specific error code. This is because error message tests are frequently changed, but error codes do not change often, if at all.
```suggestion
        assertEqual(error.message, "the specified key generator is not supported for collections with more than one shard");
```

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -140,15 +182,40 @@ function KeyGeneratorSuite () {
...
+      } finally {
+        db._drop(cn);
+      }
+    },
+
+    testAutoincrementOnOneShard: function () {
+      db._createDatabase(cn, {sharding: "single"});
```
jsteemann, line 207: OneShard is only available in the cluster && enterprise edition, so the test needs to be restricted to that combination.

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
...
+    } else if (Number(name[name.length - 1]) === 2) {
+      startingKey = "150000";
+    }
+    for (i; i < 100000; ++i) {
+      let doc = {};
+      if (startingKey === null || i > 0) {
+        doc = {"x": 0, "c": 1};
+      } else {
+        doc = {"_key": `${startingKey}`, "x": 0, "c": 1};
+      }
```
jsteemann, line 114: What is the purpose of `doc` in this test? It looks unused.

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
...
+    let url = "/_db/" + cn + "/_api/document/" + name;
+    // send documents to both coordinators
+    let i = 0;
+    let startingKey = null;
+    if (Number(name[name.length - 1]) === 1) {
+      startingKey = "abc123";
+    } else if (Number(name[name.length - 1]) === 2) {
+      startingKey = "150000";
+    }
+    for (i; i < 100000; ++i) {
```
jsteemann, line 108:
```suggestion
    for (let i; i < 100000; ++i) {
```

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
...
+    lastKey = null;
+    for (i = 0; i < res.length; ++i) {
+      let key = res[i].key;
+      assertTrue(Number(key) === Number(lastKey) + 1 || lastKey === null, {key, lastKey});
+      assertTrue(res[i].hasOwnProperty("x"));
+      assertTrue(res[i].x === 0);
+      assertTrue(res[i].hasOwnProperty("c"));
+      assertTrue(res[i].c === 1);
+      lastKey = key;
+    }
```
jsteemann, line 132: First of all, the query is unsafe, because it does not guarantee the order of the results without a SORT.
And then I think we don't need this query for anything, as we are already validating the generated keys directly on insert.

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
...
+      startingKey = "150000";
+    }
+    for (i; i < 100000; ++i) {
+      let doc = {};
+      if (startingKey === null || i > 0) {
+        doc = {"x": 0, "c": 1};
+      } else {
+        doc = {"_key": `${startingKey}`, "x": 0, "c": 1};
+      }
+      let result = sendRequest('POST', url, /*payload*/ {"x": 0, "c": 1}, {}, i % 2 === 0);
```
jsteemann, line 115: `x` and `c` are not relevant for this test.
```suggestion
      let result = sendRequest('POST', url, /*payload*/ {}, {}, i % 2 === 0);
```

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -90,18 +90,60 @@ function KeyGeneratorSuite () {
...
+    // via multiple coordinators.
+    db._create(name, {keyOptions: {type: "autoincrement"}});
+    assertNotEqual("", db[name].properties().distributeShardsLike);
+    let lastKey = null;
+    let url = "/_db/" + cn + "/_api/document/" + name;
+    // send documents to both coordinators
+    let i = 0;
+    let startingKey = null;
+    if (Number(name[name.length - 1]) === 1) {
+      startingKey = "abc123";
```
jsteemann, line 104: The purpose of a key generator is that it generates the keys itself.
Using a key such as `abc123` somewhat defeats that purpose.

What is missing here in this test is the usage of an autoincrement key generator with different `offset` and/or `increment` values. These tests should be added.

cpjulia: Will address

tests/js/client/load-balancing/load-balancing-key-generator-noauth-cluster.js
```diff
@@ -140,15 +172,39 @@ function KeyGeneratorSuite () {
...
         db._dropDatabase(cn);
       }
     },
-    
+
+    testAutoincrementOnMultipleShards: function() {
+      try {
+        db._create(cn, {numberOfShards: 5, keyOptions: {type: "autoincrement"}});
+      } catch (error) {
+        assertEqual(error.code, 501);
```
jsteemann, line 180: I would be in favor of using symbolic error names here, e.g.
```
assertEqual(error.errorNum, errors.ERROR_CLUSTER_UNSPUPPORTED.code);
```
(not tested, but something along these lines - additionally, `errors` needs to be defined first).

tests/js/client/shell/shell-keygen.js
```diff
@@ -873,10 +897,12 @@ function PersistedLastValueSuite () {
           assertEqual(0, lastValue);
 
           for (let i = 0; i < 10; ++i) {
-            c.insert({}); 
+            c.insert({});
             p = c.properties();
             let newLastValue = p.keyOptions.lastValue;
-            assertTrue(newLastValue > lastValue);
+            if (Number(numShards) === 2) {
```
jsteemann, line 903: Why this `if` condition` here? Shouldn't the assertion in the next line always hold true, for both `padded` and `autoincrement`?

jsteemann (approved):
LGTM!

PR #16375, opened 2022-06-03
Fix compilation for GCC-12 and make some refactoring

MBkkt:
### Scope & Purpose

1. Update function2 to [f569a63cfe369df867a1a4d17aaa12269156536c](https://github.com/Naios/function2/commit/f569a63cfe369df867a1a4d17aaa12269156536c)
For noexcept and some other fixes

2. Add absl::strings for working StrCat
 
3. Fix gcc-12 compilation

4. Fix missing includes on gcc-12

5. Fix initialization order warning on gcc-12

6. Make CallbackGuard better
Make noexcept check in compile time and use move only function
 
7. Make RebootTracker better
https://github.com/arangodb/arangodb/pull/16328#discussion_r888285559

8. Make RestHandler better
Don't need to use `Future<T>` here, enough `Future<Unit>`

9. Make UpgradeTasks better
Minor improvements

10. Fix RebootTracker usage

11. Fix RestHandler usage

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/RebootTracker.cpp
+++ b/arangod/Cluster/RebootTracker.cpp
@@ -144,7 +144,7 @@ CallbackGuard RebootTracker::callMeOnChange(std::string_view serverId,
   auto& rebootsMap = _callbacks[serverId];
   // For the given rebootId, get the existing callbacks map, or create a new one
   auto& callbacksMap = rebootsMap[rebootId];
-  auto const callbackId = ++_callbackId;
+  auto const callbackId = _nextCallbackId++;
   auto [_, inserted] = callbacksMap.try_emplace(
       callbackId,
       DescriptedCallback{std::move(callback), std::string{description}});
--- a/arangod/Cluster/RebootTracker.cpp
+++ b/arangod/Cluster/RebootTracker.cpp
@@ -22,7 +22,6 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "Basics/Exceptions.h"
-#include "Basics/StaticStrings.h"
 #include "Cluster/ClusterTypes.h"
 #include "Cluster/RebootTracker.h"
 #include "Logger/LogMacros.h"
@@ -30,9 +29,9 @@
 #include "Scheduler/Scheduler.h"
 #include "Scheduler/SchedulerFeature.h"
 
-#include <velocypack/Slice.h>
 #include <absl/strings/str_cat.h>
 #include <algorithm>
+#include <vector>
 
 namespace arangodb::cluster {
 namespace {
--- a/arangod/Cluster/RebootTracker.cpp
+++ b/arangod/Cluster/RebootTracker.cpp
@@ -21,25 +21,43 @@
 /// @author Tobias Gödderz
 ////////////////////////////////////////////////////////////////////////////////
 
-#include "RebootTracker.h"
-
 #include "Basics/Exceptions.h"
-#include "Basics/MutexLocker.h"
-#include "Basics/ScopeGuard.h"
 #include "Basics/StaticStrings.h"
 #include "Cluster/ClusterTypes.h"
+#include "Cluster/RebootTracker.h"
 #include "Logger/LogMacros.h"
 #include "Logger/Logger.h"
 #include "Scheduler/Scheduler.h"
 #include "Scheduler/SchedulerFeature.h"
 
-#include <velocypack/Builder.h>
 #include <velocypack/Slice.h>
-
+#include <absl/strings/str_cat.h>
 #include <algorithm>
 
-using namespace arangodb;
-using namespace arangodb::cluster;
+namespace arangodb::cluster {
+namespace {
+
+void safeInvoke(RebootTracker::DescriptedCallback& callback) noexcept {
+  try {
+    LOG_TOPIC("afdfd", TRACE, Logger::CLUSTER)
+        << "Executing callback " << callback.description;
+    callback.callback();
+  } catch (arangodb::basics::Exception const& ex) {
+    LOG_TOPIC("88a63", INFO, Logger::CLUSTER)
+        << "Failed to execute reboot callback: " << callback.description << ": "
+        << "[" << ex.code() << "] " << ex.what();
+  } catch (std::exception const& ex) {
+    LOG_TOPIC("3d935", INFO, Logger::CLUSTER)
+        << "Failed to execute reboot callback: " << callback.description << ": "
+        << ex.what();
+  } catch (...) {
+    LOG_TOPIC("f7427", INFO, Logger::CLUSTER)
+        << "Failed to execute reboot callback: " << callback.description << ": "
+        << "Unknown error.";
+  }
+}
+
+}  // namespace
 
 static_assert(
     std::is_same_v<decltype(SchedulerFeature::SCHEDULER),
@@ -55,7 +73,7 @@ static_assert(
     "SchedulerPointer is expected to point to an instance of Scheduler");
 
 RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
-    : _scheduler(scheduler) {
+    : _callbackId{0}, _scheduler{scheduler} {
   // All the mocked application servers in the catch tests that use the
   // ClusterFeature, which at some point instantiates this, do not start the
   // SchedulerFeature. Thus this dies. However, we will be able to fix that at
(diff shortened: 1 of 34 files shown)


arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-    if (rbIt != rebootMap.end()) {
-      auto& callbackSetPtr = rbIt->second;
-      TRI_ASSERT(callbackSetPtr != nullptr);
-      callbackSetPtr->erase(callbackId);
-      if (callbackSetPtr->empty()) {
-        rebootMap.erase(rbIt);
+void RebootTracker::unregisterCallback(std::string_view serverId,
+                                       RebootId rebootId,
+                                       CallbackId callbackId) noexcept {
+  std::lock_guard guard{_mutex};
```
MBkkt, line 202: There could theoretically be an exception, I ignore this.
As an alternative, I can use `absl::Mutex`, but I don't think it's necessary, because I think that in a properly loaded program in the correct code, `std::mutex` is also don't throw.
`noexcept` is needed here because the code is called from the dtor

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-
+      << "updating reboot server state from " << _state << " to " << state;
+  for (auto it = _callbacks.begin(); it != _callbacks.end();) {
+    auto next = std::next(it);
+    auto newIt = state.find(it->first);
+    TRI_ASSERT(!it->second.empty());  // we should remove it, when it was empty
     if (newIt == state.end()) {
-      // Try to schedule all callbacks for serverId.
-      // If that didn't throw, erase the entry.
-      LOG_TOPIC("88858", INFO, Logger::CLUSTER)
```
MBkkt, line 97: Maybe instead of this and "88857" messages I can write something like this
```C++
LOG_TOPIC(...) << "rebooted servers: " << std::set_difference(old_state.begin(), old_state.end(), new_state.begin(), new_state.end(), [](auto const& lhs,auto const& rhs) {
  return lhs == rhs;
});
```
If it needed of course.

I also have a feeling that we need log something like this in cluster info, not in reboot tracker

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-  auto const currentRebootId = rebootIdIt->second;
-
-  if (peerState.rebootId() < currentRebootId) {
+  auto const currRebootId = it->second;
+  if (rebootId < currRebootId) {
     // If this ID is already older, schedule the callback immediately.
-    queueCallback(DescriptedCallback{std::move(callback),
-                                     std::move(callbackDescription)});
-    return CallbackGuard{nullptr};
+    queueCallback({std::move(callback), std::string{description}});
```
MBkkt, line 126: Maybe better to create DescriptedCallback in queueCallback

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
+  TRI_ASSERT(_state.contains(it->first));
+  _scheduler->queue(RequestLane::CLUSTER_INTERNAL,
+                    [reboots = std::move(it->second)]() mutable noexcept {
+                      for (auto& callbacksIds : reboots) {
+                        for (auto& calls : callbacksIds.second) {
+                          safeInvoke(calls.second);
+                        }
+                      }
+                    });
+  _callbacks.erase(it);
```
MBkkt, line 157: `_callbacks` is flat_hash_map, so I use move + erase instead of extract

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-                          [](auto it) { return it == nullptr; }));
-  TRI_ASSERT(std::none_of(callbacks.cbegin(), callbacks.cend(),
-                          [](auto it) { return it->empty(); }));
-
-  auto cb = createSchedulerCallback(std::move(callbacks));
-  _scheduler->queue(RequestLane::CLUSTER_INTERNAL, std::move(cb));
+  std::vector<RebootIds::node_type> callbacks;
+  for (auto rbIt = reboots.begin(); rbIt->first < to;) {
+    TRI_ASSERT(rbIt != reboots.end());
+    callbacks.push_back(reboots.extract(rbIt++));
```
MBkkt, line 179: reboots are std::map, so I use extract instead of move + erase

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
+  }
+  TRI_ASSERT(!reboots.empty());
+  _scheduler->queue(RequestLane::CLUSTER_INTERNAL,
+                    [callbacks = std::move(callbacks)]() mutable noexcept {
+                      for (auto& calls : callbacks) {
+                        for (auto& e : calls.mapped()) {
+                          safeInvoke(e.second);
+                        }
+                      }
+                    });
```
MBkkt, line 189: I think I should unify it with another `queueCallbacks`

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
 #endif
 }
 
-void RebootTracker::updateServerState(
-    containers::FlatHashMap<ServerID, RebootId> const& state) {
-  MUTEX_LOCKER(guard, _mutex);
-
+void RebootTracker::updateServerState(State state) {
```
MBkkt, line 87: Maybe add rvalue reference?

arangod/Cluster/RebootTracker.cpp
```diff
@@ -21,25 +21,43 @@
...
-using namespace arangodb::cluster;
+namespace arangodb::cluster {
+namespace {
+
+void safeInvoke(RebootTracker::DescriptedCallback& callback) noexcept {
+  try {
+    LOG_TOPIC("afdfd", TRACE, Logger::CLUSTER)
+        << "Executing callback " << callback.description;
+    callback.callback();
+  } catch (arangodb::basics::Exception const& ex) {
```
MBkkt, line 45:
```suggestion
  } catch (basics::Exception const& ex) {
```

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
+                                            RebootId rebootId,
+                                            Callback callback,
+                                            std::string_view description) {
+  std::lock_guard guard{_mutex};
+  auto const it = _state.find(serverId);
+  // We MUST NOT insert something in _callbacks[serverId]
+  // unless _state[serverId] exists!
+  if (it == _state.end()) {
+    auto const error =
+        absl::StrCat("When trying to register callback '", description,
```
MBkkt, line 116: I use absl instead of basics because I had some compile time issues with basics

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-
-  auto const [iterator, inserted] = callbackMap.try_emplace(
+  auto& rebootsMap = _callbacks[serverId];
+  // For the given rebootId, get the existing callbacks map, or create a new one
+  auto& callbacksMap = rebootsMap[rebootId];
+  auto const callbackId = ++_callbackId;
+  auto [_, inserted] = callbacksMap.try_emplace(
       callbackId,
-      DescriptedCallback{std::move(callback), std::move(callbackDescription)});
+      DescriptedCallback{std::move(callback), std::string{description}});
```
MBkkt, line 138: Here I constructed DescriptedCallback, because I had compile time issue on apple clang

goedderz (commented):
Placeholder for the original review here: https://github.com/arangodb/arangodb/pull/16328#pullrequestreview-992116432

arangod/Cluster/RebootTracker.h
```diff
@@ -23,119 +23,83 @@
...
 // scheduler is destroyed.
 class RebootTracker {
  public:
-  using Callback = std::function<void(void)>;
   using SchedulerPointer = SupervisedScheduler*;
-
-  class PeerState {
-   public:
-    PeerState(ServerID serverId, RebootId rebootId)
-        : _serverId{std::move(serverId)}, _rebootId{rebootId} {}
```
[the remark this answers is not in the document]
goedderz: > Please can you show example where we use this struct and that give more readable code

It's about semantics, not convenience of use. To explain what I mean by that:

This is about the "user"-facing API, i.e. `callMeOnChange`. This binds a callback to a `PeerState`, and if that state changes, the callback is executed. In a perfect world, this would be an opaque value from the beginning: e.g., a REST request would allow you to read the `PeerState` of the client (think for example of an Coordinator, setting up AQL snippets on DBServers). You can hand this over to the RebootTracker, but do nothing else with it - possibly not even (easily) construct a `PeerState`. That would prevent pitfalls, like someone reading the RebootId of the client from the AgencyCache instead of getting it from the client itself.

Now nothing of that is currently there, but the least we can do is emphasize that these values belong together, and form the smallest unit that makes sense: Neither the ServerID has any meaning on its own, nor the RebootID.

arangod/Cluster/RebootTracker.h
```diff
@@ -23,119 +23,83 @@
...
-  CallbackId _nextCallbackId{1};
+  /// @brief Save a pointer to the scheduler for easier testing
+  SchedulerPointer _scheduler;
 
   /// @brief Last known rebootId of every server.
   /// Will regularly get updates from the agency.
   /// Updates may not be applied if scheduling the affected callbacks fails, so
   /// the scheduling will be tried again on the next update.
-  std::unordered_map<ServerID, RebootId> _rebootIds;
+  State _state;
```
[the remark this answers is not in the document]
goedderz: Could you elaborate on the last comment, i.e.

> >  I have another rebootsIds so its not convenient
>
> I don't understand, what do you mean?

?

arangod/Cluster/RebootTracker.h
```diff
@@ -23,119 +23,83 @@
...
-          std::shared_ptr<std::unordered_map<CallbackId, DescriptedCallback>>>
-          callbacks);
-  void queueCallback(DescriptedCallback callback);
+  mutable std::mutex _mutex;
 
- private:
-  mutable Mutex _mutex;
+  CallbackId _callbackId;
 
-  CallbackId _nextCallbackId{1};
```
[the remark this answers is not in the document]
MBkkt: I think we have 2 points:
1. call ctor in cpp or header
I don't agree with you (we have compiler for check this), but It's not important for me, so I can revert it
2. I wrote `_callbackId`, `{0}`, `operator++()` instead of `_nextCallbackId`, `{1}`, `operator++(int)`
I made it because I thought it a little simpler, but I also can revert it, if you want

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +84,135 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-  for (auto const& newIt : state) {
-    auto const& serverId = newIt.first;
-    auto const& rebootId = newIt.second;
-    auto const rv = _rebootIds.try_emplace(serverId, rebootId);
-    auto const inserted = rv.second;
-    // If we inserted a new server, we may NOT already have any callbacks for
-    // it!
-    TRI_ASSERT(!inserted || _callbacks.find(serverId) == _callbacks.end());
+    it = next;
   }
```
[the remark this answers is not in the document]
MBkkt: >  CallbackGuard-move constructor being noexcept, so a static_assert 
It's not necessary, check code, I don't call CallbackGuard move-ctor or assign in this function

arangod/Cluster/RebootTracker.cpp
goedderz: So regarding the log messages. I'd like to see the following:

* On INFO: When another server changes (reboots or is removed), a message when that is noticed, and which server specifically this was about. This was previously handled by `88857` and `88858`.
  This is often useful when analyzing log files.
* On DEBUG: When callbacks are executed, and which ones. This is handled by `afdfd` as before, but is now set to TRACE. I think it'd be better on DEBUG, in particular with the next point:
  This is useful when debugging something particular, e.g. why some cleanup doesn't happen or why some server is not releasing its locks or so.
* On TRACE: Changes to state as logged by `77a6e`. This was previously DEBUG (not changed by this PR), but I think it's the least useful to users, but the most noisy one, so should be changed to TRACE.

WDYT?

arangod/Cluster/RebootTracker.h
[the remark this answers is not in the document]
MBkkt: I mean I have `using RebootIds`, that used for other map, so for me if `state` will be named `_rebootIds` it will be a little confusing

arangod/Cluster/RebootTracker.cpp
MBkkt: So I see 3 questions:
1. Do we really need this log messages, or maybe only rebooted servers that had callbacks is enough?
2. What and why should be log level for log old and new state: info, debug, trace?
3. Is difference between iteration for `_callbacks` or `_state` important? 
    `_callbacks` size is `O(1)`, it's leaders and some other important servers.
    `_state` size is count of servers in cluster.
4. if 1 and 3 yes, should we think about log level higher than info? Or readability and simplicity is preferable?

MBkkt: Done?

MBkkt:
https://jenkins.arangodb.biz/job/arangodb-matrix-pr/21923/

arangod/Cluster/RebootTracker.cpp
goedderz: > Do we really need this log messages, or maybe only rebooted servers that had callbacks is enough?

Maybe. What I can say is: These log messages (88857 and 88858) have been extremely helpful in the past to debug different problems, both in support and development.

If it's always there, it helps to align (distributed) logs. That's not possible when the message isn't there.

Another drawback with logging only those will callbacks is that an expected log message doesn't appear, and you don't know whether it's because the server didn't notice the reboot, or just didn't have any callbacks.

How often that makes an impact I cannot say.

So my problem here is: I know that the logs are helpful the way they are. The will surely lose some of that, and maybe a lot: I don't know. Either way, I don't think it's worth losing them, because there's no visible improvement on the other side.

> What and why should be log level for log old and new state: info, debug, trace?

I don't get the question, I thought I've written that in the previous post? Can you be more specific?

> Is difference between iteration for _callbacks or _state important?
> _callbacks size is O(1), it's leaders and some other important servers.
> _state size is count of servers in cluster.

Regarding __callbacks size is O(1)_: I don't see how you arrive at that conclusion, unless you mean O(1) because it'll at most be 30 or so. Perhaps clarify what you mean by "leader"? That term's pretty overloaded. Note that if there are a few AQL queries running, that alone will usually add all the Coordinators on all DBServers; and all DBServers on all Coordinators.

Anyway, to answer the question: There will be a difference in size. But the amount of servers is so small that it does not matter. Also note that this is usually executed at most once a second by the heartbeat thread; and its complexity dwarfs in comparison to all the other stuff done by `ClusterInfo::flush()` (where this is called).

> if 1 and 3 yes, should we think about log level higher than info? Or readability and simplicity is preferable here?

When you say higher, do you mean WARN etc.? I can't think of anything that should be a warning here. And I don't understand the second question in this context?

MBkkt: > When you say higher, do you mean WARN etc.? I can't think of anything that should be a warning here. And I don't understand the second question in this context?

This question was about users that have more higher log level than INFO, so them don't have INFO in logs
But it's outdated

MBkkt: > I don't get the question, I thought I've written that in the previous post? Can you be more specific?

Yes, you already answered, I didn't see it, I also wrote about that

arangod/Cluster/RebootTracker.cpp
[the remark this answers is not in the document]
goedderz: Nice idea using the node_type now.

arangod/Cluster/RebootTracker.h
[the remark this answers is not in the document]
goedderz: 1. ✓

2. Yes, the current code is slightly simpler. The reasoning for the previous implementation was a different one, though: Namely that `_nextCallbackId` immediately conveys its behavior, without looking at any code: That it's the next ID of something, and that it's not yet used itself. While `_callbackId` conveys nothing but its type. Could be the ID of some special callback. Could also be the next ID. Or something else. And as I didn't have an equally expressive name for the current variant, I chose to go with one additional LOC in favor of having a better name.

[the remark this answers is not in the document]
MBkkt: 2. https://github.com/arangodb/arangodb/pull/16375/commits/3d7839382e569cea9bc4be5bd65c5faf322994b0
Done?

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +95,117 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-      strstream << "When trying to register callback '" << callbackDescription
-                << "': "
-                << "The server " << peerState.serverId() << " is not known. "
-                << "If this server joined the cluster in the last seconds, "
-                   "this can happen.";
-      return strstream.str();
-    }();
+CallbackGuard RebootTracker::callMeOnChange(RebootTracker::PeerState peer,
+                                            Callback callback,
+                                            std::string_view description) {
```
goedderz, line 120: I don't care much about whether the description is passed as a `string_view` or a `string`, but I have to ask: Why did you change that? The way I see it, we now always create a copy (when the `DescriptedCallback` is constructed). Previously, the caller had the option to move, which could optionally avoid it.

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +95,116 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-  TRI_ASSERT(std::none_of(callbacks.cbegin(), callbacks.cend(),
-                          [](auto it) { return it->empty(); }));
-
-  auto cb = createSchedulerCallback(std::move(callbacks));
-  _scheduler->queue(RequestLane::CLUSTER_INTERNAL, std::move(cb));
+  std::vector<RebootIds::node_type> callbacks;
+  for (auto rbIt = reboots.begin(); rbIt->first < to;) {
+    TRI_ASSERT(rbIt != reboots.end());
+    callbacks.push_back(reboots.extract(rbIt++));
+  }
```
goedderz, line 180: Nitpicking, but wouldn't
```suggestion
  for (auto rbIt = reboots.begin(); rbIt->first < to; ++rbIt) {
    TRI_ASSERT(rbIt != reboots.end());
    callbacks.push_back(reboots.extract(rbIt));
  }
```
be more idiomatic?

arangod/Cluster/RebootTracker.cpp
MBkkt: Ok, https://github.com/arangodb/arangodb/pull/16375/commits/5107c2495127e272174740e3cba5383327ad6072

> we now always create a copy

No, if this Peer outdated then not

> Previously, the caller had the option to move

Yes, but we never use this opportunity 


I think we have a problem in that ServerId is string (not in this code, but in general):
We almost always create copy of it: from vpacks although we store that vpacks, to other functions, etc 

Btw maybe you know why `ServerID` is `std::string`, why isn't `uint64_t`?

arangod/Cluster/RebootTracker.cpp
MBkkt: Iterator will be invalid after extract

goedderz: Oh, true.

arangod/Cluster/RebootTracker.cpp
goedderz: > No, if this Peer outdated then not

That case virtually never happens and is definitely not the case to optimize for.

> Yes, but we never use this opportunity

And now we won't in the future, either. :) Also, it's not true: In half the cases where this is called, there's a string created anyway, which is moved here. In those cases, we now make an additional copy. In the other cases, the copy is now made here instead of at the caller's site.

> I think we have a problem in that ServerId is string (not in this code, but in general):
> We almost always create copy of it: from vpacks although we store that vpacks, to other functions, etc
>
>Btw maybe you know why ServerID is std::string, why isn't uint64_t?

I don't know how that evolved historically. I'd also like to have a `uint64_t` or something else of fixed length (could even be a string), at least something we can pass around without allocating. Could probably be done by having a map somewhere, that can be consulted during (de)serialization. As a simple idea, maybe the UUIDs could be converted to an 128bit int, though that would forget the server type prefix (i.e. `PRMR-`, `COOR-`, `AGNT-`); I don't know the impact of that without more research. And server short names can't be forgotten either. So I guess that it could be done.

MBkkt: Ok, anyway done?

MBkkt: Maybe you didn't see, but I reverted to string

MBkkt: string_view better for description because we pass string literal, so in that case we made copy always if and when needed

goedderz: But we don't always pass a string literal:
```cpp
        engine->rebootTrackers().emplace_back(ci.rebootTracker().callMeOnChange(
            cluster::RebootTracker::PeerState(server, rebootId), std::move(f),
            std::move(comment)));
```
(from ExecutionEngine.cpp)

```cpp
      auto rGuard =
          std::make_unique<RebootCookie>(ci.rebootTracker().callMeOnChange(
              RebootTracker::PeerState(serverId, rebootId), std::move(f),
              std::move(comment)));
```
(from RestReplicationHandler.cpp)

Both are dynamically calculated `std::string`s. These now always copy.

And for the calls with literals, it does not make a difference: Either the copy is made at the caller's site (when passed as a std::string), or when constructing the DescriptedCallback (when passed as a std::string_view). But a copy is made either way.

So as far as I can see, this is a pessimization.

goedderz: > But if we pass always const ref string, or literal string_view always better than other possibilities, them not the same

Yes, agreed.

arangod/Cluster/RebootTracker.h
```diff
@@ -23,119 +23,86 @@
...
-
+  using Callback = fu2::unique_function<void()>;
   struct DescriptedCallback {
     Callback callback;
     std::string description;
   };
+  using State = containers::FlatHashMap<ServerID, RebootId>;
+  using CallbackId = uint64_t;
+  using RebootIds =
+      std::map<RebootId,
```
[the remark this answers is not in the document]
MBkkt: In theory yes, for example:
current reboot Id = 1
callMeOnChange(2)
callMeOnChange(10)
update current reboot Id to 3

In practice I think no, and map commonly have a size 1~2

arangod/Cluster/RebootTracker.h
```diff
@@ -23,119 +23,86 @@
...
 
+  /// @brief Save a pointer to the scheduler for easier testing
+  SchedulerPointer _scheduler;
+
   /// @brief Last known rebootId of every server.
   /// Will regularly get updates from the agency.
   /// Updates may not be applied if scheduling the affected callbacks fails, so
   /// the scheduling will be tried again on the next update.
-  std::unordered_map<ServerID, RebootId> _rebootIds;
+  State _state;
```
jsteemann, line 96: Stylistic: I somewhat dislike the name `_state` here, but would prefer to keep the old name `_rebootIds`.
RebootIds was quite a telling variable name, but `State` is not very telling.

MBkkt: https://github.com/arangodb/arangodb/pull/16375#discussion_r891388779

arangod/Cluster/RebootTracker.cpp
```diff
@@ -66,296 +95,116 @@ RebootTracker::RebootTracker(RebootTracker::SchedulerPointer scheduler)
...
-                          [](auto it) { return it == nullptr; }));
-  TRI_ASSERT(std::none_of(callbacks.cbegin(), callbacks.cend(),
-                          [](auto it) { return it->empty(); }));
-
-  auto cb = createSchedulerCallback(std::move(callbacks));
-  _scheduler->queue(RequestLane::CLUSTER_INTERNAL, std::move(cb));
+  std::vector<RebootIds::node_type> callbacks;
+  for (auto rbIt = reboots.begin(); rbIt->first < to;) {
+    TRI_ASSERT(rbIt != reboots.end());
+    callbacks.push_back(reboots.extract(rbIt++));
```
jsteemann, line 179: What happens if `rbIt++` sets the iterator to the end of the container? Is dereferencing `rbIt->first` in the for loop condition then actually valid?

MBkkt: Fixed

arangod/Cluster/RebootTracker.cpp
```diff
@@ -21,25 +21,54 @@
...
+
+void safeInvoke(RebootTracker::DescriptedCallback& callback) noexcept {
+  try {
+    LOG_TOPIC("afdfd", DEBUG, Logger::CLUSTER)
+        << "Executing callback " << callback.description;
+    callback.callback();
+  } catch (basics::Exception const& ex) {
+    LOG_TOPIC("88a63", INFO, Logger::CLUSTER)
+        << "Failed to execute reboot callback: " << callback.description << ": "
+        << "[" << ex.code() << "] " << ex.what();
```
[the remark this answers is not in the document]
MBkkt: Fixed

MBkkt:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/21969/

PR #16381, opened 2022-06-03
Documentation/open-http-interface-earlier

Simran-B:
Update `GET /_admin/status` DocuBlock

Reference: https://github.com/arangodb/docs/pull/994

### DIFF
--- a/Documentation/DocuBlocks/Rest/Administration/JSF_get_admin_status.md
+++ b/Documentation/DocuBlocks/Rest/Administration/JSF_get_admin_status.md
@@ -47,6 +47,20 @@ Whether writes are enabled.
 @RESTSTRUCT{maintenance,get_admin_status_server_info,boolean,required,}
 Whether the maintenance mode is enabled.
 
+@RESTSTRUCT{phase,get_admin_status_server_info,string,required,}
+Name of the lifecycle phase the instance is currently in. Normally one of
+`"in prepare"`, `"in start"`, `"in wait"`, `"in shutdown"`, `"in stop"`,
+or `"in unprepare"`.
+
+@RESTSTRUCT{feature,get_admin_status_server_info,string,required,}
+Internal name of the feature that is currently being prepared, started,
+stopped or unprepared.
+
+@RESTSTRUCT{recoveryTick,get_admin_status_server_info,number,required,}
+Current recovery sequence number value, if the instance is currently recovering.
+If the instance is already past the recovery, this attribute will contain the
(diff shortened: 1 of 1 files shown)


Simran-B:
- still true? "intended for manual use by the support and should never be used for monitoring or automatic tests"
- what's the hostname? Is it encoded? Why is there a reference to `ServerState` (non-existing) in the old text?
- what is a reboot ID? Do users need to know? Is it cluster only? Agents, DB-Servers, Coordinators?
- `state` always `SERVING`, what is it good for?
- `agency` cluster only? Agents, DB-Servers, Coordinators?

Simran-B:
There is actually a field `host` and a field `hostname`. The latter is only present if an environment variable `HOSTNAME` was set. The former is more complicated:
1. Use `HOST` environment variable
2. Otherwise use `NODE_NAME` environment variable
3. Otherwise use content of file `/etc/machine-id` (which contains the string that I saw)
4. On macOS, call `gethostuuid()` and hex-encode it
5. As a last resort, use the `--cluster.my-address` or `--agency.my-address` value

Regarding the `state` field, I suppose the `progress` field can be checked for as long as the `state` is `STARTUP` (once startup is completed, the progress values no longer change).

Documentation/DocuBlocks/Rest/Administration/get_admin_status.md
```diff
@@ -0,0 +1,145 @@
...
+@RESTSTRUCT{progress,get_admin_status_server_info,object,required,get_admin_status_server_info_progress}
+Startup and recovery information.
+
+You can check for changes to determine whether progress was made between two
+calls, but you should not rely on specific values as they may change between
+ArangoDB versions. The values are only expected to change during the startup and
+shutdown, i.e. while `maintenance` is `true`.
+
+You need to start _arangod_ with the `--server.early-connections` startup option
+enabled to be able to query the endpoint during the startup phase.
```
Simran-B, line 55: @jsteemann Maybe this is better?
```suggestion
enabled to be able to query the endpoint during the entire startup process.
```

jsteemann: This is not correct, because the endpoint is not open during the **entire startup process*, but only after the HTTP REST endpoint is opened. I don't want to nurture the expectation that the HTTP API is always responsive during the entire startup. Which it is not. It is only responsive after the HTTP API is opened, which is during the startup, but only at a certain point.

PR #16389, opened 2022-06-07
add getByPrimaryKey workload

jsteemann:
### Scope & Purpose

Performance-test only...

Adds a new workload to fetch data by primary key, either document keys only, or the full documents in addition.
Key values are supposed to be contiguous numeric values, optionally starting with a prefix.

Example config:
```
{
  "rocksdb": {},
  "workload": {
    "getByPrimaryKey": {
      "default": {
        "collection": "peter",
        "fillBlockCache": true,
        "fetchFullDocument": false,
        "keyPrefix": "testmann",
        "minNumericKeyValue" : 0,
        "maxNumericKeyValue" : 100000
      },
      "stopAfter": {
        "runtime": 10000
      },
      "threads": 1
    }
  }
}
```

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -
  - [ ] Backport for 3.7: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/tests/sepp/Execution.cpp
+++ b/tests/sepp/Execution.cpp
@@ -41,6 +41,14 @@ Execution::Execution(Options const& options, std::shared_ptr<Workload> workload)
 
 Execution::~Execution() {
   _state.store(ExecutionState::kStopped);
+  joinThreads();
+}
+
+void Execution::createThreads(Server& server) {
+  _threads = _workload->createThreads(*this, server);
+}
+
+void Execution::joinThreads() {
   for (auto& thread : _threads) {
     if (thread->_thread.joinable()) {
       thread->_thread.join();
@@ -48,27 +56,44 @@ Execution::~Execution() {
   }
 }
 
-void Execution::createThreads(Server& server) {
-  _threads = _workload->createThreads(*this, server);
-}
-
 ExecutionState Execution::state(std::memory_order order) const noexcept {
   return _state.load(order);
 }
 
+void Execution::signalStartingThread() noexcept { _activeThreads.fetch_add(1); }
 void Execution::signalFinishedThread() noexcept { _activeThreads.fetch_sub(1); }
 
+void Execution::advanceStatusIfNotStopped(ExecutionState state) noexcept {
+  ExecutionState value = _state.load();
+  if (state == ExecutionState::kStopped && value == state) {
+    // already stopped
+    return;
+  }
+  // never move the status backwards
+  while (value != ExecutionState::kStopped &&
+         !_state.compare_exchange_strong(value, state)) {
+    std::this_thread::yield();
+  }
+}
+
+void Execution::stop() noexcept {
+  ExecutionState value = ExecutionState::kRunning;
+  // try to advance from Running to Stopped
+  while (!_state.compare_exchange_strong(value, ExecutionState::kStopped)) {
+    std::this_thread::yield();
+  }
+}
+
 Report Execution::run() {
-  _activeThreads.store(_threads.size());
-  _state.store(ExecutionState::kPreparing);
+  advanceStatusIfNotStopped(ExecutionState::kPreparing);
 
   waitUntilAllThreadsAre(ThreadState::kRunning);
 
-  _state.store(ExecutionState::kInitializing);
+  advanceStatusIfNotStopped(ExecutionState::kInitializing);
 
   waitUntilAllThreadsAre(ThreadState::kReady);
 
-  _state.store(ExecutionState::kRunning);
+  advanceStatusIfNotStopped(ExecutionState::kRunning);
 
   auto start = std::chrono::high_resolution_clock::now();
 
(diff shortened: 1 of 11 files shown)


tests/sepp/Execution.cpp
```diff
@@ -41,34 +41,63 @@ Execution::Execution(Options const& options, std::shared_ptr<Workload> workload)
...
+  ExecutionState value = _state.load();
+  if (state == ExecutionState::kStopped && value == state) {
+    // already stopped
+    return;
+  }
+  // never move the status backwards
+  while (value != ExecutionState::kStopped &&
+         !_state.compare_exchange_strong(value, state)) {
+    std::this_thread::yield();
+  }
```
mpoeter, line 76: We can remove the if here. This if is only fulfilled if `value == ExecutionState::kStopped`, but in that case we also won't enter the loop.
```suggestion
  // never move the status backwards
  while (value != ExecutionState::kStopped &&
         !_state.compare_exchange_strong(value, state)) {
    std::this_thread::yield();
  }
```

tests/sepp/Execution.cpp
```diff
@@ -41,34 +41,63 @@ Execution::Execution(Options const& options, std::shared_ptr<Workload> workload)
...
+}
+
+void Execution::stop() noexcept {
+  ExecutionState value = ExecutionState::kRunning;
+  // try to advance from Running to Stopped
+  while (!_state.compare_exchange_strong(value, ExecutionState::kStopped)) {
+    std::this_thread::yield();
+  }
```
mpoeter, line 84: Why not just `_state.store(ExecutionState::kStopped)`?

tests/sepp/Execution.cpp
```diff
@@ -91,22 +120,26 @@ Report Execution::run() {
...
   std::chrono::duration<double, std::milli> runtime =
       std::chrono::high_resolution_clock::now() - start;
-  return buildReport(runtime.count());
-}
 
-Report Execution::buildReport(double runtime) {
-  for (auto& thread : _threads) {
-    if (thread->_thread.joinable()) {
-      thread->_thread.join();
+  joinThreads();
```
mpoeter, line 131: I don't think this is really necessary since after `waitUntilAllThreadsAre(ThreadState::kFinished)` all thread objects are fully updated so we can already access e.g. failure information even if the thread has not yet terminated. But if you prefer we can also leave it here.

tests/sepp/Workloads/GetByPrimaryKey.cpp
```diff
@@ -0,0 +1,181 @@
...
+      s = rootDB->Get(ro, docCF, keyBuilder.string(), &val);
+
+      if (!s.ok()) {
+        throw std::runtime_error("Failed to fetch document value: " +
+                                 s.ToString());
+      }
+    }
+    ++_operations;
+
+    if (_operations % 512 == 0 && execution().stopped()) {
```
mpoeter, line 167:
```suggestion
    if (_operations % 512 == 0 && (execution().stopped() || shouldStop())) {
```

tests/sepp/Workloads/GetByPrimaryKey.cpp
```diff
@@ -0,0 +1,181 @@
...
+       i < _options.maxNumericKeyValue; ++i) {
+    temp.clear();
+    if (!_options.keyPrefix.empty()) {
+      temp.append(_options.keyPrefix);
+    }
+    temp.append(std::to_string(i));
+
+    keyBuilder.constructPrimaryIndexValue(indexId, std::string_view(temp));
+
+    s = rootDB->Get(ro, indexCF, keyBuilder.string(), &val);
```
mpoeter, line 149: Should we use `RocksDBPrimaryIndex::lookupKey` instead to avoid reimplementing this and instead rely on the same code that is used when running arangod?

tests/sepp/Workloads/IterateDocuments.cpp
```diff
@@ -105,11 +111,12 @@ void IterateDocuments::Thread::run() {
   std::unique_ptr<rocksdb::Iterator> it(rootDB->NewIterator(ro, docCF));
 
   OperationOptions options;
-  std::uint64_t numIts = 0;
-  for (it->Seek(bounds.start()); it->Valid(); it->Next(), ++numIts) {
-    // TODO
+  for (it->Seek(bounds.start()); it->Valid(); it->Next()) {
+    ++_operations;
+    if (_operations % 512 == 0 && execution().stopped()) {
```
mpoeter, line 116:
```suggestion
    if (_operations % 512 == 0 && (execution().stopped() || shouldStop())) {
```
Alternatively the `execution().stopped()` check could be integrated in `shouldStop()` (I would actually prefer that).

mpoeter (changes requested):

tests/sepp/Execution.cpp
jsteemann: Done

tests/sepp/Workloads/GetByPrimaryKey.cpp
jsteemann: Done

tests/sepp/Workloads/IterateDocuments.cpp
jsteemann: Done

tests/sepp/Execution.cpp
jsteemann: I think you are right, and it is not strictly necessary to call this. However, I left this line in to make it obvious to the reader that no threads will be dangling around here.

tests/sepp/Workloads/GetByPrimaryKey.cpp
jsteemann: I internally did not use `lookupKey`, because it has a lot of code for consulting the in-memory cache etc.
In the SEPP tests I prefer to measure the storage engine performance and try to avoid most of the rest. Thus sticking to the RocksDB calls here.

jsteemann:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/21951/

PR #16400, opened 2022-06-13
Fix compile warnings/errors with clang14 and gcc11

goedderz:
### Scope & Purpose

### DIFF
--- a/arangod/IResearch/AqlHelper.cpp
+++ b/arangod/IResearch/AqlHelper.cpp
@@ -646,7 +646,8 @@ bool attributeAccessEqual(aql::AstNode const* lhs, aql::AstNode const* rhs,
 
   // TODO: is the "&" intionally. If yes: why?
   // cppcheck-suppress uninitvar; false positive
-  while (lhsValue.read(lhs, ctx) & (int)rhsValue.read(rhs, ctx)) {
+  while (lhsValue.read(lhs, ctx) &
+         static_cast<unsigned>(rhsValue.read(rhs, ctx))) {
     if (lhsValue != rhsValue) {
       return false;
     }
--- a/arangod/IResearch/AqlHelper.cpp
+++ b/arangod/IResearch/AqlHelper.cpp
@@ -646,7 +646,7 @@ bool attributeAccessEqual(aql::AstNode const* lhs, aql::AstNode const* rhs,
 
   // TODO: is the "&" intionally. If yes: why?
   // cppcheck-suppress uninitvar; false positive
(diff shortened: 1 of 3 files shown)


arangod/IResearch/AqlHelper.cpp
```diff
@@ -646,7 +646,7 @@ bool attributeAccessEqual(aql::AstNode const* lhs, aql::AstNode const* rhs,
 
   // TODO: is the "&" intionally. If yes: why?
   // cppcheck-suppress uninitvar; false positive
-  while (lhsValue.read(lhs, ctx) & rhsValue.read(rhs, ctx)) {
+  while (lhsValue.read(lhs, ctx) & (int)rhsValue.read(rhs, ctx)) {
```
MBkkt, line 649: I don't like fix:
c style cast
signed integer operand with a binary bitwise operator

MBkkt (approved):

arangod/IResearch/AqlHelper.cpp
goedderz: Good points. Would
```suggestion
  while (lhsValue.read(lhs, ctx) &
         static_cast<unsigned>(rhsValue.read(rhs, ctx))) {
```
work for you? Alternatively, we could go more verbose like
```cpp
  bool lhsRead = lhsValue.read(lhs, ctx);
  bool rhsRead = rhsValue.read(rhs, ctx);
  while (lhsRead && rhsRead) {
    if (lhsValue != rhsValue) {
      return false;
    }

    lhs = lhsValue.root;
    rhs = rhsValue.root;
    lhsRead = lhsValue.read(lhs, ctx);
    rhsRead = rhsValue.read(rhs, ctx);
  }
  ```

Comment (goedderz) [reply arangod/IResearch/AqlHelper.cpp]: Alright thanks, I've applied the first suggestion in https://github.com/arangodb/arangodb/pull/16400/commits/7e0ed87cb12c58e3299d1a50adb9f0d11527eeb1.

File: arangod/IResearch/AqlHelper.cpp [hunk 2]
```diff
@@ -646,7 +646,8 @@ bool attributeAccessEqual(aql::AstNode const* lhs, aql::AstNode const* rhs,
 
   // TODO: is the "&" intionally. If yes: why?
   // cppcheck-suppress uninitvar; false positive
-  while (lhsValue.read(lhs, ctx) & rhsValue.read(rhs, ctx)) {
+  while (lhsValue.read(lhs, ctx) &
```
maierlars, line 649: now I wonder as well 🤔

maierlars (approved):

PR #16407, opened 2022-06-14
[AQL] Add ALL_SHORTEST_PATHS feature

markuspf:
### Scope & Purpose

Add feature to compute `ALL_SHORTEST_PATHS` between two given vertices.

* Design Doc PR: https://github.com/arangodb/documents/pull/114
* Documentation PR: https://github.com/arangodb/docs/pull/1071
* Jira: https://arangodb.atlassian.net/browse/APM-531


Simran-B:
Do I get this right that it determines the length of a shortest path and then searches all shortest paths of that length, whereas k shortest path doesn't stop at the shortest length?

aMahanna:
note: the following files are out of date & need to be re-generated: `grammar.cpp`, `grammar.hpp`, `tokens.cpp`

aMahanna:
note: a decision needs to be made in regards to which approach should be taken for `ALL_SHORTEST_PATHS` behavior:

Approach 1: https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Graph/Enumerators/TwoSidedEnumerator.h#L56-L57

Approach 2:
https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Aql/EnumeratePathsNode.cpp#L398-L401

Crossroads (where the two approaches meet):
https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Graph/Enumerators/TwoSidedEnumerator.cpp#L445-L450

aMahanna (commented):
⚠️  note: a decision needs to be made in regards to which approach should be taken for `ALL_SHORTEST_PATHS` behavior:

Approach 1: https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Graph/Enumerators/TwoSidedEnumerator.h#L56-L57

Approach 2:
https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Aql/EnumeratePathsNode.cpp#L398-L401

Crossroads (where the two approaches meet):
https://github.com/arangodb/arangodb/blob/f1c9c3cb46e1561da006fd555e57d5f45ecaa3cc/arangod/Graph/Enumerators/TwoSidedEnumerator.cpp#L445-L450

Sacramentix:
Nice features, It's exactly what I was looking for!
At the moment, i need to do 2 query to do this kind of stuff.
First querythe shortest path length and then k paths query to get all shortest path.

arangod/Aql/grammar.y
```diff
@@ -730,6 +732,15 @@ k_paths_graph_info:
     }
   ;
 
+all_shortest_paths_graph_info:
+    graph_direction T_ALL_SHORTEST_PATHS expression T_STRING expression graph_subject options {
+      auto nodeStart = parser->ast()->createNodeValueInt(1);
```
mchacki, line 737: Is this actually correct?

What about the pathological case of ALL_SHORTEST_PATH "A" TO "A" ?
This should be one, and that should have an empty list of edges, and only A as a vertex.

arangod/Graph/Enumerators/TwoSidedEnumerator.cpp
```diff
@@ -477,7 +485,7 @@ template<class QueueType, class PathStoreType, class ProviderType,
 auto TwoSidedEnumerator<QueueType, PathStoreType, ProviderType,
                         PathValidator>::searchDone() const -> bool {
   return _left.noPathLeft() || _right.noPathLeft() ||
-         _left.getDepth() + _right.getDepth() > _options.getMaxDepth();
+         _left.getDepth() + _right.getDepth() > _baselineDepth;
```
mchacki, line 488: Like this :+1:

arangod/Graph/Options/TwoSidedEnumeratorOptions.cpp
```diff
@@ -34,4 +34,11 @@ TwoSidedEnumeratorOptions::TwoSidedEnumeratorOptions(size_t minDepth,
...
-size_t TwoSidedEnumeratorOptions::getMaxDepth() const { return _maxDepth; }
\ No newline at end of file
+size_t TwoSidedEnumeratorOptions::getMaxDepth() const { return _maxDepth; }
+
+bool TwoSidedEnumeratorOptions::getStopAtFirstDepth() const {
+  return _stopAtFirstDepth;
+}
+void TwoSidedEnumeratorOptions::setStopAtFirstDepth(bool stopAtFirstDepth) {
+  _stopAtFirstDepth = stopAtFirstDepth;
+}
```
mchacki, line 44:
```suggestion
}

```

tests/js/server/aql/aql-graph.js
```diff
@@ -1646,6 +1672,68 @@ function kPathsTestSuite() {
...
+      gm._drop(gn, true);
+    },
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief checks if we are able to find all shortest paths when using ANY direction with ALL_SHORTEST_PATHS
+/// One of the edges is used in both directions.
+////////////////////////////////////////////////////////////////////////////////
+
+    testAllShortestPathsAnyUseEdgeTwice: function () {
+      // TODO Anthony: Clarify what "One of the edges is used in both directions" means in ASP context
```
mchacki, line 1718: Is this TODO still open?
i am missing a bit context here, can we elaborate a bit more on this question? (Happy to do so on slack)

tests/js/server/aql/aql-graph.js
```diff
@@ -1646,6 +1672,68 @@ function kPathsTestSuite() {
...
+      ["s", "t", "a", "b", "c", "d", "e", "f", "g"].map((elem) => {
+        db[vn].insert({_key: elem});
+      });
+
+      [
+        ["s", "a"], ["s", "b"], ["a", "b"], ["a", "c"], ["b", "c"],
+        ["c", "d"], ["d", "t"],
+        ["c", "e"], ["e", "t"],
+        ["c", "f"], ["f", "t"],
+        ["c", "g"], ["g", "t"]
```
mchacki, line 1698: To make the "ANY" test different.
I would suggest to add one or two "wrong direction" edges to this graph.
Right now all paths flow from S to T.

For any we need a path like this:

S -> X <- Y <- Z -> T
which should be found
If we have another edge Y -> Z in above example, ANY should give two paths:
S -> X <- Y <- Z -> T (as above)
S -> X <- Y -> Z -> T (use the other edge from Y to Z)

tests/js/server/aql/aql-graph-all-shortest-path.js
```diff
@@ -0,0 +1,450 @@
...
+/*global assertEqual, assertTrue, assertFalse, assertMatch, fail */
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief tests for query language, graph functions
+///
+/// @file
+///
+/// DISCLAIMER
+///
+/// Copyright 2010-2012 triagens GmbH, Cologne, Germany
```
mchacki, line 11:
```suggestion
/// Copyright 2022-2022 ArangoDB GmbH, Cologne, Germany
```

tests/Graph/AllShortestPathsFinderTest.cpp
```diff
@@ -0,0 +1,644 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2020-2020 ArangoDB GmbH, Cologne, Germany
```
mchacki, line 4:
```suggestion
/// Copyright 2022-2022 ArangoDB GmbH, Cologne, Germany
```

tests/Graph/AllShortestPathsFinderTest.cpp
```diff
@@ -0,0 +1,644 @@
...
+    mockGraph.addEdge(3, 5);
+
+    /* a hexagon 6->7->8->9->10->11->6 */
+    mockGraph.addEdge(6, 7);
+    mockGraph.addEdge(7, 8);
+    mockGraph.addEdge(8, 9);
+    mockGraph.addEdge(9, 10);
+    mockGraph.addEdge(10, 11);
+    mockGraph.addEdge(11, 6);
+    // TODO Anthony: Figure out how to set direction to ANY (looks like default
```
mchacki, line 99: Are those TODOs still open?

The edge on the graph is always outbound, I think we need to look into the Index calculation.
I think for the mock graph that is indeed only outbound for now.

mchacki: I do not request to change anything here.

tests/Graph/AllShortestPathsFinderTest.cpp
```diff
@@ -0,0 +1,644 @@
...
+    EXPECT_TRUE(result.isEmpty());
+    EXPECT_TRUE(finder.isDone());
+  }
+  {
+    aql::TraversalStats stats = finder.stealStats();
+    EXPECT_EQ(stats.getScannedIndex(), 0);
+  }
+}
+
+TEST_P(AllShortestPathsFinderTest, path_depth_0) {
```
mchacki, line 296: Nice, so depth 0 is tested here!

tests/js/server/aql/aql-graph-all-shortest-path.js
```diff
@@ -0,0 +1,450 @@
...
+          RETURN path
+      `;
+      const result = db._query(query).toArray();
+      allPathsDiffer(result);
+      assertEqual(result.length, 3);
+      for (let i = 0; i < 3; ++i) {
+        isPathValid(result[i], 3, true);
+      }
+    }
+
```
mchacki, line 356: I would like to request a quick and simple test here, where SOURCE and TARGET are Identical.
The underlying C++ mechanism can do it, but I think I spottet a potential glitch in the grammar (see grammar.y comment)
This may just work out of the box as the value is ignored, but better have a test here.

mchacki (changes requested):
Very good PR.
I am happy with everything I have seen.
My comments are mostly minor items.
I think there is a small glitch if SOURCE and TARGET are identical, we should get this under test just to be save against erratic behavior.

arangod/Aql/grammar.y
aMahanna: good catch!

With `auto nodeStart = parser->ast()->createNodeValueInt(1);`, running `ALL_SHORTEST_PATHS "A" TO "A"` returns return an empty list `[]`.

With `auto nodeStart = parser->ast()->createNodeValueInt(0);`, running `ALL_SHORTEST_PATHS "A" TO "A"` returns an empty list of edges, and only A as a vertex.

This has been addressed in 7dd39c35d94

tests/js/server/aql/aql-graph.js
aMahanna: ah, it's out of date, thank you for noticing

addressed in 7dd39c35d94

tests/js/server/aql/aql-graph.js
aMahanna: added! thank you for that

addressed in 7dd39c35d94

tests/Graph/AllShortestPathsFinderTest.cpp
aMahanna: these were also out of date, addressed in 7dd39c35d94

tests/js/server/aql/aql-graph-all-shortest-path.js
aMahanna: done!

addressed in https://github.com/arangodb/arangodb/commit/7dd39c35d9422294200c784bbf804be15e0e7bab

js/apps/system/_admin/aardvark/APP/react/public/assets/src/mode-aql.js
```diff
@@ -88,7 +88,7 @@ var TextHighlightRules = require("./text_highlight_rules").TextHighlightRules;
 var AqlHighlightRules = function() {
 
     var keywords = (
-        "for|return|filter|search|sort|limit|let|collect|asc|desc|in|into|insert|update|remove|replace|upsert|options|with|and|or|not|distinct|graph|shortest_path|outbound|inbound|any|all|none|aggregate|like|k_shortest_paths|k_paths|prune|window"
+        "for|return|filter|search|sort|limit|let|collect|asc|desc|in|into|insert|update|remove|replace|upsert|options|with|and|or|not|distinct|graph|shortest_path|outbound|inbound|any|all|none|aggregate|like|k_shortest_paths|k_paths|all_shortest_paths|prune|window"
```
aMahanna, line 91: just want to double check in this case that this is indeed all that is required for frontend AQL configuration

hkernbach: We also do have:
```
    aqlKeywords: 'for|return|filter|search|sort|limit|let|collect|asc|desc|in|into|' +
      'insert|update|remove|replace|upsert|options|with|and|or|not|' +
      'distinct|graph|shortest_path|outbound|inbound|any|all|none|aggregate|like|count',
```

In: 
`frontend/js/views/spotlightView.js`

hkernbach: But of course, best solution would be to have a unified way of using this ... and generated if possible.

aMahanna: thank you! addressed in https://github.com/arangodb/arangodb/pull/16407/commits/e9b99ff481a02e06356219436d4ea6af2908a610

mchacki (approved):
LGTM, thanks again! This is a blast

PR #16409, opened 2022-06-15
[CINFRA-382] Calculate and use effective write concern

markuspf:
### Scope & Purpose

Implements the configuration change logic for changing of write concern.


markuspf:
Reviewer:
 * Pay attention to the leadership election in `Supervision.cpp`
 * This PR sadly contains some fixes for the ModelChecker, which are unrelated to the ticket
 * A lot of the code changes are actually to enable testing.

arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
```diff
@@ -261,6 +258,9 @@ struct LogCurrentSupervision {
 
   using StatusReport = std::vector<StatusMessage>;
 
+  // TODO / FIXME: This prevents assumedWriteConcern from
+  // being initialised to
```
maierlars, line 262: 0 ?

arangod/Replication2/ReplicatedLog/SupervisionAction.h
```diff
@@ -290,6 +284,50 @@ auto inspect(Inspector& f, UpdateLogConfigAction& x) {
...
+  auto execute(ActionContext& ctx) const -> void {
+    ctx.modify<LogPlanSpecification>([&](LogPlanSpecification& plan) {
+      plan.participantsConfig.config.effectiveWriteConcern =
+          newEffectiveWriteConcern;
+      plan.participantsConfig.generation += 1;
+    });
+    ctx.modify<LogCurrentSupervision>(
+        [&](LogCurrentSupervision& currentSupervision) {
+          currentSupervision.assumedWriteConcern = std::min(
+              currentSupervision.assumedWriteConcern, newEffectiveWriteConcern);
```
maierlars, line 300: I was looking for that piece of logic in the supervision code and couldn't find it. While the code is correct, comparing the design document with the code, this detail is missing. Would it make sense to move the computation of the value into the supervision and only set the value here?

tests/Replication2/Helper/ModelChecker/Predicates.h
```diff
@@ -148,4 +148,49 @@ static inline auto nonExcludedServerHasSnapshot() {
...
+
+      return currentSupervision.assumedWriteConcern <=
+             planConfig.effectiveWriteConcern;
+    }
+    return false;
+  });
+}
+
+static inline auto
+isWriteConcernUsedForCommitLessThanOrEqualToAssumedWriteConcern() {
```
maierlars, line 178: Only judging by the name, is this predicate actually something we want to assert? The writeConcern used for commits should always be bigger or equal to the assumed write concern, right? Because the assumed write concern is used for leader election and if the form quorums with fewer servers than assumedWriteConcern indicates, we might elect a leader and loose data...

tests/Replication2/Helper/ModelChecker/Predicates.h
```diff
@@ -148,4 +148,49 @@ static inline auto nonExcludedServerHasSnapshot() {
...
+    if (not state.logLeaderWriteConcern) {
+      return true;
+    }
+
+    if (state.replicatedLog and state.replicatedLog->plan and
+        state.replicatedLog->current and
+        state.replicatedLog->current->supervision) {
+      return state.replicatedLog->current->supervision->assumedWriteConcern <=
+             *state.logLeaderWriteConcern;
+    }
```
maierlars, line 191: never mind, the implementation is correct. maybe change the name?

maierlars (approved):
LGTM

tests/Replication2/Helper/ModelChecker/Predicates.h
markuspf: Name changed,

arangod/Replication2/ReplicatedLog/AgencyLogSpecification.h
markuspf: Comment expanded a bit.

arangod/Replication2/ReplicatedLog/SupervisionAction.h
markuspf: @goedderz and I stubled over exactly this yesterday too; It's a bit of an unfortunate judgement call which side to put the change on; I will put the computation on the Supervision side as well, as that makes it more obvious what hapens.

maierlars (approved):

PR #16428, opened 2022-06-20
SORT receiving input incrementally in rows 

cpjulia:
### Scope & Purpose

This PR makes the sort executor for queries with the SORT keyword receive the entries in rows, incrementally, instead of receiving a matrix containing all the input at once in memory. 

This is to be used later in conjunction with rocksdb temporary storage of the entries for queries with a large amount of entries in the input not to be held in memory and also for the rocksdb storage to store the entries in order.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  No backport required.

#### Related Information

- [ ] Docs PR: don't know if needed 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: not this, but related: https://arangodb.atlassian.net/browse/APM-84
- [ ] Design document:


arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,129 @@ SortExecutor::~SortExecutor() {
...
-  // If the execution state was not waiting it is guaranteed that we get a
-  // matrix. Maybe empty still
-  TRI_ASSERT(_input != nullptr);
-  if (_input == nullptr) {
-    THROW_ARANGO_EXCEPTION(TRI_ERROR_OUT_OF_MEMORY);
-  }
-  // After allRows the dependency has to be done
-  TRI_ASSERT(state == ExecutorState::DONE);
+  ResourceUsageScope guard(_infos.getResourceMonitor(),
+                           memoryUsageForRowIndexes);
```
jsteemann, line 144: We should use the "number of data rows" information here to reserve memory for `_rowIndexes` here, before we `emplace_back` the rows individually.

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,129 @@ SortExecutor::~SortExecutor() {
...
-  while (_returnNext < _sortedIndexes.size() && !output.isFull()) {
-    InputAqlItemRow inRow = _input->getRow(_sortedIndexes[_returnNext]);
+  ExecutorState state = ExecutorState::HASMORE;
+
+  if (!_inputReady) {
+    consumeInput(inputRange, state);
+    if (inputRange.upstreamState() == ExecutorState::HASMORE) {
+      return {state, NoStats{}, upstreamCall};
+    }
+    if (_returnNext < _rowIndexes.size()) {
```
jsteemann, line 183: Is this `if` condition still needed? When we get here, `_returnNext` should always be `0`.

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,129 @@ SortExecutor::~SortExecutor() {
...
+      _inputBlocks.emplace_back(inputBlock);
+    }
   }
 
-  if (_returnNext >= _sortedIndexes.size()) {
+  if (_returnNext >= _rowIndexes.size() && !_rowIndexes.empty()) {
     // Bail out if called too often,
     // Bail out on no elements
     return {ExecutorState::DONE, NoStats{}, 0, upstreamCall};
   }
```
jsteemann, line 238: I don't think this entire block is necessary. 
If `_inputReady` is false, it has no effect, because `_rowIndexes` will still be empty.
If `_inputReady` is true, we don't need it, because the following `if`s and `while`s take care of the situation already.

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,129 @@ SortExecutor::~SortExecutor() {
...
+  if (!_inputReady) {
+    consumeInput(inputRange, state);
+    if (inputRange.upstreamState() == ExecutorState::HASMORE) {
+      return {state, NoStats{}, 0, upstreamCall};
+    }
+    if (_returnNext < _rowIndexes.size()) {
+      doSorting();
+      _inputReady = true;
+    }
+  }
```
jsteemann, line 249: This entire `if` block should be fused with the `if` block at the beginning of this method, as it has the same `if` condition.

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,129 @@ SortExecutor::~SortExecutor() {
...
   }
 
-  while (_returnNext < _sortedIndexes.size() && call.shouldSkip()) {
-    InputAqlItemRow inRow = _input->getRow(_sortedIndexes[_returnNext]);
+  if (!_inputReady) {
+    consumeInput(inputRange, state);
+    if (inputRange.upstreamState() == ExecutorState::HASMORE) {
+      return {state, NoStats{}, 0, upstreamCall};
+    }
+    if (_returnNext < _rowIndexes.size()) {
```
jsteemann, line 245: Is this `if` condition really necessary? I think it can be omitted.

arangod/Aql/SortExecutor.cpp
cpjulia: Addressed

arangod/Aql/SortExecutor.cpp
cpjulia: Addressed

arangod/Aql/SortExecutor.cpp
cpjulia: Addressed

arangod/Aql/SortExecutor.cpp
cpjulia: Addressed

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,109 @@ SortExecutor::~SortExecutor() {
...
-  ExecutorState state;
+void SortExecutor::consumeInput(AqlItemBlockInputRange& inputRange,
+                                ExecutorState& state) {
+  size_t numDataRows = inputRange.countDataRows();
+  size_t memoryUsageForRowIndexes =
+      numDataRows * sizeof(AqlItemMatrix::RowIndex);
 
-  // We need to get data
-  std::tie(state, _input) = inputMatrix.getMatrix();
+  _rowIndexes.reserve(numDataRows);
```
jsteemann, line 143: This `reserve` call is not correct, because it will try to increase the capacity to `numDataRows`. But what if the vector already contains data?
The correct call should be `_rowIndexes.reserve(_rowIndexes.size() + numDataRows);`

arangod/Aql/SortExecutor.cpp
```diff
@@ -135,146 +134,109 @@ SortExecutor::~SortExecutor() {
...
+      return {state, NoStats{}, 0, upstreamCall};
+    }
+    doSorting();
+    _inputReady = true;
   }
 
-  while (_returnNext < _sortedIndexes.size() && call.shouldSkip()) {
-    InputAqlItemRow inRow = _input->getRow(_sortedIndexes[_returnNext]);
+  while (_returnNext < _rowIndexes.size() && call.shouldSkip()) {
+    InputAqlItemRow inRow(_inputBlocks[_rowIndexes[_returnNext].first],
```
jsteemann, line 232: Why does this `inRow` need to be built here? I can't see if used later, so we might as well not build it in the first place.

cpjulia: Addressed

arangod/Aql/SortExecutor.cpp
cpjulia: Addressed

PR #16435, opened 2022-06-21
[CINFRA-422] Replicated Log Metrics

maierlars:
### Scope & Purpose

This PR adds more metrics to the Replicated Logs. Those metrics allow together with the Replicated State Metrics (follow up PR) to observe if some parts of the replicated become slow.

To properly test the compaction counter, I added the compaction threshold to the options, and the options were missing from the LogFollower.


Documentation/Metrics/arangodb_replication2_replicated_log_number_accepted_entries.yaml
```diff
@@ -0,0 +1,13 @@
...
+help: |
+  Number of accepted (not yet committed) log entries.
+unit: number
+type: counter
+category: Replication
+complexity: simple
+exposedBy:
+  - dbserver
+description: |
+  Number of accepted (not yet committed) log entries. The entries has been inserted in to the log
```
ansoboleva, line 12: The entries have been inserted into the log
 and are actively replicated but a quorum has not yet been reached.

apetenchea (approved):

PR #16452, opened 2022-06-24
OskarTestSuitesBlockList <-> test-definitions.txt

dothebart:
### Scope & Purpose

So far we would have the `OskarTestSuitesBlockList` to stop removed testsuites to be executed in PR runs.
Since now the Branch has the `test-definitions.txt` the contents of the Blacklist can simply be removed from the test definition, 
and the blacklist be emptied.


KVS85 (approved):
LGTM.

dothebart:
unrelated unstable test in windows run.

PR #16453, opened 2022-06-24
Feature 3.9/remove blacklist contents

dothebart:
### Scope & Purpose

- Migrate this branch to use `test-definitions.txt`

- So far we would have the `OskarTestSuitesBlockList` to stop removed testsuites to be executed in PR runs.
Since now the Branch has the `test-definitions.txt` the contents of the Blacklist can simply be removed from the test definition, 
and the blacklist be emptied.


dothebart:
tests blue except unrelated lockup

KVS85 (approved):
LGTM.

maierlars (approved):

PR #16457, opened 2022-06-24
[CINFRA] Agency Supervision Refactor

maierlars:
### Scope & Purpose
This PR tries to refactor the supervision code, without changing its behavior. Additionally
- unused data types have been removed
- an unused shutdown mechanism has been removed
- The network feature included the supervision.h and this caused the methods.cpp and utils.cpp to be rebuild every time on modifies the supervision.

### DIFF
--- a/arangod/Agency/Agent.h
+++ b/arangod/Agency/Agent.h
@@ -32,7 +32,6 @@
 #include "Agency/Inception.h"
 #include "Agency/State.h"
 #include "Agency/Store.h"
-#include "Agency/Supervision.h"
 #include "Futures/Promise.h"
 #include "Basics/ConditionLocker.h"
 #include "Basics/ReadWriteLock.h"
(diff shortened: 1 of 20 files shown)


arangod/Agency/Agent.h
```diff
@@ -385,7 +386,7 @@ class Agent final : public arangodb::ServerThread<ArangodServer>,
   Constituent _constituent;
 
   /// @brief Cluster supervision module
-  Supervision _supervision;
+  std::unique_ptr<Supervision> _supervision;
```
markuspf, line 389: Out of interest: Why?

maierlars: This allows me to have this header included without making the implementation of Supervision visible. Not having it as a pointer would require every user of Agent to also include Supervision.h.

markuspf (approved):
LGTM

neunhoef (commented):
LGTM

PR #16464, opened 2022-06-27
[CINFRA-463] Added last matched index.

maierlars:
### Scope & Purpose
Add lastMatchedIndex and use this instead of lastAckedIndex. lastAckedIndex is not initalized to 0 and only incremented.


arangod/Replication2/ReplicatedLog/LogStatus.h
```diff
@@ -100,6 +100,7 @@ struct FollowerStatistics : LogStatistics {
   AppendEntriesErrorReason lastErrorReason;
   std::chrono::duration<double, std::milli> lastRequestLatencyMS;
   FollowerState internalState;
+  TermIndexPair matchedIndex;
```
goedderz, line 103: `operator==` and `operator!=` have to be adapted.

goedderz (approved):

PR #16469, opened 2022-06-27
Change default macOS CI compiler to LLVM clang-12

KVS85:
Switched default CI compiler for macOS to LLVM clang-12 (installed from Homebrew).

oskar PR: https://github.com/arangodb/oskar/pull/399


fceller (approved):
As we cannot fix the Apple version, this will hopefully help fixing the version

PR #16472, opened 2022-06-29
[3.9] Reduce maintainer mode stack usage

goedderz:
### Scope & Purpose

Partial backport of https://github.com/arangodb/arangodb/pull/16308:

Reduce stack usage of assertions in maintainer mode.

### DIFF
--- a/lib/Basics/debugging.cpp
+++ b/lib/Basics/debugging.cpp
@@ -205,7 +205,7 @@ char const conpar<false>::open = '[';
 template<>
 char const conpar<false>::close = ']';
 
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
 thread_local std::ostringstream
     arangodb::debug::AssertionLogger::assertionStringStream;
-thread_local arangodb::debug::AssertionConditionalStream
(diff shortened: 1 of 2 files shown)


lib/Basics/debugging.cpp
```diff
@@ -204,3 +204,8 @@ template<>
 char const conpar<false>::open = '[';
 template<>
 char const conpar<false>::close = ']';
+
+thread_local std::ostringstream
+    arangodb::debug::AssertionLogger::assertionStringStream;
+thread_local arangodb::debug::AssertionConditionalStream
+    arangodb::debug::AssertionConditionalLogger::assertionStringStream;
```
jsteemann, line 211: @goedderz : Do we need to define these thread_local objects when we are not in maintainer mode?
Or rather, should we put these definitions inside an `#ifdef`?

goedderz: I think they're not necessary, I've wrapped them in an `#ifdef`. However, we should then also do this in `devel` (this is just a backport). I'll remind myself to do that next week.

KVS85 (approved):
LGTM.

PR #16476, opened 2022-06-30
start converting test to use fuerte connection instead of single requsts

dothebart:
### Scope & Purpose

This test doesn't use persistent connections yet, resulting in many collections in `TIME_WAIT` state. 


jsteemann:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/22236/

PR #16477, opened 2022-06-30
Documentation: APM-296

neunhoef:
### Scope & Purpose

This PR only edits documentation snippets for the read-from-followers
feature. No code is touched.

It accompanies this PR on the documentation repo:

### DIFF
--- a/Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
+++ b/Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
@@ -6,6 +6,23 @@
 
 A JSON object describing the query and query parameters.
 
+@RESTHEADERPARAM{x-arango-allow-dirty-read,string,optional}
+If the "x-arango-allow-dirty-read" header is given, then it must contain
+the value `true`. This will allow the coordinator to ask any shard
+replica for the data and not just the leader. However, this may then
+produce "dirty reads", see
+[Dirty Reads](./document-address-and-etag.html#dirty-reads)
+for details. Note that if this operation is part of a read-only streaming
+transaction (see "x-arango-trx-id" header below), then it is the
+transaction creation operation that decides about dirty reads, and not
+this header for the individual read operation.
+
+@RESTHEADERPARAM{x-arango-trx-id,string,optional}
+If the operation shall be part of a streaming transaction (see
+[Streaming Transactions](./transaction-stream-transaction.html) for
+details), then the "x-arango-trx-id" header must be set to the
+transaction id.
+
 @RESTBODYPARAM{query,string,required,string}
 contains the query string to be executed
 
--- a/Documentation/DocuBlocks/Rest/Documents/get_read_document.md
+++ b/Documentation/DocuBlocks/Rest/Documents/get_read_document.md
@@ -24,6 +24,23 @@ If the "If-Match" header is given, then it must contain exactly one
 Etag. The document is returned, if it has the same revision as the
 given Etag. Otherwise a *HTTP 412* is returned.
 
+@RESTHEADERPARAM{x-arango-allow-dirty-read,string,optional}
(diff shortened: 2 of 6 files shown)


Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
```diff
@@ -6,6 +6,23 @@
...
+the value `true`. This will allow the coordinator to ask any shard
+replica for the data and not just the leader. However, this may then
+produce "dirty reads", see
+[Dirty Reads](./document-address-and-etag.html#dirty-reads)
+for details. Note that if this operation is part of a read-only streaming
+transaction (see "x-arango-trx-id" header below), then it is the
+transaction creation operation that decides about dirty reads, and not
+this header for the individual read operation.
+
+@RESTHEADERPARAM{x-arango-trx-id,string,optional}
```
Simran-B, line 20: Should the type be `integer`? Seems to be a `uint64` internally. Everything is a string over HTTP initially, so it's not wrong, but I believe that the Swagger type should indicate the internal type, or rather what a client should map it to.

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
```diff
@@ -6,6 +6,23 @@
 
 A JSON object describing the query and query parameters.
 
+@RESTHEADERPARAM{x-arango-allow-dirty-read,string,optional}
+If the "x-arango-allow-dirty-read" header is given, then it must contain
+the value `true`. This will allow the coordinator to ask any shard
+replica for the data and not just the leader. However, this may then
+produce "dirty reads", see
+[Dirty Reads](./document-address-and-etag.html#dirty-reads)
```
Simran-B, line 14: Internal references do not work in Swagger UI, but they (mostly) will in the docs. We can substitute with absolute URLs like done here:
https://github.com/arangodb/arangodb/blob/733402d122524aa1f42f76b9bbef062e4191b836/Documentation/DocuBlocks/Rest/Views/post_api_view_arangosearch.md?plain=1#L16
This has the disadvantage of linking to the latest stable version of the docs, which may not match the version in use and it also changes over time. Alternatively, we can remove the links entirely and leave it to the users to find related information.

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
neunhoef: I added this part of the documentation in this PR, since it was obviously forgotten. I think the user facing side of the API only has an integer in there. However, cluster-internally, this header can be more complicated and thus is a string.

Simran-B: Thanks for the explanation. Looks like it's in fact a string to users:
`{"code":201,"error":false,"result":{"id":"2199063338824","status":"running"}}`
`typeof trx.id() // string`

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor.md
```diff
@@ -4,7 +4,19 @@
 
 @RESTHEADER{POST /_api/cursor, Create cursor, createQueryCursor}
 
-A JSON object describing the query and query parameters.
+@RESTHEADERPARAMETERS
+
+@RESTHEADERPARAM{x-arango-allow-dirty-read,boolean,optional}
```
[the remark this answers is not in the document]
neunhoef: The documentation is inconsistent here. Most occurrences of such HTTP headers are described in all lower case, some in mixed case. Functionally, all incoming headers are lowercased, so the case does not matter. Just use your own judgement as to how you would like to have it in the documentation.

Simran-B (approved):

KVS85:
- [x] https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/2154/

PR #16483, opened 2022-07-01
[CINFRA-471] Leader State Manager Resign Tests

maierlars:
### Scope & Purpose
Add Tests for resign calls to the Leader State Manager.


maierlars (commented):
LGTM

PR #16493, opened 2022-07-04
[PREGEL] [APM-539] Add Metrics Feature to Pregel

markuspf:
### Scope & Purpose

Add metrics collection to Pregel subsystem.


arangod/Pregel/PregelMetrics.h
```diff
@@ -0,0 +1,54 @@
...
+  explicit PregelMetrics(metrics::MetricsFeature& metricsFeature);
+
+ private:
+  template<typename Builder, bool mock = false>
+  static auto createMetric(metrics::MetricsFeature* metricsFeature)
+      -> std::shared_ptr<typename Builder::MetricT>;
+
+ protected:
+  template<typename MFP,
+           std::enable_if_t<std::is_same_v<metrics::MetricsFeature*, MFP> ||
```
jvolmer, line 44: no sure if we can use concepts now in ArangoDB: If yes, can't this enable_if be exchanged by a concept?

arangod/Pregel/PregelMetrics.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "PregelMetrics.h"
+
+#include "Pregel/PregelMetricsDeclarations.h"
+#include "Metrics/MetricsFeature.h"
+
+using namespace arangodb::pregel;
+
+PregelMetrics::PregelMetrics(metrics::MetricsFeature& metricsFeature)
+    : ReplicatedLogMetrics(&metricsFeature) {}
```
jvolmer, line 32: Is *ReplicatedLogMetrics* here correct?

arangod/Pregel/PregelMetrics.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+template<typename Builder, bool mock>
+auto PregelMetrics::createMetric(metrics::MetricsFeature* metricsFeature)
+    -> std::shared_ptr<typename Builder::MetricT> {
+  TRI_ASSERT((metricsFeature == nullptr) == mock);
+  if constexpr (!mock) {
```
jvolmer, line 38: This mock looks odd to me. Shouldn't it be possible to create a Mock metrics feature, which would have the function addShared do the same thing as in line 41? But I see that this is how it was done already in ReplicationLogMetrics, so perhaps it is more difficult.

arangod/Pregel/PregelMetrics.cpp
```diff
@@ -0,0 +1,83 @@
...
+    return metricsFeature->addShared(Builder{});
+  } else {
+    return std::dynamic_pointer_cast<typename Builder::MetricT>(
+        Builder{}.build());
+  }
+}
+
+template<
+    typename MFP,
+    std::enable_if_t<std::is_same_v<arangodb::metrics::MetricsFeature*, MFP> ||
```
jvolmer, line 48: no sure if we can use concepts now in ArangoDB: If yes, can't this enable_if be exchanged by a concept?

arangod/Pregel/PregelMetrics.cpp
```diff
@@ -0,0 +1,83 @@
...
+PregelMetrics::PregelMetrics(MFP metricsFeature)
+    : pregelExecutions(createMetric<arangodb_pregel_executions_number, mock>(
+          metricsFeature)) {
+#ifndef ARANGODB_USE_GOOGLE_TESTS
+  static_assert(!mock);
+  static_assert(!std::is_null_pointer_v<MFP>);
+#endif
+}
+
+MeasureTimeGuard::MeasureTimeGuard(
```
jvolmer, line 61: this struct is not defined in the header file

arangod/Pregel/Worker.cpp
```diff
@@ -513,6 +527,9 @@ void Worker<V, E, M>::_finishedProcessing() {
     // count all received messages
     _messageStats.receivedCount = _readCache->containedMessageCount();
 
+    _feature.metrics()->pregelMessagesReceived->count(
+        _readCache->containedMessageCount());
```
jvolmer, line 531: you can also count them during receiving in line 455 when iterating over messages, not sure what we want

arangod/Pregel/Worker.cpp
```diff
@@ -622,7 +639,13 @@ void Worker<V, E, M>::finalizeExecution(VPackSlice const& body,
     return;
   }
 
-  auto cleanup = [self = shared_from_this(), this, cb] {
+  VPackSlice store = body.get(Utils::storeResultsKey);
+  auto const doStore = store.isBool() && store.getBool() == true;
+  auto cleanup = [self = shared_from_this(), this, doStore, cb] {
+    if (doStore) {
+      _feature.metrics()->pregelWorkersStoringNumber->fetch_sub(1);
```
jvolmer, line 646: I don't understand why you substract here first and then add later again in line 667 - should it not be the other way around?

arangod/Pregel/PregelMetrics.h
```diff
@@ -0,0 +1,69 @@
...
+
+ public:
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelConductorsNumber;
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelConductorsLoadingNumber;
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelConductorsRunningNumber;
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelConductorsStoringNumber;
+
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelWorkersNumber;
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelWorkersLoadingNumber;
+  std::shared_ptr<metrics::Gauge<uint64_t>> const pregelWorkersRunningNumber;
```
jvolmer, line 58: This is never used

jvolmer (changes requested):
I found some possible bugs with the counting, I think they should be fixed first.

arangod/Pregel/GraphStore.cpp
```diff
@@ -497,13 +499,16 @@ void GraphStore<V, E>::loadEdges(
     if (edgeBuff == nullptr || edgeBuff->remainingCapacity() == 0) {
       edges.push_back(
           createBuffer<Edge<E>>(_feature, *_config, edgeSegmentSize()));
+      _feature.metrics()->pregelMemoryUsedForGraph->fetch_add(
+          edgeSegmentSize());
```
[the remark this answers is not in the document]
markuspf: If there is one, I don't know it ;)

arangod/Pregel/Conductor.cpp
```diff
@@ -364,6 +373,8 @@ void Conductor::finishedWorkerStartup(VPackSlice const& data) {
 
   _timing.loading.finish();
   _timing.computation.start();
+
+  _feature.metrics()->pregelConductorsRunningNumber->fetch_add(1);
```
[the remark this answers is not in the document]
markuspf: So do you set them to running at the beginning of each step and stop them at the end of each?

arangod/Pregel/Worker.cpp
markuspf: The reason I chose this place is because it is where currently `_messageStats` are collected, so the risk of getting inconsistent results is lower.

arangod/Pregel/Worker.cpp
markuspf: `cleanup` is a callback function that is called in lines 670 or after storing has finished, callback passed in in line 666;

Yes, this function is badly laid out. We should refactor it post freeze.

arangod/Pregel/PregelMetrics.h
markuspf: Urgh. I wrote the code to use them but did not commit; I shall add that code.

arangod/Pregel/Worker.cpp
jvolmer: makes sense, I did not see that!

jvolmer (approved):
LGTM

PR #16497, opened 2022-07-05
make sure tests clean up by adding finally

dothebart:
### Scope & Purpose

if test would throw in the meanwhile it would leave databases behind.

- [x] :hankey: cleanup


Dronplane (approved):
LGTM

PR #16509, opened 2022-07-06
if we run with an external server, we can by pass most of this

dothebart:
### Scope & Purpose

During the rework of the instance management we broke the support for connecting an external cluster. This PR brings back this functionality.

- [x] :hankey: Bugfix


hkernbach (approved):
LGTM and works. TY!

PR #16513, opened 2022-07-06
[PREGEL] [APM-538] Correct Runtime Measurement for Pregel.

markuspf:
### Scope & Purpose

This PR does two things:
 * Correctly reports the elapsed runtime for the loading, computation, and storing stages as well as the overall runtime for a Pregel run
 * Collects the runtime of each GlobalSuperStep and reports the list of runtimes in the status output as well.


jvolmer (approved):
LGTM - just one minor comment

arangod/Pregel/Conductor.cpp
```diff
@@ -780,7 +786,7 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
 
 ErrorCode Conductor::_finalizeWorkers() {
   _callbackMutex.assertLockedByCurrentThread();
-  _finalizationStartTimeSecs = TRI_microtime();
+  _timing.finalization.start();
```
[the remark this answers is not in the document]
markuspf: I removed the measurement of finalization, as it seemed meaningless.

tests/Pregel/DurationTest.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <variant>
+
+#include <fmt/core.h>
+
+#include <Pregel/Status/ExecutionStatus.h>
+
+using namespace arangodb::pregel;
+
+struct DurationTest : ::testing::Test {};
+TEST_F(DurationTest, test_duration) {
```
jvolmer, line 33: Out of curiosity: Why is this a fixture? There is just one test, so you could just use TEST(DurationTest, test_duration)

jvolmer (approved):
LGTM

tests/Pregel/DurationTest.cpp
markuspf: Dumb question: what is the difference? I just blindly copied some other test.

jvolmer: I thought fixtures are feature of test-frameworks to use the same data for several tests. The fixture can have a SetUp function that is called at the start of each test where the fixture is used and a TearDown function which is called at the end of each of these tests.
With `TEST_F` you define which fixture to use by giving the Test Suite name (the first parameter of `TEST_F`) the same name as the fixture struct - here DurationTest - and then `TEST_F` calls the SetUp and TearDown functions automatically when this test runs.
I the code you don't have any setup - actually you just have one test - therefore the struct is irrelevant and you can just use the normal `TEST` macro instead of `TEST_F`.
In my opinion fixtures often make the tests harder to read, but if your setup is complicated they can be useful. Additionally, in my opinion it is strange that googletest requires that the TestSuiteName - here DurationTest - has to be the Fixture name.

PR #16515, opened 2022-07-06
[PREGEL] APM-538 Add reporting for loading state observables

jvolmer:
Measure number of vertices, number of edges and memory used when loading the graph into pregel.
Measurements of workers are aggregated only when required.

### DIFF
--- a/arangod/Pregel/GraphStore.h
+++ b/arangod/Pregel/GraphStore.h
@@ -28,6 +28,7 @@
 #include "Pregel/GraphFormat.h"
 #include "Pregel/Iterators.h"
 #include "Pregel/Reports.h"
+#include "Pregel/Status/Status.h"
 #include "Pregel/TypedBuffer.h"
 #include "Utils/DatabaseGuard.h"
 
@@ -72,6 +73,8 @@ class GraphStore final {
   uint64_t numberVertexSegments() const { return _vertices.size(); }
   uint64_t localVertexCount() const { return _localVertexCount; }
   uint64_t localEdgeCount() const { return _localEdgeCount; }
+
+  Status status() const { return _observables.observe(); }
 
   GraphFormat<V, E> const* graphFormat() { return _graphFormat.get(); }
 
@@ -134,6 +137,8 @@ class GraphStore final {
   std::vector<std::unique_ptr<TypedBuffer<Edge<E>>>> _edges;
   std::vector<TypedBuffer<Edge<E>>*> _nextEdgeBuffer;
   std::vector<std::unique_ptr<TypedBuffer<char>>> _edgeKeys;
+
+  Observables _observables;
 
   // cache the amount of vertices
   std::set<ShardID> _loadedShards;
--- a/arangod/Pregel/Status/ConductorStatus.h
+++ b/arangod/Pregel/Status/ConductorStatus.h
@@ -61,11 +61,11 @@ struct ConductorStatus {
     }
     return status;
   }
-  auto updateWorkerStatus(ServerID const& id, Status const& status) -> void {
+  auto updateWorkerStatus(ServerID const& id, Status&& status) -> void {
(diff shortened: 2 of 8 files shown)


jvolmer:
TODO Merge parent first

arangod/Pregel/Status/ConductorStatus.h
```diff
@@ -33,34 +34,45 @@
...
-  TimeStamp timeStamp = std::chrono::system_clock::now();
-  std::unordered_map<ServerID, WorkerStatus> workers;
+struct AccumulatedConductorStatus {
+  Status status;
+  std::unordered_map<ServerID, Status> workers;
+  bool operator==(AccumulatedConductorStatus const&) const = default;
+};
+
+template<typename Inspector>
+auto inspect(Inspector& f, AccumulatedConductorStatus& x) {
```
romanatarango, line 48: Maybe, better names for `f` and `x`.

arangod/Pregel/Status/ConductorStatus.h
```diff
@@ -33,34 +34,45 @@
...
+  return f.object(x).fields(
+      f.field("aggregatedStatus",
+              x.status),  // TODO status entries should be on top level
+      f.field("workerStatus", x.workers));
+}
 
+struct ConductorStatus {
+  std::unordered_map<ServerID, Status> workers;
```
romanatarango, line 56: Are `workers` used for anything except iterating over them? If no, we could take `vector` instead of `unordered_set`.

romanatarango (approved):

arangod/Pregel/Status/ConductorStatus.h
markuspf: These functions are usually very short, and it is clear from the type of the parameter what it is. 

Making the names longer might make the inspector functions less readable, because the inspector is used all over the place, as is the struct.

Also, I don't think deviating from the current style across the ArangoDB codebase is a good idea. 

Don't call a variable `struct` though ;)

arangod/Pregel/Status/ConductorStatus.h
```diff
@@ -33,34 +34,45 @@
...
+      status.workers.emplace(id, Status{});
     }
     return status;
   }
-  auto updateWorkerStatus(ServerID const& id, WorkerStatus const& status)
-      -> void {
-    workers.at(id) = status;
-    timeStamp = status.timeStamp;
+  auto updateWorkerStatus(ServerID const& id, Status&& status) -> void {
+    workers[id] = status;
```
markuspf, line 65: Check whether `id` exists in the workers map; at the very least use `at` to get an exception and not UB.

arangod/Pregel/Status/ConductorStatus.h
markuspf: The map is the correct solution; the only thing you *could* do is to use `FlatHashMap`

arangod/Pregel/GraphStore.h
```diff
@@ -135,6 +138,8 @@ class GraphStore final {
   std::vector<TypedBuffer<Edge<E>>*> _nextEdgeBuffer;
   std::vector<std::unique_ptr<TypedBuffer<char>>> _edgeKeys;
 
+  Observables _observables;
```
markuspf, line 141: I really like the name `Observables`

markuspf (approved):

markuspf (changes requested):
Just requesting changes to make sure that you check for presence of server in your workers map.

jvolmer:
Successful build:
Windows: https://jenkins.arangodb.biz:3456/job/arangodb-matrix-pr-windows/15103/
All others: https://jenkins.arangodb.biz:3456/view/PR/job/arangodb-matrix-pr/22466/

markuspf (approved):

PR #16521, opened 2022-07-07
[PREGEL] APM-538 Add reporting for storing state observables

jvolmer:
Add number of vertices measurements.
Additionally, make all measurement variables optional to print them only when they are set.

### DIFF
--- a/arangod/Pregel/GraphStore.cpp
+++ b/arangod/Pregel/GraphStore.cpp
@@ -37,6 +37,7 @@
 #include "Pregel/CommonFormats.h"
 #include "Pregel/IndexHelpers.h"
 #include "Pregel/PregelFeature.h"
+#include "Pregel/Status/Status.h"
 #include "Pregel/TypedBuffer.h"
 #include "Pregel/Utils.h"
 #include "Pregel/WorkerConfig.h"
@@ -115,6 +116,10 @@ template<typename V, typename E>
 void GraphStore<V, E>::loadShards(
     WorkerConfig* config, std::function<void()> const& statusUpdateCallback,
     std::function<void()> const& finishedLoadingCallback) {
+  _observables.verticesLoaded = 0;
+  _observables.edgesLoaded = 0;
+  _observables.memoryBytesUsed = 0;
+
   _config = config;
   TRI_ASSERT(_runningThreads == 0);
 
@@ -397,7 +402,7 @@ void GraphStore<V, E>::loadVertices(
       _feature.metrics()->pregelMemoryUsedForGraph->fetch_add(segmentSize);
     }
     Vertex<V, E>* ventry = vertexBuff->appendElement();
-    _observables.memoryBytesUsed += sizeof(Vertex<V, E>);
+    *_observables.memoryBytesUsed += sizeof(Vertex<V, E>);
 
     VPackValueLength keyLen;
     VPackSlice keySlice = transaction::helpers::extractKeyFromDocument(slice);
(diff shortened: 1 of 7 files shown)


arangod/Pregel/GraphStore.cpp
```diff
@@ -515,12 +520,12 @@ void GraphStore<V, E>::loadEdges(
   size_t addedEdges = 0;
   auto buildEdge = [&](Edge<E>* edge, std::string_view toValue) {
     ++addedEdges;
-    ++_observables.edgesLoaded;
+    ++(*_observables.edgesLoaded);
```
romanatarango, line 523: This should be done only when we are sure that the edge is really added and no error occurs.

arangod/Pregel/GraphStore.cpp
```diff
@@ -729,16 +735,24 @@ void GraphStore<V, E>::storeVertices(std::vector<ShardID> const& globalShards,
     }
     builder.close();
     ++numDocs;
+    ++(*_observables.verticesStored);
+    if (numDocs % 1000 == 0) {
```
romanatarango, line 739: We may want to change the constant later. I suggest to give it a name and define it where all other such constants are kept (like in `Utils.h/cpp`).

arangod/Pregel/Status/Status.h
```diff
@@ -34,17 +35,34 @@
 #include <Pregel/Common.h>
 
 namespace arangodb::pregel {
+
+template<typename T>
+auto add(std::optional<T> a, std::optional<T> b) -> std::optional<T> {
```
romanatarango, line 40: Why not const reference? If in later use `a` and/or `b` are larger than just numbers, it would make no sense to copy them.

romanatarango (changes requested):

jvolmer:
TODO merge parent first

arangod/Pregel/GraphStore.cpp
```diff
@@ -114,6 +115,10 @@ template<typename V, typename E>
 void GraphStore<V, E>::loadShards(
     WorkerConfig* config, std::function<void()> const& statusUpdateCallback,
     std::function<void()> const& finishedLoadingCallback) {
+  _observables.verticesLoaded = 0;
+  _observables.edgesLoaded = 0;
+  _observables.memoryBytesUsed = 0;
```
markuspf, line 120: is this only executed once per worker?

arangod/Pregel/Status/Status.h
```diff
@@ -55,17 +74,30 @@ auto inspect(Inspector& f, Status& x) {
...
 }
 
 struct Observables {
-  std::atomic<std::size_t> verticesLoaded{0};
-  std::atomic<std::size_t> edgesLoaded{0};
-  std::atomic<std::size_t> memoryBytesUsed{0};
+  std::optional<std::atomic<std::size_t>> verticesLoaded;
+  std::optional<std::atomic<std::size_t>> edgesLoaded;
+  std::optional<std::atomic<std::size_t>> memoryBytesUsed;
+  std::optional<std::atomic<std::size_t>> verticesStored;
```
markuspf, line 85: Is this correct? Imagine one thread observing that `verticesLoaded` is `std::nullopt` and deciding to emplace `std::atomic<std::size_t>{5}` in there; another thread also observes `std::nullopt` and puts `std::atomic<std::size_t>{15}` into `verticesLoaded`, when you actually want `20` as the total you will have either `5` or `15` depending on who is faster.

markuspf (changes requested):
I am not sure the `std::optional<std::atomic<std::size_t>>` construct is correct.

markuspf (approved):
Looks good to me.

markuspf (approved):
LGTM

jvolmer:
Successful run for all except Windows: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/22502/

PR #16524, opened 2022-07-08
[CINFRA-442] Configuration change:  waitForSync

markuspf:
### Scope & Purpose

User can change the setting of `waitForSync` for a ReplicatedState/ReplicatedLog.


maierlars (approved):
We could have the following in a followup PR:
- Add the forwarding to replicated states? (maybe this is already there)
- have js tests for replicated logs and then for replicated states?

PR #16529, opened 2022-07-09
[CINFRA-479] Convergence waits for Leadership

maierlars:
### Scope & Purpose
The ReplicatedLog Supervision now waits for the leadership of the log to be established. Without this, a log might be declared converged although it is not yet and the first request to the leader might fail.


markuspf (approved):
LGTM

apetenchea (approved):

PR #16531, opened 2022-07-11
[CINFRA-470] Follower appendEntries

apetenchea:
### Scope & Purpose

This PR implements the `appendEntries` operation for the `DocumentFollowerState`. This means that from now on we'll be able to replicate transactions on replication2 databases. The following operations are supported:
- Insert
- Update
- Replace
- Remove
- Truncate
- Commit
- Abort

Additionally, some cluster-only tests are being taken out of the `shell-transaction.js` file and moved into a new file, `shell-transaction-cluster.js`. This is because importing the log helper on a single server causes a crash when the db servers are looked up.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [X] C++ **Unit tests**
  - [X] **integration tests**
  - [ ] **resilience tests**

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Transaction/SmartContext.h
```diff
@@ -79,7 +79,8 @@ struct TransactionContextSideUser {};
 struct ManagedContext final : public SmartContext {
   ManagedContext(TransactionId globalId,
```
apetenchea, line 80: We should implement our own context class.

arangod/RocksDBEngine/ReplicatedRocksDBTransactionCollection.cpp
```diff
@@ -152,11 +152,13 @@ bool ReplicatedRocksDBTransactionCollection::ensureSnapshot() {
...
 auto ReplicatedRocksDBTransactionCollection::leaderState() -> std::shared_ptr<
     replication2::replicated_state::document::DocumentLeaderState> {
-  return _leaderState;
+  // return _leaderState;
+  return collection()->waitForDocumentStateLeader();
 }
 
 auto ReplicatedRocksDBTransactionCollection::ensureCollection() -> Result {
   auto res = RocksDBTransactionCollection::ensureCollection();
+  return res;
```
apetenchea, line 161: Don't forget to fix this.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.h
```diff
@@ -38,6 +42,19 @@ struct DocumentFollowerState
...
+        : core(std::move(core)){};
+    [[nodiscard]] bool didResign() const noexcept { return core == nullptr; }
+
+    std::unique_ptr<DocumentCore> core;
+  };
+
+  Guarded<GuardedData, basics::UnshackledMutex> _guardedData;
+
+ public:
+  std::unordered_map<TransactionId, std::shared_ptr<transaction::Methods>>
```
apetenchea, line 57: Remove this

arangod/Transaction/SmartContext.h
apetenchea: Using `ReplicatedContext`.

arangod/Transaction/Options.cpp
```diff
@@ -136,6 +136,11 @@ void Options::fromVelocyPack(arangodb::velocypack::Slice const& slice) {
       isFollowerTransaction = value.getBool();
     }
 
+    value = slice.get("isReplication2Transaction");
```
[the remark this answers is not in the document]
apetenchea: Removed after refactoring.

arangod/RocksDBEngine/ReplicatedRocksDBTransactionCollection.cpp
apetenchea: The problem was caused by the way we handled dirty reads, it is fixed now.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -45,7 +50,53 @@ auto DocumentFollowerState::acquireSnapshot(ParticipantId const& destination,
...
+        case OperationType::kInsert:
+        case OperationType::kUpdate:
+        case OperationType::kReplace:
+        case OperationType::kRemove:
+        case OperationType::kTruncate:
+          fut = trx->apply(doc);
+          break;
+        case OperationType::kCommit:
+        case OperationType::kAbort:
+          fut = trx->finish(doc);
```
maierlars, line 81: How does that work? How do you distringuish between commit and abort?

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -45,7 +50,53 @@ auto DocumentFollowerState::acquireSnapshot(ParticipantId const& destination,
...
+          break;
+        default:
+          THROW_ARANGO_EXCEPTION(TRI_ERROR_TRANSACTION_DISALLOWED_OPERATION);
+      }
+
+      // TODO parallelize?
+      fut.wait();
```
maierlars, line 88: on the follower it should always be resolved immediately right?

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
apetenchea: We look at the operation type inside the finish function. There's a case in which we have to abort anyways, even if the operation is commit. I am going to refactor this now to include a bug fix.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
maierlars: If yes, then maybe add a `ADB_PROD_ASSERT(fut.isReady());`

arangod/Replication2/StateMachines/Document/DocumentCore.h
```diff
@@ -35,19 +35,22 @@ namespace arangodb::replication2::replicated_state::document {
...
+  auto getTransactionHandler()
+      -> std::shared_ptr<IDocumentStateTransactionHandler>;
+
  private:
   GlobalLogIdentifier _gid;
   DocumentCoreParameters _params;
   ShardID _shardId;
   std::shared_ptr<IDocumentStateAgencyHandler> _agencyHandler;
   std::shared_ptr<IDocumentStateShardHandler> _shardHandler;
+  std::shared_ptr<IDocumentStateTransactionHandler> _transactionHandler;
```
maierlars, line 54: I kinda disagree with this. The transaction handler contains the currently ongoing transactions and this **not** something that should be shared across multiple instances of the document state. For example, if a follower becomes a leader, this structure is still in place. It should be located at in the follower state. Hence it is destroyed and thus all the transactions are gone as well.

The purpose of the core is to hold information that is supposed to be shared across instances. Ongoing transactions are not part of that.

apetenchea: Thanks, that's some good insight. I am currently addressing this in an upcoming commit.

maierlars (approved):

arangod/Replication2/StateMachines/Document/DocumentStateStrategy.cpp
```diff
@@ -133,4 +140,142 @@ auto DocumentStateShardHandler::createLocalShard(
...
+      return Result{
+          TRI_ERROR_TRANSACTION_INTERNAL,
+          fmt::format("Transaction of type {} with ID {} could not be applied",
+                      to_string(entry.operation), entry.tid.id())};
+  }
+
+  return std::move(fut).thenValue(
+      [self = shared_from_this()](OperationResult&& opRes) {
+        return opRes.result;
+      });
```
mpoeter, line 191: In FollowerState we have the assertion that the Future isReady anyway, so this could be simplified to:
```
TRI_ASSERT(fut.isReady());
return std::move(fut).get().result;
```

mpoeter (approved):

arangod/Replication2/StateMachines/Document/DocumentStateStrategy.cpp
apetenchea: Will be addressed in https://github.com/arangodb/arangodb/pull/16650

PR #16534, opened 2022-07-12
[CINFRA] BugFix Replicated Log Maintenance Race

maierlars:
### Scope & Purpose
Since the maintenance can create multiple actions it can happen that two replicated log actions are scheduled and both try to update a replicated log become the leader. Previously both succeeded, which resulted in the leader to be recreated. The replicated log was modified to only allow new replicated logs to be created with a higher term.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/ReplicatedLog.cpp
+++ b/arangod/Replication2/ReplicatedLog/ReplicatedLog.cpp
@@ -79,7 +79,7 @@ auto replicated_log::ReplicatedLog::becomeLeader(
   auto [leader, deferred] = std::invoke([&] {
     std::unique_lock guard(_mutex);
     if (auto currentTerm = _participant->getTerm();
-        currentTerm && *currentTerm > newTerm) {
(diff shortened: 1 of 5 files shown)


arangod/Replication2/ReplicatedLog/ReplicatedLog.cpp
```diff
@@ -79,7 +79,7 @@ auto replicated_log::ReplicatedLog::becomeLeader(
   auto [leader, deferred] = std::invoke([&] {
     std::unique_lock guard(_mutex);
     if (auto currentTerm = _participant->getTerm();
-        currentTerm && *currentTerm > newTerm) {
+        currentTerm && *currentTerm >= newTerm) {
       LOG_CTX("b8bf7", INFO, _logContext)
           << "tried to become leader with term " << newTerm
           << ", but current term is " << *currentTerm;
```
goedderz, line 85: Do you additionally want to assert `*currentTerm < newTerm` in maintainer mode?

goedderz (approved):

PR #16535, opened 2022-07-12
[PREGEL] APM-538 add reporting for running state observables

jvolmer:
Report observables in running state per gss.


jvolmer:
TODO merge parent first

PR #16538, opened 2022-07-12
we have to mimic the instanceInfo structure better in order to work with the crash-utils

dothebart:
### Scope & Purpose

When creating instance.js the crash-utils should get the structure of this object for better visualisation.
This verry case `pu.executeAndWait` would invoke this with a fake structure without this method. Here we add a dummy. 

- [x] :hankey: Bugfix


adityamukho (approved):
LGTM

PR #16539, opened 2022-07-12
[PREGEL] APM-538 Add loading state

jvolmer:
Add loading state that is used when graph is loaded into memory


markuspf (changes requested):
This needs a CHANGELOG entry.

PR #16547, opened 2022-07-14
[CINFRA-475] Test replicated state change during transaction

goedderz:
### Scope & Purpose

Changing the replicated state during a transaction should cause it to fail. Currently, it triggers an assertion, which it should not.

There is some test code moved around so the test helpers can be used on both client and server more uniformly.

- [X] :test_tube: Test


tests/js/client/replication2/replication2-cluster.js
```diff
@@ -180,6 +180,10 @@ function ddlSuite() {
 
     testCreateDatabaseWithNewReplication: function () {
       const dbName = 'replv2db';
+      const server = require('@arangodb/test-helper').getDBServers()[0].id;
+      const url = require("@arangodb/testutils/replicated-logs-helper").getServerUrl(server);
+      console.error({server, url});
```
maierlars, line 185: do you want to keep this?

maierlars (approved):

PR #16549, opened 2022-07-14
don't die on issues reading logfile lines

dothebart:
### Scope & Purpose

It seems we can see errors while reading logfiles:
```
TypeError: bad argument
    at instance.readAssertLogLines (/work/ArangoDB/js/client/modules/@arangodb/testutils/instance.js:518:26)
```
This PR try/catches them.

- [x] :hankey: Bugfix


Dronplane (approved):
LGTM

js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -509,24 +509,29 @@ class instance {
...
+              // else the server has already printed these:
+              print("ERROR: " + line);
+            }
+            this.assertLines.push(line);
           }
-          this.assertLines.push(line);
         }
       }
+    } catch (ex) {
+      print("failed to read " + this.logFile + " -> " + ex);
```
adityamukho, line 533: No way to print to `stderr`?

js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -509,24 +509,29 @@ class instance {
...
-        const line = buf.asciiSlice(lineStart, j);
-        lineStart = j + 1;
-        
-        // scan for asserts from the crash dumper
-        if (line.search('{crash}') !== -1) {
-          if (!IS_A_TTY) {
-            // else the server has already printed these:
-            print("ERROR: " + line);
+      for (let j = 0; j < maxBuffer; j++) {
+        if (buf[j] === 10) { // \n
```
adityamukho, line 518: Could this be better handled than hard-coding the ASCII character value for LF?

js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -509,24 +509,29 @@ class instance {
...
+              print("ERROR: " + line);
+            }
+            this.assertLines.push(line);
           }
-          this.assertLines.push(line);
         }
       }
+    } catch (ex) {
+      print("failed to read " + this.logFile + " -> " + ex);
+      this.assertLines.push("failed to read " + this.logFile + " -> " + ex);
```
adityamukho, line 534: `this.assertLines` seems to contain both success (ie, successfully reading a crash line from the log) and error messages (could not read line or file), without a clear key to help distinguish one kind from the other. Wondering if this could be problematic.

js/client/modules/@arangodb/testutils/instance.js
dothebart: using \n literally would not work with v8-buffers.

js/client/modules/@arangodb/testutils/instance.js
dothebart: I don't think we implemented this.

adityamukho (approved):
LGTM

PR #16551, opened 2022-07-14
[PREGEL] Delete workers only via garbage collection

jvolmer:
PREGEL_RESULT can be used in AQL to get result from a finished Pregel
run. For this to work, the conductor and workers of this Pregel run
still need to exist. A garbadge collection exists for Pregel which
runs every 20 seconds and only deletes Pregel runs (conductor and
workers) if their TTL is exceeded.
Bug: Directly after the Pregel run finished, its workers were deleted
without checking the TTL. This PR changes that such that only the garbage
collection deletes workers.


markuspf (approved):
LGTM

jvolmer:
It's a feature, not a bug!

PR #16559, opened 2022-07-15
work around BTS-903: turn of logging for the queries facility.

dothebart:
### Scope & Purpose

Work around BTS-903

- [x] :hankey: Bugfix


KVS85 (approved):
LGTM.

PR #16560, opened 2022-07-15
[PREGEL] Add aql pregel result test

jvolmer:
After finding out that the bug fixed in https://github.com/arangodb/arangodb/pull/16551 was a feature and cancelling this PR, this PR adds some tests to check this feature. Documentation improvement: https://github.com/arangodb/docs/pull/1062.

### DIFF
--- a/tests/js/common/shell/shell-pregel-basics-example.js
+++ b/tests/js/common/shell/shell-pregel-basics-example.js
@@ -261,6 +261,32 @@ function basicTestSuite() {
       if (i === 0) {
         assertTrue(false, "timeout in pregel execution");
       }
+    },
+
+    test_AQL_pregel_result_is_empty_when_pregel_results_are_stored_in_database: function () {
+      var pid = pregel.start("pagerank", graphName, { threshold: EPS / 10, store: true });
+      var i = 10000;
+      do {
+        internal.sleep(0.2);
+        var stats = pregel.status(pid);
+        if (stats.state === "done") {
+          assertEqual(stats.vertexCount, 11, stats);
+          assertEqual(stats.edgeCount, 17, stats);
+
+          let vertices = db._collection(vColl);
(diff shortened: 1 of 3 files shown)


tests/js/common/shell/shell-pregel-basics-example.js
```diff
@@ -230,6 +235,58 @@ function basicTestSuite() {
       if (i === 0) {
         assertTrue(false, "timeout in pregel execution");
       }
+    },
+
+    test_AQL_pregel_result_is_empty_after_ttl_expires: function () {
```
jvolmer, line 240: not sure if this test should be in here because it sleeps more than 25s which extends the full jenkins runtime. But it documents the behavior and the config parameter 'ttl', therefore I thought it could be good to have it.

tests/js/common/shell/shell-pregel-basics-example.js
```diff
@@ -230,6 +235,58 @@ function basicTestSuite() {
...
+      var pid = pregel.start("pagerank", graphName, { threshold: EPS / 10, store: false, ttl: 1 });
+      var i = 10000;
+      do {
+        internal.sleep(0.2);
+        var stats = pregel.status(pid);
+        if (stats.state === "done") {
+          assertEqual(stats.vertexCount, 11, stats);
+          assertEqual(stats.edgeCount, 17, stats);
+
+          // garbade collection runs every 20s, therefore we should wait at least that long plus the ttl given
```
markuspf, line 251:
```suggestion
          // garbage collection runs every 20s, therefore we should wait at least that long plus the ttl given
```

markuspf (approved):
Looks good (modulo typo.)

jvolmer:
My previous fix did not work on the cluster. What you actually want - and I just implemented: If you store the results, PREGEL_RESULT should always give back an empty array.

markuspf:
This should probably be backported to 3.10?

jvolmer:
Backport PR: https://github.com/arangodb/arangodb/pull/16647

PR #16562, opened 2022-07-15
Let arangoimport also ignore Attributes on JSON input

mchacki:
### Scope & Purpose

For the time being --remove-attribute option was limited to CSV and TSV only.
This is a bit inconvenient as the default is JSON.
This PR enables the option on JSOn as well.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


client-tools/Import/ImportFeature.cpp
```diff
@@ -165,7 +165,7 @@ void ImportFeature::collectOptions(
 
   options->addOption("--remove-attribute",
                      "remove an attribute before inserting documents"
-                     " into collection (for CSV and TSV only)",
+                     " into collection (for CSV, TSV and JSON only)",
```
Simran-B, line 168: JSON, JSONL, or both? If both, what format is left?

client-tools/Import/ImportHelper.cpp
```diff
@@ -590,6 +599,253 @@ bool ImportHelper::importJson(std::string const& collectionName,
...
+  }
+
+  // read and convert
+  int64_t totalLength;
+  std::unique_ptr<arangodb::ManagedDirectory::File> fd;
+
+  if (fileName == "-") {
+    // we don't have a filesize
+    totalLength = 0;
+    fd = directory.readableFile(STDIN_FILENO);
```
Simran-B, line 632: Does this work on Windows?

dothebart (approved):

Simran-B (commented):
Can it remove top-level attributes only? If so, that should potentially be mentioned. If it can remove nested attributes, it treat the `.` to indicate sub-attributes?

KVS85 (changes requested):
Need to have an integration test for this option.

mchacki:
@KVS85 there is an integration test here: js/client/modules/@arangodb/testsuites/importing.js

client-tools/Import/ImportHelper.cpp
mchacki: I hope so, test will show.
Anyways this is taken from the other import mechanism, if it does not work we need a separate bugfix for this in general

client-tools/Import/ImportFeature.cpp
mchacki: JSON and JSONL,
left is XML

mchacki:
@Simran-B only top-level
This has been designed for CSV/TSV imports where there is no nesting.

I did not want to modify this feature.

KVS85 (approved):
LGTM.

joerg84 (approved):

client-tools/Import/ImportFeature.cpp
Simran-B: We don't support importing XML. It's only available in arangoexport.

client-tools/Import/ImportHelper.cpp
Simran-B: I'm just surprised because I remember some ifdefs to exempt Windows from using stdin/stout in some client-tool code, but it may have changed since or it was always limited to just a few corner cases. I found the following:
- Handling of Windows' UTF-16 encoding
  https://github.com/arangodb/arangodb/blob/devel/client-tools/Shell/ShellConsoleFeature.cpp#L335-L344
- Using pipes is Linux-only in the VPack util
  https://github.com/arangodb/arangodb/blob/devel/client-tools/VPack/VPackFeature.cpp#L134-L147

PR #16566, opened 2022-07-18
QA-273: fall forward the shutdown procedure to 'force' if we reach the global deadline

dothebart:
### Scope & Purpose

once the global deadline is reached, instance manager should not be friendly to SUT anymore. 
Hence `forceterminate` is engaged. In case of `forceterminate` we also should no longer talk to instances.
We also reset the deadline, so we get to know whether we reached it, and become able to launch debuggers again.
(subprocess handling will throw with a triggered deadline - hence we need to push the pin back in.)

- [x] :pizza: New feature


KVS85 (approved):
LGTM.

PR #16414, opened 2022-06-15
[CINFRA-418] Leader State replicates modifications

apetenchea:
### Scope & Purpose

Document Leader State handles replication for replication2 databases.
The following methods were adapted to use the new replication:
- insertLocal
- modifyLocal
- removeLocal
- truncateLocal

Due to collection creating taking longer, the `shell-transaction.js` test was taking too long. We fixed that by moving it into a separate folder and changing the supervision frequency setting. In order not to separate related test suites, all transaction tests were moved into the `shell/transaction` folder.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*
  - [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #16568, opened 2022-07-18
[CINFRA-492] Add suite shell_client_transaction

goedderz:
### Scope & Purpose

Added test suite `shell_client_transaction` to the CI matrix. Was added (and some files moved there) in https://github.com/arangodb/arangodb/pull/16414/files#diff-c7ccb564e3e90efc9e6dd3d9e620cb037a3f7eada565c9615564303a9026e292.

- [X] :hankey: Bugfix

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -1059,7 +1059,8 @@ std::shared_ptr<TransactionState> RocksDBEngine::createTransactionState(
     TRI_vocbase_t& vocbase, TransactionId tid,
     transaction::Options const& options) {
   if (vocbase.replicationVersion() == replication::Version::TWO &&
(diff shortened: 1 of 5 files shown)


dothebart:
do you think arangosh accessing it via different means of communications can make a difference?

goedderz:
> do you think arangosh accessing it via different means of communications can make a difference?

The tests were previously part of `shell_client`, so I used the same settings. I don't know the tests in detail, so I'd rather be on the safe side unless someone checks and assures otherwise.

dothebart:
from what I see in the linked PR this will allways be a clustered setup? Then in general it doesn't make sense to run it in single server since the load estimates will be wrong and destabilize everything.

dothebart:
ok, maybe the linked commit isn't the only change there.

maierlars (approved):

dothebart (approved):

goedderz:
Jenkins failure is unrelated.

PR #16576, opened 2022-07-20
Workaround Compiler Error ARM+GCC+LTO

maierlars:
### Scope & Purpose
Work around internal compiler error with GCC on ARM and LTO.


markuspf (approved):
It's ok.

joerg84 (approved):

PR #16587, opened 2022-07-21
Made commit/abort/finish nodiscard

goedderz:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Graph/GraphManager.cpp
+++ b/arangod/Graph/GraphManager.cpp
@@ -311,22 +311,19 @@ bool GraphManager::graphExists(std::string const& graphName) const {
 
   Result res = trx.begin();
 
-  if (!res.ok()) {
-    return false;
+  if (res.fail()) {
+    THROW_ARANGO_EXCEPTION(res);
   }
 
   OperationOptions options;
-  try {
-    OperationResult checkDoc = trx.document(StaticStrings::GraphCollection,
-                                            checkDocument.slice(), options);
-    if (checkDoc.fail()) {
-      trx.finish(checkDoc.result);
-      return false;
-    }
-    trx.finish(checkDoc.result);
-  } catch (...) {
+  OperationResult checkDoc = trx.document(StaticStrings::GraphCollection,
+                                          checkDocument.slice(), options);
+  res = trx.finish(checkDoc.result);
+  if (res.fail()) {
+    THROW_ARANGO_EXCEPTION(res);
   }
-  return true;
+
+  return checkDoc.ok();
 }
 
 ResultT<std::unique_ptr<Graph>> GraphManager::lookupGraphByName(
@@ -448,7 +445,7 @@ OperationResult GraphManager::storeGraph(Graph const& graph, bool waitForSync,
                                                  builder.slice(), options);
 
   if (!result.ok()) {
-    trx.finish(result.result);
+    std::ignore = trx.finish(result.result);
     return result;
   }
   res = trx.finish(result.result);
--- a/arangod/Graph/GraphOperations.cpp
+++ b/arangod/Graph/GraphOperations.cpp
@@ -781,10 +781,16 @@ OperationResult GraphOperations::validateEdgeVertices(
 
   // actual result doesn't matter here
   if (!resultFrom.ok()) {
-    trx.finish(resultFrom.result);
+    auto res = trx.finish(resultFrom.result);
+    if (res.fail()) {
+      THROW_ARANGO_EXCEPTION(res);
+    }
     return OperationResult(TRI_ERROR_ARANGO_DOCUMENT_NOT_FOUND, options);
   } else if (!resultTo.ok()) {
-    trx.finish(resultTo.result);
+    auto res = trx.finish(resultTo.result);
+    if (res.fail()) {
+      THROW_ARANGO_EXCEPTION(res);
+    }
     return OperationResult(TRI_ERROR_ARANGO_DOCUMENT_NOT_FOUND, options);
   }
   return OperationResult(TRI_ERROR_NO_ERROR, options);
--- a/arangod/Transaction/Manager.cpp
+++ b/arangod/Transaction/Manager.cpp
@@ -200,7 +200,8 @@ Manager::ManagedTrx::~ManagedTrx() {
                   opts);  // own state now
     TRI_ASSERT(trx.state()->status() == transaction::Status::RUNNING);
     TRI_ASSERT(trx.isMainTransaction());
-    trx.abort();
+    // TODO may this be ignored?
+    std::ignore = trx.abort();
   } catch (...) {
     // obviously it is not good to consume all exceptions here,
     // but we are in a destructor and must never throw from here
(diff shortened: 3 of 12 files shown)


arangod/Graph/GraphManager.cpp
```diff
@@ -311,22 +311,16 @@ bool GraphManager::graphExists(std::string const& graphName) const {
...
-    }
-    trx.finish(checkDoc.result);
-  } catch (...) {
-  }
-  return true;
+  OperationResult checkDoc = trx.document(StaticStrings::GraphCollection,
+                                          checkDocument.slice(), options);
+  std::ignore = trx.finish(checkDoc.result);
+
+  return checkDoc.ok();
```
jsteemann, line 323: I think this can be simplified to just
```suggestion
  return trx.finish(checkDoc.result).ok();
```
This will slightly change the existing behavior, as now a failing "finish" call can also lead to a result of "false" being returned. This would only happen in case of error, but is probably even justified.

arangod/Graph/GraphManager.cpp
```diff
@@ -448,7 +442,7 @@ OperationResult GraphManager::storeGraph(Graph const& graph, bool waitForSync,
                                                  builder.slice(), options);
 
   if (!result.ok()) {
-    trx.finish(result.result);
+    std::ignore = trx.finish(result.result);
```
jsteemann, line 445: I would be in favor or merging this failure block with the block below.
There should only be one unconditional call to finish with `result.result` as its argument.

arangod/Graph/GraphOperations.cpp
```diff
@@ -781,10 +781,10 @@ OperationResult GraphOperations::validateEdgeVertices(
 
   // actual result doesn't matter here
   if (!resultFrom.ok()) {
-    trx.finish(resultFrom.result);
+    std::ignore = trx.finish(resultFrom.result);
```
jsteemann, line 784: I would be in favor of simplifying this code here. IMO there is no need to finish the transaction here at all in case of failure.
If a transaction object goes out of scope without having been committed or aborted, it will auto-abort anyway. And read-only transactions do not need to be committed, so for them there doesn't need to be an explicit call to finish.
So we could remove the calls to `finish` here entirely, and have the one call-site handle the trx object properly (it already does).

arangod/Transaction/Methods.h
```diff
@@ -206,20 +206,20 @@ class Methods {
   Result begin();
 
   /// @deprecated use async variant
-  auto commit() noexcept -> Result;
+  [[nodiscard]] auto commit() noexcept -> Result;
```
MBkkt, line 209: Maybe also add deprecated?
```suggestion
  [[nodiscard, deprecated("use async variant")]] auto commit() noexcept -> Result;
```

arangod/Graph/GraphOperations.cpp
```diff
@@ -781,10 +781,10 @@ OperationResult GraphOperations::validateEdgeVertices(
...
     return OperationResult(TRI_ERROR_ARANGO_DOCUMENT_NOT_FOUND, options);
   } else if (!resultTo.ok()) {
-    trx.finish(resultTo.result);
+    std::ignore = trx.finish(resultFrom.result);
```
jsteemann, line 787: Same here.

arangod/Transaction/Methods.h
```diff
@@ -206,20 +206,20 @@ class Methods {
...
   /// @deprecated use async variant
-  auto abort() noexcept -> Result;
+  [[nodiscard]] auto abort() noexcept -> Result;
   /// @brief abort the transaction
-  auto abortAsync() noexcept -> Future<Result>;
+  [[nodiscard]] auto abortAsync() noexcept -> Future<Result>;
 
   /// @deprecated use async variant
-  auto finish(Result const& res) noexcept -> Result;
+  [[nodiscard]] auto finish(Result const& res) noexcept -> Result;
```
MBkkt, line 219:
```suggestion
  [[nodiscard, deprecated("use async variant")]] auto finish(Result const& res) noexcept -> Result;
```

arangod/V8Server/v8-collection.cpp
```diff
@@ -1076,9 +1076,9 @@ static void JS_FiguresVocbaseCol(
   OperationOptions options(ExecContext::current());
   auto opRes = collection->figures(details, options).get();
 
-  trx.finish(TRI_ERROR_NO_ERROR);
+  res = trx.finish(TRI_ERROR_NO_ERROR);
```
jsteemann, line 1079: In order to simplify things here, I think we don't need the call to `finish` here at all. It is a read-only transaction anyway, so nothing needs to be explicitly committed or aborted.

goedderz: I haven't removed it entirely, but still simplified it a little.

jsteemann (approved):
In general LGTM.
I am not sure about marking the synchronous methods as "deprecated", because that will probably produce several compile warnings which we will need to ignore for now.
If those warnings don't make the build fail, it is ok for me.

goedderz:
> In general LGTM. I am not sure about marking the synchronous methods as "deprecated", because that will probably produce several compile warnings which we will need to ignore for now. If those warnings don't make the build fail, it is ok for me.

The deprecated warnings don't trigger errors in our cmake configuration.

Generally, I'd like to be able to mark deprecated things as deprecated. If we don't want the warnings that come with them on all builds (sensible), then I'd rather turn them off by default than not use `[[deprecated]]` to avoid the warnings.

goedderz:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/22754/
failures are unrelated

goedderz:
@MBkkt are you otherwise satisfied with the changes, in particular the ones regarding IResearch?

MBkkt (approved):

arangod/Transaction/Manager.cpp
```diff
@@ -200,7 +200,8 @@ Manager::ManagedTrx::~ManagedTrx() {
                   opts);  // own state now
     TRI_ASSERT(trx.state()->status() == transaction::Status::RUNNING);
     TRI_ASSERT(trx.isMainTransaction());
-    trx.abort();
+    // TODO may this be ignored?
```
mchacki, line 203: I do not really see an option here, besides logging.
We are on the destructor of the transaction.

mchacki (approved):
LGTM, makes the world better again

arangod/Transaction/Manager.cpp
goedderz, line 203:
```suggestion
```

maierlars (approved):

PR #16591, opened 2022-07-22
[CINFRA-493] Missing Stream if snapshot complete

maierlars:
### Scope & Purpose
This PR fixes two things:
1. The prototype states `applyEntries` is declared as noexcept but wasn't. This is because `getStream` could throw an exception. 
2. This pointed at another problem: the follower state manager did not properly initialise the stream reference on the state after it was constructed **and** the snapshot was already available. Added tests for that.

### DIFF
--- a/tests/Replication2/ReplicatedState/FollowerSnapshotTest.cpp
+++ b/tests/Replication2/ReplicatedState/FollowerSnapshotTest.cpp
@@ -110,6 +110,10 @@ TEST_F(FollowerSnapshotTest, basic_follower_manager_test) {
   ASSERT_EQ(nullptr, manager->getFollowerState())
       << "follower state should not be available yet";
 
+  // furthermore the state should have access to the stream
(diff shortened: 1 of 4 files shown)


tests/Replication2/ReplicatedState/FollowerSnapshotTest.cpp
```diff
@@ -110,6 +110,10 @@ TEST_F(FollowerSnapshotTest, basic_follower_manager_test) {
   ASSERT_EQ(nullptr, manager->getFollowerState())
       << "follower state should not be available yet";
 
+  // furthermore the state should have access to the stream
```
apetenchea, line 113:
```suggestion
  // furthermore the state should not have access to the stream
```

apetenchea (approved):
I can modify that comment in an upcoming PR, such that we leave Jenkins alone.

PR #16592, opened 2022-07-22
fix tests; need to load the UUID from the file

dothebart:
### Scope & Purpose

`resilience-synchronous-repl-failureAt-cluster.js` would be broken by the [instance rework](https://github.com/arangodb/arangodb/pull/16566). this fixes it.

- [x] :hankey: Bugfix


KVS85 (approved):
LGTM.

PR #16593, opened 2022-07-25
BTS-913: Abort startup if timezone is not configured properly.

dothebart:
### Scope & Purpose

instead of catching an error at AQL-Runtime, abort on startup if time zone configuration isn't working

- [x] :hankey: Bugfix
- [ ] Backports: 3.8 3.9 3.10

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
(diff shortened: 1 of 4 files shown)


jsteemann (approved):
LGTM

PR #16597, opened 2022-07-25
Fix potential deadlock in rocksdb compaction.

mpoeter:
### Scope & Purpose

There can be situations where the picker is asked for a compaction twice for the same version. In such a scenario the first call might not be able to find any eligible files and as a result set NextCompactionIndex to the total number of files. In this case the second call could return null even though we could now compact one of the files we previously had to skip. This could then result in a deadlock as explained in more detail here: https://github.com/facebook/rocksdb/issues/10257#issuecomment-1180534048

The according RocksDB PR can be found here: https://github.com/facebook/rocksdb/pull/10355

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] C++ **Unit tests** - I created a unit test for the rocksdb PR, but I did not port it back here.
- [x] Backports
  - [x] Backport for 3.9: #16598


jsteemann (approved):
LGTM!

joerg84 (approved):

jsteemann:
An upstream test has shown a potential problem:
https://circleci.com/gh/facebook/rocksdb/445331?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
Setting this PR's status to draft until clarified.

PR #16602, opened 2022-07-26
Cleanout server before testing rebalance.

maierlars:
### Scope & Purpose
Before running the rebalance tests, make sure the cluster is imbalanced by cleaning out on of the servers.


dothebart:
is this a more general utility function that could be better put into test-helper? or is this very specific to this very test?

hkernbach (approved):

KVS85 (approved):
LGTM.

PR #16605, opened 2022-07-26
[PREGEL] Add Conductor state machine and use it for the Loading state

jvolmer:
Adds a finite state machine to the conductor to improve readability: The behavior of the conductor will now be determined by the conductor state. Before, the conductor only had a enum type `_state`, which had to be checked many times. Now state changes will be defined inside a specific conductor state and can happen when the state runs or gets a message.
This is only the first PR concerning the state machine: It adds only the LoadingState and AnotherState which is just a placeholder for the following state - it will be removed later on. This PR will not break anything, the old enum type `_state` will be deleted once all conductor states are added.

### DIFF
--- a/arangod/Pregel/Conductor.cpp
+++ b/arangod/Pregel/Conductor.cpp
@@ -22,6 +22,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include <chrono>
+#include <memory>
 #include <string_view>
 #include <thread>
 
@@ -33,6 +34,8 @@
 #include "Pregel/Aggregator.h"
 #include "Pregel/AlgoRegistry.h"
 #include "Pregel/Algorithm.h"
+#include "Pregel/Conductor/State.h"
+#include "Pregel/Conductor/LoadingState.h"
 #include "Pregel/WorkerConductorMessages.h"
 #include "Pregel/MasterContext.h"
 #include "Pregel/PregelFeature.h"
@@ -165,26 +168,12 @@ Conductor::~Conductor() {
 void Conductor::start() {
   MUTEX_LOCKER(guard, _callbackMutex);
   _timing.total.start();
-  _timing.loading.start();
-
-  _globalSuperstep = 0;
-
-  updateState(ExecutionState::LOADING);
-  _feature.metrics()->pregelConductorsLoadingNumber->fetch_add(1);
-
-  LOG_PREGEL("3a255", DEBUG) << "Telling workers to load the data";
-  auto res = _initializeWorkers(Utils::startExecutionPath, VPackSlice());
-  if (res != TRI_ERROR_NO_ERROR) {
-    updateState(ExecutionState::CANCELED);
-    _feature.metrics()->pregelConductorsRunningNumber->fetch_sub(1);
-    LOG_PREGEL("30171", ERR) << "Not all DBServers started the execution";
-  }
+  changeState(conductor::StateType::Loading);
 }
 
 // only called by the conductor, is protected by the
 // mutex locked in finishedGlobalStep
 bool Conductor::_startGlobalStep() {
-  updateState(ExecutionState::RUNNING);
   if (_feature.isStopping()) {
     THROW_ARANGO_EXCEPTION(TRI_ERROR_SHUTTING_DOWN);
   }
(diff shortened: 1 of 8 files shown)


jvolmer:
TODO merge parent first

arangod/Pregel/Conductor.cpp
```diff
@@ -1141,3 +1122,15 @@ void Conductor::updateState(ExecutionState state) {
     _expires = std::chrono::system_clock::now() + _ttl;
   }
 }
+
+auto Conductor::changeState(conductor::StateType name) -> void {
```
romanatarango, line 1126: For me, `changeState` sounds as if only the state is changed. However, it is also produced and `run()` is performed. Shouldn't these three things be separated?

jvolmer: The states are designed such that their run() method should be executed as soon as a state-change happened.
Like this the conductor is either running (executes the state->run() method) or waits for a message from a worker which will change the state in the state->receive method and then triggers state->run() on the new state again.
Yes, in this state-machine the states are created just when they are triggered. I chose to do that because the conductor reaches each state only once - except if a recovery happens, but I assume this will not happen too often. That this state is created is a detail which is not relevant for the caller.

markuspf (changes requested):
Please name header files `.h` not `.hpp` for consistency

jvolmer:
Renamed header files and fixed a bug: Inside Loading::receive(Message) we collect messages from all Workers and change the state only when all servers responded. This will be refactored when all state are there, currently it is basically copied from the conductor into the state.

arangod/Pregel/Conductor/LoadingState.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "Metrics/Gauge.h"
+#include "Pregel/Conductor/State.h"
+#include "Pregel/PregelFeature.h"
+#include "Pregel/WorkerConductorMessages.h"
+
+using namespace arangodb::pregel::conductor;
+
+#define LOG_PREGEL(logId, level)          \
+  LOG_TOPIC(logId, level, Logger::PREGEL) \
+      << "[job " << conductor._executionNumber << "] "
```
markuspf, line 14: Isn't this something that should go into a common header file for Pregel?

arangod/Pregel/Conductor/LoadingState.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+Loading::Loading(Conductor& conductor) : conductor{conductor} {
+  conductor.updateState(ExecutionState::LOADING);
```
markuspf, line 17: Is this going to be removed again later? This looks like we have 2 different states now.

arangod/Pregel/Conductor/LoadingState.cpp
```diff
@@ -0,0 +1,58 @@
...
+  conductor._timing.loading.start();
+  conductor._feature.metrics()->pregelConductorsLoadingNumber->fetch_add(1);
+}
+
+Loading::~Loading() {
+  conductor._timing.loading.finish();
```
markuspf, line 23: How does this destructor interact with conductor's `ExecutionState`?

arangod/Pregel/Conductor/LoadingState.cpp
```diff
@@ -0,0 +1,58 @@
...
+    LOG_PREGEL("30171", ERR) << "Not all DBServers started the execution";
+    // TODO change state to Canceled
+    conductor.changeState(StateType::Placeholder);
+  }
+}
+
+auto Loading::receive(Message const& message) -> void {
+  if (message.type() != MessageType::GraphLoaded) {
+    LOG_PREGEL("14df4", WARN)
+        << "When loading, we expect a GraphLoaded message, nothing else";
```
markuspf, line 41: Maybe we want to say here which message we received? that might ease debugging quite a bit.

markuspf (approved):
Some questions

arangod/Pregel/Conductor/LoadingState.cpp
jvolmer: Yes, definitely, the ExecutionState will be removed when the refactoring is done. Currently we still need it at some places.

arangod/Pregel/Conductor/LoadingState.cpp
jvolmer: It does not interact at all. The state is destructed when conductor.changeState is called. In the constructor of the new state, ExecutationState is set to the new state. But we want to get rid of ExecutionState anyway.

arangod/Pregel/Conductor/LoadingState.cpp
jvolmer: ah, good point. It can only be used for the conductor states, so I will put it in the conductor namespace in Conductor/State.h

markuspf: maybe it shouldn't be called `LOG_PREGEL` then, but `LOG_PREGEL_CONDUCTOR`?

PR #16614, opened 2022-07-26
Add README for type inspection

mpoeter:
Add a README with some documentation for the type inspection library.

### DIFF
--- a/lib/Inspection/README.md
+++ b/lib/Inspection/README.md
@@ -1,0 +1,375 @@
+# Type inspection
+
+At the moment this library is rather tightly coupled with VelocyPack since the
+primary goal was to simplify (de)serialization from and to VelocyPack. It is
+planned to reduce this coupling so that the Inspector concept can also be used
+in other scenarios (e.g., deserializing from agency nodes).
+
+## Acknowledgements
+
+This library is heavily inspired by the type inspection library from the C++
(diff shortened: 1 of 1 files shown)


lib/Inspection/README.md
```diff
@@ -0,0 +1,375 @@
...
+  array types.
+- maps
+  Dynamically-sized container types with key/value pairs such as std::map.
+  At the moment we only support strings as keys.
+- objects
+  User-defined types. An object has one or more fields. Fields have a name
+  and may be optional. Further, fields may take on a fixed number of different
+  types.
+
+There is also support for `std::optional` and `std::variant`, as well as
```
apetenchea, line 33: Should we also mention `velocypack::SharedSlice` in here? It is a safe type.

lib/Inspection/README.md
```diff
@@ -0,0 +1,375 @@
...
+      value(type: "uint64_t") {
+        ...
+      }
+    }
+  }
+```
+
+Now, this type has little use on its own and only introduces unnecessary
+overhead. We do not need the additional object wrapper and could make it
+completely transparent to inspectors. This an be achieved by writing, the
```
apetenchea, line 106:
```suggestion
completely transparent to inspectors. This an be achieved by writing the
```

apetenchea (approved):

PR #16620, opened 2022-07-27
[PREGEL] Add cancel state to conductor state machine

jvolmer:
Third PR for adding a state machine to the pregel conductor.
Adds canceled state to conductor state machine

### DIFF
--- a/arangod/Pregel/Conductor.cpp
+++ b/arangod/Pregel/Conductor.cpp
@@ -37,6 +37,7 @@
 #include "Pregel/Conductor/State.h"
 #include "Pregel/Conductor/LoadingState.h"
 #include "Pregel/Conductor/StoringState.h"
+#include "Pregel/Conductor/CanceledState.h"
 #include "Pregel/WorkerConductorMessages.h"
 #include "Pregel/MasterContext.h"
 #include "Pregel/PregelFeature.h"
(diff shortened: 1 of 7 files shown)


jvolmer:
TODO merge parent first

arangod/Pregel/Conductor.cpp
```diff
@@ -428,10 +429,6 @@ VPackBuilder Conductor::finishedWorkerStep(VPackSlice const& data) {
 
     if (_state == ExecutionState::RUNNING) {
       _startGlobalStep();  // trigger next superstep
-    } else if (_state == ExecutionState::CANCELED) {
-      LOG_PREGEL("dd721", WARN)
-          << "Execution was canceled, results will be discarded.";
-      _finalizeWorkers();  // tells workers to store / discard results
```
jvolmer, line 434: If we get into the Canceled state, we do this anyway because it is part of the Canceled::run() method, which is called after a state change to this state.

arangod/Pregel/Conductor.cpp
```diff
@@ -840,42 +813,33 @@ void Conductor::finishedWorkerFinalize(VPackSlice data) {
...
-  if (_state == ExecutionState::CANCELED) {
-    auto* scheduler = SchedulerFeature::SCHEDULER;
-    if (scheduler) {
-      uint64_t exe = _executionNumber;
-      scheduler->queue(RequestLane::CLUSTER_AQL,
-                       [this, exe, self = shared_from_this()] {
-                         _feature.cleanupConductor(exe);
-                       });
-    }
+  if (_state != ExecutionState::CANCELED) {
```
jvolmer, line 816: this whole block is deleted in one of the next PRs, here just make sure that this does not run when we are in the Canceled State

PR #16627, opened 2022-07-27
[Pregel] Add recovering to conductor state machine

jvolmer:
Fifth PR for adding a state machine to the pregel conductor.
Adds InError and Recovering states to conductor state machine. Recovering is triggered from outside and is just allowed when state is Running or InError, therefore I created a new state-machine method recover(), which starts the Recovery state only from InError and Running (not yet implemented) state.

### DIFF
--- a/arangod/Pregel/Conductor/RecoveringState.cpp
+++ b/arangod/Pregel/Conductor/RecoveringState.cpp
@@ -1,0 +1,152 @@
+#include "RecoveringState.h"
+
+#include "Pregel/Algorithm.h"
+#include "Pregel/Conductor.h"
+#include "Pregel/Conductor/State.h"
+#include "Pregel/MasterContext.h"
+#include "Pregel/PregelFeature.h"
+#include "Pregel/Recovery.h"
+#include "Pregel/WorkerConductorMessages.h"
+
+using namespace arangodb::pregel::conductor;
+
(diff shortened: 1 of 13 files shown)


jvolmer:
TODO merge parent first

arangod/Pregel/Conductor/RecoveringState.cpp
```diff
@@ -0,0 +1,149 @@
...
+    return;
+  }
+
+  // we lost a DBServer, we need to reconfigure all remainging servers
+  // so they load the data for the lost machine
+  conductor._statistics.reset();
+
+  TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
+
+  // let's wait for a final state in the cluster
```
jvolmer, line 36: I'm really not sure why this has to be queued and cannot be directly executed here. Any ideas?
This queuing makes it quite hard: In line 60 we have to test again, if we are still in the Recovering state because due to stuff we did in the lines 37 to 59 we could have reached the Canceled state already. If we don't queue, we can be sure in which state we are.

markuspf: I have no immediate idea why this is the case.

I believe though that once we sorted out the state machine and interfaces to conductor and workers we will have an easier time determining which code should run asynchronously at which point we will sort these out with a fine comb.

markuspf (approved):

PR #16630, opened 2022-07-27
[PREGEL] Add fatal error state and getResults method to conductor state machine

jvolmer:
Sixth PR for adding a state machine to the pregel conductor.
- Adds FatalError to conductor state machine. 
- Pregel results are only given via the conductor state to AQL. This ensures that the results are only given if the conductor is in the Done or FatalError state (although I'm really not sure why FatalError - in my opinion results should only be available in the Done state).

### DIFF
--- a/arangod/Pregel/Conductor.cpp
+++ b/arangod/Pregel/Conductor.cpp
@@ -73,7 +73,7 @@
 #include <Inspection/VPack.h>
 #include <velocypack/Iterator.h>
 
-    using namespace arangodb;
+using namespace arangodb;
 using namespace arangodb::pregel;
 using namespace arangodb::basics;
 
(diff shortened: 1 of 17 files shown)


arangod/Pregel/Conductor.cpp
```diff
@@ -748,28 +744,7 @@ bool Conductor::canBeGarbageCollected() const {
 
 void Conductor::collectAQLResults(VPackBuilder& outBuilder, bool withId) {
   MUTEX_LOCKER(guard, _callbackMutex);
-
-  if (_state != ExecutionState::DONE && _state != ExecutionState::FATAL_ERROR) {
```
jvolmer, line 752: I'm really not sure why FatalError is here. I don't see any reason why results should be returned if the conductor is in a FatalError state. In my opinion this should not be the case, what do you think?
For now I just reproduced the behavior here and returned the results in the FatalError state but I can delete that easily.

jvolmer:
TODO merge parent first

arangod/Pregel/Conductor/CanceledState.h
```diff
@@ -33,10 +33,11 @@ namespace conductor {
 struct Canceled : State {
   Conductor& conductor;
   Canceled(Conductor& conductor);
-  ~Canceled();
+  ~Canceled(){};
```
markuspf, line 36:
```suggestion
  ~Canceled() = default;
```
?

PR #16634, opened 2022-07-27
[PREGEL] Bugfix: stop total pregel time correctly when not storing results

jvolmer:
When store=false, Done state directly follows Running state without
cleaning up. Cleanup just happens when run is canceled, therefore the
total timer ran until the Pregel run was canceled and not - as it
should - until the computation is done.
Now the total timer stops directly when entering Done state and not storing.

Additionally, correct memory label in status output


jvolmer:
Backport PR: https://github.com/arangodb/arangodb/pull/16648

markuspf (approved):
LGTM

PR #16554, opened 2022-07-14
Use std::semaphore for UnshackledMutex

MBkkt:
### Scope & Purpose

We couldn't use it early because of apple clang

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*
  - [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #16643, opened 2022-07-28
[3.10] Revert "Use std::semaphore for UnshackledMutex"

mpoeter:
Reverts arangodb/arangodb#16554

This is necessary because the libstdc++ and libc++ semaphore implementations are currently broken (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104928).

Backport of #16642


MBkkt (approved):

jsteemann (approved):
LGTM

PR #16653, opened 2022-07-29
[PREGEL] Add computing state to conductor state machine

jvolmer:
Sevent PR for adding a state machine to the pregel conductor: 
Adds the last state - the Computation state - to the state machine.

We have a problem with timings when recovering: If we recover, we attempt to start the computation timer again because we create the Computing state again, which would give an error in Duration. Therefore, in constructor and destructor of Computing it is checked if the timing was already set. This results that if we recover, the computation time would just be the time we spend in the computing state before recovering, not after. This is not correct but I am anyway not sure how we want to proceed with the recovery state, therefore I left it like this for now.


jvolmer:
TODO merge parent first

markuspf (approved):
LGTM

PR #16672, opened 2022-08-01
[PREGEL] Refactor/get rid of old state usages

jvolmer:
Get rid of almost all old state usages, this includes several small changes:
- Move expiration into state
- Give state a name
- Identify 'running' states where Pregel is active
- Add recovery to Computing state
- Delete unnecessary checks
- Add missing state changes to InError
- Add Initial state


jvolmer:
Is already merged into staging/pregel-refactor

PR #16687, opened 2022-08-02
[BUILD] Add CMakePresets.json

markuspf:
### Scope & Purpose

This PR adds `CMakePresets` which can be used to share settings for building ArangoDB between platforms and developers and make for a consistent build environment.

Example use

```
# cmake --list-presets
Available configure presets:

  "community"                       - Build ArangoDB Community Edition
  "enterprise"                      - Build ArangoDB Enterprise Edition
  "community-developer"             - Build ArangoDB Community Edition (Developer Build)
  "enterprise-developer"            - Build ArangoDB Enterprise Edition (Developer Build)
  "community-developer-tsan"        - Build ArangoDB Community Edition (TSAN Build)
  "enterprise-developer-tsan"       - Build ArangoDB Enterprise Edition (TSAN Build)
  "community-developer-asan-ubsan"  - Build ArangoDB Community Edition (ASAN and UBSAN Build)
  "enterprise-developer-asan-ubsan" - Build ArangoDB Enterprise Edition (ASAN and UBSAN Build)

# cmake --preset community-developer

...

# cmake --build --preset community-developer --parallel 4

...

# build/community-developer/bin/arangod --database.directory /tmp/foo
2022-08-08T13:34:31Z [1866168] INFO [e52b0] {general} ArangoDB 3.11.0-devel [linux] 64bit maintainer mode, using jemalloc, build heads/feature/cmake-presets-for-great-good-0-g8121a5cecf6, VPack 0.1.36, RocksDB 7.2.0, ICU 64.2, V8 7.9.317, OpenSSL 1.1.1q  5 Jul 2022

```

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -7,33 +7,124 @@
   },
   "configurePresets": [
     {
-      "name": "default",
-      "displayName": "Default Config",
-      "binaryDir": "${sourceDir}/build/default",
+      "name": "community",
+      "displayName": "Build ArangoDB Community Edition",
+      "binaryDir": "${sourceDir}/build/community",
       "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_BUILD_TYPE": "Release",
         "USE_ENTERPRISE": "Off",
         "USE_JEMALLOC": "On",
-        "USE_MAINTAINER_MODE": "Off",
-        "USE_FAILURE_TESTS": "Off",
-        "USE_FAIL_ON_WARNINGS": "Off",
-        "USE_IPO": "Off"
-      },
-      "environment": {
-
+        "USE_IPO": "On"
+      }
+    },
+    {
+      "name": "enterprise",
+      "displayName": "Build ArangoDB Enterprise Edition",
+      "binaryDir": "${sourceDir}/build/enterprise",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "USE_ENTERPRISE": "On",
+        "USE_JEMALLOC": "On",
+        "USE_IPO": "On"
+      }
+    },
+    {
+      "name": "community-developer",
+      "inherits": "community",
+      "displayName": "Build ArangoDB Community Edition (Developer Build)",
+      "binaryDir": "${sourceDir}/build/community-developer",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "USE_IPO": "Off",
+        "USE_MAINTAINER_MODE": "On",
+        "USE_FAIL_ON_WARNINGS": "On",
+        "USE_SEPARATE_REPLICATION2_TESTS_BINARY": "On"
+      }
+    },
+    {
+      "name": "enterprise-developer",
+      "inherits": "enterprise",
+      "displayName": "Build ArangoDB Enterprise Edition (Developer Build)",
+      "binaryDir": "${sourceDir}/build/enterprise-developer",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "USE_IPO": "Off",
+        "USE_MAINTAINER_MODE": "On",
+        "USE_FAIL_ON_WARNINGS": "On",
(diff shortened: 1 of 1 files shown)


jvolmer (commented):
Seems to work for me (I use now CMake 2.24).

I also have DUSE_GOOGLE_TESTS=On, DUSE_FAILURE_TESTS=On, DUSE_STRICT_OPENSSL_VERSION=Off for my build, not sure if these settings are still relevant or should be included here, just wanted to mention it.

MBkkt:
"aulsan" in that case probably better name, same as jenkins, address undefined leak (for example on apple clang leak not part of address)

CMakePresets.json
```diff
@@ -0,0 +1,134 @@
...
+        "USE_JEMALLOC": "On",
+        "USE_IPO": "On"
+      }
+    },
+    {
+      "name": "community-developer",
+      "inherits": "community",
+      "displayName": "Build ArangoDB Community Edition (Developer Build)",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
```
MBkkt, line 36: Debug. RelWithDebInfo build long and useless

CMakePresets.json
```diff
@@ -0,0 +1,134 @@
...
+        "USE_IPO": "Off",
+        "USE_MAINTAINER_MODE": "On",
+        "USE_FAIL_ON_WARNINGS": "On",
+	"USE_GOOGLE_TESTS": "On",
+	"USE_FAILURE_TESTS": "On",
+	"USE_STRICT_OPENSSL_VERSION": "Off",
+        "USE_SEPARATE_REPLICATION2_TESTS_BINARY": "On"
```
MBkkt, line 43: Why is it default?

CMakePresets.json
mchacki: I have no real preference, I have always used RelWithDebInfo in the past.
What is the benefit of Debug?

CMakePresets.json
```diff
@@ -0,0 +1,134 @@
...
+      "inherits": "enterprise",
+      "displayName": "Build ArangoDB Enterprise Edition (Developer Build)",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "USE_IPO": "Off",
+        "USE_MAINTAINER_MODE": "On",
+        "USE_FAIL_ON_WARNINGS": "On",
+	"USE_GOOGLE_TESTS": "On",
+	"USE_FAILURE_TESTS": "On",
+	"USE_STRICT_OPENSSL_VERSION": "Off",
```
mchacki, line 57: Not serious bit indention is of here ;)

CMakePresets.json
mchacki, line 43: I think there is also USE_SEPARATE_GRAPH_TESTS_BINARY.
my personal preference is to have the separate Test file (or properply split all.)
I would like to have the default of both to be ON or both to be OFF. (My preference is ON)

CMakePresets.json
```diff
@@ -0,0 +1,134 @@
...
+        "USE_SEPARATE_REPLICATION2_TESTS_BINARY": "On"
```
mchacki, line 58: See above

CMakePresets.json
markuspf: I think I will make the Replication2 and Graph test binaries proper `add_executable`s in their subdirectories and just exclude them from all.

Then they're always configured (negligible cost), and you can always build them if you're developing on that bit;

how does that sound?

CMakePresets.json
MBkkt: > What is the benefit of Debug?

Faster to build, better stacks

MBkkt:
@markuspf 

Why was it merged with completely ignoring all my comments?

CMakePresets.json
MBkkt: I use make -j 18 to build all targets, but I don't need separate binary for replication2.

I think better to make same for replication2 as for iresearch.
You always create target but it will be builded only with explicit call

MBkkt:
I also think for developers very useful flag:
`-DBASE_LD_FLAGS="-fuse-ld=lld"`

PR #16716, opened 2022-08-04
[PREGEL] Use message structs everywhere possible

jvolmer:
- Make inspector able to read and write VPackBuilder to ease the use of Inspectors for nested data structures
- Use message structs for StartGss, GssFinished and CleanupFinished messages
- Use VPackBuilder for aggregators in other structs (here I revert a change I made before to be able to create messages: Now with the VPackBuilder ability for Inspectors, we can change the aggregators back to be unique_ptrs (as they were originally).

### DIFF
--- a/arangod/Pregel/Statistics.h
+++ b/arangod/Pregel/Statistics.h
@@ -25,6 +25,7 @@
 
 #include <velocypack/Builder.h>
 #include <velocypack/Slice.h>
+#include <cstdint>
 
 #include "Pregel/Utils.h"
 #include "Logger/LogMacros.h"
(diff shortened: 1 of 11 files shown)


arangod/Pregel/Statistics.h
```diff
@@ -87,11 +88,8 @@ struct StatsManager {
     }
   }
 
-  void accumulateMessageStats(VPackSlice data) {
-    VPackSlice sender = data.get(Utils::senderKey);
-    if (sender.isString()) {
-      _serverStats[sender.copyString()].accumulate(data);
-    }
+  void accumulateMessageStats(std::string const& senderId, VPackSlice data) {
+    _serverStats[senderId].accumulate(data);
```
markuspf, line 92: I think we should be using `.at()` (which will throw an exception and not just crash if we don't know the server ID), or even have error handling here (which might be an option for later us).
```suggestion
    _serverStats.at(senderId).accumulate(data);
```

markuspf (changes requested):
One small issue (sorry!).

arangod/Pregel/Statistics.h
jvolmer: I have to investigate here: Now, an exception is thrown...

PR #16723, opened 2022-08-08
Web UI: Improve AQL highlighter

Simran-B:
### Scope & Purpose

Support identifiers in forward ticks, multi-line strings, pseudo keywords, etc.

Note that there is a bug with case insensitive matching in Ace editor: https://github.com/ajaxorg/ace/issues/4887

Our version is v1.2.6 from 2016. If the bug gets fixed, we need to upgrade to v1.8.x+, which may require additional testing (**edit:** unlike to get fixed at the moment, resorting to a workaround with character classes).

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested additions and corner cases in the web interface:
    ```
    FOR doc IN @@view SEARCH true
    FILTER doc[@attr] AT LEAST (2) > .12e3
    AND "Lorem\t\n\x
    ipsum\" \u0F6Asit" != '\
    dolor\\\'amet'
    COLLECT WITH COUNT INTO count OPTIONS { method: "hash" }
    LET `foo
    🦄
    bar` =5INSERT {x:´foo\n\uD83E\uDD84\nbar´} INTO @@coll RETURN NEW._key
    ```
    image<br><sup>(`\r\n` on Windows to make the identifier match, `\n` on Linux)</sup>
- [x] Backports
  - [x] Backport for 3.10: *(Please link PR)*


js/apps/system/_admin/aardvark/APP/react/public/assets/src/mode-aql.js
```diff
@@ -132,34 +139,67 @@ var AqlHighlightRules = function() {
...
+            caseInsensitive: true // BUG: https://github.com/ajaxorg/ace/issues/4887
+        }, {
+            token : "language.variable", // case sensitive
+            regex : /(?:CURRENT|NEW|OLD)\b/
+        }, {
+            token : keywordMapper,
+            regex : /[a-zA-Z_][a-zA-Z0-9_]*\b/
+        }, {
+            token : "identifier",
+            regex : /(?:\$?|_+)[a-zA-Z]+[_a-zA-Z0-9]*/,
```
Simran-B, line 202: Could use this regex for the keywordMapper instead, which makes all non-matching tokens "identifier" anyway.

js/apps/system/_admin/aardvark/APP/react/public/assets/src/mode-aql.js
```diff
@@ -132,34 +139,67 @@ var AqlHighlightRules = function() {
...
+            token : "punctuation",
+            regex : /[\.,]/
+        }, {
+            token : "keyword",  // language construct usable in COLLECT
+            regex : /WITH\s+COUNT\s+INTO\b/,
+            caseInsensitive: true
+        }, {
+            token : "keyword",  // handle separately because of the whitespace
+            regex : /AT\s+LEAST\b/,
+            caseInsensitive: true // BUG: https://github.com/ajaxorg/ace/issues/4887
```
Simran-B, line 193:
```suggestion
            regex : /[Aa][Tt]\s+[Ll][Ee][Aa][Ss][Tt]\b/
```

Simran-B:
It's quite robust now:

image

It handles identifiers wrapped in forward ticks and backticks, including newlines and escapes. It can also handle single- and double-quoted strings, including new lines and escapes.

Known limitations: still false-positive matches of pseudo-keywords like SEARCH, KEEP, etc. I believe `identifier` isn't actually a token type that gets styled, but I left unrecognized tokens as "identifier" because "variable" would color everything that isn't something else.

jsteemann (approved):
LGTM

ajurdak (approved):
LGTM

PR #16729, opened 2022-08-09
bug-fix/FE-148 Correct limit calculation in API

ajurdak:
### Scope & Purpose

The API automatically added "1" to the limit. Therefore the amount of nodes and the shown value in the UI was wrong. This fix corrects the calculation in the API.

- [x] :hankey: Bugfix

### Checklist

- [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-148


adityamukho (commented):
Strange behaviour for an API. Although, this workaround looks sound, it's a little ambiguous as to the behaviour when the limit is set to 0 or 1 by the user. What does the API do to the limit in these cases?

js/apps/system/_admin/aardvark/APP/aardvark.js
```diff
@@ -1137,6 +1137,9 @@ authRouter.get('/g6graph/:name', function (req, res) {
   if (config.limit !== undefined) {
     if (config.limit.length > 0 && config.limit !== '0') {
       limit = parseInt(config.limit);
+      if(limit >= 2) {
```
hkernbach, line 1140:
```suggestion
      if (limit >= 2) {
```

PR #16738, opened 2022-08-09
[PREGEL] Logging for Pregel soft shutdown fail

markuspf:
### Scope & Purpose

There currently is a problem with soft shutdowns and pregel hanging. This mostly manifests on Windows and is not (easily) reproducible.

This PR adds log outputs to get closer to the problem.


jvolmer (approved):
LGTM

PR #16752, opened 2022-08-10
[PREGEL] Separate pregel cmake build

jvolmer:


markuspf (commented):
I think if you want to move the Pregel handlers we might have to start by convincing people that REST handlers should live in the subfolders of the respective subsystem. WDYT?

jvolmer:
I think Handlers belong into the feature directory such that you can easily find it and if you decide to delete the feature, you don't have to look where else it is used.
As far as I see, some features already do that, I see RestHandler files at least in RocksDBEngine, Cluster, ClusterEngine, Agency and Aql.

markuspf (approved):
Very nice.

PR #16773, opened 2022-08-12
[CMAKE] Add cppcheck target to CMakeLists.txt

markuspf:
### Scope & Purpose

Adds a `cppcheck` target to `CMakeLists.txt`; enables the user to, after running `cmake`, run `make cppcheck` to get their source checked using `cppcheck` (if that tool is installed)

### DIFF
--- a/cmake/custom/CppCheckTarget.cmake
+++ b/cmake/custom/CppCheckTarget.cmake
@@ -3,14 +3,12 @@ find_program(CMAKE_CXX_CPPCHECK cppcheck NAMES cppcheck HINTS $ENV{PROGRAMFILES}
   
 if(CMAKE_CXX_CPPCHECK)
   # Check CppCheck version
-  set(CPP_CHECK_CMD ${CMAKE_CXX_CPPCHECK} --version)
-    execute_process(COMMAND ${CPP_CHECK_CMD}
+  execute_process(COMMAND ${CMAKE_CXX_CPPCHECK} --version
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
       RESULT_VARIABLE CPP_CHECK_RESULT
       OUTPUT_VARIABLE CPP_CHECK_VERSION
       ERROR_VARIABLE CPP_CHECK_ERROR)
     
-  # Check if version could be extracted
   if(CPP_CHECK_RESULT EQUAL 0)
     # Get number of CPU cores
     include(ProcessorCount)
@@ -30,6 +28,7 @@ if(CMAKE_CXX_CPPCHECK)
 
         # Ignore 3rdParty code
         "-i${CMAKE_SOURCE_DIR}/3rdParty"
+	"-i${CMAKE_BINARY_DIR}/"
 
         # Only show found errors
         "--quiet"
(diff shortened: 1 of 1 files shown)


markuspf:
There is also an option to, while CMake builds to project, to just run cppcheck on every file and error out if there's an issue.

That approach would have the obvious downside of slowing down compilation by quite a bit.

cmake/custom/CppCheckTarget.cmake
```diff
@@ -0,0 +1,67 @@
...
+if(CMAKE_CXX_CPPCHECK)
+  # Check CppCheck version
+  set(CPP_CHECK_CMD ${CMAKE_CXX_CPPCHECK} --version)
+    execute_process(COMMAND ${CPP_CHECK_CMD}
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      RESULT_VARIABLE CPP_CHECK_RESULT
+      OUTPUT_VARIABLE CPP_CHECK_VERSION
+      ERROR_VARIABLE CPP_CHECK_ERROR)
+    
+  # Check if version could be extracted
```
jvolmer, line 13: What do you mean with 'version' here?

cmake/custom/CppCheckTarget.cmake
```diff
@@ -0,0 +1,67 @@
+# Add Analyze with CppCheck target if CppCheck is installed
+find_program(CMAKE_CXX_CPPCHECK cppcheck NAMES cppcheck HINTS $ENV{PROGRAMFILES}/cppcheck)
+  
+if(CMAKE_CXX_CPPCHECK)
+  # Check CppCheck version
+  set(CPP_CHECK_CMD ${CMAKE_CXX_CPPCHECK} --version)
```
jvolmer, line 6: do you really need this variable, it is just used once ...

cmake/custom/CppCheckTarget.cmake
```diff
@@ -0,0 +1,67 @@
...
+
+      # Legacy from old script; unclear why this is required.
+      "-DUSE_PLAN_CACHE"
+      "-DDEFINE_FACTORY_DEFAULT"
+        )
+      
+    add_custom_target(cppcheck
+     COMMAND ${CMAKE_CXX_CPPCHECK} 2> "cppcheck.xml"
+      WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+      COMMENT "Static code analysis using ${CPP_CHECK_VERSION}"
```
jvolmer, line 63: (nitpicking:) you could also add `BYPRODUCTS cppcheck.xml` to tag this file as a byproduct of cmake. Then this file will be deleted when executing `make clean`.

jvolmer (approved):
LGTM

cmake/custom/CppCheckTarget.cmake
markuspf: I think I wanted to have a registry of versions of software that were used during the build so we can have better visibility. Not sure this belongs in this PR.

markuspf:
The target now has the problem that one has to run a full build to run cppcheck, because some generated files do not exist yet, and cppcheck insists on loading them from the build database; not sure yet whether one can ignore those somehow (-i doesn't work).

PR #16775, opened 2022-08-15
fix potential data races in dump/restore queue

jsteemann:
### Scope & Purpose

Fix potential data races in task queue used by arangodump & arangorestore.
There could have been a race between the main arangodump/arangorestore thread checking if all jobs have been finished, and a queue worker thread popping a job from the job queue and only afterwards reporting as busy.

Tries to fix https://jenkins01.arangodb.biz/job/arangodb-ANY-linux-san.x86-64/1228/EDITION=enterprise,SAN_MODE=TSan,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=linux&&test&&x86-64/artifact/tsan.log.arangorestore.139376.log

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/client-tools/Restore/RestoreFeature.cpp
+++ b/client-tools/Restore/RestoreFeature.cpp
@@ -2260,7 +2260,7 @@ void RestoreFeature::reportError(Result const& error) {
   try {
     {
       MUTEX_LOCKER(lock, _workerErrorLock);
-      _workerErrors.emplace(error);
+      _workerErrors.emplace_back(error);
     }
     _clientTaskQueue.clearQueue();
   } catch (...) {
--- a/client-tools/Restore/RestoreFeature.cpp
+++ b/client-tools/Restore/RestoreFeature.cpp
@@ -37,6 +37,8 @@
 
 #include "ApplicationFeatures/ApplicationServer.h"
 #include "Basics/FileUtils.h"
+#include "Basics/Mutex.h"
+#include "Basics/MutexLocker.h"
 #include "Basics/NumberOfCores.h"
 #include "Basics/Result.h"
 #include "Basics/StaticStrings.h"
--- a/client-tools/Restore/RestoreFeature.cpp
+++ b/client-tools/Restore/RestoreFeature.cpp
@@ -2256,8 +2256,10 @@ ClientTaskQueue<RestoreFeature::RestoreJob>& RestoreFeature::taskQueue() {
 
 void RestoreFeature::reportError(Result const& error) {
   try {
-    MUTEX_LOCKER(lock, _workerErrorLock);
-    _workerErrors.emplace(error);
+    {
+      MUTEX_LOCKER(lock, _workerErrorLock);
+      _workerErrors.emplace(error);
+    }
     _clientTaskQueue.clearQueue();
   } catch (...) {
   }
--- a/client-tools/Utils/ClientTaskQueue.h
+++ b/client-tools/Utils/ClientTaskQueue.h
@@ -110,6 +110,16 @@ class ClientTaskQueue {
   bool isQueueEmpty() const noexcept;
 
   /**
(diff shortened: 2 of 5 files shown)


client-tools/Utils/ClientTaskQueue.h
```diff
@@ -241,22 +253,37 @@ inline bool ClientTaskQueue<JobData>::spawnWorkers(
...
+  CONDITION_LOCKER(lock, _jobsCondition);
+  return _jobs.empty();
+}
+
+template<typename JobData>
+inline bool ClientTaskQueue<JobData>::isQueueEmptyAndAllWorkersIdle()
+    const noexcept {
+  // acquire both locks here, so we can avoid data races
+  CONDITION_LOCKER(lock1, _jobsCondition);
+  CONDITION_LOCKER(lock2, _workersCondition);
```
MBkkt, line 265: Suggest add try_lock for ConditionVariable(pthread and TryEnterCriticalSection (also I think better to use srwmutex)) and use std::lock to avoid possible deadlock

client-tools/Utils/ClientTaskQueue.h
```diff
@@ -394,18 +404,27 @@ inline bool ClientTaskQueue<JobData>::Worker::isIdle() const noexcept {
...
 inline void ClientTaskQueue<JobData>::Worker::run() {
   while (!isStopping()) {
+    // we must unconditionally report as busy here, regardless of if there is
+    // actually a job on the queue or not. this is because otherwise there can
+    // be a race between
+    // - this thread popping a job from the queue and only after that reporting
+    // busy.
+    // - someone checking if there are no jobs left and all workers being idle.
+    //
+    _idle.store(false);
```
MBkkt, line 414: What about set idle under queue mutex?

MBkkt (approved):
LGTM
But I think we can make basics::ConditionVariable/Mutex TRI_* better

client-tools/Utils/ClientTaskQueue.h
jsteemann: There is currently no queue mutex.
There are 2 condition variables (with associated mutexes) for _jobs and _workers, but they are located on a different object (the queue) whereas the idle flag is located on a Worker thread object.

Anyway, I changed things around so that:
- we only have one condition variable (including its mutex) for _jobs, _workers and the _queue. Previously we had 2 different ones, which allowed for all kinds of races. With the single cv everything is a lot easier. The downside is that we may have more contention than before, but that shouldn't be a problem. We are using this code only for dump/restore, and the queue management cost should not matter much here
- the atomic `_idle` variable is now set and checked under that condition as well.

client-tools/Restore/RestoreFeature.cpp
```diff
@@ -2256,8 +2258,10 @@ ClientTaskQueue<RestoreFeature::RestoreJob>& RestoreFeature::taskQueue() {
 
 void RestoreFeature::reportError(Result const& error) {
   try {
-    MUTEX_LOCKER(lock, _workerErrorLock);
-    _workerErrors.emplace(error);
+    {
+      MUTEX_LOCKER(lock, _workerErrorLock);
+      _workerErrors.emplace(error);
```
MBkkt, line 2263: emplace back?

MBkkt (changes requested):

MBkkt (approved):

PR #16776, opened 2022-08-15
[CINFRA-523] Set assumedWaitForSync when creating replicated log

markuspf:
### Scope & Purpose

Set `assumedWaitForSync` on log creation


markuspf:
Will add test as well.

apetenchea:
The failure might be related to this: https://arangodb.atlassian.net/browse/CINFRA-312

apetenchea (approved):

PR #16783, opened 2022-08-15
[CMake] Require more recent cmake

markuspf:
### Scope & Purpose

Require `CMake 3.21` for building ArangoDB; This is mostly to introduce useful `CMakePresets` as done in #16687 ; presets were introduced in cmake 3.19 and gained some more features in subsequent versions.

3.21 was chosen because this is the version that comes bundled with current CLion as to not disrupt developers workflow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # General
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


jvolmer (approved):
LGTM

PR #16785, opened 2022-08-16
remove duplicates

dothebart:
### Scope & Purpose

tests must not be twice in the test definitions without specifying parameters that create different behaviour and logfiles.


KVS85 (approved):
LGTM.

PR #16796, opened 2022-08-16
[CMake] Random cleanups

markuspf:
### Scope & Purpose

Let's start spring cleaning CMakeLists.txt.


CMakeLists.txt
```diff
@@ -287,7 +227,7 @@ endif()
 option(STATIC_EXECUTABLES "produce static executables" OFF)
 if (STATIC_EXECUTABLES)
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static")
-  if (WINDOWS)
+  if (WIN32)
```
dothebart, line 230: since we actually don't build for 32 bit windows, I'd rather remove `WIN32` from the place instead of adding more

jsteemann: `WIN32` in CMake doesn't make 32 bits.

From https://cmake.org/cmake/help/latest/variable/WIN32.html:

> Set to True when the target system is Windows, including Win64.

CMakeLists.txt
```diff
@@ -144,23 +131,11 @@ if (DEFINED ARANGODB_VERSION_RELEASE_TYPE)
 
   # devel
   if (ARANGODB_VERSION_PATCH STREQUAL "0" AND ARANGODB_VERSION_RELEASE_TYPE STREQUAL "devel")
-    set(ARANGODB_DEBIAN_UPSTREAM "${ARANGODB_VERSION_MAJOR}.${ARANGODB_VERSION_MINOR}.0~~${ARANGODB_VERSION_PATCH}")
-    set(ARANGODB_DEBIAN_REVISION "1")
-
-    set(ARANGODB_RPM_UPSTREAM "${ARANGODB_PLAIN_VERSION}")
-    set(ARANGODB_RPM_REVISION "0.1.devel")
-
```
jvolmer, line 152: Out of curiosity: Why did you delete all the variables for Debian and RPM but not for Darwin?

markuspf: This is a good question; I think I assumed that a Mac release build somehow needs this. Not sure though, will have to look/ask;

PR #16800, opened 2022-08-17
Bug fix/limit size of arangodlogs

dothebart:
### Scope & Purpose

as seen previously forgetting debug logging in tests leads to unstable CI. Assert this for green tests.

- [x] :hankey: Bugfix

### DIFF
--- a/js/client/modules/@arangodb/testutils/instance.js
+++ b/js/client/modules/@arangodb/testutils/instance.js
@@ -515,9 +515,9 @@ class instance {
 
   readAssertLogLines () {
     let size = fs.size(this.logFile);
-    if (size > 500000) {
+    if (size > this.options.maxLogFileSize) {
       // File bigger 500k? this needs to be a bug in the tests.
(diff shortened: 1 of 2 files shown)


naushniki (approved):

js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -514,6 +514,14 @@ class instance {
   // //////////////////////////////////////////////////////////////////////////////
 
   readAssertLogLines () {
+    let size = fs.size(this.logFile);
+    if (size > 500000) {
+      // File bigger 500k? this needs to be a bug in the tests.
+      let err=`ERROR: ${this.logFile} is bigger than 500kB! - %{size} Bytes!`;
+      this.assertLines.push(err);
+      print(RED + err + RESET);
+      return;
```
KVS85, line 523: I don't think it should be like this. There might be a need to enable TRACE or DEBUG and turn this check off. Moreover, 501kB doesn't sound like a huge log.

KVS85 (changes requested):

KVS85 (approved):
LGTM.

PR #16801, opened 2022-08-17
[CINFRA-528] Add commandline option to set default replication version.

markuspf:
### Scope & Purpose

When starting `arangod` one can now give the parameter `--database.default-replication-version <replicationVersion>` to set the default replication to either `1` or `2`.

### DIFF
--- a/arangod/RestServer/DatabaseFeature.h
+++ b/arangod/RestServer/DatabaseFeature.h
@@ -30,6 +30,7 @@
 #include "Metrics/Counter.h"
 #include "Metrics/Histogram.h"
 #include "Metrics/LogScale.h"
+#include "Replication2/Version.h"
 #include "RestServer/arangod.h"
 #include "Utils/VersionTracker.h"
 #include "VocBase/voc-types.h"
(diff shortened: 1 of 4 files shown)


arangod/RestServer/DatabaseFeature.h
```diff
@@ -244,6 +248,7 @@ class DatabaseFeature : public ArangodFeature {
   bool _isInitiallyEmpty;
   bool _checkVersion;
   bool _upgrade;
+  replication::Version _defaultReplicationVersion;
```
mpoeter, line 251: If possible I prefer inline initialization over constructor initializer lists, but feel free to ignore:
```suggestion
  replication::Version _defaultReplicationVersion = replication::Version::ONE;
```

mpoeter (approved):

apetenchea (approved):

arangod/RestServer/DatabaseFeature.h
markuspf: I will ignore, but with the rationale that I kept the code consistent with what was there before ;)

PR #16808, opened 2022-08-18
[PREGEL] Simpler Algorithm SDK

markuspf:
### Scope & Purpose

Draft to create a simpler way to implement Pregel Algorithms.


markuspf (commented):
Test

PR #16809, opened 2022-08-19
Require CMake 3.21

markuspf:
To enable the use of CMakePresets.json

Since apparently reopening #16783 does not work.


markuspf:
Sorry, that was a fat fingered thing. I will revert.. .again.

PR #16811, opened 2022-08-19
[PREGEL] implement load graph with future

jvolmer:
This PR introduces futures to load the graph. This is used initialize the loading at the conductor, request loading the graph for all workers and continue the work on the conductor as soon as all workers have finished loading.

This PR includes the inspector for ResultT because I need it here already. I already created a PR for devel on that (https://github.com/arangodb/arangodb/pull/16891), so you don't have to look at the code changes in lib in this PR.

### DIFF
--- a/arangod/Pregel/GraphStore.h
+++ b/arangod/Pregel/GraphStore.h
@@ -33,6 +33,7 @@
 #include "Pregel/Status/Status.h"
 #include "Pregel/TypedBuffer.h"
 #include "Pregel/Worker/WorkerConfig.h"
+#include "Pregel/WorkerConductorMessages.h"
 #include "Utils/DatabaseGuard.h"
 
 #include <atomic>
@@ -130,9 +131,9 @@ class GraphStore final {
   GraphFormat<V, E> const* graphFormat() { return _graphFormat.get(); }
 
   // ====================== NOT THREAD SAFE ===========================
-  void loadShards(WorkerConfig* config,
-                  std::function<void()> const& statusUpdateCallback,
-                  std::function<void()> const& finishedLoadingCallback);
+  auto loadShards(WorkerConfig* config,
+                  std::function<void()> const& statusUpdateCallback)
+      -> futures::Future<ResultT<GraphLoaded>>;
   void loadDocument(WorkerConfig* config, std::string const& documentID);
   void loadDocument(WorkerConfig* config, PregelShard sourceShard,
                     std::string_view key);
--- a/arangod/Pregel/WorkerInterface.h
+++ b/arangod/Pregel/WorkerInterface.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#include "Pregel/WorkerConductorMessages.h"
+
+#include <Futures/Future.h>
+
+namespace arangodb::pregel {
+
+class NewIWorker {
+ public:
+  virtual ~NewIWorker() = default;
+  [[nodiscard]] virtual auto loadGraph(LoadGraph const& graph)
+      -> futures::Future<ResultT<GraphLoaded>> = 0;
+};
+
+}  // namespace arangodb::pregel
--- a/arangod/Pregel/Conductor/Conductor.cpp
+++ b/arangod/Pregel/Conductor/Conductor.cpp
@@ -183,7 +183,6 @@ bool Conductor::_startGlobalStep() {
     THROW_ARANGO_EXCEPTION(TRI_ERROR_SHUTTING_DOWN);
   }
 
-  _callbackMutex.assertLockedByCurrentThread();
   // send prepare GSS notice
   auto prepareGssCommand = PrepareGss{.executionNumber = _executionNumber,
                                       .gss = _globalSuperstep,
@@ -405,7 +404,8 @@ void Conductor::cancel() {
   changeState(conductor::StateType::Canceled);
 }
 
-// resolves into an ordered list of shards for each collection on each server
+// resolves into an ordered list of shards for each collection on each
+// server
 static void resolveInfo(
     TRI_vocbase_t* vocbase, CollectionID const& collectionID,
     std::map<CollectionID, std::string>& collectionPlanIdMap,
@@ -455,11 +455,10 @@ static void resolveInfo(
 }
 
 /// should cause workers to start a new execution
-ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
-                                        VPackSlice additional) {
+auto Conductor::_initializeWorkers(VPackSlice additional)
+    -> futures::Future<std::vector<
+        futures::Try<arangodb::ResultT<arangodb::pregel::GraphLoaded>>>> {
   _callbackMutex.assertLockedByCurrentThread();
-
-  std::string const path = Utils::baseUrl(Utils::workerPrefix) + suffix;
 
   // int64_t vertexCount = 0, edgeCount = 0;
   std::map<CollectionID, std::string> collectionPlanIdMap;
@@ -479,9 +478,20 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
                 shardList);  // store or
   }
 
-  _dbServers.clear();
-  for (auto const& pair : vertexMap) {
-    _dbServers.push_back(pair.first);
+  if (ServerState::instance()->getRole() == ServerState::ROLE_SINGLE) {
+    TRI_ASSERT(vertexMap.size() == 1);
+    auto [server, _] = *vertexMap.begin();
+    _dbServers.push_back(server);
+    workers.emplace(server, std::make_unique<SingleServerWorkerOnConductor>(
+                                _executionNumber, server, _feature,
+                                _vocbaseGuard.database()));
+  } else {
+    for (auto const& [server, _] : vertexMap) {
+      _dbServers.push_back(server);
+      workers.emplace(server,
+                      std::make_unique<ClusterWorkerOnConductor>(
+                          _executionNumber, server, _vocbaseGuard.database()));
+    }
   }
   _status = ConductorStatus::forWorkers(_dbServers);
   // do not reload all shard id's, this list must stay in the same order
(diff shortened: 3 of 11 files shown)


arangod/Pregel/Conductor.h
```diff
@@ -46,14 +47,49 @@
...
 #include "velocypack/Builder.h"
 
 #include <chrono>
+#include <cstdint>
 #include <memory>
 
 namespace arangodb {
 namespace pregel {
 
+struct ClusterConductorWorker : NewIWorker {
```
romanatarango, line 60: Please, explain in a short comment what this class is for: it's an API of a worker whose instances run on conductors in a cluster.

arangod/Pregel/Conductor.h
```diff
@@ -46,14 +47,49 @@
...
+  [[nodiscard]] auto loadGraph(LoadGraph const& graph)
+      -> futures::Future<ResultT<GraphLoaded>> override;
+
+ private:
+  ExecutionNumber _executionNumber;
+  ServerID _serverId;
+  const DatabaseGuard _vocbaseGuard;
+};
+
+struct SingleServerConductorWorker : NewIWorker {
```
romanatarango, line 75: Please, explain in a short comment what this class is for: it's an API of a worker whose instances run on conductors in a single server.

arangod/Pregel/WorkerInterface.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "Pregel/WorkerConductorMessages.h"
+
+#include <Futures/Future.h>
+
+namespace arangodb::pregel {
+
+class NewIWorker {
```
romanatarango, line 9: Why "new"?

romanatarango (approved):
LGTM, but see small comments about comments.

arangod/Pregel/WorkerInterface.h
jvolmer: This Worker interface is only temporary because I changed only one function - loadGraph - to use futures: Currently we still have the IWorker interface we need for the Worker. After I converted all functions in IWorker to futures, we get rid of IWorker and only use NewIWorker (which I will then rename to IWorker).

arangod/Pregel/Conductor.h
jvolmer: What do you think about a better name instead: ClusterWorkerOnConductor or ClusterWorkerOnCoordinator

arangod/Pregel/Conductor.cpp
```diff
@@ -480,8 +478,22 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
...
+                    std::make_unique<SingleServerConductorWorker>(
+                        _executionNumber, serverCollections->first, _feature,
+                        _vocbaseGuard.database()));
+  } else {
+    for (auto const& pair : vertexMap) {
+      _dbServers.push_back(pair.first);
+      workers.emplace(pair.first, std::make_unique<ClusterConductorWorker>(
+                                      _executionNumber, pair.first,
+                                      _vocbaseGuard.database()));
+    }
```
markuspf, line 496:
```suggestion
    for (auto const& [server, _] : vertexMap) {
      _dbServers.push_back(server);
      workers.emplace(server, std::make_unique<ClusterConductorWorker>(
                                      _executionNumber, server,
                                      _vocbaseGuard.database()));
    }
```

arangod/Pregel/Conductor.cpp
```diff
@@ -480,8 +478,22 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
...
-    _dbServers.push_back(pair.first);
+  workers.clear();
+  if (ServerState::instance()->getRole() == ServerState::ROLE_SINGLE) {
+    TRI_ASSERT(vertexMap.size() == 1);
+    auto serverCollections = vertexMap.begin();
+    _dbServers.push_back(serverCollections->first);
+    workers.emplace(serverCollections->first,
+                    std::make_unique<SingleServerConductorWorker>(
+                        _executionNumber, serverCollections->first, _feature,
+                        _vocbaseGuard.database()));
```
markuspf, line 489:
```suggestion
    auto [server, _] = *vertexMap.begin(); // lets hope this is never empty :/
    _dbServers.push_back(server);
    workers.emplace(server,
                    std::make_unique<SingleServerConductorWorker>(
                        _executionNumber, server, _feature,
                        _vocbaseGuard.database()));
```

arangod/Pregel/Conductor.cpp
```diff
@@ -730,7 +687,6 @@ template<typename OutType, typename InType>
 auto Conductor::_sendToAllDBServers(std::string const& path,
                                     InType const& message)
     -> ResultT<std::vector<OutType>> {
-  _callbackMutex.assertLockedByCurrentThread();
```
markuspf, line 733: Is this ok? Or will it be ok once we have all the future-types sorted out?

arangod/Pregel/Conductor.cpp
```diff
@@ -455,12 +455,10 @@ static void resolveInfo(
 }
 
 /// should cause workers to start a new execution
-ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
-                                        VPackSlice additional) {
+auto Conductor::_initializeWorkers(VPackSlice additional)
+    -> std::vector<futures::Future<ResultT<GraphLoaded>>> {
```
markuspf, line 459: Could this somehow be a single future which just waits for all the insidey ones to resolve?

arangod/Pregel/Conductor.cpp
```diff
@@ -480,8 +478,22 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
   }
 
   _dbServers.clear();
-  for (auto const& pair : vertexMap) {
-    _dbServers.push_back(pair.first);
+  workers.clear();
```
markuspf, line 481: I don't  understand why this clearing is necessary; in the lifetime of a conductor this variable should only have the states "initialized empty" and "all the workers in there", shouldn't it?

Maybe it's better to have this be a `std::unique_ptr` to reflect this fact?

arangod/Pregel/Conductor.cpp
```diff
@@ -561,62 +570,10 @@ ErrorCode Conductor::_initializeWorkers(std::string const& suffix,
...
-      reqOpts.database = _vocbaseGuard.database().name();
-
-      responses.emplace_back(network::sendRequestRetry(
-          pool, "server:" + server, fuerte::RestVerb::Post, path,
-          std::move(buffer), reqOpts));
-
-      LOG_PREGEL("6ae66", DEBUG) << "Initializing Server " << server;
-    }
+    auto graph = LoadGraph{.executionNumber = _executionNumber, .details = b};
+    results.emplace_back(workers[server]->loadGraph(graph));
```
markuspf, line 574: :partying_face:

arangod/Pregel/GraphStore.h
```diff
@@ -130,9 +131,9 @@ class GraphStore final {
   GraphFormat<V, E> const* graphFormat() { return _graphFormat.get(); }
 
   // ====================== NOT THREAD SAFE ===========================
-  void loadShards(WorkerConfig* config,
-                  std::function<void()> const& statusUpdateCallback,
-                  std::function<void()> const& finishedLoadingCallback);
+  auto loadShards(WorkerConfig* config,
+                  std::function<void()> const& statusUpdateCallback)
```
markuspf, line 135: TODO: think about how to handle the status updates.

markuspf (commented):
Not quite done with the review yet, but left a couple of comments already

arangod/Pregel/Conductor.cpp
jvolmer: It should be ok when the future-types are all in place. I had problems running pregel when this was not deleted and had problems acquiring the lock and keep it for that long with the new future.

arangod/Pregel/Conductor.cpp
jvolmer: Ah, I added it because the recovery state would also call the initializeWorkers function again. We deleted recovery, so I will delete this clearing.

arangod/Pregel/Conductor.cpp
jvolmer: Good idea, I now return `futures::Future<std::vector<
        futures::Try<arangodb::ResultT<arangodb::pregel::GraphLoaded>>>>`

arangod/Pregel/GraphStore.h
jvolmer: Yes, I did not know how to improve that: We want to have status updates during the loading and not only at the end of loading, therefore the worker has to initiate the update, therefore we have to give this callback from the conductor to the worker.

arangod/Pregel/Conductor.h
romanatarango: WorkerAPIOnWhatever or WorkerInterfeaceOnWhatever. "Whatever" is probably "Conductor", but that's not that important.

arangod/Pregel/Conductor.cpp
markuspf: Not for this PR: 

In the next iteration this could do with a `using GraphLoadedFuture = std::vector<futures::Future<ResultT<GraphLoaded>>>` (and then I get ideas about how this should be done abstractly for a collection of workers.)

arangod/Pregel/Conductor.cpp
markuspf: I think this particular problem will be solved by us using `Guarded` wrappers around the datastructures we affect later; keep in mind how this callbackMutex seems to have come to be: only one thread is supposed to be in the "critical Section" of the conductor...

markuspf (approved):
I think this is OK now;

we should though start thinking about technical documentation for the intended design at least; as @Roman noticed it becomes a bit confusing quickly if we have IWorker and then ConductorWorker and WorkerWorker and friends ;)

PR #16814, opened 2022-08-19
[CMAKE] Move building  SEPP into its own place

markuspf:
### Scope & Purpose

Cleanup of CMakeLists.txt.


markuspf:
Small usability change: while SEPP is not built as a default when you run `make`, one can always build it using `make sepp` once the build is configured.

mpoeter (approved):

PR #16815, opened 2022-08-19
Add sources directly to cmake-targets in enterprise

jvolmer:
### Scope & Purpose

Add sources to cmake-targets directly without using a variable. See enterprise PR.

- Enterprise PR: https://github.com/arangodb/enterprise/pull/1012


tests/IResearch/CMakeLists.txt
```diff
@@ -72,12 +72,7 @@ set(ARANGODB_IRESEARCH_TESTS_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/IResearchViewSortedTest.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/IResearchViewCountApproximateTest.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/VelocyPackHelperTest.cpp
-  ${ADDITIONAL_IRESEARCH_TEST_SOURCES}
```
gnusi, line 75: that will remove all arangosearch related sources from tests

gnusi (changes requested):

tests/IResearch/CMakeLists.txt
jvolmer: No, this will not happen: This PR works only with the corresponding enterprise PR (https://github.com/arangodb/enterprise/pull/1012). In the enterprise PR you see that I deleted the `ADDITIONAL_IRESEARCH_TEST_SOURCES` variable, but added a specific CMakeLists.txt file in enterprise/tests/IResearch directory that directly adds the one source that was previously in `ADDITIONAL_IRESEARCH_TEST_SOURCES` directly to the targets that use it: arangodbtests and arangodbtests-iresearch.

markuspf (approved):
Very nice, thank you!

dothebart (approved):

PR #16825, opened 2022-08-19
[CMAKE] Compiler Checks

markuspf:
### Scope & Purpose

Improves and cleans up compiler version checks.

We define the minimum required version of each supported compiler (currently Clang, GCC, MSVC), and a CMake macro checks whether the compiler is supported and recent enough. 

When it is not, CMake terminates with an error message.

For the adventurous user there is he option `ARANGODB_SKIP_COMPILER_CHECK` which allows users to disable the check and run CMake anyway. 

Also remove all the compiler checks that have accumulated over time and are not used anymore.

### DIFF
--- a/cmake/CheckCompilerVersion.cmake
+++ b/cmake/CheckCompilerVersion.cmake
@@ -1,25 +1,47 @@
-macro(CheckCompilerVersion MIN_GCC_VERSION MIN_CLANG_VERSION MIN_MSVC_VERSION)
 
-if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
-  set(MIN_VERSION ${MIN_GCC_VERSION})
-elseif (CMAKE_C_COMPILER_ID STREQUAL "Clang")
-  set(MIN_VERSION ${MIN_CLANG_VERSION})
-elseif (CMAKE_C_COMPILER_ID STREQUAL "MSVC")
-  set(MIN_VERSION ${MIN_MSVC_VERSION})
-endif()
+option(ARANGODB_SKIP_COMPILER_CHECK "Skip compiler support check.")
 
-if(NOT WIN32)
-  string(ASCII 27 ESC)
-  set(COLOR_RESET "${ESC}[m")
-  set(RED        "${ESC}[31m")
-endif()
+macro(CheckMinVersionsSet)
+  if(NOT DEFINED ARANGODB_MIN_GCC_VERSION)
+    message(FATAL_ERROR "Please define ARANGODB_MIN_GCC_VERSION in CMakeLists.txt.")
+  endif()
+  if(NOT DEFINED ARANGODB_MIN_CLANG_VERSION)
+    message(FATAL_ERROR "Please define ARANGODB_MIN_CLANG_VERSION in CMakeLists.txt.")
+  endif()
+  if(NOT DEFINED ARANGODB_MIN_MSVC_VERSION)
+    message(FATAL_ERROR "Please define ARANGODB_MIN_MSVC_VERSION in CMakeLists.txt.")
+  endif()
+endmacro()
 
-if (NOT DEFINED MIN_VERSION)
-  message(WARNING "${RED}${CMAKE_C_COMPILER_ID}/${CMAKE_CXX_COMPILER_ID} compiler is not supported.${COLOR_RESET}")
-endif()
+macro(CheckMinCompilerVersionInner MIN_VERSION)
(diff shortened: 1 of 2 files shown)


MBkkt (changes requested):
We want warn when compiler not same as in CI
So == instead of >=, also only warning, not fatal error

cmake/CheckCompilerVersion.cmake
```diff
@@ -1,25 +1,47 @@
...
+  else()
+    message(
+      FATAL_ERROR
+        "ArangoDB does not support CMAKE_CXX_COMPILER_ID ${CMAKE_CXX_COMPILER_ID}"
+    )
+  endif()
 
 endmacro()
+
+CheckMinVersionsSet()
```
gnusi, line 43: The original macro dealt with arguments rather than variables set in parent scope. Please stick to the original approach.

gnusi (changes requested):

markuspf:
> We want warn when compiler not same as in CI So == instead of >=, also only warning, not fatal error

Who is "we" referring to? 

One of the reasons why ArangoDB's build and testing process is so bad is that it is never obvious what exactly went wrong and when.

I would just miss a warning. In one particular instance last week 3 developers waited quite some time for several builds of theirs to fail only to find out that clang 14 was required (on macos) to build arangodb.

Thus, I would much rather, fail early, and fail often. 

If you try compiling with a compiler that is not supported, the build fails. You can add a "yes please I want this"-flag to CMake if you really feel the need to push your luck.

If there are more compilers that can be supported, they can be added to that version check.

But furthermore, from your statement I gather that this check is supposed to solve a rather different problem from the one I was addressing: to make sure that some CI environment is reproduced.

I don't understand the added value of that check (or a warning if it fails)

cmake/CheckCompilerVersion.cmake
markuspf: I find it better to have some explicit name to the value under  scrutiny.

gnusi: Macro is a function and should work on its own, mixing execution scopes is not a good idea. You can always put any logic to the caller `CMakeLists.txt`.

maierlars (approved):

cmake/CheckCompilerVersion.cmake
mchacki: I am also in favor of having speaking names for variables, to improve maintainability (and that is highly required for our CMake...)

```
CheckCompilerVersion(
   11.2 # GCC
   14.0 # Clang
   19.32 # MSVC
 )
```

is something I highly dislike.

Can we go with:

```
CheckCompilerVersion(
  ARANGODB_MIN_GCC_VERSION,
 ARANGODB_MIN_CLANG_VERSION,
 ARANGODB_MIN_MSVC_VERSION
)
```

as a compromise?
Also I think it is rather verbose, and I like the approach from Markus better myself.

cmake/CheckCompilerVersion.cmake
```diff
@@ -1,25 +1,42 @@
...
-endif()
+macro(CheckMinCompilerVersionInner MIN_VERSION)
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS ${MIN_VERSION})
+    set(MESSAGE "ArangoDB requires at least ${CMAKE_CXX_COMPILER_ID} version ${MIN_VERSION},"
+                " detected ${CMAKE_CXX_COMPILER_VERSION}.")
 
-if (NOT DEFINED MIN_VERSION)
-  message(WARNING "${RED}${CMAKE_C_COMPILER_ID}/${CMAKE_CXX_COMPILER_ID} compiler is not supported.${COLOR_RESET}")
-endif()
+    if(${ARANGODB_WARN_COMPILER_CHECK})
```
jvolmer, line 10: is this correct or should this variable be `ARANGODB_JUST_WARN_COMPILER_CHECK`?

cmake/CheckCompilerVersion.cmake
jvolmer: Just an idea, the current approach is also fine for me: Why does it actually have to be one function / macro? We could directly write into the main CMakeLists.txt :
```
  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    CheckMinCompilerVersion(11.2)
  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
    CheckMinCompilerVersion(14.0)
  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
    CheckMinCompilerVersion(19.32)
```
Then we would just need one macro.

PR #16836, opened 2022-08-22
Added headers for graph creation examples

mchacki:
### Scope & Purpose

*Documentation only PR. adds separators between graph creation calls.*


mchacki:
Documentation only, not even example.
Merged knowingly without Jenkins.

PR #16844, opened 2022-08-22
[CINFRA-528] Use configured default replication version in more cases.

mpoeter:
Explicitly set the `replicationVersion` in the initial `_system` database entry when initializing the agency, and use the configured defaultReplicationVersion in several other cases that previously used a default fallback value of `replication::Version::ONE`.


arangod/Cluster/Maintenance.h
```diff
@@ -164,7 +165,8 @@ arangodb::Result diffPlanLocal(
     std::vector<std::shared_ptr<ActionDescription>>& actions,
     MaintenanceFeature::ShardActionMap const& shardActionMap,
     ReplicatedLogStatusMapByDatabase const& localLogs,
-    ReplicatedStateStatusMapByDatabase const& localStates);
+    ReplicatedStateStatusMapByDatabase const& localStates,
+    replication::Version defaultReplicationVersion = replication::Version::ONE);
```
apetenchea, line 169: Is this fallback value intentional?

apetenchea (approved):

arangod/Cluster/Maintenance.h
maierlars: we should remove it and make the version always explicit.

maierlars (approved):

mpoeter:
Blue except for unrelated test errors.

arangod/Cluster/Maintenance.cpp
```diff
@@ -1865,6 +1867,9 @@ arangodb::Result arangodb::maintenance::reportInCurrent(
           auto result = replication::parseVersion(rv);
           TRI_ASSERT(result.ok());
           replicationVersion = std::move(result.get());
+        } else {
+          // if "replicationVersion" field is not found this must be an old DB
+          // which defaults to version ONE.
```
markuspf, line 1872: Is this intentional (no code just else with comment)?

markuspf (approved):
This looks good to me.

Alas, how likely are we to find other places in the code where replication version is assumed to be `1` because of empty structures?

arangod/Cluster/Maintenance.cpp
mpoeter: Yes, the variable is already initialized with version::ONE in the definition. However, I wanted to place the comment here were we actually have the check instead of at the definition.

mpoeter:
> Alas, how likely are we to find other places in the code where replication version is assumed to be `1` because of empty structures?

Good question. I have checked all places where `version::ONE` has been used, but there is obviously no guarantee that I haven't missed something.

PR #16846, opened 2022-08-22
[BTS-951] Fix removeUnnecessaryRemoteScatterRule

markuspf:
### Scope & Purpose

If the topmost `REMOTE-SCATTER` block was contained in a subquery, it was removed regardless leading to an inconsistent query state.

There is an enterprise PR with a regression test for the original issue (BTS-951).

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [x] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [x] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #16862, opened 2022-08-23
[3.10] [BTS-951] Fix removeUnnecessaryRemoteScatterRule

markuspf:
### Scope & Purpose

Backport of #16846 ; there is also an enterprise branch that adds a regression test.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM, but we should add a CHANGELOG entry to this PR.

PR #16869, opened 2022-08-23
calculate documents only once, and use cached values for bulk inserts

dothebart:
### Scope & Purpose

*San is to slow. Save time by calculating documents only once.


jsteemann (approved):
LGTM

dothebart:
https://jenkins01.arangodb.biz/job/arangodb-devel-asan-and-coverage-biweekly/350/

PR #16876, opened 2022-08-24
Fix chaos test result transport

dothebart:
### Scope & Purpose

if test files are not a unique key, the testsuite needs to adopt this properly, else only the last result of same name will be represented in the summary.

Add Hook to the testrunner, implement it in chaos.js to make name uniq.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #16880, opened 2022-08-24
[CMAKE] Don't blanket-link absl

markuspf:
### Scope & Purpose

- [x] :hankey: Bugfix

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1264,15 +1264,6 @@ endif ()
 add_definitions(-DBOOST_UUID_RANDOM_PROVIDER_FORCE_POSIX=1) #fix random provider
 add_definitions(-DBOOST_ALL_NO_LIB=1) #disable boost autolink on windows
 add_subdirectory(3rdParty)
-link_libraries(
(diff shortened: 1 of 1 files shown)


MBkkt (changes requested):
Why?

markuspf:
I do not understand your question.

I was trying to find out whether removing this blanket link *on all targets* breaks something. 

Using `link_libraries` is strongly discouraged. (https://cmake.org/cmake/help/latest/command/link_libraries.html).

For my test binaries, I might just not want these libraries to be linked (or even built).

If there is a reason for this `link_libraries` to be here, please explain it to me.

MBkkt:
>If there is a reason for this link_libraries to be here, please explain it to me.

We don't want link every target with abseil, it's not convenient.

So if you want to remove this linking you need to add for all necessary targets target_link_libraries

MBkkt:
> For my test binaries, I might just not want these libraries to be linked (or even built).

It's utils, StrCat, hashmaps, etcs, replacement for part of std.

I don't want linking every target with it explicitly

PR #16882, opened 2022-08-24
throw proper error object

dothebart:
### Scope & Purpose

use a proper error object when throwing, so we have stacktraces etc. 

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

markuspf:
Would you mind changing the pull request title to something more descriptive (and maybe use labels)?

dothebart:
Sorry, but it only converts `throw <string>` to `throw new Error(<string>)` so it basically *only* does what the subject says.

PR #16886, opened 2022-08-24
UI: read & write visual graph properties bugfix [ES-1078]

hkernbach:
### Scope & Purpose

TLDR: Fixed reading and writing of visual graph properties (UI).
**Important:** This needs to be tested intensively, as we do not have automated UI tests (?) for this.

It seems that the response of (undocumented?) API `_api/user/<user>/config` has changed. 
* First, results are not handled at top-level anymore (e.g. now it is `results.graphs` instead of `graphs`).
* Second, it seems that also the modification logic changed internally. Beforehand it used to be a `PATCH` internally,
but now it is a `REPLACE`. I am pretty sure that this behaviour has changed. As this whole graph storage of properties worked once. 

Nevertheless, this PR should fix the behaviour.
Still there is a bit of future work todo regarding code read- & maintainability.

- [x] :hankey: Bugfix

### Checklist

- [x] Backports: Yes Backports are required!
  - [x] Backport for 3.10: #17443

#### Related Information

- [x] Enterprise PR: No enterprise PR required.
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/ES-1078


js/apps/system/_admin/aardvark/APP/frontend/js/models/userConfig.js
```diff
@@ -92,32 +92,91 @@ window.UserConfig = Backbone.Model.extend({
...
+        return newObject;
+      }
+
+      // Format notice
+      // dataFromDB<object>: dataFromDB.result.keyName
+      if (dataFromDB.result && dataFromDB.result[keyName]) {
+        const existingEntry = dataFromDB.result[keyName];
+        // We've found available data we need to merge
+        if (isObject(existingEntry) && isObject(newObject)) {
+          return {...existingEntry, ...newObject};
```
adityamukho, line 140: This is a shallow merge - hope it is sufficient for this use case?

adityamukho (commented):
IMO the ideal place to store custom user-specific data is in the `extra` object of the user's profile. This is read by `GET _api/user/{user}` and is patchable (by first merging the extra object with new data on the client side) using `PATCH _api/user/{user}`.

The benefit is that this endpoint is documented.

Please see my other comments inline.

js/apps/system/_admin/aardvark/APP/frontend/js/models/userConfig.js
ajurdak: After research in our specific case this solution is okay as we always save the whole object and never only change one specific property.

ajurdak (approved):
LGTM! I tested the functionality a lot.

nikhil-varma (approved):

adityamukho (approved):
LGTM

PR #16892, opened 2022-08-25
[CINFRA-427] create collection group

mchacki:
### Scope & Purpose

*First step to replace the creation collection api with a property transporting struct.
This PR should only replace the parser on the rest handler, but keep the API itself intact.
Also all following APIs should be unmodified.*


arangod/RestHandler/RestCollectionHandler.cpp
```diff
@@ -341,7 +342,45 @@ void RestCollectionHandler::handleCommandPost() {
...
     return;
   }
+  auto& cluster = _vocbase.server().getFeature<ClusterFeature>();
+  bool waitForSyncReplication = _request->parsedValue(
+      "waitForSyncReplication", cluster.createWaitsForSyncReplication());
+
+  bool enforceReplicationFactor =
+      _request->parsedValue("enforceReplicationFactor", true);
+
+#if true
```
markuspf, line 352: wut?

markuspf: That is to say, just delete the obsolete code underneath if it's not used

tests/js/common/replication_api/shell-api-replication-global.js
```diff
@@ -468,9 +468,9 @@ function dealing_with_wal_access_apiSuite () {
         fromTick = doc.parsedBody["tick"];
 
         // create collection;
-        cid = db._create("UnitTestsReplication", { waitForSync: true, globallyUniqueId: true });
+        cid = db._create("UnitTestsReplication", {waitForSync: true});
         cuid = cid.properties()["globallyUniqueId"];
-
+
```
markuspf, line 473: eslint will be very sad with you
```suggestion

```

arangod/VocBase/Properties/PlanCollection.h
```diff
@@ -0,0 +1,194 @@
...
+          f.field("cacheEnabled", planCollection.cacheEnabled).fallback(false),
+          f.field("isVolatile", planCollection.isVolatile).fallback(false),
+          f.field("syncByRevision", planCollection.syncByRevision)
+              .fallback(true),
+          f.field("usesRevisionsAsDocumentIds",
+                  planCollection.usesRevisionsAsDocumentIds)
+              .fallback(true),
+          f.field("numberOfShards", planCollection.numberOfShards)
+              .fallback(1ULL)
+              .invariant([](auto const& value) -> inspection::Status {
```
markuspf, line 120: Numpty check; when this invariant fails, will the error message tell you which value needs to be >0?

arangod/VocBase/Properties/PlanCollection.h
```diff
@@ -0,0 +1,194 @@
...
+                return {"Value has to be > 0"};
+              }),
+          f.field("replicationFactor", planCollection.replicationFactor)
+              .fallback(1ULL)
+              .invariant([](auto const& value) -> inspection::Status {
+                if (value > 0) {
+                  return inspection::Status::Success{};
+                }
+                return {"Value has to be > 0"};
+              }),
```
markuspf, line 152: This invariant might want to be its own function, as it's used multiple times.

markuspf (approved):
LGTM, bar some minor comments.

arangod/RestHandler/RestCollectionHandler.cpp
mchacki: totally agree, I just keep this in until Jenkins agrees with me that this is all green ;)
Before that I like to be able to switch to old code to see difference in the follow up method call
SO just pretend the else case is gone ;)

arangod/VocBase/Properties/PlanCollection.h
mchacki: Yes it does:
```
http://localhost:8529/_api/collection -d '{"name": "test", "numberOfShards": -2}'
{"code":400,"error":true,"errorMessage":"Number out of range on path numberOfShards","errorNum":10} 
```

```
http://localhost:8529/_api/collection -d '{"name": "test", "numberOfShards": 0}'
{"code":400,"error":true,"errorMessage":"Value has to be > 0 on path numberOfShards","errorNum":10}
```

apetenchea (approved):
I got an overview, didn't look into the details. Kudos for the Inspector work, that looks really cool!

mpoeter (approved):

mchacki:
Manually checked all red Jenkins items.
on core tests, the errors are known, and fixed in devel, this branch is lagging behind a week.
On driver test there is a known unrelated error (also in devel), everything else is blue.

PR #16766, opened 2022-08-11
Bug fix/bts 722

hkernbach:
### Scope & Purpose

When using:
- `SHORTEST_PATH`
- `K_SHORTEST_PATHS`
- `K_PATHS`
- `ALL_SHORTEST_PATHS`

and warnings are being generated during AQL Query execution, 
the warnings all reported with shortest path type "SHORTEST_PATH", instead of properly 
reporting the specific correct type.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :book: CHANGELOG entry made
- [x] Backports
  - [ ] Backport for 3.10: *(yes required)*

#### Related Information

- [ ] Enterprise PR: No enterprise PR
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-722

PR #16894, opened 2022-08-25
Bug fix/bts 722 (3.10)

hkernbach:
Backport of: https://github.com/arangodb/arangodb/pull/16766

NO EE companion.


arangod/Graph/PathType.cpp
```diff
@@ -33,6 +33,7 @@ namespace graph {
 constexpr char const* KShortestPathsName = "K_SHORTEST_PATHS";
 constexpr char const* KPathsName = "K_PATHS";
 constexpr char const* AllShortestPathsName = "ALL_SHORTEST_PATHS";
+constexpr char const* ShortestPathName = "SHORTEST_PATH";
```
jsteemann, line 36: nit: the constant naming here is inconsistent.
According to our coding conventions, all names should  `k...Name`.
Apart from that, these `char const*`s should be `std::string_views`.
That doesn't need to be fixed in 3.10, but can be addressed in devel only.

arangod/Graph/PathType.h
```diff
@@ -27,7 +27,12 @@ namespace arangodb {
 namespace graph {
 
 struct PathType {
-  enum class Type { KShortestPaths = 0, KPaths = 1, AllShortestPaths = 2 };
+  enum class Type {
+    KShortestPaths = 0,
+    KPaths = 1,
+    AllShortestPaths = 2,
+    ShortestPath = 3
```
jsteemann, line 34: Nit: according to our naming convention, all enum values should be `k...`.
This can be adjusted in devel only.

jsteemann (changes requested):
LGTM. Would be nice if the code comments could be addressed in devel later.

jsteemann (approved):

neunhoef (approved):
LGTM

KVS85:
@joerg84 will communicate https://github.com/arangodb/arangodb/pull/16894#pullrequestreview-1086869622 and https://github.com/arangodb/arangodb/pull/16894#pullrequestreview-1086868382 to @hkernbach.

PR #16897, opened 2022-08-25
fix dumping of active failover agency

dothebart:
### Scope & Purpose

at the end of the test we should dump the agency - just in case. This fixes the authentification to fix dumping.

- [x] :hankey: Bugfix


KVS85 (approved):
LGTM.

dothebart:
only unrelated issues.

PR #16913, opened 2022-08-26
Always try to join threads in tests

MBkkt:
### Scope & Purpose

*(Please describe the changes in this PR for reviewers, motivation, rationale - **mandatory**)*

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #16918, opened 2022-08-29
Introduce ThreadGuard 

markuspf:
### Scope & Purpose

In #16913 @MBkkt addressed an issue where threads created in a test were potentially not joined if an exception was thrown and the test scope was left.

This PR attempts to improve the suggested code a little bit by introducing a `ThreadGuard`, which is just a struct that wraps a vector of threads and joins them on request and on destruction.

- [X] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/lib/Basics/ThreadGuard.h
+++ b/lib/Basics/ThreadGuard.h
@@ -1,0 +1,65 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2021-2021 ArangoDB GmbH, Cologne, Germany
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <thread>
+#include <vector>
+
+namespace arangodb {
+
+// A thread guard is a wrapper around a std::vector<std::thread> that
+// facilitates joining all joinable threads on destruction or when
+// the member function joinAll is called.
+struct ThreadGuard {
+  ThreadGuard() = default;
+  explicit ThreadGuard(size_t reserve) { threads.reserve(reserve); }
+  ~ThreadGuard() {
+    // Note that this might throw, and then we're in trouble;
+    // We accept this risk here, because if joining a thread
+    // inside the destructor of ThreadGuard fails, we're in a
+    // very bad place, regardless.
+    joinAll();
+  }
+
+  template<typename Function, typename... Args>
+  auto emplace(Function&& f, Args&&... args)
+      -> std::vector<std::thread>::reference {
+    return threads.emplace_back(std::forward<Function>(f),
+                                std::forward<Args>(args)...);
+  }
+
+  void joinAll() {
+    for (auto& t : threads) {
+      if (t.joinable()) {
+        t.join();
+      }
+    }
+    threads.clear();
+  }
+
+  [[nodiscard]] auto size() const -> size_t { return threads.size(); }
+
+  std::vector<std::thread> threads;
+};
+
+}  // namespace arangodb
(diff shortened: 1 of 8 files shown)


lib/Basics/ThreadGuard.h
```diff
@@ -0,0 +1,54 @@
...
+//
+struct ThreadGuard {
+  ThreadGuard() = default;
+  explicit ThreadGuard(size_t reserve) : threads(reserve) {}
+  ~ThreadGuard() {
+    joinAll();  // Should we really?
+  }
+
+  template<typename Function, typename... Args>
+  auto emplace(Function&& f, Args... args) -> void {
```
mpoeter, line 37:
```suggestion
  auto emplace(Function&& f, Args&&... args) -> void {
```

lib/Basics/ThreadGuard.h
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  void joinAll() {
+    // TODO: Copypasta
+    for (auto& t : threads) {
+      if (t.joinable()) {
+        try {
+          t.join();
+        } catch (...) {
+        }
```
mpoeter, line 48: Do we really want to silently swallow all exceptions? IMHO join should normally never throw, but if it does we would want to know!

markuspf: I only copied this code from the original because I wanted to have exactly this discussion;

If we do not `try-catch(...)`, and call `joinAll` in the destructor, it could throw; arguably, if the `joinAll()` throws, everything is in a bad state in any case so maybe we accept that our destructor throws

lib/Basics/ThreadGuard.h
```diff
@@ -0,0 +1,54 @@
...
+  auto emplace(Function&& f, Args&&... args) -> void {
+    threads.emplace(std::forward<Function>(f), std::forward<Args...>(args...));
+  }
+
+  void joinAll() {
+    for (auto& t : threads) {
+      if (t.joinable()) {
+        t.join();
+      }
+    }
```
MBkkt, line 50: Suggest clear threads

MBkkt (approved):

lib/Basics/ThreadGuard.h
MBkkt: In general, nice, thanks!

lib/Basics/ThreadGuard.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <thread>
+#include <vector>
+
+// A thread guard is a wrapper around a std::vector<std::thread> that
+// facilitates joining all joinable threads on destruction or when
+// the member function joinAll is called.
+struct ThreadGuard {
+  ThreadGuard() = default;
+  explicit ThreadGuard(size_t reserve) : threads(reserve) {}
```
MBkkt, line 33:
```suggestion
  explicit ThreadGuard(size_t count) {
    threads.reserve(count);
  }
```

lib/Basics/ThreadGuard.h
```diff
@@ -0,0 +1,60 @@
...
+    return threads.emplace_back(std::forward<Function>(f),
+                                std::forward<Args>(args)...);
+  }
+
+  void joinAll() {
+    for (auto& t : threads) {
+      if (t.joinable()) {
+        t.join();
+      }
+    }
```
MBkkt, line 54:
```suggestion
      }
    }
    threads.clear();
```

MBkkt (changes requested):
In ctor now you create empty thread handlers and then push not empty

lib/Basics/ThreadGuard.h
markuspf: Fixed.

lib/Basics/ThreadGuard.h
markuspf: also fixed.

MBkkt (approved):
Thanks.

But honestly I don't understand few points (them all not critical)
1) Why do we need unused(?) and error prone (user can expect size() != 0 after ctor) size()
2) return thread& from push/emplace, also unused now
3) push_back, unused, and emplace did the same things.

markuspf:
> In ctor now you create empty thread handlers and then push not empty

I believe I fixed this by calling `reserve` like you suggested.
 
> For this reason I also think better to remove size() const method

Can you explain why you would like the `size()` method to be removed? 

It is used in one of the tests to determine the number of threads that were added;

Admittedly with the bug in the constructor this number would have been wrong, but with a fixed constructor this should be fine, shouldn't it?

MBkkt:
Probably some miscommunication because history of notification.

I wrote about all this in approve message.

About size, now I see test that used it. So I'm ok with it, but probably test make useless checks.
You can ignore it, I'm not insist.

markuspf:
>     3. push_back, unused, and emplace did the same things.

I was trying to make the test that used `std::back_inserter` work, but there are other issues that made me revert to a plain `for` loop for now, so I removed the `push_back` function again.

PR #16922, opened 2022-08-29
[PREGEL] Extract all conductor stuff into separate dir

jvolmer:
This is a moving and renaming refactoring PR:

- Moves Conductor.cpp and Conductor.hpp into Conductor directory
- Adds a CMakeLists for the Worker directory


arangod/Pregel/WorkerInterface.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
```
romanatarango, line 1: This file doesn't seem to be used anywhere.

romanatarango (approved):
`arangod/Pregel/WorkerInterface.h` does not seem to be used anywhere and the description does not say that also the states are moved into their own directory. Otherwise LGTM.

PR #16927, opened 2022-08-30
add a way to generate stacktraces and coredumps via the debugger

dothebart:
### Scope & Purpose

so far we would use the default coredump generation process in order to generate coredumps. 
This 
- doesn't always work on mac
- generates coredumps regardless of the expected filesize
- if one wouldn't generate the coredump, all information would be lost hence attach the debugger to the process

- [x] :pizza: New feature
- [x] Test infrastructure


KVS85:
- [ ] https://jenkins.arangodb.biz/view/Mac/job/arangodb-matrix-pr-mac.x86-64/84/

KVS85 (approved):
LGTM.

PR #16932, opened 2022-08-30
Fix stack-use-after-scope in fuertetest

markuspf:
When introducing the `ThreadGuard` I changed a lambda capture from `[=]` to
`[&]` overlooking an implicitly captured variable.

This patch changes the code to explicitly capture variables.

### Scope & Purpose

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/tests/Fuerte/ConnectionConcurrentTest.cpp
+++ b/tests/Fuerte/ConnectionConcurrentTest.cpp
@@ -77,10 +77,10 @@ TEST_P(ConcurrentConnectionF, ApiVersionParallel) {
 
   auto joins = ThreadGuard(threads());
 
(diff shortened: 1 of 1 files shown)


tests/Fuerte/ConnectionConcurrentTest.cpp
```diff
@@ -77,10 +77,10 @@ TEST_P(ConcurrentConnectionF, ApiVersionParallel) {
 
   auto joins = ThreadGuard(threads());
 
+  const size_t rep = repeat();
```
MBkkt, line 80: I suggest just move rep get before joins creation

jsteemann (approved):
LGTM!

PR #16941, opened 2022-08-30
[PREGEL] Use future to prepare global super step

jvolmer:
### Scope & Purpose

Use future to prepare the global super step

Move ConductorWorkers into their own files and in the conductor namespace and rename them to ClusterWorkerApi and SingleServerWorkerApi.


romanatarango (changes requested):
In general lgtm. The only thing: the channel file is now empty and the the *Api.h/cpp files are not mentioned in the description.

PR #16944, opened 2022-08-31
Moved escaping control and unicode chars

cpjulia:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1076

This PR moves the code from escaping control and unicode chars to the Logger.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests (there were already tests)
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
   No backports required.

#### Related Information

(Obs.: as an internal change, don't think docs are needed)

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1076
- [ ] Design document:


tests/Logger/EscaperTest.cpp
```diff
@@ -59,138 +59,129 @@ class EscaperTest : public ::testing::Test {
...
-  auto buffer = std::make_unique<char[]>(messageSize);
-  char* output = buffer.get();
+                          EscaperType& escaper) {
+  std::string output;
   escaper.writeIntoOutputBuffer(inputString, output);
-  size_t outputBufferSize = output - buffer.get();
-  std::string outputString(buffer.get(), outputBufferSize);
-  EXPECT_EQ(outputString.compare(expectedOutput), 0);
-  EXPECT_EQ(outputString, expectedOutput);
+  EXPECT_EQ(output.compare(expectedOutput), 0);
```
jsteemann, line 65: why `EXPECT_EQ(a.compare(b), 0)` instead of simply `EXPECT_EQ(a, b)`?

lib/Logger/LogAppenderFile.cpp
```diff
@@ -55,48 +54,21 @@ int LogAppenderFile::_fileMode = S_IRUSR | S_IWUSR | S_IRGRP;
...
   char* output = _buffer.get();
-  _escaper->writeIntoOutputBuffer(message, output);
-  *output++ = '\n';
-  *output = '\0';
-  return (output - _buffer.get());
+
+  strncpy(output, formattedMsg.data(), formattedMsg.size());
+  output[formattedMsg.size()] = '\0';
+
+  return (output + strlen(output) - _buffer.get());
```
jsteemann, line 66: This will create a new std::string for each message, plus a new temporary one, and then call `strlen()` in addition. That looks like horrible performance. This is a critical part of code that needs to be a fast as possible.

lib/Logger/Logger.cpp
```diff
@@ -663,6 +666,25 @@ void Logger::log(char const* logid, char const* function, char const* file,
...
+      }
+    } else {
+      if (_useUnicodeEscaped) {
+        escaper = std::make_unique<
+            Escaper<ControlCharsSuppressor, UnicodeCharsEscaper>>();
+      } else {
+        escaper = std::make_unique<
+            Escaper<ControlCharsSuppressor, UnicodeCharsRetainer>>();
+      }
+    }
```
jsteemann, line 686: It looks like we are now creating one Escaper instance on the heap for each log message. We shouldn't do that.
The logging is a critical part of code that needs to be a fast as possible.

markuspf (approved):

PR #16945, opened 2022-08-31
fix random haustier bug

jsteemann:
### Scope & Purpose

Fix a bug randomly reported via Slack about "Haustiere" results.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17024
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


markuspf (approved):
LGTM

alexbakharew (approved):
LGTM

gnusi (changes requested):
It's a breaking change for `ALL`

jsteemann:
> It's a breaking change for `ALL`

Yes, it is definitely a breaking change, but a sensible one.
Consider the following behavior of ArangoDB <= 3.10:

```
arangod> db._query("RETURN @values ALL > 3", { values: [] });
[ true ]

arangod> db._query("RETURN @values ANY > 3", { values: [] });
[ false ]
```

I find that pretty inconsistent. You ask for whether all values of an empty array are > 3 and get `true` back, and then you ask whether any value of an empty array is > 3 and get `false` back. That's not truly Vulcan logic.
So this PR proposes to change the behavior for ALL for empty arrays, so that it makes actual sense and is in line with the behavior of the ANY operator.

markuspf:
If you read "ANY > 3" as "there exists a value x such that x > 3", `false` is actually the correct answer, as there is no value ;)

jsteemann:
> If you read "ANY > 3" as "there exists a value x such that x > 3", `false` is actually the correct answer, as there is no value ;)

True. 
I think we need to agree on the desired behavior, and from that follows whether we need to make a downwards-incompatible change or not.
I am not a mathematician, so I don't know if there is a commonly accepted way of how such operations are supposed to work.

romanatarango:
> > If you read "ANY > 3" as "there exists a value x such that x > 3", `false` is actually the correct answer, as there is no value ;)
> 
> True. I think we need to agree on the desired behavior, and from that follows whether we need to make a downwards-incompatible change or not. I am not a mathematician, so I don't know if there is a commonly accepted way of how such operations are supposed to work.

There is one for ALL. ALL should be read as "every that exists", so if there is none (in the empty set), then ALL is true, regardless of the statement about it. 

ANY is considered (by some logicians and I agree with them for our case) to be sometimes unclear. Unless it is explicitly defined for some area, of course. It can be interpreted as ALL (see above), as EXISTS, which means, well, there is an element with the described property, and, if used with a negation, as NO, which means, there is no element with the property.

romanatarango (commented):
ALL can only be understood as "all that exist"; on the empty set this means: always true. ANY should mean "exists" in our case (the general use is controversial) as we don't have any EXISTS. As the empty set has no element, ANY should mean on it: false.

To sum up: the current state (without applying the PR) is correct in this respect and shouldn't be changed.

jsteemann:
Ok, I modified the PR so that the behavior regarding `ALL` does not change anymore.

romanatarango (approved):
LGTM

PR #16975, opened 2022-09-02
Reduce build time by removing unnecessary includes from header files

mpoeter:
### Scope & Purpose


jsteemann:
Further optimization: https://github.com/arangodb/arangodb/pull/16977

arangod/Cluster/AgencyCache.cpp
```diff
@@ -26,6 +26,7 @@
 #include "Agency/AgencyStrings.h"
 #include "Agency/Node.h"
 #include "ApplicationFeatures/ApplicationServer.h"
+#include "Cluster/AgencyCallback.h"
 #include "Basics/application-exit.h"
```
jsteemann, line 30:
```suggestion
#include "Basics/application-exit.h"
#include "Cluster/AgencyCallback.h"
```

arangod/Cluster/ClusterInfo.h
```diff
@@ -27,279 +27,65 @@
...
 #include "Metrics/Fwd.h"
-#include "Metrics/ClusterMetricsFeature.h"
 #include "Replication2/AgencyCollectionSpecification.h"
-#include "Replication2/ReplicatedLog/LogCommon.h"
-#include "VocBase/Identifiers/IndexId.h"
-#include "VocBase/LogicalCollection.h"
-#include "VocBase/VocbaseInfo.h"
-#include "VocBase/voc-types.h"
-#include "VocBase/vocbase.h"
+//#include "Replication2/ReplicatedLog/LogCommon.h"
```
jsteemann, line 47:
```suggestion
```

arangod/VocBase/voc-types.h
```diff
@@ -23,16 +23,7 @@
 
 #pragma once
 
-#include <cstdint>
```
jsteemann, line 26: Including `cstdint` seems to make sense here, at least a few lines below we are referring to `uint64_t`. I think we should keep this include here.

jsteemann (approved):
LGTM in general. See comments.

arangod/VocBase/voc-types.h
mpoeter: Well, `cstdint` defines `std::unint64_t`. Since we use `uint64_t`,  we should actually include `stdint.h`.

MBkkt: Yeah, it's awful. But c-headers, ".h" was depricated, so imo better to use std::size_t/uint/int/etc

jsteemann (approved):
LGTM!

MBkkt (approved):

jsteemann (approved):
LGTM!

PR #16978, opened 2022-09-02
Make assertions in soft shutdown cluster test more stable.

markuspf:
### Scope & Purpose

This PR attempts to address some test failures seen in CI due to assertions triggering when there was an unexpected number of transactions (2 vs 1);

Since the test cannot be certain how many transactions are open, as other workers on the cluster might have opened transactions, assert that there is more than one transaction present.

Also, add some messages on the assertions so that future failures are more obvious to categorise.


neunhoef (approved):
LGTM.

PR #16989, opened 2022-09-05
Delete async mode

jvolmer:
Deletes async code - including AsyncSCC. It is not tested and the code is hard to follow with it because it is hacked in.


romanatarango (approved):
Please, add a corresponding line to CHANGELOG and take care of the documentation to be changed.

markuspf (approved):
LGTM

PR #17002, opened 2022-09-06
Documentation 3.10/inverted index

Simran-B:
Describe the inverted index variant of the `/_api/index` endpoint and the `/_api/view` endpoints for `search-alias` Views


Simran-B:
The `POST /_api/index` endpoint for the inverted index has a rendering issue:
In the description of the top-level `trackListPositions` property, there is a `</div>` below the hint box for unknown reasons:

image

Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
```diff
@@ -0,0 +1,371 @@
...
+The default is an empty array.
+
+@RESTBODYPARAM{includeAllFields,boolean,optional,}
+If set to `true`, then all document attributes are indexed, excluding
+any nested attributes that are configured in the `fields` array.
+The `analyzer` and `features` properties apply to the nested fields.
+
+The default is `false`.
+
+{% hint 'warning' %}
```
ansoboleva, line 192: Doesn't it need @HINTS header?

Simran-B: `@HINTS` is only for adding admonitions to the top-level of a Rest DocuBlock. It can't be used inside of a parameter description.

Now that I think about it, `{% hint %}` shouldn't be used at all in Rest DocuBlocks other than in `@HINTS` because it won't get rendered in Swagger UI. For ArangoSearch View links, we don't actually have this hint in the DocuBlock, the DocuBlock merely links to the Manual with this hint, where it's perfectly fine to use it:
- https://www.arangodb.com/docs/stable/http/views-arangosearch.html#create-an-arangosearch-view
- https://www.arangodb.com/docs/stable/arangosearch-views.html#link-properties

Documentation/DocuBlocks/Rest/Indexes/post_api_index_fulltext.md
```diff
@@ -17,6 +17,12 @@ The collection name.
 @RESTBODYPARAM{type,string,required,string}
 must be equal to *"fulltext"*.
 
+@RESTBODYPARAM{name,string,optional,string}
+An easy-to-remember name for the index to look it up or refer to it in index hints.
+Index names are subject to the same character restrictions as collection names.
+If omitted, a name is auto-generated so that it is unique with respect to the
+collection, e.g. `idx_832910498`.
+
```
ansoboleva, line 25: test comment

Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
```diff
@@ -0,0 +1,373 @@
...
+
+Sub-properties:
+  - `type` (string, _optional_):
+    The segment candidates for the "consolidation" operation are selected based
+    upon several possible configurable formulas as defined by their types.
+    The currently supported types are:
+    - `"tier"` (default): consolidate based on segment byte size and live
+      document count as dictated by the customization attributes. If this type
+      is used, then below `segments*` and `minScore` properties are available.
+    - `"bytes_accum"`: consolidate if and only if
```
Dronplane, line 300: AFAIK we wanted to mark bytes_accum as deprecated. Let's mention this or even remove it here.

Dronplane (changes requested):
For search-alias view there are several restriction for the combined indexes: same fields should be indexed with the same analyzers and so on. Suggest to mention this in the docs.

Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
Simran-B: Removing, so that people don't start using inverted indexes with a deprecated option.

Dronplane (approved):
LGTM

ansoboleva (approved):
reviewed

ansoboleva:
https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/2407/ - verified with jenkins

PR #17004, opened 2022-09-06
Fix waitForSync for non-maintainer build

MBkkt:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17005

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/IResearch/Search.cpp
+++ b/arangod/IResearch/Search.cpp
@@ -480,7 +480,9 @@ ViewSnapshot::Links Search::getLinks() const {
   indexes.reserve(_indexes.size());
   for (auto const& [_, handles] : _indexes) {
     for (auto const& handle : handles) {
-      indexes.push_back(handle->lock());
+      if (auto index = handle->lock(); index) {
+        indexes.push_back(std::move(index));
+      }
     }
   }
   return indexes;
--- a/arangod/IResearch/ViewSnapshot.cpp
+++ b/arangod/IResearch/ViewSnapshot.cpp
@@ -48,7 +48,7 @@ class ViewSnapshotCookie final : public ViewSnapshot,
 
   void clear() noexcept;
 
-  bool compute(bool sync, std::string_view name);
+  void compute(bool sync, std::string_view name);
 
  private:
   [[nodiscard]] irs::sub_reader const& operator[](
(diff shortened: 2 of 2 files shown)


arangod/IResearch/Search.cpp
```diff
@@ -480,7 +480,9 @@ ViewSnapshot::Links Search::getLinks() const {
   indexes.reserve(_indexes.size());
   for (auto const& [_, handles] : _indexes) {
     for (auto const& handle : handles) {
-      indexes.push_back(handle->lock());
+      if (auto index = handle->lock(); index) {
```
Dronplane, line 483: Do we really need "; index" ?

Dronplane (approved):
LGTM. One minor comment

arangod/IResearch/Search.cpp
MBkkt: Yes! It's fix for single server. In single server then we get snapshot we use this method. If index already was deleted we shouldn't push it to snapshot (because acquiring snapshot will fail in this case)

MBkkt: Ok. I bad understand your comment

MBkkt: We can write it without ;index. But I don't like this style

Dronplane: So here you want to emphasise that "index" is checked?

MBkkt: it doesn't make much sense. 
I just prefer to write:
`if (auto p = ...; p)` instead of `if ((auto p = ...))`

jsteemann (approved):
LGTM!

arangod/IResearch/ViewSnapshot.cpp
```diff
@@ -176,7 +171,7 @@ ViewSnapshot* makeViewSnapshot(transaction::Methods& trx, void const* key,
 
   for (auto const& link : links) {
     if (!link) {
-      LOG_TOPIC("b054e", WARN, TOPIC)
+      LOG_TOPIC("fffff", WARN, TOPIC)
```
Simran-B, line 174: Shouldn't the ID remain the same?

MBkkt: Nope

MBkkt: This log message was copied by Jan (from compute), so now it cannot triggered in compute.

Jan should moved old logging (from compute to here).

I fix this.

But in general it's not important both check was added in 3.10

(I just like ffff more)

PR #17009, opened 2022-09-07
Fixes to `test-soft-shutdown-cluster.js`

markuspf:
### Scope & Purpose

Addressing some structural issues with the soft shutdown tests

 * Use the instance manager throughout to make sure that coordinators used in a test are left in a consistent state
 * Split out the Pregel test from the file, as it has separate issues; this way they are more visible
 * Minor fix to the Pregel test so that it converges in due course.


dothebart (approved):
LGTM, thanks!

PR #17010, opened 2022-09-07
[CINFRA-511] Improve transaction handling in leader state

mpoeter:
### Scope & Purpose

Abort transactions and register tombstones in the transaction manager for any remaining transactions when the leader resigns, as well as transactions that are still pending after finishing recovery.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**


apetenchea (approved):
Looks good, modulo the Jenkins failure

PR #17016, opened 2022-09-07
Prefetch for revision trees- BTS-989

neunhoef:
This PR attacks the problem described in
  https://arangodb.atlassian.net/browse/BTS-989
The problem is that if one orders a batch (aka
RocksDBReplicationContext), which includes a snapshot of RocksDB, it is
not guaranteed that one can subsequently fetch a revision tree for some
collection/shard which corresponds to the sequence number of the
snapshot in this batch. This is because the revision tree of the
collection might already have been rolled forward.

This PR makes it so that in case a batch is bound to a specific
collection (which happens when SynchronizeShard on the follower
tries to get in sync with the leader), then the revision tree for
that collection is prefetched and buffered in the batch.

This means that later when the revision tree is requested by the
follower, it is guaranteed that it is still available, and that we do
not have to stop revision trees from being rolled forward.

- Implement prefetch.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests (code already covered)
  - [*] **integration tests**
  - [*] **resilience tests**
- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17018
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17019
  - [ ] Backport for 3.8:

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-989


arangod/RocksDBEngine/RocksDBRestReplicationHandler.cpp
```diff
@@ -888,13 +888,24 @@ void RocksDBRestReplicationHandler::handleCommandRevisionTree() {
...
     RocksDBReplicationContext* c = _manager->find(ctx.batchId);
     RocksDBReplicationContextGuard guard(_manager, c);
+    if (c != nullptr) {  // we have the RocksDBReplicationContext!
+      // See if we can get the revision tree from the context:
+      tree = c->getPrefetchedRevisionTree(collectionName);
+      // This might return a nullptr!
+    }
+    if (tree == nullptr) {  // Still not there, try to get it directly:
+      tree = ctx.collection->getPhysical()->revisionTree(
+          c->snapshot()->GetSequenceNumber());
```
jsteemann, line 905: In theory, `snapshot()` can return a nullptr. While it is very unlikely that it does here, it is still possible. So we should check here if the returned snapshot is a nullptr or not.

arangod/RocksDBEngine/RocksDBRestReplicationHandler.cpp
```diff
@@ -888,13 +888,24 @@ void RocksDBRestReplicationHandler::handleCommandRevisionTree() {
...
+    }
     if (c != nullptr) {
-      c->removeBlocker(_request->databaseName(), _request->value("collection"));
+      c->removeBlocker(_request->databaseName(), collectionName);
```
jsteemann, line 908: It is to me what the exact contents of `collectionName` should be here. Is it a shard name or a collection guid?
It would be good to document the expected contents here, so that one does not mix shard names and guids by accident.

arangod/RocksDBEngine/RocksDBV8Functions.cpp
```diff
@@ -298,7 +298,7 @@ static void JS_CollectionRevisionTreeVerification(
     try {
       auto* physical = toRocksDBCollection(*collection);
       auto batchId = ctx->id();
-      storedTree = physical->revisionTree(batchId);
+      storedTree = physical->revisionTree(ctx->snapshot()->GetSequenceNumber());
```
jsteemann, line 301: Again, in theory the `snapshot()` function can return a nullptr, and we should check the return value here before dereferencing it.

neunhoef: Good point, I have added a check.

arangod/RocksDBEngine/RocksDBRestReplicationHandler.cpp
neunhoef: I have renamed the variables in all places to say "guid" and documented this.

arangod/RocksDBEngine/RocksDBRestReplicationHandler.cpp
neunhoef: Good point, added a check,will fail if there is no snapshot.

jsteemann (approved):
LGTM

PR #17019, opened 2022-09-08
Prefetch for revision trees - BTS-989

neunhoef:
This PR attacks the problem described in
  https://arangodb.atlassian.net/browse/BTS-989
The problem is that if one orders a batch (aka
RocksDBReplicationContext), which includes a snapshot of RocksDB, it is
not guaranteed that one can subsequently fetch a revision tree for some
collection/shard which corresponds to the sequence number of the
snapshot in this batch. This is because the revision tree of the
collection might already have been rolled forward.

This PR makes it so that in case a batch is bound to a specific
collection (which happens when SynchronizeShard on the follower
tries to get in sync with the leader), then the revision tree for
that collection is prefetched and buffered in the batch.

This means that later when the revision tree is requested by the
follower, it is guaranteed that it is still available, and that we do
not have to stop revision trees from being rolled forward.

- Implement prefetch.
- CHANGELOG.

This is the backport for 3.9.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests (code already covered)
  - [*] **integration tests**
  - [*] **resilience tests**
- [*] Backports
  - [*] This is the backport for 3.9

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-989
- [*] Devel PR: https://github.com/arangodb/arangodb/pull/17016

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,13 @@
 v3.9.4 (XXXX-XX-XX)
 -------------------
 
+* Implement prefetch for revision trees, in case a batch is created with
+  a distinguished collection as for `SynchronizeShard`. This ensures that
+  the revision tree for the batch will be available when needed, even though
+  the revision tree for the collection might already have advanced beyond
+  the sequence number of the snapshot in the batch. This ensures that
+  shards can get in sync more reliably and more quickly.
+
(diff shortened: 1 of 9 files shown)


kvahed (approved):
LGTM

PR #17036, opened 2022-09-09
[PREGEL] Send general messages between worker and conductor

jvolmer:
Adds a general message (and a response) that includes an std::variant of different message types. This way, message handling in the feature becomes more readable: We can use the same Rest path for all message types and Feature::handleWorkerRequests delegates the decision what happens with the message to the worker. The worker does pattern matching (with std::visit) on the message type to decide what to do with which message type.
This PR uses this general messages only where we already use futures. A second PR will convert the rest.

### DIFF
--- a/arangod/Pregel/Conductor/ClusterWorkerApi.cpp
+++ b/arangod/Pregel/Conductor/ClusterWorkerApi.cpp
@@ -1,22 +1,47 @@
 #include "ClusterWorkerApi.h"
+#include <variant>
 
 #include "Pregel/WorkerConductorMessages.h"
 #include "Pregel/NetworkConnection.h"
 
 using namespace arangodb::pregel::conductor;
 
+// This template enforces In and Out type of the function:
+// In enforces which type of message is sent
+// Out defines the expected response type:
+// The function returns an error if this expectation is not fulfilled
+template<typename Out, typename In>
+auto ClusterWorkerApi::execute(In const& in) -> futures::Future<ResultT<Out>> {
+  return _connection
+      .post(ModernMessage{.executionNumber = _executionNumber, .payload = {in}})
+      .thenValue([](auto&& response) -> futures::Future<ResultT<Out>> {
+        if (response.fail()) {
+          return Result{response.errorNumber(), response.errorMessage()};
+        }
+        if (!std::holds_alternative<ResultT<Out>>(response.get().payload)) {
+          return Result{
+              TRI_ERROR_INTERNAL,
+              fmt::format(
+                  "Message from worker does not include the expected {} "
+                  "type",
+                  typeid(Out).name())};
+        }
+        return std::get<ResultT<Out>>(response.get().payload);
+      });
+}
+
 auto ClusterWorkerApi::loadGraph(LoadGraph const& graph)
     -> futures::Future<ResultT<GraphLoaded>> {
-  return _connection.post<GraphLoaded>(graph);
+  return execute<GraphLoaded>(graph);
 }
 
 auto ClusterWorkerApi::prepareGlobalSuperStep(
     PrepareGlobalSuperStep const& data)
     -> futures::Future<ResultT<GlobalSuperStepPrepared>> {
-  return _connection.post<GlobalSuperStepPrepared>(data);
+  return execute<GlobalSuperStepPrepared>(data);
 }
 
 auto ClusterWorkerApi::runGlobalSuperStep(RunGlobalSuperStep const& data)
     -> futures::Future<ResultT<GlobalSuperStepFinished>> {
-  return _connection.post<GlobalSuperStepFinished>(data);
+  return execute<GlobalSuperStepFinished>(data);
 }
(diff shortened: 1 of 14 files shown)


arangod/Pregel/WorkerConductorMessages.h
```diff
@@ -279,4 +256,78 @@ auto inspect(Inspector& f, CollectPregelResults& x) {
...
       f.field("withId", x.withId).fallback(false));
 }
+
+// ---------------------- modern message ----------------------
+
+using MessagePayload =
+    std::variant<LoadGraph, PrepareGlobalSuperStep, RunGlobalSuperStep>;
+using ResponsePayload =
+    std::variant<ResultT<GraphLoaded>, ResultT<GlobalSuperStepPrepared>,
+                 ResultT<GlobalSuperStepFinished>>;
```
markuspf, line 266: I know I am being annoying, but could this be a `ResultT<std::variant<GraphLoaded, GlobalSuperStepPrepared, GlobalSuperstepFinished>`?

arangod/Pregel/Conductor/ClusterWorkerApi.cpp
```diff
@@ -1,22 +1,43 @@
 #include "ClusterWorkerApi.h"
+#include <variant>
 
 #include "Pregel/WorkerConductorMessages.h"
 #include "Pregel/NetworkConnection.h"
 
 using namespace arangodb::pregel::conductor;
 
+template<typename Out, typename In>
```
markuspf, line 9: A comment about why this is templated would not be too bad; The `In` parameter can be used to enforce what type of message to send, the `Out` parameter is used to say which type of response to expect.

(For context: my first thought was that one can just use `std::variant`throughout and be done with it, but in principle, statically asserting what type of message is expected as a response isn't a bad idea.

arangod/Pregel/WorkerConductorMessages.h
```diff
@@ -279,4 +256,78 @@ auto inspect(Inspector& f, CollectPregelResults& x) {
...
+    auto res = f.apply(v);
+    x = ResponsePayload{v};
+    return res;
+  } else {
+    auto v = ResponsePayloadSerializer{x};
+    return f.apply(v);
+  }
+}
+
+struct ModernMessage {
```
markuspf, line 312: Right now `ModernMessage` and `ModernResponse` are the same struct; 

Can you get away with just having a `ModernMessage`? Or in other words: Is the fact that something is a Response *used*?

markuspf: Further: I think at the moment you encode in the Payload already that something is a response; maybe this messaging layer should not assume a request-response model, but rather a "get my messages around" model?

markuspf (commented):
Really like how this is coming together; left a few comments we should discuss before merging.

arangod/Pregel/WorkerConductorMessages.h
jvolmer: Yes, it *is* relevant because the body in std::visit that is called in Worker::process has to be exhaustive. Here you just expect ModernMessages, no ModernResponses. (And it will be the other way around for the messages in the conductor as soon as I use the new messages in Feature::handleConductorRequest as well. 
Ok, you could pattern match only the relevant cases and do an else case for the rest, but isn't it more readable to separate the two? This way we always get a compile error if we forget to add a case when we added a new message to ModernMessage.

jvolmer: We could also call it differently: E.g. MessageFromWorker and MessageFromConductor

markuspf: I am going to double down on the "just get my messages around already" model ;) 

If you go with just `ModernMessage`, then not only do you not duplicate code here, the message handlers become more uniform. Ultimately I would like the the concept of `handleConductorRequest` to not exist anymore, I would like the Conductor to handle messages sent to it.

At some point you will have to write code that says "I did not expect this message", and in my opinion it should be the conductor's responsibility to do so; maybe we can provide some small transformation helper that specifies which messages a conductor wants to handle, and do the erroring outside transparently.

I would like it to be as simple as possible to add new message types (and also messaging paradigms). It looks to me like this split makes it harder to do so, or even impossible if we are boxed into "request-response"

arangod/Pregel/WorkerConductorMessages.h
jvolmer: Do also want this if we now smash together both Payloads?
Adding them together we would have 
```
using MessagePayload = std::variant<LoadGraph, PrepareGlobalSuperStep, RunGlobalSuperStep,  
                ResultT<GraphLoaded>, ResultT<GlobalSuperStepPrepared>, ResultT<GlobalSuperStepFinished>>;
```
With your suggestion we would e.g. wrap LoadGraph with an unneccessary ResultT. I don't see a case where it is necessary to wrap LoadGraph in an ResultT.
But perhaps you just want to make it the same for all messages, then we just do this unnecessary wrap, just wanted to mention it.

PR #17037, opened 2022-09-09
[PREGEL]Copy vertex property

romanatarango:
### Scope & Purpose

A new Pregel algorithm `readwrite` that, for each vertex, copies the value from one property of the vertex to another property. (The input property should exist and be a number, otherwise the algorithm throws an error.) The algorithm is implemented only for testing Pregel. In the background it performs (for testing) exactly `maxGSS` iterations of sending and getting messages, which are numbers, and aggregating them via `MaxAggregator`. The algorithm can be called with 
```js
var pregel = require("@arangodb/pregel");
var params = {sourceField: "input", resultField: "output", maxGSS: 36};
var execution = pregel.start("readwrite", "<graphname>", params);
```
or
```
shell> curl -X POST --header 'accept: application/json' --data-binary @- --dump - http://localhost:8529/_api/control_pregel <<EOF
{ 
  "algorithm" : "readwrite", 
  "graphName" : "<graph name>", 
  "params" : { 
    "maxGSS" : 36, 
    "sourceField" : "<input field name>" 
    "resultField" : "<output field name>" 
  } 
}
EOF
```
In addition, a new test tests the new algorithm.

- [x] :pizza: New feature
- [x] Tests
  - [x] **integration tests**
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1058


arangod/Pregel/AlgoRegistry.cpp
```diff
@@ -159,6 +162,9 @@ template<typename V, typename E, typename M>
   } else if (algorithm == "wcc") {
     return createWorker(vocbase, new algos::WCC(server, userParams), body,
                         feature);
+  } else if (algorithm == "readwrite") {
+    return createWorker(vocbase, new algos::ReadWrite(server, userParams), body,
+                        feature);
```
markuspf, line 167: Wonder whether one should:

```suggestion
#if defined(ARANGODB_ENABLE_MAINTAINER_MODE)
  } else if (algorithm == "readwrite") {
    return createWorker(vocbase, new algos::ReadWrite(server, userParams), body,
                        feature);
#endif
```

arangod/Pregel/AlgoRegistry.cpp
```diff
@@ -77,6 +78,8 @@ IAlgorithm* AlgoRegistry::createAlgorithm(
     return new algos::DMID(server, userParams);
   } else if (algorithm == "wcc") {
     return new algos::WCC(server, userParams);
+  } else if (algorithm == "readwrite") {
+    return new algos::ReadWrite(server, userParams);
```
markuspf, line 82:
```suggestion
#if defined(ARANGODB_ENABLE_MAINTAINER_MODE)
  } else if (algorithm == "readwrite") {
    return new algos::ReadWrite(server, userParams);
#endif
```

arangod/CMakeLists.txt
```diff
@@ -283,6 +283,7 @@ set(LIB_ARANGO_PREGEL_SOURCES
   Pregel/Algos/SSSP.cpp
   Pregel/Algos/ShortestPath.cpp
   Pregel/Algos/WCC.cpp
+  Pregel/Algos/ReadWrite.cpp
```
markuspf, line 286: if we exclude this from released builds this needs to be moved into an `if (USE_MAINTAINER_MODE) ` as welll

markuspf (commented):
Nice, thanks;

The only thing I wonder is whether one should disable this "algorithm" in maintainer mode.

arangod/Pregel/Algos/ReadWrite.cpp
```diff
@@ -0,0 +1,130 @@
...
+      : GraphFormat(server),
+        sourceFieldName{std::move(sourceFieldName)},
+        resultFieldName{std::move(resultFieldName)} {}
+  std::string const sourceFieldName;
+  std::string const resultFieldName;
+
+  void copyVertexData(arangodb::velocypack::Options const&,
+                      std::string const& documentId, VPackSlice document,
+                      V& targetPtr, uint64_t& /*vertexIdRange*/) override {
+    if (!document.hasKey(sourceFieldName)) {
```
gnusi, line 60: @romanatarango 
it's better to avoid `hasKey` + `get` pattern as key lookup happens twice in this case.
Better to use
```
auto const val = document.get(sourceFieldName);

if (val.isNone()) {
   // handle "no property" case
}
...
```

arangod/Pregel/Algos/ReadWrite.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+WorkerContext* ReadWrite::workerContext(VPackSlice userParams) const {
+  return new ReadWriteWorkerContext();
+}
+
+struct ReadWriteMasterContext : public MasterContext {
+  size_t maxGss;
+  explicit ReadWriteMasterContext(VPackSlice userParams) {
+    if (userParams.hasKey(Utils::maxGSS)) {
```
gnusi, line 113: same here

arangod/Pregel/Algos/ReadWrite.cpp
romanatarango: Done

arangod/Pregel/Algos/ReadWrite.cpp
romanatarango: Done

tests/js/common/shell/shell-pregel-components.js
```diff
@@ -418,14 +417,20 @@ const hitsTestSuite = pregelTestHelpers.makeHITSTestSuite(false, "", 4);
...
+// jsunity.run(componentsTestSuite);
+// jsunity.run(wccRegressionTestSuite);
+// jsunity.run(wccTestSuite);
+// jsunity.run(sccTestSuite);
+// jsunity.run(labelPropagationTestSuite);
+// jsunity.run(pagerankTestSuite);
+// jsunity.run(seededPagerankTestSuite);
+// jsunity.run(ssspTestSuite);
+// jsunity.run(hitsTestSuite);
 // jsunity.run(effectiveClosenessTestSuite);
```
jvolmer, line 431: Why is all this commented out?

romanatarango: Oh, thank you! It was for debugging and I forgot to un-comment it. Reverted now.

jvolmer (approved):

PR #17053, opened 2022-09-12
[PREGEL] use general messages for all communication between conductor and workers

jvolmer:
Use the general message for all communication between conductor and workers.

### DIFF
--- a/arangod/Pregel/Conductor/Conductor.cpp
+++ b/arangod/Pregel/Conductor/Conductor.cpp
@@ -81,6 +81,13 @@ using namespace arangodb;
 using namespace arangodb::pregel;
 using namespace arangodb::basics;
 
+namespace {
+template<class... Ts>
+struct overloaded : Ts... {
+  using Ts::operator()...;
+};
+}  // namespace
+
 #define LOG_PREGEL(logId, level)          \
   LOG_TOPIC(logId, level, Logger::PREGEL) \
       << fmt::format("[job {}]", _executionNumber)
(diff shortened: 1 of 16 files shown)


arangod/Pregel/Conductor/Conductor.cpp
```diff
@@ -231,17 +255,17 @@ auto Conductor::_preGlobalSuperStep() -> bool {
...
-void Conductor::workerStatusUpdate(VPackSlice const& data) {
+void Conductor::workerStatusUpdate(StatusUpdated const& data) {
   MUTEX_LOCKER(guard, _callbackMutex);
   // TODO: for these updates we do not care about uniqueness of responses
   // _ensureUniqueResponse(data);
 
-  auto event = deserialize<StatusUpdated>(data);
-
+  VPackBuilder event;
+  serialize(event, data);
```
romanatarango, line 264: This is only needed in debug mode.

arangod/Pregel/OutgoingCache.cpp
```diff
@@ -100,50 +102,51 @@ void ArrayOutCache<M>::flushMessages() {
...
+        {
+          VPackArrayBuilder ab2(&messagesVPack);
+          for (M const& message : messages) {
+            this->_format->addValue(messagesVPack, message);
+            if (this->_sendToNextGSS) {
+              this->_sendCountNextGSS++;
+            } else {
+              this->_sendCount++;
+            }
+          }
```
romanatarango, line 129: Can we use `serialize` for this?

romanatarango (approved):
LGTM, see comments about small things.

arangod/Pregel/OutgoingCache.cpp
jvolmer: Yes, I would like to, but it is a bit more complicated, therefore I would like to do it in a separate PR: 
Messages are of generic type M. Each algorithm defines the type M and uses a Format class of type M to do serialization and deserialization. Therefore in the PregelMessage struct we have to make sure that we always use the correct message type.
Additionally we have different types of things we send around: either an array of messages to one vertex (using ArrayOutCache and ArrayInCache) or one message to one vertex (using CombiningOutCache and CombiningInCache). This difference has to be taken into account inside the PregelMessage if we don't use a basic VPackBuilder.
We should definitely do that, but I think it is a bit more work and should be done in a separate PR.

arangod/Pregel/Conductor/Conductor.cpp
jvolmer: What is your suggestion here?

PR #17067, opened 2022-09-13
[PREGEL] Improve connection

jvolmer:
With this PR we have only one WorkerAPI (instead of SingleServerWorkerAPI and ClusterWorkerAPI before). The WorkerAPI has a pointer to a connection which can either be a direct connection (used in SingleServer mode) or a network connection (used in Cluster mode).
Additionally I improved OutgoingCache to use the network connection struct instead of doing rest calls by its own.


arangod/Pregel/Conductor/Conductor.cpp
```diff
@@ -349,18 +350,22 @@ auto Conductor::_initializeWorkers(VPackSlice additional) -> GraphLoadedFuture {
...
     for (auto const& [server, _] : vertexMap) {
       _dbServers.push_back(server);
-      workers.emplace(
-          server,
-          std::make_unique<conductor::ClusterWorkerApi>(
-              _executionNumber,
-              Connection::create(server, Utils::baseUrl(Utils::workerPrefix),
-                                 _vocbaseGuard.database())));
+      network::RequestOptions reqOpts;
+      reqOpts.timeout = network::Timeout(5.0 * 60.0);
```
romanatarango, line 361: For what is this a timeout? Why is it 5*60?

arangod/Pregel/Connection/Connection.h
```diff
@@ -0,0 +1,50 @@
...
+
+ private:
+  Type _type;
+  ServerID _id;
+  const char* typeString[2] = {"server", "shard"};
+
+ public:
+  Destination(Type type, ServerID id) : _type{type}, _id{std::move(id)} {}
+  auto toString() const -> std::string;
+};
```
romanatarango, line 43: The use of the class is unclear from the definition, please, consider write a short comment.

arangod/Pregel/Connection/Connection.h
```diff
@@ -0,0 +1,50 @@
...
+
+struct Connection {
```
romanatarango, line 45: Maybe "Sender"?

romanatarango (approved):
In general, LGTM, please read the comments.

arangod/Pregel/Conductor/Conductor.cpp
jvolmer: It was defined in Conductor::_sendToAllDBServers, which was used before to send the messages. I did not want to change the connection options, so I copied it here. I guess it means 5 minutes. We need it because we send a REST request with retry (see NetworkConnection.cpp line 79): We retry sending the request until the timeout is exceeded.

arangod/Pregel/Connection/Connection.h
jvolmer: The struct currently sends a message and receives an immediate response to it. Is this not a connection? I'm not so content with the name Sender, perhaps we come up with a better name in the process of improving the messaging.

PR #17073, opened 2022-09-13
[PREGEL] get rid of promises

jvolmer:
- Moves GraphStore into Worker directory
- Gets rid of promises: instead, uses futures::makeFutureWith which makes a real future out of a function.


arangod/Pregel/Worker/Worker.cpp
```diff
@@ -787,7 +787,9 @@ auto Worker<V, E, M>::loadGraph(LoadGraph const& graph)
 
   LOG_PREGEL("52070", WARN) << fmt::format(
       "Worker for execution number {} is loading", _config.executionNumber());
-  return std::move(_graphStore->loadShards(&_config, _makeStatusCallback()))
+  return futures::makeFutureWith([&]() {
```
MBkkt, line 790: Why not `futures::makeFuture(_graphStore->loadShards(&_config, _makeStatusCallback()))`?

MBkkt (approved):

MBkkt (changes requested):
Why do you create std::function and then call it instead of create future directly

arangod/Pregel/Worker/Worker.cpp
```diff
@@ -363,7 +367,8 @@ auto Worker<V, E, M>::_processVerticesInThreads() -> VerticesProcessedFuture {
     TRI_ASSERT(endI <= numSegments);
 
     auto vertices = _graphStore->vertexIterator(startI, endI);
-    processVertices.emplace_back(_processVertices(i, vertices));
+    processVertices.emplace_back(futures::makeFutureWith(
```
romanatarango, line 370: `processVertices` are rather `processedVertices`.

romanatarango (approved):
lgtm

arangod/Pregel/Worker/Worker.cpp
jvolmer: Thanks for the tipp, makeFuture is much better here :)

We decided that in Pregel we want to get rid of as many exceptions as possible and use ResultT instead. This makes it more obvious where errors could occur and force you to handle them directly.

PR #17077, opened 2022-09-13
[BTS-237] WWW-Authenticate header for HTTP/2

cpjulia:
### Scope & Purpose

Added WWW-Authenticate header to the HTTP response when status code is 401 for HTTP/2 as requested in the protocol. There was already coverage for HTTP/1.1.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests (already exist in tests/js/client/server_parameters, `www-authenticate.js` and `www-authenticate-off.js`
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports (probably not)

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-237
- [ ] Design document:


tests/js/client/server_parameters/www-authenticate.js
```diff
@@ -35,36 +32,83 @@ if (getOptions === true) {
...
+    setUp: function() {
+      users.save(user, "");
+      users.grantDatabase(user, '_system', 'rw');
+      helper.switchUser(user);
+    },
+
+    tearDown: function() {
+      helper.switchUser("root");
+      users.remove(user);
+      arango.reconnectWithNewPassword("");
```
jsteemann, line 61: Can't this be simplified to just
```suggestion
      connectWith("tcp", "root", "");
      users.remove(user);
```
?

tests/js/client/server_parameters/www-authenticate.js
```diff
@@ -35,36 +32,83 @@ if (getOptions === true) {
...
   return {
-    testUnauthorized : function() {
-      let result = request({ url: "/_api/version", method: "get" });
-      assertEqual(401, result.status);
-      assertTrue(result.headers.hasOwnProperty('www-authenticate'));
+
+    setUp: function() {
+      users.save(user, "");
+      users.grantDatabase(user, '_system', 'rw');
+      helper.switchUser(user);
```
jsteemann, line 55: I think it should be changed to 
```suggestion
      connectWith("tcp", "root", "");
      users.save(user, "");
      users.grantDatabase(user, '_system', 'rw');
```

tests/js/client/server_parameters/www-authenticate-off.js
```diff
@@ -36,12 +40,20 @@ if (getOptions === true) {
...
+};
+
 function testSuite() {
   return {
-    testHeader : function() {
-      let result = request({ url: "/_api/version", method: "get" });
-      assertEqual(200, result.status);
-      assertFalse(result.headers.hasOwnProperty('www-authenticate'));
+    testHeader: function() {
+      protocols.forEach((protocol) => {
```
jsteemann, line 51: nit: when the test starts, we are connected with HTTP/1, but when we are at the end of this test, we will be connected with HTTP/2. So this test has a side-effect that it permanently changes the connection type.
Ideally we ensure with setUp/tearDown that there are no side-effects after a test completes.

PR #17082, opened 2022-09-14
Testframework: Fix handling in crash analysis

dothebart:
### Scope & Purpose

- we would show errnous process flow when the arangod died in advance.
  This patch detects it, and avoids the situation alltogether. 
- properly print the instance name in information message
- aggregate the debugger processes when we shutdown-deadline-kill an arangod
- if the debugger is unable to terminate the process, we must not hand, but hard kill the process on our own.

- [x] :hankey: Bugfix
- [x]  backport 3.10: https://github.com/arangodb/arangodb/pull/17087


mchacki (approved):
LGTM.

Now we get better information if instances crash during tests

PR #17066, opened 2022-09-13
[CINFRA-483] MoveShard for Replication 2

maierlars:
### Scope & Purpose

This PR adds a special code path for replication 2 to move shard. By doing this, the old move shard API works transparently with replication 2.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #17083, opened 2022-09-14
[CINFRA-524] Add persisted state info and cleanup handler

maierlars:
### Scope & Purpose

Added persisted state info (only partially used) and a clean up handler. Used in the document store to drop the collection.


apetenchea (approved):
The code looks good to me!
We have a race when trying to drop the collection, which most of the time I can reproduce locally. Couldn't tell exactly where that is coming from, but I added some logging and it turns out that, when the collection drop blocks, the maintenance keeps seeing the shards, even after the replicated states have been dropped. I don't see how they could be somehow recreated quickly, so maybe the cleanup doesn't always do the job?

maierlars:
Yes, I saw that and I already fixed it in a follow up branch. The problem was: if the replicated log resigned faster than the replicated state, the replicated state would drop the core without destroying it. I fixed that in #17066 .

PR #17085, opened 2022-09-14
Add embedFields support for inspector

mpoeter:
### Scope & Purpose

Add support for embedding fields for inner structs with a separate inspect function. E.g.:
```
struct Inner {
   int i;
  std:: string s;
};

template<class Inspector>
auto inspect(Inspector& f, Inner & x) {
  return f.object(x).fields(f.field("i", x.i), f.field("s", x.s));
}

struct Embedded {
  int a;
  Inner inner;
  int b;
};

template<class Inspector>
auto inspect(Inspector& f, Embedded& v) {
  return f.object(v).fields(f.field("a", v.a), f.embedFields(v.inner),
                            f.field("b", v.b));
}
```
will handle objects that look like this  `{ a: ..., i: ..., s: ..., b: ...}`

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**

### DIFF
--- a/lib/Inspection/InspectorBase.h
+++ b/lib/Inspection/InspectorBase.h
@@ -320,7 +320,7 @@ namespace detail {
 template<class Inspector>
 struct Fields {
   virtual ~Fields() = default;
-  virtual Status apply(Inspector&) = 0;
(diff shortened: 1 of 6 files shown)


maierlars (approved):
LGTM!

PR #17096, opened 2022-09-15
[PREGEL] Use future for cleanup

jvolmer:
This PR uses futures for the worker cleanup (cleanup is the deletion of workers. it is happening either after storing or - if the workers still exist - after canceling).
The PR also separates the cleanup request to a worker into two requests: store and cleanup. This makes it more clear what is happening: When storing, first the store request is send to the workers, then the cleanup request. When canceled, only the cleanup request is send.
I also fixed a deadlock when shutting down.


arangod/Pregel/Conductor/States/CanceledState.cpp
```diff
@@ -17,58 +18,63 @@ Canceled::Canceled(Conductor& conductor, std::chrono::seconds const& ttl)
...
-        });
-  }
+  _cleanupUntilTimeout(std::chrono::steady_clock::now())
+      .thenValue([&](auto result) {
+        if (result.fail()) {
+          LOG_PREGEL_CONDUCTOR("f8b3c", ERR) << result.errorMessage();
+          return;
+        }
+
+        if (conductor._inErrorAbort) {
```
romanatarango, line 72: No message here?

arangod/Pregel/PregelFeature.cpp
```diff
@@ -480,15 +481,22 @@ void PregelFeature::start() {
...
+      conductors.emplace_back(it.second.conductor);
+    }
+    for (auto it : _workers) {
+      it.second.second->cancelGlobalStep(VPackSlice());
+    }
   }
-  for (auto it : _workers) {
-    it.second.second->cancelGlobalStep(VPackSlice());
+
+  // delete all conductors and workers
```
romanatarango, line 496: why also workers?

romanatarango (approved):
lgtm

PR #17099, opened 2022-09-15
Fixed issue BTS-1023

jsteemann:
### Scope & Purpose

Fixed issue https://arangodb.atlassian.net/browse/BTS-1023:

  Added Linux-specific startup option `--use-splice-syscall` to control
  whether the Linux-specific splice() syscall should be used for copying
  file contents. While the syscall is generally available since Linux 2.6.x,
  it is also required that the underlying filesystem supports the splice
  operation. This is not true for some encrypted filesystems, on which
  splice() calls thus fail.
  By setting the startup option `--use-splice-syscall` to `false`, a less
  efficient, but more portable user-space file copying method will be
  used instead, which should work on all filesystems.
  The startup option is not available on other operating systems than Linux.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17098
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17109
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1023
- [ ] Design document:


lib/ApplicationFeatures/FileSystemFeature.cpp
```diff
@@ -0,0 +1,56 @@
...
+void FileSystemFeature::collectOptions(
+    std::shared_ptr<ProgramOptions> options) {
+#ifdef __linux__
+  // option is only available on Linux
+  options
+      ->addOption("--use-splice-syscall",
+                  "Use the splice() syscall for file copying (may not be "
+                  "supported on all filesystems",
+                  new BooleanParameter(&_useSplice),
+                  options::makeFlags(Flags::DefaultNoOs, Flags::OsLinux))
```
Simran-B, line 45: Heretical question: What is the OS flag good for if a pre-processor directive removes the entire code path?
This is mainly a problem for options that Windows and/or macOS only.

jsteemann: It is probably still good for tools that, if run on Linux, will otherwise not know that the flag is Linux-specific.

Simran-B: If a not-implemented fatal error would be thrown on unsupported OSes, wouldn't that be like the status quo where the server refuses to start if an unknown startup option is passed? (I'm not sure how difficult or cumbersome that would be to implement, but it seems like a sensible middleground)

jsteemann: Why would anyone set the option on a non-Linux OS?
It is a _new_ option, so no existing deployment will use it yet, and users need to opt in.
And the documentation clearly states that it is Linux-only.

MBkkt (approved):

cpjulia (approved):
LGTM

PR #17102, opened 2022-09-15
Disable frequently failing soft shutdown pregel test

markuspf:
### Scope & Purpose

Disable the frequently hanging `test-soft-shutdown-pregel-cluster` test while debugging is in progress.


jsteemann (approved):
LGTM!

PR #17105, opened 2022-09-15
[PREGEL] Use future to collect results

jvolmer:
This PR uses futures to collect pregel results.
Collect pregel results is called either from Functions via AQL or from v8-collection via JS. I simplified the calls in both (Functions and v8-collections) by adding a function to the pregel feature that gets the results if available.


romanatarango (approved):
lgtm

PR #17110, opened 2022-09-15
Disable frequently failing pregel test... correctly

markuspf:
### Scope & Purpose

We try doing what we wanted in #17102  just actually disable the test.


jsteemann (approved):
LGTM!

PR #17123, opened 2022-09-17
Update fmt to version 9.1.0

markuspf:
### Scope & Purpose

Update fmt to 9.1.0; 

This requires a small fix in the LogicalCollection code: automatic use of `ostream<<` for `fmt` has been deprecated because of problems with ODR violations.

Two options exist: either explicit formatter specialisation with a helper struct `ostream_formatter` or using `fmt::streamed`; the latter seemed less intrusive for now.


jsteemann (approved):
LGTM

markuspf:
Thanks!

PR #17124, opened 2022-09-17
Inspector formatting (pedestrian #[derive(Debug)])

markuspf:
### Scope & Purpose

* Adds `auto serialize(T& v) -> SharedSlice` for inspectable `T`.
* Adds `inspection_formatter` that allows explicit specialisation of `fmt::formatter` for types that are inspectable. This is inspired by the `ostream_formatter` from `fmt 9.1.0`.
* Requires `fmt-9.1.0` which is added in #17123 

The helper formats using the added `serialize` function and either `toJson` (not pretty) or `toString` (prettier) using format strings `{:u}` and `{:p}`.
```
struct Foo { uint32_t v; };

template<typename Inspector>
auto inspect(Inspector f, Foo& v) {
  return f.object(v).fields(f.field("v", v.v));
}

template<>
struct fmt::formatter<Foo> : arangodb::inspection::inspection_formatter {};

fmt::print("A Foo: {}", Foo{.v = 15});
```

It is worth keeping in mind that every serialisation will create a VPackBuilder and serialise the whole object into it before printing, which might be quite inefficient. Future iterations could use a special FormattingInspector.

### DIFF
--- a/lib/Inspection/VPack.h
+++ b/lib/Inspection/VPack.h
@@ -43,7 +43,7 @@ void serialize(Builder& builder, T& value) {
 }
 
 template<class T>
-auto serialize(T& value) -> std::shared_ptr<Builder> {
+[[nodiscard]] auto serialize(T& value) -> std::shared_ptr<Builder> {
   auto builder = std::make_shared<Builder>();
   serialize(*builder, value);
   return builder;
--- a/lib/Inspection/VPack.h
+++ b/lib/Inspection/VPack.h
@@ -43,10 +43,10 @@ void serialize(Builder& builder, T& value) {
 }
 
 template<class T>
-[[nodiscard]] auto serialize(T& value) -> std::shared_ptr<Builder> {
-  auto builder = std::make_shared<Builder>();
-  serialize(*builder, value);
-  return builder;
+[[nodiscard]] auto serialize(T& value) -> SharedSlice {
+  auto builder = Builder();
+  serialize(builder, value);
+  return builder.sharedSlice();
 }
 
 namespace detail {
--- a/lib/Inspection/VPack.h
+++ b/lib/Inspection/VPack.h
@@ -46,7 +46,7 @@ template<class T>
 [[nodiscard]] auto serialize(T& value) -> SharedSlice {
   auto builder = Builder();
   serialize(builder, value);
-  return builder.sharedSlice();
+  return std::move(builder).sharedSlice();
 }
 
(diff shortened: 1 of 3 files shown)


lib/Inspection/VPack.h
```diff
@@ -42,6 +42,13 @@ void serialize(Builder& builder, T& value) {
   }
 }
 
+template<class T>
+[[nodiscard]] auto serialize(T& value) -> std::shared_ptr<Builder> {
+  auto builder = std::make_shared<Builder>();
+  serialize(*builder, value);
+  return builder;
+}
```
mpoeter, line 50: Any particular reason why we return a shared_ptr instead of a Builder?

mpoeter (approved):

lib/Inspection/VPack.h
maierlars: Builder are pretty expensive object to move around. However, I would suggest to use something else than a builder entirely. Something that still owns the slice.

markuspf: Returning a builder was the obvious thing that I knew to return; as Lars says, builders are fairly heavy, so a pointer seemed like not the worst idea.

Then I thought about having a structure that owns a slice (or a builder), would https://github.com/arangodb/velocypack/blob/main/include/velocypack/SliceContainer.h be the correct thing to use?

markuspf: Looking at this again
 * the formatter overload should be for a velocypack slice,
 * serialize could return just a SliceContainer because we don't really want to be able to fiddle about with that slice again, anyway.

lib/Inspection/VPack.h
```diff
@@ -42,6 +42,13 @@ void serialize(Builder& builder, T& value) {
   }
 }
 
+template<class T>
+[[nodiscard]] auto serialize(T& value) -> SharedSlice {
+  auto builder = Builder();
+  serialize(builder, value);
+  return builder.sharedSlice();
```
mpoeter, line 49: This will copy the buffer into a new SharedSlice. To prevent the copy you could do:
```suggestion
  return std::move(builder).sharedSlice();
```
But since I could only find one usage of this serialize function in `inspection_formatter::format` I am wondering why we don't just create a local Builder instance there and use the existing serialize function that takes the builder

markuspf: Because I imagine using this `serialize` function in other places and not just in the formatter ;)

PR #17126, opened 2022-09-18
[PREGEL] Improved explanation of Pregel SCC and tidification

romanatarango:
Improved explanation of Pregel's SCC algorithm, trivial changes according to clang-tidy.

### DIFF
--- a/arangod/Pregel/Graph.h
+++ b/arangod/Pregel/Graph.h
@@ -30,9 +30,9 @@
 #include <cstdint>
 #include <functional>
 #include <string>
+#include <utility>
 
-namespace arangodb {
-namespace pregel {
+namespace arangodb::pregel {
 
 typedef uint16_t PregelShard;
 const PregelShard InvalidPregelShard = -1;
@@ -41,8 +41,8 @@ struct PregelID {
   std::string key;    // std::string 24
   PregelShard shard;  // uint16_t
 
-  PregelID() : key(""), shard(InvalidPregelShard) {}
-  PregelID(PregelShard s, std::string const& k) : key(k), shard(s) {}
+  PregelID() : shard(InvalidPregelShard) {}
+  PregelID(PregelShard s, std::string  k) : key(std::move(k)), shard(s) {}
 
   bool operator==(const PregelID& rhs) const {
     return shard == rhs.shard && key == rhs.key;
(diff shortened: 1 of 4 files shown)


arangod/Pregel/Graph.h
```diff
@@ -41,8 +41,8 @@ struct PregelID {
   std::string key;    // std::string 24
   PregelShard shard;  // uint16_t
 
-  PregelID() : key(""), shard(InvalidPregelShard) {}
-  PregelID(PregelShard s, std::string const& k) : key(k), shard(s) {}
+  PregelID() : shard(InvalidPregelShard) {}
+  PregelID(PregelShard s, std::string k) : key(std::move(k)), shard(s) {}
```
markuspf, line 45: This looks dubious to me, did clang-tidy suggest that?

romanatarango: Yes. `key("")` because a string is initialized to the empty string by default and `std::move(k)` because the string is anyway copied into the parameter, now it can be reused.

arangod/Pregel/Graph.h
```diff
@@ -184,19 +188,20 @@ class Vertex {
...
+  [[nodiscard]] std::string_view key() const {
+    return std::string_view(_key, keyLength());
+  }
   V const& data() const& { return _data; }
   V& data() & { return _data; }
 
-  PregelID pregelId() const {
-    return PregelID(_shard, std::string(_key, keyLength()));
+  [[nodiscard]] PregelID pregelId() const {
+    return {_shard, std::string(_key, keyLength())};
```
markuspf, line 200: I found the original more clear. If it must be a brace-initialisation (why?), then
```suggestion
    return PregelID{_shard, std::string(_key, keyLength())};
```

markuspf (approved):
LGTM, needs a Jenkins.

markuspf:
Random thought: maybe update the description of the algorithm with a worst-case runtime estimate?

romanatarango:
> Random thought: maybe update the description of the algorithm with a worst-case runtime estimate?

Did it. The worst case is O(n * m) or O((n+m)^2), the average case is somewhat subtle. I could get only O(m^2 / log(m)) as an upper bound both for the running time and number of gss. Also added possible improvements.

PR #17152, opened 2022-09-21
Allow children to have children

markuspf:
ArangoDB is very trigger happy with blocking signals on startup by calling `maskAllSignals`; this propagates to children, and thus prevented me from running `ctest` from the `unittests` script.

This is an attempt at a fix: just `unmaskAllSignals` in the child after a fork.


jsteemann (approved):
Looks reasonable, and doesn't affect arangod executables.

PR #17157, opened 2022-09-21
[CINFRA-485] Intermediate Commits

apetenchea:
### Scope & Purpose

This PR adds support for intermediate commits in replication2 databases.
Changes:
- `checkIntermediateCommit` used to perform both the check and trigger an intermediate commit if one was needed. This PR splits it up into two functions, `isIntermediateCommitNeeded` and `triggerIntermediateCommit`. On the follower, `triggerIntermediateCommit` is called without a check and a read snapshot is not needed.
- `performIntermediateCommitIfRequired` now returns a `Future`, because the leader of a replicated state has to wait for the `kIntermediateCommit` entry to be committed before continuing.
- The truncate operation still blocks on `performIntermediateCommitIfRequired`
- Integration and unit tests have been updated. Also, some of the tests in `common/shell/shell-transaction.js` provide coverage for intermediate commits. We used them after changing locally the default replication type to replication2.
- We're making better use of the Inspector for `DocumentLogEntry`.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports


arangod/ClusterEngine/ClusterTransactionState.cpp
```diff
@@ -171,8 +171,13 @@ Result ClusterTransactionState::abortTransaction(
   return {};
 }
 
-Result ClusterTransactionState::performIntermediateCommitIfRequired(
-    DataSourceId cid) {
+arangodb::Result ClusterTransactionState::triggerIntermediateCommit() {
+  TRI_ASSERT(false);  // should never be called!
+  return {};
```
mpoeter, line 176: Should we just throw here (like we do in `performIntermediateCommitIfRequired`)?

arangod/ClusterEngine/ClusterTransactionState.h
```diff
@@ -52,7 +52,9 @@ class ClusterTransactionState final : public TransactionState {
   /// @brief abort a transaction
   [[nodiscard]] Result abortTransaction(transaction::Methods* trx) override;
 
-  [[nodiscard]] Result performIntermediateCommitIfRequired(
+  virtual arangodb::Result triggerIntermediateCommit() override;
```
mpoeter, line 55:
```suggestion
  Result triggerIntermediateCommit() override;
```

arangod/Transaction/Methods.cpp
```diff
@@ -3007,7 +3023,7 @@ Future<Result> Methods::replicateOperations(
             operation),
         state()->id(),
         replication2::replicated_state::document::ReplicationOptions{});
-    return Result{};
+    return performIntermediateCommitIfRequired(collection->id());
```
mpoeter, line 3026: Why? This is happening immediately after `replicateOperation`, which itself should already call `performIntermediateCommitIfRequired`. I think we should instead just return the result of the `replicateOperation` call.

mpoeter (changes requested):
I know that there have been some conflicting changes in devel recently. We should merge latest devel into staging and then update this branch with staging to fix the conflicts as part of this PR.

arangod/Transaction/Methods.cpp
mpoeter: Sorry, I was confused by the diff! All good!

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -442,7 +442,12 @@ arangodb::Result fetchRevisions(
       futures.pop_front();
       shoppingLists.pop_front();
       TRI_ASSERT(futures.size() == shoppingLists.size());
-      res = trx.state()->performIntermediateCommitIfRequired(collection.id());
+
+      auto fut =
+          trx.state()->performIntermediateCommitIfRequired(collection.id());
+      TRI_ASSERT(fut.isReady());
```
mchacki, line 448: Is this assertion correct?
This would indicate that there is not actually a "Future" here but the above method is synchronous.
Also i do not think this assert is required, with the below `get()` we will wait for the readiness, so not Production issue, may just be triggered within our tests and surprise someone.

If this is correct, please add a comment here why this future is always ready.

arangod/Replication/DatabaseInitialSyncer.cpp
```diff
@@ -2047,7 +2052,9 @@ Result DatabaseInitialSyncer::fetchCollectionSyncByRevisions(
       }
       toFetch.clear();
 
-      res = trx->state()->performIntermediateCommitIfRequired(coll->id());
+      auto fut = trx->state()->performIntermediateCommitIfRequired(coll->id());
+      TRI_ASSERT(fut.isReady());
```
mchacki, line 2056: See above comment

arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
```diff
@@ -231,6 +234,20 @@ void ReplicatedRocksDBTransactionState::addIntermediateCommits(uint64_t value) {
                                  "invalid call to addIntermediateCommits");
 }
 
+arangodb::Result
+ReplicatedRocksDBTransactionState::triggerIntermediateCommit() {
+  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL,
```
mchacki, line 239: Why is this a THROW and not a return of Result?

return would be expected, throw not

arangod/RocksDBEngine/RocksDBTransactionMethods.h
```diff
@@ -56,7 +56,10 @@ class RocksDBTransactionMethods : public RocksDBMethods {
   virtual Result abortTransaction() = 0;
 
   // Only relevant for RocksDBTrxMethods
-  virtual Result checkIntermediateCommit() { return {}; }
+  virtual bool isIntermediateCommitNeeded() { return false; }
+  virtual Result triggerIntermediateCommit() {
+    THROW_ARANGO_EXCEPTION(TRI_ERROR_NOT_IMPLEMENTED);
```
mchacki, line 61: Why is this a THROW and not a return Result?

I understand it should never be called, however return is handled gracefully, throw potentially not.

tests/Mocks/StorageEngineMock.cpp
```diff
@@ -2115,7 +2116,13 @@ TransactionStateMock::commitTransaction(arangodb::transaction::Methods* trx) {
   return arangodb::Result();
 }
 
-arangodb::Result TransactionStateMock::performIntermediateCommitIfRequired(
+arangodb::Result TransactionStateMock::triggerIntermediateCommit() {
+  THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL,
```
mchacki, line 2120: Why throw? would prefer return Result here again

mchacki (changes requested):
Some really minor changes, in general this looks very good and clean. Thank you for this!

arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
```diff
@@ -44,21 +44,28 @@ enum class OperationType {
...
-  kAbortAllOngoingTrx
+  kAbortAllOngoingTrx,
+  kIntermediateCommit,
 };
 
-auto to_string(OperationType) noexcept -> std::string_view;
+template<class Inspector>
+auto inspect(Inspector& f, OperationType& p) {
+  return f.enumeration(p).values(
+      OperationType::kInsert, "Insert",                          //
```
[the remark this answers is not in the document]
mpoeter: On the follower side we handle everything by doing a replace. However, I would like to keep those in case we want to change that in the future (provided we can figure out how we can do this by replicating only the changes and not the complete document).

arangod/Replication/DatabaseInitialSyncer.cpp
mpoeter: Yes, this is correct, because this is an intermediate commit on the local rocksdb transaction. `performIntermediateCommitIfRequired` now returns a future, because _iff_ the transaction in question is operates on a replication 2.0 database and is the leader, then the intermediate commit needs to be replicated, which is an asynchronous operation, before we perform the local commit. In all other cases, we synchronously perform the local commit only. Since this is replication1 specific code, everything is synchronous.

arangod/Replication/DatabaseInitialSyncer.cpp
mpoeter: See explanation above.

arangod/RocksDBEngine/RocksDBTransactionMethods.h
mpoeter: If this is ever called, then this is a serious bug. Not sure if this can actually be handled gracefully in a _correct_ way. So maybe we should even use a `ADB_PROD_ASSERT(false)` (which would actually terminate the process).

arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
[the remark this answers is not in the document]
mchacki: Okay :+1:

apetenchea:
Using ADB_PROD_ASSERT instead of throwing.
Previous Jenkins run is blue: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/24750/
Compiled locally. No need for another Jenkins run, as the last commit only touches dead code.

PR #17159, opened 2022-09-21
Add embedded variant format

mpoeter:
### Scope & Purpose

New variant encoding that puts the `type` indicator field on the same level as the various objects.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**


maierlars (approved):

mpoeter:
Blue except for unrelated error

PR #17163, opened 2022-09-22
Activate and fix LockStressTest.cpp

markuspf:
### Scope & Purpose

`Cache/LockStressTest.cpp` wasn't listed in CMakeLists.txt, and hence neither compiled nor run.

It was also broken.


jsteemann:
@markuspf : if the test works, let's activate it.
60 seconds test runtime does not seem to be a real burden for the CI.

MBkkt (approved):

jsteemann (approved):
LGTM!

PR #17169, opened 2022-09-22
[PREGEL] Combine worker apis

jvolmer:
This PR combines the worker apis into one api:
When we call the workers from the conductor, we send the same message to all workers (with one exception). This PR combines the worker apis into one api such that the conductor (or mostly the conductor states) don't have to manage sending the messages to the correct workers, but leave that to the worker api.
There is one exception: When loading the graph, we send different messages to each worker: Each worker gets a list of shards on its DBServer of the vertex and edge collections involved. The solution used here is not robust, it is fixed in the next PR: https://github.com/arangodb/arangodb/pull/17175.

I also made the run functions of the condcutor states better readable.

And I deleted the unused message interface inside WorkerConductorMessages.h.

### DIFF
--- a/arangod/Pregel/Conductor/WorkerApi.cpp
+++ b/arangod/Pregel/Conductor/WorkerApi.cpp
@@ -6,15 +6,21 @@
 
 using namespace arangodb::pregel::conductor;
 
-// This template enforces In and Out type of the function:
-// 'In' enforces which type of message is sent
-// 'Out' defines the expected response type:
-// The function returns an error if this expectation is not fulfilled
 template<typename Out, typename In>
-auto WorkerApi::execute(In const& in) const -> futures::Future<ResultT<Out>> {
+auto WorkerApi::sendToAll(In const& in) const -> FutureOfWorkerResults<Out> {
+  auto results = std::vector<futures::Future<ResultT<Out>>>{};
+  for (auto&& server : _servers) {
+    results.emplace_back(send<Out>(server, in));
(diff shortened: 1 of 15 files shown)


arangod/Pregel/Conductor/WorkerApi.cpp
```diff
@@ -34,30 +40,29 @@ auto WorkerApi::execute(In const& in) const -> futures::Future<ResultT<Out>> {
 
 auto WorkerApi::loadGraph(LoadGraph const& graph)
     -> futures::Future<ResultT<GraphLoaded>> {
-  return execute<GraphLoaded>(graph);
+  return send<GraphLoaded>(*_loadGraphIterator++, graph);
```
jvolmer, line 43: As mentioned in the PR description, loading the graph is an exception: Here we send different messages to each worker.
I used the iterator here to keep the function signature the same, but it is not optimal: We have to be sure that the given LoadGraph belongs to the ServerID of the iterator.
I had the idea we change the LoadGraph message such that it does not include any DBServer specifics but let the DBServer do the work (find the relevant shards) but this is not save: Due to replication we cannot be sure that we read a shard exactly once - different DBServers could in some use cases assume that they are both the leader for a specific shard and therefore both read the same shard or there is a shard where no DBServer assumes to be the leader of and so it is not read at all.
I'm happy to head suggestions how to handle this case.

markuspf (approved):

PR #17171, opened 2022-09-22
[CINFRA-556] Fix drop index operation with replication 2.0

mpoeter:
### Scope & Purpose

- [x] :hankey: Bugfix


maierlars (approved):

maierlars:
Isn't that something that could have been used to fix the EnsureIndex problem as well?

mpoeter:
@maierlars yes, initially I wanted to avoid changing the options, but since we now need it anyway I also adapted the EnsureIndex code to use this flag as well.

PR #17189, opened 2022-09-23
one more place missing bulk insert

dothebart:
### Scope & Purpose

one more test that was missing bulk inserts

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM!

PR #17201, opened 2022-09-26
[BTS-1043] Fix index executor

markuspf:
### Scope & Purpose

This fixes the `IndexExecutor` so that it skips as many rows as it can.

Before this patch the IndexExecutor only skipped once, leading to unpredictable skips and assertions being triggered in maintainer mode.

This bug should not have any impact on correctness of returned results, just on performance.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**

### DIFF
--- a/tests/js/server/aql/aql-index-skip-regression.js
+++ b/tests/js/server/aql/aql-index-skip-regression.js
@@ -1,0 +1,109 @@
+/*jshint globalstrict:false, strict:false, maxlen: 500 */
+/*global assertEqual, AQL_EXECUTE, assertTrue, fail */
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief tests for regression returning blocks to the manager
+///
+/// @file
+///
+/// DISCLAIMER
+///
+/// Copyright 2010-2014 triagens GmbH, Cologne, Germany
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is triAGENS GmbH, Cologne, Germany
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+var jsunity = require("jsunity");
+var internal = require("internal");
+var errors = internal.errors;
+var db = require("@arangodb").db,
+  indexId;
+
+// This example was produced by Jan Steeman to reproduce a
+// crash in the TraversalExecutor code
+const movieCollectionName = "MOVIE";
+
+var cleanup = function() {
+  db._drop(movieCollectionName);
+};
+
+var createData = function() {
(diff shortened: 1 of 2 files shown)


tests/js/server/aql/aql-index-skip-regression.js
```diff
@@ -0,0 +1,109 @@
...
+/// Copyright holder is triAGENS GmbH, Cologne, Germany
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+var jsunity = require("jsunity");
+var internal = require("internal");
+var errors = internal.errors;
+var db = require("@arangodb").db,
+  indexId;
```
jsteemann, line 34:
```suggestion
const jsunity = require("jsunity");
const db = require("@arangodb").db;
```

tests/js/server/aql/aql-index-skip-regression.js
```diff
@@ -0,0 +1,109 @@
...
+  var data = [];
+
+  for(let i = 1; i <= 321; i++) {
+    data.push({ test_node: null});
+  }
+  for(let i = 1; i <= 1679; i++) {
+    data.push({ test_node: i});
+  }
+  db._collection(movieCollectionName).save(data);
+};
```
jsteemann, line 56:
```suggestion
const cleanup = function() {
  db._drop(movieCollectionName);
};

const createData = function() {
  let c = db._create(movieCollectionName);

  let data = [];
  for (let i = 1; i <= 321; i++) {
    data.push({ test_node: null});
  }
  for (let i = 1; i <= 1679; i++) {
    data.push({ test_node: i});
  }
  c.insert(data);
};
```

tests/js/server/aql/aql-index-skip-regression.js
```diff
@@ -0,0 +1,109 @@
...
+
+      const bindVars = {
+        "@value0": movieCollectionName,
+        "value1": null
+      };
+
+      var actual = db._query(query, bindVars);
+      var foo = actual.toArray();
+      assertEqual(foo.length, 0, "Expecting query result to contain 0 documents");
+    }
```
jsteemann, line 99:
```suggestion

    testIndexSkipCrashes: function() {
      const query = `FOR doc IN ${cn}
                         FILTER doc['test_node'] != null
			 SORT doc._key ASC
			 LIMIT 1680, 10
                         RETURN doc`;

      let actual = db._query(query).toArray();
      assertEqual(actual.length, 0, "Expecting query result to contain 0 documents");
    }
```

tests/js/server/aql/aql-index-skip-regression.js
```diff
@@ -0,0 +1,109 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+var jsunity = require("jsunity");
+var internal = require("internal");
+var errors = internal.errors;
+var db = require("@arangodb").db,
+  indexId;
+
+// This example was produced by Jan Steeman to reproduce a
+// crash in the TraversalExecutor code
```
jsteemann, line 37: Ok, not that I am aware of :sweat_smile:

tests/js/server/aql/aql-index-skip-regression.js
markuspf: I left the bind parameters in because that is the example that was reported to me and should be on the BTS ticket (once that appears)

jsteemann (approved):
LGTM

goedderz (approved):

mchacki (approved):
LGTM!

PR #17206, opened 2022-09-27
Drop replicated states during database shutdown

goedderz:
### Scope & Purpose

Drop replicated states during database shutdown. Without this, a replicated state could be kept alive after the database is gone, possibly resulting in unexpected errors.

- [X] :hankey: Bugfix


maierlars (approved):

apetenchea (approved):

maierlars (changes requested):
WARNING: `drop` invokes the cleanup handler and actually removes the collection! What we need is a resign (and stay resigned) but we don't have that yet.

PR #17214, opened 2022-09-27
Removing intentional sleep from tests

apetenchea:
### Scope & Purpose

Previously, we had to wait a second after each replication2 test, making sure that the collection is not deleted while the follower is still processing ongoing transactions. This wait is no longer needed, because the replicated state is now responsible for removing its own collection.


mchacki (approved):
LGTM

PR #17222, opened 2022-09-27
Docs: ArangoSearch v2 corrections

Simran-B:
- Clarify inheritance of analyzer and features in inverted indexes
- `search-alias` Views accept index IDs (without collection prefix)

### DIFF
--- a/Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
+++ b/Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
@@ -19,27 +19,28 @@ If omitted, a name is auto-generated so that it is unique with respect to the
 collection, e.g. `idx_832910498`.
 
 @RESTBODYPARAM{fields,array,required,post_api_index_inverted_fields}
-An array of attribute paths as strings to index the fields with the default
-options, or objects to specify options for the fields.
+An array of attribute paths. You can use strings to index the fields with the
+default options, or objects to specify options for the fields (with the
+attribute path in the `name` property), or a mix of both.
 
 @RESTSTRUCT{name,post_api_index_inverted_fields,string,required,}
 An attribute path. The `.` character denotes sub-attributes.
+You can expand one array with `[*]`.
 
 @RESTSTRUCT{analyzer,post_api_index_inverted_fields,string,optional,}
 The name of an Analyzer to use for this field.
-Default: the value defined by the top-level `analyzer` option, or if not set,
-the default `identity` Analyzer.
+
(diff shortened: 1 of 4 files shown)


Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
```diff
@@ -95,18 +95,17 @@ An attribute path. The `.` character denotes sub-attributes.
...
+A list of Analyzer features to use for this field. You can set this option to
+overwrite what features are enabled for the `analyzer`. Possible features:
 - `"frequency"`
 - `"norm"`
 - `"position"`
 - `"offset"`
 
-Default: the value of the top-level `features` option, or if not set, the
-features defined by the Analyzer itself.
+Default: the features as defined by the Analyzer itself.
```
Dronplane, line 108:
```suggestion
Default: the features as defined by the Analyzer itself, or by parent features field in case of analyzer is not set for this field
```

Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
```diff
@@ -176,15 +181,17 @@ Defines how to compress the primary sort data. Possible values:
 The name of an Analyzer to use by default. This Analyzer is applied to the
 values of the indexed fields for which you don't define Analyzers explicitly.
 
+Default: `identity`
+
 @RESTBODYPARAM{features,array,optional,string}
-A list of Analyzer features to use by default. They define what features are
-enabled for the default `analyzer`. Possible features:
+A list of Analyzer features. You can set this option to overwrite what features
+are enabled for the default `analyzer`. Possible features:
```
Dronplane, line 188: The features have compatibility rules, like position requires frequency. Should we note this?

Simran-B: I suppose you refer to this list of dependent features? https://www.arangodb.com/docs/devel/analyzers.html#analyzer-features
I think it's sufficiently covered on that page. I don't want to add a link because we would need to use an absolute URL because of Swagger UI 😞

Dronplane: I see. Ok, then.

Documentation/DocuBlocks/Rest/Indexes/post_api_index_inverted.md
Simran-B: features are inherited from parent/top-level unless analyzer is set further down

Dronplane (approved):
Lgtm

PR #17232, opened 2022-09-28
[CMAKE] Split common test code

markuspf:
### Scope & Purpose

A first step on a longer journey; unfortunately already a good bit bigger than I would like it to be.

At the same time 
 * a cleanup of the `CMakeLists.txt` file(s) in the test directories, in particular a slight bit of cleanup of the dependencies that have developed over time between the separate sections of testing code.

 * a step towards separated test binaries.

This PR puts shared tests code into three libraries:
 1) `arangodbtests_main` contains just `main.cpp` and the main function
 2) `arangodbtests_lib` kitchen sink for utility functions. Hopefully more a temporary resting place, or a very small collection of useful functions that are only used for tests
 3) `arangodbtests_mocks` where the Mocks live that are currently shared between test libraries.

 
 Note that there is an enterprise PR, too: https://github.com/arangodb/enterprise/pull/1102


jvolmer (commented):
I like that you split up the test executable, some thoughts:
- I am not sure if the new libraries you created are self-contained, see my comment.
- I can see that beforehand several mocks were located in the IResearch where they were used but they did not really belong there (e.g. AgencyMock is not directly connected to IResearch), so putting them all in one Mock directory makes them already easier reusable. But I also don't like that we have now one Mock directory that contains all mocks: Mocks mock a specific part of the application - often a specific class of a module - therefore Mocks are easiest to find and to reuse if they are located in the module folder they belong to. If you then delete a module, the mocks in it are directly deleted with it. E.g. the AgencyMock belongs into the Agency directory. I would put as many mocks into their respective module directories as possible.
- This is really minor: I don't like the directory name TestLib because the containing directory is already named test: I would prefer Lib or Helpers.

markuspf:
>     * I am not sure if the new libraries you created are self-contained, see my comment.

They are almost certainly not, and with the ArangoDB dependency tangle I am not sure they reasonably can be; I did the split along the lines of "this looks reasonable to me, can I make *these* self contained later?" rather than "these are self-contained, hence I split them here".
 
>     * I can see that beforehand several mocks were located in the IResearch where they were used but they did not really belong there (e.g. AgencyMock is not directly connected to IResearch), so putting them all in one Mock directory makes them already easier reusable. But I also don't like that we have now one Mock directory that contains all mocks: Mocks mock a specific part of the application - often a specific class of a module - therefore Mocks are easiest to find and to reuse if they are located in the module folder they belong to. If you then delete a module, the mocks in it are directly deleted with it. E.g. the AgencyMock belongs into the Agency directory. I would put as many mocks into their respective module directories as possible.

I disagree for two reasons:
 * This is not possible (right now, maybe ever): certain mocks are used across tests in different subdirectories.
 * Why? Because mocks are not tests, mocks are mocks. My thought was to rather have the Mocks be a part of `TestLib`;
 * (Thinking of it, we should NEVER EVER have to link arangoserver to the tests....)

Maybe this situation becomes better once we have better unit testabilit? 

Yet, if we are to write more integrate-y C++ tests we might always need some mocks. Really, the tests need to be better separated into unit-y tests and integrate-y ones.


>     * This is really minor: I don't like the directory name TestLib because the containing directory is already named test: I would prefer Lib or Helpers.

This is intentional; I really dislike `Helpers`, because it says (and smells of) "Kitchen sink"; we also have *multiple* places where we call subdirectories `Basics` or `Helper`; we then include `Basics/BasicHelper.h`; god forbid these names are ambigious.

My *intent* is to call the library `TestLib` or `TestSupport` and have it under `lib` or somesuch.

markuspf:
Closing this in favour of #17467

PR #17233, opened 2022-09-28
[CMAKE] Activate graph tests

markuspf:
### Scope & Purpose

Kinda wonder whether we should be running these.


tests/CMakeLists.txt
```diff
@@ -418,6 +418,7 @@ add_executable(
   ../cmake/activeCodePage.manifest
   ${ARANGODB_TESTS_SOURCES}
   ${ARANGODB_REPLICATION2_TEST_SOURCES}
+  ${ARANGODB_GRAPH_TEST_SOURCES}
```
mchacki, line 421: I cannot targetly comment, as the lines below that are not touched do not allow it.

However can you add this line also to line 442?

```
if (USE_SEPARATE_GRAPH_TESTS_BINARY)
  add_executable(
          arangodbtests_graph
          ${ARANGODB_TESTS_COMMON_SOURCES}
          main.cpp
  )
```

=>

```
if (USE_SEPARATE_GRAPH_TESTS_BINARY)
  add_executable(
          arangodbtests_graph
          ${ARANGODB_TESTS_COMMON_SOURCES}
          ${ARANGODB_GRAPH_TEST_SOURCES}
          main.cpp
  )
```

Otherweise the "graph_only" test binary is just empty

mchacki (changes requested):
Please add it into another position:

The change itself is great.

tests/CMakeLists.txt
markuspf: Done. As the tests got broken in the meantime, I'll now look at how to fix them.

mchacki (approved):
LGTM

PR #17237, opened 2022-09-29
JS API: Fix undefined reference and logic of collection.iterate()

Simran-B:
### Scope & Purpose

To celebrate the **10th anniversary** of this bug, let's fix it so that we can actually show how the `collection.iterate()` method works in the documentation 🎉 

- `db` is undefined despite the `require("@arangodb").db` import
- higher probability values make it less likely to pick a document  (unless it's >= 1.0)

Commit from January 2013 that added the code (and example): https://github.com/arangodb/arangodb/commit/9c87607f461a34e20fd1c62123e107f166b58e3a

- [x] :hankey: Bugfix

### Checklist

I think we can ignore all of the below because no one noticed until Chris found the confusing example in the docs this year.
I manually tested the fix in `arangosh` and `arangod --console`.

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/1133
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://github.com/arangodb/docs/issues/1119
- [ ] Design document:

### DIFF
--- a/js/common/modules/@arangodb/arango-collection-common.js
+++ b/js/common/modules/@arangodb/arango-collection-common.js
@@ -33,7 +33,6 @@ var arangodb = require('@arangodb');
 
 var ArangoError = arangodb.ArangoError;
 var sprintf = arangodb.sprintf;
-var db = arangodb.db;
 
 var simple = require('@arangodb/simple-query');
 
@@ -322,8 +321,8 @@ ArangoCollection.prototype.iterate = function (iterator, options) {
     if (probability >= 1.0) {
       cursor = this.all();
     }else {
-      stmt = sprintf('FOR d IN %s FILTER rand() >= @prob RETURN d', this.name());
-      stmt = db._createStatement({ query: stmt });
+      stmt = sprintf('FOR d IN %s FILTER rand() <= @prob RETURN d', this.name());
+      stmt = arangodb.db._createStatement({ query: stmt });
 
       if (probability < 1.0) {
         stmt.bind('prob', probability);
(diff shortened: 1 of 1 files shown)


jsteemann (approved):
LGTM

js/common/modules/@arangodb/arango-collection-common.js
```diff
@@ -322,8 +321,8 @@ ArangoCollection.prototype.iterate = function (iterator, options) {
     if (probability >= 1.0) {
       cursor = this.all();
     }else {
-      stmt = sprintf('FOR d IN %s FILTER rand() >= @prob RETURN d', this.name());
-      stmt = db._createStatement({ query: stmt });
+      stmt = sprintf('FOR d IN %s FILTER rand() <= @prob RETURN d', this.name());
+      stmt = arangodb.db._createStatement({ query: stmt });
```
MBkkt, line 325: But why does it different?

It looks absolutely same

MBkkt (commented):
> I think we can ignore all of the below because no one noticed until Chris found the confusing example in the docs this year.
I manually tested the fix in arangosh and arangod --console.

Tests needed not because someone don't trust you.
They are needed to avoid the same mistake in the future

js/common/modules/@arangodb/arango-collection-common.js
Simran-B: db is defined, yet its value is `undefined`. I also tried `print(require("@arangodb").db))` and it logged `undefined`. I'm puzzled why that is.

MBkkt: I mean you remove line
var db = arangodb.db; and write it inplace.

Simran-B:
There is no future for `iterate()`, we deprecated it today for 3.11 and will likely remove it in 4.0. It's not worth the time to write tests for it in my opinion. We have more important features that don't have good test coverage.

PR #17246, opened 2022-09-30
Delay execution of addFollower job in supervision - BTS-1111

neunhoef:
### Scope & Purpose

This is about https://arangodb.atlassian.net/browse/BTS-1111.

We have often felt that an addFollower job (and a FailedFollower job) should not immediately be
executed when it seems to become necessary, since in case of a restart
of a server that server might be back very quickly, get in sync again,
and no harm is done. So far we never saw a possibility to implement this
easily, since the Supervision must not have a "memory". However, here is
an easy trick:

This PR does the following: AddFollower and FailedFollower jobs are *scheduled* right away
as before, they note their `timeCreated` time in the job description.
We now simply prevent the job from being started, if not at least x
seconds have passed since its creation. This gives us exactly the delay
we want. Setting this to 5 minutes should prevent too much havoc in case
of a simple restart and will allow the existing replicas to come back
and quickly get in sync. If this happens within 5 minutes, the
addFollower job notices that it is no longer necessary and will
terminate itself.

The only disadvantage is that in case of a real server loss, we schedule
new replicas a bit slower, increasing the time back to non-degraded
state.

Note that an `AddFollower` job which helps to increase the `replicationFactor` must
be executed right away, otherwise the user has to wait the configured delay until the change
takes action.

- [*] :hankey: Bugfix

### Checklist
- [x] Tests
  - [*] C++ **Unit tests**
- [*] Backports?
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17674
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17676
  - [*] Backport for 3.8: not planned

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1111


maierlars (approved):

neunhoef:
Basicall all unit tests for agency jobs AddFollower or FailedFollower are now broken, since I need to Mock the `config()` method, which is newly exposed in the AgencyInterface. :-(

kvahed (approved):
LGTM

arangod/Agency/AgencyFeature.cpp
```diff
@@ -168,6 +170,28 @@ regular cluster, which needs to handle only a part of the overall load.)");
                          arangodb::options::Flags::DefaultNoComponents,
                          arangodb::options::Flags::OnAgent));
 
+  options
+      ->addOption(
+          "--agency.supervision-delay-add-follower",
+          "delay in supervision, before an AddFollower job is executed [s]",
```
jsteemann, line 176: @Simran-B recently merged a PR that unifies the casing of all startup options descriptions. Convention is now to start with a capital letter and end the description with a `.`
```suggestion
          "The delay in supervision, before an AddFollower job is executed (in seconds).",
```

arangod/Agency/AgencyFeature.cpp
```diff
@@ -168,6 +170,28 @@ regular cluster, which needs to handle only a part of the overall load.)");
...
+          arangodb::options::makeFlags(
+              arangodb::options::Flags::DefaultNoComponents,
+              arangodb::options::Flags::OnAgent))
+      .setIntroducedIn(30906)
+      .setIntroducedIn(31002);
+
+  options
+      ->addOption(
+          "--agency.supervision-delay-failed-follower",
+          "delay in supervision, before a FailedFollower job is executed [s]",
```
jsteemann, line 187:
```suggestion
          "The delay in supervision, before a FailedFollower job is executed (in seconds).",
```

arangod/Agency/Agent.cpp
```diff
@@ -263,7 +263,7 @@ priv_rpc_ret_t Agent::requestVote(term_t termOfPeer, std::string const& id,
 }
 
 /// Get copy of momentary configuration
-config_t const Agent::config() const { return _config; }
+config_t const& Agent::config() const { return _config; }
```
jsteemann, line 266: The code comments isn't quite telling anymore, because the method now returns a reference, not a copy.

arangod/Agency/AgencyFeature.cpp
```diff
@@ -168,6 +170,28 @@ regular cluster, which needs to handle only a part of the overall load.)");
                          arangodb::options::Flags::DefaultNoComponents,
                          arangodb::options::Flags::OnAgent));
 
+  options
+      ->addOption(
+          "--agency.supervision-delay-add-follower",
+          "delay in supervision, before an AddFollower job is executed [s]",
+          new UInt64Parameter(&_supervisionDelayAddFollower),
```
jsteemann, line 177: Also note that this parameter is a uint64_t, whereas other time parameters in this file (e.g. a few lines above) are of type double. This is at least inconsistent.

arangod/Agency/Agent.cpp
```diff
@@ -2513,6 +2513,23 @@ void Agent::updateSomeConfigValues(velocypack::Slice data) {
     _config.setSupervisionGracePeriod(d);
     _supervision->setGracePeriod(d);
   }
+  slice = data.get("delayAddFollower");
+  uint64_t u;
```
jsteemann, line 2517: Same here. Should we use `double` instead?

arangod/Agency/Supervision.cpp
```diff
@@ -3247,8 +3258,21 @@ void arangodb::consensus::enforceReplicationFunctional(
 
               if (actualReplicationFactor < replicationFactor &&
                   apparentReplicationFactor < 2 + replicationFactor) {
+                // Note: If apparentReplicationFactor is smaller than
+                // replicationFactor, then there are fewer servers in the
+                // plan than requested by the user. This means the AddFollower
+                // job is not subject to the configurable delay and is
+                // considered more urgent. This happens, if the
+                // repliactionFactor is increased by the user.
```
jsteemann, line 3266:
```suggestion
                // replicationFactor is increased by the user.
```

arangod/Agency/AgencyFeature.cpp
neunhoef: This is intentional. I had `double` first, but this causes more trouble inside with types and chrono, so I changed this to `uint64_t`.

arangod/Agency/AgencyFeature.cpp
neunhoef: Done.

arangod/Agency/Agent.cpp
neunhoef: Changed.

arangod/Agency/Agent.cpp
neunhoef: No, the delays are intentional `uint64_t`.

jsteemann (approved):
LGTM

PR #17256, opened 2022-10-03
Hack to fix C++ exceptions on MacOS

markuspf:
### Scope & Purpose

No, I do not know what I am doing yet.

### DIFF
--- a/lib/Containers/MerkleTreeHelpers.cpp
+++ b/lib/Containers/MerkleTreeHelpers.cpp
@@ -26,7 +26,6 @@
 #include <string>
 
 #include "MerkleTreeHelpers.h"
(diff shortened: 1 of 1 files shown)


markuspf:
New surprises to be found inside ArangoDB's build system: apparently `lib/Containers/MerkleTreeHelpers.cpp` is built with `-fno-rtti` (which I deem a bad idea), but apparently this is due to snappy being built with `-fno-rtti` (which... is also a bad idea).

In any case, I suspect (but have not proven yet) that this leads to an ODR violation involving `std::exception` or `std::runtime_error`, leading to the funny test failures in the Apple tests.

markuspf:
I think we should use #17270 instead of this one.

PR #17257, opened 2022-10-03
Disable replication1 specific test for replication2

mpoeter:
### Scope & Purpose

Disable shell-following-term-id-cluster tests when running with replication 2.0 since these are specific to the old replication protocol.

- [x] :hankey: Bugfix


mpoeter:
Blue except for unrelated test error

apetenchea (approved):

PR #17271, opened 2022-10-04
avoid possible division by zero if the process may not be running etc.

dothebart:
### Scope & Purpose

If we try to acquired the process statistics of a no more rurnning process, we would run into a division by zero here.
This is mostly relevant to testing, since arangod running this on self has the warranty to acquire a running process.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #17272, opened 2022-10-04
[PREGEL] Separate dir for messages

jvolmer:
This PR moves messaging between Conductor and Workers into a new Messaging directory. There will be more files to come that are directly related to messaging and it is better maintainable if they can all be found at one place.
Additionally, it splits up the long WorkerConductorMessages.h file into 3 files: WorkerMessages.h that includes only messages coming from a worker, ConductorMessages.h that includes only messages coming from a conductor, and Messages.h which includes the general ModernMessage.


romanatarango (approved):
LGTM

PR #17273, opened 2022-10-04
[PREGEL] Add aggregate structs

jvolmer:
Adds 2 structs that will help in changing the WorkerApi to the actor model approach:
In a later PR, the conductor will call the Workers without expecting a direct Response back. The called Workers will responded back with new messages send to the conductor. The new Aggregate structs will make sure that all these responses are aggregated in the conductor accordingly. Additionally, it will return the final aggregated message when all workers responded back.


arangod/Pregel/Messaging/Aggregate.h
```diff
@@ -0,0 +1,75 @@
...
+
+template<Addable T>
+struct Aggregate {
+  Aggregate(T initialValue, uint64_t countUntilFinished)
+      : _countUntilFinished{countUntilFinished},
+        _aggregate{std::move(initialValue)} {}
+  static auto withComponentsCount(uint64_t countUntilFinished) -> Aggregate {
+    return Aggregate{T{}, countUntilFinished};
+  }
+  auto aggregate(T const& message) -> std::optional<T> {
```
romanatarango, line 45: This will copy `message` in the next line if `_aggregate.add(message)` needs to save the message. `auto aggregate(T message)` would allow the caller to pass it as an r-value and to move it. Also `_aggregate.add()` could move it then. Another option is to define both the above and `aggregate(T&& message)` with the same behavior. (@markuspf is it correct?)

arangod/Pregel/Messaging/Aggregate.h
```diff
@@ -0,0 +1,75 @@
...
+ private:
+  uint64_t _countUntilFinished;
+  T _aggregate;
+};
+
+template<typename T>
+struct AggregateCount {
+  AggregateCount(uint64_t countUntilFinished)
+      : _countUntilFinished{countUntilFinished} {}
+  auto aggregate(T const& message) -> bool {
```
romanatarango, line 63: The parameter is not used here.

romanatarango (commented):
LGTM up to two questions.

arangod/Pregel/Messaging/Aggregate.h
markuspf: Here, I think `aggregate(T message)` is the correct thing to use.

using `aggregate(T&& message)` forces the caller to use `std::move` for lvalues; this can be used to make it more obvious that an input is indeed moved.

(I find this here to be a fairly good reference-entrypoint: https://www.foonathan.net/2018/03/rvalue-references-api-guidelines/)

arangod/Pregel/Messaging/Aggregate.h
markuspf: it is part of the API though, hence it needs to be there.

jvolmer: I wanted to make sure that the type of the message is always the same for a specific instance of AggregateCount.

PR #17277, opened 2022-10-05
[CINFRA-588] Restart issues with replication2 databases

apetenchea:
### Scope & Purpose

This PR splits `getResponsibleServer` and `getResponsibleServers` into separate functions for replication1 and replication2. The replication1 code stays the same, while the replication2 functions have been modified to wait for a leader in plan.
If the leader is marked as failed (which is the case after a restart), the `currentTerm->leader` field does not contain any value.
The replication2 version needs a bigger waiting time after a restart.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6001,72 +6001,124 @@ std::vector<ServerID> ClusterInfo::getCurrentDBServers() {
 
 std::shared_ptr<std::vector<ServerID> const> ClusterInfo::getResponsibleServer(
     std::string_view shardID) {
+  if (auto result = getResponsibleServerReplication2(shardID);
+      result != nullptr) {
+    return result;
+  }
+  return getResponsibleServerReplication1(shardID);
+}
```
MBkkt, line 6009: What will be executed if replication2 not used?
Why did this case change?

goedderz: > What will be executed if replication2 not used?

`getResponsibleServerReplication2()` returns a `nullptr` for non-replication2 collections.

> Why did this case change?

Because replication2 works differently from replication1. 🤷‍♂️  I don't understand the question.

MBkkt: For me not clear why if I need responsible server from replication1 data I need to wait 100*500ms tries of replication2 fetch data (of course only in case if replication2 returns something invalid)

goedderz: A valid concern, but you don't have to wait that long: `getResponsibleServerReplication2` breaks out of the loop during the first iteration if this isn't a replication 2 database, so no sleep in case of a replication 1 database. Also note that the current implementation of `getResponsibleServerReplication2` is preliminary and will be replaced soon. There are already ongoing changes in staging, which are necessary to get this working with the recent collection groups implementation (i.e. distributeShardsLike).

apetenchea: For the replication2 case, there is a `break` out of the for loop, at line 6088 in this same file. If replication2 is not used, you will execute the code up to that point, where `it == _replicatedLogs.end()` is going to be true. After this check, we return `nullptr` and continue with the replication1 version.

PR #17278, opened 2022-10-05
Remove TRI_ASSERT from Inspector

markuspf:
### Scope & Purpose

This makes the inspector code independent of ArangoDB core code.


lib/Inspection/VPackLoadInspector.h
```diff
@@ -466,9 +464,7 @@ struct VPackLoadInspectorImpl
     velocypack::Slice slice;
     bool isPresent = false;
     if (auto it = fields.find(name); it != fields.end()) {
-      TRI_ASSERT(!it->second.second)
-          << "field " << name << " processed twice during inspection. "
-          << "Make sure field names are unique!";
+      assert(!it->second.second);
```
mpoeter, line 467:
```suggestion
      assert(!it->second.second &&
        "field processed twice during inspection. Make sure field names are unique!");
```

mpoeter (approved):
Generally LGTM, although I would prefer if we could retain the message in the last two cases.

PR #17280, opened 2022-10-05
[PREGEL] Separate sending and receiving messages to load a graph

jvolmer:
This PR separates the sending of a LoadGraph message (Conductor to Workers) and the sending of the GraphLoaded message (Worker to Conductor). 
This is the first PR that transforms the communication in Pregel to the Actor Model approach. This PR transforms only the communication when loading the graph. More PRs will follow that will transform all the other conductor-worker messaging in a similar way. 
Previously, the conductor (more specifically the conductor Loading state) sended a LoadGraph message to all workers and got a GraphLoaded answer as a direct response. Now, the conductor Loading state sends a LoadGraph message to all workers and gets an immediate Ok back from the workers that only assures that they received the message. When a worker is finished with loading the graph, it sends a new message to the conductor. The conductor Loading state receives all these worker messages, aggregates them and changes into the Computing states if it received a GraphLoaded message from all workers.


romanatarango (approved):
LGTM

markuspf (approved):
LGTM bar two typos

PR #17283, opened 2022-10-05
Implement PrettyPrintInspector

mpoeter:
### Scope & Purpose

Implement a PrettyPrintInspector that can print JSON for any inspectable data type.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**


lib/Inspection/JsonPrintInspector.h
```diff
@@ -237,18 +237,15 @@ struct JsonPrintInspector
...
     }
   }
 
   void decrementIndentationLevel() {
     if (_format == JsonPrintFormat::kPretty) {
-      assert(_indentation.size() >= IndentationPerLevel);
-      for (unsigned i = 0; i < IndentationPerLevel; ++i) {
-        _indentation.pop_back();
-      }
+      assert(_indentation.size() >= IndentationPerLevel &&
```
MBkkt, line 246: Is assert, not tri_assert is international?

mpoeter: Yes, because this is used in some smaller test libraries which should not depend on AssertionLogger etc.

MBkkt (approved):

PR #17290, opened 2022-10-06
Update the gtest module in CMakeLists.txt

markuspf:
### Scope & Purpose

A recent merge of the `staging/replication-2.0` branch changed how we use gtest;

All other submodules are updated by cmake for some reason, so we make it the same  for gtest.


MBkkt (approved):

MBkkt:
> All other submodules are updated by cmake for some reason

Because we don't want to make it manually.
I think better to use fetch content, but it doesn't work in our Jenkins (some download/connection issues)
Of course we can use fetch content without download but it's more difficult and needs change Jenkins configuration

PR #17294, opened 2022-10-06
[BTS-483] Added restriction for query cache usage on transactions

cpjulia:
### Scope & Purpose

PR for https://arangodb.atlassian.net/browse/BTS-483 which adds a restriction for queries coming from a streaming transaction or a JS transaction that is not read-only to neither store nor lookup into the query cache, as such transactions might be still ongoing and not commit, hence there would be a value that might not be persisted in the cache. 

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests WILL ADD
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-483
- [ ] Design document:


tests/js/client/shell/api/cursor-noncluster.js
```diff
@@ -304,5 +306,239 @@ function testing_the_query_cacheSuite () {
...
+          "query": `FOR doc IN ${cn} FILTER doc._key == "a" RETURN doc`,
+          "cache": false
+        });
+        assertEqual(res.code, 201);
+        assertFalse(res.error);
+        assertEqual(res.parsedBody.result[0].value, 2);
+      }
+    },
+
+    test_testing_streaming_js_transaction_aborted_set: function() {
```
Simran-B, line 477: The JavaScript Transactions are not Stream Transactions
```suggestion
    test_testing_js_transaction_aborted_set: function() {
```

tests/js/client/shell/api/cursor-noncluster.js
```diff
@@ -304,5 +306,239 @@ function testing_the_query_cacheSuite () {
...
+          "query": `FOR doc IN ${cn} FILTER doc._key == "a" RETURN doc`,
+          "bindVars": {},
+          "cache": false
+        });
+        assertEqual(res.code, 201);
+        assertEqual(res.parsedBody.result[0].value, 0);
+      }
+    },
+
+    test_testing_streaming_js_transaction_committed_set: function() {
```
Simran-B, line 511:
```suggestion
    test_testing_js_transaction_committed_set: function() {
```

Simran-B (changes requested):

arangod/Aql/Query.cpp
```diff
@@ -600,7 +599,10 @@ ExecutionState Query::execute(QueryResult& queryResult) {
         }
         // will set warnings, stats, profile and cleanup plan and engine
         auto state = finalize(*queryResult.extra);
-        if (state == ExecutionState::DONE && _cacheEntry != nullptr) {
+        if (state == ExecutionState::DONE && _cacheEntry != nullptr &&
+            !((_transactionContext->isStreaming() ||
+               _transactionContext->isTransactionJS()) &&
+              !_trx->state()->isReadOnlyTransaction())) {
```
jsteemann, line 605: This `if` condition is mind-blowing. double negation :scream: 
Can it be written in a simpler way?

arangod/Transaction/Context.h
```diff
@@ -103,6 +103,30 @@ class Context {
...
+  /// whether or not can read from query cache)
+  bool isStreaming() { return _transaction.isStreamingTransaction; }
+
+  /// @brief whether or not transaction is JS (used to know
+  /// whether or not can read from query cache)
+  bool isTransactionJS() { return _transaction.isJStransaction; }
+
+  /// @brief whether or not the transaction is read-only (used to know whether
+  /// or not can read from query cache)
+  bool isEarlyReadOnly() { return _transaction.isEarlyReadOnly; }
```
jsteemann, line 116: All these should be `const noexcept`.

arangod/Transaction/Context.h
```diff
@@ -141,6 +165,9 @@ class Context {
     TransactionId id;
     bool isReadOnlyTransaction;
     bool isFollowerTransaction;
+    bool isStreamingTransaction = false;
+    bool isEarlyReadOnly = false;
+    bool isJStransaction = false;
```
jsteemann, line 170: I am not really happy with having separate booleans for `isStreamingTransaction` and `isJSTransaction`, because this allows to set both to `true` at the same time, which would at least in theory be invalid.

arangod/V8Server/v8-vocbase.cpp
```diff
@@ -894,13 +894,21 @@ static void JS_ExecuteAql(v8::FunctionCallbackInfo<v8::Value> const& args) {
...
+      v8Context->setJStransaction();
+    }
+    if (v8g->_transactionContext->isEarlyReadOnly()) {
+      v8Context->setEarlyReadOnly();
+    }
+  }
   auto query = arangodb::aql::Query::create(
-      transaction::V8Context::Create(vocbase, true),
-      aql::QueryString(std::move(queryString)), std::move(bindVars),
+      v8Context, aql::QueryString(std::move(queryString)), std::move(bindVars),
```
jsteemann, line 911:
```suggestion
      std::move(v8Context), aql::QueryString(std::move(queryString)), std::move(bindVars),
```

arangod/Transaction/Options.h
```diff
@@ -87,6 +87,8 @@ struct Options {
   /// created.
   bool allowDirtyReads = false;
 
+  bool isEarlyReadOnly = true;
```
jsteemann, line 90: What does this flag do? Under what circumstances can it be used? At what time is its value consided valid? Code comments would help here...

jsteemann: I would prefer to **not** have `ìsEarlyReadOnly` inside the transaction's options, because it is not something that an end user should set from the outside. And the flag only seems to be used inside `transaction::Manager`, so can we remove it again from the options?

arangod/Transaction/Context.h
```diff
@@ -141,6 +165,9 @@ class Context {
     TransactionId id;
     bool isReadOnlyTransaction;
     bool isFollowerTransaction;
+    bool isStreamingTransaction = false;
+    bool isEarlyReadOnly = false;
```
jsteemann, line 169: It is puzzling that there are now 2 different read-only flags for one transaction, which may or may not contradict each other. I think we should go with only a single flag, preferrably `isReadOnlyTransaction`.
The name `isEarlyReadOnly` leaves a lot of questions, mainly, what does "early" mean, if it will be accurate after the "early" stage of the transaction etc. Better to have only a single, reliable flag.

arangod/RestHandler/RestVocbaseBaseHandler.cpp
```diff
@@ -683,7 +683,10 @@ RestVocbaseBaseHandler::createTransactionContext(AccessMode::Type mode) const {
           "illegal to start a managed transaction here");
     }
     if (value.compare(pos, std::string::npos, " aql") == 0) {
-      return std::make_shared<transaction::AQLStandaloneContext>(_vocbase, tid);
+      auto aqlStandaloneContext =
+          std::make_shared<transaction::AQLStandaloneContext>(_vocbase, tid);
+      aqlStandaloneContext->setStreaming();
```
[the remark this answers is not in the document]
cpjulia: it's because, if you have reached this region, it means that the transaction id was in the header of the request, as it happens after the 
```
 if (!found) {
    return std::make_shared<transaction::StandaloneContext>(_vocbase);
  }
 ``` 
 in line 659

File: arangod/Transaction/Context.h (hunk 5 above)
Comment (cpjulia) [reply arangod/Transaction/Context.h]: I added assertions to be sure that both are not true 
```
 /// @brief sets the transaction to be streaming (used to know whether or not
  /// can read from query cache)
  void setStreaming() noexcept {
    TRI_ASSERT(_transaction.isJStransaction == false);
    _transaction.isStreamingTransaction = true;
  }

  /// @brief sets the transaction to be JS (used to know whether or not
  /// can read from query cache)
  void setJStransaction() noexcept {
    TRI_ASSERT(_transaction.isStreamingTransaction == false);
    _transaction.isJStransaction = true;
  }

```

File: arangod/Aql/Query.cpp [hunk 10]
```diff
@@ -617,7 +616,10 @@ ExecutionState Query::execute(QueryResult& queryResult) {
         }
         // will set warnings, stats, profile and cleanup plan and engine
         auto state = finalize(*queryResult.extra);
-        if (state == ExecutionState::DONE && _cacheEntry != nullptr) {
+        bool isTransactionDisallowed = _transactionContext->isStreaming() &&
+                                       !_trx->state()->isReadOnlyTransaction();
+        if (state == ExecutionState::DONE && _cacheEntry != nullptr &&
+            !isTransactionDisallowed) {
```
jsteemann, line 622: Can we change the naming here from `isTransactionDisallowed` to `isCachingDisallowed`, because that's what it actually is about. In fact, `isCachingAllowed` would be an ever better name, because then we can avoid the double negation in `! isTransactionDisallowed` (or `! isCachingDisallowed`).

arangod/Aql/Query.cpp
```diff
@@ -860,7 +862,10 @@ QueryResultV8 Query::executeV8(v8::Isolate* isolate) {
       ss->waitForAsyncWakeup();
       state = finalize(*queryResult.extra);
     }
-    if (_cacheEntry != nullptr) {
+    bool isTransactionDisallowed = _transactionContext->isTransactionJS() &&
```
jsteemann, line 865: Same here.

arangod/Aql/Query.cpp
```diff
@@ -1380,7 +1385,11 @@ uint64_t Query::calculateHash() const {
 
 /// @brief whether or not the query cache can be used for the query
 bool Query::canUseQueryCache() const {
-  if (_queryString.size() < 8 || _queryOptions.silent) {
+  bool isTransactionDisallowed = (_transactionContext->isStreaming() ||
```
jsteemann, line 1388: Same here.

PR #17297, opened 2022-10-06
[CINFRA-579] Enabling range deletes on followers

apetenchea:
### Scope & Purpose

This PR enables range deletes for truncate operations on followers. Before this, the followers always used remove operations, regardless of what the leader did.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature


maierlars (approved):
I remember that range deletes caused performance issues in the past. Is this still a thing?

apetenchea:
> I remember that range deletes caused performance issues in the past. Is this still a thing?

From the tests I ran it is much faster than using remove operations.

PR #17306, opened 2022-10-10
[PREGEL] Have only one internal pregel endpoint

jvolmer:
This PR combines the internal pregel endpoints 'worker' and 'conductor' into one single endpoint. For this to work, I combined the process function in both worker and conductor into one process function in the feature which handles all incoming messaging. This new process function can now be called either in worker or conductor, we don't need special cases any more.
This also simplifies the direct connection: We can use a single struct for both worker and conductor.


arangod/Pregel/PregelFeature.cpp
```diff
@@ -650,89 +658,146 @@ auto PregelFeature::cleanupWorker(ExecutionNumber executionNumber)
...
+              TRI_ASSERT(created.get() != nullptr);
+              addWorker(std::move(created), executionNumber);
+              return {
+                  WorkerCreated{.senderId = ServerState::instance()->getId()}};
+            } catch (basics::Exception& e) {
+              return Result{e.code(), e.message()};
+            }
+          },
+          [&](LoadGraph const& x) -> ResultT<MessagePayload> {
+            auto w = worker(executionNumber);
```
markuspf, line 697: There's a lot of duplication going on here, but I could not yet think of a way that solves this problem just now;

As a longer-term solution the destination of the message should contain the execution number, or better yet, we just know which workers we started as conductor and speak to them directly.

PR #17314, opened 2022-10-11
[BTS-908] UI Graph Viewer: add + remove edge broken 

hkernbach:
### Scope & Purpose

WebUI Bug-fix

Problem: If a graph definition of a named graph only had one edge relation defined, new edges couldn't be created through the graph viewer.

Explanation: In cases where we distinguished between rendering either a "Select-Box" (in case we have multiple entries) or rendering a static text div (readOnlyEntry) we also need to read the "text" or "value" out of that specific element based on which html element we've rendered. In the "text" case this lead to trouble.

Example: readOnlyEntry rendered as "knows", but as soon we've read that entry element with jQuerie's text() method it got captured as: `\n     knows \n`, which then lead to a wrong request being sent to the server. 

Fix: 
Always render a "Select element", even if we only have one option available. In this specific case the element will be rendered anyway but will be disabled instead. Additionally, there is now a title set which indicates that there is only one entry element to choose - therefore disabled.

- [x] :hankey: Bugfix

### Checklist

- [x] Backports (yes required)
  - [x] Backport for devel: https://github.com/arangodb/arangodb/pull/17315
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17316
  - [x] Backport for 3.8: not required (bug got introduced during refactorings in 3.9+)

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-908


hkernbach:
Jenkins green, review still required.

adityamukho (approved):
LGTM

ajurdak (approved):
LGTM

maxkernbach (approved):

PR #17317, opened 2022-10-11
[PREGEL] Feature/ColorPropagation

romanatarango:
### Scope & Purpose

This PR introduces a new Pregel algorithm: Color Propagation. For the input it is expected that vertices have a field containing colors. (The name of the field can be passed to the algorithm as a user parameter in a call.) In addition, there should be a field containing an identifier (a number) of an equivalence class. Vertices whose values of this field are the same, belong to the same equivalence class, otherwise to different classes. (The name of the field can be passed as well.) The algorithm propagates a color from a vertex to its successors and adds it at the successors. Edges within the same equivalence class are ignored. This continues until a fixed point is reached. The resulting colors are written into a vertex property whose name can be passed as a user parameter.

The algorithm is intended as a part of the graph summarization project.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **integration tests** (will follow)
- [x] :book: CHANGELOG entry made (will follow)
- [x] :books: documentation written (release notes, API changes, ...) (will follow)

#### Related Information

- [x] Docs PR: (will follow)


arangod/Pregel/AlgoRegistry.cpp
```diff
@@ -181,7 +188,7 @@ template<typename V, typename E, typename M>
                             server, userParams),
                         body, feature);
   }
-
+  LOG_DEVEL << "Algorithm: " << algorithm;
```
markuspf, line 191:
```suggestion
```

arangod/Pregel/Algorithm.h
```diff
@@ -110,17 +112,16 @@ struct Algorithm : IAlgorithm {
...
+  [[nodiscard]] virtual uint32_t messageBatchSize(
+      WorkerConfig const& config, MessageStats const& stats) const {
     if (config.localSuperstep() == 0) {
       return 500;
     } else {
-      double msgsPerSec = stats.sendCount / stats.superstepRuntimeSecs;
-      msgsPerSec /= config.parallelism();  // per thread
+      double msgsPerSec =
+          static_cast<double>(stats.sendCount) / stats.superstepRuntimeSecs;
+      msgsPerSec /= static_cast<double>(config.parallelism());  // per thread
```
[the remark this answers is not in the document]
romanatarango: I see your point and I agree. On the other hand, (1) it will realistically never be done in another way, (2) it distracts because my IDE marks such places, (3) it is relatively easy to review if such changes are not too numerous in one PR and do not interleave with the actual changes (a lot).

arangod/Pregel/AlgoRegistry.cpp
```diff
@@ -42,6 +42,7 @@
 #include "Pregel/Utils.h"
 #if defined(ARANGODB_ENABLE_MAINTAINER_MODE)
 #include "Pregel/Algos/ReadWrite.h"
+#include "Pregel/Algos/ColorPropagation.h"
```
markuspf, line 45: This needs to be outside of the maintainermode ifdef

markuspf (approved):

PR #17322, opened 2022-10-11
setUpAll/tearDownAll for performance reasons

dothebart:
### Scope & Purpose

tests seem to get to slow during the coverage tests. 

- [x] :hankey: Bugfix

### DIFF
--- a/tests/js/client/shell/shell-iterators-read-standalone.js
+++ b/tests/js/client/shell/shell-iterators-read-standalone.js
@@ -46,7 +46,6 @@ function StandaloneAqlIteratorReadSuite() {
       run(ctx, opts);
     });
   };
-  global.read = true;
   
   let suite = {};
   deriveTestSuite(base.IteratorReadSuite(permute, true), suite, '_StandaloneAqlRead');
--- a/tests/js/client/shell/shell-iterators-read-standalone.js
+++ b/tests/js/client/shell/shell-iterators-read-standalone.js
@@ -1,0 +1,58 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global fail, assertTrue, assertFalse, assertEqual, assertNotUndefined, arango */
+
+// //////////////////////////////////////////////////////////////////////////////
+// / @brief ArangoTransaction sTests
+// /
+// /
+// / DISCLAIMER
+// /
+// / Copyright 2018 ArangoDB GmbH, Cologne, Germany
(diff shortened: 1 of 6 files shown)


tests/js/client/shell/shell-iterators.js
```diff
@@ -84,7 +84,7 @@ function IteratorSuite(permuteConfigs) {
   };
 
   return {
-    setUp: function () {
+    setUpAll: function () {
```
jsteemann, line 87: This is not safe.
There are tests in here that **modify** data in the the collection, e.g.
https://github.com/arangodb/arangodb/pull/17322/files#diff-3588a0693f1c76defd4f9f07238e7e37fca75b717e998352ac9c1fd71f21ca1aR226

jsteemann (changes requested):

tests/js/client/shell/shell-iterators-read-standalone.js
```diff
@@ -0,0 +1,58 @@
...
+    query: (...args) => internal.db._query(...args),
+    collection: (name) => internal.db[name],
+    abort: () => {},
+  };
+  const permute = function(run) {
+    [ {}, {intermediateCommitCount: 111} ].forEach((opts) => {
+      run(ctx, opts);
+    });
+  };
+  global.read = true;
```
jsteemann, line 49: What is the purpose of this global? I tried to search for it in this PR, but didn't find any clues. It seems it is just set here, but not read back?

tests/js/client/shell/shell-iterators-write-standalane.js
```diff
@@ -0,0 +1,58 @@
...
+    query: (...args) => internal.db._query(...args),
+    collection: (name) => internal.db[name],
+    abort: () => {},
+  };
+  const permute = function(run) {
+    [ {}, {intermediateCommitCount: 111} ].forEach((opts) => {
+      run(ctx, opts);
+    });
+  };
+  global.read = true;
```
jsteemann, line 49: What is the purpose of this global? I tried to search for it in this PR, but didn't find any clues. It seems it is just set here, but not read?

tests/js/client/shell/shell-iterators-read-standalone.js
dothebart: will delete. leftover of tries.

jsteemann (approved):
LGTM

PR #17359, opened 2022-10-14
[GWEN] Feature/fast wcc

romanatarango:
### Scope & Purpose

Implements an experimental WCC algorithm without Pregel. Currently, the graph should be given by a `SharedSlice` of the form 
```c++
R"({ "vertices": [ {"_key": "A", "value": 5}, ... ],  "edges":    [ {"_key": "", "_from": "A", "_to": "B"}, ... ] })"_vpack
```
the intention is to have an algorithm faster than a Pregel implementation. The algorithm uses a UnionFind structure (aka disjoint sets). Initially, each vertex is its own component. When an edge is read, the sets containing the ends of the edge are merged (the union operation). At the end, each disjoint set is a weakly connected component. An advantage of this algorithm is that it is online with respect to edges. 

- [x] :pizza: New feature
- [x] :fire: Performance improvement (to be tested)

### Checklist

- [x] C++ **Unit tests**
- [x] :book: CHANGELOG entry made (to be done if decide to incorporate)
- [x] :books: documentation written in the header file

### DIFF
--- a/arangod/Pregel/tests/DisjointSet.cpp
+++ b/arangod/Pregel/tests/DisjointSet.cpp
@@ -37,7 +37,7 @@ auto DisjointSet::addSingleton(size_t element, size_t hintSize) -> bool {
     return false;
   }
   _parent[element] = element;
-  _rank[element] = 1; // 0 reserved for "not added"
+  _rank[element] = 1;  // 0 reserved for "not added"
   return true;
 }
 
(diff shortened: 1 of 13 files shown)


arangod/Pregel/tests/DisjointSet.cpp
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+auto DisjointSet::capacity() -> size_t { return _parent.size(); }
+
+auto DisjointSet::contains(size_t element) -> bool {
+  return element < capacity() and _rank[element] != 0;
+}
+
+auto DisjointSet::print() -> void {
```
romanatarango, line 18: Will be removed before merging.

PR #17361, opened 2022-10-17
disable test until deadlock is fixed

dothebart:
### Scope & Purpose

This test will deadlock and destabilize the CI in terms of runtime and crashreport sizes.
Hence disabling it for now.


alexbakharew:
Do we have some reminder to uncomment it after the bugfix?

dothebart:
created BTS-1083 to track this.

apetenchea (changes requested):
Hi! Let me disable just that particular test file which caused the crash.

dothebart:
OK!

apetenchea (changes requested):

PR #17366, opened 2022-10-17
[PREGEL] Wait until Worker received all messages that were sent to it

jvolmer:
The previous PR #17332 merges the prepareGss and runGss message send from the conductor to the workers into a single message from conductor to workers. It turned out that when running the pregel tests several times in cluster mode, not all messages were delivered. We never check inside the Worker that for every superstep all send messages are also received. Before merging the two messages, there was possibly enough time to deliver all messages, but now this missing check results sometimes in wrong results.
This PR fixes this: In the worker - or more exact in the OutCache - we have a counter for each send message in a gss. When the worker is finished with its gss, it send this count back to the conductor - more exact the ComputingState. In the next gss, the Computing states sends to each worker the number of messages that were send to this worker. Inside the worker, before the worker actually updates its gss, it waits until all send messages are received. (Special for the CombiningOutCache: send count should is not increased when message is just combined, only when it is send to another worker).
Caveat: The worker sends messages to shards, not servers - it does not know which leader shard can be found on which server. Therefore inside the worker we count the number of messages send to each shard, send that to the conductor and here convert this map into a number of messages send to each server. (During initialization of the workers, the conductor knows the assignment of which leader shard is located on which server, we just have to save that in a conductor property to use it later during the computation).
Another annoying thing: The worker is already initialized with gss = 0. Therefore we cannot differentiate between a worker that was just initialized and gets a message from the conductor to change to gss=0 and a worker that has just finished gss=0 and gets a message from the conductor to change to gss=1. This is relevant to receive each message in its dedicated gss. I solved this by adding a new property to the worker _computationStarted that just changes to true after the message from the condctor with gss=0 arrived. When we have the Worker state machine, we should improve that code and get rid of this _computationStarted variable.

### DIFF
--- a/arangod/Pregel/Worker/Worker.h
+++ b/arangod/Pregel/Worker/Worker.h
@@ -85,6 +85,7 @@ class VertexContext;
 struct VerticesProcessed {
   VPackBuilder aggregator;
   MessageStats stats;
+  std::unordered_map<ShardID, uint64_t> sendCountPerShard;
   size_t activeCount;
 };
 
@@ -141,6 +142,10 @@ class Worker : public IWorker {
   /// current number of running threads
   size_t _runningThreads = 0;
   Scheduler::WorkHandle _workHandle;
+  // distinguishes config.globalSuperStep being initialized to 0 and
+  // config.globalSuperStep being explicitely set to 0 when the first superstep
+  // starts: needed to process incoming messages in its dedicated gss
+  bool _computationStarted = false;
 
   using VerticesProcessedFuture =
       futures::Future<std::vector<futures::Try<ResultT<VerticesProcessed>>>>;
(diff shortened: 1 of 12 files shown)


romanatarango (approved):
lgtm

arangod/Pregel/Worker/Worker.h
```diff
@@ -141,6 +142,10 @@ class Worker : public IWorker {
   /// current number of running threads
   size_t _runningThreads = 0;
   Scheduler::WorkHandle _workHandle;
+  // distinguishes config.globalSuperStep beeeing initialized to 0 and
```
markuspf, line 145:
```suggestion
  // distinguishes config.globalSuperStep being initialized to 0 and
```

arangod/Pregel/Worker/Worker.h
```diff
@@ -141,6 +142,10 @@ class Worker : public IWorker {
...
+  // config.globalSuperStep beeing explicitely set to 0 when the first superstep
```
markuspf, line 146:
```suggestion
  // config.globalSuperStep being explicitly set to 0 when the first superstep
```

markuspf (approved):

PR #17339, opened 2022-10-13
[PREGEL] Separate sending and receiving messages to run gss

jvolmer:
This PR continues the refactoring to use the actor model for communication between conductor and workers (see previous PR https://github.com/arangodb/arangodb/pull/17280)
This adapts the communication in the computation state: sending RunGlobalSuperstep from Conductor to Workers and GlobalSuperStepFinished from Workers to Conductor.

PR #17370, opened 2022-10-17
[PREGEL] Separate sending and receiving for initializing workers

jvolmer:
This PR continues the refactoring to use the actor model for communication between conductor and workers (see previous PR https://github.com/arangodb/arangodb/pull/17339)
This adapts the communication in the initial state: sending CreateWorker from Conductor to Workers and WorkerCreated from Workers to Conductor.
As a preparation, this PR moves the worker creation from the loading state into the initial state - it better fits in here. Additionally, it moves the functions in the conductor that are only responsible for the worker creation into the initial state - to tidy up the conductor which is way to large.


romanatarango (approved):
lgtm

PR #17371, opened 2022-10-18
Introduce ErrorT and improved version of (de)serializeWithStatus

markuspf:
### Scope & Purpose

Trying to use the inspection code outside of what is ArangoDB code shows that `VPack.h` currently uses ArangoDB's THROW and other exception facilities.

As a mitigation `VPack.h` contains numerous variants of `deserializeWithStatus` which returns the status of deserialisation, but takes a parameter for output.

This seemed a bit awkward to me, so I went on a journey trying to come up with a return type for `deserializeWithStatus`, which looks suspiciously not entirely unlike `ResultT`.


lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+namespace arangodb::inspection {
+
+template<typename T>
+struct StatusT {
+  using Contained = std::variant<Status, T>;
+
+  auto static error(Status&& status) -> StatusT<T> {
+    return StatusT(
+        std::move(Contained(std::in_place_index<0>, std::move(status))));
+  }
```
goedderz, line 42: Did you omit the corresponding copy constructor on purpose?

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+
+namespace arangodb::inspection {
+
+template<typename T>
+struct StatusT {
+  using Contained = std::variant<Status, T>;
+
+  auto static error(Status&& status) -> StatusT<T> {
+    return StatusT(
+        std::move(Contained(std::in_place_index<0>, std::move(status))));
```
mpoeter, line 41: Why the extra move? This should suffice:
```suggestion
    return StatusT(
        Contained(std::in_place_index<0>, std::move(status)));
```

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+  }
+  auto static ok(T const& val) -> StatusT<T> {
+    return StatusT(std::move(Contained(std::in_place_index<1>, val)));
```
mpoeter, line 44:
```suggestion
    return StatusT(Contained(std::in_place_index<1>, val));
```

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+  }
+  auto static ok(T&& val) -> StatusT<T> {
+    return StatusT(
+        std::move(Contained(std::in_place_index<1>, std::move(val))));
```
mpoeter, line 48:
```suggestion
    return StatusT(Contained(std::in_place_index<1>, std::move(val)));
```

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+      : _contained{T{}} {}
+
+  StatusT& operator=(T const& val_) {
+    _contained = val_;
+    return *this;
+  }
+  StatusT& operator=(T&& val_) {
+    _contained = std::move(val_);
+    return *this;
+  }
```
mpoeter, line 65: Can't we default those?

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+
+  T const* operator->() const { return &get(); }
+
+  T& operator*() & { return get(); }
+
+  T const& operator*() const& { return get(); }
+
+  T&& operator*() && { return std::move(get()); }
+
+  T const&& operator*() const&& { return get(); }
```
mpoeter, line 84: const rvalue references are pointless. IMHO this should be removed.

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+
+  explicit operator bool() const requires(!std::is_same_v<T, bool>) {
+    return ok();
+  }
+
+ private : StatusT(std::variant<Status, T>&& val)
+     : _contained(std::move(val)) {}
```
mpoeter, line 91: This looks strange!

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+  }
+  auto static ok(T const& val) -> StatusT<T> {
+    return StatusT(std::move(Contained(std::in_place_index<1>, val)));
+  }
+  auto static ok(T&& val) -> StatusT<T> {
+    return StatusT(
+        std::move(Contained(std::in_place_index<1>, std::move(val))));
+  }
+
+  StatusT(StatusT const& other) = delete;
```
goedderz, line 51: If you delete the copy constructor, I suppose to prevent accidental copies, do you want to add a `.copy()` method?

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+  StatusT(StatusT&& other) = default;
+
+  // T is default constructible, otherwise inspection wouldn't work
+  StatusT() requires(!std::is_nothrow_default_constructible_v<T>)
+      : _contained{T{}} {}
+
+  StatusT& operator=(T const& val_) {
+    _contained = val_;
+    return *this;
+  }
```
goedderz, line 61: I guess you want to allow copy assignment for some common patterns (e.g. reuse the same variable), even while disallowing copy construction? Perhaps you could add a small comment on the thoughts behind this.

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,95 @@
...
+#include <type_traits>
+
+#include <optional>
+
+#include "Inspection/Status.h"
+
+namespace arangodb::inspection {
+
+template<typename T>
+struct StatusT {
```
romanatarango, line 36: `T` could be requested not to be derived/convertible from/to `Status`.

tests/Basics/InspectionTest.cpp
```diff
@@ -3021,4 +3023,48 @@ TEST(ValidateInspectorContext, validate_with_context) {
...
+
+  EXPECT_EQ(res->s, "ReturnNode");
+  EXPECT_EQ(res->id, 3u);
+}
+
+TEST(VPackWithStatus, statust_test_deserialize_fail) {
+  auto testSlice = R"({
+    "s": "ReturnNode",
+    "id": 3,
+    "fehler": 2
```
romanatarango, line 3058: "error"?

tests/Basics/InspectionTest.cpp
```diff
@@ -3021,4 +3023,48 @@ TEST(ValidateInspectorContext, validate_with_context) {
...
+  })"_vpack;
+
+  auto res = deserializeWithStatusT<StatusTTest>(testSlice);
+
+  ASSERT_FALSE(res.ok()) << fmt::format("Did not detect the error we exepct");
+
+  EXPECT_EQ(res.error(), "Found unexpected attribute 'fehler'");
```
romanatarango, line 3065: error

romanatarango (approved):
Good interface.

lib/Inspection/StatusT.h
markuspf: It indeed does, I have no idea why my IDE's clang format did that.

markuspf:
I addressed some (most?) of the comments and will run a jenkins now. (That is not to say that I don't welcome more comments!).

@goedderz suggestion of making an `Either`-like type I like, I will experiment with that in a separate PR. It looks like it will turn out to be just factoring out what currently is StatusT plus an additional template parameter.

lib/Inspection/StatusT.h
romanatarango: What is the advantage of `StatusT<Status>` compared with `Status`?

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,86 @@
...
+
+  StatusT(StatusT const& other) = delete;
+  StatusT(StatusT&& other) = default;
+
+  // T is default constructible, otherwise inspection wouldn't work
+  StatusT() requires(!std::is_nothrow_default_constructible_v<T>)
+      : _contained{T{}} {}
+
+  [[nodiscard]] auto ok() const noexcept -> bool {
+    return std::holds_alternative<T>(_contained);
```
mpoeter, line 56: If you allow `T` to be `Status` then this line is not correct. But as I said, I think this would not be a good idea.

lib/Inspection/StatusT.h
mpoeter: I think `StatusT<Status>` would only cause confusion and should therefore not be allowed.

lib/Inspection/StatusT.h
```diff
@@ -0,0 +1,86 @@
...
+  }
+  [[nodiscard]] auto error() -> std::string const& {
+    return std::get<0>(_contained).error();
+  }
+  [[nodiscard]] auto path() -> std::string const& {
+    return std::get<0>(_contained).path();
+  }
+
+  [[nodiscard]] auto get() const -> T const& { return std::get<1>(_contained); }
+  [[nodiscard]] auto get() -> T& { return std::get<T>(_contained); }
```
mpoeter, line 66:
```suggestion
  [[nodiscard]] auto get() -> T& { return std::get<1>(_contained); }
```

lib/Inspection/StatusT.h
goedderz: I wouldn't disallow it. Yes, it's confusing, but it also might come up naturally when composing interfaces, and I don't see why someone would do it if not necessary (neither by accident nor on purpose), so I think there's no harm in keeping the option.

markuspf: `StatusT<Status>` just means that you have an operation that *either* yields an error, which is reflected in the *error* part of the variant, and happens to be a `Status`, *or* a successful operation that is reflected in the *get* part of the variant which happens to be a `Status`.

Sure, it might be a bad life choice to do this, but it unfortunately comes up sometimes. I don't want to stand in programmer's way too much (in particular in view of `ErrorT` introduced in #17386)

markuspf:
Jenkins on this is actually green.

PR #17382, opened 2022-10-18
[CINFRA-582] Apply Snapshot On Follower

apetenchea:
### Scope & Purpose

This PR provides the basic implementation of a snapshot transfer on follower. Currently the leader send over a couple of dummy values. On the follower, we make sure to truncate the shard, then create the appropriate transactions and apply these.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
@@ -36,7 +36,8 @@ using namespace arangodb::replication2::replicated_state::document;
 DocumentFollowerState::DocumentFollowerState(
     std::unique_ptr<DocumentCore> core,
     std::shared_ptr<IDocumentStateHandlersFactory> handlersFactory)
-    : _networkHandler(handlersFactory->createNetworkHandler(core->getGid())),
+    : shardId(core->getShardId()),
+      _networkHandler(handlersFactory->createNetworkHandler(core->getGid())),
       _transactionHandler(
           handlersFactory->createTransactionHandler(core->getGid())),
       _guardedData(std::move(core)) {}
@@ -65,14 +66,21 @@ auto DocumentFollowerState::acquireSnapshot(ParticipantId const& destination,
                   TRI_ERROR_CLUSTER_NOT_FOLLOWER);
             }
 
+            if (auto truncateRes = self->truncateLocalShard(); truncateRes.fail()) {
+              return ResultT<velocypack::SharedSlice>::error(truncateRes);
+            }
+
             // A follower may request a snapshot before leadership has been
             // established. A retry will occur in that case.
             auto leaderInterface =
                 self->_networkHandler->getLeaderInterface(destination);
             return leaderInterface->getSnapshot(waitForIndex);
           })
-      .thenValue([](auto&& result) -> futures::Future<Result> {
-        return result.result();
+      .thenValue([self = shared_from_this()](auto&& result) -> futures::Future<Result> {
+        if (result.fail()) {
+          return result.result();
+        }
+        return self->populateLocalShard(result.get());
       });
 }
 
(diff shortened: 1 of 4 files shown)


arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -96,4 +106,38 @@ auto DocumentFollowerState::applyEntries(
...
+  if (auto applyRes = _transactionHandler->applyEntry(doc); applyRes.fail()) {
+    _transactionHandler->removeTransaction(
+        TransactionId{0}.asFollowerTransactionId());
+    return applyRes;
+  }
+  auto commit = DocumentLogEntry{std::string(shardId),
+                                 OperationType::kCommit,
+                                 {},
+                                 TransactionId{0}.asFollowerTransactionId()};
+  return _transactionHandler->applyEntry(commit);
```
mpoeter, line 127: We should use `TransactionId::createFollower()` to create a new follower transactionId and use that one throughout the function.

arangod/Replication2/StateMachines/Document/DocumentStateMethods.cpp
```diff
@@ -98,6 +97,32 @@ class DocumentStateMethodsDBServer final : public DocumentStateMethods {
...
+    if (auto res = singleCollectionTrx.begin(); res.fail()) {
+      return ResultT<std::unique_ptr<ReplicationIterator>>::error(
+          res.errorNumber(), res.errorMessage());
+    }
+
+    auto* physicalCollection = logicalCollection->getPhysical();
+    TRI_ASSERT(physicalCollection != nullptr);
+
+    return physicalCollection->getReplicationIterator(
+        ReplicationIterator::Ordering::Revision, singleCollectionTrx);
```
mpoeter, line 123: This won't work like this. `getReplicationIterator` takes the `Methods`  via reference, but the `singleCollectionTrx` is going out of scope an will be destroyed.

mpoeter (changes requested):

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -35,8 +35,9 @@ using namespace arangodb::replication2::replicated_state::document;
 
 DocumentFollowerState::DocumentFollowerState(
     std::unique_ptr<DocumentCore> core,
-    std::shared_ptr<IDocumentStateHandlersFactory> handlersFactory)
-    : _networkHandler(handlersFactory->createNetworkHandler(core->getGid())),
+    std::shared_ptr<IDocumentStateHandlersFactory> const& handlersFactory)
+    : shardId(core->getShardId()),
```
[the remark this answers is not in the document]
apetenchea: The core is a guarded member. Because the `shardId` string view remains constant, I decided to store it like that, so I don't have acquire any lock to retrieve it.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
apetenchea: I didn't use that one because I was not sure how the follower transaction is generated - randomly, or incrementally. I want to avoid a scenario when the follower generates such an id and then a transaction coming from the leader uses the same.

arangod/Replication2/StateMachines/Document/DocumentStateMethods.cpp
apetenchea: Thanks!

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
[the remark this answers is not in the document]
mpoeter: Ah, ok - that makes sense then!

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
mpoeter: The transaction id is generated using a hybrid logical clock, so this is safe and there won't be any conflicts.

apetenchea: Good to know, will do the change then.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -106,4 +116,37 @@ auto DocumentFollowerState::applyEntries(
...
+  if (auto applyRes = _transactionHandler->applyEntry(doc); applyRes.fail()) {
+    _transactionHandler->removeTransaction(TransactionId::createFollower());
+    return applyRes;
+  }
+  auto commit = DocumentLogEntry{std::string(shardId),
+                                 OperationType::kCommit,
+                                 {},
+                                 TransactionId{0}.asFollowerTransactionId()};
+  return _transactionHandler->applyEntry(commit);
+}
```
mpoeter, line 137: The transaction id is used in multiple places, so we need to create it once and use it consistently:
```suggestion
  auto trxId = TransactionId::createFollower();
  auto doc = DocumentLogEntry{std::string(shardId), opType, std::move(slice), trxId};
  if (auto applyRes = _transactionHandler->applyEntry(doc); applyRes.fail()) {
    _transactionHandler->removeTransaction(trxId);
    return applyRes;
  }
  auto commit = DocumentLogEntry{std::string(shardId),
                                 OperationType::kCommit,
                                 {},
                                 trxId};
  return _transactionHandler->applyEntry(commit);
}
```

mpoeter (approved):

PR #17386, opened 2022-10-19
ErrorT

markuspf:
### Scope & Purpose

As per a suggestion of @goedderz in #17371 I tried drafting an `ErrorT` which is a `std::variant` that either contains an error code or a result value.

The intention is to enable functions that can either return a value or error, and communicate the outcome in a return value. I hope this will make for some more ergonomic code in a few places (as for instance in #17371


goedderz (commented):
Thanks, I like it!

lib/Basics/ErrorT.h
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  template<typename... Args>
+  [[nodiscard]] auto static error(Args... args) -> ErrorT<Error, T> {
+    return ErrorT(
+        Contained(std::in_place_index<0>, std::forward<Args>(args)...));
+  }
+
+  template<typename... Args>
+  [[nodiscard]] auto static ok(Args... args) -> ErrorT<Error, T> {
```
mpoeter, line 56:
```suggestion
  [[nodiscard]] auto static ok(Args&&... args) -> ErrorT<Error, T> {
```

markuspf:
I will merge this into #17371 and then run jenkins on it.

tests/Errors/main.cpp
```diff
@@ -0,0 +1,31 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+#include "gtest/gtest.h"
+
+int main(int argc, char* argv[]) {
+  ::testing::InitGoogleTest(&argc, argv);
+  GTEST_FLAG_SET(death_test_style, "threadsafe");
+
+  return RUN_ALL_TESTS();
+}
```
mpoeter, line 31: Is this file really necessary? Couldn't we use the one from the gtest repo instead?

PR #17418, opened 2022-10-21
[PREGEL] Make each state own its specific worker api

jvolmer:
This PR exchanges the WorkerApi that is used inside the conductor states to send messages to the workers with a new WorkerApi:
Now every state has its own worker api (before, the Conductor owned the worker api). This new api sends a specific type of message to the workers and collects messages of a specific type that were send back from the workers. This makes the states much more readable (they don't need any guarded or aggregate datastructure, this is all done inside the new api). The worker api becomes more generic: It does not have message-type specific functions but is generic over the message type of the expected response from the workers.
Additionally, this PR makes the storing state simpler: There is just one message - Store - send to the workers (instead of Store and  Cleanup). The cleanup happens now directly inside the worker after it finishes storing.


romanatarango (approved):
lgtm

PR #17421, opened 2022-10-21
Solve excessive memory consumption in multi-iterators - BTS-1088

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/17408

Queries with a FILTER statement like this:

  FILTER d.attr IN @list

with a very large list consume excessively large amounts of memory
during execution on dbservers.

The reason is that the query is internally translated into a
"multi-iterator" consisting of as many RocksDBIndexIterator objects
as the list has entries. Each of these subiterators keeps a RocksDB
iterator which uses a lot of memory which is unaccounted for.

The fix is to free used sub-iterators as soon as they are exhausted.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] This is the devel backport.

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/ES-1318


neunhoef (approved):
LGTM

PR #17426, opened 2022-10-24
[BTS-209] Added interpretation of payload to not treat everyone as plain text

cpjulia:
### Scope & Purpose

PR for https://arangodb.atlassian.net/browse/BTS-209. 
Fixed requests to `_admin/execute` treating every payload as plain text when they're in JSON or velocypack format, but will only treat the payload as velocypack if specified in the header's `content-type`. Otherwise, if the content-type is unset, the body will be treated as JSON and, therefore, if not according to JSON format, will raise an error (which is purposefully made). 
Behavior before:
every request body would be treated as plain text and parsed as raw payload, so, if plain text, it would work, in any other case, would raise an error. 
Behavior now:
if the request body has content-type `text/plain` or is unset, it's treated as raw payload (because the former behavior for no content-type was to be treated as such), otherwise, if JSON, it's treated as JSON and, if velocypack, it's treated as such. 
Then, to send requests to `_admin/execute`, either the content-type must be set to the body format or it will be treated as plain/text. 
Obs.: there was this separate if in the code that executes the `_admin/execute` request:
```
 if (_request->header(StaticStrings::ContentTypeHeader) == "text/plain" ||
      _request->header(StaticStrings::ContentTypeHeader)
          .starts_with("text/plain;") ||  // content-type can contain another
                                          // parameter like charset
      _request->contentType() == rest::ContentType::UNSET) {
    payload = _request->rawPayload();
  } 
```
because of this other observation below in HttpRequest class (changing it would cause a collateral effect to all other requests, hence, it was done only for `_admin/execute`.
```
else if (_contentType == ContentType::UNSET &&
             key == StaticStrings::ContentTypeHeader) {
    auto res = rest::stringToContentType(value, /*default*/ ContentType::UNSET);
    // simon: the "@arangodb/requests" module by default the "text/plain"
    // content-types for JSON in most tests. As soon as someone fixes all the
    // tests we can enable these again.
    if (res == ContentType::JSON || res == ContentType::VPACK ||
        res == ContentType::DUMP) {
      _contentType = res;
      return;
    }
```
so, in HttpRequest, if it's `text/plain`, it doesn't set the content-type to it.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-209
- [ ] Design document:


tests/js/client/shell/shell-vpack.js
```diff
@@ -158,6 +161,54 @@ function echoVpackVpack () {
...
+  };
+
+  var obj = "require(\"console\").log(\"abc\");";
+  var body = V8_TO_VPACK(obj);
+
+  var res = arango.POST_RAW(path, body, headers);
+  assertEqual(res.code, 200);
+  assertFalse(res.error);
+  assertNull(res.parsedBody);
+};
```
jsteemann, line 184:
```suggestion
}
```

tests/js/client/shell/shell-vpack.js
```diff
@@ -158,6 +161,54 @@ function echoVpackVpack () {
...
+  };
+
+  var obj = "return \"abc\"";
+  var body = V8_TO_VPACK(obj);
+
+  var res = arango.POST_RAW(path, body, headers);
+  assertEqual(res.code, 200);
+  assertFalse(res.error);
+  assertEqual(res.parsedBody, "abc");
+};
```
jsteemann, line 199:
```suggestion
}
```

tests/js/client/shell/shell-vpack.js
```diff
@@ -158,6 +161,54 @@ function echoVpackVpack () {
...
+function adminExecuteNoHeaderVpack() {
+  var path = '/_admin/execute';
+
+  var obj = "return \"abc\"";
+  var body = V8_TO_VPACK(obj);
+
+  var res = arango.POST_RAW(path, body);
+  assertEqual(res.code, 400);
+  assertTrue(res.error);
+};
```
jsteemann, line 210:
```suggestion
}
```

arangod/RestHandler/RestAdminExecuteHandler.cpp
```diff
@@ -61,9 +61,16 @@ RestStatus RestAdminExecuteHandler::execute() {
 
   TRI_ASSERT(server().getFeature<V8DealerFeature>().allowAdminExecute());
 
-  std::string_view bodyStr = _request->rawPayload();
-  char const* body = bodyStr.data();
-  size_t bodySize = bodyStr.size();
+  std::string_view payload;
+  size_t bodySize;
+  if (_request->header(StaticStrings::ContentTypeHeader) == "text/plain") {
```
jsteemann, line 66: I don't think this is sufficient. Though it will work if the content-type is exactly `text/plain`, it won't work if the content-type is `text/plain` plus any charset (see https://www.w3.org/International/articles/http-charset/index.en). This should also be supported and work.

arangod/RestHandler/RestAdminExecuteHandler.cpp
```diff
@@ -61,9 +61,16 @@ RestStatus RestAdminExecuteHandler::execute() {
...
+    payload = _request->rawPayload();
+  } else {
+    VPackSlice payloadParsed = _request->payload(false);
```
jsteemann, line 69: In case the content-type is not exactly `text/plain`, we will now always call the payload method, which will throw an error if the request is not JSON nor velocypack. What if there is no content-type set at all? Will this also fail now? IMO this case previously worked, but may now fail.

We are also assuming that what ever gets sent in the body parses to a JSON or velocypack string. If some other data type is sent, then the following `payloadParsed.stringView()` call will throw an exception. Is that intentionally not handled here? Does an appropriate error message come back to the caller in this case?

jsteemann (approved):
LGTM

PR #17427, opened 2022-10-24
[CINFRA-589] Fix race when committing trx in replication2

mpoeter:
### Scope & Purpose

With replication 2.0, an ArangoDB transaction is split into multiple RocksDB transactions (one per shard), which are committed individually and potentially concurrently. Previously, the  `RocksDBBaseTrxMethods` called back into the (shared) TransactionState to update to collections, which was wrong for several reasons: first, because the state is shared, concurrent calls that modify the state are obviously racy, and second, the state updated _all_ collections, including those that were not yet committed.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] This is covered by our current **integration tests**. A TSan run should have reported that race.


jsteemann (approved):
LGTM!

apetenchea (approved):

PR #17436, opened 2022-10-24
[DRAFT, WIP] Update S2 and abseil-cpp libraries

hkernbach:
### Scope & Purpose

DRAFT PR - ignore for now. Cleanup needed. State right now just for testing. 
This PR will update versions of s2 library and abseil-cpp. 

- [x] :pizza: 3rd party library update

### Checklist


MBkkt:
Don't forget about patches, and remove part patches for s2, now them support C++20.

hkernbach:
> Don't forget about patches, and remove part patches for s2, now them support C++20.

@MBkkt Yes, thank you :) I might just drop that PR here, I had some local compiler (upgraded to clang 15.0.3) issues and just wanted to try if newer versions of both libraries solve those issues I had. But that is not the case.

PR #17444, opened 2022-10-25
Stop dropped databases as well

goedderz:
### Scope & Purpose

During shutdown, only open databases were stopped, while dropped databases were left running. This lead to problems (in a WIP replication 2 branch), because the databases and its still running jobs were only stopped during `unprepare`, at which point the Scheduler already stopped.

- [X] :hankey: Bugfix

### DIFF
--- a/arangod/RestServer/DatabaseFeature.cpp
+++ b/arangod/RestServer/DatabaseFeature.cpp
@@ -693,13 +693,11 @@ void DatabaseFeature::stop() {
   }
 #endif
 
-  for (auto& p : theLists->_databases) {
-    TRI_vocbase_t* vocbase = p.second;
-
+  auto stopVocbase = [](TRI_vocbase_t* vocbase) {
     // iterate over all databases
     TRI_ASSERT(vocbase != nullptr);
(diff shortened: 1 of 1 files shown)


maierlars (approved):

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -693,13 +693,11 @@ void DatabaseFeature::stop() {
...
 
-  for (auto& p : theLists->_databases) {
-    TRI_vocbase_t* vocbase = p.second;
-
+  auto stopVocbase = [](TRI_vocbase_t* vocbase) {
     // iterate over all databases
     TRI_ASSERT(vocbase != nullptr);
     if (vocbase->type() != TRI_VOCBASE_TYPE_NORMAL) {
-      continue;
+      return;
```
MBkkt, line 700: why?

maierlars: The loop body was put into a lambda. Hence a continue becomes a return. The lambda is used twice below.

MBkkt: True, for some reason, I thought you put cycle to lamda

PR #17457, opened 2022-10-27
[CMAKE] Activate ctest

markuspf:
### Scope & Purpose

Just call `enable_testing()` in the main CMakeLists.txt, and add a token test (`arangodbtests_zkd`) as a first demo.

In the medium term, the monolithic `arangodbtests` binary should be split up into more manageable pieces which can be run with `ctest`;

For some binaries it could be considered to enable the GTest integration for CTest; this is absolutely not an option right now: the GTest integration generates a ctest-case for every google test and then executes the `arangodbtests` binary *for every signle one` of them. This is really slow.


jvolmer (approved):
LGTM

PR #17459, opened 2022-10-27
Add an explicit compact API call for logs.

maierlars:
### Scope & Purpose
This PR adds an explicit `compact` API. When this api is called, all participants compact as many log entries as possible. Compaction might be blocked by the state machine. 

If called on a coordinator, the request is forwarded to all participants.


apetenchea (approved):
LGMT

PR #17462, opened 2022-10-27
FORTIFY_SOURCE always with GCC and clang (when not Debug)

markuspf:
### Scope & Purpose

Enable `_FORTIFY_SOURCE=2` when compiling with GCC or clang.

This happens for all configurations except `Debug`, as _FORTIFY_SOURCE requires optimisation.

This needs a simple-performance-test benchmark to see whether `_FORTIFY_SOURCE=2` has any negative performance impact.


CMakeLists.txt
```diff
@@ -787,7 +787,7 @@ option(USE_MAINTAINER_MODE
 
 if (USE_MAINTAINER_MODE)
   add_definitions("-DARANGODB_ENABLE_MAINTAINER_MODE=1")
-  if (CMAKE_COMPILER_IS_GNUCC AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+  if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
```
jsteemann, line 790: Question: is `_FORTIFY_SOURCE` understood by MSVC as well?
If guess that even if it's not, setting the define there will not do any harm. Anyone knows?

markuspf: I would hope that jenkins tells me. alas.

markuspf: I just realised that my comment was kinda stupid; of course we can just set some `#define` and if noone picks them up they'll not do harm, really.

CI is not the correct place to blame in this case (only in the sense that setting `_FORTIFY_SOURCE` won't break anything).

So I looked around, and as far as I can tell, MSVC does not support `_FORTIFY_SOURCE`.

markuspf:
Except for what looks like a spurious failure, the CI tests pass.

I was not yet successful in getting a simple performance test to run on this branch.

PR #17463, opened 2022-10-27
[PREGEL] In conductor states: put incoming messages in queue

jvolmer:
This PR puts all messages that a the conductor state receives into a new message queue. The run function in the state processes each message in the queue synchronously, aggregates these messages and changes into a new state if all expected messages arrived (in the `WorkerApi::aggregateAllResponses` function). This way, we don't need any mutex inside the state (before we used a `Guarded<Aggregate<T>>` inside the WorkerApi that is part of each state). The receive function in the state is now very simple and only adds the message to the queue and cannot change the state any more.
The PR also changes the Computing state a bit: Now for each gss step a new Computation state is created.

### DIFF
--- a/arangod/Pregel/Conductor/States/State.h
+++ b/arangod/Pregel/Conductor/States/State.h
@@ -45,10 +45,7 @@ namespace conductor {
 
 struct State {
   virtual auto run() -> std::optional<std::unique_ptr<State>> = 0;
-  virtual auto receive(MessagePayload message)
-      -> std::optional<std::unique_ptr<State>> {
-    return std::nullopt;
-  }
+  virtual auto receive(MessagePayload message) -> void {}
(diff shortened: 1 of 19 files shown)


arangod/Pregel/Conductor/States/State.h
```diff
@@ -45,10 +45,7 @@ namespace conductor {
 
 struct State {
   virtual auto run() -> std::optional<std::unique_ptr<State>> = 0;
-  virtual auto receive(MessagePayload message)
-      -> std::optional<std::unique_ptr<State>> {
-    return std::nullopt;
-  }
+  virtual auto receive(MessagePayload message) -> void {}
```
markuspf, line 48: I was wondering whether it is reasonable to default-implement `receive` as empty? Shouldn´t this be virtual?

markuspf (approved):
LGTM, bar a small question.

PR #17464, opened 2022-10-27
[PREGEL] Cleanup

jvolmer:
This PR does some cleanup: It
-  cleans up the status update call in the worker,
- deletes the unused AggregateCount struct.

### DIFF
--- a/arangod/Pregel/Worker/Worker.cpp
+++ b/arangod/Pregel/Worker/Worker.cpp
@@ -76,11 +76,8 @@ using namespace arangodb::pregel;
 template<typename V, typename E, typename M>
 std::function<void()> Worker<V, E, M>::_makeStatusCallback() {
   return [self = shared_from_this(), this] {
-    auto statusUpdatedEvent =
-        StatusUpdated{.senderId = ServerState::instance()->getId(),
-                      .status = _observeStatus()};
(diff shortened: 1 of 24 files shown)


arangod/Pregel/Worker/Worker.cpp
```diff
@@ -76,11 +76,8 @@ using namespace arangodb::pregel;
 template<typename V, typename E, typename M>
 std::function<void()> Worker<V, E, M>::_makeStatusCallback() {
   return [self = shared_from_this(), this] {
-    auto statusUpdatedEvent =
-        StatusUpdated{.senderId = ServerState::instance()->getId(),
-                      .status = _observeStatus()};
-    _callConductor(ModernMessage{.executionNumber = _config.executionNumber(),
-                                 .payload = statusUpdatedEvent});
+    send(StatusUpdated{.senderId = ServerState::instance()->getId(),
```
markuspf, line 79: random thought: we could template the `send` function on the message type, i.e.
```
template<typename Msg, typename Args...>
send(Args&&... args) {
}
```

then this call would become `send<StatusUpdated>(ServerState::instance()->getId(), _observeStatus());`

Not sure whether this is at all better though: your code makes the struct names explicit which might be nice as well.

markuspf (approved):
So nice seeing this code  evolve into more readable code! :smile:

PR #17472, opened 2022-10-28
Repair "load indexes into memory" functions for web UI and SmartGraph edge collections

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/17478
Docs PR: https://github.com/arangodb/docs/pull/1166

Repair "Load indexes into memory" functionality in web UI, and make it work for SmartGraph edge collections, so that it actually starts the cache population.
The cache warmup is done in the background. The function will return as soon as all warmup tasks have been scheduled.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17477
  - [ ] Backport for 3.8: -

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/1166
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM. Let's collect a bit of experience with this. In the case of a collection with many shards (per dbserver), this might lead to a lot of blocked threads. Potentially, we must throttle this eventually for these cases. In most normal cases we have only very few (or only one) shard per dbserver and then everything should be fine.

jsteemann:
> LGTM. Let's collect a bit of experience with this. In the case of a collection with many shards (per dbserver), this might lead to a lot of blocked threads. Potentially, we must throttle this eventually for these cases. In most normal cases we have only very few (or only one) shard per dbserver and then everything should be fine.

Yes, this also worries me a bit.
We were previously using that `LocalTaskQueue` to dispatch the loading operations for the shards, and we additionally blocked one thread per shard that scheduled that dispatched everything and then waited for the dispatched ops to be finished.
Now it is at least better, and we don't wait for any operation to finish. Plus, the tasks for all shards etc. are completely independent, and can be scheduled independent of each other.
Still, all the tasks together can cause significant CPU and I/O load, and we currently don't have a reasonable way to limit that load somehow.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -71,17 +77,42 @@ constexpr bool EdgeIndexFillBlockCache = false;
...
       _upper(upper.data(), upper.size()) {}
 
-void RocksDBEdgeIndexWarmupTask::run() {
-  _index->warmupInternal(_trx, _lower, _upper);
+Result RocksDBEdgeIndexWarmupTask::run() {
+  DatabaseGuard databaseGuard(_databaseFeature, _dbName);
+  CollectionGuard collectionGuard(&databaseGuard.database(), _collectionName);
+
+  auto ctx = transaction::V8Context::CreateWhenRequired(
+      databaseGuard.database(), false);
```
mpoeter, line 95: Why do we use a V8Context here?

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -648,31 +679,49 @@ static std::string FindMedian(rocksdb::Iterator* it, std::string const& start,
...
+Result RocksDBEdgeIndex::scheduleWarmup() {
   if (!useCache()) {
-    return;
+    return {};
   }
 
-  // prepare transaction for parallel read access
-  RocksDBTransactionState::toState(trx)->prepareForParallelReads();
+  auto ctx =
+      transaction::V8Context::CreateWhenRequired(_collection.vocbase(), false);
```
mpoeter, line 688: Why do we use a V8Context here?

mpoeter (approved):
LGTM - not sure why we use a V8Context here, but that is what we did before as well, so...

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: I think using a V8 context is not necessary anymore. The original code used a V8 context, because the `load indexes into memory` operation could have been called via the `db.<collection>.loadIndexesIntoMemory()` function, i.e. from within V8.
As we are now always executing the loading ops in a background thread, we could change this to use a standalone context. 
I suggest we make this change in devel first to see if it has any side-effects.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: I think using a V8 context is not necessary anymore. The original code used a V8 context, because the `load indexes into memory` operation could have been called via the `db.<collection>.loadIndexesIntoMemory()` function, i.e. from within V8.
As we are now always executing the loading ops in a background thread, we could change this to use a standalone context. 
I suggest we make this change in devel first to see if it has any side-effects.

PR #17234, opened 2022-09-28
Extract arangod libraries to subdirectories

jvolmer:
This PR extracts all cmake library declarations in the arangod directory into the responsible subdirectories.
Check the commits to see all steps that have been done.

PR #17484, opened 2022-11-02
AQL: Do not lex `NOT instance` or similar expressions as `NOT IN` operator plus something

### LINKED ISSUE #17367: BUG : FILTER fails when using negation (!) on variable whose name start with "in"
## My Environment

* __ArangoDB Version__:  3.10.0      
* __Deployment Mode__: Single Server      
* __Deployment Strategy__: ArangoDB Official Docker with Docker-compose  
* __Configuration__: [Low memory recommandations from the doc](https://www.arangodb.com/docs/stable/tutorials-reduce-memory-footprint.html)           
* __Infrastructure__: OVH VPS              
* __Operating System__: Debian 10       
* __Total RAM in your machine__: 8Gb      
* __Disks in use__: HDD(?)       
* __Used Package__: Docker - official             

## Component, Query & Data

__Affected feature__: AQL query

__AQL query (if applicable)__:
```
FOR id IN ["Users/abc123"]

    FOR instance, e IN 1..1 OUTBOUND id IS_IN_INS
      FILTER !instance.HIDE
        RETURN instance
```
```
FOR id IN ["Users/abc123"]

    FOR instance, e IN 1..1 OUTBOUND id IS_IN_INS
      FILTER instance.HIDE
        RETURN instance
```
```
FOR id IN ["Users/abc123"]

    FOR istance, e IN 1..1 OUTBOUND id IS_IN_INS
      FILTER !istance.HIDE
        RETURN istance
```
__AQL explain and/or profile (if applicable)__:

```
Query String (147 chars, cacheable: true):
 FOR id IN ["Users/abc123"]
     FOR v, info IN 1..1 OUTBOUND id IS_IN_INS
       FILTER info.HIDE
         RETURN info
 

Execution plan:
 Id   NodeType            Est.   Comment
  1   SingletonNode          1   * ROOT
  2   CalculationNode        1     - LET #3 = [ "Users/abc123" ]   /* json expression */   /* const assignment */
  3   EnumerateListNode      1     - FOR id IN #3   /* list iteration */
  4   TraversalNode          5       - FOR /* vertex optimized away */, info  /* edge */ IN 1..1  /* min..maxPathDepth */ OUTBOUND id /* startnode */  IS_IN_INS
  5   CalculationNode        5         - LET #7 = info.`HIDE`   /* attribute expression */
  6   FilterNode             5         - FILTER #7
  7   ReturnNode             5         - RETURN info

Indexes used:
 By   Name   Type   Collection   Unique   Sparse   Cache   Selectivity   Fields        Stored values   Ranges
  4   edge   edge   IS_IN_INS    false    false    false       17.22 %   [ `_from` ]   [  ]            base OUTBOUND

Traversals on graphs:
 Id  Depth  Vertex collections  Edge collections  Options                                  Filter / Prune Conditions
 4   1..1                       IS_IN_INS         uniqueVertices: none, uniqueEdges: path  FILTER info.`HIDE`       

Optimization rules applied:
 Id   RuleName
  1   optimize-traversals

41 rule(s) executed, 1 plan(s) created
```

__Dataset__:

A Users collection linked to a Instances with a IS_IN_INS edge

Each User can be linked to multiple Instances and an Instance usually contains several Users

A User has the ability to HIDE an Instance he participates in and some UI will allow him to list the Instances he has while HIDE-ing the one he wants.

## Steps to reproduce

1. create the Users, Instances and IS_IN_INS collections
2. populate the collection with dummy datas linking some Users to some Instances
3. try to execute the first query

__Problem__:
The first query give 
  `Query: AQL: syntax error, unexpected not in operator near '!instance.HIDE RETURN in...' at position 4:14 (while parsing)`

Both the second and third queries work fine (although the second one obviously give back inverted results)

__Expected result__:
We should be able to use FILTER with NOT (!) operand even when our variables names starts with "in" or the doc should clearly states that it is prohibited.

__Note__:
The case was raised following an upgrade from 3.6.1 to 3.10.0 and I am sure the exact same query was fully functional in 3.6.1 so I suggest looking for optimization made in this time-frame.

Simran-B:
### Scope & Purpose

Fixes #17367

Changes in grammar.hpp/cpp are caused by the moved CMake code for flex/bison (#17234)

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.10: #17489

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/tokens.ll
```diff
@@ -214,7 +214,7 @@ class Parser;
   return T_NONE;
 }
 
-(?i:AT[ \t\r\n]+LEAST) {
+(?i:AT[ \t\r\n]+LEAST)/[ \t\r\n] {
```
Simran-B, line 217: I can't think of an AQL construct that would allow you to write e.g. `at leastblarg` and be valid somehow, so this one is probably not necessary, but maybe still good for futureproofing.

arangod/Aql/tokens.ll
```diff
@@ -126,7 +126,7 @@ class Parser;
   return T_DESC;
 }
 
-(?i:NOT[ \t\r\n]+IN) {
+(?i:NOT[ \t\r\n]+IN)/[ \t\r\n] {
```
Simran-B, line 129: `/...` is Flex' version of `(?=...)` it seems and is called [trailing context](https://www.cs.virginia.edu/~cr4bd/flex-manual/Patterns.html#Patterns).

`(?i:NOT[ \t\r\n]+IN[ \t\r\n])` should equally work but consumes the trailing whitespace - which shouldn't make a difference because whitespace is ignored anyway. However, I wonder if consuming `\n` in tokens throws off the line/column count perhaps - I'm referring to this:
```
[\n] {
  yycolumn = 0;
}
```

jsteemann: Will not work for `x NOT IN(`....

jsteemann: Nor `x NOT IN[1,2,3]`.

Simran-B: Good point, should instead check whether there is as non-variable name character

arangod/Aql/tokens.ll
Simran-B: Removed because it doesn't fix anything concrete

jsteemann (approved):
Seems to work nicely. Also checked that the line numbers are still accounted for correctly.

Simran-B:
Jenkins blue except Linux aarch because no build machine was available. Re-run: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-linux.aarch64/3962/

MBkkt (approved):

PR #17491, opened 2022-11-02
move StaticStrings into a library of its own

jsteemann:
### Scope & Purpose

Right now this doesn't provide any benefits, but it will allow linking things to arango_static_strings in the future without having to link to everything else in `lib`.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -52,6 +52,10 @@ else()
   message(STATUS "Will not compile in hand-optimized assembler code for CRC32.")
 endif()
 
+add_library(arango_static_strings STATIC
+  Basics/StaticStrings.cpp
+)
+
 add_library(arango STATIC
   ${LIB_ARANGO_APPLE}
   ${LIB_ARANGO_MSVC}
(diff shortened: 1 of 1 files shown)


MBkkt (changes requested):
I think static strings bad practice.
We should remove this kludge and use directory and namespace like StaticString instead of class with 
inline constexpr constant = "";

Also namespace better because allow to create separate files for different constant

jsteemann:
> I think static strings bad practice. We should remove this kludge and use directory and namespace like StaticString with inline constexpr constant = ""; instead of class
> 
> Also namespace better because allow to create separate files for different constant

I agree to that using constexpr would be better than using something that is prone to init-order fiascos etc., but that was not the scope of this change.
Re namespaces: yes, why not, but many of these strings here are actually used from a lot of places, including the client-tools (arangosh, arangodump etc.), so we will not be able to move them out of `lib` easily.

markuspf (approved):
I like this as a way to separate the lib directory into sligthly better units.

as @MBkkt points out, we should move to using constexpr string_views for string constants as well. 

These two moves are not mutually exclusive, though.

MBkkt (approved):

mchacki (approved):
LGTM!

PR #17514, opened 2022-11-03
test must not assert during test setup to get proper results!

dothebart:
### Scope & Purpose

test would run code outside of regular test plans, and hence produce broken results. Move code to `setUpAll`.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #17518, opened 2022-11-04
[CINFRA] Ping Replicated Log

maierlars:
### Scope & Purpose
This PR adds a new `ping` function on a replicated log. This function inserts an meta log entry and is designed for debugging. It does not interfere with the state machine based on the replicated log. The javascript function accepts an optional string parameter `message` that is put into the ping log entry. The rest handler optionally accepts an object with key `message`.


apetenchea (approved):
LGTM

PR #17534, opened 2022-11-07
Move CMake definitions of some libraries inside the lib-directory into their own directories

jvolmer:
This PR
- Moves all arango_geo cmake stuff into lib/Geo directory
- Moves all arango_v8 stuff into lib/V8. This also includes moving the file V8SecurityFeature from lib/ApplicationFeature directory into lib/V8 directory - as a source it is only used inside the arango_v8 library and therefore belongs with the other V8 stuff.
- Moves VelocyPackStringLiteral into the new lib/VelocypackUtils directory and use it as a source for new velocypack_utils library and delete its usage in arango library.


markuspf:
Love it.

What about `Basics/VPackStringBufferAdpter` `Basics/VelocyPackHelper`?

jvolmer:
As discussed, `Basics/VPackStringBufferAdpter` and `Basics/VelocyPackHelper` are more complicated to extract into the velocypack_utils library because they use a lot of stuff we don't want to link against in velocypack_utils , e.g. Exceptions. Let's do this in another PR.

markuspf:
I think we will reach the point soon at which we should decide to look at the whole underbelly of code that includes errors and exceptions and find out where to pull in a boundary.

markuspf (approved):

PR #17556, opened 2022-11-10
Actor for Pregel with tests

markuspf:
### Scope & Purpose

Small step towards making tests run.


arangod/Pregel/Actor/MPSCQueue.h
```diff
@@ -0,0 +1,96 @@
...
+  struct Node {
+    virtual ~Node() = default;
+    std::atomic<Node*> next{nullptr};
+  };
+
+  MPSCQueue() : stub{}, head(&stub), tail(&stub) {}
+
+  auto push_internal(Node *value) -> void {
+    auto prev = head.exchange(value);
+    prev->next.store(value);
```
maierlars, line 47: What if between line 46 and 47, `prev` was popped and released? Or even worse, it was pop, released, reallocated and pushed somewhere else. Now you write to some random memory address. I think this is the classical ABA problem and memory reclamation problem. I have no idea, so maybe ask @mpoeter .

markuspf: This should not happen (according to my current understanding)

Since you're inserting at the head, the prev's old next pointer would be a nullptr (before you store something in it);

The pop function then would not pop directly, but add a pointer to stub exactly to avoid this situation.

mpoeter: Correct - the node referenced by head can never be released/reclaimed as long as the next pointer is null. That is the beauty of this queue, because it does not need a memory reclamation scheme.

arangod/Pregel/Actor/Actor.h
```diff
@@ -0,0 +1,83 @@
...
+struct Actor {
+  Actor(Scheduler& schedule, State initialState)
+      : schedule(schedule), state(initialState) {}
+
+  Actor(Scheduler& schedule, State initialState, std::size_t batchSize)
+      : batchSize(batchSize), schedule(schedule), state(initialState) {}
+
+  void process(std::unique_ptr<Message> msg) {
+    inbox.push(std::move(msg));
+    kick();
```
MBkkt, line 44: You can use mpsc stack to understand need kick or not.

Check async mutex or strand implemention: https://github.com/YACLib/YACLib/blob/main/src/exe/strand.cpp#L29

MBkkt (changes requested):

arangod/Pregel/Actor/MPSCQueue.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include <memory>
+#include <atomic>
+
+// This lock-free multi-producer-single-consumer-queue is inspired by the
+// code published at
+//
+// https://www.1024cores.net/home/lock-free-algorithms/queues/intrusive-mpsc-node-based-queue
+//
+// under a Apache-2.0 license.
```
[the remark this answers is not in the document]
mpoeter: I am not sure I understand your argument. The Vyukov MPSC queue is pretty simple already. Yes, a stack with a reverse might be simpler, but has more contention in push (as you mentioned) and linear overhead for the reverse. This MPSC queue is not linearizable, but that doesn't matter in our case. The way it is used it is not just lock-free, but even wait-free.  It would be quite simple to extend this algorithm for batching if we need it.

With n SPSC queue you not longer have a strict FIFO order.

arangod/Pregel/Actor/Actor.h
mpoeter: I don't understand! When we push a message, we always need to call `kick` to schedule a task to ensure that it will be processed. Not sure what an async mutex or strand has to do with it. Also, why do you think `kick` requires a global mutex?

MBkkt: Because kick push task to executor, which commonly shared between threads.

Not you don't need kick every push you have cycle in work, while you're in cycle you don't need to call kick.
Question: how to synchronize kick with end of cycle? One of answers: strand code

MBkkt:
```auto* expected = _jobs.load(std::memory_order_relaxed);
  do {
    job.next = expected == Mark() ? nullptr : expected;
  } while (!_jobs.compare_exchange_weak(expected, &job));
  if (expected == Mark()) {
   kick();
  }

```

MBkkt: Also current code contains bug, what if I make two pushes, from different threads:

Them doing something like

Thread1:
Add task1
kick
pop task1
thinks what's all
Thread2:
push task2
kick
thinks other thread execute it (cas on busy)
Thread1
store false to the busy
end

arangod/Pregel/Actor/MPSCQueue.h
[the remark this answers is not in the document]
MBkkt: You have allocation for task, and kick (which commonly have internal synchronization and probably allocation in your case) on every push.
So more contention bad argument.

Linear overhead sounds strange, because in general it single additional operation for every task.

About simple, I don't think you right, you known with other more difficult algorithms, but in general pop not simple in Vyukov queue.

And talking about wait-free incorrect, wait-free about guarantees, not about speed.

Anyway, if we talk about this case, it have (incorrect) synchronization on every kick (if (cas)) and probably additional synchronization in push to executor.

MBkkt (commented):
It's pseudocode, but it explains why don't need kick on every proccess

```cpp
struct Actor {
  void process(Msg* msg) {
    auto* expected = head.load();
    do { 
       msg->next = expected == Mark() ? nullptr : expected; 
    } while (!_jobs.cas(expected, msg));
    if (expected == Mark()) { 
      kick(); 
    }
  }

  void kick() {
    // Make sure that *someone* works here
    schedule([this]() { this->work(); });
  }

  void work() {
    if (!tail) {
      tail = head.exchange(0);
      reverse(tail);
    }
    auto i = 0;
    do {
      state = std::visit(MessageHandler{std::move(state)}, *tail);
      tail = tail->next;
    } while (++i < batchSize && tail);
    if (!tail && head.load() == nullptr && head.cas(nullptr, Mark()) {
      return;
    }
    kick();
  }
};
```

arangod/Pregel/Actor/MPSCQueue.h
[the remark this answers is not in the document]
mpoeter: You always have the allocation for the task, but since the queue itself is intrusive there is not additional allocation there. The scheduler itself is lock-free. ATM it uses the boost lock-free queue which handles its own internal freelist. So yes, pushing tyo the scheduler may require an allocation, but that is negligible. FWIW - I was thinking about using a different queue in the scheduler instead which does not require node allocations (since we only store pointers in the queue).

The reverse is linear in the number of entries in the queue. Granted, I also don't see this as a problem, but it is a difference in complexity to the current algorithm. Yes, the pop is more complex than with an MPSC stack, but not overly so.

I am well aware that lock-freedom/wait-freedom is about progress guarantees and not performance. I never mentioned anything about performance. That said, the current implementation is wait-free population oblivious - the push consists of only two instructions - it doesn't get much faster than this (at least on x86/x64; on ARM the exchange has to be simulated with a LL/SC retry loop).

arangod/Pregel/Actor/Actor.h
mpoeter: True, we do not need to schedule a new task if one is already scheduled or being executed. Yes, this matches the execution model of a strand, but we do not have something like that in our code right now, and it is easy enough to simulate. I haven't checked the code yet, so it is possible there are still some bugs.
FWIW - this is just a poor mans implementation of an actor model by using our current scheduler (which is not really designed for this). The main goal is to gather experience, so this is not meant to be tuned for performance.

MBkkt: Your message for me doesn't matches with implementation Vyukov queue.

arangod/Pregel/Actor/MPSCQueue.h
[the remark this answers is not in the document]
MBkkt: > You always have the allocation for the task (and you will have that one with every other data structure as well).

It's not required, it can be on stack, or on another allocated memory.

> but since the queue itself is intrusive there is no additional allocation there

Agree, but I don't talk about it.

> The scheduler itself is lock-free. ATM it uses the boost lock-free queue which handles its own internal freelist. So yes, pushing tyo the scheduler may require an allocation, but that is negligible. FWIW - I was thinking about using a different queue in the scheduler instead which does not require node allocations (since we only store pointers in the queue).

It depends about what scheduler do we use, in arangodb thread pool it's not very effective:
1. Always make allocation for functor: https://github.com/arangodb/arangodb/blob/devel/arangod/Scheduler/Scheduler.h#L175 doesn't have some free list node or something like that
2. Then it makes push to very slow boost lockfree queue, which use free list node etc
3. Then a lot of atomics loads/writes
4. Then sometime try to wakeup waiting thread (lock mutex etc)

> FWIW - I was thinking about using a different queue in the scheduler instead which does not require node allocations (since we only store pointers in the queue).

It's can be simple intrusive forward list under mutex, it already be a lot of better.
Of course it can be even better with scheduler like in golang/rust tokio for example. But this schedulers doesn't have different priority.
So I think better to start with few simple intrusive forward lists.

> That said, the current implementation is wait-free population oblivious - the push consists of only two instructions - it doesn't get much faster than this (at least on x86/x64; on ARM the exchange has to be simulated with a LL/SC retry loop).

No, from practical side yes, but formally no, Vyukov mentioned it:
> Push function is blocking wrt consumer. I.e. if producer blocked in (*), then consumer is blocked too. Fortunately 'window of inconsistency' is extremely small - producer must be blocked exactly in (*). Actually it's disadvantage only as compared with totally lockfree algorithm. It's still much better lockbased algorithm.

> on ARM the exchange has to be simulated with a LL/SC retry loop

Btw not true for arm starts from v8.1, if I remember correctly

I agree with Vyukov queue better for mpsc scenario in vacuum, because smaller contention in push, but currently I don't see the reason to write more difficult and less effective (in our scenario) code.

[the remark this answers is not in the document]
mpoeter: > It's not required, it can be on stack, or on another allocated memory.
No, it cannot be on the stack if we just want to push the message, schedule the handler and go in with some other work.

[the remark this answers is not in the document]
MBkkt: It can if we know actor in which we push message handle it before end of scope.

But yes for actors, it's should be pretty rare

markuspf:
Well, this PR seems to have escalated a little bit.

For context, I had some quite clear (and fairly simple) goals with the code I wrote (rather am writing) in this PR;

 * Have a type that I can send messages to that will be executed in a sequential way. It was first called `MessagableT`, and its now called `Actor`
 * Not use mutexes/locks, as I did not want to get stuck in the ArangoDB scheduler detail hell.
 * Have tests that *compile and execute* in <1 minute from a fresh checkout, and ideally on github.
 * It be transparent enough that someone can understand the details of the code, but not need to understand all the details of ArangoDB.
 * Make all that somehow usable in the context of our Pregel code.
 
Following a suggestion by @mpoeter, I implemented the MPSC queue that is in this PR, and I intend to stay with that queue right now. It is a good learning and teaching experience.

I don't want to assume anything about what the ArangoDB scheduler does. The only thing I want to do is be able to say "schedule this thing, please" to make the Actor go.

Note also, that this PR goes into `staging/pregel-refactor`, not into `devel`

MBkkt (commented):
I don't agree with it, but ok

PR #17562, opened 2022-11-10
Fix SEARCH-411: double commit [2/2]

MBkkt:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -67,15 +68,15 @@ namespace {
 
 class IResearchFlushSubscription final : public FlushSubscription {
  public:
-  explicit IResearchFlushSubscription(TRI_voc_tick_t tick = 0) noexcept
+  explicit IResearchFlushSubscription(uint64_t tick = 0) noexcept
```
Dronplane, line 71: Why `tick` type is changed?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -550,53 +589,152 @@ void ConsolidationTask::operator()() {
...
-      _avgCleanupTimeMs{nullptr},
-      _consolidationTimeNum{0},
-      _avgConsolidationTimeMs{nullptr},
-      _metricStats{nullptr} {
+      _id(iid) {
   // initialize transaction callback
-  _trxCallback = [this](transaction::Methods& trx,
-                        transaction::Status status) -> void {
+
+  _trxBeforeFinish = [this](transaction::Methods& trx) {
```
Dronplane, line 602: NIT
Technically this is "PreCommit"  as transaction may be finished with Abort an this callback would not be called in such case.
```suggestion
  _trxPreCommit = [this](transaction::Methods& trx) {
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -550,53 +589,152 @@ void ConsolidationTask::operator()() {
...
+      }
+    }
+    else {
+      ctx._ctx.AddToFlush();
+    }
+#else
+    ctx._ctx.AddToFlush();
+#endif
+  };
+  _trxAfterFinish = [this](transaction::Methods& trx,
```
Dronplane, line 669: NIT This is called on every status change. Not only on commit.
```suggestion
  _trxStatusChange = [this](transaction::Methods& trx,
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1516,6 +1709,37 @@ Result IResearchDataStore::remove(transaction::Methods& trx,
 
   return {};
 }
+
+bool IResearchDataStore::exists(IResearchDataStore::Snapshot const& snapshot,
+                                LocalDocumentId documentId, bool nested,
+                                TRI_voc_tick_t const* recoveryTick) const {
+  if (recoveryTick == nullptr) {
+    TRI_ASSERT(false) << "Call exists not in recovery";
```
Dronplane, line 1717: Why it is wrong to call this method not in the recovery?

Dronplane (changes requested):

arangod/IResearch/IResearchDataStore.cpp
MBkkt: It's uint64_t from rocksdb so I think it should be uint64_t, anyway it's just using

arangod/IResearch/IResearchDataStore.cpp
MBkkt: Probably nothing, but now we don't have such code

arangod/IResearch/IResearchDataStore.cpp
MBkkt: Hmm, yes, maybe I need to change it

arangod/IResearch/IResearchDataStore.cpp
MBkkt: In general probably name same as callback type is good, I agree

arangod/IResearch/IResearchDataStore.cpp
Dronplane: Then maybe remove this assert?  Or at least  please leave a comment, why this assert is here

MBkkt: Ok

arangod/IResearch/IResearchDataStore.cpp
Dronplane: That will confuse readers. As in base class we have  TRI_voc_tick_t . So I'd prefer to be consistent with typenames. Even if for now these are the same.

MBkkt: The main issue for me rocksdb tick not TRI_vocbase, so I think we should also replace to uint64_t

MBkkt: Internally in iresearch also uint64

Dronplane: IMO: On server level we operate the abstraction of engine ticks with type "TRI_voc_tick_t". That is not necessary RocksDB ticks (we may replace storage engine in future).  The only requirement is that underlying engine ticks should fit into this type. And all "FlushFeature" related stuff should operate `TRI_voc_tick_t`.

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -828,36 +963,79 @@ Result IResearchDataStore::commitUnsafeImpl(
...
+    std::move(stageTwoGuard).Cancel();
+    if (!commitTwo) {
+      LOG_TOPIC("21bda", TRACE, TOPIC)
+          << "Second commit for ArangoSearch index '" << id()
+          << "' is no changes tick " << lastTickBeforeCommitTwo;
+      TRI_ASSERT(_lastCommittedTickOne <= lastTickBeforeCommitTwo);
+      TRI_ASSERT(_lastCommittedTickTwo <= lastTickBeforeCommitTwo);
+      // no changes, can release the latest tick before commit two
+      _lastCommittedTickOne = lastTickBeforeCommitTwo;
+      _lastCommittedTickTwo = lastTickBeforeCommitTwo;
```
Dronplane, line 1031: Github is fooling me or failpoint for crash after stage two is missing? https://github.com/arangodb/arangodb/blob/a4eb4d658241b53f3ede34f983ba31f74609a00d/arangod/IResearch/IResearchLink.cpp#L1163

Dronplane (approved):
LGTM

arangod/IResearch/IResearchDataStore.cpp
MBkkt: True

Issue #1160, opened 2014-12-04
V8 update for 2.4?


CoDEmanX:
A V8 update is scheduled for 2.4 xmas release according to the roadmap. Which version will it be, and which harmony features will be available?

I'm particularly interested template strings, which allow for multi-line string literals in backticks without end-of-line escaping:

```
`a long
multi-line string`

"the current"\
"way to go"
```

It would allow for nicely formatted AQL query strings in JavaScript files.

CoDEmanX:
@pluma: filed report https://github.com/arangodb/aqbjs/issues/5

@jsteemann: thanks for checking. I asked some V8 guys and they told me template string are available in the yet unstable 3.31. According to them, an upgrade from 3.29 to 3.31 shouldn't be much pain - I hope you will update V8 in ArangoDB to 3.31+ as it becomes stable.

Rest parameters, the spread operator and destructuring assignments aren't coded as of now, although there's some work going on. It appears to be rather difficult whatsoever to implement these ES6 features and I wouldn't expect them to be available in V8 any time soon.

jsteemann:
@CoDEmanX : interesting, thanks for sharing the info.

PR #17577, opened 2022-11-14
[CMAKE] Make futures into a library with separate testing

markuspf:
### Scope & Purpose

This makes the `lib/Futures` directory into its own library, trying to decouple the ArangoDB Futures from the rest of ArangoDB.

To test ArangoDB futures, you should now be able to run

```
# cmake --preset community-developer
# cmake --build --preset community-developer --parallel 20 --target arangodbtests_futures
# ctest --preset community-developer -R futures 
```
(or alternatively run `build/community-developer/bin/arangodbtests_futures` at your own leisure)

Note that this PR depends on two other PRs, namely  #17616 and #17609.

This PR has grown an enterprise buddy  #1160


markuspf:
Sorry guys, I think there's still a bit of cleaning up to be done in this PR before it's ready for review.

lib/Futures/SharedState.h
```diff
@@ -210,6 +211,7 @@ class SharedState {
         [[fallthrough]];
 
       default:
+        break;
         TRI_ASSERT(false);  // unexpected state
```
jvolmer, line 215: Is the break on purpose before the TRI_ASSERT - then you can just delete the TRI_ASSERT I guess.

lib/Futures/SharedState.h
```diff
@@ -26,9 +26,8 @@
 #include <atomic>
 #include <function2.hpp>
 
-#include "Basics/debugging.h"
-#include "Futures/Try.h"
-#include "Logger/LogMacros.h"
+#include "Assertions/Assert.h"
```
jvolmer, line 29: With this include, would you not need to link arango_assertions into arango_futures?

lib/Futures/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+add_library(arango_futures STATIC
+  Future.cpp)
+
+target_include_directories(arango_futures PRIVATE
+  ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
+  PUBLIC
+  ${PROJECT_SOURCE_DIR}/lib
```
jvolmer, line 7: Do we really need this? In this PR you already changed all "Futures/...." includes into "...". If you need more, perhaps it needs to be linked? (see my comment below about the assertions)

lib/Futures/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+  if (MSVC)
+    target_compile_options(arango_futures PRIVATE /WX /D_WINSOCK_DEPRECATED_NO_WARNINGS96)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") # clang-cl.exe
+      # There's a warning about /MP being unused if compiling only a single file
+      target_compile_options(arango_futures PRIVATE -Wno-unused-command-line-argument)
+    endif()
+  else ()
+    target_compile_options(arango_futures PRIVATE -Werror -Wno-error=deprecated-declarations)
+  endif ()
+endif ()
```
jsteemann, line 24: Can be removed once we merge https://github.com/arangodb/arangodb/pull/17611.

jsteemann (commented):
LGTM except Julia's comment and the one comment I made

lib/Futures/SharedState.h
```diff
@@ -173,6 +173,7 @@ class SharedState {
         }
         [[fallthrough]];
       default:
+        break;
         TRI_ASSERT(false);  // unexpected state
```
maierlars, line 177:
```suggestion
        TRI_ASSERT(false);  // unexpected state
        break;
```

maierlars (approved):

lib/Futures/SharedState.h
markuspf: yup

lib/Futures/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+target_include_directories(arango_futures
+  PUBLIC
+  ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
+  .)
+
+target_link_libraries(arango_futures arango_assertions)
+
+if (MSVC)
+  target_compile_options(arango_futures PUBLIC "-J")
```
MBkkt, line 12: Needs comment why it's necessary.
Also `/J` not `-J`

lib/Futures/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_library(arango_futures STATIC
+  Future.cpp)
+
+target_include_directories(arango_futures
+  PUBLIC
+  ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
+  .)
```
MBkkt, line 7:
```suggestion
  )
```

lib/Futures/Future.cpp
```diff
@@ -21,7 +21,7 @@
 /// @author Simon Grätzer
 ////////////////////////////////////////////////////////////////////////////////
 
-#include "Futures/Future.h"
+#include "Future.h"
```
MBkkt, line 24: revert this please

lib/Futures/Future.h
```diff
@@ -28,10 +28,10 @@
...
 #include <thread>
 
-#include "Futures/Exceptions.h"
-#include "Futures/Promise.h"
-#include "Futures/SharedState.h"
-#include "Futures/Unit.h"
+#include "Exceptions.h"
+#include "Promise.h"
+#include "SharedState.h"
+#include "Unit.h"
```
MBkkt, line 34: same here

lib/Futures/Promise.h
```diff
@@ -23,9 +23,9 @@
 
 #pragma once
 
-#include "Futures/Exceptions.h"
-#include "Futures/SharedState.h"
-#include "Futures/Unit.h"
+#include "Exceptions.h"
+#include "SharedState.h"
+#include "Unit.h"
```
MBkkt, line 28: and here

lib/Futures/SharedState.h
```diff
@@ -26,9 +26,9 @@
 #include <atomic>
 #include <function2.hpp>
 
-#include "Basics/debugging.h"
-#include "Futures/Try.h"
-#include "Logger/LogMacros.h"
+#include "Assertions/Assert.h"
+
+#include "Try.h"
```
MBkkt, line 31: and here

lib/Futures/Utilities.h
```diff
@@ -29,8 +29,8 @@
 #include <mutex>
 #include <vector>
 
-#include "Futures/Future.h"
-#include "Futures/Unit.h"
+#include "Future.h"
+#include "Unit.h"
```
MBkkt, line 33: and here

MBkkt (changes requested):
Everything looks good, except unnecessary changes "include paths" (and it makes it worse)

lib/Futures/Future.cpp
maierlars: has it ever occurred to you, that it might be completely obvious to you, but it isn't to others? Why do you want him to revert? It is not very respectful to just dump whatever you want. We are not your code monkeys. Please be better!

MBkkt: I think if it obvious for Markus, he write to me why.
Sorry if it sounds rude. But for me, without explanation in PR description it looks like changes for changes

markuspf: The file `Future.h` is in the same directory as `Future.cpp` and can hence be included as such without a path. 

We can have a separate discussion about whether it is desirable for a library to not put its headers into a dedicated subdirectory; 
The library is declared in `CMakeLists.txt` to have just its own path as include directory.

That would mean having a subdirectory `lib/Futures/Futures` under `lib/Futures`; I found that unnecessarily verbose.

This becomes a problem once there is `lib/Lib1/Utils.h` and `lib/Lib2/Utils.h` and no way to disambiguate; so yes, sure, we can add the header directory.

MBkkt: > The file Future.h is in the same directory as Future.cpp and can hence be included as such without a path.

Relative includes more slow on compilation stage.

But you don't use it, you use absolute include because you add it to include directories in cmake

> The library is declared in CMakeLists.txt to have just its own path as include directory.

But why? Maybe it a little safer from one side, but we don't develop library, it's just small part of arangodb, in separate directory, I don't understand the reason to do it.

> That would mean having a subdirectory lib/Futures/Futures under lib/Futures; I found that unnecessarily verbose.
This becomes a problem once there is lib/Lib1/Utils.h and lib/Lib2/Utils.h and no way to disambiguate; so yes, sure, we can add the header directory.

Yes, so you can just use include paths from "lib" not "lib/Futures" (as we do before for all other libs) and don't add additional stuff to cmake for compiler.
Also it protects us from situation that you describe

MBkkt: I think overall your PR is good changes. And will be good to merge it, but maybe you can make separate PR with "include" changes, where we can discuss pros and cons?
As you write:

> We can have a separate discussion about whether it is desirable for a library to not put its headers into a dedicated subdirectory;

jsteemann (approved):

MBkkt (approved):

PR #17581, opened 2022-11-14
experimental: automatically refill index cache entries

jsteemann:
### Scope & Purpose

Add experimental new option `refillIndexCaches` to 

* AQL INSERT/UPDATE/REPLACE modification operations
* single-document insert, update and replace operations
* multi-document insert, update and replace operations

If the option is set for an operation, a currently running transaction will keep track of which edge cache index entries were invalidated by the transaction, and will try to (re-)populate them later.
Example usages:
* `db.<collection>.insert({ _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
* `db.<collection>.update(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
* `db.<collection>.replace(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
* `INSERT { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
* `UPDATE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
* `REPLACE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`

The repopulating of the in-memory edge caches is performed by a background thread, so the modification operations shouldn't be slowed down a lot. The background thread may however cause additional I/O for looking up the data in RocksDB and repopulating the caches.
The background repopulating is done in a best-effort way and is not guaranteed to always succeed, e.g. if there is no memory available memory for the cache subsystem, or when an in-memory cache table is currently in a migration phase (grow/shrink operation).
Currently only edge indexes are supported.

For testing, there is an experimental startup option `--rocksdb.refill-index-caches` for DB servers, which currently defaults to `false`. If it is set to true, the cache repopulation will be turned on automatically for all modification operations, so that it doesn't need to be specified on the per-operation/per-query level.

Please mind the following caveats:
* the option `refillIndexCaches` is currently not forwarded to followers, because it is unclear if this is needed. Forwarding the option would be trivial to add later though.
* when (re)starting an instance, it currently doesn't automatically populate its in-memory caches for edge indexes. This still has to be done manually.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/ExecutionPlan.cpp
+++ b/arangod/Aql/ExecutionPlan.cpp
@@ -955,6 +955,8 @@ ModificationOptions ExecutionPlan::parseModificationOptions(
           options.validate = !value->isTrue();
         } else if (name == StaticStrings::KeepNullString) {
           options.keepNull = value->isTrue();
+        } else if (name == StaticStrings::RefillIndexCachesString) {
+          options.refillIndexCaches = value->isTrue();
         } else if (name == StaticStrings::MergeObjectsString) {
           options.mergeObjects = value->isTrue();
         } else if (name == StaticStrings::Overwrite) {
--- a/arangod/Aql/ModificationExecutorHelpers.cpp
+++ b/arangod/Aql/ModificationExecutorHelpers.cpp
@@ -216,6 +216,7 @@ OperationOptions ModificationExecutorHelpers::convertOptions(
   // in.exclusive;
   out.overwriteMode = in.overwriteMode;
   out.ignoreRevs = in.ignoreRevs;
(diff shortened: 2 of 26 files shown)


jsteemann:
Obsolete

PR #17602, opened 2022-11-17
Documentation-3.9/arangosearch-cache-v3.9.5

Simran-B:
Add the `cache` option to the `storedValues` DocuBlock of arangosearch Views.

Follow-up to arangodb/docs#1175 and related to arangodb/docs#1190


Simran-B:
Merged into https://github.com/arangodb/arangodb/pull/17603 because it's too late for inclusion in v3.9.5

PR #17608, opened 2022-11-17
[CMAKE] Some CMakeLists.txt cleanups

markuspf:
### Scope & Purpose

 * Removes commented out lines
 * Removes Debian/Darwin/RPM packaging


hkernbach (approved):
LGTM

CMakeLists.txt
```diff
@@ -1232,20 +1122,6 @@ endif ()
...
-# V8_LIBS
-# V8_INCLUDE_DIR
-#
-# ICU_VERSION
-# ICU_LIBS
-# ICU_INCLUDE_DIR
-#
-# ZLIB_VERSION
-# ZLIB_LIBS
-# ZLIB_INCLUDE_DIR
```
jsteemann, line 1247: If this comment is accurate (don't know if it is), then I would vote for leaving it there.
This is because the stuff in 3rdParty populates these variables in the parent scope, and it is otherwise not immediately visible that this happens.

jsteemann (commented):
Can't comment on the packaging sections in this PR.
Everything else looks good to me, except the one comment I made.

CMakeLists.txt
```diff
@@ -1466,14 +1342,6 @@ add_custom_target(clean_autogenerated_files
 
 message(STATUS "building for git revision: ${ARANGODB_BUILD_REPOSITORY}")
 
-# if (VERBOSE)
```
neunhoef, line 1469: This looks like debugging code which might be useful one day.

neunhoef (approved):
LGTM. Cannot comment about packaging either.

CMakeLists.txt
markuspf: There are CMake GUI tools with which one can look at the variables set in the current environment; there is a curses variant (ccmake) and a QT variant with mouse interface and all.

I believe that would be much more helpful for debugging.

CMakeLists.txt
markuspf: These are only partially accurate; the ccmake tool can give better insight into what's defined and whats not;

I will look into the necessity of exporting values to the parent scope at all and hopefully find beter ways to deal wit hthis.

PR #17609, opened 2022-11-17
[LIB] Split assertions out of lib/Basics/debugging.h

markuspf:
To resolve the dependency of every bit of code on heavy core functionality, split assertion code out of debugging.h and provide TRI_ASSERT and ADB_PROD_ASSERT separately from include files in three modes

 * Light eassertions: these do not depend on anything more than the C++ stdlib and call std::terminate() to end the process.
 * Maintainer mode assetions: these use ArangoDBs CrashHandler to crash the program if an assertion fails in the case of TRI_ASSERT and ADB_PROD_ASSERT
 * Production mode assertions: these do not crash the arangod process for TRI_ASSERT, but do crash for ADB_PROD_ASSERT


lib/Assertions/Assert.h
```diff
@@ -0,0 +1,72 @@
...
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+#pragma once
+
+// Ensures that the macro TRI_ASSERT is defined. There are three possible modes
+//
+// * USE_ARANGODB_LIGHT_ASSERTIONS -- a light version of assertions that does not
+//   require ArangoDB's CrashHandler
+//   The motivation for "light mode" is to resduce dependencies on ArangoDB code
```
jsteemann, line 29:
```suggestion
//   The motivation for "light mode" is to reduce dependencies on ArangoDB code
```

lib/Assertions/Assert.h
```diff
@@ -0,0 +1,72 @@
...
+//   for C++ unittests.
+// * MAINTAINER_MODE -- in maintainer mode, when the condition in TRI_ASSERT is
+//   false, the program crashes using ArangoDB's CrashHandler
+// * Production mode: Assertions are compiled away. Except that the expression
+//   in the assertion is actually evaluated, still, but the process does not
```
jsteemann, line 34: "actually evaluted": this should be clarified, because the expression evaluation code is generated during compile time, but cannot be reached at runtime. Very likely it is optimized away by the compiler completely.
At runtime, nothing of the expression evaluation code should ever be executed, as the `expr` code is only executed if the expression `false` is `true`:
``` 
#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/          \
  (true) ? ((false) ? (void)(expr) : (void)nullptr)   \
         : ::arangodb::debug::AssertionNoOpLogger{} & \
               ::arangodb::debug::NoOpStream {}
```

lib/Assertions/AssertionLightLogger.h
```diff
@@ -0,0 +1,44 @@
...
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include "Basics/CrashHandler.h"
```
jsteemann, line 26: This include seems to be superfluous.
And according to the comments in https://github.com/arangodb/arangodb/pull/17609/files#diff-a3f3d97412e29d24227a8dd685fb6022bc17036472c9151c603e3895433546feR27-R28, including the CrashHandler here defeats the purpose of the PR.

jsteemann: Should rather include
```
#include <exception> // std::terminate
#include <ostream>  // std::flush
#include <sstream>  // std::ostringstream

File: lib/Assertions/ProdAssert.h [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+// Ensures that the macro ADB_PROD_ASSERT is defined. There are three possible modes
+//
+// * USE_ARANGODB_LIGHT_ASSERTIONS -- a light version of assertions that does not
+//   require ArangoDB's CrashHandler
+//   The motivation for "light mode" is to resduce dependencies on ArangoDB code
```
jsteemann, line 30:
```suggestion
//   The motivation for "light mode" is to reduce dependencies on ArangoDB code
```

maierlars (approved):
See Jans comment. Otherwise, LGTM

markuspf:
Sorry for jumping the gun on this one; It turned out that light mode was not just superflous, but actually did not really work.

One would have had to compile every bit of code in two modes. I added a "light" crash handler instead in #17616 Now one can compile every library and link the light crash handler for testing.

At some point we can get rid of the "light" crash handler again: Once I understood the dependencies from the CrashHandler through the logger downwards, we can likely agree on a base library of arangodb stuff that compiles very quickly and gives valuable information on crashes.

lib/Assertions/AssertionLogger.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace arangodb::debug {
+struct AssertionLogger {
+  [[noreturn]] void operator&(std::ostringstream const& stream) const {
+    std::string message = stream.str();
+    arangodb::CrashHandler::assertionFailure(
+        file, line, function, expr,
+        message.empty() ? nullptr : message.c_str());
+  }
+  // can be removed in C++20 because of LWG 1203
```
jsteemann, line 38: As we are using C++20 already, can we remove this now?

jsteemann (approved):
LGTM

lib/Assertions/AssertionLogger.h
markuspf: I tried and it didn't work; will look again, maybe with @maierlars ' help

maierlars: Well, I think compilers are not there yet.

PR #17611, opened 2022-11-17
fix USE_FAIL_ON_WARNINGS CMake option

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1147

Fix `USE_FAIL_ON_WARNINGS` CMake option.

Previously it only failed if there were errors in some libraries. but several other libraries were not affected by the settings and compiled successfully even despite compile warnings being present.
This PR will change compiler settings so that builds will fail when `USE_FAIL_ON_WARNINGS` is set and there is a compile warning in any of our own libraries or executables.
Tests are currently still excluded, but can be added as well if the approach works in general.
Third-party libraries also should be ignored.

Nothing should change if the CMake option `USE_FAIL_ON_WARNINGS` is set to off.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1147
- [ ] Design document:


markuspf:
Thanks for having a look at this!

I wonder whether it would be less intrusive and verbose to just add `add_compile_options(-Werror)` to the CMakeLists.txt in `arangod`, `lib`, `tests` and `client-tools`?

CMakeLists.txt
```diff
@@ -975,6 +975,20 @@ function(add_cxx_flags_if_supported var)
   set(${var} "${${var}}" PARENT_SCOPE)
 endfunction()
 
+function(add_compile_warnings_flags lib)
+  if (USE_FAIL_ON_WARNINGS)
+    if (MSVC)
+      target_compile_options(arango PRIVATE /WX /D_WINSOCK_DEPRECATED_NO_WARNINGS)
```
Dronplane, line 981: Should here also be `lib` ?
```suggestion
      target_compile_options(${lib} PRIVATE /WX /D_WINSOCK_DEPRECATED_NO_WARNINGS)
```

jsteemann: Yes!

maierlars (approved):
I was under the impression that `add_compile_options` sets the flags for all targets. Why do we need to call `add_compile_warnings_flags´ twice?

gnusi (approved):

arangod/IResearch/CMakeLists.txt
```diff
@@ -92,13 +92,13 @@ target_link_libraries(arango_iresearch
   arango_cluster_engine
   arango_indexes
   arango_vocbase
-  boost_locale # required by iResearch
+  boost_locale # required by IResearch
```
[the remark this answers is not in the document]
jsteemann: Have removed it now, and it compiled and linked fine locally.

PR #17616, opened 2022-11-18
Make CrashHandler its own thing

markuspf:
### Scope & Purpose

This PR separates CrashHandler out of the `lib/Basics` directory and introduces a "Light" crash handler with the same interface as the "Heavy" one in ArangoDB. The result is two libraries: `arango_crashhandler` and `arango_crashhandler_light`.

The "Light" crash handler comes with its own (really lightweight) tests, and just terminates if anything tries to call a function that is not implemented.

The main purpose of this exercise is to make dependencies and interfaces clearer and to enable being able to test ArangoDB code that uses `TRI_ASSERT` (which uses the CrashHandler to crash) without compiling "all of ArangoDB", by having tests only link against `arango_crashhandler_light`.

It is perfectly possible that at some point we can remove the light crashhandler again.

I will rebase the work in #17609 on this because this causes less verbosity and overhead (and it will actually work)


jsteemann (approved):
LGTM

PR #17630, opened 2022-11-22
[CINFRA-432] Fixing recoverEntries on leader

apetenchea:
### Scope & Purpose

Since we started supporting the compareExchange operation, we introduced the `_ongoingStates` variable into the core. This is a deque to which we always append states containing the latest changes to the prototype. For example, when a key is inserted, it is immediately applied to the `_store`, which is then appended to `_ongoingStates`. This is now considered to be the latest state, but only after the key is committed, we can begin using this state for reading. Hence, when a log entry is committed, we start popping from `_ongoingStates` until we reach the correct state where the modifications coming from that log entry become visible. This is done in the `PrototypeCore::update` method.
There are a few problems with the way `_ongoingStates` are being used:
1. These are now shared by leader and follower, although a follower doesn't need them for anything. If a follower becomes leader, we should reset `_ongoingStates`, since these are now obsolete. We shouldn't read any data from them!
2. When a leader is doing recovery, the current code behaves as if the `_ongoingStates` which are already present in the core are valid. This is not the case, as they might contain uncommitted entries. During recovery the leader should discard any ongoing states and start fresh!

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


maierlars (approved):
LGTM

PR #17632, opened 2022-11-22
[testing]if after a restore attempt the server isn't healthy anymore properly abort the test so we don't clean up

dothebart:
### Scope & Purpose

we would recently see servers fail during asan dump_with_crashes runs. 
However, logfiles would be cleared, hence no post mortem analysis of the incident would be possible.
This PR will check the server health and turn the status to bad if the server is gone. 

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

KVS85 (approved):

KVS85:
Unrelated failure in Jenkins.

PR #17635, opened 2022-11-22
[PREG-180] Split Pregel V8 methods into their own file

markuspf:
### Scope & Purpose

This moves the V8 methods that access Pregel functionality into a file of  their own.


jvolmer (approved):
LGTM

PR #17638, opened 2022-11-22
[PREG-173] Extract collection source stuff into specified classes

jvolmer:
This PR moves stuff that sets the collection sources in the conductor into separate classes: Mainly GraphSource does all the work. 
I added an interface for collections which is implemented by ClusterCollection and SingleServerCollection. The main difference is that a ClusterCollection can be made out of several Collections (happening if the collection is smart). With this interface we can handle collections in single server and cluster the same way.
Additionally, this PR combines several Pregel configuration variables into a struct PregelOptions.


arangod/Pregel/PregelOptions.h
```diff
@@ -0,0 +1,70 @@
...
+  std::vector<std::string> edgeCollections;
+};
+
+struct GraphName {
+  std::string graph;
+};
+
+using VertexCollectionID = CollectionID;
+using EdgeCollectionID = CollectionID;
+struct EdgeCollectionRestrictions {
```
markuspf, line 48: I wonder whether `EdgeCollectionRestrictions` should be a part of `GraphDataSource`?

arangod/Pregel/tests/InspectorTest/main.cpp
```diff
@@ -0,0 +1,98 @@
...
+            std::vector<std::string>{"some_collection_name"});
+  ASSERT_EQ(alternative.edgeCollections,
+            (std::vector<std::string>{"some_collection_name",
+                                      "another_collection_name"}));
+}
+
+auto main(int argc, char** argv) -> int {
+  testing::InitGoogleTest();
+  return RUN_ALL_TESTS();
+}
```
markuspf, line 98: You could use `gtest_main` now and lose this bit of code.

arangod/Pregel/tests/InspectorTest/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_executable(pregel_inspector main.cpp ${PROJECT_SOURCE_DIR}/lib/Basics/VelocyPackStringLiteral.cpp)
+
+target_include_directories(pregel_inspector PRIVATE ${PROJECT_SOURCE_DIR}/lib ${PROJECT_SOURCE_DIR}/arangod)
+
+target_link_libraries(pregel_inspector gtest fmt velocypack)
```
markuspf, line 5:
```suggestion
target_link_libraries(pregel_inspector gtest_main fmt velocypack)
```

arangod/Pregel/Utils.cpp
```diff
@@ -71,7 +71,7 @@ ErrorCode Utils::resolveShard(ClusterInfo& ci, WorkerConfig const* config,
   std::shared_ptr<LogicalCollection> info;
   auto const& it = planIDMap.find(collectionName);
   if (it != planIDMap.end()) {
-    info = ci.getCollectionNT(config->database(), it->second);  // might throw
+    info = ci.getCollectionNT(config->database(), it->second);
```
markuspf, line 74: We should make a TODO card to (re)move `resolveShard` from Pregel. This should have no business being here ;)

markuspf (changes requested):
I think this looks good overall.

I do wonder though whether the whole set of `Collections` deserve a comment about what they do, because this still seems very complicated,

Are they meaningfully testable? Don't worry if they're not; that just means we will have to look at the data source design a bit harder!

arangod/Pregel/Utils.cpp
jvolmer: https://arangodb.atlassian.net/browse/PREG-206

arangod/Pregel/PregelOptions.h
jvolmer: Good point

jvolmer:
I added some descriptions to the collections, is that what you had in mind?
Testing is difficult by using a StorageEngineMock, very similar to LogicalDataSourceTest, but that does not seem to work for ClusterCollection: The test has problems with the ClusterFeature, probably the test is not even executed in a cluster environment, I'm not sure how this works at the moment.

arangod/Pregel/Conductor/GraphSource.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (!collection->hasAccessRights(auth::Level::RW)) {
+      return Result{TRI_ERROR_FORBIDDEN};
+    }
+    if (storeResults && !collection->hasAccessRights(auth::Level::RW)) {
+      return Result{TRI_ERROR_FORBIDDEN};
+    }
+  }
+
+  if (!ServerState::instance()->ROLE_SINGLE) {
```
jvolmer, line 59: I just noticed that the sharding needs only to be checked in cluster mode. I don't like that I have to check the role explicitly here, I would like the collection by itself to know what to do. Not sure if there is a bitter way, options I thought about:
- The isShardingCorrect function is very pregel-specific therefore I would not like to move it inside the collection interface
- Have a pregelCollection on top of collection that has a isShardingCorrect function, this seems to be too much effort, because then this pregelcollection has to implement all collection functions and you possibly also need a PregelCollection*s* struct as well.

markuspf (approved):

PR #17645, opened 2022-11-23
BTS-1129: [testing] work around test setup limiting maximum path length

dothebart:
### Scope & Purpose

unix domain socket paths in tests might get too long, add workaround for BTS-1129

- [x] :hankey: Bugfix


dothebart:
https://jenkins01.arangodb.biz/job/arangodb-ANY-mac-matrix.x86-64/3857/

dothebart:
we fixed the path lengths instead.

PR #17662, opened 2022-11-24
[PREG-182] Delete current customizable pregel

jvolmer:
### Scope & Purpose

To simplify the pregel refactoring, this PR deletes AIR (customizable pregel) inside Pregel, related Pregel code and the Greenspun lib which was only used inside Pregel AIR. The customizable pregel is currently not sufficiently tested and not really used. It will be added or rewritten after we finished the Pregel refactoring.


markuspf (changes requested):
Please add an entry to CHANGELOG about the removal of AIR/Greenspun.

We still have to find out what is there in terms of documentation to remove.

Otherwise this looks good to me.

markuspf (approved):

PR #17674, opened 2022-11-25
Delay AddFollower and FailedFollower jobs by a configurable amount - BTS-1111

neunhoef:
### Scope & Purpose

This is about https://arangodb.atlassian.net/browse/BTS-1111.

This is a backport from devel to 3.10 for
https://github.com/arangodb/arangodb/pull/17246.

We have often felt that an addFollower job (and a FailedFollower job) should not immediately be
executed when it seems to become necessary, since in case of a restart
of a server that server might be back very quickly, get in sync again,
and no harm is done. So far we never saw a possibility to implement this
easily, since the Supervision must not have a "memory". However, here is
an easy trick:

This PR does the following: AddFollower and FailedFollower jobs are *scheduled* right away
as before, they note their `timeCreated` time in the job description.
We now simply prevent the job from being started, if not at least x
seconds have passed since its creation. This gives us exactly the delay
we want. Setting this to 5 minutes should prevent too much havoc in case
of a simple restart and will allow the existing replicas to come back
and quickly get in sync. If this happens within 5 minutes, the
addFollower job notices that it is no longer necessary and will
terminate itself.

The only disadvantage is that in case of a real server loss, we schedule
new replicas a bit slower, increasing the time back to non-degraded
state.

Note that an `AddFollower` job which helps to increase the `replicationFactor` must
be executed right away, otherwise the user has to wait the configured delay until the change
takes action.

- [*] :hankey: Bugfix

### Checklist
- [x] Tests
  - [*] C++ **Unit tests**
- [*] Backports?

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1111


arangod/Agency/AddFollower.cpp
```diff
@@ -193,6 +202,22 @@ bool AddFollower::start(bool&) {
...
   }
 
+  // Check if configured delay has passed since the creation of the job,
+  // or we are in "urgent" mode:
+  if (!_notBefore.empty()) {
+    auto now = std::chrono::system_clock::now();
+    std::chrono::system_clock::time_point notBefore =
+        stringToTimepoint(_notBefore);
+    if (now < notBefore) {
+      LOG_TOPIC("1de2c", DEBUG, Logger::SUPERVISION)
```
maierlars, line 212: I just noticed: shouldn't this be `TRACE`?

arangod/Agency/FailedFollower.cpp
```diff
@@ -173,6 +178,18 @@ bool FailedFollower::start(bool& aborts) {
 
   // Exclude servers in failoverCandidates for some clone and those in Plan:
   auto shardsLikeMe = clones(_snapshot, _database, _collection, _shard);
+
+  // Check if configured delay has passed since the creation of the job:
+  auto now = std::chrono::system_clock::now();
+  if (now < _notBefore) {
+    LOG_TOPIC("1de2d", DEBUG, Logger::SUPERVISION)
```
maierlars, line 185: same as above.

maierlars (approved):

PR #17675, opened 2022-11-25
[CINFRA-618] Abort failed transaction on followers

apetenchea:
### Scope & Purpose

There are two scenarios in which the leader and the follower could get out of sync, due to a failing transaction.

1. We are on the leader. Suppose we want to insert a document. This implies an insert operation and a commit. We replicate the insert operation successfully and eventually try to commit it. This is done inside `ReplicatedRocksDBTransactionState::doCommit()`. The leader always replicates the commit operation first, waits for enough followers to confirm the commit, then commits it locally. It is possible that the leader fails to commit locally (e.g. due to `OUT_OF_DISC_SPACE`).  At this point, the follower and the leader are already out of sync: followers have applied the transaction locally and are already able to forget about it, while the leader has to abort and roll back. We've debated possible options, and the best solution for now is to crash the leader. This might be enough to alert humans in care of low disc space.
2. We are on the follower and receive a transaction that cannot be applied, perhaps due to low disk space. In this case, the leader and the rest of follower continue, while the problematic one remains stuck. The best we could for now is stop the process. This is handled in `DocumentFollowerState::applyEntries`.

This PR contains some minor log output improvements and tinkering with the `ErrorCode` class, which I found useful to keep around.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**


arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
```diff
@@ -134,7 +136,9 @@ futures::Future<Result> ReplicatedRocksDBTransactionState::doCommit() {
         for (auto& res : results) {
           auto result = res.get();
           if (result.fail()) {
-            return result;
+            // We finished replication, but failed to commit locally. There is
+            // no going back.
+            FATAL_ERROR_EXIT_CODE(result.errorNumber().value());
```
maierlars, line 141: I think we have different exit code which do not agree with error codes. In particular I would suggest to log a precise error statement es well.

maierlars (changes requested):

arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
apetenchea: > I think we have different exit code which do not agree with error codes.
I did not know that, thank you!

apetenchea:
> Could you describe in a little more detail what the scenario looks like, and in particular how it's handled after this change? I don't quite get it from looking at the diff.

Gladly, I updated the description.

goedderz:
> > Could you describe in a little more detail what the scenario looks like, and in particular how it's handled after this change? I don't quite get it from looking at the diff.
> 
> Gladly, I updated the description.

Perfect, thanks!

maierlars (approved):

PR #17691, opened 2022-11-29
QA-335: [testing] on windows arangod writes coredumps on its own. delete them after this test

dothebart:
### Scope & Purpose

this recovery test was recently allowed to be ran on windows. 
On linux the system is responsible to write coredumps, which is disabled by scripts. 
On windows arangods write minidumps on their own. Delete them so we don't get crash reports. 

Edit: really fixing its issues on Windows is complicated, and the added value isn't that high. So disalble them. 
- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

Issue #9702, opened 2019-08-13
Feature: avoid write-write conflict errors for single document updates


mpoeter:
When using RocksDB, two concurrent update operations on the same document can result in a write-write conflict (error-code 1200). This type of error often causes confusion among users (this is a frequently asked question in Slack) and puts additional burden on the developer - one has to be aware of this error and how to handle it. I know that it is possible to use exclusive mode, but this locks the whole collection, which is overkill in many cases.
However, I think it should be possible to avoid this error for almost all operations that update only a single document.

At the moment, a developer has two choices to handle this error: either propagate it, or perform a retry. IMHO propagating it is hardly ever the right choice - if we have performed an unconditional update (i.e., not specified a specific document revision), then I simply want this update to be applied - regardless of document's state and any other potential concurrent updates. Performing a retry causes additional overhead during development (has to be implemented) and runtime.

The main point is that the error is not deterministic - if the requests are processed with slightly different timings (such that they do not overlap), then the updates are simply applied without any error.

So what it comes down to: *If there are multiple concurrent updates for the same document, then they should be serialized in some way.*

I would like to trigger a discussion about if and how this could be improved.

graetzer:
impossible to do without adding the retry logic inside the database. Especially considering that we do not know the intention of the developer.

jsteemann:
I think to free developers from the burden of retrying or conflict handling, IMHO there are at least the following options:
* serializing writes to (specific) collections automatically, regardless of whether keys conflict. This would be an easy solution, both to implement on the server side and for developers to use, but would pessimize writes for which there are no conflicts.
* retrying writes on conflict automatically for transactions/operations that are flagged as such. This would require minimal changes in client applications, and allows fine-grained control of when to retry (and when to fail with conflicts).

Do you see any further options?

mpoeter:
Well, as a developer, if I send a request to update a document in the DB, I expect that update to be applied. If I want to update a *specific revision*, then I can set `ignoreRevs` to false, and if the specified revision is no longer valid, then reporting a conflict error is perfectly fine.
But when I don't specify a revision, I simply want the update to be applied *unconditionally*.

AFAIK the RocksDB engine acquires a write-lock on a to-be-updated document. If this fails (because another request is currently updating the document), you abort with said error. Wouldn't it be possible to simply wait for some (limited) time until we could acquire the lock, so we can perform the update?

dothebart:
[...]

PR #17693, opened 2022-11-29
Avoid write-write conflicts for single document operations

mpoeter:
### Scope & Purpose

Avoid write-write conflicts for single document operations performed via the document REST API (i.e., no AQL, but also no streaming transactions). This is achieved by locking the key of each document before performing the actual modification. This lock acquisition effectively serializes all operations on the same document. To avoid starvation, the lock acquisition is limited to one second. This lock timeout value is currently hardcoded but will be made configurable in the future. If the lock cannot be acquired within this time, the operation fails with a write-write conflict error as before.

Performing changes to a unique index entry also requires us to lock that index entry to ensure uniqueness. This lock acquisition is subject to the same lock timeout as locking the document key.

We are planning to generalize this for multi-document operations as well as AQL and streaming transactions in the future.

In case we cannot acquire the lock on the key of the document we want to insert/modify, the error message will be `Timeout waiting to lock key - in index primary of type primary over '_key'; conflicting key: <key>` where `<key>` corresponds to the key of the document we tried to modify. In addition, the error object will contain `_key`, `_id` and `_rev` fields. The `_key` and `_id` correspond to the document we tried to insert/modify, and `_rev` will correspond to the current revision of the document from the DB if available, and otherwise empty.

In case we cannot acquire the lock on a unique index entry, the error message will be `Timeout waiting to lock key - in index <indexName> of type persistent over '<fields>'; document key: <key>; indexed values: [<values>]` where `<indexName>` is the name of the index in which we tried to lock the entry, `<fields>` is the list of fields included in that index, `<key>` corresponds to the key of the document we tried to insert/modify, and `<values>` corresponds to the indexed values from our document. In addition, the error object will contain `_key`, `_id` and `_rev` fields. The `_key` and `_id` correspond to the document we tried to insert/modify, and `_rev` will correspond to the current revision of the document from the DB if available, and otherwise empty.

This addresses Github issue #9702 and APM-522

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :books: documentation written (release notes, API changes, ...)

#### Related Information

- [x] Jira ticket: https://arangodb.atlassian.net/browse/APM-522
- [x] Design document: https://github.com/arangodb/documents/blob/master/DesignDocuments/01_IDEAS/WriteWriteConflictPrevention.md
- [x] Release notes PR: https://github.com/arangodb/docs/pull/1217


arangod/RestHandler/RestVocbaseBaseHandler.cpp
```diff
@@ -553,18 +553,17 @@ void RestVocbaseBaseHandler::extractStringParameter(std::string const& name,
 
 std::unique_ptr<transaction::Methods> RestVocbaseBaseHandler::createTransaction(
     std::string const& collectionName, AccessMode::Type type,
-    OperationOptions const& opOptions) const {
+    OperationOptions const& opOptions, transaction::Options trxOpts) const {
```
jsteemann, line 556: Should the options be passed by const reference? They mostly consist of booleans, but it seems they also carry the PeerState, which we probably don't want to copy unnecessarily.
```suggestion
    OperationOptions const& opOptions, transaction::Options const& trxOpts) const {
```

arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
```diff
@@ -106,7 +109,20 @@ TRI_voc_tick_t RocksDBTrxBaseMethods::lastOperationTick() const noexcept {
...
+bool RocksDBTrxBaseMethods::ensureSnapshot() {
+  if (_rocksTransaction->GetSnapshot() == nullptr) {
+    _rocksTransaction->SetSnapshot();
+    _readOptions.snapshot = _rocksTransaction->GetSnapshot();
+    TRI_ASSERT(_readOptions.snapshot != nullptr);
+    // we at least are at this point
+    TRI_ASSERT(_db || _readOptions.snapshot);
+    _lastWrittenOperationTick = _readOptions.snapshot
+                                    ? _readOptions.snapshot->GetSequenceNumber()
+                                    : _db->GetLatestSequenceNumber();
```
jsteemann, line 121: As `_readOptions.snapshot` is guaranteed to be a non-nullptr here, the ternary operator doesn't seem necessary anymore.

arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
```diff
@@ -172,8 +188,9 @@ rocksdb::Status RocksDBTrxBaseMethods::GetForUpdate(
   TRI_ASSERT(cf != nullptr);
   TRI_ASSERT(_rocksTransaction);
   rocksdb::ReadOptions const& ro = _readOptions;
-  TRI_ASSERT(ro.snapshot != nullptr);
-  return _rocksTransaction->GetForUpdate(ro, cf, key, val);
+  TRI_ASSERT(ro.snapshot != nullptr || _state->options().delaySnapshot);
+  auto res = _rocksTransaction->GetForUpdate(ro, cf, key, val);
+  return res;
```
jsteemann, line 193: nit
```suggestion
  return _rocksTransaction->GetForUpdate(ro, cf, key, val);
```

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1056,44 +1056,48 @@ Result RocksDBCollection::lookupKey(
     transaction::Methods* trx, std::string_view key,
     std::pair<LocalDocumentId, RevisionId>& result,
     ReadOwnWrites readOwnWrites) const {
+  return doLookupKey(trx, key, result, readOwnWrites, false);
+}
+
+Result RocksDBCollection::lookupKeyForUpdate(
+    transaction::Methods* trx, std::string_view key,
+    std::pair<LocalDocumentId, RevisionId>& result) const {
+  return doLookupKey(trx, key, result, ReadOwnWrites::yes, true);
```
jsteemann, line 1065: The `ReadOwnWrites::yes` here may be overkill for single document operations. In these, I don't think it is necessary to observe our own writes (because it is a single document transaction). But I may be wrong.

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -40,7 +41,7 @@ class Manager;
 }  // namespace cache
 
 class LogicalCollection;
-class RocksDBPrimaryIndex;
+class ManagedDocumentResult;
```
jsteemann, line 44: I removed ManagedDocumentResult from our entire codebase a few weeks ago. We shouldn't bring it back.
```suggestion
```

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -148,7 +155,20 @@ class RocksDBCollection final : public RocksDBMetaCollection {
                         bool fillCache,
                         ReadOwnWrites readOwnWrites) const override;
 
+  // @brief return the primary index
+  // WARNING: Make sure that this instance
+  // is somehow protected. If it goes out of all scopes
+  // or it's indexes are freed the pointer returned will get invalidated.
```
jsteemann, line 161:
```suggestion
  // or its indexes are freed the pointer returned will get invalidated.
```

arangod/RocksDBEngine/RocksDBCollection.h
```diff
@@ -25,6 +25,7 @@
 
 #include "Statistics/ServerStatistics.h"
 #include "RocksDBEngine/RocksDBMetaCollection.h"
+#include "RocksDBEngine/RocksDBPrimaryIndex.h"
```
jsteemann, line 28: It seems that it is not necessary to include this file here. The forward declaration would have been sufficient.

arangod/RocksDBEngine/RocksDBPrimaryIndex.cpp
```diff
@@ -825,10 +764,11 @@ Result RocksDBPrimaryIndex::insert(transaction::Methods& trx,
   rocksdb::Status s =
       mthd->Put(_cf, key.ref(), value.string(), /*assume_tracked*/ true);
   if (!s.ok()) {
-    res.reset(rocksutils::convertStatus(s, rocksutils::index));
+    Result res = rocksutils::convertStatus(s, rocksutils::index);
     addErrorMsg(res, keySlice.copyString());
```
jsteemann, line 768: The string parameter of `addErrorMsg` is changed to `std::string_view` in this PR, so we can pass the key as a string_view here.
```suggestion
    addErrorMsg(res, keySlice.stringView());
```

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -1660,151 +1659,188 @@ Result RocksDBVPackIndex::insert(transaction::Methods& trx,
...
+      // not necessary during index creation, because nothing
+      // will be in the in-memory cache.
+      invalidateCacheEntry(::lookupValueFromSlice(key.string()));
+    }
+
+    rocksdb::Status s;
+    if (performChecks) {
+      // auto oldLockTimeout = mthds->SetLockTimeout(0);
+      s = mthds->GetForUpdate(_cf, key.string(), &existing);
+      // mthds->SetLockTimeout(oldLockTimeout);
```
jsteemann, line 1721: Are the commented out lines still needed here?

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -1660,151 +1659,188 @@ Result RocksDBVPackIndex::insert(transaction::Methods& trx,
...
+          [&](LocalDocumentId const&, VPackSlice doc) {
+            IndexOperationMode mode = options.indexOperationMode;
+            VPackSlice key = transaction::helpers::extractKeyFromDocument(doc);
+            if (mode == IndexOperationMode::internal) {
+              // in this error mode, we return the conflicting document's key
+              // inside the error message string (and nothing else)!
+              res = Result{res.errorNumber(), key.copyString()};
+            } else {
+              // normal mode: build a proper error message
+              addErrorMsg(res, key.copyString());
```
jsteemann, line 1753:
```suggestion
              addErrorMsg(res, key.stringView());
```

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -1660,151 +1659,188 @@ Result RocksDBVPackIndex::insert(transaction::Methods& trx,
...
-          state->findCollection(_collection.id()));
-      TRI_ASSERT(trxc != nullptr);
-      for (uint64_t hash : hashes) {
-        trxc->trackIndexInsert(id(), hash);
-      }
+    TRI_ASSERT(key.containsLocalDocumentId(documentId));
+    s = mthds->PutUntracked(_cf, key, value.string());
+    if (!s.ok()) {
+      Result res = rocksutils::convertStatus(s, rocksutils::index);
+      return addErrorMsg(res, doc.get(StaticStrings::KeyString).copyString());
```
jsteemann, line 1830:
```suggestion
      return addErrorMsg(res, doc.get(StaticStrings::KeyString).stringView());
```

tests/js/common/shell/shell-document.js
```diff
@@ -302,6 +302,18 @@ function CollectionDocumentSuite () {
       });
     },
 
+    testSaveInvalidDocumentKeyValueWithOverwriteMode : function () {
+      [ "", " ", "  ", " a", "a ", "/", "|", "#", "a/a", "\0", "\r", "\n", "\t", "\"", "[", "]", "{", "}", "\\" ].forEach(function (key) {
+        try {
+          collection.save({ _key: key }, { overwriteMode: "update" });
+          fail();
+        }
+        catch (err) {
```
jsteemann, line 311:
```suggestion
        } catch (err) {
```

tests/sepp/Options.h
```diff
@@ -76,15 +77,17 @@ auto inspect(Inspector& f, Setup& o) {
...
+    workloads::InsertDocuments::Options, workloads::IterateDocuments::Options>;
 namespace workloads {
 // this inspect function must be in namespace workloads for ADL to pick it up
 template<class Inspector>
 inline auto inspect(Inspector& f, WorkloadVariants& o) {
   namespace insp = arangodb::inspection;
   return f.variant(o).unqualified().alternatives(
+      insp::type<workloads::WriteWriteConflict::Options>(
+          "write-write-conflict"),
       insp::type<workloads::GetByPrimaryKey::Options>("getByPrimaryKey"),
```
jsteemann, line 91: Casing is incosistent here: `write-write-conflict` vs `getByPrimaryKey`. We should go with only convention here.

tests/sepp/Workloads/WriteWriteConflict.cpp
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+
+    ++_operations;
+
+    if (_operations % 512 == 0 && shouldStop()) {
+      break;
+    }
+  }
+  std::cout << ("Conflicts: " + std::to_string(_conflicts) + "\n");
```
jsteemann, line 145:
```suggestion
  std::cout << "Conflicts: " << _conflicts << std::endl;
```

tests/sepp/Workloads/WriteWriteConflict.h
```diff
@@ -0,0 +1,117 @@
+
```
jsteemann, line 1:
```suggestion
```

tests/sepp/Workloads/WriteWriteConflict.h
```diff
@@ -0,0 +1,117 @@
...
+    insert,
+    update,
+    replace,
+  };
+
+  struct Options;
+  struct ThreadOptions;
+  struct Thread;
+
+  WriteWriteConflict(Options const& options) : _options(options) {}
```
jsteemann, line 52:
```suggestion
  explicit WriteWriteConflict(Options const& options) : _options(options) {}
```

arangod/RestHandler/RestVocbaseBaseHandler.cpp
mpoeter: The options are modified internally, but since they default to a default constructed temp value we cannot use a non-const ref. We could take them by rvalue ref though.

arangod/RocksDBEngine/RocksDBCollection.cpp
mpoeter: That is true for single document operations, but since this is generic code we do not have the relevant information. Besides, for single document operations we have a specialized RocksDBMethods implementation, so it doesn't really matter performance wise.

arangod/RocksDBEngine/RocksDBCollection.h
mpoeter: Unfortunately it is necessary because I redefine `primaryIndex` with a covariant return type (`RocksDBPrimaryIndex` instead of `Index`), so a forward declaration is not sufficient because the compiler needs to know the definition to check that it is a subtype of `Index`.

Documentation/DocuBlocks/Rest/Documents/delete_remove_document.md
```diff
@@ -64,6 +64,13 @@ is returned if the document was removed successfully and
 is returned if the collection or the document was not found.
 The response body contains an error document in this case.
 
+@RESTRETURNCODE{409}
+is returned if we fail to lock the document key due to some
+concurrent operation that is operating on the same document.
+This also also referred to as a write-write conflict.
+The response body contains an error document in this case the
```
jsteemann, line 71:
```suggestion
The response body contains an error document in this case with the
```

Documentation/DocuBlocks/Rest/Documents/patch_update_document.md
```diff
@@ -145,8 +145,15 @@ an error document in this case.
 is returned if the collection or the document was not found.
 
 @RESTRETURNCODE{409}
-is returned if the update causes a unique constraint violation in 
-a secondary index.
+there are two possible reasons for this error:
+1) if the update causes a unique constraint violation in  a secondary
```
jsteemann, line 149:
```suggestion
1) if the update causes a unique constraint violation in a secondary
```

Documentation/DocuBlocks/Rest/Documents/patch_update_document.md
```diff
@@ -145,8 +145,15 @@ an error document in this case.
...
+index. The response body contains an error document in
+this case with the errorNum set to 1210 (`ERROR_ARANGO_UNIQUE_CONSTRAINT_VIOLATED`).
+2) if we fail to lock the document key or some unique index entry
+due to some concurrent operation that is operating on the same
+document. This also also referred to as a write-write conflict.
+The response body contains an error document in this case the
```
jsteemann, line 155:
```suggestion
The response body contains an error document in this case with the
```

Documentation/DocuBlocks/Rest/Documents/put_replace_document.md
```diff
@@ -124,8 +124,15 @@ an error document in this case.
 is returned if the collection or the document was not found.
 
 @RESTRETURNCODE{409}
-is returned if the replace causes a unique constraint violation in 
-a secondary index.
+there are two possible reasons for this error:
+1) if the replace causes a unique constraint violation in  a secondary
```
jsteemann, line 128:
```suggestion
1) if the replace causes a unique constraint violation in a secondary
```

Documentation/DocuBlocks/Rest/Documents/put_replace_document.md
```diff
@@ -124,8 +124,15 @@ an error document in this case.
...
+index. The response body contains an error document in
+this case with the errorNum set to 1210 (`ERROR_ARANGO_UNIQUE_CONSTRAINT_VIOLATED`).
+2) if we fail to lock the document key or some unique index entry
+due to some concurrent operation that is operating on the same
+document. This also also referred to as a write-write conflict.
+The response body contains an error document in this case the
```
jsteemann, line 134:
```suggestion
The response body contains an error document in this case with the
```

PR #17643, opened 2022-11-22
Automatically refill in-memory index cache entries after insert/update/replace/remove

jsteemann:
### Scope & Purpose

Automatically refill in-memory index cache entries after insert/update/replace/remove, if so desired.

* Added new per-operation option `refillIndexCaches` to write operations,
  namely:

  - AQL INSERT/UPDATE/REPLACE/REMOVE modification operations
  - single-document insert, update, replace and remove operations
  - multi-document insert, update, replace and remove operations

  If the option is set to `true` any such operation, every currently running 
  transaction will keep track of which in-memory index cache entries were 
  invalidated by the transaction, and will try to (re-)fill them later.
  Currently edge indexes and velocypack-based indexes (persistent, hash, 
  sklipist index) are supported. For velocypack-based indexes, the refilling
  will only happen if the index was set up with an in-memory cache (i.e. the
  `cacheEnabled` flag was set during index creation).

  Example usages:
  - `db.<collection>.insert({ _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `db.<collection>.update(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `db.<collection>.replace(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `db.<collection>.remove(key, { refillIndexCaches: true });`
  - `INSERT { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
  - `UPDATE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
  - `REPLACE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
  - `REMOVE { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`

  The refilling of the in-memory caches for indexes is performed by a 
  background thread, so that the foreground write operation shouldn't be 
  slowed down a lot. The background thread may however cause additional 
  I/O for looking up the data in RocksDB and for repopulating the caches.

  The background refilling is done in a best-effort way and is not guaranteed 
  to always succeed, e.g. if there is no memory available memory for the cache 
  subsystem, or when an in-memory cache table is currently in a migration phase
  (grow/shrink operation).

  There is a new startup option `--rocksdb.auto-refill-index-caches-on-modify`
  for DB-Servers and single servers, which currently defaults to `false`. If 
  it is set to `true`, the cache refilling will be turned on automatically for
[...]

PR #17696, opened 2022-11-30
Automatically refill in-memory index cache entries after insert/update/replace

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/17643

Automatically refill in-memory index cache entries after insert/update/replace, if so desired.

* Added experimental per-operation option `refillIndexCaches` to write 
  operations, namely:

  - AQL INSERT/UPDATE/REPLACE modification operations
  - single-document insert, update and replace operations
  - multi-document insert, update and replace operations

  If the option is set to `true` any such operation, every currently running 
  transaction will keep track of which in-memory edge index cache entries were 
  invalidated by the transaction, and will try to (re-)fill them later.
  Currently only edge indexes are supported. ArangoDB 3.11 will add support
  for velocypack-based indexes (persistent, hash, sklipist index).

  Example usages:
  - `db.<collection>.insert({ _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `db.<collection>.update(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `db.<collection>.replace(key, { _from: ..., _to: ..., ... }, { refillIndexCaches: true });`
  - `INSERT { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
  - `UPDATE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`
  - `REPLACE { ... } WITH { ... } INTO <collection> OPTIONS { refillIndexCaches: true }`

  The refilling of the in-memory caches for indexes is performed by a 
  background thread, so that the foreground write operation shouldn't be 
  slowed down a lot. The background thread may however cause additional 
  I/O for looking up the data in RocksDB and for repopulating the caches.

  The background refilling is done in a best-effort way and is not guaranteed 
  to always succeed, e.g. if there is no memory available memory for the cache 
  subsystem, or when an in-memory cache table is currently in a migration phase
  (grow/shrink operation).

  The experimental startup option `--rocksdb.auto-refill-index-caches-on-modify`
  for DB-Servers and single servers, which currently defaults to `false`. If 
  it is set to `true`, the cache refilling will be turned on automatically for
  all insert/update/replace operations, so that it doesn't need to be specified
  on the per-operation/per-query level.

  The experimental option `--rocksdb.auto-refill-index-caches-queue-capacity`
  can be used to limit the number of index cache entries that the background 
  thread will queue. This is a safeguard to keep the memory usage at bay in 
  case the background thread is slower than concurrent threads that perform 
  ingestions.

  There are also new experimental startup options to control whether or not 
  the in-memory caches should automatically be seeded upon server restart.
  The option `--rocksdb.auto-fill-index-caches-on-startup` for DB-Servers and 
  single servers turns on this functionality. It currently defaults to `false`.
  If it is set to `true`, the in-memory caches of all eligible indexes will be
  automatically preseeded after the server startup. Note that this may cause 
  additional CPU and I/O load.
  The option `--rocksdb.max-concurrent-index-fill-tasks` is available to limit
  the impact of the automatic index filling at startup. It controls how many 
  concurrent full index filling operations can execute concurrently. The lower 
  this number is, the lower the impact of cache filling, but the longer it will
  take. The default value for this option depends on the number of available 
  cores, and is at least `1`. A value of `0` cannot be used. 
  This option is only relevant if `--rocksdb.auto-fill-index-caches-on-startup` 
  is set to `true`.

  The PR also adds the following metrics:
  - `rocksdb_cache_auto_refill_loaded_total`: Total number of queued items for 
    in-memory index caches refilling. It will always report a value of zero on 
    coordinators.
  - `rocksdb_cache_auto_refill_dropped_total`: Total number of dropped items 
    for in-memory index caches refilling (because number of queued items would 
    exceed the value of `--rocksdb.auto-refill-index-caches-queue-capacity`). 
    It will always report a value of zero on coordinators.
  - `rocksdb_cache_full_index_refills_total`: Total number of in-memory index
    caches refill operations for entire indexes. The counter gets increased for
    every index automatically loaded (because startup option 
    `--rocksdb.auto-fill-index-caches-on-startup` is set to `true`) or when
    full indexes are loaded into memory manually. 
    In cluster deployments the counter will be increased once per eligible 
    index per shard. It will always report a value of zero on coordinators.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17705/
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


Simran-B (approved):

KVS85:
- [ ] https://jenkins.arangodb.biz/view/Documentation/job/arangodb-ANY-examples/2707/

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
```diff
@@ -0,0 +1,363 @@
...
+      ->addOption("--rocksdb.auto-fill-index-caches-on-startup",
+                  "Whether to automatically fill the in-memory edge cache with "
+                  "entries on server startup.",
+                  new options::BooleanParameter(&_fillOnStartup),
+                  arangodb::options::makeFlags(
+                      options::Flags::DefaultNoComponents,
+                      options::Flags::OnDBServer, options::Flags::OnSingle,
+                      options::Flags::Uncommon, options::Flags::Experimental))
+      .setIntroducedIn(30906)
+      .setIntroducedIn(31020)
```
neunhoef, line 102: I think this must be `31002`.

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
```diff
@@ -0,0 +1,363 @@
...
+                  "Whether to automatically (re-)fill the in-memory edge "
+                  "cache with entries on insert/update/replace/remove "
+                  "operations by default.",
+                  new options::BooleanParameter(&_autoRefill),
+                  arangodb::options::makeFlags(
+                      options::Flags::DefaultNoComponents,
+                      options::Flags::OnDBServer, options::Flags::OnSingle,
+                      options::Flags::Uncommon, options::Flags::Experimental))
+      .setIntroducedIn(30906)
+      .setIntroducedIn(31020)
```
neunhoef, line 119: 31002

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
```diff
@@ -0,0 +1,363 @@
...
+          "--rocksdb.auto-refill-index-caches-queue-capacity",
+          "How many changes can be queued at most for automatically refilling "
+          "the edge cache.",
+          new options::SizeTParameter(&_maxCapacity),
+          options::makeFlags(options::Flags::DefaultNoComponents,
+                             options::Flags::OnDBServer,
+                             options::Flags::OnSingle, options::Flags::Uncommon,
+                             options::Flags::Experimental))
+      .setIntroducedIn(30906)
+      .setIntroducedIn(31020)
```
neunhoef, line 148: 31002

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
```diff
@@ -0,0 +1,363 @@
...
+                  "The maximum number of index fill tasks that can run "
+                  "concurrently on server startup.",
+                  new options::SizeTParameter(&_maxConcurrentIndexFillTasks,
+                                              /*minValue*/ 1),
+                  options::makeFlags(
+                      options::Flags::DefaultNoComponents,
+                      options::Flags::OnDBServer, options::Flags::OnSingle,
+                      options::Flags::Uncommon, options::Flags::Experimental))
+      .setIntroducedIn(30906)
+      .setIntroducedIn(31020)
```
neunhoef, line 165: 31002

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
jsteemann, line 102:
```suggestion
      .setIntroducedIn(31002)
```

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
jsteemann, line 119:
```suggestion
      .setIntroducedIn(31002)
```

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
jsteemann, line 148:
```suggestion
      .setIntroducedIn(31002)
```

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
jsteemann, line 165:
```suggestion
      .setIntroducedIn(31002)
```

mpoeter (approved):
LGTM, but due to limited time I could only do a superficial review

neunhoef (approved):
LGTM

PR #17717, opened 2022-12-01
QA-342: mark tests that will test v8 API-Functionality not directly connected to arangodb itself

dothebart:
### Scope & Purpose

Pull all tests that only test V8<->arangodb native APIs into their own new testsuite.
- shell_v8: run once without spawning an arangod (like the client_permissions suite does) 
- shell_server_v8: run once with a server
- run both testsuites only on the single server, since there is no added value from running it in the cluster again


jsteemann (approved):
LGTM

js/client/modules/@arangodb/testsuites/aql.js
```diff
@@ -26,6 +26,8 @@
 // //////////////////////////////////////////////////////////////////////////////
 
 const functionsDocumentation = {
+  'shell_v8': 'Aranodb V8 bridges',
+  'shell_server_v8': 'Aranodb V8 bridges run inside of coordinator / dbserver',
```
jsteemann, line 30:
```suggestion
  'shell_v8': 'Arangodb V8 integration',
  'shell_server_v8': 'Arangodb V8 integration run inside of coordinator / dbserver',
```

PR #17721, opened 2022-12-01
Use minizip from zlib

maierlars:
### Scope & Purpose
We had our own copy of minizip in libarango. However, there is a copy present in zlib. Use this instead.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


dothebart:
Our own copy of the lib is patched 
- to at least have a minimal amount of error feedback to our users (error handling is ... in there)
- support utf-16 filenames.
and several other itsy bitsy bugfixes; Please make sure they're present in the upstream code

PR #17725, opened 2022-12-02
BTS-418: Suboptimal index range calculation with redundant conditions

jsteemann:
### Scope & Purpose

Fixed BTS-418.
Partly redundant conditions, such as in 
```
FOR doc IN col
  LET lowerBound = 10
  LET upperBound = 100
  FILTER doc.uid >= lowerBound AND doc.uid <= upperBound AND doc.uid < 1234
  RETURN doc
```
were not always properly folded into simpler conditions. That resulted in extra filter conditions in the query, which did not affect correctness of results, but could slow queries down unnecessarily.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-418
- [ ] Design document:


CHANGELOG
```diff
@@ -1,14 +1,14 @@
...
 
+* Fixed BTS-418: Suboptimal index range calculation with redundant conditions.
+
 * Add missing metrics for user traffic: Histograms:
     `arangodb_client_user_connection_statistics_bytes_received`
     `arangodb_client_user_connection_statistics_bytes_sent`
   These numbers were so far only published via the statistics API.
   This is needed for Oasis traffic accounting.
 
-* Deleted customizable Pregel (AIR) and Greenspun library
```
jsteemann, line 10: intentional change: this changelog entry appeared in the changelog twice before.

tests/js/server/aql/aql-optimizer-conditions-merging.js
```diff
@@ -0,0 +1,195 @@
...
+      runTests(100, 2);
+    },
+    
+    testConditionsMerging3Conditions : function () {
+      runTests(100, 3);
+    },
+    
+    testConditionsMerging4Conditions : function () {
+      runTests(100, 4);
+    },
```
goedderz, line 188: Nice idea to do some random testing!

One question on this: Do you know (e.g. by testing) how many (percent) of the generated conditions are empty, especially for 4 conditions? I could imagine it's relatively high (haven't done a calculation though), and if so, we might want to tweak the random generation so it gets less likely (and if not, its fine).

jsteemann: Definitely, the more condition parts we have, the more of the conditions yield empty results.
I did a few runs and counted the percentage of empty conditions. On average we get:
* 25% empty conditions for 2 condition parts
* 55% empty conditions for 3 condition parts
* 66% empty conditions for 4 condition parts
That's far from perfect, but I think it's still a good enough test for now.
Adjusting the condition generation so that less empty conditions are produced is doable, but also not 100% trivial, so I would prefer to not do it right now.

goedderz: Agreed, the test is already pretty thorough this way. Thanks for investigating!

PR #17728, opened 2022-12-02
validate `replicationFactor` when creating new databases

jsteemann:
### Scope & Purpose

Validate `replicationFactor` when creating new databases.
Disallow creating new databases with a `replicationFactor` value set to a value lower than `--cluster.min-replication-factor` or higher than `--cluster.max-replication-factor`. Previously the `replicationFactor` settings for new databases were not bounds-checked, only for new collections.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17729
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17730
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +350,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
...
               .getNumber<decltype(vocbaseOptions.replicationFactor)>();
+      if (haveCluster) {
+        uint32_t const minReplicationFactor =
+            server.getFeature<ClusterFeature>().minReplicationFactor();
+        uint32_t const maxReplicationFactor =
+            server.getFeature<ClusterFeature>().maxReplicationFactor();
+        // make sure the replicationFactor value is between the configured min
+        // and max values
+        if (vocbaseOptions.replicationFactor > maxReplicationFactor &&
+            maxReplicationFactor > 0) {
```
MBkkt, line 361: NIT
```suggestion
        if (0 < maxReplicationFactor && maxReplicationFactor < vocbaseOptions.replicationFactor) {
```

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +350,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
       vocbaseOptions.replicationFactor =
           replicationSlice
               .getNumber<decltype(vocbaseOptions.replicationFactor)>();
+      if (haveCluster) {
+        uint32_t const minReplicationFactor =
+            server.getFeature<ClusterFeature>().minReplicationFactor();
+        uint32_t const maxReplicationFactor =
```
MBkkt, line 356: NIT
```suggestion
        auto const minReplicationFactor =
            server.getFeature<ClusterFeature>().minReplicationFactor();
        auto const maxReplicationFactor =
```

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +350,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
...
+            server.getFeature<ClusterFeature>().maxReplicationFactor();
+        // make sure the replicationFactor value is between the configured min
+        // and max values
+        if (vocbaseOptions.replicationFactor > maxReplicationFactor &&
+            maxReplicationFactor > 0) {
+          THROW_ARANGO_EXCEPTION_MESSAGE(
+              TRI_ERROR_BAD_PARAMETER,
+              std::string("replicationFactor must not be higher than "
+                          "maximum allowed replicationFactor (") +
+                  std::to_string(maxReplicationFactor) + ")");
```
MBkkt, line 366:
```suggestion
              absl::StrCat("replicationFactor must not be higher than "
                          "maximum allowed replicationFactor (", maxReplicationFactor, ")"));
```

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +350,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
...
+        } else if (vocbaseOptions.replicationFactor < minReplicationFactor &&
+                   minReplicationFactor > 0) {
+          THROW_ARANGO_EXCEPTION_MESSAGE(
+              TRI_ERROR_BAD_PARAMETER,
+              std::string("replicationFactor must not be lower than "
+                          "minimum allowed replicationFactor (") +
+                  std::to_string(minReplicationFactor) + ")");
```
MBkkt, line 373:
```suggestion
              absl::StrCat("replicationFactor must not be lower than "
                          "minimum allowed replicationFactor (",
                  minReplicationFactor, ")"));
```

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +350,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
...
+        // and max values
+        if (vocbaseOptions.replicationFactor > maxReplicationFactor &&
+            maxReplicationFactor > 0) {
+          THROW_ARANGO_EXCEPTION_MESSAGE(
+              TRI_ERROR_BAD_PARAMETER,
+              std::string("replicationFactor must not be higher than "
+                          "maximum allowed replicationFactor (") +
+                  std::to_string(maxReplicationFactor) + ")");
+        } else if (vocbaseOptions.replicationFactor < minReplicationFactor &&
+                   minReplicationFactor > 0) {
```
MBkkt, line 368: NIT
```suggestion
        } else if (0 < minReplicationFactor && vocbaseOptions.replicationFactor < minReplicationFactor) {
```

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -232,10 +226,10 @@ Result CreateDatabaseInfo::extractUsers(VPackSlice const& users) {
   return Result();
 }
 
-Result CreateDatabaseInfo::extractOptions(VPackSlice const& options,
-                                          bool extractId, bool extractName) {
+Result CreateDatabaseInfo::extractOptions(VPackSlice options, bool extractId,
+                                          bool extractName) {
   if (options.isNone() || options.isNull()) {
-    return Result();
+    options = VPackSlice::emptyObjectSlice();
```
MBkkt, line 232: why did it change?

MBkkt (approved):

arangod/VocBase/VocbaseInfo.cpp
jsteemann: This is exactly the bugfix. When no extra data was sent along with the "create database" request, we exited early, and then didn't set any defaults correctly (e.g. replicationFactor was then hard-coded to 1 and not to a sensible minimum value >= `--cluster.min-replication-factor`).

arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -357,6 +352,29 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options) {
...
+            server.getFeature<ClusterFeature>().minReplicationFactor();
+        auto const maxReplicationFactor =
+            server.getFeature<ClusterFeature>().maxReplicationFactor();
+        // make sure the replicationFactor value is between the configured min
+        // and max values
+        if (0 < maxReplicationFactor &&
+            maxReplicationFactor < vocbaseOptions.replicationFactor) {
+          THROW_ARANGO_EXCEPTION_MESSAGE(
+              TRI_ERROR_BAD_PARAMETER,
+              absl::StrCat("replicationFactor must not be higher than "
```
maierlars, line 366: Is StringUtils::concatT now deprecated in favor of absl or are we going to have both?

maierlars (approved):

arangod/VocBase/VocbaseInfo.cpp
jsteemann: Good question. I would favor using abseil because it is maintained for us.

MBkkt: Also StringUtils::concatT isn't same.
It cannot convert numbers to string.
And few other issues

arangod/VocBase/VocbaseInfo.cpp
MBkkt: Thanks

arangod/VocBase/VocbaseInfo.cpp
maierlars: what issues though? we should fix them, it is used in all active branches.

MBkkt: I didn't mean bugs, so don't think necessary to change concatT code.

If you interested you can compare them.

maierlars: also it does convert number and any other type using `to_string` via adl.

MBkkt: Good, forgot about to_string

Do you have some issue with abseil implementation? 

If you interested to use concat with custom strinify it's also possible

MBkkt: I think my position about concanT same as Jan.

It's ok, but I don't see the reason to maintain own implementation of this function.

For example StrCat can use resize without memse for grow string, but now it's only for libc++.

Of course we can write our own better implementation which can do it for libstdc++ (similar code exist in folly), but is it necessary? I don't think so, but on 3rdParty library works a lot of people, and maybe sometime we update it and it will be done (issue already was created by someone).

PR #17738, opened 2022-12-05
[CINFRA-612] Handle shard creation errors

apetenchea:
### Scope & Purpose

When a collection is created, we start creating the corresponding replicated state, which in turn starts creating the core. The core is responsible for creating the shard, but for that, it needs to look up the collection properties in the agency. There is a slight race in there, because the collection could get removed, before the core gets to look up the properties, in which case it will find nothing.
Currently we crash the server with an error message. This PR makes `getCollectionPlan` return a `ResultT`. The `DocumentCore` checks the value and throws and exception in case something went wrong. This would get caught by `UpdateReplicatedStateAction`.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentCore.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentCore.cpp
@@ -27,7 +27,8 @@
 #include "Replication2/StateMachines/Document/DocumentStateHandlersFactory.h"
 #include "Replication2/StateMachines/Document/DocumentStateMachine.h"
 #include "Replication2/StateMachines/Document/DocumentStateShardHandler.h"
-#include "Basics/application-exit.h"
+
+#include <Basics/application-exit.h>
 
 using namespace arangodb::replication2::replicated_state::document;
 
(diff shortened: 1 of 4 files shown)


goedderz:
This is a valid solution. I wonder though, would it be possible to pass on that information? What I mean is, when the creation of the state is being triggered, I suppose we already have some version of the Plan in hand. Could the collection properties be taken from there and passed down to the Core constructor, so we use the same "snapshot", so this kind of error can't occur? Or is that infeasible?

apetenchea:
> I suppose we already have some version of the Plan in hand. Could the collection properties be taken from there and passed down to the Core constructor, so we use the same "snapshot", so this kind of error can't occur? Or is that infeasible?

This might be possible. When we pass down the collection ID to the core, we have access to the `ClusterCollectionCreationInfo` structure. Among a few other member variables, this contains a `velocypack::Slice json`. I will check that out, maybe it contains the plan information for the collection. I have to investigate this a bit further and see whether it's feasible.

apetenchea:
This PR is currently waiting for #17679, since there's been a refactoring of the shard creation code in there.

apetenchea:
Due to a current refactoring, this PR is no longer relevant. The core is no longer responsible for creating the shard. This is done by the maintenance on the leader, which write a `CreateShard` entry into the log.

PR #17744, opened 2022-12-06
Print the pid of the process which sent the signal

mpoeter:
### Scope & Purpose

- [x] :pizza: New feature

### Checklist


CHANGELOG
```diff
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Print the pid of the process which sent the signal.
```
```suggestion
* Print the pid of the process which sent a SIGABRT or other fatal 
  signal that shuts down ArangoDB ungracefully.
```

PR #17746, opened 2022-12-06
Updated immer to 0.8.0 and made it a submodule.

maierlars:
### Scope & Purpose
Update Immer to 0.8.0 and make it a submodule.


.gitmodules
```diff
@@ -16,3 +16,7 @@
 [submodule "3rdParty/abseil-cpp"]
 	path = 3rdParty/abseil-cpp
 	url = https://github.com/arangodb/abseil-cpp.git
+[submodule "3rdParty/immer"]
+	path = 3rdParty/immer
+	url = https://github.com/arximboldi/immer
```
jsteemann, line 21: IMO I don't think it is good to refer to a repository which is not under our control.
For example, if the maintainer shuts down the repo or withdraws the release, our builds will all fail. Though this is an unlikely situation, we can avoid it entirely and also apply whatever changes/patches on top if we fork the original repository and make the submodule point to the fork.

jsteemann:
The following version info should also be updated:
https://github.com/arangodb/arangodb/blob/devel/LICENSES-OTHER-COMPONENTS.md#immer

.gitmodules
maierlars: How can do such a fork?

jsteemann (approved):
LGTM

PR #17749, opened 2022-12-06
reduce dependency on StringBuffer

jsteemann:
### Scope & Purpose

Reduce dependency on `StringBuffer` where it is not actually needed.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/lib/Rest/GeneralRequest.cpp
+++ b/lib/Rest/GeneralRequest.cpp
@@ -24,8 +24,6 @@
 
 #include "GeneralRequest.h"
 
-#include "Basics/StaticStrings.h"
-#include "Basics/StringBuffer.h"
 #include "Basics/StringUtils.h"
 #include "Basics/VelocyPackHelper.h"
 #include "Basics/debugging.h"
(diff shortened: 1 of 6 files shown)


lib/Rest/GeneralRequest.cpp
```diff
@@ -198,7 +194,7 @@ std::string const& GeneralRequest::header(std::string const& key,
   auto it = _headers.find(key);
   if (it == _headers.end()) {
     found = false;
-    return StaticStrings::Empty;
```
MBkkt, line 201: maybe better to return std::string_view?

jsteemann: I prefer not to do it as part of this PR. There are lots of callers that do `std::string const& value = _request->header(name, found);` which would all need to be changed.
I am not against it in general, but it would make the scope of the PR larger. This is intended to be only a small change.

PR #17759, opened 2022-12-06
Deadlock in DocumentFollowerState

apetenchea:
### Scope & Purpose

PR continued in: #17850

The deadlock involves two threads, A and B, working on the same replicated state.
A is *being dropped*, like this:
1. `VocBaseLogManager::dropReplicatedState` → the `VocBaseLogManager::_guardedData` lock is acquired
2. `ReplicatedState<S>::drop` → the `ReplicatedState::guardedData` lock is acquired
3. `FollowerStateManager<S>::resign` → the `FollowerStateManager<S>::_guardedData` lock is acquired
4. `DocumentFollowerState::resign` → this *blocks* on acquiring the `DocumenFolowerState::_guardedData` lock (race with B1)

B is *applying entries*:
1. `DocumentFollowerState::applyEntries` → the `DocumentFollowerState::_guardedData` lock is acquired (before A4)
2. During `applyEntries`, the stream might be released. This calls into `LogFollower::release` which uses its guard to wait on `_appendEntriesInFlightCondVar` .

In order to unlock this situation, we would need another thread to call `notify_one()` on the condition variable. This is done in `appendEntries`, called from the `RestInternalLogHandler`. The problem is that in order to make the call, we first need to locate the follower, which can be done by `_vocbase.getReplicatedLogFollowerById`, which has to acquire `VocBaseLogManager::_guardedData` lock, which is being held at A1. More so, we might never get another appendEntries with or without the lock, as maybe the test is about to end, in which case we are stuck forever.

A solution to this is to call stream->release() outside of the lock in `applyEntries`, similar to what we do on the leader.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


goedderz:
> In order to unlock this situation, we would need another thread to call notify_one() on the condition variable. This is done in appendEntries, called from the RestInternalLogHandler. The problem is that in order to make the call, we first need to locate the follower, which can be done by _vocbase.getReplicatedLogFollowerById, which has to acquire VocBaseLogManager::_guardedData lock, which is being held at A1. More so, we might never get another appendEntries with or without the lock, as maybe the test is about to end, in which case we are stuck forever.

@apetenchea I talked with Lars, and we don't think this part can be quite right. `release()` only waits on `_appendEntriesInFlightCondVar` while `_appendEntriesInFlight` is `true`. This gets set to `true` at the beginning of a call to `LogFollower::appendEntries()`, and gets set to `false` after its done. And by "being done" I don't mean when it returned, but after the log entries have been written to storage. However, I don't see a reason why this should block, and thus (as far as I can see) should not result in a deadlock.

So now I'm confused. 😊

goedderz (approved):
Alex and I discussed it, and a possible reason for this deadlock is a misuse of the condition variable `_appendEntriesInFlightCondVar`, where we don't unlock the corresponding mutex after setting `_appendEntriesInFlight` to `false`.

The concern noted in [this comment](https://github.com/arangodb/arangodb/pull/17759#discussion_r1043206422) is valid and will be addressed in the refactoring branch.

As we still think the changes here are beneficial, this PR can be merged at any time from my POV.

apetenchea:
Continued in #17850

PR #17760, opened 2022-12-06
better indicate test state during shutdown

dothebart:
### Scope & Purpose

We had a test lock up without precise information. Improve the situation.

- [x] :hankey: Bugfix


MBkkt (approved):

jsteemann (approved):
LGTM

PR #17765, opened 2022-12-07
[CINFRA-623] Enhanced logging

apetenchea:
### Scope & Purpose

During `testFailoverDuringTransaction` we encountered a very strange behavior, on Windows, which we were not able to reproduce nor explain. After a failover has occurred successfully, we start a streaming transaction, but when we try to commit it, the coordinator seems to contact the old leader.
This PR does not provide a bug fix, but it enhances test logging.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature


mchacki (approved):
LGTM, this logging is good in general

PR #17769, opened 2022-12-07
clang-format.ps1: Fix for call operator change in pwsh 7.3.0

Simran-B:
Unbreaks the PowerShell wrapper script. Starting with pwsh v7.3.0, the call operator seems to forward a variable with an empty string as such, causing the `clang-format.sh` script not to find any files (unless explicitly specified).


Dronplane (approved):
LGTM

KVS85 (approved):

PR #17791, opened 2022-12-12
Rid of data protector

MBkkt:
### Scope & Purpose

Make wait deterministic and avoid unnecessary copies

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -1032,26 +990,25 @@ ErrorCode DatabaseFeature::dropDatabase(std::string_view name,
...
 
-      newLists->_databases.erase(it);
-      newLists->_droppedDatabases.insert(vocbase);
+      next = _databases.make(prev);
+      next->erase(name);
+      _droppedDatabases.insert(vocbase);
     } catch (...) {
-      delete newLists;
       return TRI_ERROR_OUT_OF_MEMORY;
     }
```
jsteemann, line 1005: Now that we do not need to prevent leaks by calling `delete` in the catch block, we can make the whole `try...catch` thing obsolete, and simply let the exception escape from here. That could change the effective error code in case an exception occurs, but it will simplify the code and thus probably should be done now.

jsteemann (approved):
LGTM

arangod/RestServer/DatabaseFeature.h
```diff
@@ -257,14 +249,39 @@ class DatabaseFeature : public ArangodFeature {
   std::unique_ptr<DatabaseManagerThread> _databaseManager;
   std::unique_ptr<IOHeartbeatThread> _ioHeartbeatThread;
 
-  std::atomic<DatabasesLists*> _databasesLists;
-  // TODO: Make this again a template once everybody has gcc >= 4.9.2
-  // arangodb::basics::DataProtector<64>
-  mutable basics::DataProtector _databasesProtector;
-  mutable Mutex _databasesMutex;
+  using DatabasesList = containers::FlatHashMap<std::string, TRI_vocbase_t*>;
```
markuspf, line 252: wouldn´t it be better to just use the Guarded data structure here?

MBkkt: It's not same, we don't want write block reads

MBkkt: Maybe it's possible, but it's big behavior difference

MBkkt: More similar rcu approach, where data have epoch

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -72,6 +71,17 @@ using namespace arangodb::basics;
...
 namespace {
+
+template<typename T>
+void waitUnique(std::shared_ptr<T> const& ptr) {
+  // It's safe because we guarantee we have not weak_ptr
+  while (ptr.use_count() != 1) {
+    // 250ms from old DataProtector code authored by Max Neunhoeffer
+    std::this_thread::sleep_for(std::chrono::microseconds(250));
+  }
+  std::atomic_thread_fence(std::memory_order_acquire);
```
mpoeter, line 82: What is this atomic-fence for? And why do we need waitUnique at all if we now use a shared_ptr for the database list?

MBkkt: As I mentioned exists other approaches but I think this most close to old code

mpoeter: Why do we need a HB relation between releasing the shared_ptr in read mode? A thread that is accessing the database list in read mode cannot change anything, so I don't see why we need to have a HB relation. And all changes to the list are protected by a mutex anyway.

MBkkt: Because someone have access to databases pointers, then release shared pointer then we need unique access to this pointer

MBkkt: Let's see example

t1:
read shared_ptr

t2:
lock mutex
update shared_ptr
move to vocbase ptr from shared_ptr list to dropped
// no wait or wait without acquire
unlock mutex

t3 -- gc thread:
lock mutex
access to dropped databese.


We want to t3 should be after t1 release it shared_ptr

MBkkt: So I have chain:
t1 shared_ptr release synchronize with t2 wait
t2 mutex.unlock synchronize with t3 mutex.lock

MBkkt: > In the old code we had to wait until it was ensured that no other thread had a reference to the memory, so it was safe to delete it.

It's pretty same, we have shared_ptr to **map of raw pointers**.
So wait in old and new code protect this raw pointers

mpoeter: Discussed offline - the shared_ptr only protects the databases list, but we actually need to protect reclamation of the raw vocbase pointers from that list, which is why we need to wait until all other shared_ptrs are dropped.

PR #17801, opened 2022-12-13
remove type from vocbase

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1176

Remove `_type` attribute from `TRI_vocbase_t`, in order to simplify setup.
The type attribute is set during construction only, and seems to be set to `TRI_VOCBASE_COORDINATOR` on coordinators and `TRI_VOCBASE_NORMAL` in all other cases. Instead of passing the attribute to the constructor, adjust the few places that actually distinguished between the two different types.
The few differences in behavior could also be handled by two different subclasses of `TRI_vocbase_t`. That would however require larger refactoring now.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1176
- [ ] Design document:


tests/IResearch/IResearchViewCoordinatorTest.cpp
```diff
@@ -122,8 +120,7 @@ TEST_F(IResearchViewCoordinatorTest, test_rename) {
       "{ \"name\": \"testView\", \"type\": \"arangosearch\", \"id\": \"1\", "
       "\"collections\": [1,2,3] }");
 
-  Vocbase vocbase(TRI_vocbase_type_e::TRI_VOCBASE_TYPE_COORDINATOR,
-                  testDBInfo(server.server()));
+  Vocbase vocbase(testDBInfo(server.server()));
```
neunhoef, line 123: This is a coordinator test. Is it clear that the code executed with this Vocbase object will behave the same as before? Does the code "feel" that it is supposed to run in coordinator mode here?

neunhoef (approved):
LGTM. See one little question in a comment.

tests/IResearch/IResearchViewCoordinatorTest.cpp
jsteemann: Previously there have been some `if...else...` branches inside `TRI_vocbase_t` based on the vocbase's `_type` attribute. 
That means for vocbases of type `TRI_VOCBASE_TYPE_COORDINATOR` a few other branches were taken than for vocbases of type `TRI_VOCBASE_TYPE_NORMAL`. This now still happens, but it is not tied to the `_type` attribute anymore, but to the return value of `ServerState::instance()->isCoordinator()`.
There was a direct relationship between 
* `ServerState::instance()->isCoordinator()` => `TRI_VOCBASE_TYPE_COORDINATOR`
* `! ServerState::instance()->isCoordinator()` => `TRI_VOCBASE_TYPE_NORMAL`
As the unit test still passes with the changes from this PR, it probably does not make a difference.

At some point we should also get rid of all the `ServerState::instance()->isXXX()` throughout the code, because that IMO is also not ideal. But the `_type` attribute was yet redundant to the `ServerState` values, so it is for sure not needed.

neunhoef: OK

arangod/RestServer/DatabaseFeature.h
```diff
@@ -39,6 +39,8 @@
 
 #include <condition_variable>
 #include <mutex>
+#include <memory>
```
MBkkt, line 42: Here we have two "lists" droppedDatabases and databases.
I think they can store std::unique_ptr instead of raw pointer.
In such case will be smaller changes in .cpp

MBkkt (changes requested):

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -643,48 +640,46 @@ void DatabaseFeature::stop() {
...
 #endif
 
-  auto stopVocbase = [](TRI_vocbase_t* vocbase) {
-    // iterate over all databases
-    TRI_ASSERT(vocbase != nullptr);
-    if (vocbase->type() != TRI_VOCBASE_TYPE_NORMAL) {
-      return;
-    }
-
+  if (!ServerState::instance()->isCoordinator()) {
```
MBkkt, line 643: in such case we don't need 
```cpp
  std::unique_lock lock{_databasesMutex};
  auto databases = _databases.load();
```

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -771,19 +766,18 @@ void DatabaseFeature::recoveryDone() {
 
   _pendingRecoveryCallbacks.clear();
 
-  auto databases = _databases.load();
+  if (!ServerState::instance()->isCoordinator()) {
```
MBkkt, line 769: early return?

arangod/RestServer/DatabaseFeature.cpp
jsteemann: Done

arangod/RestServer/DatabaseFeature.cpp
jsteemann: Done

arangod/RestServer/DatabaseFeature.h
jsteemann: @MBkkt : It's actually quite complicated to change the "list" types from `TRI_vocbase_t*` to `std::unique_ptr<TRI_vocbase_t>`.
It requires changes in lots of places in DatabaseFeature.cpp, but the most probablematic thing is the ownership "switch" when moving a unique_ptr from `_droppedDatabases` to some temporary local variable for further cleanup.
`absl::FlatHashSet` only has iterators that return a const reference to the pointed-to member, so moving the contents of a `unique_ptr` from such set is actually not easily possible. We could make the vocbases in the lists `std::shared_ptr`s, which would solve this problem, and potentially is even better for the future, so we can return an `std::shared_ptr` from a database lookup function instead of a raw pointer.
But this will require us to change hundreds of internal APIs, so I am reluctant to do it now.
If you still see any room for improvement in this PR, please feel free to push the changes directly into the PR.
Thanks!

MBkkt: Ok, I understood

MBkkt (approved):

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -124,26 +125,25 @@ void DatabaseManagerThread::run() {
...
       {
         std::lock_guard lock{databaseFeature._databasesMutex};
         auto& dropped = databaseFeature._droppedDatabases;
         // check if we have to drop some database
         for (auto it = dropped.begin(), end = dropped.end(); it != end; ++it) {
           TRI_ASSERT(*it != nullptr);
           if ((*it)->isDangling()) {
-            database = *it;
+            database.reset(*it);
             dropped.erase(it);
```
MBkkt, line 139: I'm not sure, is erase noexcept (on practice)?
Anyway it can be easy fixed:
auto* ptr = *it;
erase(it)
database.reset(ptr);

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -633,6 +631,10 @@ void DatabaseFeature::stop() {
   StorageEngine& engine = server().getFeature<EngineSelectorFeature>().engine();
   engine.cleanupReplicationContexts();
 
+  if (ServerState::instance()->isCoordinator()) {
+    return;
+  }
+
   std::unique_lock lock{_databasesMutex};
```
MBkkt, line 638: Please remove lock and load, they are unnecessary.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -771,19 +770,18 @@ void DatabaseFeature::recoveryDone() {
 
   _pendingRecoveryCallbacks.clear();
 
-  auto databases = _databases.load();
+  if (!ServerState::instance()->isCoordinator()) {
+    auto databases = _databases.load();
```
MBkkt, line 774: lock and load should be here.

arangod/RestServer/DatabaseFeature.cpp
MBkkt: Not here?) But it's NIT, so it's ok for me

PR #17802, opened 2022-12-13
[PREG-183] Remove async mode

jvolmer:
Deletes async code - including AsyncSCC and related code. It is not tested and the code is hard to follow with it because it is hacked in.

Docs PR: https://github.com/arangodb/docs/pull/1221


markuspf (approved):
LGTM

PR #17803, opened 2022-12-13
[CINFRA] Split Up LogFollower into resuable and testable

maierlars:
# THIS IS A DRAFT: THERE WILL BE A SEPARATE PR CONTAINING THE `libarango` CHANGES!

## Scope & Purpose

The purpose of this PR is to split up a huge class `LogFollower` into smaller reusable components. These components are not yet reused, but are soon going to be reused in a similar PR for the `LogLeader`. This refactoring allowed us to make the tests much more fine grained and explicit.

Additionally, we improved the locking structure. Locks are now ordered in a tree like structure, which prevents deadlocks. This allowed use to remove some posts onto the scheduler.

### Test Binary
As a side quest, this PR creates a `arango_lightweight` library, that contains the bare minimum to run the tests. It does not include V8, ICU, RocksDB, Aql, Pregel, RestHandler, OpenSSL, ZLib, S3 and much more. This has reduced the test binary to 66MB including all debug symbols. It still requires abseil, which now dominates the build time.

Furthermore, the replication 2 source was split up into two libraries. One library, `arango_replication2_pure` contains code that does not directly interact with arangodb. Any interaction required is decoupled using virtual interfaces. In the `arangodbtests_replication2_pure` test binary, these interfaces are implemented with mocks, usually using gmock.

Then there is `arangodb_replication2` which is used if `arangod` is build. It contains everything that is in `arangodb_replication2_pure` but also contains the arangodb specific implementations of aforementioned interfaced.

### Changes in `libarango`
- The `arango` library uses openssl and zlib in multiple places to compress or encrypt strings. The functions remain in the same header files, but their implementation was moved into `*-advanced.cpp` files. Those cpp files are not build using the `arango_lightweight` library and using those functions will result in a linker error.
- Comparing velocypack always requires ICU to be build, because we have to compare unicode strings. One can disable this using a runtime flag, but at linktime the symbols have to be available. Therefore, comparing velocypack is not possible using the `arango_lightweight` library **for now**. I'm not against having an implementation that only compares strings in the ascii range.


lib/CMakeLists.txt
```diff
@@ -195,20 +238,24 @@ target_link_libraries(arango PUBLIC
...
+  arango_lightweight
+  arango_crashhandler
+=======
   absl::flat_hash_set
   absl::flat_hash_map
   absl::node_hash_set
   absl::node_hash_map
   absl::synchronization
   absl::strings
+>>>>>>> origin/feature/merge-replicated-logs-into-replicated-states
```
markuspf, line 258: Whups?

Documentation/Metrics/arangodb_replication2_rocksdb_write_batch_size.yaml
```diff
@@ -0,0 +1,13 @@
...
+help: |
+  Histogram containing the sizes of write batches issued.
+unit: byte
+type: histogram
+category: Replication
+complexity: medium
+exposedBy:
+  - dbserver
+description: |
+  Histogram containing the sizes of write batches issued. Worker collect input from multiple replicated logs into
```
nerpaula, line 12: worker or workers?

maierlars:
```suggestion
  Histogram containing the sizes of write batches issued. Workers collect input from multiple replicated logs into
```

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+      Result result = co_await asResult(std::move(f));
+      // lock guard to make effects visible
+      if (result.ok()) {
+        LOG_CTX("b6cbf", TRACE, self->loggerContext)
+            << "storage operation completed";
+        // resolve and continue
+        req.promise.setValue(std::move(result));
+        guard = self->guardedData.getLockedGuard();
+        guard->onDiskLog = std::move(req.logResult);
+        guard->onDiskMapping = std::move(req.mappingResult);
```
goedderz, line 216: Is it OK to resolve the promise before the value is visible in the data structures? E.g., if we do an appendEntries, followed by an updateCommitIndex; I assume it's OK because the FollowerCommitManager calculates a resolveIndex from both commit index and the visible log?

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -0,0 +1,358 @@
...
+        auto queue = std::move(guard->queue);
+        guard->queue.clear();
+        guard.unlock();
+        // resolve everything else
+        req.promise.setValue(std::move(result));
+        for (auto& r : queue) {
+          LOG_CTX("507fe", INFO, self->loggerContext)
+              << "aborting storage operation because of error in previous "
+                 "operation";
+          r.promise.setValue(TRI_ERROR_ARANGO_CONFLICT);
```
goedderz, line 236: Would it be prudent to choose another error code? If I saw something fail with a conflict on a follower, I'd probably panic and assume other things have gone horribly wrong. Maybe some `TRI_ERROR_REPLICATION_REPLICATED_LOG_SUBSEQUENT_FAULT` or `TRI_ERROR_REPLICATION_REPLICATED_LOG_STORAGE_TRANSACTION_ABORTED` or so...

arangod/Replication2/ReplicatedLog/Components/LogFollower2.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+auto FollowerManager::resign()
+    -> std::tuple<std::unique_ptr<replicated_state::IStorageEngineMethods>,
+                  std::unique_ptr<IReplicatedStateHandle>, DeferredAction> {
+  // 1. resign the state and receive its handle
+  auto handle = stateHandle->resign();
+  // 2. resign the storage manager to receive the storage engine methods
+  auto methods = storage->resign();
+  // TODO resign CommitManager
```
goedderz, line 169: This sounds like it should be resolved, or is there no race that would allow the commit manager to access the storage- or state handle manager after they're destroyed?

goedderz (commented):
As discussed, I focused on the components. Really nice to read, way cleaner and easier to verify and understand than before.

arangod/Replication2/ReplicatedLog/Components/CompactionManager.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  if (first == last) {
+    return {first, {CompactionStopReason::NothingToCompact{}}};
+  } else if (newCompactionIndex == releaseIndex) {
+    return {newCompactionIndex,
+            {CompactionStopReason::NotReleasedByStateMachine{releaseIndex}}};
+  } else {
+    TRI_ASSERT(newCompactionIndex == largestIndexToKeep);
+    return {newCompactionIndex,
+            {CompactionStopReason::LeaderBlocksReleaseEntry{}}};
```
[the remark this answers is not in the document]
maierlars: yes

arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
```diff
@@ -0,0 +1,74 @@
...
+    std::unique_ptr<IReplicatedStateHandle> stateHandlePtr)
+    : stateHandle(std::move(stateHandlePtr)) {
+  ADB_PROD_ASSERT(stateHandle != nullptr);
+}
+
+auto StateHandleManager::resign() noexcept
+    -> std::unique_ptr<IReplicatedStateHandle> {
+  auto guard = guardedData.getLockedGuard();
+  // TODO assert same methods
+  std::ignore = guard->stateHandle->resignCurrentState();
```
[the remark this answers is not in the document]
maierlars: Right now a state can return just any instance of methods. And I thought we maybe want to assert that it is the same pointer? It is not that important.

arangod/Replication2/ReplicatedLog/Components/ExclusiveBool.h
```diff
@@ -0,0 +1,80 @@
...
+ private:
+  friend struct ExclusiveBoolGuard;
+  std::atomic_flag value;
+};
+
+inline void ExclusiveBoolGuard::reset() noexcept {
+  if (ptr) {
+    ptr->value.clear();
+  }
+}
```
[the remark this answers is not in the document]
maierlars: yes, otherwise we might clear something twice.

arangod/Replication2/ReplicatedLog/Components/LogFollower2.cpp
maierlars: By design, if you depend on another component, it will outlive you. However, it might refuse to act and return an error instead.

I initially planned to resign the commit manager, but it turns out, that there is nothing to resign, except maybe for some promises.

arangod/Replication2/ReplicatedLog/Components/SnapshotManager.h
```diff
@@ -0,0 +1,69 @@
...
+  explicit SnapshotManager(
+      IStorageManager& storage, IStateHandleManager& stateHandle,
+      std::shared_ptr<FollowerTermInformation const> termInfo,
+      std::shared_ptr<ILeaderCommunicator> leaderComm,
+      LoggerContext const& loggerContext);
+  auto invalidateSnapshotState() -> Result override;
+  auto checkSnapshotState() noexcept -> SnapshotState override;
+
+  auto setSnapshotStateAvailable(MessageId msgId, std::uint64_t version)
+      -> Result;
```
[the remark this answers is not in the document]
maierlars: Yes, that was intentional. Or lets put it differently, I didn't see, why `ISnapshotManager` should have this function. If we need it, it is easy to add later.

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
maierlars: Yes, I think this i fine, because who every looks into the log will see spearhead and not what is on disk.

maierlars: Ah not true. There is `getCommitedLog` 🤔 In that case we have to lock the mutex, update the onDisk state and then unlock, resolve, lock.

maierlars: Fixed it.

arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
```diff
@@ -0,0 +1,74 @@
...
+    std::unique_ptr<IReplicatedLogFollowerMethods> ptr) {
+  auto guard = guardedData.getLockedGuard();
+  guard->stateHandle->becomeFollower(std::move(ptr));
+}
+
+void StateHandleManager::acquireSnapshot(const ParticipantId& leader,
+                                         std::uint64_t version) noexcept {
+  auto guard = guardedData.getLockedGuard();
+  // TODO is the correct log index required here?
+  guard->stateHandle->acquireSnapshot(leader, LogIndex{0}, version);
```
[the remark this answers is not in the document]
maierlars: done. Alex removed the parameter already.

arangod/Replication2/ReplicatedLog/Components/AppendEntriesManager.cpp
```diff
@@ -0,0 +1,216 @@
...
+  MeasureTimeGuard timer(*metrics->replicatedLogFollowerAppendEntriesRtUs);
+  // TODO more metrics?
+
+  LoggerContext lctx =
+      loggerContext.with<logContextKeyMessageId>(request.messageId)
+          .with<logContextKeyPrevLogIdx>(request.prevLogEntry);
+  LOG_CTX("7f407", TRACE, lctx) << "receiving append entries";
+
+  Guarded<GuardedData>::mutex_guard_type guard = guarded.getLockedGuard();
+  auto self = shared_from_this();  // required for coroutine to keep this alive
```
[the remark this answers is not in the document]
maierlars: 😵

lib/Basics/errors.dat
```diff
@@ -167,6 +167,7 @@ ERROR_REPLICATION_REPLICATED_LOG_INVALID_TERM,1425,"an invalid term was given","
 ERROR_REPLICATION_REPLICATED_LOG_UNCONFIGURED,1426,"log participant unconfigured","Will be raised when a participant is currently unconfigured, i.e. neither a leader nor a follower."
 ERROR_REPLICATION_REPLICATED_STATE_NOT_FOUND,1427,"replicated state {id:} of type {type:} not found","Will be raised when a specific replicated state was not found."
 ERROR_REPLICATION_REPLICATED_STATE_NOT_AVAILABLE,1428,"replicated state {id:} of type {type:} is unavailable","Will be raised when a specific replicated state was accessed but is not (yet) available."
+ERROR_REPLICATION_REPLICATED_LOG_SUBSEQUENT_FAULT,1429,"operation aborted because a previous operation failed","Will be raised when an operation is aborted to a cascading failure."
```
goedderz, line 170:
```suggestion
ERROR_REPLICATION_REPLICATED_LOG_SUBSEQUENT_FAULT,1429,"operation aborted because a previous operation failed","Will be raised when an operation is aborted due to a cascading failure."
```

goedderz (approved):

goedderz:
Regarding Jenkins results: Compile error on Windows exists in the staging branch already. Failure in linux/single/enterprise was Jenkins-related, successful retry: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-linux.x86-64/27731/

PR #17824, opened 2022-12-15
make RocksDB lock manager faster

jsteemann:
### Scope & Purpose

Slightly modify the internal map container type and LockMap API of RocksDB's built-in lock manager.
This mostly replaces internal usages of `std::unordered_map` with Abseil's `flat_hash_map`, because it turns out to be more cache-friendly and faster on average.
Also simplifies an API for returning a LockMap that previously returned it as an `std::shared_ptr`. Now make it return a raw pointer only. This is safe in our use cases.

Originally I tried creating a new, custom lock manager derived from `rocksdb::LockManager`, and basically copy over and adjust the behavior from `rocksdb::PointLockManager`. However, this turned out to be infeasible, because `rocksdb::PointLockManager` depends on a lot of RocksDB-internal components which are not in its public include directory, and thus cannot be used by code outside of RocksDB.
So replicating all functionality from `rocksdb::PointLockManager` outside of RocksDB would have been a lot of work, or would have required us to include private header files.
The change made in this PR directly modifies the RocksDB source files instead, but it is a small change only, and thus easier to maintain.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

mpoeter (approved):

PR #17828, opened 2022-12-15
fix assertion failure in chaos tests

jsteemann:
### Scope & Purpose

Fix an assertion failure happening in chaos tests (and elsewhere), when write-write conflicts occurred during document updates.
In case of batch updates, upon a ww conflict, 2 results were written to the result VPackBuilder, one for the conflicting document, and one for the error. So the result VPackBuilder contained more entries than the input slice/number of updates.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Transaction/Methods.cpp
```diff
@@ -1901,7 +1901,7 @@ Future<OperationResult> transaction::Methods::modifyLocal(
     Result res = collection->getPhysical()->lookupKeyForUpdate(
         this, key.stringView(), lookupResult);
     if (res.fail()) {
-      if (res.is(TRI_ERROR_ARANGO_CONFLICT)) {
+      if (res.is(TRI_ERROR_ARANGO_CONFLICT) && !isArray) {
```
mpoeter, line 1904: In insertLocal we have a similar check where we also consider the replication type:
https://github.com/arangodb/arangodb/blob/e858626cc7eef181f3366aa11810963754f1c019/arangod/Transaction/Methods.cpp#L1460-L1461
Shouldn't we do the same here? If yes, then this would also apply to the check in line 1947.

arangod/Transaction/Methods.cpp
```diff
@@ -2029,12 +2030,7 @@ Future<OperationResult> transaction::Methods::modifyLocal(
   TRI_ASSERT(replicationType != ReplicationType::FOLLOWER ||
              replicationData->slice().isEmptyArray());
   TRI_ASSERT(!newValue.isArray() || options.silent ||
-             resultBuilder.slice().length() == newValue.length())
-      << "isUpdate: " << isUpdate << ", silent: " << options.silent
-      << ", leader: " << (replicationType == ReplicationType::LEADER)
-      << ", follower: " << (replicationType == ReplicationType::FOLLOWER)
-      << ", newValue: " << newValue.toJson()
-      << ", resultBuilder: " << resultBuilder.slice().toJson();
```
mpoeter, line 2037: Don't we want to keep that output?

mpoeter (approved):

PR #17830, opened 2022-12-15
QA-336: Bug fix/cleanup successfull recovery

dothebart:
### Scope & Purpose

One failing recovery test would add the working directory of all successfull tests as well resulting in huge reports. 

- [x] :hankey: Bugfix

### DIFF
--- a/tests/test-definitions.txt
+++ b/tests/test-definitions.txt
@@ -90,7 +90,7 @@ resilience_failover priority=750 cluster -- --dumpAgencyOnError true
 resilience_move priority=600 cluster -- --dumpAgencyOnError true
 resilience_sharddist cluster -- --dumpAgencyOnError true
 
-recovery_cluster priority=2000 cluster buckets=8 -- --test search
+recovery_cluster priority=2000 cluster buckets=8
 
 # different number of buckets in cluster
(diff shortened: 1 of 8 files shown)


dothebart:
To demonstrate the effect, one run will be done with two sabotaged tests.

tests/test-definitions.txt
```diff
@@ -90,7 +90,7 @@ resilience_failover priority=750 cluster -- --dumpAgencyOnError true
 resilience_move priority=600 cluster -- --dumpAgencyOnError true
 resilience_sharddist cluster -- --dumpAgencyOnError true
 
-recovery_cluster priority=2000 cluster buckets=8 -- --test search
```
jsteemann, line 93: Why this change? Is it established that the non-search recovery tests work in cluster?

dothebart: the cluster suite now does that filtering: 
```
                                       // At the moment only view-tests supported by cluster recovery tests:
                                      ).filter(testname => testname.search('search') >= 0);
```
since I don't think that we should abuse test filtering for this.

jsteemann (approved):
LGTM

PR #17831, opened 2022-12-15
[PREG-203] Move pregel rest handling into pregel directory

jvolmer:
Rest handling cpp files are only used inside Pregel and should therefore be inside Pregel directory.


markuspf (approved):
LGTM; can you try Jenkins once more just to see whether it becomes blue for once?

PR #17832, opened 2022-12-15
Bug fix/SEARCH-441 force index hint fix

Dronplane:
### Scope & Purpose

+ Moves failed forced index hints check to the end of the optimization. To avoid it being triggered for some alternative plan 
    that would be discarded as too costly;
+ PR fixes attribute name check for ByTerm filter with non-constant argument
+ aql-index-hits tests are now runs with all rules enabled to track possible conflicts

Enterprise PR:https://github.com/arangodb/enterprise/pull/1181

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/ExecutionPlan.h
```diff
@@ -433,6 +435,9 @@ class ExecutionPlan {
   /// prefetching on the node level should be executed.
   bool _isAsyncPrefetchEnabled{false};
 
+  // Flag there are collection nodes with forceIndexHint:true
+  unsigned _hasForcedIndexHints{false};
```
gnusi, line 439:
```suggestion
bool _hasForcedIndexHints{false};
```

arangod/Aql/Optimizer.cpp
```diff
@@ -391,6 +391,26 @@ void Optimizer::createPlans(std::unique_ptr<ExecutionPlan> plan,
 
   estimateCosts(queryOptions, estimateAllPlans);
 
+  // Best plan should not have forced hints left.
+  // There might be other plans that has, but we don't care
+  if (_plans.list.front().first->hasForcedIndexHints()) {
```
gnusi, line 396: NIT to remove duplication
```
suggestion
if (auto& plan = _plans.list.front().first; first->hasForcedIndexHints()) {
```

gnusi (changes requested):

arangod/Aql/ExecutionPlan.cpp
```diff
@@ -526,8 +526,16 @@ bool ExecutionPlan::contains(ExecutionNode::NodeType type) const {
 }
 
 /// @brief increase the node counter for the type
-void ExecutionPlan::increaseCounter(ExecutionNode::NodeType type) noexcept {
+void ExecutionPlan::increaseCounter(ExecutionNode const& node) noexcept {
+  auto const type = node.getType();
   ++_typeCounts[type];
+  if (type == ExecutionNode::ENUMERATE_COLLECTION) {
```
jsteemann, line 532: I guess nodes of type `INDEX` are not considered here because they all start up as nodes of type ENUMERATE_COLLECTION initially, and are only converted to INDEX nodes by the optimizer, which already considers index hints in the `use-indexes` optimizer rule.
If so, adding a code comment with the explanation would be nice. Thanks!

jsteemann (approved):
See @gnusi 's comment. Otherwise LGTM

arangod/Aql/ExecutionPlan.h
Dronplane: Forgot to change type when moved away from counting 🤦‍♂️

arangod/Aql/ExecutionPlan.cpp
Dronplane: Yes you are right. Will add a comment

gnusi (approved):

PR #17834, opened 2022-12-16
acquire an index snapshot for document write ops

jsteemann:
### Scope & Purpose

Acquire a snapshot of the (list of) indexes when starting document insert, update/replace and remove operations, and use that snapshot through the operation.
Previously, the list of indexes was acquired multiple times during an operation, and it was (at least in theory) possible that the list of indexes changed between the individual acquistions.
The PR also contains an optimization to not fetch the full document from the storage engine for remove and replace operations in case the full document is not needed to process the operation. This is the case when the collection does not contain any secondary indexes and `returnOld` is not used.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Transaction/IndexesSnapshot.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "IndexesSnapshot.h"
+
+#include "Indexes/Index.h"
+#include "StorageEngine/PhysicalCollection.h"
+
+namespace arangodb {
+
+IndexesSnapshot::IndexesSnapshot(PhysicalCollection& collection)
+    : _locker(collection._indexesLock, collection._indexesLockWriteOwner,
+              __FILE__, __LINE__),
```
mpoeter, line 33: I don't really like that we are accessing some internals of the collection here. Can we instead add a `acquireIndexLock` function to `PhysicalCollection` that returns a Locker?

arangod/Transaction/IndexesSnapshot.cpp
```diff
@@ -0,0 +1,64 @@
...
+      _valid(true) {
+  _indexes.reserve(collection._indexes.size());
+  for (auto const& idx : collection._indexes) {
+    _indexes.emplace_back(idx);
+  }
```
mpoeter, line 38: Why do we copy out the indexes? We only allow access to our index list as long as we have the locker, but if we have the locker, couldn't we just as well access to collection index list?

arangod/Transaction/Methods.cpp
```diff
@@ -913,6 +952,8 @@ struct RemoveProcessor : ReplicatedProcessorBase<RemoveProcessor> {
   // builder for a single, old version of document (will be recycled for each
   // document)
   transaction::BuilderLeaser _previousDocumentBuilder;
+  // whether or not we need to read the previous document version
+  bool const _needToFetchOldDocument;
```
mpoeter, line 956: I think this could me moved to ReplicatedProcessorBase.

mpoeter (approved):

arangod/Transaction/IndexesSnapshot.cpp
jsteemann: Ok, changed

arangod/Transaction/IndexesSnapshot.cpp
jsteemann: In there we could access the `_indexes` property from `PhysicalCollection` directly, without having 2 copy anything.
There are two reasons why I would still prefer creating a copy:
1. `PhysicalCollection::_indexes` is of type `IndexesContainerType`, which in an `std::set` with a custom comparator. Some operations, such as `_indexes[0]` don't feel natural when performed on an `std::set`. I will probably also make more use of the IndexesSnapshot in the future, and will then potentially access members from position 1 on etc. All this is possible with an `std::set` somehow as well, but it feels awkward.
2. If I use a const reference to the original `_indexes` member of `PhysicalCollection` instead of creating a copy, it is still possible that someone gives up the lock (e.g. by calling `IndexesSnapshot::release()`) before accessing the member, which would be unsafe. Or acquire the reference, then call `release()` and after that use the reference. All these potential issues can be avoided by creating a copy.

mpoeter: Ok, if we use a different container then it makes more sense.

PR #17844, opened 2022-12-19
Refactor transaction methods

mpoeter:
### Scope & Purpose

First step in refactoring transaction methods to reduce duplicate code and make it easier to maintain.
Subsequent steps with more reduction will follow.


arangod/Transaction/Methods.h
```diff
@@ -413,7 +413,7 @@ class Methods {
 
   Result triggerIntermediateCommit();
 
- private:
+  // private:
```
jsteemann, line 416: Is that intentional?

arangod/Transaction/Methods.h
```diff
@@ -526,7 +506,7 @@ class Methods {
   Future<OperationResult> truncateLocal(std::string const& collectionName,
                                         OperationOptions& options);
 
- protected:
+  // protected:
```
jsteemann, line 509: intentional?

arangod/Transaction/Methods.cpp
```diff
@@ -409,1386 +409,1607 @@ OperationResult emptyResult(OperationOptions const& options) {
...
+                this->_resultBuilder.slice().isEmptyObject()));
+    TRI_ASSERT(_replicationData->slice().isArray());
+    TRI_ASSERT(_replicationType != Methods::ReplicationType::FOLLOWER ||
+               _replicationData->slice().isEmptyArray());
+    TRI_ASSERT(!this->_value.isArray() || this->_options.silent ||
+               this->_resultBuilder.slice().length() == this->_value.length())
+        << "operation: " << _operationType
+        << ", silent: " << this->_options.silent
+        << ", replicationType: " << static_cast<int>(this->_replicationType)
+        << ", isArray: " << this->_value.isArray();
```
jsteemann, line 628: This should be removed, because if it triggers, it may crash the AssertionLogger with stack overflow!

arangod/Transaction/Methods.cpp
```diff
@@ -409,1386 +409,1607 @@ OperationResult emptyResult(OperationOptions const& options) {
...
-  }
+    if (res.fail()) {
+      return res;
+    }
 
-  // new
-  if (newDoc != nullptr) {
-    builder.add(StaticStrings::New, newDoc->slice());
-  }
+    bool excludeFromReplication = _excludeAllFromReplication;
```
jsteemann, line 778: IMO `excludeFromReplication` will always have the same value as `_excludeAllFromReplication` in a remove operation.
So we can get rid of this local variable here.

arangod/Transaction/Methods.cpp
```diff
@@ -409,1386 +409,1607 @@ OperationResult emptyResult(OperationOptions const& options) {
...
+        if (index) {
+          return index->addErrorMsg(result, key);
+        }
+        return result;
+      } else {
+        TRI_ASSERT(lookupResult.first.isSet());
+        TRI_ASSERT(lookupResult.second.isSet());
+        std::tie(oldDocumentId, oldRevisionId) = lookupResult;
+      }
+    } else if (res.errorNumber() != TRI_ERROR_ARANGO_DOCUMENT_NOT_FOUND) {
```
jsteemann, line 1051:
```suggestion
    } else if (res.isNot(TRI_ERROR_ARANGO_DOCUMENT_NOT_FOUND)) {
```

arangod/Transaction/Methods.cpp
```diff
@@ -409,1386 +409,1607 @@ OperationResult emptyResult(OperationOptions const& options) {
...
+            &_methods, key, lookupResult, ReadOwnWrites::yes);
+        TRI_ASSERT(!lookupRes.ok() || lookupResult.second.isSet());
+        _methods.buildDocumentIdentity(_collection, _resultBuilder,
+                                       _trxColl.id(), key, lookupResult.second,
+                                       RevisionId::none(), nullptr, nullptr);
+        auto index = _collection.getPhysical()->primaryIndex();
+        if (index) {
+          return index->addErrorMsg(res, key);
+        }
+        return res;
```
jsteemann, line 1069:
```suggestion
```

arangod/Transaction/Methods.cpp
```diff
@@ -409,1386 +409,1607 @@ OperationResult emptyResult(OperationOptions const& options) {
...
+    bool didReplace = false;
+    bool excludeFromReplication = _excludeAllFromReplication;
 
-/// @brief add a collection to the transaction for read, at runtime
-DataSourceId transaction::Methods::addCollectionAtRuntime(
-    std::string const& collectionName, AccessMode::Type type) {
-  if (collectionName == _collectionCache.name && !collectionName.empty()) {
-    return _collectionCache.cid;
-  }
+    if (not oldDocumentId.isSet()) {
```
jsteemann, line 1083: Our current convention (used in >= 99.9999% of our code) is to use `!` and not `not` (pun intended).
```suggestion
    if (!oldDocumentId.isSet()) {
```

arangod/Transaction/Methods.cpp
mpoeter: I have adapted the output so that we no longer print the actual payload, so I think this should be fine.

arangod/Transaction/Methods.h
mpoeter: Ah, this was necessary as a temporary workaround to access some private functions. Will fix.

arangod/Transaction/Methods.h
mpoeter: Same as above.

arangod/Transaction/Methods.cpp
mpoeter: I think I saw somewhere in the code `not` instead if `!` which is the only reason I used it. :)
I actually prefer `not` because I find it much more readable - a `!` can be overlooked very easily.

arangod/Transaction/Methods.cpp
jsteemann: Ok, we can leave it here, although I don't think the message will be great help now. Anyway, it's fine.

jsteemann (approved):
LGTM

PR #17862, opened 2022-12-25
[SEARCH-442, BTS-1169] Make GeoJson parsing faster

MBkkt:
### Scope & Purpose

Corresponding to Intel Vtune on some data and query (`SEARCH GEO_INTERSECTS(index_polygon, user_rect)`)
~~I expect 150+% speedup for most queries with polygons in index.
And some speedup for other.~~
> After real life tests we see speedup 250% :)
 6 seconds instead of 16 seconds

* Make GeoJson parsing faster
* Make ShapeContainer intersects/contains faster
* Remove unnecessary functions/classes
* Use FlatHashSet for deduplication uint64_t in query
* Use GeoJson parsing without validation for IResearch GeoFilter
* Was implemented `GEO_INTERSECTS(something, multi_polyline/points)`, but not  `GEO_INTERSECTS(multi_polyline/points, something)` it's fixed now 
* Fix centroid compute for S2LatLngRect in ShapeContainer
* Fix GEO_CONTAINS(S2Polygon, S2LatLngRect), was only angle point checking, just imagine polygon which looks like "X"
* Fix S2LatLng -> S2Point works only for valid S2LatLng:
  It was really strange.  In some cases we produce error to user. In other ignore and works incorrect. In third makes normalize and works correct.
 I think always produce error is better, but in such case we need way to update old user data. It's impossible now, so I think better to always normalize
  So I change behavior: In some queries user couldn't specify not normalized points. It was error/invalid behavior, now they can and it will be valid

### TODO for other PRs

* Using parsing without validation not only for iresearch indexes but also for GeoIndex
* Using binary s2 representation in indexes to avoid geojson -> s2 parsing&transformation overhead
* Enable s2 checks in maintainer mode, problem with S2LatLng -> S2Point because of it wasn't enabled
* Fuzzing tests for our GEO, because now we don't have a lot of tests

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17866

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/IResearch/Geo.cpp
```diff
@@ -21,91 +21,63 @@
...
-        shape.resetCoordinates(ll);
+template<bool Validation>
+bool parseShape(velocypack::Slice vpack, geo::ShapeContainer& region,
+                bool onlyPoint) {
+  Result r;
+  if (vpack.isObject()) {
+    if (onlyPoint) {
+      S2LatLng latLng;
+      r = geo::json::parsePoint<Validation>(vpack, latLng);
+      if (!Validation || r.ok()) {
```
Dronplane, line 45: Why check "Validation" here? "r" is always OK in case of Validation false as I see. Maybe just ASSERT is enough?

arangod/IResearch/Geo.cpp
```diff
@@ -21,91 +21,63 @@
...
         << "Geo JSON or array of coordinates expected, got '"
-        << slice.typeName() << "'";
-
+        << vpack.typeName() << "'";
     return false;
   }
-
-  if (res.fail()) {
-    LOG_TOPIC("4549c", DEBUG, arangodb::iresearch::TOPIC)
+  if (Validation && r.fail()) {
```
Dronplane, line 61: Same here? Is it expected for "parseXXX" without Validation  to return failure?

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,765 @@
...
+
+constexpr std::string_view kTypeStringPoint = "point";
+constexpr std::string_view kTypeStringPolygon = "polygon";
+constexpr std::string_view kTypeStringLineString = "linestring";
+constexpr std::string_view kTypeStringMultiPoint = "multipoint";
+constexpr std::string_view kTypeStringMultiPolygon = "multipolygon";
+constexpr std::string_view kTypeStringMultiLineString = "multilinestring";
+constexpr std::string_view kTypeStringGeometryCollection = "geometrycollection";
+
+// TODO(MBkkt) t as parameter looks better, but compiler produce strange error
```
Dronplane, line 81: "but compiler produce strange error"
Could you please elaborate a bit more? At first glance "toString(t)" it works well  when called from template https://godbolt.org/z/hvMsna5dT

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,765 @@
...
+template<bool Validation>
+Result parseLineImpl(velocypack::Slice vpack, std::vector<S2Point>& vertices) {
+  auto r = parseImpl<Validation, true>(vpack, vertices);
+  if (Validation && ADB_UNLIKELY(!r.ok())) {
+    return r;
   }
-
-  if (Type::MULTI_POINT != type(vpack)) {
-    return TRI_ERROR_BAD_PARAMETER;
+  removeAdjacentDuplicates(vertices);
```
Dronplane, line 242: Do we need this call in case of Validation=false? I guess in this case we expect that someone already called this while parsing with Validation=true before. Or we store "non-normalized" version? Then it is ok to leave it here.

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,765 @@
...
-  VPackSlice coordinates = vpack;
-  if (vpack.isObject()) {
-    coordinates = vpack.get(Fields::kCoordinates);
+  // S2Loop doesn't like duplicates
+  removeAdjacentDuplicates(vertices);
+  if (Validation && ADB_UNLIKELY(vertices.front() != vertices.back())) {
+    return {TRI_ERROR_BAD_PARAMETER, "Loop not closed"};
+  }
+  // S2Loop automatically add last edge
+  if (ADB_LIKELY(vertices.size() != 1)) {
```
Dronplane, line 291: 0 is also "!=1" But popping on empty vector is UB
```suggestion
  if (ADB_LIKELY(!vertices.size() > 1)) {
```

Dronplane, line 291: 0 is also "!=1" But popping on empty vector is UB
```suggestion
  if (ADB_LIKELY(vertices.size() > 1)) {
```

Dronplane, line 291: 0 is also "!=1" But popping on empty vector is UB
```suggestion
  if (ADB_LIKELY(vertices.size() > 1)) {
```

Dronplane (changes requested):

arangod/IResearch/Geo.cpp
MBkkt: Not but in such case will be runtime condition, here it's compile time

lib/Geo/GeoJson.cpp
MBkkt: Just not works, I described it, didn't I?
It just write it's not constant if I use it from another template

Probably compiler bug

lib/Geo/GeoJson.cpp
MBkkt: We checked < 4 above, you cannot here less than one

lib/Geo/GeoJson.cpp
MBkkt: Yes :( As I wrote with Validation false we don't need Validation but still need transforming from valid vpack GeoJson to valid s2 input

lib/Geo/Utils.cpp
```diff
@@ -37,40 +35,38 @@
 #include "Geo/ShapeContainer.h"
 #include "Geo/karney/geodesic.h"
 
-namespace arangodb {
-namespace geo {
-namespace utils {
+namespace arangodb::geo::utils {
 
-Result indexCellsLatLng(VPackSlice const& data, bool isGeoJson,
+Result indexCellsLatLng(velocypack::Slice data, bool geoJson,
```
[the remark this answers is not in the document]
MBkkt: But doesn't work forward declaration.
Btw, I discussed with Jan, a some time ago, and he said VPack using probably outdated

arangod/IResearch/Geo.cpp
MBkkt: Answered in other place

lib/Geo/GeoJson.cpp
Dronplane: Just curious. What was the code to trigger the problem? You description gives not much hints )

MBkkt: I couldn't understand, it was with clang 14 if you curious

lib/Geo/GeoJson.cpp
Dronplane: But "removeAdjacentDuplicates" could change number of `vertices` couldn't it ?

lib/Geo/GeoJson.cpp
Dronplane: Ah I see. I've checked only latest version. Clang 15 works fine. And older gcc/msvc also compiles fine. Looks indeed like a bug in clang 14 and below. And they fixed it in the 15. Maybe add a TODO for removing that kludge once clang has a fix in older versions/we switch to 15?

lib/Geo/GeoJson.cpp
MBkkt: unique on non empty array never return empty array, so I don't want such always negative conditions

MBkkt: If you insist it's really necessary I can change it, buy if not I prefer this way

Dronplane: "unique on non empty array never return empty array" True. Overlooked that. I guess it is safe to leave it as is now.

lib/Geo/GeoJson.cpp
MBkkt: I will add todo to try it in future version but it's only stylistic difference

arangod/IResearch/Geo.cpp
Dronplane: For disjunction this is not true. But anyway  let's leave it as is. This is not performance critical.

Dronplane: Ah no. for "!Validation" it is compile time indeed

MBkkt: In general I think most critical time validation is FindValidationError probably Contains, vpack double validation, so maybe I can don't use it in with r.ok

MBkkt:
After real life tests we see speedup 250% :)

lib/Geo/S2/S2MultiPointRegion.cpp
```diff
@@ -20,31 +20,35 @@
...
-  }
-}
+namespace arangodb::geo {
 
-S2MultiPointRegion::S2MultiPointRegion(S2MultiPointRegion const* other)
-    : num_points_(other->num_points_), points_(nullptr) {
-  if (num_points_ > 0) {
-    points_ = new S2Point[num_points_];
-    memcpy(points_, other->points_, num_points_ * sizeof(S2Point));
+S2Point S2MultiPointRegion::GetCentroid() const noexcept {
```
neunhoef, line 35: I know this algorithm has been in there before the PR (with an additional Normalize at the end). But I think it is seriously broken anyway. Imagine a bunch of points near the international date line, so this side, some on the other. Their Centroid according to this algorithm could easily be near the zero meridian at approximately the same lattitude, which is utter nonsense.
What should we do about this?

lib/Geo/S2/S2MultiPointRegion.cpp
```diff
@@ -53,60 +57,30 @@ S2Cap S2MultiPointRegion::GetCapBound() const {
...
 
-bool S2MultiPointRegion::Contains(S2Point const& p) const {
-  for (int k = 0; k < num_points_; k++) {
-    if (points_[k] == p) {
-      return true;
-    }
-  }
-  return false;
-}
+bool S2MultiPointRegion::Contains(S2Cell const& cell) const { return false; }
```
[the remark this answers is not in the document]
neunhoef: As far as I see this method is not called anywhere, why bother?

lib/Basics/DownCast.h
```diff
@@ -36,7 +36,9 @@ constexpr auto* downCast(From* from) noexcept {
   static_assert(!std::is_reference_v<To>, "'To' shouldn't be reference");
   using CastTo =
       std::conditional_t<std::is_const_v<From>, std::add_const_t<To>, To>;
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
```
neunhoef, line 39: Why is this necessary? `TRI_ASSERT` is only active in maintainer mode anyway?

tests/js/server/aql/aql-queries-geo.js
```diff
@@ -100,10 +100,6 @@ function ahuacatlLegacyGeoTestSuite () {
       db._drop("UnitTestsAhuacatlLocationsNon");
     },
 
-    testLegacyNearInvalidCoordinate : function () {
```
neunhoef, line 103: Why was this test removed?

tests/js/server/aql/aql-queries-geo.js
```diff
@@ -168,10 +164,6 @@ function ahuacatlLegacyGeoTestSuite () {
       actual = runQuery("FOR x IN NEAR(" + locations.name() + ", -70, 70, 2) SORT x.latitude, x.longitude DESC LIMIT 3 RETURN x");
       assertEqual(expected, actual);
     },
-
```
neunhoef, line 171: Why was this test removed?

tests/js/common/shell/shell-index-geo.js
```diff
@@ -289,10 +289,11 @@ function GeoIndexErrorHandlingSuite() {
       var d1 = collection.save({ a : 1 });
       var d2 = collection.save({ loc : null });
       var d3 = collection.save({ loc : [0] });
+      // Normalized to -90, 160
```
neunhoef, line 292: What is normalized here? I do not see a change?

tests/js/common/shell/shell-index-geo.js
```diff
@@ -289,10 +289,11 @@ function GeoIndexErrorHandlingSuite() {
...
       var d4 = collection.save({ loc : [ -100, -200 ] });
       collection.save({ loc : [ -10, -20 ]});
 
-      assertEqual(1, collection.near(0,0).toArray().length);
+      assertEqual(2, collection.near(0,0).toArray().length);
```
neunhoef, line 296: Why was the test adjusted?

tests/js/common/shell/shell-index-geo.js
```diff
@@ -306,7 +307,7 @@ function GeoIndexErrorHandlingSuite() {
       collection.replace(d3, { loc : [ 0 ] });
       collection.replace(d4, { loc : [ -100, -200 ] });
 
-      assertEqual(1, collection.near(0,0).toArray().length);
+      assertEqual(2, collection.near(0,0).toArray().length);
```
neunhoef, line 310: Why was the test adjusted?

neunhoef (commented):
I am through 35 of the 44 files. I have looked at the new geojson parsing algorithms. What is basically missing is some iresearch stuff, the `ShapeContainer` class and its tests. I have to stop for today, but I will finish this some time tomorrow. I submit my comments so far. Nothing serious, just a few questions.

lib/Geo/GeoJson.h
```diff
@@ -52,58 +63,121 @@ enum class Type : uint8_t {
...
+/// @note
+/// In other words I make assumption if Validation is false all parsing
+/// is successful.
+/// @note
+/// In Maintainer mode under x86 linux Validation is false,
+/// use same code as Validation is true. But make TRI_ASSERT that result is ok.
+/// @note
+/// For all functions without Validation parameter it's implicitly equal true.
+/// @note
+/// For all functions without legacy parameter it's implicitly equal false.
```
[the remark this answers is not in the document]
MBkkt: > Why can we afford this? How can we ensure that only data, which has previously been parsed successfully, is parsed without validation?

Because it was in code. We use this function only in some tests and Functions.cpp

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,750 @@
...
-namespace {
-inline bool sameCharIgnoringCase(char a, char b) noexcept {
-  return arangodb::basics::StringUtils::toupper(a) ==
-         arangodb::basics::StringUtils::toupper(b);
+// The main idea is check in our CI 3 compile time branches:
+// 1. Validation true
+// 2. Validation false kIsMaintainer true
+// 3. Validation false kIsMaintainer false
+#if defined(ARANGODB_ENABLE_MAINTAINER_MODE) && defined(__linux__) && \
+    !defined(__aarch64__)
```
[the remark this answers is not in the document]
MBkkt: It's not no validation in maintainer mode on arm, it's needed to test other code path when user specify no validation

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,750 @@
...
-                 ShapeContainer::Type::S2_MULTIPOINT);
-  }
-  return res;
+    auto r = parseLineImpl<Validation>(slice, vertices);
+    if (Validation && ADB_UNLIKELY(!r.ok())) {
+      return r;
+    }
+    // TODO(MBkkt) single unnecessary copy here, needs to patch S2!
+    lines.emplace_back(vertices, S2Debug::DISABLE);
+    // TODO should be FindValidationError instead of assert?
```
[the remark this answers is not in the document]
MBkkt: If no other critical issue, will be fixed with test in next PR

lib/Geo/S2/S2MultiPointRegion.cpp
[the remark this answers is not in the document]
MBkkt: cell have square, square of point is zero, so point cannot contains cell?

Anyway it's from old code?

lib/Basics/DownCast.h
MBkkt: It's my paranoia, now TRI_ASSERT always generate some code, yes it's always false, but there is not guarantee at compile time

tests/js/server/aql/aql-queries-geo.js
MBkkt: Because 

> Fix S2LatLng -> S2Point works only for valid S2LatLng:
It was really strange. In some cases we produce error to user. In other ignore and works incorrect. In third makes normalize and works correct.
I think always produce error is better, but in such case we need way to update old user data. It's impossible now, so I think better to always normalize
So I change behavior: In some queries user couldn't specify not normalized points. It was error/invalid behavior, now they can and it will be valid

tests/js/server/aql/aql-queries-geo.js
MBkkt: Because

> Fix S2LatLng -> S2Point works only for valid S2LatLng:
It was really strange. In some cases we produce error to user. In other ignore and works incorrect. In third makes normalize and works correct.
I think always produce error is better, but in such case we need way to update old user data. It's impossible now, so I think better to always normalize
So I change behavior: In some queries user couldn't specify not normalized points. It was error/invalid behavior, now they can and it will be valid

tests/js/common/shell/shell-index-geo.js
MBkkt: > Fix S2LatLng -> S2Point works only for valid S2LatLng:
It was really strange. In some cases we produce error to user. In other ignore and works incorrect. In third makes normalize and works correct.
I think always produce error is better, but in such case we need way to update old user data. It's impossible now, so I think better to always normalize
So I change behavior: In some queries user couldn't specify not normalized points. It was error/invalid behavior, now they can and it will be valid

tests/js/common/shell/shell-index-geo.js
MBkkt: Because now we have two valid points

arangod/Aql/Functions.cpp
```diff
@@ -6163,8 +6161,10 @@ AqlValue functions::GeoEquals(ExpressionContext* expressionContext,
   AqlValueMaterializer mat2(vopts);
 
   geo::ShapeContainer first, second;
-  Result res1 = geo::geojson::parseRegion(mat1.slice(p1, true), first, false);
-  Result res2 = geo::geojson::parseRegion(mat2.slice(p2, true), second, false);
+  auto res1 = geo::json::parseRegion(mat1.slice(p1, true), first,
```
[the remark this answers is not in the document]
MBkkt: I really believe user of some C++ functions never should write `Result` or `Result<T>` because it's inconvenient and oblivious.

But if you insist I can revert such changes if you point they.

lib/Geo/S2/S2MultiPointRegion.cpp
MBkkt: Maybe we should compute centroid of S2Point(3 component vector)? It's still something strange in general, but at least we don't really suffer because meridian issue.

MBkkt: It's strange because it will be "planar" centroid not on "sphere"

arangod/Aql/Functions.cpp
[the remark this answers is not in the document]
neunhoef: I do not insist, this is your style, so go for it. But here is why I do not share your believe: Typically, code is written once and read many times. For the writer it might be a tiny bit more convenient to write "auto" instead of "Result" or even "Result<T>" (after all, the writer potentially would have to check the signature of the function he calls, but the writer will have to do this anyway). But for the reader, it is suddenly obvious what the type of `res1` is, which is an enormous advantage when reading code. The reader does not have to check the signature of `geo::json::parseRegion` to find out what type `res1` is to judge if subsequent code is correct. Therefore I believe that it is well spent time for the writer to save time and effort for the many readers of this code.

lib/Geo/GeoJson.cpp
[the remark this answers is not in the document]
neunhoef: I do not understand this argument. For my money, this `!defined(__aarch64__)` means that if I compile in maintainer mode on Linux ARM, then I will get
```
constexpr bool kIsMaintainer = false;
```
Therefore, validation will not be turned on in maintainer mode on ARM in cases where it is usually (for efficiency reasons) turned off like this one:
```
template<bool Validation>
Result parseRegion(velocypack::Slice vpack, ShapeContainer& region,
                   std::vector<S2Point>& cache, bool legacy) {
  auto r = parseRegionImpl<(Validation || kIsMaintainer)>(vpack, region, cache,
                                                          legacy);
  TRI_ASSERT(Validation || r.ok()) << r.errorMessage();
  return r;
}
```
Why this difference on ARM? If you want to expose this (unvalidated) code path in tests and use the "ARM" builds for this, then please add a comment to explain this, because otherwise a reader will think "Is there anything which is not working on ARM and is hidden by this condition?".

lib/Geo/GeoJson.cpp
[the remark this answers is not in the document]
neunhoef: Fine.

lib/Geo/S2/S2MultiPointRegion.cpp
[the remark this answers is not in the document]
neunhoef: Well, according to https://s2geometry.io/devguide/s2cell_hierarchy we have
"Cells can be used to represent both points and regions: points are generally represented as leaf cells, while regions are represented as collections of cells at any level(s)."
So I think in some cases cells at the bottom level are used to represent points, although they in actual fact represent small regions of approx. 1cm diameter.
In any case, this is new code, the `S2MultiPointRegion` did not have a method `Contains` for cells in the old code. And as far as I see, the method is not used at all.
So another possibility would be to remove the method again, if it is not clear what it should do.

lib/Geo/S2/S2MultiPointRegion.cpp
neunhoef: I think the algorithm as presented does not make any sense. We could use the (normalized) vectors in R^3, take the centroid there and then project back to the sphere by normalizing. However, this has a singularity. If I would, for example, take two antipodal points, I would end up at (0,0,0) and can no longer produce a result. At the same time, the "centroid" isn't well defined anyway in this case. Maybe we should just not implement this, but this would break backward compatibility.
I do agree that the `Normalize` at the end of the old code is unnecessary.
How about the following: We add a comment saying that this algorithm does not make sense, that the concept is not well defined and that we just leave the old broken algorithm for backwards compatibility?
Furthermore, we say so in the documentation.
What do you think?

lib/Basics/DownCast.h
neunhoef: :-) Fine.

tests/js/server/aql/aql-queries-geo.js
neunhoef: Sorry, I do not get it. As far as I see, the `ToPoint` method of `S2LatLng` works for arbitrary latitude/longitute pairs, since it only applies `sin` and `cos` functions to the angles in radians. Therefore, the S2LatLng values do not have to be normalized at all for this to work.
Where in the PR is any change which changes behaviour which would merit removing the tests?
What are these cases in which we would produce an error vs. ignoring things?

tests/js/server/aql/aql-queries-geo.js
neunhoef: I found it. There is a subtle change in the geo index, which now normalizes points (in the non-geojson case) rather than rejecting them. I think this is a good change, I approve of it. I now understand why the tests had to be changed.

tests/js/common/shell/shell-index-geo.js
neunhoef: I found it. There is a subtle change in the geo index, which now normalizes points (in the non-geojson case) rather than rejecting them. I think this is a good change, I approve of it. I now understand why the tests had to be changed.

tests/js/common/shell/shell-index-geo.js
neunhoef: Understood. It would have been helpful for me to get the information what actual change made the point valid. I have now spent approx. one hour to look for it. In the end, this leads to massively prolonged times for reviews and is a bit frustrating. Anyway, is resolved now.

MBkkt: Everywhere where lat lon point constructed from another points, or from user input, I add assert or normalized.

I think here it's from parseImpl(..., S2LatLng& point) from Geo/Json.cpp

lib/Geo/S2/S2MultiPointRegion.cpp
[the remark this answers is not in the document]
MBkkt: It's not unused, used in some other code

(I wrote in all this methods not implemented, then run tests, and see it failed)


And it had in header
``
bool Contains(S2Cell const& cell) const override { return false; }
``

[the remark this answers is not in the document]
MBkkt: If I remember correct it used in cover compute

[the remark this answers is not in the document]
MBkkt: Anyway I think it's correct, check this
https://github.com/arangodb/arangodb/blob/devel/3rdParty/s2geometry/master/src/s2/s2point_region.h#L55

tests/js/common/shell/shell-index-geo.js
MBkkt: Or maybe from old line 127 in Index.cpp

lib/Geo/S2/S2MultiPointRegion.cpp
MBkkt: I think in most cases it wouldn't be 0, 0, 0, so it can be normalized to unit length, and I think it can make some sense.

I understand it's not correct, but for near points it will pretty normal.
And at the same time, it's not suffer because lat closed to pole/or because 180 meridian.

Anyway I don't against any comments here, but if you allow I will add it in next pr, with other stuff

lib/Geo/GeoJson.cpp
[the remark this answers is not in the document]
MBkkt: Ok, first of all this about Validation and isMaintainer.

User of this function now only about Validation, and then he used parseRegion/etc. He pass Validation parameter.

If passed Validation == false shouldn't be any error, but at the same time shouldn't be any checks (because performance reasons).

So I decide to make Validation == false in Maintainer Mode same as Validation == true + assert which assume no errors was.

But our code only compiled  (and provided to user) but not tested without maintainer. So I decide to don't have enable my checks on arm, to see other code path also works


Excuse me, I though it was clear from comment

lib/Geo/GeoJson.h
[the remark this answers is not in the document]
MBkkt: > Why can we afford this? How can we ensure that only data, which has previously been parsed successfully, is parsed without validation?

Or I don't understand your questions, can you explain what do you mean?

In general this comments just clarification, not behavior changes.

We have one parameters for all functions:
Validation, when Validation true, data will be checked, if Validation false we assume it's correct (because when we insert it we already checked it, and insert only Valid data, at least for arangosearch)

Legacy, it's only about polygons (parsePolygon, parseMultiPolygon, parseRegion), first two functions used only with legacy=false, so I changed they, now they don't have such parameter

neunhoef (changes requested):
This looks very good and promising. I like the speedup potential. I doubt that we will be able to replicate the speedups for normal geo indexes, since we do not store the indexed data in the geo index. I have found a few minor issues, please see comments. Once these are addressed, I think this is good to go.
I doubt that this can sensibly be ported back to 3.9, but 3.10 would be possible. However, it is a large change and large parts (`ShapeContainer` class and `GeoJson.cpp`) have essentially been implemented anew. I have my doubts that it is wise to backport to 3.10, although we probably must, since customers are waiting for the changes.

lib/Geo/S2/S2MultiPolyline.cpp
```diff
@@ -21,91 +21,64 @@
...
+      bounder.AddPoint(point);
     }
   }
   return bounder.GetBound();
 }
 
-S2Cap S2MultiPolyline::GetCapBound() const {
-  return GetRectBound().GetCapBound();
-}
+bool S2MultiPolyline::Contains(S2Cell const& /*cell*/) const { return false; }
```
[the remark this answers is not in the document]
MBkkt: It's used and was in old code. And it's correct

https://github.com/google/s2geometry/blob/master/src/s2/s2polyline.h#L298

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
   }
-
-  resetCoordinates(lat.getNumber<double>(), lng.getNumber<double>());
-  return TRI_ERROR_NO_ERROR;
+  auto subregionBound = polygon.GetSubRegionBound();
+  if (ADB_LIKELY(!subregionBound.Contains(rect))) {
+    return false;
+  }
+  // TODO(MBkkt) ask and check is it really necessary
+  if (ADB_UNLIKELY(rect.is_full() && polygon.is_empty())) {
```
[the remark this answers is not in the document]
MBkkt: Yes, here is typo. Thanks!
Will be fixed.
Should be
```
if (polygon.is_empty() && rect.is_full()) return false;
```

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
-    return Result(TRI_ERROR_BAD_PARAMETER, "Invalid coordinate pair");
+static bool contains(S2Polygon const& polygon, S2LatLngRect const& rect) {
+  if (ADB_UNLIKELY(rect.is_point())) {
+    TRI_ASSERT(rect.lo().is_valid());
+    return polygon.Contains(rect.lo().ToPoint());
   }
-
-  resetCoordinates(lat.getNumber<double>(), lng.getNumber<double>());
-  return TRI_ERROR_NO_ERROR;
+  auto subregionBound = polygon.GetSubRegionBound();
```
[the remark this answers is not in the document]
MBkkt: It's correct because
```
  // Since "bound_" is not exact, it is possible that a polygon A contains
  // another polygon B whose bounds are slightly larger.  "subregion_bound_"
  // has been expanded sufficiently to account for this error, i.e.
  // if A.Contains(B), then A.subregion_bound_.Contains(B.bound_).
  S2LatLngRect subregion_bound_;
```
```
bool S2Polygon::Contains(const S2Polygon& b) const {
  // It's worth checking bounding rectangles, since they are precomputed.
  // Note that the first bound has been expanded to account for possible
  // numerical errors in the second bound.
  if (!subregion_bound_.Contains(b.bound_)) {
    // simplifield
    return false;
  }
```
In general I just moved it from S2Polygon::Contains to not construct S2Polygon from S2LatLngRect

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
-      return equals(static_cast<S2LatLngRect const*>(cc->_data));
+    case Type::S2_POLYLINE: {
+      auto const& lhs = basics::downCast<S2Polyline>(*_data);
+      auto const& rhs = basics::downCast<S2Polyline>(*other._data);
+      return lhs.Equals(rhs);
     }
-    case ShapeContainer::Type::S2_POLYGON: {
-      return equals(static_cast<S2Polygon const*>(cc->_data));
+    case Type::S2_LATLNGRECT: {
+      auto const& lhs = basics::downCast<S2LatLngRect>(*_data);
```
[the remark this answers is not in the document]
MBkkt: It's already exist for S2_LATLNGRECT. But ok, it doesn't matter for me

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
       }
-
-      for (int k = 0; k < pts1->num_points(); k++) {
-        if (!equals(
-                Coordinate(S2LatLng::Latitude(pts1->point(k)).degrees(),
-                           S2LatLng::Longitude(pts1->point(k)).degrees()),
-                Coordinate(S2LatLng::Latitude(pts2->point(k)).degrees(),
-                           S2LatLng::Longitude(pts2->point(k)).degrees()))) {
+      for (size_t i = 0; i != size; ++i) {
+        if (lhsPoints[i] != rhsPoints[i]) {
```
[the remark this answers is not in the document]
MBkkt: Completely agree, it's definitely wrong, but fix not cheap and easy, so now I just leave as it was

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
+      auto const& rhsLines = rhs.Impl();
+      auto const size = lhsLines.size();
+      if (size != rhsLines.size()) {
         return false;
       }
-
-      for (size_t k = 0; k < lines2->num_lines(); k++) {
-        if (!this->equals(&(lines1->line(k)), &(lines2->line(k)))) {
+      for (size_t i = 0; i != size; ++i) {
+        if (!lhsLines[i].ApproxEquals(rhsLines[i])) {
```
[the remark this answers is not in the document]
MBkkt: Yep

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
-    case ShapeContainer::Type::S2_POINT: {
-      S2Point const& p = static_cast<S2PointRegion*>(cc->_data)->point();
-      return equals(S2LatLng::Latitude(p).degrees(),
-                    S2LatLng::Longitude(p).degrees());
-    }
-    case ShapeContainer::Type::S2_POLYLINE: {
-      return equals(static_cast<S2Polyline const*>(cc->_data));
+  switch (_type) {
+    case Type::EMPTY:
+      return true;
```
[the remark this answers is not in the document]
MBkkt: Yes, I think it doesn't need comment because code with if just a few line above

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
   if (e.flattening() == 0.0) {
     switch (_type) {
-      case Type::S2_LATLNGRECT:
-        // only used in legacy situations
-        return static_cast<S2LatLngRect*>(_data)->Area() *
-               kEarthRadiusInMeters * kEarthRadiusInMeters;
-      case Type::S2_POLYGON:
-        return static_cast<S2Polygon*>(_data)->GetArea() *
-               kEarthRadiusInMeters * kEarthRadiusInMeters;
+      case Type::S2_LATLNGRECT: {
```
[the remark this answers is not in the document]
MBkkt: Ok

lib/Geo/S2/S2MultiPolyline.cpp
```diff
@@ -21,91 +21,64 @@
...
-void S2MultiPolyline::Encode(Encoder* const encoder) const {
-  encoder->Ensure(10);  // sufficient
-  encoder->put8(kCurrentEncodingVersionNumber);
-  encoder->put32(lines_.size());
-  for (size_t k = 0; k < lines_.size(); k++) {
-    S2Polyline const& line = lines_[k];
-    line.Encode(encoder);
-  }
-  assert(encoder->avail() >= 0);
+bool S2MultiPolyline::Contains(S2Point const& /*p*/) const {
```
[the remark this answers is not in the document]
MBkkt: It's old behavior! And it's used

For contains we don't warn user, only for intersects, for contains it's implicitly false

lib/Geo/ShapeContainer.cpp
```diff
@@ -21,882 +21,652 @@
...
   auto loop = std::make_unique<S2Loop>(v, S2Debug::DISABLE);
   return S2Polygon{std::move(loop), S2Debug::DISABLE};
 }
 
-}  // namespace
-
-Result ShapeContainer::parseCoordinates(VPackSlice const& json, bool geoJson) {
-  if (!json.isArray() || json.length() < 2) {
-    return Result(TRI_ERROR_BAD_PARAMETER, "Invalid coordinate pair");
+static bool contains(S2LatLngRect const& rect, S2Polyline const& polyline) {
```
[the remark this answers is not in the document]
MBkkt: Old code, just merged to single place.

lib/Geo/ShapeContainer.cpp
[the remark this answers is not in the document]
neunhoef: ok. Understood.

lib/Geo/GeoJson.cpp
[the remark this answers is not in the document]
neunhoef: OK, now I understand. Then let's add a comment in the code so that others understand, too.

lib/Geo/S2/S2MultiPointRegion.cpp
neunhoef: Fine.

lib/Geo/S2/S2MultiPolyline.cpp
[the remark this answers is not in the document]
neunhoef: ok.

lib/Geo/S2/S2MultiPolyline.cpp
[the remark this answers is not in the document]
neunhoef: ok.

lib/Geo/ShapeContainer.cpp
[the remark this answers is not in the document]
neunhoef: ok.

lib/Geo/ShapeContainer.cpp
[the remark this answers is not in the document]
neunhoef: Well, at least I have needed some time to see this. You should be aware that every time somebody reads your code and does not immediately understand, this somebody will be a tiny bit angry. Anything which can help to overcome this is worth it. But OK.

lib/Geo/GeoJson.cpp
```diff
@@ -39,616 +41,750 @@
...
-}  // namespace
 
-namespace {
-inline bool sameCharIgnoringCase(char a, char b) noexcept {
-  return arangodb::basics::StringUtils::toupper(a) ==
-         arangodb::basics::StringUtils::toupper(b);
+// The main idea is check in our CI 3 compile time branches:
+// 1. Validation true
+// 2. Validation false kIsMaintainer true
+// 3. Validation false kIsMaintainer false
```
neunhoef, line 55:
```suggestion
// 3. Validation false kIsMaintainer false
// Therefore, we run tests with validation always on on x86_64 and with no validation on aarch64
// to get good test coverage for both cases.
```

MBkkt: I will do it in next PR! this branch: https://github.com/arangodb/arangodb/tree/update-geo-s2-analyzer

Dronplane (approved):
LGTM

PR #17865, opened 2022-12-27
upgrade velocypack to newest version

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1184

The newest velocypack version introduces some incompatible changes, e.g. the default parameter for the iteration order in ObjectIterator got removed. This PR makes the ArangoDB code compatible with the new velocypack version.

None of the changes in this PR should affect functionality, but as object iteration order may be changed, there may be unntended side effects caused by these changes.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1184
- [ ] Design document:

### DIFF
--- a/arangod/Agency/Job.cpp
+++ b/arangod/Agency/Job.cpp
@@ -100,14 +100,16 @@ write_ret_t singleWriteTransaction(AgentInterface* _agent,
       VPackArrayBuilder onePair(&envelope);
       {
         VPackObjectBuilder mutationPart(&envelope);
-        for (auto pair : VPackObjectIterator(trx[0])) {
+        for (auto pair :
+             VPackObjectIterator(trx[0], /*useSequentialIteration*/ false)) {
           envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                        pair.value);
         }
       }
       if (trx.length() > 1) {
         VPackObjectBuilder preconditionPart(&envelope);
-        for (auto pair : VPackObjectIterator(trx[1])) {
+        for (auto pair :
+             VPackObjectIterator(trx[1], /*useSequentialIteration*/ false)) {
           envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                        pair.value);
         }
@@ -142,14 +144,16 @@ trans_ret_t generalTransaction(AgentInterface* _agent,
           VPackArrayBuilder onePair(&envelope);
           {
             VPackObjectBuilder mutationPart(&envelope);
-            for (auto pair : VPackObjectIterator(singleTrans[0])) {
+            for (auto pair : VPackObjectIterator(
+                     singleTrans[0], /*useSequentialIteration*/ false)) {
               envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                            pair.value);
             }
           }
           if (singleTrans.length() > 1) {
             VPackObjectBuilder preconditionPart(&envelope);
-            for (auto pair : VPackObjectIterator(singleTrans[1])) {
+            for (auto pair : VPackObjectIterator(
+                     singleTrans[1], /*useSequentialIteration*/ false)) {
               envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                            pair.value);
             }
@@ -190,14 +194,16 @@ trans_ret_t transient(AgentInterface* _agent,
       VPackArrayBuilder onePair(&envelope);
       {
         VPackObjectBuilder mutationPart(&envelope);
-        for (auto pair : VPackObjectIterator(trx[0])) {
+        for (auto pair :
+             VPackObjectIterator(trx[0], /*useSequentialIteration*/ false)) {
           envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                        pair.value);
         }
       }
       if (trx.length() > 1) {
         VPackObjectBuilder preconditionPart(&envelope);
-        for (auto pair : VPackObjectIterator(trx[1])) {
+        for (auto pair :
+             VPackObjectIterator(trx[1], /*useSequentialIteration*/ false)) {
           envelope.add("/" + Job::agencyPrefix + pair.key.copyString(),
                        pair.value);
         }
@@ -339,7 +345,8 @@ bool Job::finish(std::string const& server, std::string const& shard,
         addRemoveJobFromSomewhere(finished, "Pending", _jobId);
 
         if (operations.length() > 0) {
-          for (auto oper : VPackObjectIterator(operations)) {
+          for (auto oper : VPackObjectIterator(
+                   operations, /*useSequentialIteration*/ true)) {
             finished.add(oper.key.stringView(), oper.value);
           }
         }
@@ -357,7 +364,8 @@ bool Job::finish(std::string const& server, std::string const& shard,
       if (preconditions.isObject() &&
           preconditions.length() > 0) {  // preconditions --
         VPackObjectBuilder precguard(&finished);
-        for (auto prec : VPackObjectIterator(preconditions)) {
+        for (auto prec : VPackObjectIterator(preconditions,
+                                             /*useSequentialIteration*/ true)) {
           finished.add(prec.key.stringView(), prec.value);
         }
       }  // -- preconditions
@@ -927,7 +935,7 @@ void Job::addPutJobIntoSomewhere(Builder& trx, std::string const& where,
       trx.add("timeFinished",
               VPackValue(timepointToString(std::chrono::system_clock::now())));
     }
-    for (auto obj : VPackObjectIterator(job)) {
+    for (auto obj : VPackObjectIterator(job, /*useSequentialIteration*/ true)) {
       trx.add(obj.key.stringView(), obj.value);
     }
     if (!reason.empty()) {
--- a/arangod/Agency/RestAgencyPrivHandler.cpp
+++ b/arangod/Agency/RestAgencyPrivHandler.cpp
@@ -246,8 +246,9 @@ RestStatus RestAgencyPrivHandler::execute() {
             redirectRequest(slice.get("id").copyString());
             return RestStatus::DONE;
           }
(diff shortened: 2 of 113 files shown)


MBkkt (approved):

arangod/Agency/Job.cpp
```diff
@@ -927,7 +935,7 @@ void Job::addPutJobIntoSomewhere(Builder& trx, std::string const& where,
       trx.add("timeFinished",
               VPackValue(timepointToString(std::chrono::system_clock::now())));
     }
-    for (auto obj : VPackObjectIterator(job)) {
+    for (auto obj : VPackObjectIterator(job, /*useSequentialIteration*/ true)) {
```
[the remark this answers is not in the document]
MBkkt: Honestly all false needs comments why

MBkkt:
https://github.com/arangodb/velocypack/pull/147

kvahed (commented):
We need a true performance measurement test.
Are we making theses changed on a hunch, a theoretical chapter in a book?

arangod/Replication2/Methods.cpp
```diff
@@ -632,9 +630,9 @@ struct ReplicatedLogMethodsCoordinator final
           std::vector<LogIndex> indexes;
           indexes.reserve(payloadSize);
           auto indexIter = velocypack::ArrayIterator(result.get("indexes"));
-          std::transform(
-              indexIter.begin(), indexIter.end(), std::back_inserter(indexes),
-              [](auto const& it) { return it.template extract<LogIndex>(); });
+          for (auto it : indexIter) {
+            indexes.push_back(it.template extract<LogIndex>());
+          }
```
kvahed, line 635: WTF?
Why would you do that? `std::transform` does autovectorizationand other cool stuff. Like AVX2 on nummeric transformations etc. Loop unrolling. Why?

arangod/Agency/RestAgencyPrivHandler.cpp
```diff
@@ -246,8 +246,9 @@ RestStatus RestAgencyPrivHandler::execute() {
             redirectRequest(slice.get("id").copyString());
             return RestStatus::DONE;
           }
-          for (auto const& obj : VPackObjectIterator(ret->slice())) {
-            result.add(obj.key.copyString(), obj.value);
+          for (auto obj : VPackObjectIterator(
```
maierlars, line 249: having `auto` instead of `auto const&` now makes a copy. However, the type returned by `operator*` is just two pointers, so its cheap to copy. If the implementation would not be visible to the translation unit, this would have been translated to a pass by reference return value, since the ABI only allows for a single register as return value. (and making it `auto` instead of `auto const&` would have had no effect)

But the implementation is known, so it will be optimized. In both cases, the compiler produces identical code. https://godbolt.org/z/fPo1oa1P8

arangod/Aql/AqlCall.cpp
```diff
@@ -145,13 +145,14 @@ auto AqlCall::fromVelocyPack(velocypack::Slice slice) -> ResultT<AqlCall> {
...
-    auto const keySlice = it.key;
+  for (auto it :
+       velocypack::ObjectIterator(slice, /*useSequentialIteration*/ true)) {
+    auto keySlice = it.key;
     if (ADB_UNLIKELY(!keySlice.isString())) {
       return Result(TRI_ERROR_TYPE_ERROR,
                     "When deserializating AqlCall: Key is not a string");
     }
-    auto const key = keySlice.stringView();
+    auto key = keySlice.stringView();
```
maierlars, line 155: why was `const` removed? does it have implications for code generation?

tests/js/server/aql/aql-geo.js
```diff
@@ -1045,8 +1045,10 @@ function searchAliasGeoSuite() {
   return suite;
 }
 
+/*
 jsunity.run(arangoSearchGeoSuite);
 jsunity.run(searchAliasGeoSuite);
+*/
```
maierlars, line 1051: Disable two test suits? intentional?

tests/Aql/QueryHelper.cpp
```diff
@@ -45,16 +45,18 @@ auto vpackHasNoneRecursive(VPackSlice slice) -> bool {
...
-    auto iter = VPackArrayIterator(slice);
-    return std::any_of(iter.begin(), iter.end(),
-                       [](auto slice) { return vpackHasNoneRecursive(slice); });
-  }
-  if (slice.isObject()) {
-    auto iter = VPackObjectIterator(slice);
-    return std::any_of(iter.begin(), iter.end(), [](auto pair) {
-      return vpackHasNoneRecursive(pair.key) ||
-             vpackHasNoneRecursive(pair.value);
-    });
```
maierlars, line 57: It took me this long to figure out why you rewrite all these stl algorithms, because `end` now returns a sentinel, right?

arangod/Aql/Functions.cpp
```diff
@@ -9138,20 +9134,18 @@ AqlValue functions::Interleave(aql::ExpressionContext* expressionContext,
...
   while (!iters.empty()) {  // in this loop we only deal with nonempty arrays
     for (auto i = iters.begin(); i != iters.end();) {
-      builder->add(i->current.value());  // thus this will always be valid on
-                                         // the first iteration
-      i->current++;
-      if (i->current == i->end) {
+      builder->add(i->value());  // thus this will always be valid on
+                                 // the first iteration
+      i->next();
+      if (*i == std::default_sentinel_t{}) {
```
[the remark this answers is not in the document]
maierlars: there is `std::default_sentinel`.

tests/Aql/QueryHelper.cpp
maierlars: You could instead use `std::ranges::any_of(iter, [](auto pair) { ... });`. These are updated versions to support a different end-iterator type (in this case the default_sentinel)

arangod/Aql/Functions.cpp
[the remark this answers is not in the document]
MBkkt: I know, it's same asm, but from valid () looks better.
So >nit

arangod/Replication2/Methods.cpp
MBkkt: std::transform never do auto vectorization for Object/ArrayIterator, because they are forward iterator.

And btw if it's auto vectorization it's same as for loop auto vectorization

tests/js/server/aql/aql-geo.js
MBkkt: I think it's my fault. It's not intentional.
Thanks!

tests/Aql/QueryHelper.cpp
MBkkt: std::ranges not supported in clang + libstdc++ configuration.

I think it's common for our development proccess (sanitizers or local runs)

MBkkt: But it can be boost...

arangod/Replication2/Methods.cpp
MBkkt: It was done because std::transform not supported different types of iterators

MBkkt:
@kvahed 

> We need a true performance measurement test.
Are we making theses changed on a hunch, a theoretical chapter in a book

I don't think remove creation extra object, remove extra conditions, remove extra payload (same alignment)
Needs additional tests, it's just obliviously better.

If you talk about true vs false in ObjectIterator, I think it's also obliviously, because with false (old default) you need to make lookup in Slice in every operator*!
So it's probably was O(n*log(n)) instead of O(n)

tests/Aql/QueryHelper.cpp
MBkkt: https://godbolt.org/z/WMhzrceWf proof

maierlars:
> If you talk about true vs false in ObjectIterator, I think it's also obliviously, because with false (old default) you need to 
 make lookup in Slice in every operator*!
> So it's probably was O(n*log(n)) instead of O(n).
Lookup of the nth element should be constant unless its a compact object, but in that case we always do a sequential scan.

tests/Aql/QueryHelper.cpp
MBkkt: https://godbolt.org/z/ec6vTKxjr proof

But any_of probably here probably can work

If I correct no problem with algorithm header but have with ranges

MBkkt:
> Lookup of the nth element should be constant unless its a compact object, but in that case we always do a sequential scan.

You are right, my mistake, but still compare computing offset with computing size of element (which just lookup head for type and size if necessary)

Anyway, I think this PR can be done better, and I want to suggest to Jan separate ObjectIterator to ObjectIterator (which will be sequential) and OrderedObjectIterator (which will be not sequential)

I think it will be  more convenient

markuspf:
Token generic comment

If all places are touched to put a boolean parameter in for `VPackObjectIterator`, wouldn't it be nicer to have two different Iterators, the old one and `VPackSequentialObjectIterator`? 

That would make it explicit what the iterator does without the need for an obscure boolean parameter which needs a comment to explain what it means.

MBkkt:
@markuspf already was suggested and accepted see comment above.

But one note, if you're see PR the main idea you always almost need sequential iterator so current ObjectIterator just became sequential (btw same implementation as array) and if you need special sorted/ordered iteration you need to use special type of iterator.

jsteemann:
Obsoleted by simpler PR https://github.com/arangodb/arangodb/pull/18009

PR #17870, opened 2023-01-03
jemalloc cache_oblivious configurable, default false

neunhoef:
This "fixes" an option of jemalloc which can safe considerable amounts
of memory without impacting performance (according to our measurements).
In this PR, we add configurability of this option from the environment
(forgotten in jemalloc) and at the same time set the default to `false`.

- Make `cache_oblivious` configurable for jemalloc from ENV.
- Make `cache_oblivious:false` the default for jemalloc.

This option does the following:

If set to true (old default), jemalloc allocates for every allocation
of 16384 or more bytes one additional page (4k) to change the base
address that it is not divisible by 4096. This could potentially make
sense since it helps the CPU caches, in the case that lots of accesses
to the beginning of such blocks happen (because of the limited
associativity) of the CPU caches.

However, on the flip side, it "wastes" 4K for every allocation. Since
the RocksDB buffer cache does most of its allocations for 16384 bytes,
this is 25% additional RAM usage, which hurts.

Setting the option to `false` helps to not do this optimization. So far,
our measurements suggest that this does not have a bad influence on
performance. I guess that the RocksDB buffer cache does not access the
beginning of allocated blocks more frequently than other places.

The plan is to backport this to 3.10 and 3.9, but only the change to
make this configurable, and to keep the default of `true`.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17870
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/17872

### DIFF
--- a/3rdParty/jemalloc/CMakeLists.txt
+++ b/3rdParty/jemalloc/CMakeLists.txt
@@ -28,7 +28,7 @@ if (LINUX OR DARWIN)
   else ()
     set(JEMALLOC_CC_TMP "${CMAKE_C_COMPILER}")
     set(JEMALLOC_CXX_TMP "${CMAKE_CXX_COMPILER}")
-    set(JEMALLOC_CONFIG "background_thread:true")
+    set(JEMALLOC_CONFIG "background_thread:true,cache_oblivious:false")
   endif ()
 
   if (USE_JEMALLOC_PROF)
--- a/3rdParty/jemalloc/v5.3.0/src/jemalloc.c
+++ b/3rdParty/jemalloc/v5.3.0/src/jemalloc.c
@@ -30,9 +30,9 @@
 /* Data. */
 
 /* Runtime configuration options. */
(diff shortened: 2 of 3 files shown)


jsteemann (approved):
LGTM!

dothebart:
has this been submitted upstream?
Please add a hint to this to `3rdParty/README_maintainers.md` so it might not be forgotten in case of upgrade!

jsteemann:
New versions of jemalloc provide that configure option. So there is no need to submit the change upstream.

dothebart:
do we want to upgrade any time soon?

jsteemann:
No, there hasn't been a newer jemalloc release since the release we are currently on. 
And jemalloc doesn't seem to release very often. So there are no plans to upgrade any soon.

PR #17877, opened 2023-01-04
Added retry request for the last batch

cpjulia:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,11 @@
 devel
 -----
 
+* APM-592: Added a retriable flag for queries in which, case the batch response
+  object isn't successfully received, by sending a POST request with 
+  `_api/cursor/<cursorId>/<batchId>`, the response object from the `batchd` that
+  had not been received before would be returned. The request is idempotent.
```
jsteemann, line 7: Sorry, I am not able to parse that sentence. Can it be rephrased and simplified?

arangod/Aql/QueryOptions.cpp
```diff
@@ -201,6 +202,9 @@ void QueryOptions::fromVelocyPack(VPackSlice slice) {
   if (value = slice.get("stream"); value.isBool()) {
     stream = value.getBool();
   }
+  if (value = slice.get("retriable"); value.isBool()) {
+    retriable = value.getBool();
```
jsteemann, line 206: Can be done for this assignment and the ones above and below it:
```suggestion
    retriable = value.isTrue();
```
Difference: `getBool()` checks if the slice contains a boolean and will throw otherwise. Because we already established that the slice contains a boolean, we can directly call `isTrue()` on the slice.

arangod/Utils/Cursor.h
```diff
@@ -70,6 +74,8 @@ class Cursor {
 
   inline bool hasCount() const { return _hasCount; }
 
+  inline bool isRetriable() const { return _isRetriable; }
+
   inline double ttl() const { return _ttl; }
 
   inline double expires() const {
```
jsteemann, line 81:
```suggestion
  bool hasCount() const noexcept { return _hasCount; }

  bool isRetriable() const noexcept { return _isRetriable; }

  double ttl() const noexcept { return _ttl; }

  double expires() const noexcept {
```

arangod/Utils/CursorRepository.h
```diff
@@ -73,7 +73,7 @@ class CursorRepository {
   //////////////////////////////////////////////////////////////////////////////
 
  public:
-  Cursor* createFromQueryResult(aql::QueryResult&&, size_t, double, bool);
+  Cursor* createFromQueryResult(aql::QueryResult&&, size_t, double, bool, bool);
```
jsteemann, line 76:
```suggestion
  Cursor* createFromQueryResult(aql::QueryResult&& result, size_t batchSize, double ttl, bool hasCount, bool isRetriable);
```

arangod/Aql/QueryCursor.cpp
```diff
@@ -451,6 +460,13 @@ ExecutionState QueryStreamCursor::writeResult(VPackBuilder& builder) {
   if (hasMore) {
     builder.add("id", VPackValue(std::to_string(id())));
   }
+  if (isRetriable()) {
```
jsteemann, line 463: It looks like this part is identical between QueryResultCursor and QueryStreamCursor. Can it be moved to the shared base class?

arangod/Utils/Cursor.h
```diff
@@ -85,6 +91,18 @@ class Cursor {
 
   void setDeleted() { _isDeleted = true; }
 
+  void setLastQueryBatchObject(velocypack::Builder& builder) {
+    _currentBatchResult.second = builder;
+  }
```
jsteemann, line 96: This API is a bit unintuitive, because we are passing the Builder by mutable reference, but are then copying it anyway.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +678,73 @@ RestStatus RestCursorHandler::createQueryCursor() {
...
+
+  TRI_ASSERT(_cursor != nullptr);
+
+  std::shared_ptr<transaction::Context> ctx = _cursor->context();
+
+  VPackBuilder builder;
+
+  size_t batchId;
+  std::stringstream batchSuffix(suffixes[1]);
+  batchSuffix >> batchId;
```
jsteemann, line 732:
```suggestion
  size_t batchId = basics::StringUtils::uint64(suffixes[1]);
```

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +678,73 @@ RestStatus RestCursorHandler::createQueryCursor() {
...
+  if (!_cursor->isRetriable()) {
+    generateError(rest::ResponseCode::BAD, TRI_ERROR_HTTP_BAD_PARAMETER,
+                  "expecting retriable option to be true");
+    return RestStatus::DONE;
+  }
+
+  _cursor->setWakeupHandler(withLogContext(
+      [self = shared_from_this()]() { return self->wakeupHandler(); }));
+
+  TRI_ASSERT(_cursor != nullptr);
```
jsteemann, line 724: I think this assert can be removed. The pointer was already dereferenced twice before it, so the assert is of no value.

arangod/Utils/Cursor.h
```diff
@@ -145,10 +163,13 @@ class Cursor {
...
   CursorId const _id;
   size_t const _batchSize;
+  size_t _currentBatchId;
   double _ttl;
   std::atomic<double> _expires;
   bool const _hasCount;
   bool _isDeleted;
+  bool _isRetriable;
   std::atomic<bool> _isUsed;
+  std::pair<size_t, VPackBuilder> _currentBatchResult;
```
jsteemann, line 173: I don't like storing a full `Builder` here. All we need is the slice data contained in the Builder, which the Builder stores in its underlying Buffer.
So why don't we just store a Buffer here?

js/client/modules/@arangodb/arango-query-cursor.js
```diff
@@ -219,7 +221,18 @@ ArangoQueryCursor.prototype.next = function () {
...
+
+      try {
+        arangosh.checkRequestResult(requestResult);
+      } catch (err) {
+        if (latestBatchId !== undefined) {
+          requestResult = this._database._connection.POST(this._baseurl() + '/' + encodeURIComponent(latestBatchId), null);
+          arangosh.checkRequestResult(requestResult);
+        } else {
+          // throw the error again for the case in which the request to the latest batch is not retryable
+          throw(err);
```
jsteemann, line 233: `throw` is a keyword, doesn't need parentheses.
```suggestion
          throw err;
```

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -618,10 +623,24 @@ RestStatus RestCursorHandler::generateCursorResult(rest::ResponseCode code) {
...
     builder.add(StaticStrings::Code, VPackValue(static_cast<int>(code)));
     builder.close();
+    _cursor->setLastQueryBatchObject(builder);
 
     _response->setContentType(rest::ContentType::JSON);
+    TRI_IF_FAILURE("MakeConnectionErrorForRetry") { return RestStatus::FAIL; }
+
     generateResult(code, std::move(buffer), std::move(ctx));
   } else {
+    VPackBuilder builder;
```
jsteemann, line 633: We seem to already have a Builder object on the stack here.
Why do we need yet another one?

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -618,10 +623,24 @@ RestStatus RestCursorHandler::generateCursorResult(rest::ResponseCode code) {
     builder.add(StaticStrings::Error, VPackValue(false));
     builder.add(StaticStrings::Code, VPackValue(static_cast<int>(code)));
     builder.close();
+    _cursor->setLastQueryBatchObject(builder);
```
jsteemann, line 626: We should not store anything if the cursor is not retriable. Otherwise we will only add overhead for something the user didn't ask for.

Instead of copying the Builder object here and storing it in the cursor, we should store the Builder's Buffer in the cursor. We can even create a Buffer on heap using `make_shared`, and then make the Builder point to it while building the result. And afterwards simply store the shared_ptr in the cursor.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -618,10 +623,24 @@ RestStatus RestCursorHandler::generateCursorResult(rest::ResponseCode code) {
...
+    VPackBuilder builder;
+    builder.openObject();
+    builder.add(StaticStrings::Code,
+                VPackValue(static_cast<int>(
+                    GeneralResponse::responseCode(r.errorNumber()))));
+    builder.add(StaticStrings::Error, VPackValue(true));
+    builder.add("errorMessage", VPackValue(r.errorMessage()));
+    builder.add("errorNum", VPackValue(r.errorNumber()));
+    builder.close();
+    _cursor->setLastQueryBatchObject(builder);
```
jsteemann, line 642: same here: we shouldn't store a full Builder, but rather a shared_ptr to a Buffer.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +678,73 @@ RestStatus RestCursorHandler::createQueryCursor() {
   return registerQueryOrCursor(body);
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// @brief shows the batch given by <batch-id> if it's the last cached batch
+/// response on a retry, and does't advance cursor
+////////////////////////////////////////////////////////////////////////////////
+
+RestStatus RestCursorHandler::showLastBatchRetry() {
```
jsteemann, line 686: I find the method name hard to understand. `showLastBatch()` I do understand, but the extra `Retry` at the end puzzles me a bit.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+  return {
+    setUpAll: function () {
+      db._drop(cn);
+      let cid = db._create(cn);
+
+      let docs = [];
+      for (let i = 0; i < 2001; i++) {
+        docs.push({"_key": `test${i}`});
+      }
+      db[cn].save(docs);
```
jsteemann, line 1230: `cid` is not used later in this function. So I suggest using it.
```suggestion
      let c = db._create(cn);

      let docs = [];
      for (let i = 0; i < 2001; i++) {
        docs.push({"_key": `test${i}`});
      }
      c.insert(docs);
```

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 201);
+      assertEqual(typeof doc.parsedBody['id'], "string");
+      assertMatch(reId, doc.parsedBody['id']);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      let cursorId = doc.parsedBody['id'];
+      let nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1254: Do we expect `nextBatchId` to contain any value here? Which value? I think we should assert for it.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+
+      assertEqual(doc.code, 200);
+      assertEqual(doc.headers['content-type'], contentType);
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 200);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1277: Are we expecting a certain nextBatchId here? Which one? Let's add an assert!

arangod/Aql/QueryCursor.cpp
```diff
@@ -128,6 +128,13 @@ Result QueryResultCursor::dumpSync(VPackBuilder& builder) {
 
     builder.add("cached", VPackValue(_cached));
 
+    if (isRetriable()) {
+      _currentBatchResult.first = ++_currentBatchId;
+      if (hasNext()) {
+        builder.add("nextBatchId", _currentBatchId + 1);
```
jsteemann, line 134: Cursor ids are strings, but the value of `nextBatchId` is a number. Shall we change the value of `nextBatchId` in the response to be a string, too? That would make it consistent. 
I also think this wouldn't cause any problem, because the client application/driver is not supposed to do any math (e.g. +1, -1) with the nextBatchId, but can only resend the very same batch id to the server.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+    },
+
+    tearDownAll: function () {
+      internal.debugClearFailAt();
+      db._drop(cn);
+    },
+
+    test_cursor_non_stream_request_retriable: function () {
+      let cmd = api;
+      let body = {"query": `FOR u IN ${cn} RETURN u`, "options": {"stream": true, "retriable": true}};
```
jsteemann, line 1240: The name of the test is `non_stream`, but the test performs a streaming query anyway. Looks wrong.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 201);
+      assertEqual(typeof doc.parsedBody['id'], "string");
+      assertMatch(reId, doc.parsedBody['id']);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      let cursorId = doc.parsedBody['id'];
+      let nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1325: What value of `nextBatchId` is expected here?

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,225 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+        if (i === 1900) {
+          internal.debugClearFailAt();
+        }
+        assertEqual("test" + i, nextValue._key);
+        assertEqual(i !== 2000, cursor.hasNext());
+      }
+      assertFalse(cursor.hasNext());
+    },
+  };
+}
```
jsteemann, line 1425: There should also be a test with `retriable: false`, that ensures that no `nextBatchId` is returned with the cursor results, and that a request to fetch the next id actually fails.

CHANGELOG
Simran-B, line 7: Here is a suggestion:

```suggestion
* APM-592: Added a `retriable` flag for AQL queries. If enabled, a `nextBatchId` attribute is
  included in the response object of every cursor batch. Clients can use it to retrieve the
  subsequent result batch without advancing the cursor by using the new
  `POST /_api/cursor/<cursorId>/<batchId>` endpoint. This enables clients to retry receiving
  query results should a transfer fail. The request is idempotent.
```

I'm a bit confused, however. Is the client expected to remember the `nextBatchId` after receiving the first batch, then fetch the second batch normally? If It fails to retrieve the second batch, is uses the remembered ID from the first batch and retries to transfer the second batch? What if the initial batch fails already? And are cursors not destroyed as soon as the client asks for the last batch if the `retriable` option is enabled because the client might ask for that batch again? Does the server keep it around until the cursor TTL is reached?

Simran-B (commented):
As per Google's style guide, the terms `retriable` and `retryable` should be avoided, but if necessary, the latter appears to be recommended: https://developers.google.com/style/word-list#retriable

CHANGELOG
cpjulia: The first batch starts with 1, if you don't retrieve it successfully, then you won't get the `nextBatchId` returned to you. 
As the value of `nextBatchId` is deterministic and based on single increment per batch, then the user can either use the `nextBatchId` value when it's returned in the batch's response object or memorize it. 
The first batch's id value is 1. I'll add it to the documentation. 
Cursors are advanced every time you run `_api/cursor/<cursorId>` successfully and destroyed after the last batch, so you can get the in-between batches starting from the first (only the latest batch, the previous batches are not cached). 
The server keeps the latest batch cached until the next batch's response object is created, then it stores the next batch in place of the former one. The request to `_api/cursor/<cursorId>/<batchId>` doesn't advance the cursor

CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* APM-592: In batched query results, when executing requests for `_api/cursor`, 
+  there might be a connection error and the user might not be able to retrieve
+  the latest batch in `cursorId`. For that, a query option flag `retriable` was
```
```suggestion
* APM-592: In batched query results, when executing requests for `/_api/cursor`, 
  there might be a connection error and the user might not be able to retrieve
  the latest batch from the cursor. For that, a query option flag `retriable` was
```

CHANGELOG
```diff
@@ -1,6 +1,14 @@
...
+  added. When set to `true`, if the latest batch response object wasn't
+  successfully returned, the user can send a retry request to receive it with
```
jsteemann, line 8:
```suggestion
  successfully received, the user can send a retry request to receive it with 
```

CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* APM-592: In batched query results, when executing requests for `_api/cursor`, 
+  there might be a connection error and the user might not be able to retrieve
+  the latest batch in `cursorId`. For that, a query option flag `retriable` was
+  added. When set to `true`, if the latest batch response object wasn't
+  successfully returned, the user can send a retry request to receive it with 
+  a POST request to `_api/cursor/<cursorId>/<batchId>`. Only the latest batch
+  is cached, meaning former batches cannot be retrieved with this request.
```
```suggestion
  is cached, meaning former batches cannot be retrieved again later.
```

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -618,17 +622,30 @@ RestStatus RestCursorHandler::generateCursorResult(rest::ResponseCode code) {
...
+
+    generateResult(code, builder.slice(), std::move(ctx));
   } else {
+    if (_cursor->isRetriable()) {
+      builder.add(StaticStrings::Code,
+                  VPackValue(static_cast<int>(
+                      GeneralResponse::responseCode(r.errorNumber()))));
+      builder.add(StaticStrings::Error, VPackValue(true));
+      builder.add("errorMessage", VPackValue(r.errorMessage()));
+      builder.add("errorNum", VPackValue(r.errorNumber()));
```
jsteemann, line 639: Please use `StaticStrings` here.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +676,68 @@ RestStatus RestCursorHandler::createQueryCursor() {
...
+  _cursor->setWakeupHandler(withLogContext(
+      [self = shared_from_this()]() { return self->wakeupHandler(); }));
+
+  std::shared_ptr<transaction::Context> ctx = _cursor->context();
+
+  VPackBufferUInt8 buffer;
+
+  std::string const& batchId = suffixes[1];
+
+  auto const r = _cursor->getLastBatchResult(batchId, buffer);
```
jsteemann, line 728: Why yet another Buffer here? Can't we return the shared_ptr to the existing Buffer and simply use it?

arangod/Utils/Cursor.h
```diff
@@ -85,6 +92,30 @@ class Cursor {
 
   void setDeleted() { _isDeleted = true; }
 
+  void setLastQueryBatchObject(
+      std::shared_ptr<velocypack::Buffer<uint8_t>> buffer) {
+    _currentBatchResult.second = buffer;
```
jsteemann, line 97: The shared_ptr is passed by value, so it can be moved here. Otherwise there will be 2 copies of it.
```suggestion
    _currentBatchResult.second = std::move(buffer);
```

arangod/Utils/Cursor.h
```diff
@@ -85,6 +92,30 @@ class Cursor {
...
+  }
+
+  Result getLastBatchResult(std::string const& batchId,
+                            VPackBufferUInt8& buffer) {
+    if (_currentBatchResult.first.compare(batchId) != 0) {
```
jsteemann, line 102: Why compare? :thinking: 
```suggestion
    if (_currentBatchResult.first != batchId) {
```

arangod/Utils/Cursor.h
```diff
@@ -85,6 +92,30 @@ class Cursor {
...
+      return Result(TRI_ERROR_HTTP_NOT_FOUND, "batch id not found");
+    }
+    VPackBuilder builder(*(_currentBatchResult.second.get()));
```
jsteemann, line 105: `_currentBatchResult` contains a shared_ptr to the Buffer already. Why not simply return it? Creating another Builder on top is just unnecessary overhead.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertEqual(doc.parsedBody['code'], 201);
+      assertEqual(typeof doc.parsedBody['id'], "string");
+      assertMatch(reId, doc.parsedBody['id']);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      let cursorId = doc.parsedBody['id'];
+      assertEqual(doc.parsedBody['nextBatchId'], 2);
+      let nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1255: Is `nextBatchId` an integer or a string? I think we should also check its type here. 
```suggestion
      let nextBatchId = doc.parsedBody['nextBatchId'];
      assertEqual(nextBatchId, "2");
      
```

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertEqual(doc.code, 200);
+      assertEqual(doc.headers['content-type'], contentType);
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 200);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      assertEqual(doc.parsedBody['nextBatchId'], 3);
+      nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1279:
```suggestion
      nextBatchId = doc.parsedBody['nextBatchId'];
      assertEqual(nextBatchId, "3");      
```

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      doc = arango.POST_RAW(cmd, "");
+
+      assertEqual(doc.code, 200);
+      assertEqual(doc.headers['content-type'], contentType);
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 200);
+      assertFalse(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['id'], undefined);
+      assertEqual(doc.parsedBody['result'].length, 1);
+      assertFalse(doc.parsedBody['cached']);
```
jsteemann, line 1300: Are we expected a `nextBatchId` in the response here`? If not, we should verify that the response does not contain one.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertEqual(doc.parsedBody['code'], 201);
+      assertEqual(typeof doc.parsedBody['id'], "string");
+      assertMatch(reId, doc.parsedBody['id']);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      let cursorId = doc.parsedBody['id'];
+      assertEqual(doc.parsedBody['nextBatchId'], 2);
+      let nextBatchId = doc.parsedBody['nextBatchId'];
```
jsteemann, line 1328: Same comments apply as in previous test.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      doc = arango.POST_RAW(cmd, "");
+
+      assertEqual(doc.code, 200);
+      assertEqual(doc.headers['content-type'], contentType);
+      assertEqual(doc.parsedBody['code'], 200);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      assertEqual(doc.parsedBody['nextBatchId'], 3);
```
jsteemann, line 1351: Same comments apply as in previous test.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+
+      cmd = api + `/${cursorId}`;
+      doc = arango.POST_RAW(cmd, "");
+
+      assertEqual(doc.code, 200);
+      assertEqual(doc.headers['content-type'], contentType);
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 200);
+      assertFalse(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['id'], undefined);
```
jsteemann, line 1371: Same comments apply as in previous test.

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+      assertFalse(doc.parsedBody['error']);
+      assertEqual(doc.parsedBody['code'], 201);
+      assertEqual(typeof doc.parsedBody['id'], "string");
+      assertMatch(reId, doc.parsedBody['id']);
+      assertTrue(doc.parsedBody['hasMore']);
+      assertEqual(doc.parsedBody['result'].length, 1000);
+      assertFalse(doc.parsedBody['cached']);
+
+      let cursorId = doc.parsedBody['id'];
+      assertEqual(doc.parsedBody['nextBatchId'], undefined);
```
jsteemann, line 1400:
```suggestion
      assertUndefined(doc.parsedBody['nextBatchId']);
```

tests/js/client/shell/api/cursor.js
```diff
@@ -1212,9 +1212,268 @@ function dealing_with_cursorsSuite_fetching_floating_point_valuesSuite () {
...
+        if (i === 1900) {
+          internal.debugClearFailAt();
+        }
+        assertEqual("test" + i, nextValue._key);
+        assertEqual(i !== 2000, cursor.hasNext());
+      }
+      assertFalse(cursor.hasNext());
+    },
+  };
+}
```
jsteemann, line 1469: There should also be a test that tries to receive an invalid batch ids, e.g. when the current batch is is 1, it should try to retrieve batch id 2, or when the current id is 2, it should try to retrieve 0. 
Or try to receive batch id `"piff"` or such.

arangod/RestHandler/RestCursorHandler.cpp
cpjulia: You can't because, in the following lines, you might have to use `std::move` on the data pointed to by the shared_ptr, and then, you'll have to use it again and doesn't contain the data anymore:
```
 auto const r = _cursor->getLastBatchResult(batchId, buffer);

  if (r.ok()) {
    _response->setContentType(rest::ContentType::JSON);
    generateResult(rest::ResponseCode::OK, std::move(buffer), std::move(ctx));
  }
```
`generateResult()` expects `Payload &&`, rvalue

tests/js/client/shell/api/cursor.js
cpjulia: Added within the same test functions, but expecting a 404 for getting the `nextBatchId` when it's one batch ahead (not yet fetched) and for batch id 0

CHANGELOG
Simran-B: Shouldn't the last batch remain cached until the cursor TTL is exceeded or the cursor is deleted manually? In the case of streaming cursors, the last batch contains additional important information. If retrieving the last batch fails, you cannot retry and need to re-run the entire query...

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor_identifier.md
```diff
@@ -77,13 +85,11 @@ Valid request for next batch
...
 
-    var body = response.body.replace(/\\/g, '');
-    var _id = JSON.parse(body).id;
-    response = logCurlRequest('POST', url + '/' + _id, '');
+    response = logCurlRequest('POST', url + '/' + response.parsedBody.id, '');
     assert(response.code === 200);
 
     logJsonResponse(response);
-  ~ db._drop(cn);
+    db._drop(cn);
```
jsteemann, line 92: Shouldn't the `~` remain here? I think it indicates that the output of the operation is not part of the generated example, i.e. here the `~` would hide the result of the `db._drop(cn);` operation, which is probably desired.

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor_identifier_batch.md
```diff
@@ -0,0 +1,118 @@
...
+
+@RESTREPLYBODY{code,integer,required,integer}
+The HTTP status code.
+
+@RESTREPLYBODY{errorNum,integer,required,int64}
+A server error number (if `error` is `true`).
+
+@RESTREPLYBODY{errorMessage,string,required,string}
+A descriptive error message (if `error` is `true`).
+
```
jsteemann, line 75: These are actually *optional* attributes and not *required* attributes.
They are only set in case of errors.

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +681,66 @@ RestStatus RestCursorHandler::createQueryCursor() {
...
+
+  std::shared_ptr<transaction::Context> ctx = _cursor->context();
+
+  std::string const& batchId = suffixes[1];
+  auto const [buffer, r] = _cursor->getLastBatchResult(batchId);
+
+  if (r.ok()) {
+    _response->setContentType(rest::ContentType::JSON);
+    generateResult(rest::ResponseCode::OK, VPackSlice(buffer->data()),
+                   std::move(ctx));
```
jsteemann, line 735:
```suggestion
  std::string const& batchId = suffixes[1];
  auto const [buffer, r] = _cursor->getLastBatchResult(batchId);

  if (r.ok()) {
    _response->setContentType(rest::ContentType::JSON);
    generateResult(rest::ResponseCode::OK, VPackSlice(buffer->data()),
                   _cursor->context());
```

arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -659,6 +681,66 @@ RestStatus RestCursorHandler::createQueryCursor() {
...
+  } else {
+    // builder can be in a broken state here. simply return the error
```
jsteemann, line 737:
```suggestion
    // no Buffer available here
```

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor_identifier.md
Simran-B: Nope, that's only the case for `@EXAMPLE_ARANGOSH_OUTPUT` blocks but not `@EXAMPLE_ARANGOSH_RUN` blocks. Here, only the parts that are explicitly logged will be visible (`logCurlRequest()`, `logJsonResponse()`, etc.).

Documentation/DocuBlocks/Rest/Cursors/post_api_cursor_identifier_batch.md
Simran-B: Well, this specific reply body is for the error case, where they are always set. I consider this to be an oddity of our DocuBlocks. I would agree if he had a unified description of the response structure, but then all non-error fields would need to be optional...

jsteemann (approved):
LGTM

Simran-B (approved):

Simran-B (approved):

PR #17887, opened 2023-01-05
Split SHA256Functor and TRI_SlurpDecryptFile out of files-advanced.cpp

markuspf:
### Scope & Purpose

Some gratuitous cleanup that I asked  for on #17864; 

In particular this removes checks for outdated OpenSSL versions in the code for SHA256Functor and hence makes it quite a bit more concise.

Requires enterprise buddy https://github.com/arangodb/enterprise/pull/1188 and is based on #17864


lib/Basics/files-slurp-decrypt.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "Enterprise/Encryption/EncryptionFeature.h"
+
+using namespace arangodb;
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief slurps in a file that is encrypted and return unencrypted contents
+////////////////////////////////////////////////////////////////////////////////
+
+/// @brief read buffer size (used for bulk file reading)
+#define READBUFFER_SIZE 8192
```
jsteemann, line 41: As this define is only used in the following function, can we turn it into a constexpr?

markuspf: Done!

jsteemann (approved):
LGTM

PR #17888, opened 2023-01-05
Remove superflous OPENSSL version guards

markuspf:
### Scope & Purpose

These guards checked whether OpenSSL was version < OpenSSL 1.1.1, which are not supported anymore and should hence always evaluate to `false` anyway.


jsteemann (approved):
LGTM

PR #17889, opened 2023-01-05
[CINFRA-647] Make the Stream throw an exception after resign

goedderz:
### Scope & Purpose

Since #17850, the DocumentFollowerState no longer synchronizes resigning with calling release on the stream. But calls to release on a resigned stream caused a nullptr-deref. This PR changes that, so it throws an exception instead.

- [X] :hankey: Bugfix

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-647

### DIFF
--- a/arangod/Replication2/ReplicatedState/ReplicatedState.tpp
+++ b/arangod/Replication2/ReplicatedState/ReplicatedState.tpp
@@ -141,8 +141,7 @@ void ReplicatedStateManager<S>::leadershipEstablished(
                     guard->_currentManager))
           .resign();
   ADB_PROD_ASSERT(oldMethods == nullptr);
-  auto stream = std::make_shared<
-      ProducerStreamProxy<EntryType, Deserializer, Serializer>>(
+  auto stream = std::make_shared<typename LeaderStateManager<S>::StreamImpl>(
       std::move(methods));
   auto leaderState = _factory->constructLeader(std::move(core));
   // TODO Pass the stream during construction already, and delete the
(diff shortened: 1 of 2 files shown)


arangod/Replication2/ReplicatedState/ReplicatedState.tpp
```diff
@@ -252,10 +251,38 @@ auto ReplicatedStateManager<S>::resign() && -> std::unique_ptr<CoreType> {
...
+}
+
+template<typename S, template<typename> typename Interface,
+         ValidStreamLogMethods ILogMethodsT>
+auto StreamProxy<S, Interface, ILogMethodsT>::waitFor(LogIndex index)
+    -> futures::Future<WaitForResult> {
+  // TODO As far as I can tell right now, we can get rid of this:
+  //      Delete this, also in streams::Stream.
+  return _logMethods->waitFor(index).thenValue(
+      [](auto const&) { return WaitForResult(); });
```
apetenchea, line 280: `waitFor` is used in `DocumentLeaderState::replicateOperation`, in case we have to wait for commit (line 181).

apetenchea (commented):
Apart from `waitFor`, we also use `stream->insert()` in `DocumentLeaderState::replicateOperation`. Should we also throw an exception for them as well?

arangod/Replication2/ReplicatedState/ReplicatedState.tpp
goedderz: > Apart from `waitFor`, we also use `stream->insert()` in `DocumentLeaderState::replicateOperation`. Should we also throw an exception for them as well?

I'm unsure. It's not necessary, as `insert` can not be called after the state resigned, and the stream is purposefully resigned only after the state. So it'd also be plausible to add an assertion that `insert` is not called after being resigned. On the other hand, maybe it's better to have a consistent interface on the stream regarding resign. What's your take on this?

goedderz:
@apetenchea 
> Apart from `waitFor`, we also use `stream->insert()` in `DocumentLeaderState::replicateOperation`. Should we also throw an exception for them as well?

I'm unsure. It's not necessary, as `insert` can not be called after the state resigned, and the stream is purposefully resigned only after the state. So it'd also be plausible to add an assertion that `insert` is not called after being resigned. On the other hand, maybe it's better to have a consistent interface on the stream regarding resign. What's your take on this?

goedderz:
Jenkins passed: https://jenkins01.arangodb.biz/job/rlog-matrix-pr/212/

maierlars (approved):

apetenchea (approved):

PR #17894, opened 2023-01-05
[CINFRA] Some performance improvements and more metrics

maierlars:
### Scope & Purpose

The performance improvements include:
- resolving replications `waitFor` futures multithreaded. Previously they were resolved sequentially. This adds more parallelism.
- Enable blod-db for the replicated log column family. This reduced the compaction debt by a lot and since replicated logs are written to be deleted instead of read, this is a fair trade off.
- We also added more metrics to precisely pin down where the performance is lost. Those new metrics point to writing into rocksdb. As expected, RocksDB is not the best (TM) storage engine for writing replicated logs.


arangod/Replication2/MetricsHelper.h
```diff
@@ -59,7 +58,11 @@ struct GaugeScopedCounter {
   void fire() noexcept;
 
  private:
-  std::shared_ptr<metrics::Gauge<std::uint64_t>> _metric;
+  struct noop {
+    template<typename T>
+    void operator()(T*) {}
+  };
```
MBkkt, line 64: Why? You already delete copy ctor/assign. Why not raw pointer??
Also unique_ptr by default doesn't have trivial abi

MBkkt (commented):
Btw PR sounds good

arangod/Replication2/MetricsHelper.h
maierlars: You are incorrect. Raw pointer don't provide the same move semantic. `fire` should trigger exactly once, and if the type is moved away, it should not trigger, but only if the moved-to object is destroyed. Thus a non-trivial ABI is required.

MBkkt: 1. Your object is already have move only semantics
2. In fire you manually call .reset(), and manually check nullptr.

This code will be completely same (=nullptr instead of .reset) without unique_ptr

MBkkt: > non-trivial ABI is required

Any code where unique_ptr and pointer is equivalent not trivial ABI required

maierlars: No it will not be the same, because the move constructor/assignment operator differs.

MBkkt: Ok, I see now, u use default move, sorry.

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -742,9 +746,9 @@ auto replicated_log::LogLeader::GuardedLeaderData::prepareAppendEntry(
       << ", current commit index = " << _commitIndex
       << ", last acked lci = " << follower->lastAckedLowestIndexToKeep
       << ", current lci = " << _lowestIndexToKeep;
-  if (follower->nextPrevLogIndex == lastAvailableIndex.index &&
-      _commitIndex == follower->lastAckedCommitIndex &&
-      _lowestIndexToKeep == follower->lastAckedLowestIndexToKeep) {
+  if (follower->nextPrevLogIndex == lastAvailableIndex.index) {  // &&
+    //_commitIndex == follower->lastAckedCommitIndex &&
+    //_lowestIndexToKeep == follower->lastAckedLowestIndexToKeep) {
```
goedderz, line 751: IIRC we didn't want to keep this.

arangod/RocksDBEngine/RocksDBPersistedLog.cpp
```diff
@@ -255,7 +272,11 @@ auto RocksDBAsyncLogWriteBatcher::queue(AsyncLogWriteContext& ctx,
         lane._pendingPersistRequests.emplace_back(ctx, std::move(action));
     bool const wantNewThread = lane._activePersistorThreads == 0 ||
                                (lane._pendingPersistRequests.size() > 100 &&
-                                lane._activePersistorThreads < 2);
+                                lane._activePersistorThreads < 5);
```
goedderz, line 275: IIRC this didn't make a difference; do we want to keep the change?

arangod/Replication2/ReplicatedState/ReplicatedState.tpp
```diff
@@ -521,34 +524,62 @@ void FollowerStateManager<S>::handleApplyEntriesResult(arangodb::Result res) {
...
   if (_stream == nullptr) {
     return std::nullopt;
   }
   if (_commitIndex > _lastAppliedIndex and
       not _applyEntriesIndexInFlight.has_value()) {
-    // TODO MeasureTimeGuard rttGuard(_metrics->replicatedStateApplyEntriesRtt);
     auto log = _stream->methods().getLogSnapshot();
-    _applyEntriesIndexInFlight = _commitIndex;
+    _applyEntriesIndexInFlight =
+        std::min(_commitIndex, _lastAppliedIndex + 1000);
```
goedderz, line 553: Maybe add some comment?
```suggestion
    // Apply at most 1000 entries at once, so we have a smoother progression.
    _applyEntriesIndexInFlight =
        std::min(_commitIndex, _lastAppliedIndex + 1000);
```

PR #17895, opened 2023-01-05
SEARCH-323 SEARCH-279 Have a dedicated storage engine snapshot for ArangoSearch index  and aggressive LM

Dronplane:
### Scope & Purpose

SEARCH-279: 
Preparation to fix. Now each ArangoSearch index have a storage snapshot for materializig. This PR adds storage snapshot storage and using this snapshot for Views and for LateMaterialized views. Inverted index would be adressed in a separate PR
Snapshot is currently is still not synced with iresearch commit.

SEARCH-323
LocalDocumentId is not read in view executor. Rather SearchDoc is emitted and LocalDocumentId is actually read in materialize executor.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/IResearch/ViewSnapshot.h
```diff
@@ -43,17 +43,23 @@ class Methods;
 }  // namespace transaction
 namespace iresearch {
 
+using ViewSegment =
+    std::tuple<DataSourceId, irs::sub_reader const*, StorageSnapshot const&>;
```
MBkkt, line 47: struct please!

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1942,7 +1963,7 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
     }
 
     // update reader
-    _dataStore._reader = reader;
+    std::atomic_store(&_dataStore._snapshot, reader);
```
MBkkt, line 1966: I think very unsafe to have _snapshot a public, I think it should be stored and loaded every time by separate function. We can avoid it only under commit mutex, because it also protects writes

arangod/Aql/MaterializeExecutor.h
```diff
@@ -144,6 +172,12 @@ class MaterializeExecutor {
         ReadContext& ctx);
   };
 
+  void fillBuffer(AqlItemBlockInputRange& inputRange);
+  using BufferRecord = std::tuple<iresearch::SearchDoc, LocalDocumentId,
+                                  LogicalCollection const*>;
```
MBkkt, line 177: struct?

arangod/Aql/MaterializeExecutor.h
```diff
@@ -47,11 +48,36 @@ class RegisterInfos;
 template<BlockPassthrough>
 class SingleRowFetcher;
 
+// two storage varians as we need
+// collection name to be sotred only
```
MBkkt, line 52:
```suggestion
// collection name to be stored only
```

arangod/Aql/MaterializeExecutor.h
```diff
@@ -47,11 +48,36 @@ class RegisterInfos;
...
+// when needed.
+class NoCollectionNameHolder {};
+class StringCollectionNameHolder {
+ public:
+  StringCollectionNameHolder(std::string const& name)
+      : _collectionSource(name) {}
+
+ protected:
+  std::string const& _collectionSource;
```
MBkkt, line 61: Maybe string_view better?

arangod/Aql/MaterializeExecutor.h
```diff
@@ -68,11 +94,13 @@ class MaterializerExecutorInfos {
 
   aql::QueryContext& query() const { return _query; }
 
-  T collectionSource() const { return _collectionSource; }
+  template<typename Source = T>
+  std::enable_if_t<std::is_same_v<Source, std::string const&>, Source>
```
MBkkt, line 98: Just assert?

Or even better just move it to public method of base class

arangod/Aql/MaterializeExecutor.h
```diff
@@ -47,11 +48,36 @@ class RegisterInfos;
...
 template<typename T>
-class MaterializerExecutorInfos {
+class MaterializerExecutorInfos
+    : private std::conditional_t<std::is_same_v<T, std::string const&>,
+                                 StringCollectionNameHolder,
+                                 NoCollectionNameHolder> {
  public:
-  MaterializerExecutorInfos(T collectionSource, RegisterId inNmDocId,
-                            RegisterId outDocRegId, aql::QueryContext& query);
+  using Base =
```
MBkkt, line 70: If you write this using out of class you can use it twice instead of duplicate conditional_t

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -118,14 +205,31 @@ arangodb::aql::MaterializeExecutor<T>::produceRows(
...
 
-    // FIXME(gnusi): use rocksdb::DB::MultiGet(...)
-    written =
-        collection->getPhysical()
-            ->read(&_trx,
-                   LocalDocumentId(input.getValue(docRegId).slice().getUInt()),
-                   callback, ReadOwnWrites::no)
-            .ok();
-
+    if constexpr (std::is_same<T, std::string const&>::value) {
```
MBkkt, line 208:
```suggestion
    if constexpr (std::is_same_v<T, std::string const&>) {
```

But better to have flag in base class

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -55,22 +72,93 @@ MaterializeExecutor<T>::ReadContext::copyDocumentCallback(ReadContext& ctx) {
...
-
 template<typename T>
 arangodb::aql::MaterializeExecutor<T>::MaterializeExecutor(
     MaterializeExecutor<T>::Fetcher& /*fetcher*/, Infos& infos)
     : _trx(infos.query().newTrxContext()),
       _readDocumentContext(infos),
       _infos(infos) {}
 
+template<typename T>
+void arangodb::aql::MaterializeExecutor<T>::fillBuffer(
```
MBkkt, line 83:
```suggestion
void MaterializeExecutor<T>::fillBuffer(
```

arangod/Aql/IResearchViewExecutor.h
```diff
@@ -562,7 +547,7 @@ class IResearchViewExecutorBase {
   std::vector<ColumnIterator> _storedValuesReaders;
   std::array<char, arangodb::iresearch::kSearchDocBufSize> _buf;
   bool _isInitialized;
-
+  std::vector<float_t> _singleDocScoreBuffer;
```
MBkkt, line 550:
```suggestion
  std::vector<float> _singleDocScoreBuffer;
```
We don't support in arangodb float_t/double_t. In general I think we can drop it support in iresearch too

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2824,14 +2824,14 @@ ExecutionNode* AsyncNode::clone(ExecutionPlan* plan, bool withDependencies,
 }
 
 namespace {
-const char* MATERIALIZE_NODE_IN_NM_COL_PARAM = "inNmColPtr";
 const char* MATERIALIZE_NODE_IN_NM_DOC_PARAM = "inNmDocId";
 const char* MATERIALIZE_NODE_OUT_VARIABLE_PARAM = "outVariable";
+const char* MATERIALIZE_NODE_MULTI_NODE_PARAM = "multiNode";
```
MBkkt, line 2829:
```suggestion
constexpr std::string_view kMaterializeNodeInNmDocParam = "inNmDocId";
constexpr std::string_view kMaterializeNodeOutVariableParam = "outVariable";
constexpr std::string_view kMaterializeNodeMultiNodeParam = "multiNode";
```

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -1329,6 +1357,34 @@ bool IResearchViewHeapSortExecutor<ExecutionTraits>::fillBufferInternal(
...
 }
 
+template<typename ExecutionTraits>
+bool IResearchViewExecutor<ExecutionTraits>::next() {
+  auto reset = [&] {
+    ++_readerOffset;
+    _currentSegmentPos = 0;
+    _itr.reset();
+    _doc = nullptr;
+  };
```
MBkkt, line 1367: Maybe make it private function method?

MBkkt (commented):
Not finished

arangod/Aql/MaterializeExecutor.cpp
Dronplane: it is done like that already. Just not pushed )) PR is still WIP

MBkkt: Ok)

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -458,6 +458,23 @@ class RocksDBEngine final : public StorageEngine {
   }
 #endif
 
+  class RocksDBSnapshot : public StorageSnapshot {
+   public:
+    RocksDBSnapshot(rocksdb::DB* db) : _snapshot(db) {}
+
+    TRI_voc_tick_t tick() const noexcept {
+      return static_cast<TRI_voc_tick_t>(
```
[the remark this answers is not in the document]
Dronplane: I'll add static assert for future

arangod/Aql/IResearchViewExecutor.h
Dronplane: True. Actually, more correct would be to use 'irs::score_t'. Fixed like this.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -3733,4 +3733,8 @@ bool RocksDBEngine::checkExistingDB(
 
   return dbExisted;
 }
+
+std::shared_ptr<StorageSnapshot> RocksDBEngine::currentSnapshot() {
```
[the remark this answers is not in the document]
Dronplane: No. "currentSnapshot" is interface function. And std::shared_ptr<RocksDBSnapshot> would not be a valid return type. Technically we could make StorageSnaphot and its descendants to have some internal reference counting and indeed return RocksDBSnapshot. But currently I don't think it is necessary.

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -458,6 +458,23 @@ class RocksDBEngine final : public StorageEngine {
...
+          _snapshot.snapshot()->GetSequenceNumber());
+    }
+
+    auto getSnapshot() const { return _snapshot.snapshot(); }
```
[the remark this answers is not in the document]
Dronplane: What that gives us?

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -458,6 +458,23 @@ class RocksDBEngine final : public StorageEngine {
   }
 #endif
 
+  class RocksDBSnapshot : public StorageSnapshot {
```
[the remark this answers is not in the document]
Dronplane: Currently yes. I'll add it

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -969,6 +969,25 @@ bool RocksDBCollection::lookupRevision(transaction::Methods* trx,
...
+    StorageSnapshot const& snapshot) const {
+  ::ReadTimeTracker timeTracker(
+      _statistics._readWriteMetrics,
+      [](TransactionStatistics::ReadWriteMetrics& metrics,
+         float time) noexcept { metrics.rocksdb_read_sec.count(time); });
+
+  if (!token.isSet()) {
+    return Result{TRI_ERROR_ARANGO_DOCUMENT_KEY_BAD,
+                  "invalid local document id"};
+  }
```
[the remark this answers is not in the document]
Dronplane: I just follow the behaviour we have in the "read" function. Why we should have something different here?

arangod/RocksDBEngine/Methods/RocksDBSstFileMethods.h
```diff
@@ -55,6 +55,13 @@ class RocksDBSstFileMethods final : public RocksDBMethods {
 
   ~RocksDBSstFileMethods();
 
+  rocksdb::Status GetFromSnapshot(rocksdb::ColumnFamilyHandle*,
+                                  rocksdb::Slice const&,
+                                  rocksdb::PinnableSlice*, ReadOwnWrites,
+                                  rocksdb::Snapshot const*) override {
```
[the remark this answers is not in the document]
Dronplane: Class RocksDBSstFileMethods is final by itself

arangod/RocksDBEngine/Methods/RocksDBBatchedWithIndexMethods.h
```diff
@@ -39,6 +39,10 @@ class RocksDBBatchedWithIndexMethods final : public RocksDBMethods {
   RocksDBBatchedWithIndexMethods(rocksdb::TransactionDB* db,
                                  rocksdb::WriteBatchWithIndex*);
 
+  rocksdb::Status GetFromSnapshot(rocksdb::ColumnFamilyHandle*,
+                                  rocksdb::Slice const&,
+                                  rocksdb::PinnableSlice*, ReadOwnWrites,
+                                  rocksdb::Snapshot const*) override;
```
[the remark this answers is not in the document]
Dronplane: class is final by itself

arangod/RocksDBEngine/Methods/RocksDBBatchedMethods.h
```diff
@@ -32,6 +32,13 @@ class RocksDBBatchedMethods final : public RocksDBMethods {
  public:
   explicit RocksDBBatchedMethods(rocksdb::WriteBatch*);
 
+  rocksdb::Status GetFromSnapshot(rocksdb::ColumnFamilyHandle*,
+                                  rocksdb::Slice const&,
+                                  rocksdb::PinnableSlice*, ReadOwnWrites,
+                                  rocksdb::Snapshot const*) override {
```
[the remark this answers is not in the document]
Dronplane: class is final by itself

arangod/IResearch/ViewSnapshot.h
Dronplane: Why? )

arangod/IResearch/SearchDoc.h
```diff
@@ -27,16 +27,26 @@
...
 #include "Aql/AqlValue.h"
+#include "VocBase/Identifiers/DataSourceId.h"
 #include "utils/type_limits.hpp"
 
-namespace iresearch {
+namespace irs {
 
 struct sub_reader;
 
 }
```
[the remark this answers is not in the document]
Dronplane: But compiler thinks it is still needed )

arangod/IResearch/IResearchDataStore.cpp
Dronplane: It is protected member, not intended to be used from outside. So what is unsafe here?

arangod/Aql/MaterializeExecutor.h
Dronplane: We anyway have a string reference here (it comes from the collection"name() call). And there is no way to have anything else. So let it be a reference for now.

arangod/IResearch/IResearchDataStore.cpp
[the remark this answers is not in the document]
Dronplane: Could you please elaborate more?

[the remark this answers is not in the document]
MBkkt: I think in all this functions should be something like atomic_load/store_explicit(..., memory_order_acquire/release)

arangod/IResearch/SearchDoc.h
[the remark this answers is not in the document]
MBkkt: Ok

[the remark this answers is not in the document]
MBkkt: It's just not understandable how it works before your change

arangod/IResearch/IResearchDataStore.cpp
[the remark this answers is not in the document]
Dronplane: Good point. Let's do it

arangod/Aql/MaterializeExecutor.h
MBkkt: Ok

arangod/IResearch/IResearchDataStore.cpp
MBkkt: It's probably not for this PR, but in general I believe we shouldn't have access to this members

MBkkt: unsafe because you need to think where is safe to touch part of snapshot, and where is it unsafe.

I think everywhere except commit should use 
_snapshot.load() -> shared_ptr<Data>

And commit should use .store(Data)

Un noone shouldn't have access to _snapshot members

Dronplane: That looks fearsible. The only place where we read only reader part  is metrics. But it would be no harm to load entire _snapshot. And that might be useful in future if we will include some data from storage snapshot into metrics

arangod/RocksDBEngine/Methods/RocksDBBatchedMethods.h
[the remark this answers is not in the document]
MBkkt: :( writing always final is simpler!
Also it simpler for reader, and easy to make class not final if sometime will be needed.

override very special keyword needed only for virtual function which is not final.

arangod/IResearch/ViewSnapshot.h
MBkkt: Difficult to understand what's going ok, of course it's not like we cannot merge with tuple, but without tuple it's better.

Just see on your own code like get<0>/etc after few weeks

Do you agree?

arangod/RocksDBEngine/RocksDBEngine.h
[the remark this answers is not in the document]
MBkkt: Your function just using for internal field function.

Suggested option makes it clear.

If it will be just auto, if return type changed to reference here will be not intentional copy

arangod/IResearch/ViewSnapshot.h
Dronplane: No. It`s quite simple construct, But if you strongly insist I could make it a struct )

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -458,6 +458,23 @@ class RocksDBEngine final : public StorageEngine {
...
+   public:
+    RocksDBSnapshot(rocksdb::DB* db) : _snapshot(db) {}
+
+    TRI_voc_tick_t tick() const noexcept {
```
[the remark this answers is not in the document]
MBkkt: At least override

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2824,14 +2824,14 @@ ExecutionNode* AsyncNode::clone(ExecutionPlan* plan, bool withDependencies,
...
-const char* MATERIALIZE_NODE_IN_NM_COL_PARAM = "inNmColPtr";
-const char* MATERIALIZE_NODE_IN_NM_DOC_PARAM = "inNmDocId";
-const char* MATERIALIZE_NODE_OUT_VARIABLE_PARAM = "outVariable";
+constexpr std::string_view MATERIALIZE_NODE_IN_NM_DOC_PARAM = "inNmDocId";
+constexpr std::string_view MATERIALIZE_NODE_OUT_VARIABLE_PARAM = "outVariable";
+constexpr std::string_view MATERIALIZE_NODE_MULTI_NODE_PARAM = "multiNode";
 }  // namespace
 
 MaterializeNode* materialize::createMaterializeNode(
     ExecutionPlan* plan, arangodb::velocypack::Slice const& base) {
```
MBkkt, line 2833: slice is just a pointer
```suggestion
    ExecutionPlan* plan, arangodb::velocypack::Slice base) {
```

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -1083,24 +1082,23 @@ IResearchViewNode::IResearchViewNode(aql::ExecutionPlan& plan,
...
   auto const viewNameSlice = base.get(NODE_VIEW_NAME_PARAM);
   auto const viewName =
       viewNameSlice.isNone() ? "" : viewNameSlice.stringView();
   auto const viewIdSlice = base.get(NODE_VIEW_ID_PARAM);
+
+  if (base.hasKey("outNmColPtr")) {
+    // Old coordinator tries to run query on
+    // new dbserver. Registers are not compatible.
+    // We must abort.
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_BAD_PARAMETER,
```
MBkkt, line 1098: Please add TODO about this condition can be removed after support last version with it (probably any 3.9 or <= 3.10.2) will be dropped

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -1885,7 +1862,7 @@ std::unique_ptr<aql::ExecutionBlock> IResearchViewNode::createBlock(
     TRI_ASSERT(_view || _meta);
     auto executorInfos = aql::IResearchViewExecutorInfos{
         std::move(reader),
-        std::move(outRegister),
+        outRegister,
```
MBkkt, line 1865: Is it just int? If not maybe better to remove const and keep moving

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -56,52 +62,158 @@ MaterializeExecutor<T>::ReadContext::copyDocumentCallback(ReadContext& ctx) {
...
+        if (ADB_LIKELY(pkReader)) {
+          docValue = irs::get<irs::payload>(*pkReader);
+        } else {
+          // skip segment without PK column
+          do {
+            ++doc;
+          } while (doc != end &&
+                   lastSegment == std::get<0>(_bufferedDocs[*doc]).segment());
+          continue;
+        }
```
MBkkt, line 173: Maybe?
```suggestion
        if (ADB_UNLIKELY(pkReader == nullptr)) {
          // skip segment without PK column
          do {
            ++doc;
          } while (doc != end &&
                   lastSegment == std::get<0>(_bufferedDocs[*doc]).segment());
          continue;
        }
        docValue = irs::get<irs::payload>(*pkReader);
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1942,7 +1984,7 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
     }
 
     // update reader
-    _dataStore._reader = reader;
+    _dataStore.storeSnapshot(reader);
```
MBkkt, line 1987:
```suggestion
    _dataStore.storeSnapshot(std::move(reader));
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1930,7 +1966,13 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
...
-    auto reader = _dataStore._reader.reopen(_format);
+    auto oldSnapshot = _dataStore.loadSnapshot();
+    // we reuse storage snapshot. Technikally this is not
+    // right as old storage snapshot is most likely outdated.
+    // but index is empty so it makes no sense as we will not
+    // materialize anything anyway.
+    auto reader = std::make_shared<DataSnapshot>(
+        oldSnapshot->_reader.reopen(_format), oldSnapshot->_snapshot);
 
     if (!reader) {
```
MBkkt, line 1977: maybe make reader still reader and move make_shared to the end?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1930,7 +1966,13 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
     _lastCommittedTickTwo = _lastCommittedTickOne;
 
     // get new reader
-    auto reader = _dataStore._reader.reopen(_format);
+    auto oldSnapshot = _dataStore.loadSnapshot();
+    // we reuse storage snapshot. Technikally this is not
+    // right as old storage snapshot is most likely outdated.
+    // but index is empty so it makes no sense as we will not
+    // materialize anything anyway.
```
MBkkt, line 1973: Maybe just nullptr in such case?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1306,11 +1322,21 @@ Result IResearchDataStore::initDataStore(
...
-                                                       nullptr, readerOptions);
+      auto engineSnapshot = _engine->currentSnapshot();
+      if (ADB_UNLIKELY(!engineSnapshot)) {
+        return {TRI_ERROR_INTERNAL,
+                absl::StrCat("failed to get storage snapshot while "
+                             "initializing ArangoSearch index '",
+                             index().id().id(), "'")};
+      }
+      _dataStore.storeSnapshot(std::make_shared<DataSnapshot>(
+          irs::directory_reader::open(*(_dataStore._directory), nullptr,
```
MBkkt, line 1333: open returned value also probably should be checked

MBkkt, line 1333: instead of `nullptr` probably should be `_format`

I think I missed it

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1425,13 +1451,23 @@ Result IResearchDataStore::initDataStore(
...
+    }
+    _dataStore.storeSnapshot(std::make_shared<DataSnapshot>(
+        irs::directory_reader::open(*(_dataStore._directory), _format,
+                                    readerOptions),
+        engineSnapshot));
   }
 
-  if (!_dataStore._reader) {
+  auto snapshot = _dataStore.loadSnapshot();
+  if (!snapshot) {
```
MBkkt, line 1470: I think in DataSnapshot ctor should be TRI_ASSERT(reader != nullptr);
Otherwise checks like this incorrect

And we should store nullptr to storeSnapshot or use old

MBkkt (changes requested):
load/storeSnapshot usage in IResearchDataStore looks broken for me.

Maybe it's too late and I just missing something, let's discuss tomorrow

arangod/Aql/IResearchViewNode.cpp
Dronplane: I guess we do not prevent upgrading from EOLed versions. So this code should stay.

arangod/Aql/IResearchViewNode.cpp
MBkkt: Yep

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1982,7 +2024,7 @@ IResearchDataStore::Stats IResearchDataStore::stats() const {
 IResearchDataStore::Stats IResearchDataStore::updateStatsUnsafe() const {
   TRI_ASSERT(_dataStore);
   // copy of 'reader' is important to hold reference to the current snapshot
-  auto reader = _dataStore._reader;
+  auto reader = _dataStore.loadSnapshot()->_reader;
   if (!reader) {
```
MBkkt, line 2028:
```suggestion
  auto snapshot = _dataStore.loadSnapshot();
  if (!snapshot) {
```

arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.cpp
```diff
@@ -469,3 +470,14 @@ Result RocksDBTrxBaseMethods::doCommitImpl() {
...
 }
+
+rocksdb::Status RocksDBTrxBaseMethods::GetFromSnapshot(
+    rocksdb::ColumnFamilyHandle* family, rocksdb::Slice const& slice,
+    rocksdb::PinnableSlice* pinnable, ReadOwnWrites rw,
+    rocksdb::Snapshot const* snapshot) {
+  auto oldSnapshot = _readOptions.snapshot;
+  auto restoreSnapshot = absl::Cleanup{
+      [oldSnapshot, this]() { _readOptions.snapshot = oldSnapshot; }};
+  _readOptions.snapshot = snapshot;
```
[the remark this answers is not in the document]
gnusi: I'd discuss possibility to our own dedicated read-only transaction (sub-transaction) for search needs.

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2824,14 +2824,14 @@ ExecutionNode* AsyncNode::clone(ExecutionPlan* plan, bool withDependencies,
...
-const char* MATERIALIZE_NODE_OUT_VARIABLE_PARAM = "outVariable";
+constexpr std::string_view MATERIALIZE_NODE_IN_NM_DOC_PARAM = "inNmDocId";
+constexpr std::string_view MATERIALIZE_NODE_OUT_VARIABLE_PARAM = "outVariable";
+constexpr std::string_view MATERIALIZE_NODE_MULTI_NODE_PARAM = "multiNode";
 }  // namespace
 
 MaterializeNode* materialize::createMaterializeNode(
-    ExecutionPlan* plan, arangodb::velocypack::Slice const& base) {
-  if (base.hasKey(MATERIALIZE_NODE_IN_NM_COL_PARAM)) {
+    ExecutionPlan* plan, arangodb::velocypack::Slice const base) {
```
jsteemann, line 2833: If this code runs on DB servers during upgrade, the following may happen:
* the coordinator is running version 3.10 and sends AQL snippet data containing the "inNmColPtr" key in order to create a MaterializeMultiNode.
* the DB server is running version 3.11 already and does not interpret the `inNmColPtr` key anymore, and thus creates a MaterializeSingleNode.

Dronplane: This is handled  in arangod/Aql/IResearchViewNode.cpp:1124   - We will fail View enumeration node creation.

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2824,14 +2824,14 @@ ExecutionNode* AsyncNode::clone(ExecutionPlan* plan, bool withDependencies,
...
+  if (base.hasKey(MATERIALIZE_NODE_MULTI_NODE_PARAM)) {
```
jsteemann, line 2834: Nit: here we only check if the key is present, but we don't check its value.
It is at least theoretically possible that in the future someone changes the code so we will send `{ "multiNode": false }`, and will wrongly interpret it and creating a multi node.

arangod/Aql/IResearchViewExecutor.h
```diff
@@ -322,8 +319,6 @@ class IndexReadBuffer {
       if (newMemoryUsage != tracked) {
         if (newMemoryUsage > tracked) {
           _memoryTracker.increase(newMemoryUsage - tracked);
-        } else {
```
jsteemann, line 325: If this branch cannot be reached, we should change the nested if conditions from
```
if (newMemoryUsage != tracked) {
  if (newMemoryUsage > tracked) {
```
to just
```
if (newMemoryUsage > tracked) {
```

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -681,28 +681,34 @@ ViewSnapshotPtr snapshotSingleServer(IResearchViewNode const& node,
     if (options.forceSync) {
       syncViewSnapshot(*snapshot, view.name());
     }
-    if (options.restrictSources) {
-      return std::make_shared<ViewSnapshotView>(*snapshot, options.sources);
-    }
+    // if (options.restrictSources) {
+    //  return std::make_shared<ViewSnapshotView>(*snapshot, options.sources);
+    //}
```
jsteemann, line 686: Still needed?

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -681,28 +681,34 @@ ViewSnapshotPtr snapshotSingleServer(IResearchViewNode const& node,
...
     return {ViewSnapshotPtr{}, snapshot};
   }
-  if (!options.forceSync && trx.isMainTransaction()) {
+
+  // TODO: remove here restrincSources condition when switching back to
```
jsteemann, line 690:
```suggestion
  // TODO: remove here restrictSources condition when switching back to
```

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -681,28 +681,34 @@ ViewSnapshotPtr snapshotSingleServer(IResearchViewNode const& node,
...
     }
   }();
   snapshot = makeViewSnapshot(trx, key, options.forceSync, view.name(),
                               std::move(links));
-  if (options.restrictSources && snapshot != nullptr) {
-    return std::make_shared<ViewSnapshotView>(*snapshot, options.sources);
-  }
+  // if (options.restrictSources && snapshot != nullptr) {
+  //  return std::make_shared<ViewSnapshotView>(*snapshot, options.sources);
+  //}
```
jsteemann, line 711: Still needed?

arangod/Aql/IResearchViewNode.cpp
```diff
@@ -1333,19 +1361,18 @@ std::pair<bool, bool> IResearchViewNode::volatility(
...
-    return (_options.restrictSources || _view == nullptr)
-               ? static_cast<void const*>(this)
-               : static_cast<void const*>(_view.get());
-  } else if (ServerState::instance()->isSingleServer()) {
-    return _view.get();
-  }
-
-  TRI_ASSERT(false);
-  return nullptr;
+  // if (ServerState::instance()->isDBServer()) {
```
jsteemann, line 1364: Old code fragements still needed here?

arangod/Aql/IResearchViewNode.cpp
Dronplane: We want to restore Snapshot "views" in future PRs. Just now it is broken :( , so we disabled this code temporary

arangod/Aql/IResearchViewNode.cpp
Dronplane: We want to restore Snapshot "views" in future PRs. Just now it is broken :( , so we disabled this code temporary

arangod/Aql/IResearchViewNode.cpp
Dronplane: We want to restore Snapshot "views" in future PRs. Just now it is broken :( , so we disabled this code temporary

arangod/Aql/IResearchViewExecutor.cpp
```diff
@@ -2101,83 +2071,54 @@ template class ::arangodb::aql::IResearchViewMergeExecutor<ExecutionTraits<
...
 template class ::arangodb::aql::IResearchViewHeapSortExecutor<ExecutionTraits<
     true, false, false,
     MaterializeType::NotMaterialize | MaterializeType::UseStoredValues>>;
-template class ::arangodb::aql::IResearchViewHeapSortExecutor<ExecutionTraits<
-    true, false, false,
-    MaterializeType::LateMaterialize | MaterializeType::UseStoredValues>>;
+// template class
+// ::arangodb::aql::IResearchViewHeapSortExecutor<ExecutionTraits<
+//    true, false, false,
+//    MaterializeType::LateMaterialize | MaterializeType::UseStoredValues>>;
```
jsteemann, line 2095: Still needed?

Dronplane: No. Leftovers.

arangod/Aql/ExecutionNode.cpp
Dronplane: True. Fixed

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -56,52 +62,158 @@ MaterializeExecutor<T>::ReadContext::copyDocumentCallback(ReadContext& ctx) {
...
+          _bufferedDocs.push_back(
+              std::make_tuple(searchDoc, LocalDocumentId{}, lastCollection));
+        }
+      }
+    };
+
+    if (block->hasShadowRows()) {
+      readInputDocs.template operator()<false>();
+    } else {
+      readInputDocs.template operator()<true>();
```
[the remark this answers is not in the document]
Dronplane: 👍

arangod/Aql/IResearchViewExecutor.h
Dronplane: Fixed

arangod/Aql/MaterializeExecutor.h
```diff
@@ -47,11 +49,35 @@ class RegisterInfos;
 template<BlockPassthrough>
 class SingleRowFetcher;
 
+// two storage varians as we need
```
jsteemann, line 52:
```suggestion
// two storage variants as we need
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -999,6 +1007,9 @@ Result IResearchDataStore::commitUnsafeImpl(
       _lastCommittedTickOne = lastTickBeforeCommitOne;
       _lastCommittedTickTwo = lastTickBeforeCommitOne;
       impl.tick(_lastCommittedTickOne);
+      auto reader = _dataStore.loadSnapshot()->_reader;
+      _dataStore.storeSnapshot(
+          std::make_shared<DataSnapshot>(std::move(reader), engineSnapshot));
```
jsteemann, line 1012: If this throws an exception, are things then left in a bad state (e.g. `_lastCommittedTickTwo` was already advanced before)?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1064,8 +1075,26 @@ Result IResearchDataStore::commitUnsafeImpl(
...
 
     // update reader
-    TRI_ASSERT(_dataStore._reader != reader);
-    _dataStore._reader = reader;
+    TRI_ASSERT(_dataStore.loadSnapshot()->_reader != reader);
+    auto const readerSize = reader->size();
+    auto const docsCount = reader->docs_count();
+    auto const liveDocsCount = reader->live_docs_count();
+
+    // For now we want to retain old behaviour that
```
jsteemann, line 1083: AE
```suggestion
    // For now we want to retain old behavior that
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1064,8 +1075,26 @@ Result IResearchDataStore::commitUnsafeImpl(
...
+    // storage snapshot is not older than iresearch snapshot
+    // TODO: Remove this as soon as index_writer will start accepting limiting
+    // tick for commit
+    auto engineSnapshotAfterCommit = _engine->currentSnapshot();
+    if (ADB_UNLIKELY(!engineSnapshotAfterCommit)) {
+      return {
+          TRI_ERROR_INTERNAL,
+          absl::StrCat("Failed to get engine snapshot while finishig commit "
```
jsteemann, line 1091:
```suggestion
          absl::StrCat("Failed to get engine snapshot while finishing commit "
```

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1064,8 +1075,26 @@ Result IResearchDataStore::commitUnsafeImpl(
...
+                       "ArangoSearch index '",
+                       index().id().id(), "'")};
+    }
+
+    _dataStore.storeSnapshot(std::make_shared<DataSnapshot>(
```
jsteemann, line 1096: This can also throw. Do we need to roll back then?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1908,20 +1946,18 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
     _lastCommittedTickOne = std::max(tick, _lastCommittedTickOne);
     _lastCommittedTickTwo = _lastCommittedTickOne;
 
+    auto oldSnapshot = _dataStore.loadSnapshot();
+    // we reuse storage snapshot. Technikally this is not
```
jsteemann, line 1950:
```suggestion
    // we reuse storage snapshot. Technically this is not
```

arangod/IResearch/IResearchDataStore.cpp
Dronplane: I guess no. In the worst case we will have outdated snapshot like commit has failed. And will try to update  it as soon as next changes will arrive.

arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -448,6 +448,22 @@ class RocksDBEngine final : public StorageEngine {
...
+    explicit RocksDBSnapshot(rocksdb::DB& db) noexcept : _snapshot(&db) {}
+
+    TRI_voc_tick_t tick() const noexcept override {
+      return _snapshot.snapshot()->GetSequenceNumber();
+    }
+
+    decltype(auto) getSnapshot() const { return _snapshot.snapshot(); }
+
+   private:
+    mutable rocksdb::ManagedSnapshot _snapshot;
```
jsteemann, line 462: As we are using `rocksdb::ManagedSnapshot` here now, we should add an explicit include at the top of the file to include `<rocksdb/snapshot.h>`.

arangod/StorageEngine/StorageEngine.h
```diff
@@ -88,6 +88,16 @@ struct Options;
 
 }  // namespace transaction
 
+class StorageSnapshot {
+ public:
+  StorageSnapshot() = default;
+  StorageSnapshot(const StorageSnapshot&) = delete;
+  StorageSnapshot& operator=(const StorageSnapshot&) = delete;
```
jsteemann, line 95: Following our convention, it should rather be:
```suggestion
  StorageSnapshot(StorageSnapshot const&) = delete;
  StorageSnapshot& operator=(StorageSnapshot const&) = delete;
```

arangod/IResearch/IResearchDataStore.cpp
Dronplane: Worst case - we will have old snapshot in the link. Will try to update it next time. Index will still be consistent, just more "eventual"

MBkkt (approved):

gnusi (approved):

PR #17897, opened 2023-01-06
improve query plan serialization time

jsteemann:
### Scope & Purpose

Improve query plan serialization time

* Added new stage "instantiating executors" to the query profiling output. The time spent in "instantiating executors" is the time needed to create the query executors from the final query execution time. In cluster mode, this stage also includes the time needed for physically distributing the query snippets to the participating database servers. Previously, the time spent for instantiating executors and the physical distribution was contained in the "optimizing plan" stage, which was misleading.

* Remove constant values for query variables from query plan serialization in cases they were not needed. Previously, constant values of query variables were always serialized for all occurrences of a variable in a query plan. If the constant values were large, this contributed to higher serialization and thus query setup times. Now the constant values are only serialized for relevant parts of query execution plans.

The relevant change in this PR that has the largest effect on performance is that the `toVelocyPack(...)` method of `aql::Variable` now doesn't include constant values anymore. There is new overload of that method that will also include constant values, but it has to be called explicitly. This is now done only in a few dedicated places. All other places that call `toVelocyPack(...)` will call the method that doesn't serialize the constant values.
The rest of this PR contains a bit of cleanup, which is not relevant for the performance improvement.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17900
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,22 @@
 devel
 -----
 
+* Added new stage "instantiating executors" to the query profiling output.
+  The time spent in "instantiating executors" is the time needed to create
+  the query executors from the final query execution time. In cluster mode,
+  this stage also includes the time needed for physically distributing the 
+  query snippets to the participating database servers.
+  Previously, the time spent for instantiating executors and the physical
+  distribution was contained in the "optimizing plan" stage, which was
+  misleading.
+
+* Remove constant values for query variables from query plan serialization 
+  in cases they were not needed. Previously, constant values of query variables
+  were always serialized for all occurrences of a variable in a query plan.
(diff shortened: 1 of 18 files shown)


maierlars (approved):
LGTM

PR #17901, opened 2023-01-06
[CINFRA-620] Fixing remove operations and adding extra tests

apetenchea:
### Scope & Purpose

Two smaller tasks are handled in one PR, because one depends on the other:
1. [CINFRA-620](https://arangodb.atlassian.net/browse/CINFRA-620) - remove operations are normal to fail with `DOCUMENT_NOT_FOUND`, when applied on followers or during leader recovery. We treat this case in the exact same way we treat `UNIQUE_CONSTRAINT_VIOLATED` during an insert operation.
2. [CINFRA-615](https://arangodb.atlassian.net/browse/CINFRA-615) - more tests have been added, mainly for `DocumentFollowerState::applyEntries`.

While I was adding the tests, I realized there was a bug in `applyEntries`. We did not calculate the release index correctly in case of commit/abort operations. The release index should be set as the lowest index of an already completed transaction, not the current index. For `kAbortAllOngoingTrx`, `_activeTransactions.getReleaseIndex()` will return the current index, after all active transactions have been cleared.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**

[CINFRA-620]: https://arangodb.atlassian.net/browse/CINFRA-620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CINFRA-615]: https://arangodb.atlassian.net/browse/CINFRA-615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -100,7 +100,7 @@ auto DocumentFollowerState::applyEntries(
         }
 
         return basics::catchToResultT([&]() -> std::optional<LogIndex> {
-          auto releaseIndex = std::make_optional<LogIndex>();
+          std::optional<LogIndex> releaseIndex{std::nullopt};
```
maierlars, line 103: `std::optional<T>` default constructs as `std::nullopt`.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -115,17 +115,13 @@ auto DocumentFollowerState::applyEntries(
...
+                  self->_activeTransactions.getReleaseIndex(entry->first);
             } else if (doc.operation == OperationType::kCommit ||
                        doc.operation == OperationType::kAbort) {
               self->_activeTransactions.erase(doc.tid);
-              if (!releaseIndex.has_value() ||
-                  releaseIndex.value() < entry->first) {
-                releaseIndex = entry->first;
-              }
+              releaseIndex =
+                  self->_activeTransactions.getReleaseIndex(entry->first);
```
maierlars, line 124: the old code computed the maximum of some values. I assume that `getReleaseIndex` returns a monotonically increasing index?

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -135,6 +144,7 @@ auto DocumentStateTransactionHandler::applyEntry(DocumentLogEntry doc)
       }
       case OperationType::kAbortAllOngoingTrx:
         TRI_ASSERT(false);  // should never happen as it should be handled above
+        break;
```
maierlars, line 147: instead maybe do an abort?

maierlars (approved):

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
apetenchea: Yes, it returns a monotonically increasing index. The difference is that `getReleaseIndex` makes sure that the returned index does not split an active transaction in two halves. In other words, it returns the greatest index so far, that is less than the first occurrence index of any *active* transaction.

PR #17903, opened 2023-01-09
[CINFRA-686] Recreate states on start up.

maierlars:
### Scope & Purpose
Before this PR existing states were not recreated (in fact, we would just crash). However, now they are re-instantiated with the correct type and parameters.

Two additional things needed to be changed:

- The DocumentStateFeature now registers the document state machine during `prepare()` already, so it is available when the DatabaseFeature loads the databases (and meanwhile the replicated states)
- The RebootTracker is no longer used to rebuild participants when a server's own rebootId changes, as it is not yet available when the replicated states are loaded during startup. Instead, the maintenance is used to compare the rebootIds of the replicated logs with the server's current one.

Related to CINFRA-650.

See https://arangodb.atlassian.net/browse/CINFRA-686


apetenchea (approved):
LGTM

PR #17904, opened 2023-01-09
[cmake] Ignore user presets

jvolmer:
Adds cmake user presets file to gitignore - Each user should be able to define his/her own presets.

Links zkd tests library into arangodbtests instead of adding test files manually to arangodbtests.


markuspf (approved):
LGTM

PR #17905, opened 2023-01-09
Make AqlValue::getTypeString better

MBkkt:
### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/AqlValue.cpp
+++ b/arangod/Aql/AqlValue.cpp
@@ -334,21 +334,46 @@ bool AqlValue::isArray() const noexcept {
   return true;
 }
 
-char const* AqlValue::getTypeString() const noexcept {
-  if (isNone()) {
-    return "none";
-  } else if (isNull(true)) {
-    return "null";
-  } else if (isBoolean()) {
-    return "bool";
-  } else if (isNumber()) {
-    return "number";
-  } else if (isString()) {
-    return "string";
-  } else if (isObject()) {
-    return "object";
-  } else if (isArray()) {
-    return "array";
+std::string_view AqlValue::getTypeString() const noexcept {
+  auto get = [](auto t) -> std::string_view {
+    switch (t) {
+      case velocypack::ValueType::Null:
+        return "null";
+      case velocypack::ValueType::Bool:
+        return "bool";
+      case velocypack::ValueType::Array:
+        return "array";
+      case velocypack::ValueType::Object:
+        return "object";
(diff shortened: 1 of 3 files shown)


Dronplane (approved):
LGTM

arangod/Aql/AqlValue.cpp
```diff
@@ -334,21 +334,46 @@ bool AqlValue::isArray() const noexcept {
...
+        return "object";
+      case velocypack::ValueType::Double:
+      case velocypack::ValueType::Int:
+      case velocypack::ValueType::UInt:
+      case velocypack::ValueType::SmallInt:
+        return "number";
+      case velocypack::ValueType::String:
+        return "string";
+      default:
+        return "none";
```
jsteemann, line 356: The behavior now is not equivalent to the previous behavior before the change.
Note that before this change, `"null"` was returned even for None values, because of the following code part:
```
} else if (isNull(true)) {
    return "null";
...
```
`isNull(true)` returns `true` for real Null values, but also for None values.

jsteemann (changes requested):
The PR slightly changes the behavior for None values. This should be fixed.

arangod/Aql/AqlValue.cpp
MBkkt: Yes, but it already checked none above, so no difference?

jsteemann: Turns out that my comment does not matter. The old code checked for `Slice::isNone()` first and then returned `"none"` anyway.

jsteemann (approved):
LGTM

markuspf:
I would be nice if the PR description described what is done more in detail.

If you add a test for the function you changed that would in future catch changes that change behaviour of the function more easily.

MBkkt:
@markuspf 

> I would be nice if the PR description described what is done more in detail.

In general it's just 1-2 switch instead of 1-7. And I thought code pretty easy to understand. If not feel free to ask, it's not difficult to answer for me.

> If you add a test for the function you changed that would in future catch changes that change behaviour of the function more easily.

Tests already exist, it's used in aql typename function.

PR #17906, opened 2023-01-09
[CINFRA-645] acquireSnapshot could be called concurrently

apetenchea:
### Scope & Purpose

The user may remove and add the server again. The leader might do a compaction which the follower won't notice if the process is stopped. From the follower's perspective, the leader compacted something without waiting for itself. Hence, a new snapshot is required. This can happen so quick, that one snapshot transfer is not yet completed before another one is requested.
Before we start transferring the snapshot, we truncate the shard. In case an older snapshot transfer is still ongoing, we make sure that it's not going to apply transactions anymore.
As a remark, currently there's no way to abort a snapshot on the leader, except sending a `finish` request. This would be changed later on, but for now it does suffice.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentFollowerState.h
+++ b/arangod/Replication2/StateMachines/Document/DocumentFollowerState.h
@@ -82,7 +82,6 @@ struct DocumentFollowerState
   Guarded<GuardedData, basics::UnshackledMutex> _guardedData;
   ActiveTransactionsQueue _activeTransactions;
   std::atomic<std::uint64_t> _snapshotsCount;
-  std::mutex _shardRewriteMutex;
 };
 
 }  // namespace arangodb::replication2::replicated_state::document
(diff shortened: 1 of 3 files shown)


arangod/Replication2/StateMachines/Document/DocumentFollowerState.h
```diff
@@ -81,6 +81,7 @@ struct DocumentFollowerState
   std::unique_ptr<IDocumentStateTransactionHandler> _transactionHandler;
   Guarded<GuardedData, basics::UnshackledMutex> _guardedData;
   ActiveTransactionsQueue _activeTransactions;
+  std::atomic<std::uint64_t> _snapshotsCount;
```
mpoeter, line 84: It looks like _snapshotsCount is only access inside `_guardedData.doUnderLock`, so I think we can move it into GuardedData and make it non-atomic.

mpoeter (commented):
In general LGTM, but I would prefer to use the name `_currentSnapshotId` or `_currentSnapshotVersion` instead of  `_snapshotCount`, because I think it better reflects the meaning of the variable.

mpoeter (approved):

apetenchea:
Thanks, I've been looking for a better name!

PR #17911, opened 2023-01-10
Small improvements for cmake presets

jsteemann:
### Scope & Purpose

Small improvements for cmake presets

- turn off jemalloc for ASan/UBSan builds, because no other memory allocator should be used with ASan
- use the same ASan/UBSan options as we currently do in Oskar (they were recently changed there)
- fix "incorrect" preset names

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -61,7 +61,8 @@
       "inherits": "community-developer",
       "displayName": "Build ArangoDB Community Edition (TSAN Build)",
       "cacheVariables": {
-        "CMAKE_CXX_FLAGS": "-fsanitize=thread"
+        "CMAKE_CXX_FLAGS": "-fsanitize=thread",
+        "CMAKE_C_FLAGS": "-fsanitize=thread"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


MBkkt (commented):
We don't use vptr checks, I have plan to fix and enable it but now it's not checked in CI

https://github.com/arangodb/oskar/blob/master/scripts/buildArangoDB6.fish#L78

In general it's about local presets I don't really care about it

jsteemann:
@MBkkt : thank you. I adjusted the options.

jvolmer (approved):
LGTM

markuspf:
> In general it's about local presets I don't really care about it

Keep in mind that these presets are not just local, they are part of the ArangoDB codebase as much as `arangod/Cluster/ClusterTypes.h` is.

The goal is to have well defined presets that developers use for reproducibility, and that are also used in CI.

MBkkt:
> Keep in mind that these presets are not just local, 

while they don't used on Jenkins they are local.
(for tests and customer images)

> they are part of the ArangoDB codebase as much as

So not really, for example now they were incorrect for sanitizers, and it doesn't have real effect.

I don't have anything against they but anyone can write own preset local and it's used only local.

markuspf (approved):
Thanks!

PR #17912, opened 2023-01-10
Simplify status handling and dropping of collections

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1190

Simplify status handling and dropping of collections.

Previously, collections had statuses `corrupted`, `loaded`, and `deleted`. In addition there was an atomic boolean flag that indicated whether a collection was deleted or not.
This PR removes all the different collection status, and only keeps the deletion flag. A collection will report as being `loaded` in case it is not deleted, and `deleted` otherwise. The status `corrupted` is not necessary, and hasn't been effectively used before. It only played a role with MMFiles collections, which are not present anymore.
This also allowed removing functions for return the status of a collection. To keep things downwards-compatible, a few places that return a collection's status had to be hard-coded.

In addition, a collection's `_useRevisionsAsDocumentIds` flag was changed from an atomic boolean to a non-atomic boolean, because its value is only set in the constructor of LogicalCollection, and then never changed later.

The dropping of collections inside `TRI_vocbase_t` has also been simplified. Previously dropping was started in a `dropCollection` method, which called into a `dropCollectionWorker()` function. Both functions had internal loops to repeat the drop action in case it didn't succeed initially. This is now simplified, so that only `dropCollectionWorker()` retries until success.
The max timeout for dropping a collection has also been removed from the `dropCollection()` call, as dropping should always succeed eventually. If this is not the case, it indicates a bug or a resource leak (leaked collection object with the status lock still acquired).

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1190
- [ ] Design document:


MBkkt (approved):

arangod/IResearch/IResearchLink.cpp
```diff
@@ -396,7 +396,7 @@ Result IResearchLink::unload() noexcept {
   // the collection was still there. But when we did an unload link,
   // the collection was already lazily deleted.
   if (index().collection().deleted()  // collection deleted
-      || index().collection().status() == TRI_VOC_COL_STATUS_DELETED) {
+      || index().collection().deleted()) {
```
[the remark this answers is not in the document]
jsteemann: Fixed

arangod/V8Server/v8-vocbase.cpp
```diff
@@ -1407,35 +1407,29 @@ static void MapGetVocBase(v8::Local<v8::Name> const name,
...
-      auto status = collection->status();
+    if (!ServerState::instance()->isCoordinator() && collection &&
+        &(collection->vocbase()) == &vocbase) {
       auto cid = collection->id();
-      auto internalVersion = collection->v8CacheVersion();
 
       // check if the collection is still alive
-      if (status != TRI_VOC_COL_STATUS_DELETED && cid.isSet() &&
-          !ServerState::instance()->isCoordinator()) {
+      if (!collection->deleted() && cid.isSet()) {
```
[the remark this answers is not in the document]
jsteemann: Adjusted

arangod/VocBase/LogicalDataSource.h
```diff
@@ -24,18 +24,20 @@
 #pragma once
 
 #include <atomic>
-#include <string_view>
+#include <cstdint>
+#include <string>
 
 #include "Basics/Result.h"
+#include "Basics/debugging.h"
```
MBkkt, line 31:
```suggestion
#include "Assertions/Assert.h"
```

Btw for TRI_ASSERT now can be included only it.
Thanks to @markuspf

arangod/VocBase/LogicalDataSource.h
```diff
@@ -57,16 +59,28 @@ class LogicalDataSource {
...
+
   void deleted(bool deleted) noexcept {
-    _deleted.store(deleted, std::memory_order_relaxed);
+    TRI_ASSERT(deleted);
+    // intentionally do not use memory_order_relaxed here, as
+    // the deleted(true) call will only happen once in the lifetime
+    // of a data source, and it is not necessary to optimize the
+    // performance of it.
+    _deleted.store(deleted);
   }
```
MBkkt, line 74:
```suggestion
  void setDeleted() noexcept {
    #ifdef ARANGODB_MAINTAINER_MODE
      TRI_ASSERT(!_deleted.exchange(true));
    #else
      // relaxed here and in load ok because we don't need
      // happens before between them.
      _deleted.store(true, std::memory_order_relaxed);
    #endif
}
```

jsteemann:
```
In file included from /work/ArangoDB/arangod/VocBase/LogicalCollection.h:37,
                 from /work/ArangoDB/arangod/Utils/Events.h:32,
                 from /work/ArangoDB/arangod/VocBase/VocbaseInfo.cpp:36:
/work/ArangoDB/arangod/VocBase/LogicalDataSource.h:68:63: error: missing binary operator before token "("
   68 | #if defined(ARANGODB_ENABLE_MAINTAINER_MODE) && !__has_feature(thread_sanitizer)
      |               
```

jsteemann:
@MBkkt : I think we should not complicate it any further. This surroundings of setting the atomic boolean are a detail that doesn't affect behavior or performance.
I think the PR is ok as it is (provided Jenkins likes it too). My only goal now for be to land this PR.

neunhoef (approved):
LGTM

arangod/VocBase/LogicalDataSource.h
```diff
@@ -94,6 +107,12 @@ class LogicalDataSource {
   TRI_vocbase_t& vocbase() const noexcept { return _vocbase; }
 
  protected:
+  // revert a setDeleted() call later. currently only used by LogicalView.
+  // TODO: should be removed
+  void setUndeleted() noexcept {
+    _deleted.store(false, std::memory_order_seq_cst);
```
mpoeter, line 113: It does not really make sense to make this seq-cst if the other operations on _deleted are all relaxed. It doesn't really hurt, but it might indicate to some reader that there is some synchronization which actually does not exist.

arangod/VocBase/vocbase.cpp
```diff
@@ -819,45 +805,41 @@ arangodb::Result TRI_vocbase_t::loadCollection(
...
+  std::string const& dbName = _info.getName();
 
-  StorageEngine& engine = server().getFeature<EngineSelectorFeature>().engine();
-  engine.prepareDropCollection(*this, *collection);
+  arangodb::Result res;
 
-  double endTime = TRI_microtime() + timeout;
+  auto guard = scopeGuard([&res, &colName, &dbName]() noexcept {
+    try {
+      events::DropCollection(dbName, colName, res.errorNumber());
```
mpoeter, line 830: Previously there were cases where we returned from the function _without_ calling events::DopCollection (e.g. TRI_ERROR_SHUTTING_DOWN). With the scopeGuard this behavior changes. Is this intentional?

mpoeter (approved):

arangod/VocBase/LogicalDataSource.h
MBkkt: +, but also this function should be deleted in future, so why not

arangod/VocBase/vocbase.cpp
jsteemann: Yes!

PR #17915, opened 2023-01-11
[CINFRA] State Creation Speedup

maierlars:
### Scope & Purpose
Collection and Database creation was slow with replication 2. The reason for this was, that changes in replicated states were not reported to the agency asap. This was, because the database was not marked as dirty. This PR changes that. Once a leader is established, the database is marked as dirt, which triggers a maintenance run, which then reports the changes to the agency. Similar code was already in place for updates.

Additionally I changes the times in `waitFor` for testing. Usually replicated states react faster than .5s. I was able (at least on my machine) to push the test run time for `replication2-supervision-replicated-log-target-cluster.js` from 2.5 minutes to 1.5 minutes. Most of the time was spend in waitFor, waiting for the system to converge (which it already did).


apetenchea (approved):

apetenchea:
The PR is related to [CINFRA-529](https://arangodb.atlassian.net/browse/CINFRA-529)

PR #17920, opened 2023-01-11
[CINFRA] Replace Participant Race with Agency Cache

maierlars:
### Scope & Purpose
I did two things in this PR:
1. I added `waitForLastCommitIndex()` to the agency cache. The function calls the agency and looks for the current commit index. Then it waits for this index to be visible in the local cache. Doing that ensures, that writes that were performed on a different coordinator before a call to `waitForLastCommitIndex` are visible locally.
2. To implement the above function I added `getCurrentCommitIndex` to the AsyncAgencyComm.
3. The replace-participant call, that failed due to a wrong cache, is not patched. See below for more details, why the test failed.
4. I added a precondition, that was missing and allowed the write, although something else was already present in the agency. Just having this precondition, wouldn't have fixed the test. Instead it would have failed with precondition-failed.

#### Details ####
This is a caching issue. The test does the following: it creates a replicated log, without an explicit leader set. Then it sets the leader explicitly to the current leader. After that it replaces the leader with another server.
The test now asserts, that the explicitly set leader is replaced with the new server, that the new server becomes the leader and that the old leader is removed.
However, in this run, the coordinator agency cache did not see that the leader was explicitly set, because the cache did not reflect the first write. Hence it did not generate a transaction that replaces the explicitly set leader.

### DIFF
--- a/arangod/Agency/AsyncAgencyComm.cpp
+++ b/arangod/Agency/AsyncAgencyComm.cpp
@@ -434,8 +434,8 @@ namespace arangodb {
 
 futures::Future<consensus::index_t> AsyncAgencyComm::getCurrentCommitIndex()
     const {
-  auto future = sendWithFailover(fuerte::RestVerb::Get, "_api/agency/config",
-                                 120s, RequestType::CUSTOM, {});
+  auto future = sendWithFailover(fuerte::RestVerb::Get, "/_api/agency/config",
+                                 120s, RequestType::READ, {});
   return std::move(future).thenValue([](AsyncAgencyCommResult&& response) {
     if (auto result = response.asResult(); result.fail()) {
       THROW_ARANGO_EXCEPTION(result);
--- a/arangod/Agency/AsyncAgencyComm.cpp
+++ b/arangod/Agency/AsyncAgencyComm.cpp
@@ -432,6 +432,20 @@ arangodb::AsyncAgencyComm::FutureResult agencyAsyncSend(
 
 namespace arangodb {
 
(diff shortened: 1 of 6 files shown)


goedderz (approved):

neunhoef (approved):
LGTM

arangod/Agency/AsyncAgencyComm.cpp
```diff
@@ -432,6 +432,20 @@ arangodb::AsyncAgencyComm::FutureResult agencyAsyncSend(
 
 namespace arangodb {
 
+futures::Future<consensus::index_t> AsyncAgencyComm::getCurrentCommitIndex()
+    const {
+  auto future = sendWithFailover(fuerte::RestVerb::Get, "_api/agency/config",
```
jsteemann, line 437: Nit: in most other places that call agency REST APIs, the paths start with a forward slash. Here the path doesn't start with a forward slash. It probably still works, but it would be more consistent to make the path start with a `/` here as well.
```suggestion
  auto future = sendWithFailover(fuerte::RestVerb::Get, "/_api/agency/config",
```

arangod/Agency/AsyncAgencyComm.cpp
```diff
@@ -432,6 +432,20 @@ arangodb::AsyncAgencyComm::FutureResult agencyAsyncSend(
...
+                                 120s, RequestType::CUSTOM, {});
```
jsteemann, line 438: Question: if we use `RequestType::CUSTOM` here, we don't get a retry in case of errors or even redirects. Is that intentional? If not, we may want to use `RequestType::READ`.

jsteemann (approved):
LGTM

PR #17922, opened 2023-01-11
[PREGEL] Provide test binary for Pregel

markuspf:
### Scope & Purpose

Adds `CMakeLists.txt` for `tests/Pregel`, which adds a test executable `arangodbtests_pregel` and a test case `pregel`;

Note that `typedbuffer.cpp` is not included in this test-binary intentionally because of its large dependency footprint and the fact that we plan to remove that class. It is still tested by `arangodbtests`.


jvolmer (approved):
LGTM

PR #17924, opened 2023-01-12
[CINFRA] Snapshot Version

maierlars:
### Scope & Purpose

It can happen that during a snapshot transfer, this transfer has to be restarted. Although it is more of a theoretic problem, we still want to cover it. I introduced the snapshot version, so that a completion event of a snapshot does not mark a snapshot as available, if it does not belong to the correct version.


goedderz (approved):

goedderz (commented):
I think this should get at least one basic test.

goedderz (approved):

PR #17925, opened 2023-01-12
[PREG-207] Use struct for pregel options

jvolmer:
Uses a struct for the pregel options instead of handling everything with a velocypack.
Extracting the pregel parameters (e.g. "store", "shartKeyAttribute", ...) is a bit more work because it involves using structs in all algorithms for their algorithm-specific parameters instead of plain velocypack, therefore this will be done later in another PR.

Additionally, I moved the code that processes the options from the REST handler into the feature, such that the REST handler is doing no business work and processing the options is done at one place: The feature.


markuspf (approved):
LGTM

PR #17926, opened 2023-01-12
[CINFRA-654] Introduce shard operations on follower

apetenchea:
### Scope & Purpose

**Context**
The plan is to get rid of the `agencyHandler` from the `DocumentCore`. The maintenance will create the shard on the leader, which in turn will instruct the followers to do the same.

**In this PR**
We are introducing two new operations: `CreateShard` and `DropShard`. We make sure to handle them during `applyEntries` on the follower and `recoverEntries` on the leader.

The `DocumentStateShardHandler` no longer uses the `shardId` of the core, but expects to get one as parameter.

Shard operations are not currently issued by the leader, so in reality the follower is not going to see these in the log.

[CINFRA-653](https://arangodb.atlassian.net/browse/CINFRA-654) is also handled here.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

[CINFRA-653]: https://arangodb.atlassian.net/browse/CINFRA-653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
@@ -105,6 +105,14 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
 
     while (auto entry = ptr->next()) {
       auto doc = entry->second;
+      if (doc.operation == OperationType::kCreateShard ||
+          doc.operation == OperationType::kDropShard) {
+        // TODO
+        // If the maintenance sees that shards are missing, it is going to
+        // create them on the leader.
+        // Perhaps we should check that they're already created.
+        continue;
+      }
       auto res = transactionHandler->applyEntry(doc);
       if (res.fail()) {
         return res;
--- a/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
@@ -159,8 +159,6 @@ auto DocumentLeaderState::replicateOperation(velocypack::SharedSlice payload,
                         // returning a dummy index
   }
 
-  // TODO there is a race here, what happens if the leader resigns after this
-  // point? (CINFRA-613)
   auto const& stream = getStream();
   auto entry =
       DocumentLogEntry{std::string(shardId), operation, std::move(payload),
--- a/arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
+++ b/arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
@@ -46,6 +46,8 @@ enum class OperationType {
   kAbort,
   kAbortAllOngoingTrx,
   kIntermediateCommit,
+  kCreateShard,
+  kDropShard
 };
 
 template<class Inspector>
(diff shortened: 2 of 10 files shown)


arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -105,6 +105,14 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
     while (auto entry = ptr->next()) {
       auto doc = entry->second;
+      if (doc.operation == OperationType::kCreateShard ||
+          doc.operation == OperationType::kDropShard) {
+        // TODO
+        // If the maintenance sees that shards are missing, it is going to
+        // create them on the leader.
+        // Perhaps we should check that they're already created.
+        continue;
+      }
```
maierlars, line 115: If we do not create the shard immediately, how do we apply log entries for those shards (which might already exist in the log)

arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
```diff
@@ -59,7 +61,9 @@ auto inspect(Inspector& f, OperationType& p) {
       OperationType::kCommit, "Commit",                          //
       OperationType::kAbort, "Abort",                            //
       OperationType::kAbortAllOngoingTrx, "AbortAllOngoingTrx",  //
-      OperationType::kIntermediateCommit, "IntermediateCommit"   //
+      OperationType::kIntermediateCommit, "IntermediateCommit",  //
+      OperationType::kCreateShard, "CreateShard",                //
+      OperationType::kDropShard, "DropShard"                     //
```
maierlars, line 66: Not sure if we need a `ModifyShard` operation as well? Someone should look into which properties of a shard can change 🤔

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: True! This is a problem I've been thinking about yesterday. I was guessing what would happen in the maintenance at the point where a replicated state becomes leader. For the maintenance to create shards on the leader, it would have to know that the local replicated state is the leader of the shard sheaf in which the missing shard belongs to. This implies that the leader state is instantiated before the maintenance tries to create the shard, which I don't know if it's always the case.
I think a possible solution would be to let the leader create the shard in recovery, and when the maintenance runs, it would instruct the leader to create the shard again, but it would only do so unless it wasn't already created yet.

arangod/Replication2/StateMachines/Document/DocumentLogEntry.h
apetenchea: I will cover that up. I didn't wanted to add `ModifyShard` in here until I know for sure what it is supposed to do.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: I created [CINFRA-656](https://arangodb.atlassian.net/browse/CINFRA-656) to tackle this problem. I suggest to let the core create the shards in this PR.

PR #17940, opened 2023-01-13
[CINFRA] Use Persisted LogIterator on Follower

maierlars:
### Scope & Purpose
The purpose of this PR is to understand the impact of serving a log iterator from RocksDB instead of from a InMemoryLog. In this branch every iterator is backed by RocksDB. Performance tests are advised. In future we probably want to keep a reasonable amount of log in memory and evict everything else.


arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -314,6 +314,44 @@ auto StorageManager::getTermIndexMapping() const -> TermIndexMapping {
   return guardedData.getLockedGuard()->onDiskMapping;
 }
 
+auto StorageManager::getCommittedLogIterator(LogRange range) const
+    -> std::unique_ptr<TypedLogRangeIterator<LogEntryView>> {
+  // return guardedData.getLockedGuard()->onDiskLog.getIteratorRange(range.from,
+  //                                                                range.to);
+
```
goedderz, line 321:
```suggestion
```

goedderz (approved):

PR #17792, opened 2022-12-12
CircleCI configuration

markuspf:
This PR is a first step in our project to build a new developer CI. Details can also be found [here](https://arangodb.atlassian.net/wiki/spaces/DEV/pages/1947271266/New+CI)

 * Add configuration files for CircleCI,
 * Add a Dockerfile to specify the build container used in CI builds
 * *only* builds the community version on x86-64 linux
 * runs C++ tests
 * It is based on @maierlars' previous work

PR #17959, opened 2023-01-17
[CMake] Add PR presets and clean up a bit

markuspf:
### Scope & Purpose

This PR adds some of the work that is currently in #17792 to `devel`.

The `pr` presets are used to build arangodb for testing on circleci and are currently set up to mimic the setup that is used on the Jenkins testing infrastructure.

I have also hidden a few of the "base" presets that do not constitute a usable preset for configuring/building and added a few missing build presets.

We should probably think about how to test that presets are valid and usable (beyond the fact that if you make a syntax mistake in the presets file, CI won't even run if it relies on cmake presets).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 tags
 build
+build-presets
 /core.*
 *.vim
 .deps
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -18,6 +18,54 @@
       }
     },
     {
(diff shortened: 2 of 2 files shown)


jvolmer (approved):
Nice, I like the multiple inheritance, that makes the definitions more concise.

jvolmer (approved):

PR #17962, opened 2023-01-17
[PREGEL] Remove ReportManager

markuspf:
### Scope & Purpose

The ReportManager was introduced as part of the CustomPregel experimental feature which has already been removed.

New means to observe Pregel's state are going to be introduced in forthcoming PRs


jvolmer (commented):
LGTM

PR #17963, opened 2023-01-17
[PREGEL] Split Graph.h, make PregelShard its own struct

markuspf:
### Scope & Purpose

* Split `Graph.h` into `Vertex.h`, `VertexID.h`, `Edge.h`, and `PregelShard.h` so there is one type per header; this is to facilitate the changes coming to the GraphStore in a forthcoming PR
* Make `PregelShard` a proper type, not just a typedef; this should facilitate the change to making the VertexIDs in Pregel fixed size later.
* Add some small tests for PregelShard

Note that is is based off of #17922


jvolmer (approved):
LGTM

PR #17964, opened 2023-01-17
Make inspection formatter independent of exceptions

jvolmer:
The inspection formatter uses now serializeWithErrorT instead of serialize in formatter. This comes handy if you want to use the formatter, but don't want to link to the full arango that includes exceptions.


markuspf (approved):
lgtm; need a green jenkins though.

PR #17970, opened 2023-01-18
[PREGEL] Add actor framework

jvolmer:
Adds a self-written actor framework that implements the actor model.
This framework will be used for messaging in Pregel and will simplify the Pregel code. Several PRs will follow in which this framework will be used inside Pregel. In the long run it can be beneficial to use the C++ Actor Framework instead of our own implementation because it is much more powerful, but for the first usage we did not want to pull a full framework into ArangoDB.

The idea of the actor model is that you have several units of computation (the actors) that run concurrently and possibly distributed. Actors only interact via messages. In our implementation, an actor has a message queue where receiving messages are buffered. Inside the actor everything runs non-concurrently: An actor takes a message from the queue, then processes it, then takes the next message from the queue and processes that and so on. This makes the processing inside the actor much less error-prone. Additionally, the framework does the sending of messages for you (also between different servers), such that your code using the framework does not need to handle that.

## Usage
One actor runtime runs on each server. This runtime has a map of all actors running on this server and is responsible for spawning new actors and communication between actors (also across servers).

To create your own custom actor you need to define
- an actor state,
- message types the actor can receive (combined inside a variant) and 
- a handler that defines what happens when a specific message type arrives. 
Have a look at example actors we used in the tests: TrivialActor.cpp, SpawnActor.cpp or PingPongActor.cpp.

Inside the handler, the actor's state can be changed, new actors can be spawned and messages can be send to other actors. Actors are typesafe, meaning they can only receive messages of their specified message types (and some predefined error message types).

When spawning an actor, a specific state and a start message have to be given. Then the runtime spawns an actor in the specific state and directly sends the start message to it.

Actors are identified via their PID, which consists of a server ID and an actor ID on that server. The handler can only send messages to other actors if it knows the receiving actor's PID. This PID can e.g. come from a received message or it is part of the actor's state.

Errors can happen when messages are send: If an error happens during sending - e.g. the receiving actor cannot be found - an error message is sent back to the sending actor. The handler can define the behavior of the custom actor if a specific error type arrives or define a catch-all for all message types it does not define a handle for. The error types are defined in Message.h; the following errors can currently occur: 
- UnknownMessage if the actor tried to send a message to another actor which does not recognize the sent message type, 
- ActorNotFound if the receiving actor cannot be found on the receiving server or
- ServerNotFound if the receiving server cannot be found

### Actor
An actor receives messages (serialized or non-serialized) via the _process_ method, tries to convert the message into its own message type or an error-message and - if successful - pushes this message to its queue. This starts a work cycle that is moved to the scheduler. In this work cycle, _batchsize_ messages from the queue are processed sequentially. The scheduler is customizable, in ArangoDB we will use the ArangoDB scheduler. We use a lock-free multi-producer-single-consumer message queue (defined in MPSCQueue.h) such that we will not get any troubles when using the ArangoDB scheduler. The message queue queues messages of type MessageOrError which is a variant of the actor-specific messages and pre-defined errors. This way, the actor can receive messages and errors.

### Runtime
An actor runtime is running on each server and knows all actors running on this server - all having the same ActorBase interface. This runtime is responsible for
- spawning new actors
- sending a message to another actor (which can be an actor on the same server - meaning inside the same runtime) or on another server, and
- distributing messages coming from another server to the receiving actor in this runtime.
The runtime decides whether a message is sent locally - because the receiving actor is on the same server - or sent externally to an actor on another server. 
If a message is sent to an actor on another server, the message is serialized to velocypack and an external dispatcher sends the serialized message. The external dispatcher is customizable: A very basic example of an external dispatcher is given in MultiRuntimeTest.cpp, for ArangoDB we have to write one that sends the message via our REST interface.

### HandlerBase
The HandlerBase provides method calls of the runtime (spawning a new actor and dispatching a message to another actor) to a custom actor handler. This way the custom handler needs to inherit from HandlerBase to spawn and dispatch without having a handle on the runtime itself. Additionally, it provides the current actor state, the actor's own PID and the sender PID of the message it processes.

## TODO
We still need to implement the cleanup of actors inside the runtime, e.g. when shutting down the runtime.

### DIFF
--- a/arangod/Pregel/Actor/CMakeLists.txt
+++ b/arangod/Pregel/Actor/CMakeLists.txt
@@ -1,3 +1,8 @@
 add_library(arango_actor INTERFACE)
 target_include_directories(arango_actor INTERFACE include/)
 target_link_libraries(arango_actor INTERFACE arango_crashhandler velocypack fmt)
+
+add_library(arango_actor_standalone INTERFACE)
+target_include_directories(arango_actor_standalone INTERFACE include/)
+target_link_libraries(arango_actor_standalone INTERFACE velocypack fmt)
+target_compile_definitions(arango_actor_standalone INTERFACE STANDALONE)
--- a/arangod/Pregel/Actor/CMakeLists.txt
+++ b/arangod/Pregel/Actor/CMakeLists.txt
@@ -1,3 +1,3 @@
 add_library(arango_actor INTERFACE)
 target_include_directories(arango_actor INTERFACE include/)
-target_link_libraries(arango_actor INTERFACE velocypack fmt)
+target_link_libraries(arango_actor INTERFACE arango_crashhandler velocypack fmt)
--- a/arangod/Pregel/Actor/CMakeLists.txt
+++ b/arangod/Pregel/Actor/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_library(arango_actor INTERFACE)
+target_include_directories(arango_actor INTERFACE include/)
+target_link_libraries(arango_actor INTERFACE velocypack fmt)
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -25,7 +25,6 @@
 
 #include <Inspection/VPackWithErrorT.h>
 #include <memory>
-#include <iostream>
 #include <string_view>
 #include <type_traits>
 
@@ -66,7 +65,7 @@ concept IsInspectable = requires(typename A::Message message,
   {fmt::format("{}", message)};
   {fmt::format("{}", state)};
 };
-};  // namespace
+}  // namespace
 template<typename Runtime, typename A>
 concept Actorable = IncludesAllActorRelevantTypes<Runtime, A> &&
     IncludesName<A> && HandlerInheritsFromBaseHandler<Runtime, A> &&
@@ -84,20 +83,18 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
         std::size_t batchSize)
       : pid(pid),
         runtime(runtime),
-        state(std::move(initialState)),
-        batchSize(batchSize) {}
+        batchSize(batchSize),
+        state(std::move(initialState)) {}
 
   ~Actor() = default;
 
   auto typeName() -> std::string_view override { return Config::typeName(); };
 
-  void process(ActorPID sender,
-               std::unique_ptr<MessagePayloadBase> msg) override {
-    if (auto* m =
-            dynamic_cast<MessagePayload<typename Config::Message>*>(msg.get());
+  void process(ActorPID sender, MessagePayloadBase& msg) override {
+    if (auto* m = dynamic_cast<MessagePayload<typename Config::Message>*>(&msg);
         m != nullptr) {
       push(sender, std::move(m->payload));
-    } else if (auto* n = dynamic_cast<MessagePayload<ActorError>*>(msg.get());
+    } else if (auto* n = dynamic_cast<MessagePayload<ActorError>*>(&msg);
                n != nullptr) {
       push(sender, std::move(n->payload));
     } else {
@@ -107,7 +104,7 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
     }
   }
 
-  void process(ActorPID sender, velocypack::SharedSlice msg) override {
+  void process(ActorPID sender, velocypack::SharedSlice const msg) override {
     if (auto m =
             inspection::deserializeWithErrorT<typename Config::Message>(msg);
         m.ok()) {
@@ -141,6 +138,10 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
     return res.get();
   }
 
+  template<typename R, typename C, typename Inspector>
+  requires Actorable<R, C>
+  friend auto inspect(Inspector& f, Actor<R, C>& x);
+
  private:
   void push(ActorPID sender, typename Config::Message&& msg) {
     pushToQueueAndKick(std::make_unique<InternalMessage>(
@@ -164,7 +165,7 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
   }
 
   void work() {
-    if (finished.load() == true) {
+    if (finished.load()) {
       return;
     }
     auto i = batchSize;
@@ -174,15 +175,15 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
           typename Config::template Handler<Runtime>{
               {pid, msg->sender, std::move(state), runtime}},
           msg->payload->item);
-      i--;
-      if (i == 0) {
-        // push more work to scheduler if queue is still not empty
-        if (not inbox.empty()) {
-          kick();
-          return;
-        }
+      if (--i == 0) {
         break;
       }
+    }
+
+    // push more work to scheduler if queue is still not empty
+    if (not inbox.empty()) {
+      kick();
+      return;
     }
 
     idle.store(true);
@@ -215,20 +216,21 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
 
     // only push work to scheduler if actor is idle (meaning no work is waiting
     // on the scheduler and no work is currently processed in work())
-    auto isIdle = idle.load();
-    if (isIdle and idle.compare_exchange_strong(isIdle, false)) {
+    auto isIdle = true;
+    if (idle.load() and idle.compare_exchange_strong(isIdle, false)) {
       kick();
     }
   }
 
- public:
   ActorPID pid;
   std::atomic<bool> idle{true};
-  std::atomic<bool> finished;
+  std::atomic<bool> finished{false};
   arangodb::pregel::mpscqueue::MPSCQueue<InternalMessage> inbox;
   std::shared_ptr<Runtime> runtime;
-  std::unique_ptr<typename Config::State> state;
   std::size_t batchSize{16};
+
+ public:
+  std::unique_ptr<typename Config::State> state;
 };
 
 // TODO make queue inspectable
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -212,11 +212,6 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
     ActorPID sender;
     std::unique_ptr<MessageOrError<typename Config::Message>> payload;
   };
-  template<typename Inspector>
-  auto inspect(Inspector& f, InternalMessage& x) {
-    return f.object(x).fiels(f.field("sender", x.sender),
-                             f.field("payload", x.payload));
-  }
 
   auto pushToQueueAndKick(std::unique_ptr<InternalMessage> msg) -> void {
     // don't add new messages when actor is finished
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -1,4 +1,4 @@
-////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////
 /// DISCLAIMER
 ///
 /// Copyright 2014-2022 ArangoDB GmbH, Cologne, Germany
@@ -38,18 +38,17 @@ namespace arangodb::pregel::actor {
 
 namespace {
 template<typename Runtime, typename A>
-concept IncludesAllActorRelevantTypes =
-    std::is_class<typename A::State>::value &&
-    std::is_class<typename A::template Handler<Runtime>>::value &&
-    std::is_class<typename A::Message>::value;
+concept IncludesAllActorRelevantTypes = std::is_class_v<typename A::State> &&
+    std::is_class_v<typename A::template Handler<Runtime>> &&
+    std::is_class_v<typename A::Message>;
 template<typename A>
 concept IncludesName = requires() {
   { A::typeName() } -> std::convertible_to<std::string_view>;
 };
 template<typename Runtime, typename A>
 concept HandlerInheritsFromBaseHandler =
-    std::is_base_of < HandlerBase<Runtime, typename A::State>,
-typename A::template Handler<Runtime> > ::value;
+    std::is_base_of_v < HandlerBase<Runtime, typename A::State>,
+typename A::template Handler<Runtime> > ;
 template<typename Runtime, typename A>
 concept MessageIsVariant = requires(ActorPID pid,
                                     std::shared_ptr<Runtime> runtime,
@@ -94,20 +93,17 @@ struct Actor : ActorBase {
 
   void process(ActorPID sender,
                std::unique_ptr<MessagePayloadBase> msg) override {
-    auto* ptr = msg.release();
-    if (auto* m = dynamic_cast<MessagePayload<typename Config::Message>*>(ptr);
+    if (auto* m =
+            dynamic_cast<MessagePayload<typename Config::Message>*>(msg.get());
         m != nullptr) {
       push(sender, std::move(m->payload));
-      delete m;
-    } else if (auto* n = dynamic_cast<MessagePayload<ActorError>*>(ptr);
+    } else if (auto* n = dynamic_cast<MessagePayload<ActorError>*>(msg.get());
                n != nullptr) {
       push(sender, std::move(n->payload));
-      delete n;
     } else {
       runtime->dispatch(
           pid, sender,
           ActorError{UnknownMessage{.sender = sender, .receiver = pid}});
-      delete ptr;
     }
     kick();
   }
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -74,7 +74,7 @@ concept Actorable = IncludesAllActorRelevantTypes<Runtime, A> &&
 
 template<typename Runtime, typename Config>
 requires Actorable<Runtime, Config>
-struct Actor : ActorBase {
+struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
   Actor(ActorPID pid, std::shared_ptr<Runtime> runtime,
         std::unique_ptr<typename Config::State> initialState)
       : pid(pid), runtime(runtime), state(std::move(initialState)) {}
@@ -105,7 +105,6 @@ struct Actor : ActorBase {
           pid, sender,
           ActorError{UnknownMessage{.sender = sender, .receiver = pid}});
     }
-    kick();
   }
 
   void process(ActorPID sender, velocypack::SharedSlice msg) override {
@@ -127,7 +126,6 @@ struct Actor : ActorBase {
         std::abort();
       }
     }
-    kick();
   }
 
   auto finish() -> void override { finished.store(true); }
@@ -145,21 +143,24 @@ struct Actor : ActorBase {
 
  private:
   void push(ActorPID sender, typename Config::Message&& msg) {
-    inbox.push(std::make_unique<InternalMessage>(
+    pushToQueueAndKick(std::make_unique<InternalMessage>(
         sender,
         std::make_unique<MessageOrError<typename Config::Message>>(msg)));
   }
   void push(ActorPID sender, ActorError&& msg) {
-    inbox.push(std::make_unique<InternalMessage>(
+    pushToQueueAndKick(std::make_unique<InternalMessage>(
         sender,
         std::make_unique<MessageOrError<typename Config::Message>>(msg)));
   }
 
   void kick() {
     // Make sure that *someone* works here
-    // TODO make sure that "this" still exist when this->work() is called
-    //      or add check befor call to this->work()
-    (*runtime->scheduler)([this]() { this->work(); });
+    (*runtime->scheduler)([self = this->weak_from_this()]() {
+      auto me = self.lock();
+      if (me != nullptr) {
+        me->work();
+      }
+    });
   }
 
   void work() {
@@ -201,6 +202,13 @@ struct Actor : ActorBase {
   auto inspect(Inspector& f, InternalMessage& x) {
     return f.object(x).fiels(f.field("sender", x.sender),
                              f.field("payload", x.payload));
+  }
+
+  auto pushToQueueAndKick(std::unique_ptr<InternalMessage> msg) -> void {
+    inbox.push(std::move(msg));
+    if (not busy.load()) {
+      kick();
+    }
   }
 
  public:
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -24,13 +24,15 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 #include <string_view>
 #include <type_traits>
 
-#include "ActorBase.h"
-#include "HandlerBase.h"
-#include "Message.h"
-#include "MPSCQueue.h"
+#include "Actor/ActorBase.h"
+#include "Actor/Assert.h"
+#include "Actor/HandlerBase.h"
+#include "Actor/Message.h"
+#include "Actor/MPSCQueue.h"
 
 namespace arangodb::pregel::actor {
 
@@ -115,13 +117,13 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
       auto error =
           ActorError{UnknownMessage{.sender = sender, .receiver = pid}};
       auto payload = inspection::serializeWithErrorT(error);
-      ADB_PROD_ASSERT(payload.ok());
+      ACTOR_ASSERT(payload.ok());
       runtime->dispatch(pid, sender, payload.get());
     }
   }
 
   auto finish() -> void override { finished.store(true); }
-  auto finishedAndIdle() -> bool override {
+  auto isFinishedAndIdle() -> bool override {
     return finished.load() and idle.load();
   }
 
@@ -129,8 +131,16 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
 
   auto serialize() -> velocypack::SharedSlice override {
     auto res = inspection::serializeWithErrorT(*this);
-    ADB_PROD_ASSERT(res.ok());
+    ACTOR_ASSERT(res.ok());
     return res.get();
+  }
+
+  auto getState() -> std::optional<typename Config::State> {
+    auto s = state.get();
+    if (s == nullptr) {
+      return std::nullopt;
+    }
+    return *s;
   }
 
   template<typename R, typename C, typename Inspector>
@@ -232,8 +242,6 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
   std::shared_ptr<Runtime> runtime;
   // tunable parameter: maximal number of processed messages per work() call
   std::size_t batchSize{16};
-
- public:
   std::unique_ptr<typename Config::State> state;
 };
 
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -220,6 +220,7 @@ struct Actor : ActorBase, std::enable_shared_from_this<Actor<Runtime, Config>> {
   std::atomic<bool> finished{false};
   arangodb::pregel::mpscqueue::MPSCQueue<InternalMessage> inbox;
   std::shared_ptr<Runtime> runtime;
+  // tunable parameter: maximal number of processed messages per work() call
   std::size_t batchSize{16};
 
  public:
--- a/arangod/Pregel/Actor/include/Actor/Actor.h
+++ b/arangod/Pregel/Actor/include/Actor/Actor.h
@@ -1,0 +1,223 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2022 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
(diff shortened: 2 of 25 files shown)


arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,223 @@
...
+concept IncludesAllActorRelevantTypes =
+    std::is_class<typename A::State>::value &&
+    std::is_class<typename A::template Handler<Runtime>>::value &&
+    std::is_class<typename A::Message>::value;
+template<typename A>
+concept IncludesName = requires() {
+  { A::typeName() } -> std::convertible_to<std::string_view>;
+};
+template<typename Runtime, typename A>
+concept HandlerInheritsFromBaseHandler =
```
MBkkt, line 50: Do we use concept? I'm not sure all our compilers good with it

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,223 @@
...
+        batchSize(batchSize) {}
+
+  ~Actor() = default;
+
+  auto typeName() -> std::string_view override { return Config::typeName(); };
+
+  void process(ActorPID sender,
+               std::unique_ptr<MessagePayloadBase> msg) override {
+    auto* ptr = msg.release();
+    if (auto* m = dynamic_cast<MessagePayload<typename Config::Message>*>(ptr);
```
MBkkt, line 98: dynamic cast to resolve type of message is a bad idea.
It's very slow, and it's additional place where we require rtti

For example you can add tag virtual member which will be return enum class and then make switch on it.
And do static_cast.
It's more effective and universal.
Also it will be single virtual call on all process, instead of N dynamic_cast

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,223 @@
...
+          ActorError{UnknownMessage{.sender = sender, .receiver = pid}};
+      auto payload = inspection::serializeWithErrorT(error);
+      if (payload.ok()) {
+        runtime->dispatch(pid, sender, payload.get());
+      } else {
+        fmt::print("Error serializing UnknownMessage");
+        std::abort();
+      }
+    }
+    kick();
```
MBkkt, line 134: kick on every process is a bad idea, I already wrote about this
https://github.com/arangodb/arangodb/pull/17556#pullrequestreview-1181338093

But in last PR it wasn't devel branch, and I agreed it can be merged, in devel I don't think so

arangod/Pregel/Actor/include/Actor/MPSCQueue.h
```diff
@@ -0,0 +1,130 @@
...
+
+// This lock-free multi-producer-single-consumer-queue is inspired by the
+// code published at
+//
+// https://www.1024cores.net/home/lock-free-algorithms/queues/intrusive-mpsc-node-based-queue
+//
+// under a Apache-2.0 license.
+namespace arangodb::pregel::mpscqueue {
+
+template<typename T>
```
MBkkt, line 37: You need to remove this code and use "queue" which allow you to call kick() not on every process.

Or modify actor code, to not call kick on every proccess with this queue.
But in such case you also should use weaker memory orders here

MBkkt (changes requested):
https://github.com/arangodb/arangodb/pull/17556#pullrequestreview-1181338093

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: Jenkins ran through for this PR so I guess it is fine for all tested compilers. At least in Replication2 a concept is used as well, it improves the error message a lot.

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,223 @@
...
+  }
+  void push(ActorPID sender, ActorError&& msg) {
+    inbox.push(std::make_unique<InternalMessage>(
+        sender,
+        std::make_unique<MessageOrError<typename Config::Message>>(msg)));
+  }
+
+  void kick() {
+    // Make sure that *someone* works here
+    (*runtime->scheduler)([this]() { this->work(); });
```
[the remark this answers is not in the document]
jvolmer: You are right: Our plan is to introduce an actor state to keep track whether the actor is still processing stuff, we still have to implement that.

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: I implemented Manuel's suggestion in https://github.com/arangodb/arangodb/pull/17970#discussion_r1073501129

MBkkt: My comment about dynamic_cast, in your fix you still make dynamic_cast in worst case N-1 times

jvolmer:
I rebased on devel and added garbage collection to the actor runtime as one new commit.

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+#pragma once
+
+#include "Actor/ActorBase.h"
+#include "Actor/ActorPID.h"
+#include "Basics/Guarded.h"
+
+namespace arangodb::pregel::actor {
+
+struct ActorList {
+  struct ActorMap : std::unordered_map<ActorID, std::shared_ptr<ActorBase>> {};
```
mpoeter, line 34: Do you really want to derive? Or would this suffice:
```suggestion
  using ActorMap = std::unordered_map<ActorID, std::shared_ptr<ActorBase>>;
```

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+    return actors.doUnderLock(
+        [id](ActorMap& map)
+            -> std::optional<
+                std::reference_wrapper<std::shared_ptr<ActorBase>>> {
+          auto a = map.find(id);
+          if (a == std::end(map)) {
+            return std::nullopt;
+          }
+          auto& [_, actor] = *a;
+          return actor;
```
mpoeter, line 49: NIT: I suppose this could be simplified to
```suggestion
          return a->second;
```

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+        });
+  }
+
+  auto add(ActorID id, std::shared_ptr<ActorBase> actor) -> void {
+    actors.doUnderLock([id, actor = std::move(actor)](ActorMap& map) {
+      map.emplace(id, std::move(actor));
+    });
```
mpoeter, line 56: NIT: we can simply capture by reference:
```suggestion
    actors.doUnderLock([&id, &actor](ActorMap& map) {
      map.emplace(id, std::move(actor));
    });
```

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  auto remove(ActorID id) -> void {
+    auto actor = find(id);
+    if (actor.has_value()) {
+      actors.doUnderLock([id](ActorMap& map) { map.erase(id); });
+    }
```
mpoeter, line 63: NIT: In case the actor exists this has to acquire the lock twice.
No need to change now, just something to keep in mind in case we find that lock contention on the actor map to be a problem.

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+
+  auto removeIf(
+      std::function<bool(std::shared_ptr<ActorBase> const&)> deletable)
+      -> void {
+    actors.doUnderLock([deletable = std::move(deletable)](ActorMap& map) {
+      std::erase_if(map, [deletable = std::move(deletable)](const auto& item) {
+        auto& [_, actor] = item;
+        return deletable(actor);
+      });
+    });
```
mpoeter, line 74: I think this could be simplified a bit:
```suggestion
    actors.doUnderLock([&deletable](ActorMap& map) {
      std::erase_if(map, [&deletable](const auto& item) {
        return deletable(actor->second);
      });
    });
```
By capturing by reference we avoid move-construction and destruction of a few additional shared_ptr instances (micro-optimization).

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  auto apply(std::function<void(std::shared_ptr<ActorBase>&)> fn) -> void {
+    actors.doUnderLock([fn = std::move(fn)](ActorMap& map) {
+      for (auto&& [id, actor] : map) {
+        fn(actor);
+      }
+    });
```
mpoeter, line 82:
```suggestion
    actors.doUnderLock([&fn](ActorMap& map) {
      for (auto&& [id, actor] : map) {
        fn(actor);
      }
    });
```

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,133 @@
...
+  auto size() const -> size_t {
+    return actors.doUnderLock([](ActorMap const& map) { return map.size(); });
+  }
+
+  Guarded<ActorMap> actors;
+
+  struct ActorInfo {
+    ActorID id;
+    std::string_view type;
+  };
```
mpoeter, line 104: Should this be public?

arangod/Pregel/Actor/include/Actor/ActorList.h
markuspf: yes, but it would then obscure the inteded thing to be gotten out of the pair ;)

arangod/Pregel/Actor/include/Actor/ActorList.h
jvolmer: I used inheritance because I need a specific inspector for ActorMap because the ActorBase inside the shared pointer is an abstract class and therefore have to define the inspection explicitly here.

MBkkt (changes requested):
Also except queue tests all tests are single threaded, it's a lot of synchronization code in other places which also should be tested

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: fixed, kick is now only called when actor is idle

arangod/Pregel/Actor/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_library(arango_actor INTERFACE)
```
MBkkt, line 1: ?
```suggestion
add_library(arango_pregel_actor INTERFACE)
```

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Markus Pfeiffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <Inspection/VPackWithErrorT.h>
+#include <memory>
+#include <iostream>
```
MBkkt, line 28: Is it necessary?
```suggestion
```

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+    (*runtime->scheduler)([self = this->weak_from_this()]() {
+      auto me = self.lock();
+      if (me != nullptr) {
+        me->work();
+      }
+    });
+  }
+
+  void work() {
+    if (finished.load() == true) {
```
MBkkt, line 167:
```suggestion
    if (finished.load(std::memory_order_relaxed)) {
```

Now you don't have any guarantee about finished, it's not used for synchronization, only for stop execution

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+
+    // only push work to scheduler if actor is idle (meaning no work is waiting
+    // on the scheduler and no work is currently processed in work())
+    auto isIdle = idle.load();
+    if (isIdle and idle.compare_exchange_strong(isIdle, false)) {
+      kick();
+    }
+  }
+
+ public:
```
MBkkt, line 224: why? It looks like details implementation
```suggestion
 private:
```

```suggestion
 protected:
```

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+                             f.field("payload", x.payload));
+  }
+
+  auto pushToQueueAndKick(std::unique_ptr<InternalMessage> msg) -> void {
+    inbox.push(std::move(msg));
+
+    // only push work to scheduler if actor is idle (meaning no work is waiting
+    // on the scheduler and no work is currently processed in work())
+    auto isIdle = idle.load();
+    if (isIdle and idle.compare_exchange_strong(isIdle, false)) {
```
MBkkt, line 219: Now you kick almost always, even in case when idle was false.

Btw do you understand what with idle you just write spinlock with try_lock and unlock?
Please specify memory_orders

```suggestion
    auto isIdle = true
    if (idle.load() == isIdle and idle.compare_exchange_strong(isIdle, false)) {
```

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+      kick();
+    }
+  }
+
+ public:
+  ActorPID pid;
+  std::atomic<bool> idle{true};
+  std::atomic<bool> finished;
```
MBkkt, line 227:
```suggestion
  std::atomic<bool> finished{false};
```

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+  arangodb::pregel::mpscqueue::MPSCQueue<InternalMessage> inbox;
+  std::shared_ptr<Runtime> runtime;
+  std::unique_ptr<typename Config::State> state;
+  std::size_t batchSize{16};
```
MBkkt, line 231: Why 16?

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+    if (finished.load() == true) {
+      return;
+    }
+    auto i = batchSize;
+
+    while (auto msg = inbox.pop()) {
+      state = std::visit(
+          typename Config::template Handler<Runtime>{
+              {pid, msg->sender, std::move(state), runtime}},
+          msg->payload->item);
```
MBkkt, line 176: Why do you need other part of payload?

What about using two vector<variant> queue instead of allocate every message?

arangod/Pregel/Actor/include/Actor/MPSCQueue.h
MBkkt: Thanks, first part done.

Now please do second part, specify memory_orders or remove ineffective lock-free code.
If you don't know how, do you think is good idea to merge such code in arangodb?

arangod/Pregel/Actor/include/Actor/Message.h
```diff
@@ -0,0 +1,134 @@
...
+}  // namespace
+
+namespace arangodb::pregel::actor {
+
+struct MessagePayloadBase {
+  virtual ~MessagePayloadBase() = default;
+};
+
+template<typename Payload>
+struct MessagePayload : MessagePayloadBase {
```
MBkkt, line 52: Maybe? In such case it will be single allocation instead of two for process
```suggestion
struct MessagePayload : MessagePayloadBase, Node {
```

arangod/Pregel/Actor/include/Actor/Message.h
```diff
@@ -0,0 +1,134 @@
...
+template<typename... Args0, typename... Args1>
+struct concatenator<std::variant<Args0...>, std::variant<Args1...>> {
+  std::variant<Args0..., Args1...> item;
+  concatenator(std::variant<Args0...> a) {
+    std::visit(overloaded{[this](auto&& arg) { this->item = std::move(arg); }},
+               std::move(a));
+  }
+  concatenator(std::variant<Args1...> b) {
+    std::visit(overloaded{[this](auto&& arg) { this->item = std::move(arg); }},
+               std::move(b));
```
MBkkt, line 112: I'm pretty sure overloaded don't needed here, please excuse me and provide compile error if I'm wrong.

```suggestion
    std::visit([this](auto&& arg) { this->item = std::move(arg); },
               std::move(a));
  }
  concatenator(std::variant<Args1...> b) {
    std::visit([this](auto&& arg) { this->item = std::move(arg); },
               std::move(b));
```

arangod/Pregel/Actor/include/Actor/Runtime.h
```diff
@@ -0,0 +1,206 @@
...
+    if (actor.has_value()) {
+      actor.value()->process(sender, msg);
+    } else {
+      auto error = ActorError{ActorNotFound{.actor = receiver}};
+      auto payload = inspection::serializeWithErrorT(error);
+      if (payload.ok()) {
+        dispatch(receiver, sender, payload.get());
+      } else {
+        fmt::print("Error serializing ActorNotFound");
+        std::abort();
```
MBkkt, line 120: fatal error exit

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+      } else {
+        fmt::print("Error serializing UnknownMessage");
+        std::abort();
+      }
+    }
+  }
+
+  auto finish() -> void override { finished.store(true); }
+  auto finishedAndIdle() -> bool override {
+    return finished.load() and idle.load();
```
MBkkt, line 133: Now it's possible it will be never true.

I think it's a bug if you want to support such API

MBkkt (changes requested):
Please fix bugs:
1. finishedAndIdle
2. pushToQueueAndKick

And improve or simplify ineffective implementation

arangod/Pregel/Actor/include/Actor/MPSCQueue.h
```diff
@@ -0,0 +1,130 @@
...
+struct MPSCQueue {
```
MBkkt, line 38: Also leak here:
create actors
push some messages
finish();
destroy actor -> all not executed messages are leaked

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+           runtime}},
+      message)};
+};
+template<typename A>
+concept IsInspectable = requires(typename A::Message message,
+                                 typename A::State state) {
+  {fmt::format("{}", message)};
+  {fmt::format("{}", state)};
+};
+};  // namespace
```
MBkkt, line 69:
```suggestion
}  // namespace
```

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,139 @@
...
+
+ public:
+  ActorList(ActorMap map) : actors{std::move(map)} {}
+  ActorList() = default;
+
+  auto find(ActorID id) -> std::optional<std::shared_ptr<ActorBase>> {
+    return actors.doUnderLock(
+        [id](ActorMap& map)
+            -> std::optional<
+                std::reference_wrapper<std::shared_ptr<ActorBase>>> {
```
MBkkt, line 50: awesome...

arangod/Pregel/Actor/include/Actor/ActorList.h
MBkkt: @jvolmer 

You don't need find here at all

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,139 @@
...
+  struct ActorInfo {
+    ActorID id;
+    std::string_view type;
+  };
+
+ public:
+  ActorList(ActorMap map) : actors{std::move(map)} {}
+  ActorList() = default;
+
+  auto find(ActorID id) -> std::optional<std::shared_ptr<ActorBase>> {
```
MBkkt, line 46: Maybe ?
```suggestion
  auto find(ActorID id) const -> std::optional<std::shared_ptr<ActorBase>> {
```

arangod/Pregel/Actor/include/Actor/Actor.h
mpoeter: Your suggestion is semantically equivalent.
```
    auto isIdle = idle.load();
    if (isIdle and idle.compare_exchange_strong(isIdle, false)) {
```
We first load the value of `idle` and store it in `isIdle`, then we check that `isIdle` (the local variable) is true, and only if that is the case we try the compare_exchange.

Where do you see a spinlock? With the idle flag just ensure that only a single worker can be enqueued at any point in time.

We deliberately did not specify any memory orders yet since this requires careful analysis.

MBkkt: > Your suggestion is semantically equivalent.

You're right I misread.

> Where do you see a spinlock? With the idle flag just ensure that only a single worker can be enqueued at any point in time.

Here is try_lock, when we store false it's unlock.

> We deliberately did not specify any memory orders yet since this requires careful analysis.

It's devel, we can use simpler solution, because here we have really tricky memory orders (correct for queue and correct for spinlock, don't work together)

mpoeter: > Here is try_lock, when we store false it's unlock.
I don't get it... there is not loop involved and we are not spinning on anything. This compare_exchange is simply used to ensure that at most one thread is scheduling a worker, nothing more.

MBkkt: I didn't write somewhere here is loop.

I wrote here is try_lock (just cas), and in the another place is unlock (store false)

In general nevermind it's not the issue but you can separate this part and make it more clear

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: What do you mean with "other part"?

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: I'm not sure I understand your idea with the enum: I understand that you want to add a virtual method to MessgePayloadBase, e.g.:
```
virtual auto type() -> Type;
```
where `Type` is an enum. Then this enum has to include ActorError and the Message types of all actors in this runtime. If I understood this correct, there is a usability problem with this:
We would like only the Actor itself to know its own Message types it can handle: that way, when you define your own specific actor - e.g. the TrivialActor (in tests/Pregel/Actor/Actors/TrivialActor.h) - you only define the message types once in the specific actor definition (see TrivialMessages in TrivialActor.h). With the enum you would need to add an alternative to the enum each time you define a new specific actor.

arangod/Pregel/Actor/include/Actor/Message.h
jvolmer: I'm not sure I understand why this removes one allocation, which two allocations do you mean?

arangod/Pregel/Actor/CMakeLists.txt
jvolmer: The actor is currently added to be used in the pregel system but it is not dependent on pregel in any way and could possibly be used in other arangodb code as well, therefore I like arango_actor better.

arangod/Pregel/Actor/include/Actor/Message.h
MBkkt: Now you need one allocation for Message another for payload

arangod/Pregel/Actor/include/Actor/Actor.h
mpoeter: I don't understand why you interpret this as lock/unlock, but I wanted to add some comments anyway.

MBkkt: I don't wrote "lock"

```
class SpinLock {
  public:
    bool try_lock() {
       auto locked = _locked.load(relaxed)
       return !locked && _locked.cas(locked, false, acquire, relaxed);
    }
    void unlock() noexcept {
       locked.store(release);
    }
  private:
    std::atomic_bool _locked{false};
};
```

It's spinlock, in this code it's same.
As I already wrote it's ok, in general we are protecting from multiple consumers of queue

arangod/Pregel/Actor/include/Actor/Actor.h
MBkkt: If I don't misread you use only `->item` from `payload`, why we need to make payload separate struct if used only single field ?

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,248 @@
...
+      push(sender, std::move(n.get()));
+    } else {
+      auto error =
+          ActorError{UnknownMessage{.sender = sender, .receiver = pid}};
+      auto payload = inspection::serializeWithErrorT(error);
+      if (payload.ok()) {
+        runtime->dispatch(pid, sender, payload.get());
+      } else {
+        fmt::print("Error serializing UnknownMessage");
+        std::abort();
```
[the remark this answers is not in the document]
MBkkt: Can you explain please? Not just resolve

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: We want to limit the number of messages per work call, but the actual number 16 is arbitrarily chosen and can possibly be tuned.

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: Initially I wanted the concatenator in Message.h to inherit from std::variant<Args0..., Args1...> and don't have any item. But here we call visit on a variant and unfortunately gcc 11.2 - which we use on Jenkins - has a bug: I cannot call visit on an inherited variant (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90943) - it is only fixed in gcc 11.3. Therefore this is a workaround.

arangod/Pregel/Actor/include/Actor/Actor.h
MBkkt: Comment? 

I mean 16 can be not good in some cases, for an example in case of something like data parallelism we want work in really big batches at the same time for big jobs for an example lsm compaction we want don't want something bigger than 1

arangod/Pregel/Actor/include/Actor/Actor.h
MBkkt: Ok

arangod/Pregel/Actor/include/Actor/ActorPID.h
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+#include "Inspection/Format.h"
+
+namespace arangodb::pregel::actor {
+struct ActorID {
+  size_t id;
+
+  auto operator<=>(ActorID const& other) const = default;
+};
+template<class Inspector>
```
markuspf, line 36:
```suggestion
template<typename Inspector>
```
nitpick

arangod/Pregel/Actor/include/Actor/Message.h
jvolmer: I still don't understand how we can get rid of one allocation:
In Runtime::dispatchLocally, we create the MessagePayload (I guess that is the first allocation you refer to). In the upper Actor::push we create an InternalMessage out of this message payload (I guess that is the second allocation). If I understand your correctly, you say that we can get rid of the second allocation if MessagePayload inherits from Node. 
Internal Message inherits from Node and the Queue we use in the Actor is a Queue with a Node pointing at an InternalMessage type. InternalMessage consists not only of the MessagePayload, but also of a sender, so how can we get rid of the second allocation when MessagePayload is a Node?

arangod/Pregel/Actor/include/Actor/Actor.h
jvolmer: Thanks, I fixed that by checking for finished in pushToQueueAndKick instead of work. Like this, when actor is set to finished, no messages can be pushed to queue but the remaining messages in the actor will be worked off.

arangod/Pregel/Actor/include/Actor/MPSCQueue.h
jvolmer: Thanks, that's true. I added a destructor in MPSCQueue that should fix the leak.

arangod/Pregel/Actor/include/Actor/MPSCQueue.h
jvolmer: It can be true that the code is more efficient if we specify the memory orders. But it is a lot of work to figure out the most efficient and correct memory order for each atomic operation and we don't even have a performance test that shows that this effort would be worth the time spent. Therefore I would like to stick to the sequentially consistent memory order for now which will give correct results.

arangod/Pregel/Actor/include/Actor/Actor.h
```diff
@@ -0,0 +1,262 @@
...
+    InternalMessage(
+        ActorPID sender,
+        std::unique_ptr<MessageOrError<typename Config::Message>>&& payload)
+        : sender(sender), payload(std::move(payload)) {}
+    ActorPID sender;
+    std::unique_ptr<MessageOrError<typename Config::Message>> payload;
+  };
+  template<typename Inspector>
+  auto inspect(Inspector& f, InternalMessage& x) {
+    return f.object(x).fiels(f.field("sender", x.sender),
```
markuspf, line 217:
```suggestion
    return f.object(x).fields(f.field("sender", x.sender),
```

arangod/Pregel/Actor/include/Actor/Message.h
MBkkt: > In Runtime::dispatchLocally, we create the MessagePayload (I guess that is the first allocation you refer to).

Now it's (external message) created on stack (if I see all usages), so it's ok for me. 
You fixed it in some commit when you fixed other comment about `unique_ptr`.
So I resolve it.

> If I understand your correctly, you say that we can get rid of the second allocation if MessagePayload inherits from Node.

Yes, comment was about that.

> InternalMessage consists not only of the MessagePayload, but also of a sender, so how can we get rid of the second allocation when MessagePayload is a Node?

Idea was Just protected inherit ExternalMessage from InternalMessage.
So when user allocate ExternalMessage he is also allocate InternalMessage.
So in `proccess(..., std::unique_ptr<ExternalMessage> p)` you can access to the InternalMessage and don't need to allocate it, only change maybe.

MBkkt (changes requested):
Few small fixes needed

arangod/Pregel/Actor/include/Actor/ActorList.h
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  auto remove(ActorID id) -> void {
+    actors.doUnderLock([id](ActorMap& map) { map.erase(id); });
+  }
+
+  auto removeIf(
+      std::function<bool(std::shared_ptr<ActorBase> const&)> deletable)
+      -> void {
+    actors.doUnderLock([&deletable](ActorMap& map) {
```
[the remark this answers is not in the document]
jvolmer: But I want to make sure that the function given to removeIf has the given signature: If it does not have this signature - meaning one input of ActorBase type and the return type bool - removeIf will not work. I don't see why we should make this function more general although it will not work in all other cases.

PR #17971, opened 2023-01-18
[CINFRA-665] Drop replicated states along with their database

goedderz:
### Scope & Purpose

When a database is dropped, also drop all its replicated states.

- [X] :pizza: New feature


arangod/RocksDBEngine/RocksDBPersistedLog.cpp
```diff
@@ -429,6 +429,7 @@ auto RocksDBLogStorageMethods::drop() -> Result {
   }
 
   auto range = RocksDBKeyBounds::LogRange(ctx.objectId);
+  // TODO should we remove rocksutils::removeLargeRange instead?
```
maierlars, line 432:
```suggestion
  // TODO should we remove using rocksutils::removeLargeRange instead?
```
did you meant that?

maierlars (approved):

tests/IResearch/common.cpp
```diff
@@ -631,7 +631,7 @@ void setDatabasePath(arangodb::DatabasePathFeature& feature) {
 
   path /= TRI_GetTempPath();
   path /= std::string("arangodb_tests.") + std::to_string(TRI_microtime());
-  const_cast<std::string&>(feature.directory()) = path.string();
+  feature.setDirectory(path.string());
```
jsteemann, line 634: :+1:

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2536,6 +2536,41 @@ void RocksDBEngine::pruneWalFiles() {
...
 
+Result RocksDBEngine::dropReplicatedStates(TRI_voc_tick_t databaseId) {
+  rocksdb::ReadOptions readOptions;
+  std::unique_ptr<rocksdb::Iterator> iter(_db->NewIterator(
+      readOptions, RocksDBColumnFamilyManager::get(
+                       RocksDBColumnFamilyManager::Family::Definitions)));
+  auto rSlice = rocksDBSlice(RocksDBEntryType::ReplicatedState);
+  for (iter->Seek(rSlice); iter->Valid() && iter->key().starts_with(rSlice);
+       iter->Next()) {
+    if (databaseId != RocksDBKey::databaseId(iter->key())) {
```
jsteemann, line 2547: It looks like that the database id is always stored in bytes 1 to 9 of the RocksDB key (starting counting bytes at 0).
Right now we are looking for all RocksDB keys that start with `rSlice`, which is a one byte type prefix.
Instead, we could change it to look for keys that have a combined prefix of the type byte **AND** the database id, i.e. check the first 9 bytes.
With that change, we could also use `iterate_upper_bound` in the read options, and set it to prefix **AND** (database id + 1).
Because we are iterating over the definitions column family here and there are only few entries in it, I don't expect a large performance impact here. However, we are looking at more keys than necessary.
I suggest at least leaving a TODO code comment.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2536,6 +2536,41 @@ void RocksDBEngine::pruneWalFiles() {
...
+      continue;
+    }
+
+    auto slice =
+        VPackSlice(reinterpret_cast<uint8_t const*>(iter->value().data()));
+
+    auto info = velocypack::deserialize<RocksDBReplicatedStateInfo>(slice);
+    auto methods = std::make_unique<RocksDBLogStorageMethods>(
```
jsteemann, line 2555: Why is the `RocksDBLogStorageMethods` object created on the heap here?
Wouldn't it suffice to create it on the stack?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2536,6 +2536,41 @@ void RocksDBEngine::pruneWalFiles() {
...
+        info.objectId, databaseId, info.stateId, _logPersistor, _db,
+        RocksDBColumnFamilyManager::get(
+            RocksDBColumnFamilyManager::Family::Definitions),
+        RocksDBColumnFamilyManager::get(
+            RocksDBColumnFamilyManager::Family::ReplicatedLogs));
+    auto res = methods->drop();
+    if (res.fail()) {
+      return res;
```
jsteemann, line 2563: It is questionable whether we should drop out here prematurely.
If there are multiple replicated logs for a database, this method may drop only some of them and leave the others around.
Why not try to delete all of them, but still return an error upon failure (e.g. the first error that this method picked up)?

arangod/RocksDBEngine/RocksDBEngine.cpp
goedderz: Good point! I changed it as suggested, and also changed it in `loadReplicatedStates` to do the same.

arangod/RocksDBEngine/RocksDBEngine.cpp
goedderz: True, fixed.

arangod/RocksDBEngine/RocksDBEngine.cpp
goedderz: That is sensible, I've changed it accordingly.

maierlars (approved):

PR #17972, opened 2023-01-18
[PREG-226] Use structs for worker messages

jvolmer:
Use structs instead of plain velocypack for messages sent from worker to conductor.

This is a preparation step for using a proper messaging framework for Pregel.

### DIFF
--- a/arangod/Pregel/Conductor/Conductor.cpp
+++ b/arangod/Pregel/Conductor/Conductor.cpp
@@ -1,3 +1,4 @@
+#include "Pregel/Worker/Messages.h"
 ////////////////////////////////////////////////////////////////////////////////
 /// DISCLAIMER
 ///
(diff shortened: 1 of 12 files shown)


arangod/Pregel/Worker/Worker.cpp
```diff
@@ -455,49 +461,49 @@ void Worker<V, E, M>::_finishedProcessing() {
...
+
+  auto gssFinished =
+      GlobalSuperStepFinished{.executionNumber = _config.executionNumber(),
+                              .sender = ServerState::instance()->getId(),
+                              .gss = _config.globalSuperstep(),
+                              .messageStats = _messageStats};
+  auto serialized = inspection::serializeWithErrorT(gssFinished);
+  if (!serialized.ok()) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(
+        TRI_ERROR_INTERNAL, "Cannot serialize GlobalSuperStepFinished message");
```
markuspf, line 496: we could throw the error message of the serializer as well here

arangod/Pregel/Conductor/Conductor.cpp
```diff
@@ -710,16 +714,21 @@ void Conductor::collectAQLResults(VPackBuilder& outBuilder, bool withId) {
       .executionNumber = _executionNumber, .withId = withId};
   auto serialized = inspection::serializeWithErrorT(collectResults);
   if (!serialized.ok()) {
-    THROW_ARANGO_EXCEPTION(Result{TRI_ERROR_FAILED});
+    THROW_ARANGO_EXCEPTION_MESSAGE(
+        TRI_ERROR_FAILED, "Cannot serialize CollectPregelResults message");
```
markuspf, line 718: could throw the error message of the serializer here.

arangod/Pregel/Conductor/Conductor.cpp
```diff
@@ -307,35 +318,32 @@ bool Conductor::_startGlobalStep() {
 
 // The worker can (and should) periodically call back
 // to update its status
-void Conductor::workerStatusUpdate(VPackSlice const& data) {
+void Conductor::workerStatusUpdate(StatusUpdated&& update) {
```
markuspf, line 321: Is there a particular reason why you use rvalue-refs here (but as far as I saw) nowhere else?

markuspf (changes requested):
A couple of (what I believe are) improvements requested.

arangod/Pregel/Conductor/Conductor.cpp
jvolmer: Yes, because `updateWorkerStatus` is called in this function, which requires an rvalue-ref for the status. I could of  course use `workerStatusUpdate(StatusUpdate const& update)`, and copy `update` but I thought this would be more convenient.
I guess nothing speaks against using rvalue-refs for all messages that are given to a function in conductor or worker.

markuspf (approved):
LGTMn

PR #17975, opened 2023-01-19
Fix hotbackup restore w.r.t. permuted servers, BTS-1196

neunhoef:
### Scope & Purpose

This addresses https://arangodb.atlassian.net/browse/BTS-1196.
See there for details.

It is possible that after hotbackup/download and hotbackup/restore some
of the data in the hotbackup is lost. A collection might suddenly be
empty or the data of a subset of the shards may be missing.

This only happens if the data was taken on one cluster and has been
uploaded and then downloaded to a different cluster with different
DBServer ids.

NOTE: I have intentionally not changed it that the `backup.json`is modified
on download, since it is part of a hotbackup and might be subject to consistency
checks and checksum checking. Furthermore, if we would rewrite the backup.json
it would no longer fit to the directory names in the backup. Therefore, the `backup.json`
is still modified at the hotbackup restore stage and we just use the fact that the matching
function is now canonical and will produce the same result on download and on restore.

-  [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/17976

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1196.


arangod/Cluster/ClusterMethods.cpp
```diff
@@ -3301,24 +3301,36 @@ arangodb::Result matchBackupServersSlice(VPackSlice const planServers,
...
   std::copy(dbServers.begin(), dbServers.end(),
             std::inserter(localCopy, localCopy.end()));
 
-  // Skip all direct matching names in pair and remove them from localCopy
-  std::unordered_set<std::string>::iterator it;
+  // Skip all direct matching names in pair and remove them from localCopy.
+  // If later a server does not occur it means that no translation has to
+  // happen for it.
   for (auto planned : VPackObjectIterator(planServers)) {
     auto const plannedStr = planned.key.copyString();
```
maierlars, line 3322: If we already talk about relying on the order of something, does it matter how we iterate through the `planServers` object? I remember that there were some changed to the `VPackObjectIterator` lately, that affected the order. `planServers` object might change, while being semantically the same?

jsteemann: IMO the order of iteration over `planServers` should not matter here.
For every string contained in it, we only remove it from `localCopy` if it is contained in `localCopy`. If it's not contained in `localCopy`, then we emplace it into `match`.
However, `localCopy` is an std::set and thus unique and ordered. The order in which we remove from it should thus not matter.
And `match` is an `std::map` and thus unique and ordered as well. The order in which we insert into it should not matter.

tests/HotBackup/HotBackupCoordinatorTest.cpp
```diff
@@ -88,6 +88,21 @@ int countSubstring(std::string const& str, std::string const& sub) {
 std::vector<std::string> dbsPath{"arango", "Plan", "DBServers"};
 std::vector<std::string> colPath{"arango", "Plan", "Collections"};
 
+namespace {
+void checkAlphabetic(std::map<ServerID, ServerID> const& matches) {
```
jsteemann, line 92: I think this could be replaced by `std::is_sorted()` (3) from https://en.cppreference.com/w/cpp/algorithm/is_sorted

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -3301,24 +3301,36 @@ arangodb::Result matchBackupServersSlice(VPackSlice const planServers,
...
       localCopy.erase(it);
     } else {
       match.try_emplace(plannedStr, std::string());
     }
   }
   // match all remaining
   auto it2 = localCopy.begin();
-  for (auto& m : match) {
-    m.second = *it2++;
+  for (auto& m : match) {  // alphabetical order!
```
jsteemann, line 3332: Is it guaranteed here that `match` and `localCopy` have the same amount of members?
If not then `it2` could run over the bounds of the `localCopy` container.
I think we should at least add an assertion here that the number of members in `localCopy` is greater or equal to the number of members in `match`.

arangod/Cluster/ClusterMethods.cpp
neunhoef: This was exactly my thinking.

tests/HotBackup/HotBackupCoordinatorTest.cpp
neunhoef: I will look into this this afternoon. The .first entries are sorted by definition, I need to check that also the .second entries are sorted. If this is possible with `is_sorted`, then I will use this.

arangod/Cluster/ClusterMethods.cpp
neunhoef: Initially, we check that dbServers has at least as many members as planServers, otherwise we bail out.
`localCopy` is initially a copy of dbServers.
Then we only remove from `localCopy` things which are in `planServers`,
for those servers in planServers, which are not in localCopy, we create an entry in `match`. Therefore it follows, that in the end, localCopy has at least as many members as match, as far as I see. We could add a comment to this effect and an assertion, Will do.

tests/HotBackup/HotBackupCoordinatorTest.cpp
jsteemann: You can write a custom compare function, such as `std::is_sorted(container.begin(), container.end(), [](auto const& lhs, auto const& rhs) { return lhs.second < rhs.second; });`

neunhoef: Yes, this works. Done it.

jsteemann (approved):
LGTM

maierlars (approved):

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -3301,24 +3301,49 @@ arangodb::Result matchBackupServersSlice(VPackSlice const planServers,
...
+  //  - this function is supposed to return a canonical result, which
+  //    does not depend on the order of the input in `planServers` or
+  //    `dbServers`, therefore we use a sorted set here and a sorted
+  //    map as the result.
+  //  - Performance does not matter at all, this method is called for
+  //    hotbackup download and hotbackup restore, both methods are
+  //    using considerable amounts of time. The server lists are usually
+  //    quite small (3 or 5 or a few dozen at most).
+  // So please do not "optimize" this.
+  std::set<std::string> localCopy;
```
MBkkt, line 3314: std::less<> as second parameter, in such case possible to find string_view without unnecessary conversation: is_transparent
https://en.cppreference.com/w/cpp/utility/functional/less_void

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -3301,24 +3301,49 @@ arangodb::Result matchBackupServersSlice(VPackSlice const planServers,
...
+  // dbServers should be a list of pairwise different servers, this
+  // is usually ensured since it is a vector manufactured from the keys
+  // of a map:
+  TRI_ASSERT(localCopy.size() == dbServers.size());
+
+  // Skip all direct matching names in pair and remove them from localCopy.
+  // If later a server does not occur it means that no translation has to
+  // happen for it.
   for (auto planned : VPackObjectIterator(planServers)) {
     auto const plannedStr = planned.key.copyString();
```
MBkkt, line 3326: Unnecessary copy here

arangod/Cluster/ClusterMethods.cpp
neunhoef: Sorry, I do not understand this. According to cppreference the second template argument of `std::set` is by default `std::less`. Please elaborate.

MBkkt: > According to cppreference the second template argument of std::set is by default std::less. Please elaborate.

Not really :(

https://stackoverflow.com/questions/67456235/why-does-the-c-standard-not-change-stdset-to-use-stdless-as-its-default

neunhoef: OK, understood. Thanks for the hint, but I will ignore it. As argued previously, this is not a performance relevant place at all. Lists of servers are typically size 3, in very rare cases they are 20. And we are in the context of a backup download or restore, which are both relatively slow operations. So it is not worthwhile to
 - clutter the code with useless optimizations
 - make the reader think about why this particular optimization was done here
Finally, the linked article even contains this:
"Moving from std::set<T, std::less<T>> to std::set<T, std::less<>> can make key-finding algorithms more efficient, if the search-key never has to be converted.

Conversely, it can make them less efficient if the conversion occurrs on every call to the comparator, instead of once in the caller on starting the algorithm. Those conversions can be quite expensive."
So without measurement we should not embark on this. And since sizes are small, I doubt it that we can even perform a sensible measurement.

arangod/Cluster/ClusterMethods.cpp
neunhoef: Agreed, but not relevant here.

arangod/Cluster/ClusterMethods.cpp
MBkkt: > Lists of servers are typically size 3, in very rare cases they are 20.

It's not about size of elements in map, it's about unnecessary construction of string (copy and allocation) from string_view (commonly can be received from vpack/constant) in case of find in this map.

> "Moving from std::set<T, std::less> to std::set<T, std::less<>> can make key-finding algorithms more efficient, if the search-key never has to be converted.

For find string_view in map/set of strings it's never converting.

Single conversation (same as with default less but not required in case of inserting existing key) will be in case of insert/emplace


Anyway, my comment not required you to do something, but it's useful tip, in c++ for map/set strings almost always better to write less<>.

It probably could be changed in standard if not ABI issues, but because of ABI we should make it explicit default.

PR #17976, opened 2023-01-19
Fix hotbackup restore w.r.t. permuted servers, BTS-1196

neunhoef:
### Scope & Purpose

This addresses https://arangodb.atlassian.net/browse/BTS-1196.
See there for details.

It is possible that after hotbackup/download and hotbackup/restore some
of the data in the hotbackup is lost. A collection might suddenly be
empty or the data of a subset of the shards may be missing.

This only happens if the data was taken on one cluster and has been
uploaded and then downloaded to a different cluster with different
DBServer ids.

NOTE: I have intentionally not changed it that the backup.jsonis modified
on download, since it is part of a hotbackup and might be subject to consistency
checks and checksum checking. Furthermore, if we would rewrite the backup.json
it would no longer fit to the directory names in the backup. Therefore, the backup.json
is still modified at the hotbackup restore stage and we just use the fact that the matching
function is now canonical and will produce the same result on download and on restore.

This is the backport for 3.10.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1196.
- [*] Original PR: https://github.com/arangodb/arangodb/pull/17975


maierlars (approved):

jsteemann (approved):
LGTM

PR #17981, opened 2023-01-19
RDBCoveringIterator doesn't work but never created

MBkkt:
### Scope & Purpose

This PR fixes RDBCoveringIterator. So far, it has never worked, since it was never activated. This PR
also activates TDBCoveringIterator under suitable circumstances.

Bugs found and fixed:

1. endless loop because _gotIntervals was always false, in general just typo, fixed
2. endless loop because _scanInterval was not advanced
3. endless loop because `break` did not set _scanInternal to the end.

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] C++ **Unit tests**
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18006


neunhoef (approved):
LGTM

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
```diff
@@ -762,9 +776,9 @@ std::unique_ptr<IndexIterator> RocksDBGeoIndex::iteratorForCondition(
   if (!params.sorted &&
       (params.filterType == geo::FilterType::CONTAINS ||
        params.filterType == geo::FilterType::INTERSECTS) &&
-      (params.minDistanceRad() < geo::kRadEps &&
-       params.maxDistanceRad() >
-           geo::kMaxRadiansBetweenPoints - geo::kRadEps)) {
+      (params.minDistance <= geo::kMetersEps &&
+       params.maxDistance >=
+           geo::kMaxDistanceBetweenPoints - geo::kMetersEps)) {
```
MBkkt, line 781: What about test like:

```
LET P = ...
FOR d in ... FILTER GEO_DISTANCE(p, d) > 0
```

I mean what if user don't want same points, only distinct?
I still think here will be better to have boolean condition

MBkkt (commented):
Other LGTM

arangod/RocksDBEngine/RocksDBGeoIndex.cpp
neunhoef: OK, I will try to think of something...

neunhoef: This is indeed a bug, as it seems. Because even with `GEO_DISTANCE(p, d) >= 0` (which should return the point), it is not returned. Looking into it...

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Activate RDB_CoveringIterator and use it for some geo index queries.
```
jsteemann, line 4: Because the CHANGELOG is end-user-visible, what is the benefit of the RDB_CoveringIterator for end users?
Is there any way to see from query execution plans that an RDB_CoveringIterator will be used?
CHANGELOG entry does not necessarily need to be adjusted, but I am just curious what the RDB_CoveringIterator can do for me. :sweat_smile:

MBkkt: At least you don't need to sort results by distance to their centroids

lib/Geo/GeoParams.h
```diff
@@ -53,6 +53,7 @@ inline constexpr double kMaxRadiansBetweenPoints = kPi + kRadEps;
 // constexpr double kEarthRadiusInMeters = (6378.137 * 1000);
 // Volumetric mean radius
 inline constexpr double kEarthRadiusInMeters = (6371.000 * 1000);
+inline constexpr double kMetersEps = kRadEps * kEarthRadiusInMeters;
```
MBkkt, line 56: Is it still necessary?

jsteemann (approved):
LGTM

KVS85 (approved):

PR #17982, opened 2023-01-19
improve object lifetime management in replication manager

jsteemann:
### Scope & Purpose

The RocksDBReplicationManager owns RocksDBReplicationContexts instances, which were previously managed as raw pointers.
Whenever a caller asked the manager to return replication context by id, the manager returned a raw pointer.
Callers could also set a deleted flag on a replication context, marking it as to-be-deleted for the manager, which would eventually garbage-collect it.
This was unnecessarily complex and is now replaced in this PR with std::shared_ptrs.
Handing out a replication context from the manager now happens only by returning a RAII guard object, RocksDBReplicationContextGuard, which will automatically make sure the borrowed context is now deleted while used.
All call sites have been adjusted.

The PR also fixes an object lifetime issue with the RocksDB snapshot that was passed to RocksDBReplicationIterator. Previously, the snapshot was passed to RocksDBReplicationIterator as a raw pointer. The pointer was however managed by the RocksDBReplicationContext object, which obfuscated the ownership.
Now the RocksDBReplicationContext creates the snapshot using an std::shared_ptr, so that the snapshot can easily be passed to multiple sub-iterators inside the same RocksDBReplicationContext, without risking lifetime issues for the snapshot.

The PR should not change behavior, but only simplify the replication context handling code, make it easier to understand, and ideally a bit more robust to use.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -1085,29 +1090,18 @@ double RocksDBReplicationContext::expires() const {
...
   MUTEX_LOCKER(locker, _contextLock);
-  TRI_ASSERT(!_isDeleted);
 
-  ++_users;
-  ttl = std::max(_ttl, ttl);
+  if (ttl < 0.0) {
+    ttl = _ttl;
+  } else {
+    ttl = std::max(_ttl, ttl);
+  }
```
MBkkt, line 1102: I don't understand this change

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -1085,29 +1090,18 @@ double RocksDBReplicationContext::expires() const {
...
-  return (_users > 0);
-}
+void RocksDBReplicationContext::extendLifetime(double ttl) {
+  double now = TRI_microtime();
 
-void RocksDBReplicationContext::use(double ttl) {
   MUTEX_LOCKER(locker, _contextLock);
-  TRI_ASSERT(!_isDeleted);
 
-  ++_users;
```
MBkkt, line 1107: why is it removed?

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -1120,49 +1114,26 @@ void RocksDBReplicationContext::use(double ttl) {
...
+    TRI_ASSERT(_snapshot != nullptr);
+    TRI_ASSERT(_snapshot->snapshot() != nullptr);
+    _snapshotTick = _snapshot->snapshot()->GetSequenceNumber();
   }
 }
 
 RocksDBReplicationContext::CollectionIterator::CollectionIterator(
     TRI_vocbase_t& vocbase, std::shared_ptr<LogicalCollection> const& coll,
-    bool sorted, rocksdb::Snapshot const* snapshot)
+    bool sorted, std::shared_ptr<rocksdb::ManagedSnapshot> snapshot)
```
MBkkt, line 1131: What is the difference between ManagedSnapshot and Snapshot?

Also it looks like it can be unique_ptr, but maybe I'm wrong

arangod/RocksDBEngine/RocksDBReplicationContextGuard.cpp
```diff
@@ -0,0 +1,66 @@
...
+    RocksDBReplicationManager& manager,
+    std::shared_ptr<RocksDBReplicationContext> ctx)
+    : _manager(manager), _ctx(std::move(ctx)), _deleted(false) {}
+
+RocksDBReplicationContextGuard::RocksDBReplicationContextGuard(
+    RocksDBReplicationContextGuard&& other) noexcept
+    : _manager(other._manager),
+      _ctx(std::move(other._ctx)),
+      _deleted(other._deleted) {
+  other._deleted = false;
```
MBkkt, line 47: NIT
```suggestion
      _deleted(std::exchange(other._deleted, false)) {
```

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
jsteemann: We don't need hand-rolled usage counters anymore, as we managed the object via shared_ptr now.
The replication contexts manager maintains a map of these shared_ptrs. Any users of a replication context will get a shared_ptr copy from the manager. Once a user is done, the shared_ptr will go out of scope and reduce its use count.

Additionally, users can request deletion of an object from the manager, which will then simply remove it from its internal map. Or the manager will GC outdated/stale replication contexts, again by removing them from its map.
This simplifies the object handling a lot, and not deletion flag or usage counter is needed on the replication contexts anymore.

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
jsteemann: A ManagedSnapshot is a RAII implementation from RocksDB, that automatically acquires a snapshot upon creation, and release it upon destruction.
So it is not necessary to call `CreateSnapshot()` and `ReleaseSnapshot()` on the DB anymore.

In theory we could use a unique_ptr here, but as callers can request the iterator's snapshot, I could either return a raw pointer or a shared_ptr. I decided to use the latter one, because the code is not performance-critical, and I would like avoid the use of raw pointers here (that was part of the motivation for this PR).

MBkkt: Ok

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
MBkkt: I see

MBkkt (approved):

PR #17984, opened 2023-01-20
[BTS-899] _admin/status crash on agents

maierlars:
### Scope & Purpose
[BTS](https://arangodb.atlassian.net/browse/BTS-899)

Do not expose the reboot id on agents, because there is none.

### DIFF
--- a/arangod/RestHandler/RestStatusHandler.cpp
+++ b/arangod/RestHandler/RestStatusHandler.cpp
@@ -145,11 +145,12 @@ RestStatus RestStatusHandler::executeStandard(ServerSecurityFeature& security) {
 
     if (!isStartup && !serverState->isSingleServer()) {
       result.add("persistedId", VPackValue(serverState->getPersistedId()));
-      if (auto rid = serverState->getRebootId(); rid.initialized()) {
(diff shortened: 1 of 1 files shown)


arangod/RestHandler/RestStatusHandler.cpp
```diff
@@ -145,11 +145,12 @@ RestStatus RestStatusHandler::executeStandard(ServerSecurityFeature& security) {
...
     if (!isStartup && !serverState->isSingleServer()) {
       result.add("persistedId", VPackValue(serverState->getPersistedId()));
-      if (auto rid = serverState->getRebootId(); rid.initialized()) {
-        result.add("rebootId", VPackValue(rid.value()));
-      }
 
       if (!serverState->isAgent()) {
+        // agents do not initialize the reboot id
+        if (auto rid = serverState->getRebootId(); rid.initialized()) {
+          result.add("rebootId", VPackValue(rid.value()));
```
maierlars, line 152: The problem here is, that `getRebootId` asserts that the reboot id is initialized.

jsteemann (approved):
LGTM

PR #17985, opened 2023-01-20
if shell-general-graph-creation-cluster.js is ran with coverage instrumented binaries, this test will not finish any time soon

dothebart:
### Scope & Purpose

creating the maximum number of available shards with the preconfigured number of replications is quiet resource intense. 
Disable replication for this test, since it primarily wants to check whether the limit is respected.

### DIFF
--- a/tests/js/server/shell/shell-general-graph-creation-cluster.js
+++ b/tests/js/server/shell/shell-general-graph-creation-cluster.js
@@ -1,5 +1,5 @@
 /*jshint globalstrict:false, strict:false */
-/*global assertEqual, assertTrue, assertFalse, fail */
+/*global print, assertEqual, assertTrue, assertFalse, fail */
 
 ////////////////////////////////////////////////////////////////////////////////
 /// @brief test the general-graph class
@@ -72,6 +72,10 @@ function GeneralGraphClusterCreationSuite() {
     },
 
     testCreateMoreShardsThanAllowed : function () {
+      if (internal.ccoverage) {
+        print("Skipping testCreateMoreShardsThanAllowed in coverage tests");
+        return;
+      }
       let max = internal.maxNumberOfShards;
       try {
         graph._create(gn, edgeDef, null, { numberOfShards: max + 1 });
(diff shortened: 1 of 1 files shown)


markuspf (commented):
I am not sure I understand the purpose of this PR.

First, I would appreciate if the title of the PR would describe the thing it does, with more detail in the description. 

What does "in coverage" mean? What does "this test" that is modified here do? Why is this a problem "in coverage"?

Leaving that aside, the solution proposed in this PR looks ad-hoc and prone to lead to problems later. To propose a better solution, I will have to understand the problem, though.

dothebart:
gcov instrumented binaries perform way slower than regular binaries. Hence this runs into a timeout. 
A Probable better solution would be to pin-point this to `replication_factor=1` unconditionally if the API permits that.

MBkkt:
@dothebart Can you do it only for gcov? for an example in gcov build you can set some environment variable.
I don't think it's an a good idea to disable replication in some tests.

dothebart:
I did that first - `internal.coverage` - However, I don't think that replication in this case is functional part of this test - but the bottleneck.

mpoeter:
I fully agree with @dothebart here! These tests are also among the longest running in CircleCI. Since replication isn't really a relevant aspect of these tests I also think we should just disable it. BTW: there is also a similar enterprise test

tests/js/server/shell/shell-general-graph-creation-cluster.js
```diff
@@ -1,5 +1,5 @@
 /*jshint globalstrict:false, strict:false */
-/*global assertEqual, assertTrue, assertFalse, fail */
+/*global print, assertEqual, assertTrue, assertFalse, fail */
```
hkernbach, line 2:
```suggestion
/*global assertEqual, assertTrue, assertFalse, fail */
```

mchacki (approved):
LGTM.
Replication is not explicitly part of the test here, we just want to create many shards.

mpoeter:
Blue except for unrelated arangosearch test error

PR #17988, opened 2023-01-20
move IOHeartbeatThread into its own file

jsteemann:
### Scope & Purpose

Move IOHeartbeatThread into its own file.
Previously, the IOHeartbeatThread was located in DatabaseFeature.h/.cpp. With the IOHeartbeatThread in its own file, several standard library includes in DatabaseFeature.h can be avoided.
This is just a structural refactoring. No functionality or behavior are changed.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/DatabaseFeature.h
```diff
@@ -27,17 +27,13 @@
...
-#include "Metrics/Histogram.h"
-#include "Metrics/LogScale.h"
 #include "Replication2/Version.h"
 #include "RestServer/arangod.h"
 #include "Utils/DatabaseGuard.h"
 #include "Utils/VersionTracker.h"
 #include "VocBase/voc-types.h"
 #include "VocBase/Methods/Databases.h"
 
-#include <condition_variable>
```
MBkkt, line 40: All this to just avoid condition_variable and metrics (for metrics it's possible just use Fwd.h)?

I'm not strong against such changes, but is it really worth?
git history will be broken for IOHeartbeatThread, same about backports

arangod/RestServer/IOHeartbeatThread.h
```diff
@@ -0,0 +1,68 @@
...
+///
+/// @author Max Neunhoeffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include "Basics/Thread.h"
+#include "Metrics/Counter.h"
+#include "Metrics/Histogram.h"
+#include "Metrics/LogScale.h"
```
MBkkt, line 29: #include "Metrics/Fwd.h"

neunhoef (approved):
LGTM

markuspf (approved):

arangod/RestServer/DatabaseFeature.h
jsteemann: Yes, but is it relevant that classes move between files? Especially an easy to locate class such as IOHeartbeatThread?
We have other PRs that change 200+ files and we don't complain.
The benefit of this PR IMO is not the we avoid including a few files in a central header file. That benefit will even be hard to measure. The main benefit IMO is that it improves modularization. If you need an IOHeartbeatThread, you now can have one (and only this) by including the appropriate header file. You don't get an IOHeartbeatThread as a (wanted or unwanted) side effect by including DatabaseFeature.h anymore. IMHO it should have been like this from the get-go, and this PR just rectifies it. (same for DatabaseManagerThread by the way, but I like to do changes in small iterations).

PR #17999, opened 2023-01-23
[CINFRA-666] Pass vocbase to document state machine

goedderz:
#### Let vocbase stay accessible to the document state machine

Previously, the document state machine looked up its vocbase by name. However, that leads to a race:
When the database is dropped, it's first dropped from the list of databases. Then the states are resigned, and last the object is destroyed.

So looking up by name can fail before the states are resigned, while the actual vocbase object is still valid.

This PR passes a reference to the vocbase to the DocumentCore for the document state machine to access.

#### Never use LogCore during participant (re)construction

If exceptions occured during participant construction (e.g. due to the vocbase not being available, before the previous change, but also other possible errors), this might mean the LogCore could be destroyed and lost.

This PR now makes sure that doesn't happen, either by making constructors or functions taking a `std::unique_ptr<LogCore>` `noexcept`, or by changing the argument into an rvalue reference, and restoring if necessary.

- [X] :hankey: Bugfix


arangod/Replication2/StateMachines/Document/DocumentCore.cpp
```diff
@@ -32,10 +32,12 @@
 using namespace arangodb::replication2::replicated_state::document;
 
 DocumentCore::DocumentCore(
-    GlobalLogIdentifier gid, DocumentCoreParameters coreParameters,
+    TRI_vocbase_t& vocbase, GlobalLogIdentifier gid,
+    DocumentCoreParameters coreParameters,
     std::shared_ptr<IDocumentStateHandlersFactory> const& handlersFactory,
     LoggerContext loggerContext)
     : loggerContext(std::move(loggerContext)),
+      _vocbase(vocbase),
```
mpoeter, line 40: How do we ensure that the vocbase lives long enough? I suppose we need a DatabaseGuard here.

mpoeter (approved):
One comment, otherwise LGTM

arangod/Replication2/StateMachines/Document/DocumentCore.cpp
maierlars: The vocbase strictly outlives all replicated states, in the sense that `resign` is called on all of them before the vocbase is destroyed.

apetenchea: Apart from what Lars mentioned, a `DatabaseGuard` would still rely on `DatabaseFeature::useDatabase`, which is what causes the issues when looking up by name. Because Tobias decided to use a `vocbase` pointer, we'll be able to access it until the resignation of the replicated state, even though ''others" can no longer lookup the database.

apetenchea (approved):

PR #18004, opened 2023-01-24
Disable "maybe-uninitialized" warnings as errors for GCC

goedderz:
### Scope & Purpose

Disable `maybe-uninitialized` warnings as errors on GCC. The warning stays enabled, but it's no longer a build error.
This is changed because we've seen too many false positives over the last years.

- [X] :building_construction: Build change


jsteemann (approved):
LGTM

PR #18007, opened 2023-01-24
Add missing include for gcc 11.2 with glibc

neunhoef:
### Scope & Purpose

I found that I cannot compile devel with maintainer mode switched off on
Ubuntu with gcc 11.2.0 and glibc. It seems an include is missing. This
PR adds the include.

- [*] :hankey: Bugfix

### DIFF
--- a/lib/Assertions/ProdAssert.h
+++ b/lib/Assertions/ProdAssert.h
@@ -52,6 +52,8 @@
 
 #include "AssertionLogger.h"
 
+#include "Basics/system-compiler.h"
+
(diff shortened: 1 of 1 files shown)


jsteemann (approved):
LGTM

MBkkt (approved):
Is it needed to be backported?
Or it's used only in replication2 code now?

neunhoef:
I think it is only in changed Pregel code. I have not yet had a problem to build 3.10 so far.

PR #18010, opened 2023-01-24
Make ArangoDB compile with clang-15

maierlars:
### Scope & Purpose

Make ArangoDB compile with clang-15. clang-15 now errors on unused but set variables. I removed all of them.
These changes are necessary for me, because otherwise there is no compiler available on arch linux that compiles arangodb. gcc-12.2.1 does not work because of known template errors.

There is one small catch: The automatically generated code in `grammar.cpp` declares a variable `Aqlnerrs` which we do not use. Thus I disabled the warning for that file.

### DIFF
--- a/arangod/Agency/RemoveFollower.cpp
+++ b/arangod/Agency/RemoveFollower.cpp
@@ -230,11 +230,8 @@ bool RemoveFollower::start(bool&) {
                  });
 
   size_t inSyncCount = 0;
-  size_t notGoodCount = 0;
   for (auto const& pair : overview) {
-    if (pair.second == -1) {
-      ++notGoodCount;
-    } else if (static_cast<size_t>(pair.second) >= shardsLikeMe.size()) {
+    if (static_cast<size_t>(pair.second) >= shardsLikeMe.size()) {
       ++inSyncCount;
     }
   }
--- a/arangod/Aql/CMakeLists.txt
+++ b/arangod/Aql/CMakeLists.txt
@@ -216,3 +216,5 @@ endif ()
 if (CMAKE_COMPILER_IS_GNUCC)
   set_source_files_properties(tokens.cpp PROPERTIES COMPILE_FLAGS -Wno-sign-compare)
 endif ()
+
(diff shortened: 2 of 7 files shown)


arangod/Aql/CMakeLists.txt
```diff
@@ -216,3 +216,5 @@ endif ()
 if (CMAKE_COMPILER_IS_GNUCC)
   set_source_files_properties(tokens.cpp PROPERTIES COMPILE_FLAGS -Wno-sign-compare)
 endif ()
+
+set_source_files_properties(grammar.cpp PROPERTIES COMPILE_FLAGS -Wno-unused-but-set-variable)
```
MBkkt, line 220: NIT
```suggestion
// TODO fix it?
set_source_files_properties(grammar.cpp PROPERTIES COMPILE_FLAGS -Wno-unused-but-set-variable)
```

arangod/Agency/RemoveFollower.cpp
```diff
@@ -230,11 +230,8 @@ bool RemoveFollower::start(bool&) {
                  });
 
   size_t inSyncCount = 0;
-  size_t notGoodCount = 0;
   for (auto const& pair : overview) {
-    if (pair.second == -1) {
-      ++notGoodCount;
-    } else if (static_cast<size_t>(pair.second) >= shardsLikeMe.size()) {
+    if (static_cast<size_t>(pair.second) >= shardsLikeMe.size()) {
```
MBkkt, line 234:
```suggestion
    if (pair.second != -1 && static_cast<size_t>(pair.second) >= shardsLikeMe.size()) {
```

For -1 it will be always true, so here your changes incorrect.

MBkkt (changes requested):
Please fix `inSyncCount` compute, other LGTM

arangod/Agency/RemoveFollower.cpp
MBkkt: Another option is cast `size()`/both to int/int64

PR #18012, opened 2023-01-25
[CINFRA] Report state machine state to the agency. Useful for tests.

maierlars:
### Scope & Purpose
A replicated log is _ready_ when the leadership has been established. This is not true for the associated replicated state. The replicated state can do recovery only after the leadership is established and therefore might not be ready immediately after. Some tests fail, because they try to access a not-yet-ready state.

In this PR we add an additional field to Current/ReplicatedLogs in the local states of each dbserver. This field reflects the progress of the replicated state becoming ready. (i.e. is recovery completed)

In fact, this is what we want to wait for during collection creation as well.


arangod/Replication2/ReplicatedState/ReplicatedState.tpp
```diff
@@ -384,6 +396,8 @@ void LeaderStateManager<S>::recoverEntries() {
         if (auto self = weak.lock(); self != nullptr) {
           auto guard = self->_guardedData.getLockedGuard();
           guard->_leaderState->onRecoveryCompleted();
+          guard->_recoveryCompleted = true;
+          LOG_CTX("1b3d0", DEBUG, self->_loggerContext) << "recovery completed";
```
apetenchea, line 400: This log ID is the same as the previous one.

arangod/Replication2/ReplicatedLog/types.h
```diff
@@ -236,4 +236,30 @@ struct QuorumData {
...
+namespace {
+constexpr std::string_view kStringUnconfigured = "Unconfigured";
+constexpr std::string_view kStringRecovery = "RecoveryInProgress";
+constexpr std::string_view kStringOperational = "ServiceOperational";
+}  // namespace
+
+enum class LocalStateMachineStatus { kUnconfigured, kRecovery, kOperational };
+auto to_string(LocalStateMachineStatus) noexcept -> std::string_view;
+
+struct LocalStateMachineStatusStringTransformer {
```
apetenchea, line 248: Since this transformer is not implemented anywhere, I think it's best to not have it here. I could see myself trying to use it in the future and reporting that there's a problem.

lib/Logger/LogTopic.cpp
```diff
@@ -142,8 +142,8 @@ LogTopic Logger::MMAP("mmap");
 LogTopic Logger::PREGEL("pregel", LogLevel::INFO);
 LogTopic Logger::QUERIES("queries", LogLevel::INFO);
 LogTopic Logger::REPLICATION("replication", LogLevel::INFO);
-LogTopic Logger::REPLICATION2("replication2", LogLevel::WARN);
-LogTopic Logger::REPLICATED_STATE("rep-state", LogLevel::WARN);
+LogTopic Logger::REPLICATION2("replication2", LogLevel::DEBUG);
```
apetenchea, line 145: Are we sure we want this?

apetenchea (approved):
LGTM! Left some minor comments.

arangod/Replication2/ReplicatedState/ReplicatedState.tpp
maierlars: correct. But did jenkins not fail with that error?

lib/Logger/LogTopic.cpp
maierlars: I'm sure we don't want this 😆

arangod/Replication2/ReplicatedState/ReplicatedState.tpp
apetenchea: `arangodb-ANY-cppcheck-pr` failed, but there is only `sourceInfo.log` in there. I suppose it was the log ids, it's not obvious where to look for the failure in this case.

maierlars: `arangodb-ANY-check-guidelines` is the one that checks log ids.

maierlars: ah, I know why. This is `LOG_CTX`. And the bash script used to check for log ids does not recognize this. Because it is (of course) a different implementation and does not use `utils/checkLogIds.py`

arangod/Replication2/ReplicatedLog/types.h
maierlars: Gone!

maierlars:
Failure unreleated.

PR #18019, opened 2023-01-26
Experimental Change on SmartGraphs

mchacki:
Internal EE Change only.
See description on EE part
EE:https://github.com/arangodb/enterprise/pull/1197


CHANGELOG
```diff
@@ -1,6 +1,13 @@
 devel
 -----
 
+* Fixed EE: Concurrent batch insert/update CRUD operations into
+  SmartEgeCollections on conflicting edge keys, could get the
```
```suggestion
  SmartEdgeCollections on conflicting edge keys could get the
```

arangod/ClusterEngine/ClusterTransactionState.cpp
```diff
@@ -43,6 +43,7 @@
 #include "Transaction/ManagerFeature.h"
 #include "Transaction/Methods.h"
 #include "VocBase/LogicalCollection.h"
+#include "Utils/CollectionNameResolver.h"
```
jsteemann, line 46: nit: convention is to sort includes alphabetically
```suggestion
#include "Utils/CollectionNameResolver.h"
#include "VocBase/LogicalCollection.h"
```

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -107,7 +108,13 @@ RequestLane RestDocumentHandler::lane() const {
 RestStatus RestDocumentHandler::execute() {
   // extract the sub-request type
   auto const type = _request->requestType();
-
+#ifdef ARANGODB_ENABLE_FAILURE_TESTS
+  std::vector<std::string> const& suffixes = _request->decodedSuffixes();
+  TRI_IF_FAILURE("failOnCRUDAPI" + suffixes[0]) {
```
jsteemann, line 113: Is accessing `suffixes[0]` unconditionally safe here? What if someone just calls `/_api/document` without any suffixes?
I think it is safer to put the `TRI_IF_FAILURE...` into an `if (!suffixes.empty()) { ... }` condition.

mchacki: Ah good point. The if-def will be there in all of our Test Builds. So it may have undesired side effects. Will fix!

arangod/RestHandler/RestDocumentHandler.cpp
```diff
@@ -784,9 +788,8 @@ RestStatus RestDocumentHandler::removeDocument() {
 
   _activeTrx = createTransaction(cname, AccessMode::Type::WRITE, opOptions,
                                  std::move(trxOpts));
-  if (suffixes.size() == 2 || !search.isArray()) {
-    _activeTrx->addHint(transaction::Hints::Hint::SINGLE_OPERATION);
-  }
+
+  addTransactionHints(cname, isMultiple, false);
```
jsteemann, line 792: The case `suffixes.size() == 2` is not optimized anymore, but it should be:
```suggestion
  addTransactionHints(cname, isMultiple || suffixes.size() != 2, false);
```

mchacki: At actually use, because the suffixes == 2 case will Produce an Object, not an array, and isMultiple is only true if we have an array in the search clause.
So we hand in false, false here, which will trigger the SINGLE_OPERATION hint

markuspf (approved):

mchacki:
The red test in Jenkins is unrelated, it tests internals of a different feature and has no shared code-path

PR #18022, opened 2023-01-26
clang-format some files that were flagged up by an experimental format of the codebase

markuspf:
### Scope & Purpose

clang-format some stragglers.


markuspf:
The discrepancies are caused by using clang-format-14 over clang-format-12; closing this for now as updates to newer clang-format can better be achieved by a PR once CircleCI setup is used.

PR #18024, opened 2023-01-26
[PREGEL] Use struct for pregel constants

jvolmer:
Combines all constant values of a pregel run in one struct to tidy up the conductor. For this, all theses constant value during a pregel run are specified in the feature, just before the conductor is created.

Additionally, I did some small cleanups like
- deleted redundant function input to Conductor::initializeWorkers
- deleted unused method Conductor::ensureUniqueResponse(VPackSlice)


markuspf (commented):
This looks good; 

I am not sure about the name "constants"; I agree that the values are constant for the lifetime of that Pregel execution, but I'd still call this "parameters" or "arguments"; then again, all names are similarly nondescriptive.

Maybe this is an `ExecutionSpecification`?

markuspf (approved):

markuspf (approved):
LGTM

PR #18028, opened 2023-01-27
This test could use some more bulknig.

dothebart:
### Scope & Purpose

bulk the testsetup part that doesn't set failurepoints to hopefully stabilize the test.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

dothebart:
```
tests\js\common\shell\shell-collection-noncluster.js
``` 
failed - close call language wise but other test.

PR #18030, opened 2023-01-27
[CINFRA] Linking shards with collection groups

apetenchea:
### Scope & Purpose

In order to get working on the `DocumentState`, we need some minimalist support for collection groups. Their introduction brings some breaking changes in the way we'll use replicated states from now on: a replicated state now corresponds to a _shardSheaf_ and handles multiple shards. This PR ensures that there is just enough base to extend the `DocumentState` for this purpose.

**Main changes**
1. The `shardSheaves` attribute of a `CollectionGroup` (in target), among with the `groupId` and `shardsR2` of a collection entry (in plan) are now written to the agency. This probably creates some minor conflicts with #18031 
2. When creating a new `CollectionGroup`, the ID of the replicated states corresponding to its shard sheaves are randomly generated. Thus, for collections created through the REST API, the mapping between a shard ID `s1234` and the corresponding state ID `1234` is no longer valid. However, for system collections, we still go through the `ClusterInfo` methods, which don't involve `CollectionGroups` yet. Therefore, for system collections, the old mapping remains. 
3. Due to the changes introduced at (2), we had to modify how collection creation and collection dropping works in `ClusterInfo`. In these methods we had to lookup the replicated state corresponding to each shard, and we have to handle both cases: when there's a collection group and when there's not. This also affected a couple of JS test suites, where we had to take collection groups into account when mapping shards to their replicated states. The next step is to let the supervision take care of the removal of replicated states.
4. Jobs such as `MoveShard` had to be adjusted too. Due to system collections not belonging to any `CollectionGroup`, we fallback to the old shardID to stateID mapping in case we can't find the corresponding group of a collection. This is mean to be temporary. Once we have full support for `CollectionGroups`, there will be no fallback option.
5. Part of `ClusterCollectionMethods::prepareCollectionGroups` has been temporarily disabled, such that every collection gets its own group, ignoring the `distributeShardsLike` attribute. This is because currently the `DocumentState` cannot handle more than one shard.
6. When the leader wants to replicate an operation, it has to lookup the corresponding replicated state of a `LogicalCollection`. We've added the `_replicatedStateId` property to that, set only for replication2 collections in `CreateCollection::first()`. I am not particularly happy with this solution, but for now it does the trick.

**Extra changes**

1. Removed replicated state JS functions that used obsolete agency paths.
2. A few small improvements to unrelated JS tests.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports


arangod/Cluster/ClusterInfo.cpp
```diff
@@ -4008,9 +4007,42 @@ Result ClusterInfo::dropCollectionCoordinator(  // drop collection
   auto replicatedStatesCleanup = futures::Future<Result>{std::in_place};
   if (coll->replicationVersion() == replication::Version::TWO) {
     std::vector<replication2::LogId> stateIds;
-    for (auto pair : VPackObjectIterator(shardsSlice)) {
-      auto shardId = pair.key.copyString();
-      stateIds.emplace_back(LogicalCollection::shardIdToStateId(shardId));
+    velocypack::Slice groupIdSlice = collectionSlice.get("groupId");
```
mchacki, line 4010: I am looking forward to be able to use Inspectors here, would make this a lot simpler.

We should pair again as soon as this PR is merged, and my PR to add Inspectors for Plan/Collection this can be really simplified \o/

arangod/Cluster/Utils/PlanCollectionEntryReplication2.cpp
```diff
@@ -66,17 +66,17 @@ PlanShardToServerMapping PlanCollectionEntryReplication2::getShardMapping()
...
 PlanCollectionEntryReplication2::getReplicatedLogForTarget(
-    ShardID const& shardId, ResponsibleServerList const& serverIds,
-    std::string_view databaseName) const {
+    replication2::LogId const& logId, ResponsibleServerList const& serverIds,
+    std::string_view databaseName, ShardID const& shardId) const {
   using namespace replication2::replicated_state;
 
   replication2::agency::LogTarget spec;
 
-  spec.id = LogicalCollection::shardIdToStateId(shardId);
```
mchacki, line 75: Is the shardIdToStateId function now obsolete?

Maybe we can already remove it now.

arangod/Replication2/AgencyCollectionSpecificationInspectors.h
```diff
@@ -31,7 +31,8 @@ template<class Inspector>
 auto inspect(Inspector& f, CollectionGroup& x) {
   return f.object(x).fields(f.field(StaticStrings::Id, x.id),
                             f.field("collections", x.collections),
-                            f.field("attributes", x.attributes));
+                            f.field("attributes", x.attributes),
```
mchacki, line 34: Expecting conflict with my work here, should be straight forward to solve.

tests/js/server/replication2/replication2-replicated-state-document-store.js
```diff
@@ -609,14 +623,18 @@ const replicatedStateDocumentStoreSuiteDatabaseDeletionReplication2 = function (
     tearDown: lh.registerAgencyTestEnd,
 
     testDropDatabase: function() {
+      /*
```
mchacki, line 626: Just for clarification:

This is commented out because right now we are missing one step on delete, of collections, that should be done on CollectionGroups, and need help from supervision to figure out that the ReplicatedState does not have any participant anymore correct?

mchacki (approved):
LGTM


I have one clarification question only.

tests/js/server/replication2/replication2-replicated-state-document-store.js
apetenchea: Eventually the supervision is going to figure out that the `CollectionGroup` is empty, and clean up all the replicated states corresponding to its shard sheaves. Until that is implemented, I have modified `ClusterInfo::dropCollectionCoordinator` to lookup the replicated states corresponding to the collection shards and drop them. This test is currently commented because dropping the database goes through `ClusterInfo::dropDatabaseCoordinator`, which I have not yet adapted.
I am going to provide a PR description soon, once I'm done with that.

arangod/Cluster/Utils/PlanCollectionEntryReplication2.cpp
apetenchea: It is still needed unfortunately, for `ClusterInfo` methods which rely on the old mapping between `shardId` and `logId`. After we manage to move that to collection groups, we'll safely remove it.

arangod/Cluster/ClusterInfo.cpp
apetenchea: Sounds great, looking forward to that!

PR #18031, opened 2023-01-27
[CINFRA] Feature/cinfra structs for collection groups in agency

mchacki:
### Scope & Purpose

Add Structs & Insepectors for Collections & Collection Groups in Agency Plan and Target.
They are not yet wired up, but the pathes where they should land in are defined.
So we can now start working the Supervision, as well as using those structs from createCollectionAPI independently.


maierlars (approved):
lgtm

apetenchea (approved):

PR #18034, opened 2023-02-01
[CINFRA-357] NodeLoadInspector

mpoeter:
### Scope & Purpose

This PR adds a node load inspector that allows to read of data structures from agency nodes.

This is currently a draft because it is based on the changes in #17283. Once that PR is merged I will rebase this branch onto devel.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**

### DIFF
--- a/arangod/Agency/Node.cpp
+++ b/arangod/Agency/Node.cpp
@@ -1119,14 +1119,14 @@ std::vector<std::string> Node::exists(
 }
 
 std::vector<std::string> Node::exists(std::string const& rel) const {
-  return exists(Store::split(rel));
+  return exists(split(rel));
 }
 
 bool Node::has(std::vector<std::string> const& rel) const {
   return exists(rel).size() == rel.size();
 }
 
-bool Node::has(std::string const& rel) const { return has(Store::split(rel)); }
+bool Node::has(std::string const& rel) const { return has(split(rel)); }
 
 std::optional<int64_t> Node::getInt() const noexcept {
   if (type() == NODE || !slice().isNumber<int64_t>()) {
--- a/arangod/Agency/Node.cpp
+++ b/arangod/Agency/Node.cpp
@@ -70,6 +70,46 @@ Node::Node(std::string const& name, Store* store)
 /// @brief Default dtor
 Node::~Node() = default;
 
+/// @brief Split strings by forward slashes, omitting empty strings,
+/// and ignoring multiple subsequent forward slashes
+std::vector<std::string> Node::split(std::string_view str) {
+  std::vector<std::string> result;
+
+  char const* p = str.data();
+  char const* e = str.data() + str.size();
+
+  // strip leading forward slashes
+  while (p != e && *p == '/') {
+    ++p;
+  }
+
+  // strip trailing forward slashes
+  while (p != e && *(e - 1) == '/') {
+    --e;
+  }
+
+  char const* start = nullptr;
+  while (p != e) {
+    if (*p == '/') {
+      if (start != nullptr) {
+        // had already found something
+        result.emplace_back(start, p - start);
+        start = nullptr;
+      }
+    } else {
+      if (start == nullptr) {
+        start = p;
+      }
+    }
+    ++p;
+  }
+  if (start != nullptr) {
+    result.emplace_back(start, p - start);
+  }
+
+  return result;
+}
+
 /// @brief Get slice to value buffer
 Slice Node::slice() const {
   // Some array
@@ -373,14 +413,14 @@ std::optional<std::reference_wrapper<Node const>> Node::get(
 }
 
 /// @brief lh-value at path
-Node& Node::getOrCreate(std::string const& path) {
-  return getOrCreate(Store::split(path));
+Node& Node::getOrCreate(std::string_view path) {
+  return getOrCreate(split(path));
 }
 
 /// @brief rh-value at path
 std::optional<std::reference_wrapper<Node const>> Node::get(
-    std::string const& path) const {
-  return get(Store::split(path));
+    std::string_view path) const {
+  return get(split(path));
 }
 
 // lh-store
@@ -1130,6 +1170,20 @@ bool Node::isString() const {
   return slice().isString();
 }
 
+bool Node::isArray() const {
+  if (type() == NODE) {
+    return false;
+  }
+  return _isArray;
+}
+
+bool Node::isObject() const {
+  if (type() == NODE) {
+    return true;
+  }
+  return false;
+}
+
 bool Node::isUInt() const {
   if (type() == NODE) {
     return false;
@@ -1251,6 +1305,13 @@ std::optional<std::string> Node::getString() const {
     return std::nullopt;
   }
   return slice().copyString();
+}
+
+std::optional<std::string_view> Node::getStringView() const {
+  if (type() == NODE || !slice().isString()) {
+    return std::nullopt;
+  }
+  return slice().stringView();
 }
 
 std::optional<Slice> Node::getArray() const {
--- a/arangod/Agency/Node.h
+++ b/arangod/Agency/Node.h
@@ -145,6 +145,10 @@ class Node final {
   /// @brief Child nodes
   typedef std::unordered_map<std::string, std::shared_ptr<Node>> Children;
 
+  /// @brief Split strings by forward slashes, omitting empty strings,
+  /// and ignoring multiple subsequent forward slashes
+  static std::vector<std::string> split(std::string_view str);
+
   /// @brief Construct with name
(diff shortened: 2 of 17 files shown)


arangod/Agency/Node.h
```diff
@@ -334,15 +344,18 @@ class Node final {
...
   /// @brief Get node specified by path string
   std::optional<std::reference_wrapper<Node const>> get(
-      std::string const& path) const;
+      std::string_view path) const;
 
   /// @brief Get string value (throws if type NODE or if conversion fails)
   std::optional<std::string> getString() const;
 
+  /// @brief Get string value (throws if type NODE or if conversion fails)
+  std::optional<std::string_view> getStringView() const;
```
MBkkt, line 357: Maybe you can return null string as nullopt?

arangod/Agency/Node.cpp
```diff
@@ -70,6 +70,46 @@ Node::Node(std::string const& name, Store* store)
 /// @brief Default dtor
 Node::~Node() = default;
 
+/// @brief Split strings by forward slashes, omitting empty strings,
+/// and ignoring multiple subsequent forward slashes
+std::vector<std::string> Node::split(std::string_view str) {
```
MBkkt, line 75: We can use absl::StrSplit instead of inventing wheel

arangod/Agency/NodeLoadInspector.h
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  template<class Func>
+  auto doProcessObject(Func&& func)
+      -> decltype(func(std::string_view(),
+                       std::declval<consensus::Node const*>())) {
+    assert(_node);
+    assert(_node->isObject());
+    for (auto& [k, v] : _node->children()) {
+      if (auto res = func(k, v.get()); not res.ok()) {
```
MBkkt, line 182: I think we have codestyle, why do you write code in other style?

Is it too difficult to write code in same style in same directory? :(

arangod/Agency/NodeLoadInspector.h
```diff
@@ -0,0 +1,311 @@
...
+    } else {
+      return {};
+    }
+  }
+
+  Status checkArrayLength(std::size_t arrayLength) {
+    auto slice = _node->getArray();
+    assert(slice != nullptr);
+    if (slice->length() != arrayLength) {
+      return {"Expected array of length " + std::to_string(arrayLength)};
```
MBkkt, line 298: StrCat

arangod/Agency/NodeLoadInspector.h
```diff
@@ -0,0 +1,311 @@
...
+
+  Status loadTypeField(std::string_view fieldName, std::string_view& result) {
+    auto v = _node->get(fieldName);
+    if (not v.has_value()) {
+      return {"Variant type field \"" + std::string(fieldName) +
+              "\" is missing"};
+    }
+    auto val = v->get().getStringView();
+    if (not val.has_value()) {
+      return {"Variant type field \"" + std::string(fieldName) +
```
MBkkt, line 242: StrCat

lib/Inspection/JsonPrintInspector.h
```diff
@@ -0,0 +1,263 @@
...
+      for (unsigned i = 0; i < IndentationPerLevel; ++i) {
+        _indentation.push_back(' ');
+      }
+    }
+  }
+
+  void decrementIndentationLevel() {
+    if (_format == JsonPrintFormat::kPretty) {
+      assert(_indentation.size() >= IndentationPerLevel);
+      for (unsigned i = 0; i < IndentationPerLevel; ++i) {
```
MBkkt, line 249: Please use resize

lib/Inspection/JsonPrintInspector.h
```diff
@@ -0,0 +1,263 @@
...
+      if (++it != end) {
+        _stream << ',' << _linebreak;
+      }
+    }
+    return {};
+  }
+
+  void incrementIndentationLevel() {
+    if (_format == JsonPrintFormat::kPretty) {
+      for (unsigned i = 0; i < IndentationPerLevel; ++i) {
```
MBkkt, line 240: Please use append

arangod/Agency/Node.h
mpoeter: I prefer to stick to the convention used in this file.

arangod/Agency/NodeLoadInspector.h
mpoeter: Why would I want to introduce a dependency on abseil just for creating an error message?

lib/Inspection/JsonPrintInspector.h
mpoeter: Good call, I did not know that there was an append overload for this.

arangod/Agency/Node.h
MBkkt: Ok

arangod/Agency/NodeLoadInspector.h
MBkkt: Why introduce?
We already depends, if sometime for some reason we decide we don't want replace StrCat by any custom function is easy.

mpoeter: No, this code does not. For example, I locally have a small test executable that only contains the inspector tests, and that executable currently has no dependency on abseil, and I see absolutely no reason to add one.

What exactly would be the advantage of using StrCat here? This is a very unlikely error case that definitely does not warrant any kind of optimization (performance or otherwise).

arangod/Agency/NodeLoadInspector.h
mpoeter: I am not sure if we actually have this in our coding guidelines. Either way, we have quite some code that uses word operators, and I find it much more readable. In particular a single `!` can easily be overlooked. You are right, that the style in this file/directory is not fully consistent, because some of it is older code. I can happily update everything to use word operators though.

lib/Inspection/JsonPrintInspector.h
mpoeter: Fixed in #17283

arangod/Agency/NodeLoadInspector.h
MBkkt: > No, this code does not.

It is agency, not lib inspection, agency depends on arangodb library, which depends on abseil (because of few reasons s2, hashtables, strings, mutex)

So why doesn't?

> What exactly would be the advantage of using StrCat here?

It simpler (don't need to_string, which can allocate with libstdc++ (sso size 15)).

Also if we always will concat strings via abseil, it will be unified in all our code, smaller chance to write inefficient code in not error path

arangod/Agency/NodeLoadInspector.h
MBkkt: > and I find it much more readable

For me any is ok, but in general seeing not/and/or in c++ code is very confusing, I always knew it works, but almost no one write like this. Personally I know only replication 2 team which start to write like this in last year

arangod/Agency/NodeLoadInspector.h
mpoeter: Yes, this file currently lives in Agency instead of lib/inspection, but I am not sure if it will stay there. Nevertheless, this particular header file currently has no dependency on abseil, and since this is an unlikely error case I see not reason for "efficiency optimizations". Regarding simplicity - syntactically we trade one function call for another, so I don't see the big gain here.

MBkkt: > but I am not sure if it will stay there

Maybe todo in such case?
About moving it out

About discussion at all.
I think in agency, and other arangod parts, abseil dependency completely ok, because we have more largest stuff, like v8, rocksdb, etc.
So build time for absl::strings is not the issue.

At the same time, code looks at least the same or better (because you don't need to explicit call for each numeric argument). And works also better (I agree it's commonly not important, but it's still true) 

So I think for implementation, not lib code, using abseil almost always better than manual concat.

Anyway fill free to resolve it, if I didn't convince you or you thinking about move this code out of agency

arangod/Agency/NodeLoadInspector.h
mpoeter: Yes, I also used to write "classic" C++ code, but in particular the logical not operator `!` often caused confusion because I misread some condition. A lot of other languages (Ruby, Python, ...) use word operators which I find much more readable. And since we also have those in C++, I would like to use them. But I agree that we should try to establish a common style.

maierlars (approved):

maierlars:
error unrelated and fixed by https://github.com/arangodb/arangodb/pull/18044

PR #18046, opened 2023-02-06
launch it printing version information

dothebart:
### Scope & Purpose

we don't have any tests outputting our version - launch arangosh printing its version details.

### DIFF
--- a/js/client/modules/@arangodb/testsuites/arangosh.js
+++ b/js/client/modules/@arangodb/testsuites/arangosh.js
@@ -190,7 +190,7 @@ function arangosh (options) {
           'Starting arangosh printing the version:',
           '',
           0,
-          {'version': true});
+          {'version': 'true'});
   print();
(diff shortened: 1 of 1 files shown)


js/client/modules/@arangodb/testsuites/arangosh.js
```diff
@@ -185,6 +186,13 @@ function arangosh (options) {
     tmpMgr.destructor(ret[section]['status'] && options.cleanup);
   }
 
+  runTest('testArangoshExitVersion',
+          'Starting arangosh printing the version:',
+          '',
+          0,
+          {'version': true});
```
jsteemann, line 193: Probably works as is, but IMO all command-line arguments will end up as strings anyway.
```suggestion
          {'version': 'true'});
```

jsteemann (approved):
LGTM

PR #18065, opened 2023-02-07
properly handle the covearge switch via cmake

dothebart:
### Scope & Purpose

use the standard cmake way of handling external switches to enable code coverage


markuspf (commented):
I wonder whether this change is better added as a CMakePreset; there are currently presets for asan, ubsan and tsan.

Adding a preset for coverage seems more consistent.

Regardless of whether this change is just made into a preset, I would quite like to see a `-coverage` preset in `CMakePresets.json`

CMakePresets.json
```diff
@@ -83,6 +83,16 @@
         "USE_JEMALLOC": "Off"
       }
     },
+    {
+      "hidden": true,
+      "name": "gcov",
+      "cacheVariables": {
+        "USE_COVERAGE": "On",
+        "CMAKE_CXX_FLAGS": "-fno-stack-protector -fprofile-arcs -ftest-coverage",
+        "CMAKE_C_FLAGS": "-fno-stack-protector -fprofile-arcs -ftest-coverage",
```
markuspf, line 92:
```suggestion
```
I now think it is best to leave the option in `CMakeLists.txt`, but certainly to remove these two lines.

Rationale: This way we could maybe switch coverage on only for *our* code, as I believe at the moment we just enable coverage for *everything*.

tests/js/common/shell/shell-index.js
```diff
@@ -29,7 +29,7 @@
 const jsunity = require("jsunity");
 const internal = require("internal");
 const errors = internal.errors;
-const testHelper = require("@arangodb/test-helper").Helper;
+const { Helper, versionHas } = require("@arangodb/test-helper");
```
jsteemann, line 32: why `Helper` with a capital `H`?
```suggestion
const { helper, versionHas } = require("@arangodb/test-helper");
```

jsteemann (approved):
LGTM

tests/js/common/shell/shell-index.js
dothebart: oh, this will not work. As long as the function/object  comes from inside of testhelper, it must keep the name

CMakeLists.txt
```diff
@@ -202,6 +202,13 @@ if (USE_DTRACE)
   add_definitions("-DUSE_DTRACE=1")
 endif ()
 
+# enable code coverage
+option(USE_COVERAGE "enable gcov code coverage" OFF)
+if (USE_COVERAGE)
+  add_compile_options(-fno-stack-protector -fprofile-arcs -ftest-coverage)
```
markuspf, line 208: Why `-fno-stack-protector` for coverage?

dothebart: I did not question
https://github.com/arangodb/oskar/blob/master/scripts/buildAlpine6.fish#L58

if its not needed, we can remove it.

KVS85 (approved):

CMakeLists.txt
KVS85: This is a copy of current options unrelated to functional changes of this PR.

PR #18067, opened 2023-02-07
[CINFRA] Collection Group Supervision

maierlars:
### Scope & Purpose
This PR adds a minimal working version of the supervision for collection groups. It is far from perfect and will be optimized in future PRs. The primary goal is to make it work.


maierlars:
Remaining tests are unrelated.

PR #18068, opened 2023-02-08
improve memory usage of in-memory edge cache

jsteemann:
### Scope & Purpose

Improve memory usage of in-memory edge cache. This is achieved by compressing the ids of lookup keys and mapped values (i.e. `_from` and `_to` values) in case the collection name in `_from` or `_to` matches the collection name cached by the in-memory index.
That cached collection name is automatically determined by the first operation on the in-memory edge index by the first `_from` / `_to` values used.
For the index on `_from`, the collection name part of the first lookup value will be used as prefix of cache keys, and the collection name part of the first mapped `_to` value will be used a prefix of cached values.
For the index on `_to`, the collection name part of the first lookup value will be used as prefix of cache keys, and the collection name part of the first mapped `_from` value will be used a prefix of cached values.
That means the cached collection names in `_from` and `_to` are independent. 
The compressed `_from` and `_to` values will only be used in memory, but will never be stored as such in RocksDB.

The optimization can help to reduce memory usage of the in-memory edge cache if many edges share the initially used (cached) collection name. It should not help much if many different edge collections are used.
Currently the optimization is enabled automatically and cannot be turned off.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18121
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -236,39 +273,54 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
-        LocalDocumentId docId{
-            _builderIterator.value().getNumericValue<uint64_t>()};
-        _builderIterator.next();
-        TRI_ASSERT(_builderIterator.valid());
-        // For now we store the complete opposite _from/_to value
-        TRI_ASSERT(_builderIterator.value().isString());
+    std::string const* cacheKeyCollection = nullptr;
+    std::string const* cacheValueCollection = nullptr;
+
+    fu2::unique_function<void()> handleSingleResult = [&, this]() {
```
MBkkt, line 289: Why not auto?

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -971,3 +1070,90 @@ void RocksDBEdgeIndex::recalculateEstimates() {
...
+    }
+
+    // here we must have a collection name
+    TRI_ASSERT(previous != nullptr);
+
+    // now check if the collection name in the value we got matches the cached
+    // name.
+    TRI_ASSERT(!previous->empty());
+    // must have at least 'c/k'
+    TRI_ASSERT(value.size() > 2);
```
MBkkt, line 1140: You have code above which assign `{}` to value.

jsteemann: Correct, but that assignment `value = {}` is in a different branch:
```
  if (pos == std::string_view::npos || pos == 0 || pos + 1 == value.size() ||
      value[pos + 1] == '/') {
    value = {};
```
Here we are in that `if`'s `else` block, so it does not matter.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -971,3 +1070,90 @@ void RocksDBEdgeIndex::recalculateEstimates() {
...
+  }
+  TRI_ASSERT(previous != nullptr);
+  return previous;
+}
+
+RocksDBEdgeIndex::CachedCollectionName::CachedCollectionName() noexcept
+    : _name(nullptr) {}
+
+RocksDBEdgeIndex::CachedCollectionName::~CachedCollectionName() {
+  delete _name.load(std::memory_order_relaxed);
```
MBkkt, line 1097: One thread create string, another destroy string, formally you need happens before (maybe it's done by another synchronization) and in practice it doesn't really matter.
But I believe honest acquire will be better for this, because it's doesn't look like performance critical.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -971,3 +1070,90 @@ void RocksDBEdgeIndex::recalculateEstimates() {
...
+      TRI_ASSERT(value[1] != '/');
+    }
+  }
+
+  return value;
+}
+
+std::string const* RocksDBEdgeIndex::CachedCollectionName::get()
+    const noexcept {
+  return _name.load(std::memory_order_relaxed);
```
MBkkt, line 1158: I don't think relaxed is correct here. See other comments

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -282,57 +334,65 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
-
-              _builderIterator.next();
-              limit--;
+        // try to read from cache
+
+        // build actual cache lookup key. this is either the value of fromTo,
+        // or a suffix of it in case fromTo starts with the cached collection
+        // name, and an empty string_view if the lookup value is invalid.
+        cacheKey = _index->buildCompressedCacheKey(cacheKeyCollection, fromTo);
+        TRI_ASSERT(cacheKey.empty() || cacheKey == fromTo ||
```
MBkkt, line 347: NIT
Maybe move assert under !cacheKey.empty()?

In such case first part of disjunction can be removed

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -282,57 +334,65 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
+            // We got sth. in the cache
+            VPackSlice cachedData(finding.value()->value());
+            TRI_ASSERT(cachedData.isArray());
+            if (cachedData.length() / 2 < limit) {
+              // Directly return it, no need to copy
+              _builderIterator = VPackArrayIterator(cachedData);
+              while (_builderIterator.valid()) {
+                handleSingleResult();
+                _builderIterator.next();
+                TRI_ASSERT(limit > 0);
```
MBkkt, line 369: Could you explain this assert for me, please? Why is limit == 0 not ok here?

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -282,57 +334,65 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
+          auto finding = _cache->find(cacheKey.data(),
+                                      static_cast<uint32_t>(cacheKey.size()));
+
+          if (finding.found()) {
+            incrCacheHits();
+            needRocksLookup = false;
+            // We got sth. in the cache
+            VPackSlice cachedData(finding.value()->value());
+            TRI_ASSERT(cachedData.isArray());
+            if (cachedData.length() / 2 < limit) {
```
MBkkt, line 363: Do I understand correctly cachedData is always even length?
Maybe add such assert?

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -236,39 +273,54 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
+      // For now we store the complete opposite _from/_to value
+      VPackSlice fromTo = _builderIterator.value();
+      TRI_ASSERT(fromTo.isString());
+
+      std::string_view v = fromTo.stringView();
+      TRI_ASSERT(!v.empty());
+      if (v.front() == '/') {
+        // prefix-compressed value
+        cacheValueCollection =
+            _index->cachedValueCollection(cacheValueCollection);
```
MBkkt, line 304: I don't understand why this is needed.

If I see correct it will do load only for case when cacheValueCollection is nullptr.
But it's true only if it's first call for cacheValueCollection.
If it's possible I don't understand why is cache in _index always already filled (because return value is never null, at least in assert).


Anyway, why don't move this load from first handleSingleResult call to this line
```
std::string const* cacheValueCollection = nullptr;
```

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
MBkkt: Hmm
> maybe it's done by another synchronization

This probably should always to be true, because otherwise we don't have hb between dtor and calling method

arangod/Aql/DocumentProducingHelper.h
```diff
@@ -138,21 +139,21 @@ struct DocumentProducingFunctionContext {
...
   /// @brief Builder that is reused to generate projection results
-  arangodb::velocypack::Builder _objectBuilder;
+  velocypack::Builder _objectBuilder;
 
   /// @brief set of already returned documents. Used to make the result distinct
   containers::FlatHashSet<LocalDocumentId> _alreadyReturned;
 
-  RegisterId const _outputRegister;
   Variable const* _outputVariable;
+  RegisterId const _outputRegister;
```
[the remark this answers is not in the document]
jsteemann: It is an unrelated reordering of members to make the type more compact.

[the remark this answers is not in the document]
MBkkt: Hmm is RegisterId has other alignment?

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: Because we are subtracting 1 from it. If `limit == 0`, then it would underflow. That is exactly what the assertion should uncover if it happens.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: The idea was to do this only lazily, when it is actually needed to perform the atomic load.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
MBkkt: Overflow is defined for unsigned integer.

Anyway my question more about why if I have limit like 1000
Is not possible to produce exactly 1000 documents.

Oh I probably understand, it's because `< limit`, so our limit at least here is not included? `[0; limit)`

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
MBkkt: I mean maybe just do always single load is better?

It's single load in all cases compared to 0 load in one cases and 1 load + n conditions in other.

Probably compiler can optimize conditions if you remove type erasure.
And probably branch prediction is very good.

But it looks like too much "probably" for single load (per function at all, not per handleSingleResult) in other cases.

Is it really optimize something?

Maybe I missed something

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: We don't want the limit value to underflow, even if the underflow is defined here.
That is the motivation for the assert.

The entire `while` loop in addition is only executed if the condition `if (cachedIterator.size() / 2 < limit)` is true. That's why the limit will not go to 0 before the subtraction is happening.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: The values of `cacheKeyCollection` and `cacheValueCollection` are data-dependent.
For example, `cacheKeyCollection` will be populated with the collection name of the edge on the first lookup in the index.
We really don't want to call this if the from/to value is invalid, if we don't have the cache enabled etc. That is also a reason why it is called lazily.

MBkkt (approved):

mpoeter (approved):

arangod/Aql/DocumentProducingHelper.h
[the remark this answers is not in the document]
jsteemann: `RegisterId` has the size of a uint32_t.
So having a `RegisterId` and then a pointer and then a bool would waste space.
Rearranging them so that we have a pointer, then a uint32_t-sized type and then a boolean is likely better.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -971,3 +1069,86 @@ void RocksDBEdgeIndex::recalculateEstimates() {
...
+
+std::string_view RocksDBEdgeIndex::CachedCollectionName::buildCompressedValue(
+    std::string const*& previous, std::string_view value) const {
+  // split lookup value to determine collection name and key parts
+  auto pos = value.find('/');
+
+  if (pos == std::string_view::npos || pos == 0 || pos + 1 == value.size() ||
+      value[pos + 1] == '/') {
+    // totally invalid lookup value
+    value = {};
```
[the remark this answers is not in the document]
jsteemann: done

PR #18075, opened 2023-02-09
wip: improve memory usage of in-memory edge cache

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/18068

Improve memory usage of in-memory edge cache. This is achieved by compressing the ids of lookup keys and mapped values (i.e. `_from` and `_to` values) in case the collection name in `_from` or `_to` matches the collection name cached by the in-memory index.
That cached collection name is automatically determined by the first operation on the in-memory edge index by the first `_from` / `_to` values used.
For the index on `_from`, the collection name part of the first lookup value will be used as prefix of cache keys, and the collection name part of the first mapped `_to` value will be used a prefix of cached values.
For the index on `_to`, the collection name part of the first lookup value will be used as prefix of cache keys, and the collection name part of the first mapped `_from` value will be used a prefix of cached values.
That means the cached collection names in `_from` and `_to` are independent. 
The compressed `_from` and `_to` values will only be used in memory, but will never be stored as such in RocksDB.

The optimization can help to reduce memory usage of the in-memory edge cache if many edges share the initially used (cached) collection name. It should not help much if many different edge collections are used.
Currently the optimization is enabled automatically and cannot be turned off.
We have to run performance tests to see if the change can actually make performance worse in some case (e.g. many different edge collections). We also have to check how far this change can reduce memory usage of the in-memory edge index.
There is also currently some inefficiency while building back the full `_from` / `_to` value from a compressed key / value, which very likely can be improved with a bit of effort.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
+++ b/arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
@@ -45,7 +45,6 @@
 #include "RocksDBEngine/RocksDBIndexCacheRefillFeature.h"
 #include "RocksDBEngine/RocksDBKey.h"
 #include "RocksDBEngine/RocksDBKeyBounds.h"
-#include "RocksDBEngine/RocksDBSettingsManager.h"
 #include "RocksDBEngine/RocksDBTransactionMethods.h"
 #include "RocksDBEngine/RocksDBTransactionState.h"
 #include "RocksDBEngine/RocksDBTypes.h"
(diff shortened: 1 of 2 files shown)


jsteemann:
Obsolete, needs to be started over.

PR #18076, opened 2023-02-09
feat: sort collections in user permissions view [FE-133]

palashkaria:
### Scope & Purpose

Currently, collections on user permissions tab are sorted in the order of creation (because `collections` is an object, which [maintains enumeration order](https://stackoverflow.com/questions/39054764/show-original-order-of-object-properties-in-console-log/39055008#39055008) in some JS engines). This PR sorts the collections in alphabetical order by the collection name.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested
#### Related Information
- [x] Jira ticket: https://arangodb.atlassian.net/browse/FE-133

[x] Current behaviour - no sorting:
image

[x] Updated behaviour - sort alphabetically:
image


hkernbach (approved):
LGTM. Only very minor comment: 

Readability in "userPermissionView.ejs" decreases slightly a little bit as we do use index access when accessing the collections array (e.g. permissionCollection[1] or permissionCollection[0] - which is the collection name). We could tmp store them in meaningful variable names - but this is only optional. Technically all good.

palashkaria:
> LGTM. Only very minor comment:
> 
> Readability in "userPermissionView.ejs" decreases slightly a little bit as we do use index access when accessing the collections array (e.g. permissionCollection[1] or permissionCollection[0] - which is the collection name). We could tmp store them in meaningful variable names - but this is only optional. Technically all good.

I agree, and I have made the changes for the same. This also results in a much smaller diff. Thanks for the suggestion @hkernbach !

ajurdak (approved):
LGTM.

PR #18072, opened 2023-02-08
Make gharial transactions globally managed [BTS-1220].

neunhoef:
### Scope & Purpose

This is about https://arangodb.atlassian.net/browse/BTS-1220.

See details there.

A transaction on a coordinator which does not use the
transaction::Manager or is a top level AQL transaction, will not
be "globally managed", rather it will be "legacy" and can therefore lead
to discrepancies in leader and follower data if an abort is
forced at some late stage when the followers have already committed.
The problem is that for such a legacy transaction no transaction headers are sent
along with requests to leaders (and as a consequence to followers), such that
all operations will immediately commit, in particular on followers. If then something
goes wrong on the leader after something has already be committed on the follower
(for example a conflict which leads to a transaction abort), we have a silent discrepancy
between leader and follower shards.

This PR tries to fix this by finding all places in the code which
need an additional hint `GLOBAL_MANAGED` to force proper distributed behaviour.

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **integration tests**
- [*] :book: CHANGELOG entry made
- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18077
  - [ ] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18078
  - [*] Backport for 3.8: not done

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1220
- [*] Enterprise PR (for tests): https://github.com/arangodb/enterprise/pull/1203

PR #18078, opened 2023-02-09
Make gharial transactions globally managed [BTS-1220].

neunhoef:
### Scope & Purpose

This is about https://arangodb.atlassian.net/browse/BTS-1220.
See details there.

This is the backport for 3.9, original PR: https://github.com/arangodb/arangodb/pull/18072

A transaction on a coordinator which does not use the
transaction::Manager or is a top level AQL transaction, will not
be "globally managed", rather it will be "legacy" and can therefore lead
to discrepancies in leader and follower data if an abort is
forced at some late stage when the followers have already committed.
The problem is that for such a legacy transaction no transaction headers are sent
along with requests to leaders (and as a consequence to followers), such that
all operations will immediately commit, in particular on followers. If then something
goes wrong on the leader after something has already be committed on the follower
(for example a conflict which leads to a transaction abort), we have a silent discrepancy
between leader and follower shards.

This PR tries to fix this by finding all places in the code which
need an additional hint `GLOBAL_MANAGED` to force proper distributed behaviour.

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.9: This is it.

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1220
- [*] Enterprise PR (for tests): https://github.com/arangodb/enterprise/pull/1204

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,10 +1,10 @@
 v3.9.9 (XXXX-XX-XX)
 -------------------
 
-* Made all transactions used by the gharial API on coordinators and a few
-  others marked "globally managed". This fixes an issue where
-  transaction conflicts could lead to a silent out of sync situation
-  between a leader shard and its followers.
+* Made all transactions used by the gharial API on coordinators and a few others
+  marked "globally managed". This fixes an issue where transaction conflicts
+  could lead to a silent out of sync situation between a leader shard and its
+  followers.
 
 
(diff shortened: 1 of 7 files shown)


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #18080, opened 2023-02-09
[CINFRA-672] Use explicit shards for replicated operations.

maierlars:
### Scope & Purpose
Use explicit shards for replicated operations.

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
@@ -119,7 +119,7 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
       }
     }
 
-    auto doc = DocumentLogEntry{std::string(self->shardId),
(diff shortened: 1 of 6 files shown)


arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -119,7 +119,7 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
       }
     }
 
-    auto doc = DocumentLogEntry{std::string(self->shardId),
+    auto doc = DocumentLogEntry{"",  // this affects all shards
```
mpoeter, line 122: We should consider making DocumentLogEntry a variant.

mpoeter (approved):

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
maierlars: We agreed to do this in a separate PR.

PR #18085, opened 2023-02-09
[testing]: it seems ARM tends to be slower on shutdown; lets give it some more time

dothebart:
### Scope & Purpose

we had ARM testruns where subsequent launches would overlap with old runs. give it more time.


jsteemann (approved):
LGTM

mpoeter (approved):

PR #18087, opened 2023-02-10
feat: adds query search [FE-131]

palashkaria:
- adds a searchOptions to the query collection (to be used for sorting later too)
- triggers search on input change and on keydown

### Scope & Purpose

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-131

Screenshots:

image

image (1)


ajurdak (approved):
LGTM.

hkernbach (approved):
Code LGTM. 

Two minor things:
1.) 
image

Search box height does not align to the height of the button. 

2.) `restrictToSearchPhraseKey` and `restrictToSearchPhrase` (search mechanism) looks familiar. Somewhen - must not be this PR - we should move this code (as we do have it on multiple places) to a single dedicated location where it can be reused. Every view does have it's own added logic. This logic we could pass via. callbacks. Reason: In cases we do have issues here, we might need to adjust multiple views. With a single place, we will improve maintainability. But as I said already - very minor right now. Something for later. If we start the transition from backbone js related views to react based - we should not implement my suggestion at all.

palashkaria:
> Code LGTM.
> 
> Two minor things: 1.) image
> 
> Search box height does not align to the height of the button.
> 
> 2.) `restrictToSearchPhraseKey` and `restrictToSearchPhrase` (search mechanism) looks familiar. Somewhen - must not be this PR - we should move this code (as we do have it on multiple places) to a single dedicated location where it can be reused. Every view does have it's own added logic. This logic we could pass via. callbacks. Reason: In cases we do have issues here, we might need to adjust multiple views. With a single place, we will improve maintainability. But as I said already - very minor right now. Something for later. If we start the transition from backbone js related views to react based - we should not implement my suggestion at all.

Noted, good suggestions. I'd refrain from doing this right now and wait for a transition to React! But will keep this in mind in case we touch these files again

PR #18089, opened 2023-02-10
Make geo index cheap again [BTS-1219].

neunhoef:
This addresses BTS-1219: https://arangodb.atlassian.net/browse/BTS-1219

We found that 3.10 has improved an optimizer rule `move-filters-into-enumerate`. Furthermore, the cost computation of a full collection enumeration with included filter was wrong. Together with the fact that the geo index did not properly advertise its cost savings the query optimizer suddenly preferred a full table scan over the usage of a geo index for some query with nested FOR loops.

The fix consists of:
 - fix the cost computation for full collection scans with and without included filter
 - make the geo index reduce its cost estimation to better advertise its capabilities to the optimizer

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.10: This is the backport for 3.10

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1219
- [*] Original PR for devel: https://github.com/arangodb/arangodb/pull/18095


tests/js/server/aql/aql-queries-geo.js
```diff
@@ -1358,7 +1358,36 @@ function geoLegacyComparison() {
...
+    },
+
+    testNestedFor : function() {
+      let query = `FOR e IN ${cn}
+                     FOR d IN ${cn}
+                       FILTER GEO_INTERSECTS(d.geo, e.geo)
+                       RETURN {d: d._key, e: e._key}`;
+      let compact = helper.getCompactPlan(AQL_EXPLAIN(query))
+        .map(function(node) { return node.type; });
+      assertEqual([ "SingletonNode", "EnumerateCollectionNode",
```
jsteemann, line 1388: IMO we should at least check here that the selected index is the expected geo index, and not for example the primary index.

tests/js/server/aql/aql-queries-geo.js
```diff
@@ -1358,7 +1358,36 @@ function geoLegacyComparison() {
...
+      db._drop(cn);
+      coll = db._create(cn);
+      coll.ensureIndex({type:"geo", fields:["geo"], geoJson: true});
+    },
+
+    tearDown : function() {
+      db._drop(cn);
+    },
+
+    testNestedFor : function() {
```
jsteemann, line 1381: I also suggest adding a additional tests that check that the geo index is used for varying collection sizes (small, medium, large), and also when using some filter conditions on the collection. This is because index cost formulas may take into account the number of to-be-processed documents.
Another question is also which index should be used if there are other candidate indexes, e.g.
```
FOR e IN ${cn}
  FOR d IN ${cn}
    FILTER GEO_INTERSECTS(d.geo, e.geo)
    FILTER d.value == @value 
    RETURN ...
```
when there is a persistent index present on `value`.

tests/js/server/aql/aql-queries-geo.js
neunhoef: Indeed, done.

tests/js/server/aql/aql-queries-geo.js
neunhoef: Indeed, I have added more tests.

neunhoef: Done.

jsteemann (approved):
LGTM

mchacki (approved):
LGTM

PR #18102, opened 2023-02-13
[CINFRA-656] Maintenance drop shard on leader + Supervision wait for Current

maierlars:
### Scope & Purpose
This PR does two things:
1. It adds dropping a shard using the maintenance via the Document state leader. 
2. It adds an additional check to the supervision to wait for a collection to appear in Current.


arangod/Cluster/Maintenance.cpp
```diff
@@ -466,8 +466,10 @@ static void handleLocalShard(
     if (replicationVersion != replication::Version::TWO) {
```
apetenchea, line 466: Activating this PR would mean an `|| isLeading` inside this conditional, as far as I see.

apetenchea (approved):
LGTM! I get that it's currently blocked, but I think it's ok to merge it if you want. Might make merge conflicts easier to fix along the way.

arangod/Cluster/Maintenance.cpp
maierlars: 😵‍💫 why is it not in this PR but was in a different one...

apetenchea (approved):

maierlars:
Last error is unrelated to this PR.

PR #18107, opened 2023-02-14
Replace JSON.parse(response.body) by response.parsedBody in HTTP examples

Simran-B:
Avoid unnecessary parsing of the payload body in DocuBlocks.

Also adds a missing logJsonResponse() to delete cursor example.


dothebart (approved):
good catch. this should be available in all current stable branches.

ansoboleva (approved):

ansoboleva:
verified with jenkins https://jenkins.arangodb.biz:3456/view/Documentation/job/arangodb-ANY-examples/2873/

PR #18109, opened 2023-02-14
[CINFRA-673] Multi shard snapshot transfer

mpoeter:
### Scope & Purpose

Prepare for transferring multiple shards during snapshot transfer.

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**

#### Related Information

- [x] Jira ticket: https://arangodb.atlassian.net/browse/CINFRA-673

### DIFF
--- a/arangod/Replication2/StateMachines/Document/CollectionReader.cpp
+++ b/arangod/Replication2/StateMachines/Document/CollectionReader.cpp
@@ -77,11 +77,11 @@ CollectionReader::CollectionReader(
     std::shared_ptr<LogicalCollection> logicalCollection,
     ReplicationTransaction& trx)
     : _logicalCollection(std::move(logicalCollection)) {
-  Result res = trx.addCollection(*logicalCollection);
+  Result res = trx.addCollection(*_logicalCollection);
 
   if (!res.ok()) {
     LOG_TOPIC("c49e5", ERR, Logger::REPLICATION2)
-        << "Failed to add collection " << logicalCollection->name()
+        << "Failed to add collection " << _logicalCollection->name()
         << " to a transaction: " << res.errorMessage();
     THROW_ARANGO_EXCEPTION(res);
   }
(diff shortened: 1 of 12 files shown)


maierlars (approved):
LGTM

arangod/Replication2/StateMachines/Document/CollectionReader.cpp
```diff
@@ -28,31 +28,66 @@
...
+auto ReplicationTransaction::options() -> transaction::Options {
   transaction::Options options;
   options.requiresReplication = false;
+  options.fillBlockCache = false;  // TODO - check whether this is a good idea
+  return options;
+}
+
+auto ReplicationTransaction::addCollection(LogicalCollection const& collection)
+    -> Result {
+  return transaction::Methods::addCollection(collection.id(), collection.name(),
```
apetenchea, line 49: I had a quick look at the Jenkins failure, and this seems to be the culprit. Unfortunately, we cannot call `addCollection` after the transaction has status `CREATED`, which happens after it is constructed. We might have to create a transaction for each shard, which clearly would be less elegant than what you were trying to achieve here, or find some other workaround.

apetenchea (changes requested):

arangod/Replication2/StateMachines/Document/CollectionReader.cpp
mpoeter: We definitely should be able to add readonly collections to an already startet trx. Will look into this!

apetenchea: When we try to do that, an exception is thrown in `Methods.cpp:3001`: "cannot add collection to a previously started top-level transaction". But, that can work with `addCollectionAtRuntime`. I don't understand the inner workings of these methods either, just tried it out and it works.

apetenchea (approved):
LGTM!

PR #18115, opened 2023-02-14
ES-1428: max number of V8 contexts depends on max threads

jsteemann:
### Scope & Purpose

https://arangodb.atlassian.net/browse/ES-1428
https://arangodb.atlassian.net/browse/FRB-57

* ES-1428: make the maximum number of V8 contexts depend on the maximum number of server threads, if `--javascript.v8-contexts` is not set explicitly. Previously the maximum number of V8 contexts was hard-coded to 16 when the option `--javascript.v8-contexts` option was not set explicitly. Now the maximum number defaults to the value of `--server.maximal-threads`, regardless of if this was is explicitly configured or the default value is used. A server with default configuration should now not block waiting for V8 contexts to become available, but it may use more memory for the additional V8 contexts if there are many concurrent that invoke JavaScript actions (e.g. using the web UI or Foxx).

It seems that no documentation PR is needed for 3.10 and devel, as the documentation for startup options is auto-generated from the ArangoDB source code.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: 
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


dothebart:
can you please relax that number for mac as well?

jsteemann:
> can you please relax that number for mac as well?

Test didn't fail on macOS yet. Is it worth adjusting the test once more?

dothebart:
do you actively watch it on macos? the CI doesn't run any .js tests on mac.

jsteemann:
> do you actively watch it on macos? the CI doesn't run any .js tests on mac.

Well, if we don't run this test on macOS, why should it then be adjusted for macOS? :thinking:

dothebart:
Nighlies run it.
This was one of the ~dozend tests that drove it to the limit.
Mac machines are small; less core count (<20 vs. > 32 for Windows & Linux), less ram.

dothebart (approved):

PR #18123, opened 2023-02-15
[CINFRA-679] Document State Tracks Multiple Shards

apetenchea:
### Scope & Purpose

We are adding support for multiple shards in `DocumentCore`. Until now, we used to have the `shardId` property for referencing the corresponding shard. This PR replaces that with a map, *ShardId -> ShardProperties*.
Because on the followers we create and drop shards through the use of log entries, we are now in a situation in which, during a replay of the log, a transaction might reference a dropped shard. Or, the follower tries to create an already existing shard. The document state checks for the existence of a shard before applying any transaction.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**

### DIFF
--- a/arangod/Cluster/CreateCollection.cpp
+++ b/arangod/Cluster/CreateCollection.cpp
@@ -273,7 +273,8 @@ bool CreateCollection::createReplication2Shard(CollectionID const& collection,
         replication2::replicated_state::document::DocumentLeaderState>(
         state.get()->getLeader());
     if (leaderState != nullptr) {
-      // TODO we are blocking the maintenance here, but we should not
+      // It is necessary to block here to prevent creation of an additional
+      // action while we are waiting for the shard to be created.
       leaderState
           ->createShard(
               shard, collection,
--- a/arangod/Cluster/CreateCollection.cpp
+++ b/arangod/Cluster/CreateCollection.cpp
@@ -273,9 +273,12 @@ bool CreateCollection::createReplication2Shard(CollectionID const& collection,
         replication2::replicated_state::document::DocumentLeaderState>(
         state.get()->getLeader());
     if (leaderState != nullptr) {
-      leaderState->createShard(
-          shard, collection,
-          velocypack::SharedSlice(_description.properties()->bufferRef()));
+      // TODO we are blocking the maintenance here, but we should not
+      leaderState
+          ->createShard(
+              shard, collection,
+              velocypack::SharedSlice(_description.properties()->bufferRef()))
+          .get();
     }
   }
 
--- a/arangod/Replication2/StateMachines/Document/DocumentCore.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentCore.cpp
@@ -53,7 +53,12 @@ DocumentCore::DocumentCore(
   TRI_ASSERT(shardResult.ok()) << "Shard creation failed for replicated state"
                                << _gid << ": " << shardResult;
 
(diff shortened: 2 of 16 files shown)


arangod/Cluster/CreateCollection.cpp
```diff
@@ -273,9 +273,12 @@ bool CreateCollection::createReplication2Shard(CollectionID const& collection,
...
     if (leaderState != nullptr) {
-      leaderState->createShard(
-          shard, collection,
-          velocypack::SharedSlice(_description.properties()->bufferRef()));
+      // TODO we are blocking the maintenance here, but we should not
+      leaderState
+          ->createShard(
+              shard, collection,
+              velocypack::SharedSlice(_description.properties()->bufferRef()))
+          .get();
```
maierlars, line 281: We can simply put shard creation on the slow lane. This slow lane is designed for executing long running processes like synchronize shard (which takes days). We can wait for a shard log entry to be replicated.

arangod/Replication2/StateMachines/Document/DocumentCore.cpp
```diff
@@ -47,30 +47,32 @@ DocumentCore::DocumentCore(
...
+  _shards.emplace(shardId, std::move(shardProperties));
   LOG_CTX("b7e0d", TRACE, this->loggerContext)
-      << "Created shard " << _shardId << " for replicated state " << _gid;
+      << "Created shard " << shardId << " for replicated state " << _gid;
 }
 
-auto DocumentCore::getShardId() -> ShardID const& { return _shardId; }
+auto DocumentCore::getShardId() -> ShardID const& {
+  return _shards.begin()->first;
+}
```
maierlars, line 68: delete this function?

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -132,6 +135,15 @@ auto DocumentFollowerState::applyEntries(
...
                   << collectionId;
             } else {
+              // Even though a shard was dropped before acquiring the snapshot,
+              // we could still see transactions referring to that shard.
+              if (!data.core->isShardAvailable(doc.shardId)) {
+                LOG_CTX("1970d", INFO, self->loggerContext)
+                    << "Will not apply transaction " << doc.tid << " for shard "
+                    << doc.shardId << " because it is not available";
+                continue;
+              }
```
maierlars, line 145: Good catch!

```
set key=x shard=12 <- log from here
delete shard=12
set key=y shard=13 <- snapshot from here
```

arangod/Replication2/StateMachines/Document/DocumentCore.cpp
mpoeter: We will, but for now we still need it. This depends on the snapshot transfer PR to be merged first.

apetenchea: Not yet, unfortunately. We'll wait for #18109 to be merged.

maierlars (approved):

arangod/Cluster/CreateCollection.cpp
maierlars: you do by calling `get` on a future?

PR #18127, opened 2023-02-16
[GORDO-1518] Fix memory leak in tests

markuspf:
### Scope & Purpose

ASAN/UBSAN detected a load of memory leaks, all caused by not shutting down the runtimes for actors. 

This patch fixes the immediate problem; We will look at better solutions for shutting down runtimes gracefully.

To verify this locally I used the following commands

```
# cmake --preset community-developer-asan-ubsan
# cmake --build --preset community-developer-asan-ubsan --target arangodbtests_actor
# build-presets/bin/arangodbtests_actor
```


hkernbach (approved):
LGTM

PR #18130, opened 2023-02-16
Adjusted default threadCount for dump/restore.

joerg84:
### Scope & Purpose

This changes the default number of threads in `arangodump` and `arangorestore` to 4. Previously, this was the number of cores
of the client machine, which does not really make much sense. The value 4 is designed to not overwhelm small instances on ArangoGraph and can easily be adjusted by a command line option.

ATTENTION

Since https://arangodb.atlassian.net/browse/BTS-1232 was fixed, this PR is questionable. First, the problem this PR addresses does no longer exist, secondly, this PR could create a new problem, namely that many imports are much slower, since they do not use the full capabilities for parallelism of the client machine, the network and the database itself.
Therefore we should probably give up on it.

- [x] :hankey: Bugfix

### Checklist

- [ ] Backports
  - [ ] Backport for 3.10:


client-tools/Dump/DumpFeature.cpp
```diff
@@ -640,9 +640,8 @@ DumpFeature::DumpFeature(Server& server, int& exitCode)
   using arangodb::basics::FileUtils::buildFilename;
   using arangodb::basics::FileUtils::currentDirectory;
   _options.outputPath = buildFilename(currentDirectory().result(), "dump");
-  _options.threadCount =
-      std::max(uint32_t(_options.threadCount),
-               static_cast<uint32_t>(NumberOfCores::getValue()));
+  // Default ThreadCount 4
+  _options.threadCount = std::max(uint32_t(_options.threadCount), 4);
```
neunhoef, line 644: A simpler solution would be to remove this code here and simply adjust the default for `Options._threadCount` (from currently 2) to 4 in `client-tools/Dump/DumpFeature.h`.

client-tools/Restore/RestoreFeature.cpp
```diff
@@ -1670,9 +1670,8 @@ RestoreFeature::RestoreFeature(Server& server, int& exitCode)
   using arangodb::basics::FileUtils::buildFilename;
   using arangodb::basics::FileUtils::currentDirectory;
   _options.inputPath = buildFilename(currentDirectory().result(), "dump");
-  _options.threadCount =
-      std::max(uint32_t(_options.threadCount),
-               static_cast<uint32_t>(NumberOfCores::getValue()));
+  // Default ThreadCount 4
+  _options.threadCount = std::max(uint32_t(_options.threadCount), 4);
```
neunhoef, line 1674: Same here with `client-tools/Restore/RestoreFeature.h` and `struct Options` there.

neunhoef (changes requested):
This would work, but make it unnecessarily complicated.

neunhoef (approved):
LGTM

jsteemann:
Seems ok, but will make dump/restore slower in some situations. IMHO there is no "good" default value for the number of threads that does not depend on the deployment's capabilities, so we can as well adjust the default to 4 via this PR.

jsteemann (approved):
LGTM

neunhoef:
Since https://arangodb.atlassian.net/browse/BTS-1232 was fixed, this PR is questionable. First, the problem this PR addresses does no longer exist, secondly, this PR could create a new problem, namely that many imports are much slower, since they do not use the full capabilities for parallelism of the client machine, the network and the database itself.
Therefore we should probably give up on it.

PR #18133, opened 2023-02-17
[testing] compress agency dumps (#18083)

dothebart:
backport of https://github.com/arangodb/arangodb/pull/18083

### DIFF
--- a/js/client/modules/@arangodb/testutils/instance-manager.js
+++ b/js/client/modules/@arangodb/testutils/instance-manager.js
@@ -67,8 +67,11 @@ const termSignal = 15;
 
 const instanceRole = inst.instanceRole;
 
+let instanceCount = 1;
+
 class instanceManager {
   constructor(protocol, options, addArgs, testname, tmpDir) {
+    this.instanceCount = instanceCount++;
     this.protocol = protocol;
     this.options = options;
     this.addArgs = addArgs;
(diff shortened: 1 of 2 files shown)


js/client/modules/@arangodb/testutils/instance-manager.js
```diff
@@ -524,20 +527,48 @@ class instanceManager {
...
+    if (fs.isFile(zipfn)) {
+      fs.remove(zipfn);
+    };
+    if (fs.exists(dumpdir)) {
+      fs.list(dumpdir).forEach(file => {
+        const fn = fs.join(dumpdir, file);
+        if (fs.isFile(fn)) {
+          fs.remove(fn);
+        }
+      });
```
alexbakharew, line 541: Why we can't just remove all dumpdir without iterating over it? 
Why we delete here only files but not directories?

May be we can use something like this?
```
fs.rmSync(dir, { recursive: true, force: true });
```

js/client/modules/@arangodb/testutils/instance-manager.js
```diff
@@ -524,20 +527,48 @@ class instanceManager {
...
         }
       }
     });
+    let zipfiles = [];
+    fs.list(dumpdir).forEach(file => {
+      const fn = fs.join(dumpdir, file);
+      if (fs.isFile(fn)) {
+        zipfiles.push(file);
+      }
+    });
```
alexbakharew, line 565: Same question here: Why we add to zip only files but not directories and content of them?

alexbakharew (changes requested):
Need to figure out some questions

js/client/modules/@arangodb/testutils/instance-manager.js
dothebart: there may be zips and other json reports in this directory,

js/client/modules/@arangodb/testutils/instance-manager.js
dothebart: `..` is a directory ;)

naushniki (approved):
LGTM.

js/client/modules/@arangodb/testutils/instance-manager.js
alexbakharew: So we will delete only files on the first leve of `dumpdir`, right? Directories will be not processed?

alexbakharew (approved):
LGTM

KVS85 (approved):

PR #18079, opened 2023-02-09
[testing] make sure tests cleanup their instances directories and temporary files

dothebart:
### Scope & Purpose

- dump tests would not cleanup their work data on success in all cases, fix this.
- export tests would not clean up their database directories
- server_permissions tests would not clean up temporary files created along with the tests
- `shell-dump-integration.js` & `shell-restore-integration.js` would not clean up

PR #18134, opened 2023-02-17
Bug fix 3.10/only remove subdir (#18079)

dothebart:
backport of https://github.com/arangodb/arangodb/pull/18079


alexbakharew (approved):
LGTM

tests/js/client/shell/shell-restore-integration.js
```diff
@@ -1292,126 +1154,105 @@ function restoreIntegrationSuite() {
...
+        assertInstanceOf(arangodb.ArangoCollection, col);
+        assertEqual(colJson.parameters.name, col.name());
+        assertEqual(colJson.parameters.type, col.type());
+        if (isCluster) {
+          assertEqual(colJson.parameters.distributeShardsLike, col.properties().distributeShardsLike);
         }
-        fs.removeDirectoryRecursive(path, true);
-      } finally {
-        db._useDatabase("_system");
       }
```
mpoeter, line 1324: There are a few places where you left the `_useDatabase` in a finally. Was this intentional? Should we keep the finally here, or can we remove the try-finally in the other cases as well?

mpoeter (approved):

tests/js/client/shell/shell-restore-integration.js
dothebart: I guess not being in the system database is potentially bad for following up tests, hence if the test changes the database, it should change it back?
but probably `teardown` could be doing that as well, but I don't think we should do this in the backport.

mpoeter: Agreed about not doing this in the backport. I just wanted to point out that the changes are not fully consistent.

PR #18137, opened 2023-02-17
Coverage Presets, Small fix for CMakeLists.txt to work with clang

markuspf:
### Scope & Purpose

* Add CMake presets to build with coverage
* Fix `CMakeLists.txt` to work when using clang.


dothebart:
build with gcc + coverage successfull.

dothebart (approved):

PR #18149, opened 2023-02-20
Sort queries by modified timestamp [FE-132]

palashkaria:
### Scope & Purpose

This adds a new `modified_at` timestamp for queries, and adds a way to sort using the same.
Also updates the sorting UI, has these options now:
- sort by created date
- name
- modified date
- switch sorting order

Screenshot 2023-02-20 at 1 00 55 AM

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-132


js/apps/system/_admin/aardvark/APP/frontend/js/views/queryView.js
```diff
@@ -1566,11 +1597,57 @@
...
+        } else {
+          x = 0;
+        }
+        return x;
+      }
+      function sortByModifiedAt (a, b) {
+        var x;
+        if (a.modified_at > b.modified_at) {
+          x = -1;
+        } else if (a.modified_at < b.modified_at) {
```
hkernbach, line 1619: Do we need to check here if `modified_at` is available? Will it cause issues if it is not available here? e.g. produce false positives in case of `undefined`?

hkernbach (approved):
Apart from one comment LGTM. 

One very minor CSS-only addition: 
image

Borders of the search box are not aligned with the boxes in the background. If easy fixable, we could do this. If a fix would require too much time fiddling around with CSS, just ignore.

js/apps/system/_admin/aardvark/APP/frontend/js/views/queryView.js
palashkaria: True, good catch! I will add an explicit check, as undefined behaviour can be unpredictable

palashkaria:
I realized while testing that there is an issue with this task, and I will need input from @ansoboleva ; I have added a comment on the Jira ticket outlining the problem. We need to take a call on default behaviour

Comment here: https://arangodb.atlassian.net/browse/FE-132?focusedCommentId=86517

Pausing this for now and picking up the next ticket meanwhile!

ajurdak (approved):
LGTM.

PR #18153, opened 2023-02-20
[CINFRA] Fixing Windows build

apetenchea:
### Scope & Purpose

This PR fixes some MSVC compile errors and failing tests.

- Fixes compiler errors on Windows.
- Fixes a C++ unit test that became unstable due to differences in map initialization between platforms
- Fixes a restart test for replication2. Basically, after a restart we try to fetch all the documents of a collection, which we expect to get on the first try. That works on most platforms, but sometimes on Windows we get a broken connection error. I traced it to point of making the network request, and I observed that if we retry after 1 second or so, with exactly the same port and address, the request is successful. For now, I decided to stop the investigation there and make the test retry a few times, concluding that Windows is just slower at restarting processes.
- `getResponsibleServer` and `getResponsbileServers` for replication2 have been adapted for collection groups. This was a hidden bug which we didn't notice due to the newly added supervision for collection groups. I created a task to simplify the code in the future: [CINFRA-700](https://arangodb.atlassian.net/browse/CINFRA-700)

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

[CINFRA-700]: https://arangodb.atlassian.net/browse/CINFRA-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mchacki (changes requested):
I think this is an unintended behavior change, can we get awa without changing the behavior?

arangod/Cluster/Utils/TargetCollectionAgencyWriter.cpp
```diff
@@ -165,8 +165,7 @@ TargetCollectionAgencyWriter::prepareCreateTransaction(
     // Insert an empty object, we basically want to occupy the key here for
     // preconditions
     writes = std::move(writes).set(
-        collectionNamePath->collection(entry.getName())->str(),
-        [](VPackBuilder& builder) {});
+        collectionNamePath->collection(entry.getName())->str(), "");
```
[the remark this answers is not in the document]
apetenchea: Seems like `emplace_object` won't work either. I'll leave the PR open for now until we can come up with something else.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6215,19 +6215,61 @@ ClusterInfo::getResponsibleServerReplication2(std::string_view shardID) {
...
 
-  auto logId = LogicalCollection::shardIdToStateId(shardID);
   auto result = std::shared_ptr<std::vector<ServerID>>{nullptr};
 
   while (true) {
     {
+      if (++tries >= 100 || _server.isStopping()) {
+        break;
+      }
+
```
mchacki, line 6225: Should we introduce some kind of delay / backoff here?
The continue loop here is quite short.
We free the lock,
We add a number,
We ask for the lock.
=> I think there could be some chance that the writeLock never gets in between to update protected data.

Should we add something like (sleep 0.1 * planUpdateInteraval) for every consecutive try to increase the chances for write lock?

mchacki: In General I would prefer if we could get away without busy looping here in the first place.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6215,19 +6215,61 @@ ClusterInfo::getResponsibleServerReplication2(std::string_view shardID) {
...
+      // TODO we have to simplify this by using some adequate data structures
+      // lookup the collection name
+      auto colName = _shardToName.find(shardID);
+      if (colName == _shardToName.end()) {
+        continue;
+      }
+
+      // fetch the logical collection containing the shard
+      std::shared_ptr<LogicalCollection> logical{nullptr};
+      for (auto const& [dbID, col] : _plannedCollections) {
```
mchacki, line 6237: Wait a second, this sounds too complicated...

Can we talk about this in person?
We should be able to identify the group & shardSheaf index for a a shard much easier.

arangod/Cluster/ClusterInfo.cpp
mchacki: Why is the retry actually required?
Can't we proof that you only get the shardID in your hands, after all _planXXX entries have seen it?

arangod/Cluster/Utils/TargetCollectionAgencyWriter.cpp
```diff
@@ -205,9 +205,12 @@ TargetCollectionAgencyWriter::prepareCreateTransaction(
 
     // Insert an empty object, we basically want to occupy the key here for
     // preconditions
-    writes = std::move(writes).set(
+    writes = std::move(writes).emplace_object(
         collectionNamePath->collection(entry.getName())->str(),
-        [](VPackBuilder& builder) {});
+        [](VPackBuilder& builder) {
+          builder.openObject();
```
mchacki, line 211: Great thanks :+1:

mchacki (commented):
I am not yet decided if I want to approve / reject this PR.

The code itself looks good to me.
However I think the logic is more complex than necessary, would be great if we can talk about it.

Also happy to do this in 2 Steps, e.g. merge this one, and do a simplification followup.

arangod/Cluster/ClusterInfo.cpp
apetenchea: There is a sleep at line 6301, for 500ms. Until now, I never experienced problems with this sleep interval. Nevertheless, if you think it's worth it, we could increase it with each iteration.
The retry was there originally and it is required to make sure that leadership has been established, similar to `getResponsibleServerReplication1`, which also has a retry.
Perhaps some of the `continue` statements could be eliminated, but other than that I don't think there's a way to ensure that the log has a leader, apart from waiting.

arangod/Cluster/ClusterInfo.cpp
apetenchea: I did not find an easier way to do that and created a ticket for simplifying this in the future: [CINFRA-700](https://arangodb.atlassian.net/browse/CINFRA-700). I think we're missing some of the necessary structures in `ClusterInfo`, but I'll gladly change it if there's an easier workaround.

mchacki (approved):
LGTM

PR #18160, opened 2023-02-20
Add missing sccache links

mpoeter:
### Scope & Purpose

Somehow the sccache links in the build container got lost - this PR fixes that.
Note - no build is needed for this PR. The container is already built and pushed and used in current CircleCI builds.

- [x] :hankey: Bugfix


markuspf (approved):
LGTM

PR #18161, opened 2023-02-21
CINFRA: ShardKeys are not part of Target yet, and hence are forgotten

mchacki:
This fixes an internal replication 2.0 bug. We forgot to forward ShardKeys of collections, this has some undesired side-effects.


maierlars (approved):

mchacki:
Red tests are unrelated.

PR #18163, opened 2023-02-21
[CINFRA-674][CINFRA-675]  Replicated operations refactoring

apetenchea:
### Summary

This pull request refactors DocumentLogEntries to use variants instead of enums and fixes several bugs that were discovered during the refactoring.

- Refactored DocumentLogEntries to use variants instead of enums
- Fixed several bugs that were discovered during the refactoring, including handling the IntermediateCommit operation
- Introduced a new ReplicatedOperation structure to indicate the command that the replicated state machine has to execute
- Fixed a potential problem caused by early compaction on the leader
- Separated the decision to replicate a transaction from replicateOperations and moved that logic into a separate method: needsReplication
- Added additional C++ tests as needed and refactored existing tests

### Gory details

1. We are refactoring `DocumentLogEntries` to use variants instead of enum. Due to this refactoring alone, several bugs were found and fixed, as pointed out by the compiler (for example, we forgot to handle the `IntermediateCommit` operation in several places). The `DocumentLogEntry` structure remains as the main mechanism for *transporting* log entries into the stream, while the newly added `ReplicatedOperation` *indicates the command* that the replicated state machine has to execute. Currently, the log entry is only a wrapper on top of the operation, but I believe this separation could prove to be beneficial in the future. This change is responsible for most lines modified in this PR, and brought with itself a necessary refactoring in the unit tests (both C++ and JS). Additional C++ tests were added as needed. This point address [CINFRA-674](https://arangodb.atlassian.net/browse/CINFRA-674).
2. A potential problem caused by early compaction was fixed on the leader. When dealing with `Commit` and `CreateShard` operations, the leader has to replicate them first, wait for the entries to be committed (in the log sense), commit the operation in RocksDB (or create the shard) and *only then* call release on the stream. This is now achieved by introducing a `release` method on the leader. There are currently two versions, one for handling user transactions that carry an official `TransactionID` and cannot simply be split in half by compaction, and one taking just a log index, which is enough for identifying an internal transaction (such as `DropShard`). This could be improved in the future, such that we no longer have to handle two different cases, perhaps we could generate transaction IDs for internal operations as well. The first plan was to use a `ScopeGuard` here, but I found that a bit tricky due to potential exceptions being thrown in `doCommit()`. Nevertheless, the focus here was to get a working version, while making sure the current refactoring allows for such improvements in the future. See [CINFRA-675](https://arangodb.atlassian.net/browse/CINFRA-675) for reference.
3. We separated the decision to replicate a transaction from `replicateOperations` and moved that logic into a separate method: `needsReplication`. This makes it more visible where are the places in which we decide what to replicate or not. However, as single commit/abort operations can still be encountered on the follower, we've made sure these are safely ignored. See the comments on [CINFRA-695](https://arangodb.atlassian.net/browse/CINFRA-694).
4. Some additional bug-fixes and minor TODO's were covered in the rest of this PR.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

[CINFRA-674]: https://arangodb.atlassian.net/browse/CINFRA-674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CINFRA-675]: https://arangodb.atlassian.net/browse/CINFRA-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -101,79 +108,101 @@ auto DocumentStateTransactionHandler::getTrx(TransactionId tid)
...
-                  .with<logContextKeyLogId>(_gid.id);
-
-          LOG_CTX("0da00", INFO, logContext)
-              << "Result ignored while applying transaction " << doc.tid
-              << " with operation " << int(doc.operation) << " on shard "
-              << doc.shardId << ": " << res;
+auto DocumentStateTransactionHandler::applyEntry(ReplicatedOperation operation)
+    -> Result try {
+  return std::visit(
+      [this, &operation](auto&& op) -> Result {
```
mchacki, line 120: Same as before, we probably want to consider using override visitor here.
With this many cases it gets a bit hard to visually see that all `else if constexpr` are correct, and we not only have a single if at one place.

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -101,79 +108,101 @@ auto DocumentStateTransactionHandler::getTrx(TransactionId tid)
...
+            return shardRes.result();
+          }
+          if (!shardRes.get()) {
+            LOG_CTX("c98e9", INFO, _logContext)
+                << "Shard " << op.shard << " already exists";
+          }
+          return Result{};
+        } else if constexpr (std::is_same_v<T,
+                                            ReplicatedOperation::DropShard>) {
+          auto transactions = getTransactionsForShard(op.shard);
```
mchacki, line 187: Please wrap this 2 lines into ifdef MAINTAINER_MODE

The transactions are only used within the Assert, so we can help the compiler that this is actually only required in Maintainer mode, and not required in release. (We probably end up with "unused" compiler warning here anyhow.

arangod/Replication2/StateMachines/Document/ReplicatedOperation.cpp
```diff
@@ -0,0 +1,112 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2023 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
```
mchacki, line 5:
```suggestion
/// Copyright 2023-2023 ArangoDB GmbH, Cologne, Germany
```

arangod/Replication2/StateMachines/Document/ReplicatedOperation.h
```diff
@@ -0,0 +1,267 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2023 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
```
mchacki, line 5:
```suggestion
/// Copyright 2023-2023 ArangoDB GmbH, Cologne, Germany
```

arangod/Replication2/StateMachines/Document/ReplicatedOperation.h
```diff
@@ -0,0 +1,267 @@
...
+
+#include <variant>
+
+namespace arangodb::replication2::replicated_state::document {
+
+/*
+ * When a log entry is received, the ReplicatedOperation tells the state machine
+ * what it has to do (i.e. start a transaction, drop a shard, ...).
+ */
+struct ReplicatedOperation {
```
mchacki, line 40: This comment holds for all of the following file.

Do we need these elements exactly where they are serialized/deserialzed as well?
Or are they passed around, and only serialized/deserialized in fewer files?

If the second is true, we should have a follow up task to split "Inspectors" from "Structs" into two separate header files.
The inspector has high impact on compile time, that we can avoid for files just using the structs, but have nothing to do with their outside representation

arangod/RocksDBEngine/ReplicatedRocksDBTransactionCollection.cpp
```diff
@@ -227,24 +221,27 @@ auto ReplicatedRocksDBTransactionCollection::ensureCollection() -> Result {
 futures::Future<Result>
 ReplicatedRocksDBTransactionCollection::performIntermediateCommitIfRequired() {
   if (_rocksMethods->isIntermediateCommitNeeded()) {
+    // TODO due to distributeShardsLike, it is possible that we'll replicate
```
mchacki, line 224: Do we have a ticket for this?

mchacki (changes requested):
In general a very good PR.

My request change is centered around the TRI_ASSERT with side-effect, I think it is important to get this sorted out before merge. (Otherwise TestCode and ReleaseCode have different behavior)

All other comments can be considered nice to have, and I would not block the PR for any of them.

arangod/Replication2/StateMachines/Document/ActiveTransactionsQueue.cpp
```diff
@@ -30,70 +30,62 @@
...
+std::optional<LogIndex> ActiveTransactionsQueue::getReleaseIndex() const {
+  if (_logIndices.empty()) {
+    return std::nullopt;
   }
-}
-
-std::size_t ActiveTransactionsQueue::size() const noexcept {
-  return _transactions.size();
+  auto const& idx = _logIndices.front();
+  TRI_ASSERT(idx.second == Status::kActive)
```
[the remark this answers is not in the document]
apetenchea: Sure, doesn't hurt to have an ADB_PROD_ASSERT

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -89,93 +87,129 @@ auto DocumentFollowerState::acquireSnapshot(ParticipantId const& destination,
...
-                    self->_activeTransactions.getReleaseIndex(entry->first);
+    return basics::catchToResultT([&]() -> std::optional<LogIndex> {
+      std::optional<LogIndex> releaseIndex;
+      auto const& transactionHandler = data.core->getTransactionHandler();
+
+      while (auto entry = ptr->next()) {
+        auto doc = entry->second;
+
+        auto res = std::visit(
+            [&](auto& op) -> Result {
```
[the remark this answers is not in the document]
apetenchea: I think the overloaded version looks nicer indeed, but my goal was to make sure that I handle all cases, which now I can using the `static_assert(always_false_v<T>, "non-exhaustive visitor!");` at line 189. Unfortunately I don't think that such an assertion is possible with the overloaded approach.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.h
```diff
@@ -83,7 +81,6 @@ struct DocumentFollowerState
   };
 
   std::shared_ptr<IDocumentStateNetworkHandler> _networkHandler;
-  std::unique_ptr<IDocumentStateTransactionHandler> _transactionHandler;
   Guarded<GuardedData, basics::UnshackledMutex> _guardedData;
   ActiveTransactionsQueue _activeTransactions;
```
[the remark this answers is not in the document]
apetenchea: You are right, it's only being used inside the `_guardedData` lock anyway.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -156,76 +176,125 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
-      !_activeTransactions.getLockedGuard()->erase(transactionId)) {
-    // we have not replicated anything for a transaction with this id, so
-    // there is no need to replicate the abort/commit operation
-    return LogIndex{};  // TODO - can we do this differently instead of
-                        // returning a dummy index
-  }
+auto DocumentLeaderState::needsReplication(ReplicatedOperation const& op)
+    -> bool {
+  return std::visit(
+      [&](auto&& op) -> bool {
```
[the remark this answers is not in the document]
apetenchea: I used `auto&&` because afaik in that case the compiler can choose whatever fits best.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -269,77 +338,105 @@ auto DocumentLeaderState::allSnapshotsStatus() -> ResultT<AllSnapshotsStatus> {
...
+      return result.result();
+    }
+    auto logIndex = result.get();
+
+    return self->_guardedData.doUnderLock([&](auto& data) -> Result {
+      if (data.didResign()) {
+        return TRI_ERROR_REPLICATION_REPLICATED_LOG_LEADER_RESIGNED;
+      }
+
+      // TODO we clear snapshot here, to release the shard lock. This is
```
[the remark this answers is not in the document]
apetenchea: Proper deletion of snapshots will be handled with [CINFRA-676](https://arangodb.atlassian.net/browse/CINFRA-676)

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.h
```diff
@@ -39,26 +38,42 @@ namespace arangodb::replication2::replicated_state::document {
...
+                                     std::shared_ptr<VPackBuilder> properties)
+      -> Result;
+  auto executeDropCollectionAction(ShardID shard, CollectionID collection)
+      -> Result;
 
  private:
   GlobalLogIdentifier _gid;
   MaintenanceFeature& _maintenanceFeature;
+  ServerID _server;
+  struct {
```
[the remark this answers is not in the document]
apetenchea: I wanted to use a `shared_lock` for reading data, which I didn't manage to implement with a `GuardedData`. I don't think it can differentiate between *shared* and *unique* locking.

arangod/Replication2/StateMachines/Document/DocumentStateTransaction.cpp
```diff
@@ -45,26 +45,30 @@ auto DocumentStateTransaction::apply(DocumentLogEntry const& entry)
...
+                              ReplicatedOperation::Update,
+                              ReplicatedOperation::Replace>) {
+          opOptions.overwriteMode = OperationOptions::OverwriteMode::Replace;
+          return _methods->insert(operation.shard, operation.payload.slice(),
+                                  opOptions);
+        } else if constexpr (std::is_same_v<T, ReplicatedOperation::Remove>) {
+          return _methods->remove(operation.shard, operation.payload.slice(),
+                                  opOptions);
+        } else if constexpr (std::is_same_v<T, ReplicatedOperation::Truncate>) {
+          // TODO Think about correctness and efficiency.
```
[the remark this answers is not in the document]
apetenchea: I created one: [CINFRA-697](https://arangodb.atlassian.net/browse/CINFRA-697)

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
apetenchea: My problem is that I'm trying to have these `static_assert(always_false_v<T>, "Unhandled operation type. This should never happen.");` help me make sure that I handle all cases at compile time. In the end, the overload solution would work fine for me, I don't have a strong preference on how to achieve this visitor thing. We could bring that as a topic of debate to one of our team meetings and decide upon a common approach.

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
apetenchea: Good catch!

arangod/RocksDBEngine/ReplicatedRocksDBTransactionCollection.cpp
apetenchea: Created one: [CINFRA-698](https://arangodb.atlassian.net/browse/CINFRA-698)

arangod/Replication2/StateMachines/Document/ReplicatedOperation.h
apetenchea: That's a good observation, I have split them into separate files!

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -156,76 +176,125 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
-        operation != OperationType::kAbort) {
-      activeTransactions.emplace(transactionId, idx);
-    } else {
-      // TODO - must not release commit entry before trx has actually been
-      // committed!
-      stream->release(activeTransactions.getReleaseIndex(idx));
-    }
-    return idx;
+  auto&& insertionRes = basics::catchToResultT([&] {
+    return _activeTransactions.doUnderLock([&](auto& activeTransactions) {
```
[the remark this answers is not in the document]
apetenchea: I expanded the current comment

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.h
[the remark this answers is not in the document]
apetenchea: I created a ticket for that: [CINFRA-701](https://arangodb.atlassian.net/browse/CINFRA-701)

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -98,54 +97,77 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
-        }
 
-        auto res = transactionHandler->applyEntry(doc);
-        if (res.fail()) {
-          return res;
-        }
+      auto res = std::visit(
+          [&](auto&& op) -> Result {
+            using T = std::decay_t<decltype(op)>;
+            if constexpr (ModifiesUserTransaction<T>) {
```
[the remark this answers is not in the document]
apetenchea: The problem is a bit more complicated for `recoverEntries`, because there are several places where we return early from the `std::visit` call. There is also some common code at the end, the `transactionHandler->applyEntry` call, which might happen or not, depending on the logic contained in the if statements above.

apetenchea:
I created [CINFRA-702](https://arangodb.atlassian.net/browse/CINFRA-702) for ideas of future improvements.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -100,82 +98,39 @@ auto DocumentFollowerState::applyEntries(
...
-                self->_activeTransactions.erase(doc.tid);
-                releaseIndex =
-                    self->_activeTransactions.getReleaseIndex(entry->first);
-              } else {
-                self->_activeTransactions.emplace(doc.tid, entry->first);
-              }
+            auto [index, doc] = *entry;
+
+            auto currentReleaseIndex = std::visit(
+                [&data, index = entry->first](
```
mpoeter, line 104:
```suggestion
            auto [index, doc] = *entry;

            auto currentReleaseIndex = std::visit(
                [&data, index](
```

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -100,82 +98,39 @@ auto DocumentFollowerState::applyEntries(
...
+                doc.getInnerOperation());
+
+            if (currentReleaseIndex.fail()) {
+              LOG_CTX("0aa2e", FATAL, self->loggerContext)
+                  << "failed to apply entry " << doc
+                  << " on follower: " << currentReleaseIndex.result();
+              FATAL_ERROR_EXIT();
+            }
+            if (currentReleaseIndex->has_value()) {
+              releaseIndex = std::move(currentReleaseIndex).get();
```
mpoeter, line 117: Looks like `releaseIndex` is only set here, so we don't really need it and can return `currentReleaseIndex` instead.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -363,4 +309,124 @@ auto DocumentFollowerState::handleSnapshotTransfer(
...
+        << " because it is not active";
+    return ResultT<std::optional<LogIndex>>{std::nullopt};
+  }
+  auto const& transactionHandler = core->getTransactionHandler();
+  if (auto res = transactionHandler->applyEntry(op); res.fail()) {
+    return res;
+  }
+
+  // We don't need to update the release index after an intermediate commit.
+  return ResultT<std::optional<LogIndex>>::success(std::nullopt);
```
mpoeter, line 353: Theoretically we can release everything in this transaction up to this point and update the start LogIndex of this transaction to the current LogIndex. However, I would not do this now but maybe put it down so we can revisit this later, perhaps in the refactoring where we try to unify code between the follower state and the leader recovery.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -98,54 +94,79 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
-    stream->insert(doc);
+    auto abortAll = ReplicatedOperation::buildAbortAllOngoingTrxOperation();
+    auto abortAllRes =
+        self->replicateOperation(abortAll, ReplicationOptions{}).get();
+    if (abortAllRes.fail()) {
+      LOG_CTX("b4217", FATAL, self->loggerContext)
+          << "failed to replicate AbortAllOngoingTrx operation during "
+             "recovery: "
+          << abortAllRes.result();
+      FATAL_ERROR_EXIT();
```
mpoeter, line 168: Do we really need to block here until the entry is replicated?

mchacki (approved):
LGTM

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -26,14 +26,14 @@
...
-#include "Transaction/ReplicatedContext.h"
-#include "DocumentStateHandlersFactory.h"
+#include "VocBase/AccessMode.h"
 
-namespace {
-
-auto shouldIgnoreError(arangodb::OperationResult const& res) noexcept -> bool {
+namespace arangodb::replication2::replicated_state::document {
+auto DocumentStateTransactionHandler::shouldIgnoreError(
+    arangodb::OperationResult const& res) noexcept -> bool {
```
mpoeter, line 36: NIT: any particular reason why you moved this function into the DocumentStateTransactionHandler? I usually keep function that are not really related to a class (low cohesion, ie., do not rely on its members/functions) outside as separate functions, especially when they are not used outside of this module. This also helps keep the class' interface clean since a member function also has to be declared in the header.

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -59,8 +59,8 @@ auto shouldIgnoreError(arangodb::OperationResult const& res) noexcept -> bool {
   return true;
 }
 
-auto makeResultFromOperationResult(arangodb::OperationResult const& res,
-                                   arangodb::TransactionId tid)
+auto DocumentStateTransactionHandler::makeResultFromOperationResult(
+    arangodb::OperationResult const& res, arangodb::TransactionId tid)
```
mpoeter, line 63: NIT: same as above

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -101,103 +104,145 @@ auto DocumentStateTransactionHandler::getTrx(TransactionId tid)
...
-auto DocumentStateTransactionHandler::applyEntry(DocumentLogEntry doc)
-    -> Result {
-  if (doc.operation == OperationType::kAbortAllOngoingTrx) {
-    _transactions.clear();
-    return Result{};
-  }
+void DocumentStateTransactionHandler::setTrx(
+    TransactionId tid, std::shared_ptr<IDocumentStateTransaction> trx) {
+  auto [_, isInserted] = _transactions.emplace(tid, std::move(trx));
+  TRI_ASSERT(isInserted) << "Transaction " << tid << " already exists";
```
mpoeter, line 110: Should we make this an `ADB_PROD_ASSERT`?

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
```diff
@@ -101,103 +104,145 @@ auto DocumentStateTransactionHandler::getTrx(TransactionId tid)
...
+  auto trx = getTrx(op.tid);
+  ADB_PROD_ASSERT(trx != nullptr);
+
+  auto res = Result{};
+  using T = std::decay_t<decltype(op)>;
+  if constexpr (std::is_same_v<T, ReplicatedOperation::Commit>) {
+    res = trx->commit();
+  } else if constexpr (std::is_same_v<T, ReplicatedOperation::Abort>) {
+    res = trx->abort();
   }
```
mpoeter, line 160: I would prefer something like this:
```suggestion
  auto res = std::invoke(
    overload{
      [&](ReplicatedOperation::Commit const&) { return trx->commit(); },
      [&](ReplicatedOperation::Abort const&) { return trx->abort); }
    }, op);
```
I think this is a bit easier to read and it would also cause a compile error if we ever add a new type to `FinishesUserTransaction`.

arangod/Transaction/Methods.cpp
```diff
@@ -2644,18 +2644,17 @@ Future<OperationResult> transaction::Methods::truncateLocal(
...
+        replication2::replicated_state::document::ReplicatedOperation::
+            buildTruncateOperation(state()->id(), trxColl->collectionName());
+    // Should finish immediately
+    auto replicationRes =
+        leaderState
+            ->replicateOperation(
+                std::move(operation),
+                replication2::replicated_state::document::ReplicationOptions{})
+            .get();
+    return OperationResult{replicationRes.result(), options};
```
mpoeter, line 2657: Why should this finish immediately? Shouldn't this be async?

arangod/Transaction/Methods.cpp
```diff
@@ -3139,17 +3138,19 @@ Future<Result> Methods::replicateOperations(
...
+            replicationData.sharedSlice());
+    // Should finish immediately
+    auto replicationRes =
+        leaderState
+            ->replicateOperation(
+                std::move(replicatedOp),
+                replication2::replicated_state::document::ReplicationOptions{})
+            .get();
+    if (replicationRes.fail()) {
+      return replicationRes.result();
```
mpoeter, line 3153: Why should this finish immediately? Shouldn't this be async?

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -100,82 +98,39 @@ auto DocumentFollowerState::applyEntries(
...
-                    self->_activeTransactions.getReleaseIndex(entry->first);
-              } else if (doc.operation == OperationType::kCommit ||
-                         doc.operation == OperationType::kAbort) {
-                self->_activeTransactions.erase(doc.tid);
-                releaseIndex =
-                    self->_activeTransactions.getReleaseIndex(entry->first);
-              } else {
-                self->_activeTransactions.emplace(doc.tid, entry->first);
-              }
+            auto [index, doc] = *entry;
```
maierlars, line 101: copy of document log entry might be expensive if we write huge batches.

mpoeter (approved):

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
maierlars: I don't think so.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -156,76 +177,125 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
...
-                        // returning a dummy index
-  }
+auto DocumentLeaderState::needsReplication(ReplicatedOperation const& op)
+    -> bool {
+  return std::visit(
+      overload{
+          [&](FinishesUserTransactionOrIntermediate auto const& op) -> bool {
+            // We don't replicate single abort/commit operations
+            // in case we have not replicated anything else for
+            // a transaction.
```
maierlars, line 201: This comment was confusing. Is it about not replicating the commit or abort for an empty transaction?

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
maierlars: forget about that, the log entry contains a `ShardSlice` which is cheap to copy.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
apetenchea: I use `releaseIndex` in order to preserve only the last `currentReleaseIndex` that's calculated in the while loop, I can't just return `currentReleaseIndex` directly.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
apetenchea: Good, noted that

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: We are not actually blocking, because we don't wait for commit. I will add an assertion to make sure that the future is always ready after the `replicateOperation` call.

maierlars: ah because `waitForCommit` is not set in the options 😉

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: Updated the comment

arangod/Transaction/Methods.cpp
apetenchea: It finishes immediately because we are not waiting for the operation to be committed in the log. I added an assertion and comment on that.

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
apetenchea: I updated the method.

arangod/Transaction/Methods.cpp
apetenchea: Same as above

arangod/Replication2/StateMachines/Document/DocumentStateTransactionHandler.cpp
apetenchea: While working with concepts and visitors, I had some linker issues which I wanted to get out of the way quickly, so I moved these functions around. That's no longer the case now, hence I moved them back to an anonymous namespace in the cpp file. I agree with your points.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
mpoeter: Ah, good point. 🤦‍♂️

PR #18172, opened 2023-02-22
Slick GEO tests to just create each set of data *once* at the beginning in individual collections / views

dothebart:
### Scope & Purpose

We improve the test performance by 
1) [x] DRY the test case creation
2) [x] use setupAll
3) [x] join the suites to just have one set of these

- [x] :hankey: Bugfix

Same set of binaries, 
`time ./scripts/unittest auto --test aql-geo-3  --cluster true` 
=>  Default:
```
real    2m3.704s
user    3m24.658s
sys     0m14.395s
```
=> this PR:
```
real    0m56.104s
user    0m58.925s
sys     0m5.747s
```

### DIFF
--- a/tests/js/server/aql/aql-geo.inc
+++ b/tests/js/server/aql/aql-geo.inc
@@ -486,7 +486,8 @@
         Object.keys(withView).forEach(key => { withView[key].drop(); });
         Object.keys(noIndex).forEach(key => { noIndex[key].drop(); });
         Object.keys(withIndex).forEach(key => { withIndex[key].drop(); });
-        Object.keys(testAnalyzers).forEach(key => { analyzers.remove(testAnalyzers[key].name()); });
+        Object.keys(testAnalyzers).forEach(key => {
+          analyzers.remove(testAnalyzers[key].name(), true); });
       },
 
       ////////////////////////////////////////////////////////////////////////////////
--- a/tests/js/server/aql/aql-geo.inc
+++ b/tests/js/server/aql/aql-geo.inc
@@ -483,17 +483,10 @@
 
       tearDownAll: function () {
         Object.keys(plainView).forEach(key => { plainView[key].drop(); });
+        Object.keys(withView).forEach(key => { withView[key].drop(); });
         Object.keys(noIndex).forEach(key => { noIndex[key].drop(); });
         Object.keys(withIndex).forEach(key => { withIndex[key].drop(); });
-        Object.keys(withView).forEach(key => { withView[key].drop(); });
-        //Object.keys(testAnalyzers).forEach(key => { testAnalyzers[key].drop(); });
-      },
-
-      ////////////////////////////////////////////////////////////////////////////////
-      /// @brief tear down
-      ////////////////////////////////////////////////////////////////////////////////
-
-      tearDown: function () {
+        Object.keys(testAnalyzers).forEach(key => { analyzers.remove(testAnalyzers[key].name()); });
       },
 
       ////////////////////////////////////////////////////////////////////////////////
--- a/tests/js/server/aql/aql-geo.inc
+++ b/tests/js/server/aql/aql-geo.inc
@@ -47,7 +47,7 @@
       return {
         type: "Polygon",
         coordinates: [[[lon[0], lat[0]], [lon[1], lat[0]],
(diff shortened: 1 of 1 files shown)


MBkkt (changes requested):
In this test single analyzer per test run.
So you don't need to have multiple analyzers, and also parameter for queries

MBkkt (commented):
Parameterization of tests is probably ok, but creating multiple analyzers with different names but same definition looks bad to me (I know it was, but if you change test, I think you can also fix this)

tests/js/server/aql/aql-geo.inc
```diff
@@ -216,916 +105,885 @@
...
-      truncateAll();
-    },
+      let cmpResFullScanWithIndexes = compareKeyLists("without index", resFullScanNoIndexes, "with index", resFullScanWithIndexes);
+      let cmpResView = compareKeyLists("without index", resFullScanNoIndexes, "with view", resView);
+      let cmpResInvertedIndex = compareKeyLists("without index", resFullScanNoIndexes, "with inverted index", resInvertedIndex);
 
-////////////////////////////////////////////////////////////////////////////////
-/// @brief test single point
-////////////////////////////////////////////////////////////////////////////////
+      // PLEASE UNCOMMENT LINES BELOW AFTER FIXING https://arangodb.atlassian.net/browse/BTS-1184
```
MBkkt, line 233: https://github.com/arangodb/arangodb/pull/18218

tests/js/server/aql/aql-geo.inc
```diff
@@ -56,158 +71,37 @@
...
+    function createCollectionSet(basename, theanalyzer) {
+      let collWithoutIndex = "UnitTestsGeoWithoutIndex";
+      let collWithIndex = "UnitTestsGeoWithIndex";
+      let collWithView = "UnitTestsGeoWithView";
+      let viewName = "UnitTestsGeoView";
+      noIndex[basename] = db._create(basename + '_' + collWithoutIndex);
+      withIndex[basename] = db._create(basename + '_' + collWithIndex);
+      withIndex[basename].ensureIndex({type: "geo", geoJson: true, fields: ["geo"]});
+      withView[basename] = db._create(basename + '_' + collWithView);
+      testAnalyzers[basename] = theanalyzer;
```
MBkkt, line 101: NIT
https://stackoverflow.com/questions/6600868/set-default-value-of-javascript-object-attributes

tests/js/server/aql/aql-geo.inc
```diff
@@ -216,916 +110,889 @@
...
-      db._drop(collWithoutIndex);
-      db._drop(collWithIndex);
-      db._drop(collWithView);
-      let analyzers = require("@arangodb/analyzers");
-      analyzers.remove("geo_json");
-    },
+    function compare(query, queryView) {
+      let fullScanQuery = `FOR d IN @@coll ${query} RETURN d._key`;
+      let viewQuery = `FOR d IN ${viewName} ${queryView} RETURN d._key`;
+      let invertedIndexQuery = `FOR d IN ${collWithIndex} OPTIONS {indexHint: "inverted", forceIndexHint: true, waitForSync: true} ${query} RETURN d._key`;
```
MBkkt, line 227: In general will be better if you move it the setupAll, same as views.
Also it can be improved because search-alias view use inverted indexes (but it question for @alexbakharew why didn't he does it). Not necessary to create separate view and inverted index collection. Also not necessary to create double count of inverted indexes.
```suggestion
      let invertedIndexQuery = `FOR d IN ${collWithIndex} OPTIONS {indexHint: "inverted", forceIndexHint: true } ${query} RETURN d._key`;
```

MBkkt (approved):

tests/js/server/aql/aql-geo.inc
MBkkt: Please uncomment this test, it anyway will be conflict between your PR and @cpjulia

tests/js/server/aql/aql-geo.inc
dothebart: this doesn't copy the `theanalyzer` which is perfectly fine for us here, since we don't intend to edit it afterwards, we just want to find it under that key.

PR #18173, opened 2023-02-22
Make inspection its own library

markuspf:
### Scope & Purpose

When developing new features for the inspection library it comes in handy to be able to compile the inspector tests by themselves.

This still has a few more dependency-tentacles into `lib` than I would like, but it is a start.


mpoeter (approved):

markuspf:
Jenkins test failures are unrelated.

PR #17682, opened 2022-11-29
Started foundations of telemetrics

cpjulia:
### Scope & Purpose

*(Please describe the changes in this PR for reviewers, motivation, rationale - **mandatory**)*

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #18174, opened 2023-02-23
[APM-294]

cpjulia:
### Scope & Purpose

Added telemetrics API that gathers info about a deployment for the receiving side to generate statistics. It's accessible via `_admin/server-info`. It's on by default and can be turned off with server startup parameter `--server.send-telemetrics` set to false. If set to true, every time the user logs into arangosh, one request is sent to gather telemetrics from `_admin/server-info`. Only users that are admin might get a successful http response with the telemetrics object.
Obs.: In this PR, there's no persistence or server feature that keeps sending telemetrics every interval of time. It's fetched on the fly and sent when requested. If needed, the feature from the old PR that has persistence can be added. Also, there was a startup parameter with the time interval in which telemetrics owuld be fetched, not all happens on the fly.
Related to https://github.com/arangodb/arangodb/pull/17682.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/APM-294
- [ ] Design document:


dothebart:
hm, I would rather not name this something with telemetry as long as it doesn't support OpenTelemetry.

CHANGELOG
```diff
@@ -24,6 +32,7 @@ devel
   and `_to` parts. Lookups from the in-memory edge cache will not return
   compressed values but the full-length edge ids. The compressed values will
   also be used in memory only and will not be persisted on disk.
+>>>>>>> aff4448a4f54c9e197f3616f683db9d0bfc31ca3
```
Simran-B, line 35: Merge conflict leftover

arangod/GeneralServer/GeneralServerFeature.h
```diff
@@ -87,6 +87,8 @@ class GeneralServerFeature final : public ArangodFeature {
 
   void countVstConnection() { _vstConnections.count(); }
 
+  bool isTelemetricsEnabled() { return _enableTelemetrics; }
```
jsteemann, line 90:
```suggestion
  bool isTelemetricsEnabled() const noexcept { return _enableTelemetrics; }
```

lib/V8/v8-utils.h
```diff
@@ -26,14 +26,22 @@
...
 #include <v8.h>
 
 #include "Basics/ErrorCode.h"
 
 namespace arangodb {
 class Result;
 }
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief downloads data from a URL
```
jsteemann, line 39: the description `downloads data from a URL` sounds strange when looking at the function name `getEndpoint`.
So either the function should be renamed or the description should be adjusted.
Also, as it is a public API now, the meaning of the parameters (and the 3 strings in the return value) should be somehow clarified.

CHANGELOG
```diff
@@ -1,6 +1,18 @@
...
+  the receiving side to generate statistics. It's accessible via 
+  `_admin/telemetrics`. It's on by default, only disabled in maintainer 
+  mode,  and can be turned on/off with server startup parameter 
+  `--server.send-telemetrics` set to false. 
+  If set to true, every time the user logs into arangosh, one request is
+  sent to gather telemetrics from `_admin/server-info`. Only users that are 
+  admin might get a successful http response with the telemetrics object.
+  To send telemetrics to our endpoint for testing in maintainer mode, 
+  the shell must be started with parameter 
+  `--telemetrics.send-to-endpoint` set to true.
```
jsteemann, line 14:
```suggestion
* APM-294: Added telemetrics API that gathers anonymous feature usage
  statistics from a deployment. The API is accessible via the endpoint
  `/_admin/telemetrics`. The API is enabled by default in release builds, but
  disabled by default in maintainer mode. It can be explicitly turned on/off 
  with the server startup parameter `--server.send-telemetrics`. 
  The required access privileges to access the telemetrics API can be 
  configured via the server startup option `--server.support-info-api`.
  The telemetrics API is used by the arangosh: every time the arangosh is 
  started, it will send a request to the connected server to gather telemetrics
  from the `/_admin/server-info` endpoint. The telemetrics data are then 
  sent to an aggregation service that is run by ArangoDB.
```

arangod/GeneralServer/GeneralServerFeature.cpp
jsteemann, line 1: The description for the startup option `--server.support-info-api` needs to be extended, because with this PR it does not only control the policy for exposing the support info API, but also the telemetrics API.

arangod/RestHandler/RestTelemetricsHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+RestTelemetricsHandler::RestTelemetricsHandler(ArangodServer& server,
+                                               GeneralRequest* request,
+                                               GeneralResponse* response)
+    : RestBaseHandler(server, request, response) {}
+
+RestStatus RestTelemetricsHandler::execute() {
+  GeneralServerFeature& gs = server().getFeature<GeneralServerFeature>();
+  if (!gs.isTelemetricsEnabled()) {
+    generateError(rest::ResponseCode::FORBIDDEN, TRI_ERROR_HTTP_FORBIDDEN,
+                  "telemetrics is disabled. Must enable with startup parameter "
```
jsteemann, line 48:
```suggestion
                  "telemetrics API is disabled. Must enable with startup parameter "
```

arangod/RestHandler/RestTelemetricsHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+                  "`--server.send-telemetrics`.");
+    return RestStatus::DONE;
+  }
+
+  auto const& apiPolicy = gs.supportInfoApiPolicy();
+  TRI_ASSERT(apiPolicy != "disabled");
```
jsteemann, line 54: This assertion does not hold because the telemetrics API is added unconditionally in the GeneralServerFeature. If the user sets `--server.support-info-api-policy` to `disabled`, then the assertion would be violated here.
See https://github.com/arangodb/arangodb/pull/18174/files#r1138455710

arangod/RestHandler/RestTelemetricsHandler.cpp
```diff
@@ -0,0 +1,79 @@
...
+    return RestStatus::DONE;
+  }
+
+  VPackBuilder result;
+
+  bool isLocal = _request->parsedValue("local", false);
+  SupportInfoBuilder::buildInfoMessage(result, _request->databaseName(),
+                                       _server, isLocal, true);
+  generateResult(rest::ResponseCode::OK, result.slice());
+  _response->setHeader(StaticStrings::Location, _request->requestPath());
```
jsteemann, line 76: Why is a location header set here unconditionally? I don't think we should return a location header here outside of a well-defined test, especially as the redirect will redirect to itself...

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+                                         ArangodServer& server) {
+  DatabaseFeature& dbFeature = server.getFeature<DatabaseFeature>();
+
+  std::vector<std::string> databases = methods::Databases::list(server, "");
+
+  containers::FlatHashMap<std::string_view, uint32_t> dbViews;
+  for (auto const& database : databases) {
+    auto vocbase = dbFeature.lookupDatabase(database);
+
+    LogicalView::enumerate(*vocbase,
```
jsteemann, line 93: It is not safe to use the return value of `dbFeature.lookupDatabase(...)` unconditionally. The API returns a raw pointer, and it can return a nullptr!
Instead, please use `dbFeature.useDatabase(...)`, which returns a VocbasePtr RAII object. You still need to check whether it is a nullptr though.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+    size_t numGraphColls;
+    size_t numSmartColls;
+    size_t numDisjointGraphs;
+    VPackBuilder builder;
+  };
+
+  std::unordered_map<std::string_view, DbCollStats> visitedDatabases;
+  containers::FlatHashMap<size_t, containers::FlatHashSet<std::string_view>>
+      visitedColls;
+  containers::FlatHashMap<size_t, size_t> collNumDocs;
```
jsteemann, line 113: Some comments here about what the `size_t` keys in these maps mean would help the reader of the code.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+  // merge all collections belonging to the database into
+  // the same object, as the database might be in more
+  // than one dbserver
+  for (auto const dbItFromServers : velocypack::ObjectIterator(infoSlice)) {
+    for (auto const dbIt : velocypack::ArrayIterator(dbItFromServers.value)) {
```
jsteemann, line 118: No need to make const here
```suggestion
  for (auto dbItFromServers : velocypack::ObjectIterator(infoSlice)) {
    for (auto dbIt : velocypack::ArrayIterator(dbItFromServers.value)) {
```

client-tools/Shell/ShellFeature.cpp
```diff
@@ -80,6 +83,10 @@ void ShellFeature::collectOptions(
 
   options->addOption("--javascript.run-main", "Execute main function.",
                      new BooleanParameter(&_runMain));
+
+  options->addOption("--telemetrics.send-to-endpoint",
+                     "Send telemetrics to endpoint",
+                     new BooleanParameter(&_sendToEndpoint));
```
jsteemann, line 89: If this option is only useful for testing purposes, shouldn't we
1. mark it as Uncommon
2. only enable it in maintainer mode
?

client-tools/Shell/ShellFeature.cpp
```diff
@@ -190,4 +202,44 @@ void ShellFeature::start() {
...
+void ShellFeature::getTelemetricsInfo(VPackBuilder& builder) {
+  if (_telemetricsHandler != nullptr) {
+    _telemetricsHandler->getTelemetricsInfo(builder);
+  }
+}
+
+void ShellFeature::startTelemetrics() {
+  if (_telemetricsHandler != nullptr) {
+    _telemetricsHandler.reset(nullptr);
+  }
```
jsteemann, line 227: An extra reset is not necessary because `_telemetricsHandler` is assigned to as the next operation.
Assigning a new object to an existing unique_ptr will always discard the old object before.

client-tools/Shell/ShellFeature.h
```diff
@@ -63,7 +81,9 @@ class ShellFeature final : public ArangoshFeature {
   std::vector<std::string> _positionals;
   std::string _unitTestFilter;
   std::vector<std::string> _scriptParameters;
+  std::unique_ptr<TelemetricsHandler> _telemetricsHandler;
   bool _runMain{false};
+  bool _sendToEndpoint{false};
```
jsteemann, line 86: Should this only be present in maintainer mode?
The member should also be renamed to something more telling. `sendToEndpoint` is nice, but send **what**? Tribbles?
```suggestion
  bool _sendToEndpoint{false};
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+    TRI_ASSERT(newEndpoint.get() != nullptr);
+    std::unique_ptr<httpclient::GeneralClientConnection> connection(
+        httpclient::GeneralClientConnection::factory(
+            client.getCommFeaturePhase(), newEndpoint, 30, 60, 3, sslProtocol));
+
+    TRI_ASSERT(connection != nullptr);
+
+    if (_httpClient != nullptr) {
+      _httpClient.reset(nullptr);
+    }
```
jsteemann, line 287: No need for an extra reset call here

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+      numRedirects = 0;
+    }
+    if (numRedirects == 0) {
+      timeoutInSecs *= 3;
+      timeoutInSecs = std::min<uint32_t>(600, timeoutInSecs);
+      _runCondition.wait_for(lk, std::chrono::seconds(timeoutInSecs),
+                             [this] { return _server.isStopping(); });
+    }
+    _httpClient.reset(nullptr);
+    connection.reset(nullptr);
```
jsteemann, line 328: No need to reset connection to a nullptr, because it is an on-stack `std::unique_ptr`, which will go out of scope just after this life anyway.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+                VPackValue(_telemetricsResponse.errorMessage()));
+    builder.close();
+  }
+}
+
+void TelemetricsHandler::arrangeTelemetrics() {
+  try {
+    this->fetchTelemetricsFromServer();
+
+    if (_telemetricsResponse.ok()) {
```
jsteemann, line 352: data race: `_telemetricsResponse` accessed without holding the mutex.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+void TelemetricsHandler::sendTelemetricsToEndpoint() {
+  std::string const originalUrl =
+      "https://europe-west3-telemetrics-project.cloudfunctions.net/"
+      "telemetrics-cf-3";
+
+  std::string url = originalUrl;
+
+  Result result = {TRI_ERROR_NO_ERROR};
+  uint32_t timeoutInSecs = 10;
+  uint64_t sslProtocol = TLS_V12;
```
jsteemann, line 242: Whenever possible, we should use TLS1.3. It has many advantages over TLS1.2 and should be understood by up-to-date servers.
```suggestion
  uint64_t sslProtocol = TLS_V13;
```

client-tools/Shell/TelemetricsHandler.h
```diff
@@ -0,0 +1,75 @@
...
+  void sendTelemetricsToEndpoint();
+  void arrangeTelemetrics();
+
+  ArangoshServer& _server;
+  std::mutex _mtx;
+  std::condition_variable _runCondition;
+  std::thread _telemetricsThread;
+  std::unique_ptr<httpclient::SimpleHttpClient> _httpClient;
+  Result _telemetricsResponse;
+  Result _response;
```
jsteemann, line 70: What is the difference between these two members? Are they still both needed?

client-tools/Shell/V8ClientConnection.cpp
```diff
@@ -1012,6 +1010,132 @@ static void ClientConnection_httpPostRaw(
...
+  TRI_V8_RETURN_TRUE();
+
+  TRI_V8_TRY_CATCH_END
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief ClientConnection method "sendTelemetricsToEndpointTestRedirect"
+////////////////////////////////////////////////////////////////////////////////
+
+static void ClientConnection_sendTelemetricsToEndpointTestRedirect(
```
jsteemann, line 1079: Not sure why this function exists. Can't we just somehow make the original sendTelemetrics stuff use a different URL if a test requires so?

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+      std::string_view dbName = dbIt.get("name").stringView();
+      if (visitedDatabases.find(dbName) == visitedDatabases.end()) {
+        visitedDatabases.insert(std::make_pair(dbName, DbCollStats()));
+        visitedDatabases[dbName].builder.openObject();
+        visitedDatabases[dbName].builder.add("n_views",
+                                             VPackValue(dbViews[dbName]));
+        visitedDatabases[dbName].builder.add(
+            "single_shard", VPackValue(dbIt.get("single_shard").getBoolean()));
+        visitedDatabases[dbName].builder.add("colls",
+                                             VPackValue(VPackValueType::Array));
```
jsteemann, line 128: This repeatedly looks up the `dbName` key in `visitedDatabases`. Works, but we just inserted to key, so we don't need to look it up repeatedly.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+                           [this] { return _server.isStopping(); });
+    timeoutInSecs *= 2;
+    timeoutInSecs = std::min<uint32_t>(100, timeoutInSecs);
+  }
+}
+
+// Tests the redirection when telemetrics is sent to an endpoint, though not
+// exactly the same thing, because here the endpoint is local, hence the same
+// client is used and a new endpoint is not created.
+void TelemetricsHandler::sendTelemetricsToEndpointTestRedirect(
```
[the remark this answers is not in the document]
cpjulia: This is the function that we agreed that I'd create for testing redirection using foxx, expose it to the js test and call it

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+        } else {
+          auto amountDocs = collNumDocs[planId];
+          collNumDocs[planId] +=
+              collIt.get("n_docs").getNumber<decltype(amountDocs)>();
+          visitedDatabases[dbName].builder.add(collIt);
+          std::string_view collType = collIt.get("type").stringView();
+          if (collType == "document") {
+            visitedDatabases[dbName].numDocColls++;
+          } else if (collType == "edge") {
+            visitedDatabases[dbName].numGraphColls++;
```
jsteemann, line 153: If we are counting edge collections here, shouldn't the statistics member be named `numEdgeColls` instead?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+    connection.reset(nullptr);
+  }
+}
+
+void TelemetricsHandler::getTelemetricsInfo(VPackBuilder& builder) {
+  // as this is for printing the telemetrics result in the tests, we send the
+  // telemetrics object if the request returned ok and the error if not to
+  // parse the error in the tests
+  std::unique_lock lk(_mtx);
+  if (_telemetricsResponse.ok()) {
```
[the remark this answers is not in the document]
cpjulia: This constructor already locks the mutex calling `_mtx.lock()`

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+          }
+          if (collIt.get("smart_graph").getBoolean()) {
```
jsteemann, line 155: Less code and less error-prone (works even when the value of `smart_graph` is not a boolean value, whereas the previous code will just throw)
```suggestion
          if (collIt.get("smart_graph").isTrue()) {
```

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+            visitedDatabases[dbName].numSmartColls++;
+          }
+          if (collIt.get("disjoint").getBoolean()) {
```
jsteemann, line 158: Less code and less error-prone (works even when the value of `smart_graph` is not a boolean value, whereas the previous code will just throw)
```suggestion
          if (collIt.get("disjoint").isTrue()) {
```

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+            visitedDatabases[dbName].numDisjointGraphs++;
```
jsteemann, line 159: Isn't it the number of disjoint SmartGraph edge collections rather than the number of disjoint graphs that we are counting here?
The member name in the statistics object (and also in the result that we build up) should IMO reflect that.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+      }
+    }
+    result.close();
+  }
+}
+
+// if it's a telemetrics request, keys can only have lowercase letters and
+// underscores
+void SupportInfoBuilder::normalizeKeyForTelemetrics(std::string& key) {
+  key = arangodb::basics::StringUtils::tolower(key.data());
```
jsteemann, line 215:
```suggestion
  arangodb::basics::StringUtils::tolowerInPlace(key);
```

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+      !isLocal;
+
+  result.openObject();
+
+  if (isSingleServer && !isActiveFailover) {
+    result.add("deployment", VPackValue(VPackValueType::Object));
+    if (isTelemetricsReq) {
+      // it's single server, but we maintain the format the same as cluster
+      std::string envValue;
+      TRI_GETENV("STARTUP_MODE", envValue);
```
jsteemann, line 248: Environment variables under our control should always be prefixed with `ARANGODB_`
```suggestion
      TRI_GETENV("ARANGODB_STARTUP_MODE", envValue);
```

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+      result.add("startup_mode", VPackValue(envValue));
+      result.add("persistedId",
+                 VPackValue("single_" + std::to_string(serverId)));
```
jsteemann, line 251: Why is it `startup_mode` but `persistedId`, i.e. both snake_case and camelCase mixed?

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+          result.add("persistedId",
+                     VPackValue(arangodb::basics::StringUtils::tolower(
+                         ServerState::instance()->getPersistedId())));
+        } else {
+          result.add("persistedId", VPackValue(serverId));
+        }
+
+        std::string envValue;
+        TRI_GETENV("STARTUP_MODE", envValue);
+        result.add("startup_mode", VPackValue(envValue));
```
jsteemann, line 298: For both, see above comments.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+
+      result.add("agents", VPackValue(agents));
+      result.add("coordinators", VPackValue(coordinators));
+      std::string dbServersKey = "dbServers";
+      std::string shardsStatitsics = "shards";
+      if (isTelemetricsReq) {
+        dbServersKey = "db_servers";
+        shardsStatitsics = "shards_statistics";
+      }
+      result.add(dbServersKey, VPackValue(dbServers));
```
jsteemann, line 413: This mix of snake_case and camelCase looks really weird.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+  }
+
+  result.add("role", VPackValue(ServerState::roleToString(
+                         ServerState::instance()->getRole())));
+  result.add("maintenance",
+             VPackValue(ServerState::instance()->isStartupOrMaintenance()));
+  containers::FlatHashMap<std::string_view, std::string_view> keys;
+
+  if (isTelemetricsReq) {
+    keys["readOnly"] = "read_only";
```
jsteemann, line 482: ok, why?
Instead of doing all this if...then...else stuff here in the server code, can't we just make the gathering endpoint use the attribute names that we had already in use. That would simplify a lot.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+
+  result.add("databases", VPackValue(VPackValueType::Array));
+  for (auto const& database : databases) {
+    auto vocbase = dbFeature.lookupDatabase(database);
+
+    result.openObject();
+    result.add("name", database);
+
+    size_t numDocColls = 0;
+    size_t numGraphColls = 0;
```
jsteemann, line 604: Isn't that rather `numEdgeColls`?

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,762 @@
...
+    size_t numSmartColls = 0;
+    size_t numDisjointGraphs = 0;
```
jsteemann, line 606: And here, numDisjointSmartGraphEdgeColls?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+      lk.lock();
+
+      result = this->checkHttpResponse(response, true);
+      if (auto errorNum = result.errorNumber();
+          errorNum == TRI_ERROR_NO_ERROR ||
+          errorNum == TRI_ERROR_HTTP_FORBIDDEN) {
+        _telemetricsResponse = std::move(result);
+        break;
+      } else {
+        _telemetricsResponse = std::move(result);
```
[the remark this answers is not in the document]
cpjulia: Will store it in the function that fetches telemetrics

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,380 @@
...
+    std::unique_ptr<httpclient::SimpleHttpResult> response(_httpClient->request(
+        rest::RequestType::POST, relative, compressedBody.data(),
+        compressedBody.size(), headers));
+    lk.lock();
+    result = this->checkHttpResponse(response, false);
+    if (result.ok()) {
+      auto returnCode = response->getHttpReturnCode();
+      if (returnCode == 200) {
+        break;
+      } else if (returnCode == 301 || returnCode == 302 || returnCode == 307) {
```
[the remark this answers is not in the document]
cpjulia: Made it retry only if got a 500

client-tools/Shell/TelemetricsHandler.h
cpjulia: Removed

arangod/Utils/SupportInfoBuilder.cpp
cpjulia: As agreed, I used the same builder structure for API `/_admin/support-info` and `/_admin/telemetrics`, so I have to make this distinction

cpjulia: To be able to gather the telemetrics response following its rules and maintaining `support-info` as it is, with no changes. In the former comment, `persistedId` was actually wrong and I changed it to `persisted_id`.

arangod/Utils/SupportInfoBuilder.cpp
cpjulia: I agree, but, but was told no

cpjulia: Though I've been getting OK from the endpoint, so it seems it can parse without these rules

client-tools/Shell/TelemetricsHandler.cpp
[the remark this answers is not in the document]
cpjulia: I can add a condition variable there that would be signaled by `fetchTelemetricsFromServer`, but then, if this function that would wait for the condition variable is called before `fetchTelemetricsFromServer` is called, it would wait for the condition to be satisfied until it timed out, which would take a few seconds, because it would not be useful to have a very small timeout

arangod/GeneralServer/GeneralServerFeature.cpp
```diff
@@ -202,14 +208,23 @@ void GeneralServerFeature::collectOptions(
   options->addOldOption("server.keep-alive-timeout", "http.keep-alive-timeout");
   options->addOldOption("no-server", "server.rest-server");
 
+  options
+      ->addOption("--server.send-telemetrics",
+                  "Whether to enable the telemetrics API.",
```
jsteemann, line 213: The name `send-telemetrics` is probably misleading. 
It is more about whether the server _has_ a telemetrics API or not.
Apart from that the switch is a bit problematic because you can enable the telemetrics API with this option, but then set `--server.support-info-api-policy` to `disabled`, which will effectively turn the API off again.

client-tools/Shell/ShellFeature.cpp
```diff
@@ -80,6 +83,11 @@ void ShellFeature::collectOptions(
 
   options->addOption("--javascript.run-main", "Execute main function.",
                      new BooleanParameter(&_runMain));
+
+  options->addOption(
```
jsteemann, line 87: Isn't this startup option relevant only for maintainer mode?
If it's disabled, what does it mean? That we fetch telemetrics from the server, but don't send it out to the collecting service? That sounds pretty useless, so the question is if we can somehow use a failure point or something similar instead of a (user-visible) startup option.

jsteemann: Aah, just saw that the option is now inside an `#ifdef ARANGODB_ENABLE_MAINTAINER_MODE` section.
So that's ok. But can we use something like a failure point here instead of a startup option?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    : _server(server), _sendToEndpoint(sendToEndpoint) {}
+
+TelemetricsHandler::~TelemetricsHandler() {
+  if (_telemetricsThread.joinable()) {
+    _telemetricsThread.join();
+  }
+}
+
+Result TelemetricsHandler::checkHttpResponse(
+    std::unique_ptr<httpclient::SimpleHttpResult> const& response) {
```
jsteemann, line 86: nice, this function is now side-effects free (except from throwing exceptions), so we can mark it as `const` now!

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+  if (response == nullptr || !response->isComplete()) {
+    return {TRI_ERROR_INTERNAL, this->_httpClient->getErrorMessage()};
+  }
+  if (response->wasHttpError()) {
+    auto errorNum = TRI_ERROR_INTERNAL;
+    std::string errorMsg = response->getHttpReturnMessage();
+    std::shared_ptr<velocypack::Builder> bodyBuilder;
+    // Handle case of no body:
+    try {
+      bodyBuilder = response->getBodyVelocyPack();
```
jsteemann, line 98:
```suggestion
    // Handle case of no body:
    try {
      auto bodyBuilder = response->getBodyVelocyPack();
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+      result = this->checkHttpResponse(response);
+      if (auto errorNum = result.errorNumber();
+          errorNum == TRI_ERROR_NO_ERROR ||
+          errorNum == TRI_ERROR_HTTP_FORBIDDEN) {
+        _telemetricsFetchedInfo.add(response->getBodyVelocyPack()->slice());
+        _telemetricsFetchResponse = std::move(result);
+        break;
+      } else {
+        _telemetricsFetchResponse = std::move(result);
+      }
```
jsteemann, line 144: `_telemetricsFetchResponse` is set in both the `if` and `else` cases, so the code can be simplified to:
```suggestion
      _telemetricsFetchResponse = std::move(result);
      if (auto errorNum = result.errorNumber();
          errorNum == TRI_ERROR_NO_ERROR ||
          errorNum == TRI_ERROR_HTTP_FORBIDDEN) {
        _telemetricsFetchedInfo.add(response->getBodyVelocyPack()->slice());
        break;
      }
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    }
+    _runCondition.wait_for(lk, std::chrono::seconds(timeoutInSecs),
+                           [this] { return _server.isStopping(); });
+    timeoutInSecs *= 2;
+    timeoutInSecs = std::min<uint32_t>(100, timeoutInSecs);
+  }
+}
+
+// Sends telemetrics to the endpoint and ests the redirection when telemetrics
```
jsteemann, line 153:
```suggestion
// Sends telemetrics to the endpoint and tests the redirection when telemetrics
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+          encoding::gzipCompress(reinterpret_cast<uint8_t const*>(body.data()),
+                                 body.size(), compressedBody);
+      if (res != TRI_ERROR_NO_ERROR) {
+        break;
+      }
+      headers.emplace(StaticStrings::ContentLength,
+                      std::to_string(body.size()));
+      headers.emplace(StaticStrings::ContentEncoding,
+                      StaticStrings::EncodingGzip);
+    }
```
jsteemann, line 199: Why is gzip compression only used for the originalUrl and not for other URLs?
If there is a rationale behind this, there should be a code comment explaining the reason.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+// created.
+std::optional<VPackBuilder> TelemetricsHandler::sendTelemetricsToEndpoint(
+    std::string const& reqUrl) {
+  VPackBuilder testResponseBuilder;
+
+  bool isLocal = false;
+  if (reqUrl.starts_with('/')) {
+    isLocal = true;
+  }
+  bool isOriginalUrl = (reqUrl.compare(kOriginalUrl) == 0) ? true : false;
```
jsteemann, line 165:
```suggestion
  bool isOriginalUrl = reqUrl == kOriginalUrl;
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+// endpoint is local, hence the same client is used and a new endpoint is not
+// created.
+std::optional<VPackBuilder> TelemetricsHandler::sendTelemetricsToEndpoint(
+    std::string const& reqUrl) {
+  VPackBuilder testResponseBuilder;
+
+  bool isLocal = false;
+  if (reqUrl.starts_with('/')) {
+    isLocal = true;
+  }
```
jsteemann, line 164:
```suggestion
  bool isLocal = reqUrl.starts_with('/');
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+          httpclient::GeneralClientConnection::factory(
+              client.getCommFeaturePhase(), newEndpoint, 30, 60, 3,
+              sslProtocol));
+
+      if (connection == nullptr) {
+        continue;
+      }
+      httpclient::SimpleHttpClientParams const params(30, false);
+      _httpClient = std::make_unique<httpclient::SimpleHttpClient>(
+          connection.get(), params);
```
jsteemann, line 230: Passing `connection.get()` to the SimpleHttpClient is problematic here, because the SimpleHttpClient then will not assume ownership of the connection. It will just store and use the raw pointer to the connection.
However, after line 230 the connection object will go out of scope and will be deleted. 
So the SimpleHttpClient will end up with a pointer to a deleted connection. UB.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+      auto [endpoint, relative, error] =
+          getEndpoint(nullptr, endpoints, url, lastEndpoint);
+      relativeUrl = std::move(relative);
+      ClientFeature& client =
+          _server.getFeature<HttpEndpointProvider, ClientFeature>();
+      std::unique_ptr<Endpoint> newEndpoint(Endpoint::clientFactory(endpoint));
+
+      lk.lock();
+
+      TRI_ASSERT(newEndpoint.get() != nullptr);
```
jsteemann, line 219: Does this assertion always hold?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+      std::unique_ptr<httpclient::GeneralClientConnection> connection(
+          httpclient::GeneralClientConnection::factory(
+              client.getCommFeaturePhase(), newEndpoint, 30, 60, 3,
+              sslProtocol));
+
+      if (connection == nullptr) {
+        continue;
```
jsteemann, line 226: If we `continue` here, we will have a busy loop.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+      if (!isOriginalUrl) {
+        auto& clientParams = _httpClient->params();
+        clientParams.setRequestTimeout(30);
+      }
+      lk.unlock();
+      if (isOriginalUrl) {
+        std::unique_ptr<httpclient::SimpleHttpResult> response(
+            _httpClient->request(rest::RequestType::POST, relativeUrl,
+                                 compressedBody.data(), compressedBody.size(),
+                                 headers));
```
jsteemann, line 242: The `response` object is created here and immediately after goes out of scope. Does not look right.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    timeoutInSecs *= 2;
+    timeoutInSecs = std::min<uint32_t>(100, timeoutInSecs);
+  }
+}
+
+// Sends telemetrics to the endpoint and ests the redirection when telemetrics
+// is sent to an endpoint, though not exactly the same thing, because here the
+// endpoint is local, hence the same client is used and a new endpoint is not
+// created.
+std::optional<VPackBuilder> TelemetricsHandler::sendTelemetricsToEndpoint(
```
jsteemann, line 157: This function looks very complicated and should be simplified.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+  // as this is for printing the telemetrics result in the tests, we send the
+  // telemetrics object if the request returned ok and the error if not to
+  // parse the error in the tests
+  std::unique_lock lk(_mtx);
+
+  if (_telemetricsFetchResponse.ok() && !_telemetricsFetchedInfo.isEmpty()) {
+    builder.add(_telemetricsFetchedInfo.slice());
+  } else if (_telemetricsFetchResponse.fail()) {
+    builder.openObject();
+    builder.add("errorNum",
```
jsteemann, line 306:
```suggestion
    builder.add(StaticStrings::ErrorNum,
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+                VPackValue(_telemetricsFetchResponse.errorNumber()));
+    builder.add("errorMessage",
```
jsteemann, line 308:
```suggestion
    builder.add(StaticStrings::ErrorMessage,
```

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+  }
+}
+
+void TelemetricsHandler::getTelemetricsInfo(VPackBuilder& builder) {
+  // as this is for printing the telemetrics result in the tests, we send the
+  // telemetrics object if the request returned ok and the error if not to
+  // parse the error in the tests
+  std::unique_lock lk(_mtx);
+
+  if (_telemetricsFetchResponse.ok() && !_telemetricsFetchedInfo.isEmpty()) {
```
jsteemann, line 302: Can there be a race here, e.g. if no telemetrics data have been fetched **yet**, but this function is already called?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,348 @@
...
+      if (_sendToEndpoint) {
+        this->sendTelemetricsToEndpoint();
+      }
+#else
+      this->sendTelemetricsToEndpoint();
+#endif
+    }
+  } catch (std::exception const& err) {
+    LOG_TOPIC("e1b5b", WARN, arangodb::Logger::FIXME)
+        << "Exception on handling telemetrics " << err.what();
```
jsteemann, line 331:
```suggestion
        << "Exception on handling telemetrics: " << err.what();
```

client-tools/Shell/TelemetricsHandler.h
```diff
@@ -0,0 +1,81 @@
...
+  Result checkHttpResponse(
+      std::unique_ptr<httpclient::SimpleHttpResult> const& response);
+  void fetchTelemetricsFromServer();
+
+  void arrangeTelemetrics();
+
+  ArangoshServer& _server;
+  std::mutex _mtx;
+  std::condition_variable _runCondition;
+  std::condition_variable _runCondition2;
```
jsteemann, line 73: Not needed?
```suggestion
```

client-tools/Shell/TelemetricsHandler.h
```diff
@@ -0,0 +1,81 @@
...
+  void getTelemetricsInfo(VPackBuilder& builder);
+
+  void sendTelemetricsToEndpointTestRedirect(VPackBuilder& builder);
+  std::optional<VPackBuilder> sendTelemetricsToEndpoint(
+      std::string const& reqUrl = kOriginalUrl);
+
+ private:
+  static constexpr char kOriginalUrl[] =
+      "https://europe-west3-telemetrics-project.cloudfunctions.net/"
+      "telemetrics-cf-3";
```
jsteemann, line 62: Why not use a constexpr `std::string_view`?
```suggestion
  static constexpr char kOriginalUrl[] =
      "https://europe-west3-telemetrics-project.cloudfunctions.net/"
      "telemetrics-cf-3";
```

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,94 @@
...
+      if (res !== undefined || numSecs >= 16) {
+        break;
+      }
+      internal.sleep(numSecs);
+      numSecs *= 2;
+    }
+    assertNotUndefined(res);
+    return res;
+  } catch (err) {
+  }
```
jsteemann, line 56: Why catch all errors here and silence them? This will silence the potential throw by the `assertNotUndefined(res)` in line 53 too.
Catching exceptions here is unnecessary anyway, as the caller of this function assumes a certain response structure anyway, and will fail when this function returns `undefined`.
So it would be better and simpler to just remove the try ... catch.

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function telemetricsOnShellTestsuite() {
+
+  return {
+
+    testTelemetricsShellRequestByUserNotEnabled: function () {
+      try {
+        arango.startTelemetrics();
+        const res = getTelemetricsResult();
```
jsteemann, line 66: There is a potential race here: `startTelemetrics()` can return before the result has been fetched, so `getTelemetrics()` will see an empty result.

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,94 @@
...
+        assertTrue(res.hasOwnProperty("errorNum"));
+        assertTrue(res.hasOwnProperty("errorMessage"));
+        assertEqual(res.errorNum, internal.errors.ERROR_HTTP_FORBIDDEN.code);
+        assertTrue(res.errorMessage.includes("telemetrics is disabled"));
```
jsteemann, line 70: This cannot work, as the server-side error message does not contain this string anymore:
https://github.com/arangodb/arangodb/pull/18174/files#diff-dd5cc47b8e7a09f138a1ee123e08cf7b05f5697d89e782bbeb520dcc3c58f763R54

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,94 @@
...
+      } catch (err) {
+      }
```
jsteemann, line 72: If all errors are silenced here anyway, the test is completely useless.
Please remove the try ... catch.

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,94 @@
...
+
+function telemetricsApiUsageTestsuite() {
+
+  return {
+
+    testTelemetricsApiRequestByUserNotEnabled: function () {
+      arango.startTelemetrics();
+      const res = arango.GET("/_admin/telemetrics");
+      assertEqual(res.errorNum, internal.errors.ERROR_HTTP_FORBIDDEN.code);
+      assertTrue(res.errorMessage.includes("telemetrics is disabled"));
```
jsteemann, line 86: This cannot work, as the server-side error message has been changed and does not contain this text:
https://github.com/arangodb/arangodb/pull/18174/files#diff-dd5cc47b8e7a09f138a1ee123e08cf7b05f5697d89e782bbeb520dcc3c58f763R54

tests/js/client/server_parameters/test-statistics-enabled-false.js
jsteemann, line 1: All changes in this file can and should be reverted.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,766 @@
...
+            }
+
+            containers::FlatHashMap<std::string, size_t> idxTypesToAmounts;
+
+            auto flags = Index::makeFlags(Index::Serialize::Estimates,
+                                          Index::Serialize::Figures);
+            std::vector<std::string_view> idxTypes = {
+                "edge",      "geo",        "hash",      "fulltext", "inverted",
+                "no-access", "persistent", "iresearch", "skiplist", "ttl",
+                "zkd",       "primary",    "unknown"};
```
jsteemann, line 691: I don't think that an index of type "no-access" actually exists. Where is the type from?
I think older deployments can also have old indexes of type "geo1" and "geo2". These should be mapped to "geo" simply.

tests/js/client/server_parameters/test-telemetrics-false-auth.js
cpjulia: That's why `getTelemetricsResult()` has a loop, for trying to read the telemetrics until the `startTelemetrics()` function the thread executes has fetched the telemetrics and stored the json it fetched

client-tools/Shell/ShellFeature.h
```diff
@@ -63,7 +90,11 @@ class ShellFeature final : public ArangoshFeature {
   std::vector<std::string> _positionals;
   std::string _unitTestFilter;
   std::vector<std::string> _scriptParameters;
+  std::unique_ptr<TelemetricsHandler> _telemetricsHandler;
   bool _runMain{false};
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
```
jsteemann, line 95: Here, the member is made available if `ARANGODB_ENABLE_MAINTAINER_MODE`, but above (https://github.com/arangodb/arangodb/pull/18174/files#diff-e43190b071f8b4c4829e9e3080f74b3481e1a8d06f9f14ceada0485931022dfcR59) the member is modified if `ARANGODB_ENABLE_FAILURE_TESTS`. 
This is inconsistent.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+  }
+}
+
+// Sends telemetrics to the endpoint and tests the redirection when telemetrics
+// is sent to an endpoint, though not exactly the same thing, because here the
+// endpoint is local, hence the same client is used and a new endpoint is not
+// created.
+std::optional<VPackBuilder> TelemetricsHandler::sendTelemetricsToEndpoint(
+    std::string const& reqUrl) {
+  VPackBuilder testResponseBuilder;
```
jsteemann, line 155: why is this variable called `testResponseBuilder`?

tests/js/client/server_parameters/test-telemetrics-false-auth.js
```diff
@@ -0,0 +1,90 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global getOptions, assertEqual, assertTrue, assertNotUndefined, arango */
+
+// //////////////////////////////////////////////////////////////////////////////
+// / @brief dropping followers while replicating
+// /
```
jsteemann, line 6:
```suggestion
```

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global getOptions, assertEqual, assertTrue, assertNotNull, assertNotUndefined, arango */
+
+// //////////////////////////////////////////////////////////////////////////////
+// / @brief dropping followers while replicating
+// /
```
jsteemann, line 6:
```suggestion
```

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+      if (res !== undefined || numSecs >= 16) {
+        break;
+      }
+      internal.sleep(numSecs);
+      numSecs *= 2;
+    }
+    assertNotUndefined(res);
+    return res;
+  } catch (err) {
+  }
```
jsteemann, line 86: This try ... catch unconditionally silences all errors. It should not be there at all.

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
```
jsteemann, line 104: This try ... catch unconditionally silences all errors. It should not be there.

client-tools/Shell/TelemetricsHandler.cpp
cpjulia: Because it's the builder from the response that is gonna be sent when the function `sendTelemetricsToEndpoint()` is used to test redirection

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+        db._useDatabase(databaseName);
+
+        let numShards = 1;
+        if (isCluster) {
+          numShards = 2;
+        }
+        db._create(cn, {numberOfShards: numShards, replicationFactor: 1});
+        for (let i = 0; i < 2000; ++i) {
+          db[cn].insert({value: i + 1, name: "abc"});
+        }
```
jsteemann, line 270: Please willify this test setup, i.e. use a single insert only.

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+              assertEqual(coll["n_persistent"], 0);
+              if (!isCluster) {
+                assertEqual(coll["rep_factor"], 1);
+              }
+              assertEqual(coll["n_geo"], 0);
+            });
+          }
+        });
+      } finally {
+        try {
```
jsteemann, line 330: There should be a `db._useDatabase("_system");` in the finally part.
It currently is somewhere in the try part, but it is not guaranteed that it was reached, so the finally may run and try to drop the database while not being in the `_system` database itself. Then the `_dropDatabase` in the finally will throw.

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+        db._useDatabase(databaseName);
+
+        let numShards = 1;
+        if (isCluster) {
+          numShards = 2;
+        }
+        db._create(cn, {numberOfShards: numShards, replicationFactor: 1});
+        for (let i = 0; i < 2000; ++i) {
+          db[cn].insert({value: i + 1, name: "abc"});
+        }
```
jsteemann, line 394: Please willify

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+              assertEqual(coll["n_persistent"], 0);
+              if (!isCluster) {
+                assertEqual(coll["rep_factor"], 1);
+              }
+              assertEqual(coll["n_geo"], 0);
+            });
+          }
+        });
+      } finally {
+        try {
```
jsteemann, line 456: Please add `db._useDatabase("_system");` here too.

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+        db._useDatabase(databaseName);
+
+        let numShards = 1;
+        if (isCluster) {
+          numShards = 2;
+        }
+        db._create(cn, {numberOfShards: numShards, replicationFactor: 1});
+        for (let i = 0; i < 2000; ++i) {
+          db[cn].insert({value: i + 1, name: "abc"});
+        }
```
jsteemann, line 513: Please willify.

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+              assertEqual(coll["n_persistent"], 0);
+              if (!isCluster) {
+                assertEqual(coll["rep_factor"], 1);
+              }
+              assertEqual(coll["n_geo"], 0);
+            });
+          }
+        });
+      } finally {
+        try {
```
jsteemann, line 571: please add `db._useDatabase("_system");`

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -0,0 +1,777 @@
...
+        arango.startTelemetrics();
+        getTelemetricsResult();
+        const telemetrics = getTelemetricsSentToEndpoint();
+        assertForTelemetricsResponse(telemetrics);
+      } finally {
+        try {
+          db._dropDatabase(databaseName);
+        } catch (err) {
+        }
+      }
```
jsteemann, line 760: Why? I don't see a database being created by this test.

arangod/Utils/SupportInfoBuilder.cpp
```diff
@@ -0,0 +1,765 @@
...
+            } else {
+              result.add("disjoint", VPackValue(false));
+            }
+
+            containers::FlatHashMap<std::string, size_t> idxTypesToAmounts;
+
+            auto flags = Index::makeFlags(Index::Serialize::Estimates,
+                                          Index::Serialize::Figures);
+            std::vector<std::string_view> idxTypes = {
+                "edge",     "geo",      "geo1",       "geo2",      "hash",
```
jsteemann, line 689: why is adding `geo1` and `geo2` here necessary if you normalize the index name further down below (https://github.com/arangodb/arangodb/pull/18174/files#diff-7de15e18bf4e086618a2d93c76dd3837c798908d1ee5647300ca7a384a9198cbR730-R732) anyway?

client-tools/Shell/TelemetricsHandler.cpp
jsteemann: Yes, but is `test` any relevant in a function that is called `sendTelemetricsToEndpoint`? That function could be called from anywhere, at least in theory. Why limit it artifically and assume the reader of the code knows it is called only from tests? That makes things just puzzling for the reader.
Why not simply name it `responseBuilder`, because that is exactly what this variable is used for. Regardless of whether the function is called from a test or from somewhere else.

jsteemann (approved):
LGTM

PR #18179, opened 2023-02-23
Partial fix for PRESUPP-539

jsteemann:
### Scope & Purpose

Partial fix for https://arangodb.atlassian.net/browse/PRESUPP-539

* Partial fix for PRESUPP-539: account for memory used during AQL condition transformation to disjunctive normal form (DNF). This transformation can use a lot of memory for complex filter conditions, which was previously not accounted for. Now if the transformation uses a lot of memory and hits the configured query memory limit, the query will rather be aborted than overuse memory.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/PRESUPP-539
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Condition.cpp
+++ b/arangod/Aql/Condition.cpp
@@ -38,6 +38,7 @@
 #include "Basics/ScopeGuard.h"
 #include "Basics/StaticStrings.h"
 #include "Containers/FlatHashSet.h"
+#include "Containers/SmallVector.h"
 #include "Indexes/Index.h"
 #include "Logger/LogMacros.h"
 #include "Transaction/CountCache.h"
@@ -71,7 +72,7 @@ namespace {
 // is a condition that excludes null (e.g. != null). if this is tracked first,
 // we are sure the index attribute value cannot be null and we can still use
 // the sparse index
-int operationWeight(AstNode const* node) {
+int operationWeight(AstNode const* node) noexcept {
   switch (node->type) {
     case NODE_TYPE_OPERATOR_BINARY_NE:
       // != before ==, e.g. attr != null && attr == FUNC(abc) for hash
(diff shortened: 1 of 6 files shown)


arangod/Aql/Condition.cpp
```diff
@@ -1914,29 +1974,60 @@ AstNode* Condition::transformNodePostorder(
...
+
+      // check if there would be too many members in the DNF version of
+      // the condition. we may suffer from combinatorial explosion here.
+      if (orMembers >= maxNumberOfConditionMembers) {
+        // throw a special error about a too complex query condition.
+        // this will be handled by the callers, so that they will continue
+        // with a simplified version of the condition (which may not be
+        // usable for index lookups).
+        THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_FAILED,
+                                       "too complex query condition");
```
goedderz, line 2003: If I read that correctly, you want to catch only this specific error in `transformCondition`. However, it seems to be used for other things as well - wouldn't it make sense to introduce a specific error code for this, or did I misunderstand the intention?

goedderz (approved):

arangod/Aql/Condition.cpp
jsteemann: Yes, the intention was to catch only this specific error. I reused TRI_ERROR_FAILED because it is not used often in other places (at least in AQL). I can introduce a specific error for this case to make it more future-proof.

jsteemann: Addressed.

goedderz (approved):

PR #18182, opened 2023-02-24
[CINFRA-513] Remove syncShardsWithLogs from tests

apetenchea:
### Scope & Purpose

Until the Collection Group Supervision was implemented, we had to reply on a hack to synchronize the participants of a shard in `Plan/Collections` with the participants of the corresponding replicated log. This is now done in the supervision.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature


maierlars (approved):

maierlars:
only the already known shutdown hanger.

PR #18188, opened 2023-02-24
Fix arangorestore API to create _rev on shard leader [BTS-1232].

neunhoef:
### Scope & Purpose

This is about https://arangodb.atlassian.net/browse/BTS-1232 .

Short description: 

arangorestore uses the (undocumented) API /_api/replication/restore-data
implemented in the RestReplicationHandler. Since ArangoDB 3.8, we
create new _rev values during the restore operation. Unfortunately,
this creation of _rev values was done on the coordinator(s) and since
these are hybrid logical clock (HLC) values, it can happen that the
generated _rev values on two different coordinators are the same for
two different documents. This will clash on the leader of the OneShard
database, since there, the newly generated _rev values are used as
internal document IDs, which directly translate into a RocksDB key in
the DocumentsColumnFamily. This leads to a collision and this manifests
itself as a lock timeout, since two different RocksDB transactions try
to touch the same key. However, even if no lock timeout would have
happened, the consequences would have been dire: Namely, there would
have been a corruption between primary index and data in the Documents
column family. We would have had two primary index entries pointing to
the same document.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1232
- [*] Original PR: https://github.com/arangodb/arangodb/pull/18187


jsteemann (approved):
LGTM

mchacki (approved):
LGTM

PR #18195, opened 2023-02-25
Reduce build time

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/1295

Cleanup includes, split up files, use pimpl idiom for some classes to avoid expensive includes in the header (asio!), ...

I compared this PR agains vanilla devel with the following cmake command:
```
cmake -DCMAKE_CXX_COMPILER="/usr/bin/clang++-14" -DCMAKE_C_COMPILER="/usr/bin/clang-14" -DCMAKE_C_COMPILER_AR=/usr/bin/llvm-ar-14 -DCMAKE_CXX_COMPILER_AR=/usr/bin/llvm-ar-14 -DCMAKE_C_COMPILER_RANLIB=/usr/bin/llvm-ranlib-14 -DCMAKE_CXX_COMPILER_RANLIB=/usr/bin/llvm-ranlib-14 -DUSE_MAINTAINER_MODE=On -DUSE_ENTERPRISE=On -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_GOOGLE_TESTS=On -DUSE_FAILURE_TESTS=On -DUSE_STRICT_OPENSSL_VERSION=Off -DUSE_FAIL_ON_WARNINGS=On -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" ..
```

I ran a complete rebuild from scractch (no cache) for the targets arangod and arangodbtests. These are the results from my machine:
```
This PR:
real    13m08.175s
user    187m36.230s
sys     8m9.790s

devel:
real    14m27.133s
user    213m24.737s
sys     9m38.821s
```
So we have a speedup of almost 10%.

There is a lot more that can be done, but I will do that in separate PRs.


ankit-gautam23 (commented):
Could you plz split the PR

mpoeter:
@ankit-gautam23 can you elaborate why you want me to split the PR? I tend to keep this PR as is and only update it to latest devel so it can be merged eventually. I will not add any more changes but create separate PRs for further cleanup.

neunhoef (approved):
LGTM

arangod/GeneralServer/CommTask.cpp
```diff
@@ -516,7 +515,6 @@ void CommTask::executeRequest(std::unique_ptr<GeneralRequest> request,
     return;
   }
 
-  TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
   SchedulerFeature::SCHEDULER->trackCreateHandlerTask();
```
MBkkt, line 518: Why?

MBkkt (commented):
LGTM for search part.

About adding new virtual methods for scheduler and make pimpl for connection pool, I don't think it's good idea but I don't strong against it.

arangod/Aql/Functions.cpp
```diff
@@ -5286,11 +5288,7 @@ AqlValue functions::IpV4ToNumber(ExpressionContext* expressionContext,
 
     struct in_addr addr;
     memset(&addr, 0, sizeof(struct in_addr));
-#if _WIN32
-    int result = InetPton(AF_INET, &buffer[0], &addr);
-#else
     int result = inet_pton(AF_INET, &buffer[0], &addr);
```
jsteemann, line 5291: Glad that we can simplify this code.
Is `inet_pton` guaranteed to be available in all Windows environments we are targeting?

arangod/GeneralServer/RestHandler.cpp
```diff
@@ -73,7 +73,6 @@ RestHandler::~RestHandler() {
     // someone forgot to call trackTaskEnd 🤔
     TRI_ASSERT(PriorityRequestLane(determineRequestLane()) ==
                RequestPriority::LOW);
-    TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
```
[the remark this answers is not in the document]
mpoeter: That was by mistake. Will revert.

arangod/GeneralServer/CommTask.cpp
```diff
@@ -516,7 +515,6 @@ void CommTask::executeRequest(std::unique_ptr<GeneralRequest> request,
     return;
   }
 
-  TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
```
jsteemann, line 519: nit: I think the assertion is still valid. I would vote to keep it.

arangod/GeneralServer/RestHandler.cpp
jsteemann, line 76: Same here.

arangod/Aql/Functions.cpp
mpoeter: Yes, I checked that.

arangod/GeneralServer/CommTask.cpp
mpoeter: Thanks! The assertions got lost because I previously had another change which affected the scheduler, and when I reverted those and removed the changes I apparently also removed those assertions. Will check all places and restore them.

jsteemann (approved):
LGTM

PR #18206, opened 2023-02-28
improve debuggability of errors upon database creation failure

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/18207

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1218

This PR combines 2 independent improvements: Improve debuggability of database creation failures (in case of exceptions) and make a test more robust.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [x] Backport for 3.10: this PR
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1218
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #18207, opened 2023-02-28
improve debuggability of errors upon database creation failure

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1217

This PR combines 2 independent improvements: Improve debuggability of database creation failures (in case of exceptions) and make a test more robust.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18206
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1217
- [ ] Design document:


MBkkt (approved):

tests/test-definitions.txt
```diff
@@ -41,7 +41,7 @@ shell_client_multi priority=2500 parallelity=2 single suffix=vst -- --vst true
 shell_client_transaction priority=750 single suffix=http -- $EncryptionAtRest
 shell_client_transaction priority=1750 single suffix=http2 -- --http2 true
 shell_client_transaction priority=2750 single suffix=vst -- --vst true
-shell_client_traffic priority=250 parallelity=1 single suffix=http -- http true
+shell_client_traffic priority=250 parallelity=1 single suffix=http
```
dothebart, line 44: both is wrong. it needs to be `-- --http true`. With this change the default protocol (which could be different) would be used.

tests/js/client/shell/shell-auto-refill-index-caches-on-modify-noncluster.js
```diff
@@ -31,15 +31,45 @@ const getMetric = require('@arangodb/test-helper').getMetricSingle;
...
 const n = 5000; // documents
 
-const waitForMetrics = () => {
-  // wait for metrics to settle, using an informal API
+const waitForPendingRefills = () => {
+  // wait for pending refill operations to have finished, using an informal API
   arango.POST('/_api/index/sync-caches', {});
 };
 
+const runWithRetry = (cb) => {
```
dothebart, line 39: This function should be put to test-helper.js

dothebart (changes requested):
Please move the utility function to the test-helpers, and make sure the transport protocol remains pinned properly.

tests/js/client/shell/shell-auto-refill-index-caches-on-modify-noncluster.js
dothebart: https://github.com/arangodb/release-test-automation/blob/main/test_data/makedata.js#L139 already has a similar function.

tests/test-definitions.txt
jsteemann: I know it should be `-- --http true` to explicitly make it use the HTTP protocol.
But I just looked at the other lines in the file (e.g. line 41 from the above excerpt), where we **DO NOT SET** the protocol explicitly when it is HTTP.
So for consistency I just did the same here.

tests/js/client/shell/shell-auto-refill-index-caches-on-modify-noncluster.js
jsteemann: This `runWithRetry` function does something specific to a collection used in the test, e.g.
```
      if (tries++ === 3) {
        throw err;
      }   

      db[cn].truncate();
      require("internal").sleep(2);
```
In order to make this function generic, I would need to add a callback function for the failure case, make the sleep value configurable and also the number of tries. This would add 3 more parameters to the generic function.
```
const runWithRetry = (cb, failureCb, tries, sleepTimeout) => { ... }
```
I would also have to adjust all callers of this function.

Not sure if that's really worth it. I doubt it.

tests/test-definitions.txt
dothebart: then all of the should be nailed to make this shure please.

tests/js/client/shell/shell-auto-refill-index-caches-on-modify-noncluster.js
dothebart: Y? 
If s.b. needs it, he can adapt it to get more functionality? 
I have cleaned up so many of these "just need them here" helper functions through out the tests in the last 24 months, we definitely should stop that now.

jsteemann: done

dothebart (approved):

PR #18215, opened 2023-02-28
[CINFRA-669] Prevent replicated state usage before they're ready

goedderz:
### Scope & Purpose

Replicated states mustn't be used before they're ready. For a leader state, that means its state machine must not be accessible before recovery finished successfully. For a follower state it means its state machine must not be accessible before it has a "current" snapshot, i.e. it will be able to catch up to the leader with its log and snapshot without having to acquire a new snapshot.

This was implemented earlier, but (deliberately) temporarily removed in #17528. This reimplements that protection.

Trivial changes that are necessary for this were moved into its own PR to make it easier to review: https://github.com/arangodb/arangodb/pull/18276
That one must be merged first.

Ticket: https://arangodb.atlassian.net/browse/CINFRA-669

- [x] :hankey: Bugfix

### Checklist

  - [X] C++ **Unit tests**


arangod/Replication2/ReplicatedState/StateStatus.cpp
maierlars, line 1: file is essentially empty

maierlars (approved):

PR #18221, opened 2023-03-01
add options for BlobDB configuration

jsteemann:
### Scope & Purpose

* Provide options for configuring and enabling RocksDB's blob storage (BlobDB)
  for large documents in the documents column family.
  This is currently an experimental feature.

  The following experimental options are available:

  - `--rocksdb.enable-blob-files`: Enable the usage of blob files for the
    documents column family. This option defaults to `false`. All following
    options are only relevant if this option is set to `true`.
  - `--rocksdb.min-blob-size`: Size threshold for storing large documents in
    blob files (in bytes, 0 = store all documents in blob files).
  - `--rocksdb.blob-file-size`: Size limit for blob files in the documents
    column family (in bytes).
  - `--rocksdb.blob-compression-type`: Compression algorithm to use for blob
    data in the documents column family.
  - `--rocksdb.enable-blob-cache`: Enable caching of blobs in the block cache
    for the documents column family.
  - `--rocksdb.enable-blob-garbage-collection`: Enable blob garbage collection
    during compaction in the documents column family.
  - `--rocksdb.blob-garbage-collection-age-cutoff`: Age cutoff for garbage
    collecting blob files in the documents column family (percentage value from
    0 to 1 determines how many blob files are garbage collected during
    compaction).
  - `--rocksdb.blob-garbage-collection-force-threshold`: Garbage ratio
    threshold for scheduling targeted compactions for the oldest blob files
    in the documents column family.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


dothebart:
oops, wrong browser window, sorry

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2894,6 +2894,12 @@ DECLARE_GAUGE(rocksdb_block_cache_capacity, uint64_t,
 DECLARE_GAUGE(rocksdb_block_cache_pinned_usage, uint64_t,
               "rocksdb_block_cache_pinned_usage");
 DECLARE_GAUGE(rocksdb_block_cache_usage, uint64_t, "rocksdb_block_cache_usage");
+#ifdef ARANGODB_ROCKSDB8
+// DECLARE_GAUGE(rocksdb_block_cache_entries, uint64_t,
+//                    "rocksdb_block_cache_entries");
+// DECLARE_GAUGE(rocksdb_block_cache_charge_per_entry, uint64_t,
+//                    "rocksdb_block_cache_charge_per_entry");
+#endif
```
mpoeter, line 2902: Should we remove the comments since this is already ifdefed?

arangod/RocksDBEngine/RocksDBOptionFeature.cpp
```diff
@@ -1276,12 +1504,29 @@ void RocksDBOptionFeature::validateOptions(
...
   if (_compactionStyle != ::kCompactionStyleLevel) {
     LOG_TOPIC("6db54", WARN, Logger::ENGINES)
         << "using compaction style '" << _compactionStyle
         << "' is experimental and not supported for production usage";
   }
 
+  if (_blockCacheType == ::kBlockCacheTypeHyperClock) {
+#ifndef ARANGODB_ROCKSDB8
+    // cannot be reached with RocksDB 7.2
+    TRI_ASSERT(false);
```
mpoeter, line 1522: Should we make this a ADB_PROD_ASSERT? I know we should never be able get here, but let's play safe. :)

mpoeter (approved):

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: The comments need to be present, because without them the `check-guidelines` check in Jenkins will fail.
This check verifies that for all declared metrics there is a corresponding file in `Documentation`. The comments disable the check for these metrics.

PR #18214, opened 2023-02-28
[GORDO-1554] Key generation ASSERT inspection

hkernbach:
### Scope & Purpose

Fixes invalid key insertion issue in EnterpriseGraph Vertex/Documents.
There is a EE Companion: https://github.com/arangodb/enterprise/pull/1220

Details:
[GORDO-1554] Key generation ASSERT inspection

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :book: CHANGELOG entry made
- [x] Backports
  - [ ] Backport for 3.10: *(Please link PR)* YES REQUIRED

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/GORDO-1554
- [x] EE PR: https://github.com/arangodb/enterprise/pull/1220

[GORDO-1554]: https://arangodb.atlassian.net/browse/GORDO-1554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #18230, opened 2023-03-02
[GORDO-1554] Key generation ASSERT inspection

hkernbach:
3.10 Backport of: https://github.com/arangodb/arangodb/pull/18214
Please see all details there.

**There is also an EE Companion:** 
* https://github.com/arangodb/enterprise/pull/1222

### DIFF
--- a/tests/js/common/shell/shell-document-babies.js
+++ b/tests/js/common/shell/shell-document-babies.js
@@ -29,6 +29,7 @@ var ERRORS = arangodb.errors;
 var db = arangodb.db;
 var internal = require("internal");
 var wait = internal.wait;
+const cluster = require("internal").isCluster();
 
 ////////////////////////////////////////////////////////////////////////////////
 /// @brief test suite: babies for documents
@@ -45,7 +46,7 @@ function CollectionDocumentSuiteBabies() {
     /// @brief set up
     ////////////////////////////////////////////////////////////////////////////////
 
-    setUp: function() {
+    setUp: function () {
       db._drop(cn);
       collection = db._create(cn, {
         waitForSync: false
@@ -56,7 +57,7 @@ function CollectionDocumentSuiteBabies() {
     /// @brief tear down
     ////////////////////////////////////////////////////////////////////////////////
 
-    tearDown: function() {
+    tearDown: function () {
       if (collection) {
         collection.unload();
         collection.drop();
@@ -69,7 +70,7 @@ function CollectionDocumentSuiteBabies() {
     /// @brief insert multiple documents and remove them again
     ////////////////////////////////////////////////////////////////////////////////
 
-    testInsertRemoveMulti: function() {
+    testInsertRemoveMulti: function () {
       var docs = collection.insert([{}, {}, {}]);
       assertEqual(docs.length, 3);
       collection.remove(docs);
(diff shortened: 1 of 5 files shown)


hkernbach:
Unrelated jenkins failure

markuspf (approved):
LGTM

tests/js/common/shell/shell-document-babies.js
```diff
@@ -305,30 +306,43 @@ function CollectionDocumentSuiteBabies() {
...
-        assertEqual(docs[1].errorNum, ERRORS.ERROR_ARANGO_DOCUMENT_KEY_BAD.code);
-        collection.remove("a");
+        }];
+
+        if (!cluster) {
+          const docs = collection.insert(batch);
+          assertEqual(docs.length, 2);
+          assertEqual(docs[0]._key, "a");
+          assertTrue(docs[1].error);
+          assertEqual(docs[1].errorNum, ERRORS.ERROR_ARANGO_DOCUMENT_KEY_BAD.code);
```
dothebart, line 325: sequence of parameters should be similar; a third parameter with the full document would be even better.

tests/js/common/shell/shell-document-babies.js
```diff
@@ -305,30 +306,43 @@ function CollectionDocumentSuiteBabies() {
...
+          collection.remove("a");
+          assertEqual(collection.count(), 0);
+        } else {
+          try {
+            collection.insert(batch);
```
jsteemann, line 330: Wouldn't this different test code mean that we establish a new behavior difference between single server and cluster?
An insert should not have different behavior between single server and cluster. When a batch of documents is inserted, the overall insert should **not** fail even if some/all of the operations in the batch fail. Instead, the caller must check the individual results of each operation.

tests/js/common/shell/shell-keygen.js
```diff
@@ -32,29 +32,246 @@ const jsunity = require("jsunity");
...
+  // single insert, using document API
+  invalidKeys.forEach(invalidKey => {
+    try {
+      db[vn].save({_key: invalidKey, value: "43"});
+    } catch (ignore) {
+    }
+  });
+
+  // batch insert, using document API
+  const batch = [];
```
jsteemann, line 219: Should `batch` be const here? I know JavaScript is quite special about `const`, but I always keep forgetting what exactly is constant when you declare a JS variable `const`.

tests/js/common/shell/shell-keygen.js
```diff
@@ -32,29 +32,246 @@ const jsunity = require("jsunity");
...
+
+  // Note: Using save() <-> insert() should be the same.
+  // Using both here to double-check that the assumption is
+  // proven.
+
+  // single insert, using document API
+  invalidKeys.forEach(invalidKey => {
+    try {
+      db[vn].insert({_key: invalidKey, value: "43"});
+    } catch (ignore) {
```
jsteemann, line 207: If this insert attempt is supposed to fail, we should add `fail()` after it for clarity.
I know at the end we are checking the number of docs in the collection, but checking for failures immediately here would be preferrable IMO.

tests/js/common/shell/shell-keygen.js
hkernbach: Yes it is fine. Explanation can be found here e.g.: https://stackoverflow.com/a/61517230

tests/js/common/shell/shell-keygen.js
```diff
@@ -32,29 +32,246 @@ const jsunity = require("jsunity");
...
+    }
+  });
+  // single insert, using document API
+  invalidKeys.forEach(invalidKey => {
+    try {
+      db[vn].save({_key: invalidKey, value: "43"});
```
jsteemann, line 213: same here (and for all following occurrences)

tests/js/common/shell/shell-keygen.js
hkernbach: Basically it means, you cannot change the type of batch. It must stay an array.

hkernbach:
Jenkins green

jsteemann (approved):
LGTM

PR #18232, opened 2023-03-02
[GORDO-1557] Create proper worker actor state

jvolmer:
Adds vocbase and algorithm to worker actor state. Algorithm makes the worker generic over Vertex-, Edge- and Message-Type which are all dependent on the algorithm used.


markuspf (approved):
LGTM

PR #18237, opened 2023-03-03
Added `--rocksdb.auto-refill-index-caches-on-followers`

jsteemann:
### Scope & Purpose

Added startup option `--rocksdb.auto-refill-index-caches-on-followers` to control whether automatic refilling of in-memory caches should happen on followers or just leaders. The default value is `true`, i.e. refilling happens on followers too.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18239
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18249
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2868,6 +2868,18 @@ void RocksDBEngine::scheduleFullIndexRefill(std::string const& database,
   f.scheduleFullIndexRefill(database, collection, iid);
 }
 
+bool RocksDBEngine::autoRefillIndexCaches() const {
```
MBkkt, line 2871: NIT 
maybe better to get user access to the RocksDBIndexCacheRefillFeature& instead of two methods?

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
```diff
@@ -165,6 +166,20 @@ or cache-enabled persistent indexes.)");
...
+          "Whether or not to automatically (re-)fill the in-memory index "
+          "caches on followers as well.",
+          new options::BooleanParameter(&_autoRefillOnFollowers),
+          arangodb::options::makeFlags(options::Flags::DefaultNoComponents,
+                                       options::Flags::OnDBServer,
+                                       options::Flags::OnSingle))
+      .setIntroducedIn(31005)
+      .setLongDescription(R"(Set this to `false` to only (re-)fill in-memory
+index caches on leaders and save memory on followers. 
+Note that the value of this option should be identical for all DBServers.)");
```
MBkkt, line 182: We really need cluster wide options. 
It's not first case when we want to force user set same value per all DBServers/Coordinators

arangod/Transaction/Methods.cpp
```diff
@@ -3160,13 +3159,42 @@ Future<Result> Methods::replicateOperations(
...
+  if (options.refillIndexCaches == RefillIndexCaches::kDontRefill) {
+    // index cache refilling opt-out
+    reqOpts.param(StaticStrings::RefillIndexCachesString, "false");
+
+    TRI_IF_FAILURE("RefillIndexCacheOnFollowers::failIfFalse") {
+      THROW_ARANGO_EXCEPTION(TRI_ERROR_DEBUG);
+    }
+  } else {
     // this attribute can have 3 values: default, true and false. only
     // expose it when it is not set to "default"
```
MBkkt, line 3173: This statement is not true for your code.
Also I interested why

arangod/Transaction/Methods.cpp
```diff
@@ -3160,13 +3159,42 @@ Future<Result> Methods::replicateOperations(
...
+                       .server()
+                       .template getFeature<EngineSelectorFeature>()
+                       .engine();
+
+    bool refill = (options.refillIndexCaches == RefillIndexCaches::kRefill) ||
+                  (options.refillIndexCaches == RefillIndexCaches::kDefault &&
+                   engine.autoRefillIndexCaches());
+    if (!engine.autoRefillIndexCachesOnFollowers()) {
+      refill = false;
+    }
```
MBkkt, line 3184:
```suggestion
    bool const refill = engine.autoRefillIndexCachesOnFollowers() && 
                  ((options.refillIndexCaches == RefillIndexCaches::kRefill) ||
                  (options.refillIndexCaches == RefillIndexCaches::kDefault &&
                   engine.autoRefillIndexCaches()));
```

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: I intentionally didn't want storage engine-agnostic code (such as transaction/Methods.cpp) to access RocksDB-specific code/classes.
That's why I generalized the access via virtual functions in the generic storage engine interface.

arangod/Transaction/Methods.cpp
jsteemann: Ok, internally it is an enum with 3 values.
On the API level, it is a boolean that can be either true (maps to `kRefill`) or false (maps to `kDontRefill). However, not specifying the URL parameter at all in a request leads to us not having any boolean value. This maps to `kDefault` (no value set, but use configured default value).

arangod/RocksDBEngine/RocksDBIndexCacheRefillFeature.cpp
jsteemann: It is not critical if this option is set to different values on different DB servers, i.e. there will be no crashes, no data loss etc. if the option is set to different values.
It's just the functionality then won't work as desired.

I think I don't want to introduce (forced) cluster-wide options now as part of this PR.

MBkkt: Of course, it's not message for some immediately action. Just thoughts

arangod/RocksDBEngine/RocksDBEngine.cpp
MBkkt: Ok

MBkkt (approved):

PR #18238, opened 2023-03-03
[GORDO-1554] 2

hkernbach:
### Scope & Purpose

Follow up PR to related to: https://github.com/arangodb/arangodb/pull/18214
Soft-error in case a non-string value has been used.


jsteemann (approved):
LGTM

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6028,10 +6028,13 @@ std::vector<ServerID> ClusterInfo::getCurrentDBServers() {
...
 std::shared_ptr<std::vector<ServerID> const> ClusterInfo::getResponsibleServer(
     std::string_view shardID) {
-  if (auto result = getResponsibleServerReplication2(shardID);
-      result != nullptr) {
-    return result;
+  if (!shardID.empty()) {
+    if (auto result = getResponsibleServerReplication2(shardID);
+        result != nullptr) {
+      return result;
+    }
```
markuspf, line 6035: Is this really what you mean to do? Can replication2 not deal with empty shardIDs? and should then always the replication1 function be called, even if replication2 is used?

apetenchea: If the `shardID` is empty, the replication2 version will return `nullptr` (or look for the replicated log with ID 0, which is even worse). Not great, but for now I don't see the change in this PR as a breaking change, it's fine to stay.

markuspf (approved):

PR #18243, opened 2023-03-04
fix races in in-memory cache system

jsteemann:
### Scope & Purpose

Partially address https://arangodb.atlassian.net/browse/BTS-1255:

Fix shutdown races (shutdown of a cache could overlap with spawned FreeMemory and Migrate tasks for the same cache, leading to assertion failures because the amount of allocated memory went below zero). This fixes errors we have seen in cache unit tests in devel recently, although the race conditions have been for longer.
In order to prevent the races, a new per-cache `_shutdownLock` is introduced, which is acquired in exclusive mode by the shutdown, and acquired in shared mode while FreeMemory or Migrate tasks execute. That way the shutdown is mutually exclusive with a cache's task execution.
Note: the `shutdown()` function of a cache will be called only when the cache manager is being shut down (shutdown of the entire cache subsystem) or when the cache manager's static `destroyCache(cache)` function is called for a particular cache. The latter happens during index destructors (for indexes with in-memory caches), after `truncate()` calls (when the caches can be wiped and recreated) and after `unload()` calls.

Also added more assertions for verifying some of the cache subsystem's memory usage invariants.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cache/Cache.cpp
+++ b/arangod/Cache/Cache.cpp
@@ -32,7 +32,6 @@
 
 #include "Basics/SpinLocker.h"
 #include "Basics/SpinUnlocker.h"
-#include "Basics/cpu-relax.h"
 #include "Basics/voc-errors.h"
 #include "Cache/CachedValue.h"
 #include "Cache/Common.h"
@@ -53,8 +52,7 @@ Cache::Cache(Manager* manager, std::uint64_t id, Metadata&& metadata,
              std::shared_ptr<Table> table, bool enableWindowedStats,
              std::function<Table::BucketClearer(Metadata*)> bucketClearer,
              std::size_t slotsPerBucket)
-    : _taskLock(),
-      _shutdown(false),
+    : _shutdown(false),
       _enableWindowedStats(enableWindowedStats),
       _findHits(),
       _findMisses(),
@@ -329,6 +327,8 @@ std::shared_ptr<Table> Cache::table() const {
 }
 
 void Cache::shutdown() {
+  SpinLocker shutdownGuard(SpinLocker::Mode::Write, _shutdownLock);
+
   SpinLocker taskGuard(SpinLocker::Mode::Write, _taskLock);
   auto handle = shared_from_this();  // hold onto self-reference to prevent
                                      // pre-mature shared_ptr destruction
@@ -343,7 +343,7 @@ void Cache::shutdown() {
       }
 
       SpinUnlocker taskUnguard(SpinUnlocker::Mode::Write, _taskLock);
-      std::this_thread::sleep_for(std::chrono::microseconds(10));
+      std::this_thread::sleep_for(std::chrono::microseconds(20));
     }
 
     std::shared_ptr<cache::Table> table = this->table();
@@ -375,15 +375,6 @@ bool Cache::canResize() noexcept {
 
   SpinLocker metaGuard(SpinLocker::Mode::Read, _metadata.lock());
   return !(_metadata.isResizing() || _metadata.isMigrating());
-}
-
-bool Cache::canMigrate() noexcept {
-  if (ADB_UNLIKELY(isShutdown())) {
-    return false;
-  }
-
-  SpinLocker metaGuard(SpinLocker::Mode::Read, _metadata.lock());
-  return !_metadata.isMigrating();
 }
 
 /// TODO Improve freeing algorithm
@@ -431,6 +422,12 @@ bool Cache::freeMemory() {
 
 bool Cache::migrate(std::shared_ptr<Table> newTable) {
   if (ADB_UNLIKELY(isShutdown())) {
+    // unmarking migrating flag
+    SpinLocker metaGuard(SpinLocker::Mode::Write, _metadata.lock());
+
+    TRI_ASSERT(_metadata.isMigrating());
+    _metadata.toggleMigrating();
+    TRI_ASSERT(!_metadata.isMigrating());
     return false;
   }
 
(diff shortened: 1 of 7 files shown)


arangod/Cache/Table.cpp
```diff
@@ -220,15 +220,16 @@ std::uint32_t Table::logSize() const noexcept { return _logSize; }
 
 Table::BucketLocker Table::fetchAndLockBucket(std::uint32_t hash,
                                               std::uint64_t maxTries) {
+  std::uint32_t index = (hash & _mask) >> _shift;
```
MBkkt, line 223: NIT

uint32_t & uint64_t -> uint64_t
uint64_t >> uint32_t -> uint64_t

so implicit cast?

arangod/Cache/Manager.cpp
```diff
@@ -196,6 +204,12 @@ void Manager::shutdown() {
     if (!_shuttingDown) {
       _shuttingDown = true;
     }
+
+    while (globalProcessRunning()) {
+      // wait for rebalancer and migration tasks to complete
+      std::this_thread::yield();
```
MBkkt, line 210: yield doesn't work like this at least on linux.

In general issue is scheduler algorithm which give every thread quant of time.
Now your thread executed -> so then you call sched_yield scheduler see that now time for your thread and your thread continue executing.

As Linus said using sched_yield almost always mistake (exception probably is not default scheduler something like real-time scheduler, fifo scheduler etc -- slow but with some guarantees)

cpu_relax than sleep should be used.
Example https://github.com/facebook/folly/blob/main/folly/synchronization/detail/Sleeper.h

arangod/Cache/Manager.cpp
```diff
@@ -338,11 +357,14 @@ std::tuple<bool, Metadata, std::shared_ptr<Table>> Manager::registerCache(
   }
 
   if (ok) {
-    metadata =
-        Metadata(Cache::kMinSize, fixedSize, table->memoryUsage(), maxSize);
-    ok = increaseAllowed(metadata.allocatedSize - table->memoryUsage(), true);
+    std::uint64_t memoryUsage = table->memoryUsage();
+    metadata = Metadata(Cache::kMinSize, fixedSize, memoryUsage, maxSize);
+    TRI_ASSERT(metadata.allocatedSize >= memoryUsage);
```
MBkkt, line 362: is equal ok?
It looks unnecessary to call other part of code in such case

arangod/Cache/Manager.cpp
```diff
@@ -621,7 +649,7 @@ ErrorCode Manager::rebalance(bool onlyCalculate) {
 
 void Manager::shrinkOvergrownCaches(Manager::TaskEnvironment environment) {
   TRI_ASSERT(_lock.isLockedWrite());
-  for (auto it : _caches) {
+  for (auto& it : _caches) {
     std::shared_ptr<Cache>& cache = it.second;
```
MBkkt, line 653: NIT 
maybe to make it more clear
```suggestion
  for (auto& [_, cache] : _caches) {
```

arangod/Cache/Cache.cpp
```diff
@@ -343,7 +343,10 @@ void Cache::shutdown() {
       }
 
       SpinUnlocker taskUnguard(SpinUnlocker::Mode::Write, _taskLock);
-      std::this_thread::sleep_for(std::chrono::microseconds(10));
+      SpinUnlocker shutdownUnguard(SpinUnlocker::Mode::Write, _shutdownLock);
+
+      // sleep a bit without holding the locks
+      std::this_thread::sleep_for(std::chrono::microseconds(20));
```
MBkkt, line 349: Why do we need acquire this spinlocks before first cycle iteration, but we don't need to require afrer cycle gone or on next operation?

In general looks like bug.

MBkkt (changes requested):

arangod/Cache/Cache.cpp
jsteemann: Mmh, don't understand.
We have the spinlock acquired for each iteration in that loop.

arangod/Cache/Table.cpp
jsteemann: All of these (`hash`, `_mask`, `_shift`) are uint32_t values, so I don't think an implicit cast is actually happening here.

arangod/Cache/Manager.cpp
jsteemann: `metadata.allocatedSize` will probably always be bigger than `memoryUsage` here. But this is an implementation detail.
But for the assertions that care about `_globalAllocation` and `_fixedAllocation` this should not matter too much. For these two, we always (and only) want to check the invariant that `_globalAllocation` is always >= `_fixedAllocation`.

arangod/Cache/Manager.cpp
jsteemann: Done

arangod/Cache/Manager.cpp
jsteemann: Ok, I added `cpu_relax()` here.

arangod/Cache/Cache.cpp
MBkkt: Can you name line number for me?
I don't see where lock acquired in cycle.

MBkkt: Hmm or Unlocker call unlock in ctor and then call lock in dtor?

Ok I understand now.

MBkkt (approved):

neunhoef (approved):
LGTM

mpoeter (approved):

PR #18245, opened 2023-03-04
improve query optimizer

jsteemann:
### Scope & Purpose

Addresses issues in https://arangodb.atlassian.net/browse/PRESUPP-546

* Allow usage of projections and covering indexes in more cases.
  Previously, projections were not used if there were complex filter conditions
  on the index attribute(s) that contained the `[*]` expansion operator
  with inline FILTERs or RETURNs, e.g. 
  `FILTER doc.addrs[* FILTER CURRENT.country == 'US'].zip`.
* make AQL optimizer rule `simplify-conditions` correctly report that it was
  triggered. Previously that rule never reported that it was triggered
  although even though it actually was.
* support stringification (and thus comparison) of Ast NOP nodes
  the stringification is used to compare conditions against each other,
  e.g. for removing duplicate condition parts, or remove conditions from
  FILTERs that are also covered by indexes.
  not being able to stringify some conditions led to duplicate condition
  parts to not always be removed (for some special nodes, such as array
  expansions with inline filters, e.g. doc.a[* FILTER CURRENT.b > 3]`).

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18244
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/PRESUPP-546
- [ ] Design document:


arangod/Aql/OptimizerRules.cpp
```diff
@@ -2608,9 +2608,6 @@ void arangodb::aql::simplifyConditionsRule(Optimizer* opt,
       AstNode* simplified = plan->getAst()->traverseAndModify(root, visitor);
       if (simplified != root) {
         nn->expression()->replaceNode(simplified);
-      }
-      // cppcheck-suppress knownConditionTrueFalse
-      if (modifiedNode) {
```
MBkkt, line 2613: "if" always was false?
Could you explain for me why did it change to always true?

arangod/Aql/Condition.cpp
```diff
@@ -1754,8 +1754,13 @@ bool Condition::canRemove(ExecutionPlan const* plan, ConditionPart const& me,
         }
       }
     }
-  } catch (...) {
-    // simply ignore any errors and return false
+  } catch (std::exception const& ex) {
```
MBkkt, line 1757: What about callers this function, do they except exceptions from this code?

arangod/Aql/Ast.cpp
```diff
@@ -2846,8 +2853,11 @@ bool Ast::getReferencedAttributesRecursive(
         }
       }
       state.seen.clear();
-      // don't descend into the expansion itself (already handled it)
-      return false;
+      // if we are looking at a traversal output variable, we don't
+      // descend into the expansion itself (already handled it).
+      // however, we want and must descend into subnodes in case we were
+      // not called for a traversal output variable.
+      return expectedAttribute.empty();
```
MBkkt, line 2860: Maybe not empty?
I'm not sure, just tried to understand

Dronplane (approved):
LGTM

arangod/Aql/OptimizerRules.cpp
MBkkt: @Dronplane explained.

Dronplane: Looks like `modifiedNode` value is not used anymore (only written) and could be removed

Dronplane (approved):
LGTM

arangod/Aql/Condition.cpp
jsteemann: Previously all exceptions were caught and silenced, without letting the caller know about it.
Now we do effectively the same (catching all exceptions of type `std::exception`, which is the only type of exception that could be thrown by any of the code in question). We are still not returning any indication of the exception to the caller, but simply log it in trace mode.
It is normally not a problem if an exception is caught here and silenced. It only means that we will miss some potential optimization opportunity for an AQL condition expression. But the query should still function normally with or without the exception happening.

arangod/Aql/OptimizerRules.cpp
jsteemann: Fixed

MBkkt (approved):

PR #18253, opened 2023-03-06
Remove dependency on ApplicationServer from GraphFormat

markuspf:
### Scope & Purpose

Long standing TODO item: Remove dependency of GraphFormat on ApplicationServer.


jvolmer (approved):
nice

PR #18256, opened 2023-03-06
Implement Inspection for std::set, std::unordered_set and the like

markuspf:

### DIFF
--- a/lib/Inspection/include/Inspection/Access.h
+++ b/lib/Inspection/include/Inspection/Access.h
@@ -60,6 +60,8 @@ template<class Inspector, class T>
     return f.tuple(x);
   } else if constexpr (detail::IsMapLike<TT>::value) {
     return f.map(x);
+  } else if constexpr (detail::IsSetLike<TT>::value) {
+    return f.set(x);
   } else if constexpr (detail::IsListLike<TT>::value) {
     return f.list(x);
   }
--- a/lib/Inspection/include/Inspection/Access.h
+++ b/lib/Inspection/include/Inspection/Access.h
@@ -60,9 +60,8 @@ template<class Inspector, class T>
     return f.tuple(x);
   } else if constexpr (detail::IsMapLike<TT>::value) {
     return f.map(x);
-  } else if constexpr (detail::IsSetLike<TT>::value) {
-    return f.set(x);
-  } else if constexpr (detail::IsListLike<TT>::value) {
+  } else if constexpr (detail::IsListLike<TT>::value or
+                       detail::IsSetLike<TT>::value) {
     return f.list(x);
   }
 }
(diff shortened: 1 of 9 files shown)


lib/Inspection/include/Inspection/Access.h
```diff
@@ -60,6 +60,8 @@ template<class Inspector, class T>
     return f.tuple(x);
   } else if constexpr (detail::IsMapLike<TT>::value) {
     return f.map(x);
+  } else if constexpr (detail::IsSetLike<TT>::value) {
+    return f.set(x);
```
markuspf, line 64: Might be able to use list here now

lib/Inspection/include/Inspection/LoadInspectorBase.h
```diff
@@ -413,6 +421,21 @@ struct LoadInspectorBase : InspectorBase<Derived, Context> {
...
+  template<class T>
+  Status processSet(T& set) {
+    std::size_t idx = 0;
+    return this->self().doProcessList([&](auto value) -> Status {
+      auto ff = this->self().make(value);
+      typename T::value_type val;
+      if (auto res = process(ff, val); !res.ok()) {
+        return {std::move(res), std::to_string(idx), Status::ArrayTag{}};
+      }
+      set.insert(std::move(val));
```
markuspf, line 433: I couldn't find a neater way to process a set, beacuse  the function we found with `IsSetLike` has to be used after all.

mpoeter: I think we can ditch this whole function and handle everything in `processList` like this:
```
  template<class T>
  Status processList(T& list) {
    constexpr isSet = detail::IsSetLike<T>::value;
    std::size_t idx = 0;
    return this->self().doProcessList([&](auto value) -> Status {
      auto ff = this->self().make(value);
      typename T::value_type val;
      if (auto res = process(ff, val); !res.ok()) {
        return {std::move(res), std::to_string(idx), Status::ArrayTag{}};
      }
      if constexpr (isSet) {
        list.insert(std::move(val));
      } else {
        list.push_back(std::move(val));
      }
      ++idx;
      return {};
    });
  }
  ```

File: lib/Inspection/include/Inspection/SaveInspectorBase.h [hunk 3]
```diff
@@ -78,6 +78,14 @@ struct SaveInspectorBase : InspectorBase<Derived, Context> {
...
   }
 
+  template<class T>
+  [[nodiscard]] auto set(T const& set) {
+    auto& f = this->self();
+    return f.beginArray()                                                     //
+           | [&]() { return f.processList(std::begin(set), std::end(set)); }  //
+           | [&]() { return f.endArray(); };                                  //
+  }
+
```
mpoeter, line 88: IMHO this should not be needed anymore.

mpoeter (approved):

lib/Inspection/include/Inspection/SaveInspectorBase.h
markuspf: Indeed.

PR #18260, opened 2023-03-07
[GORDO-1548] Be able to use master context on every algorithm

jvolmer:
This PR makes it possible to create a MasterContext for every Algorithm running in Pregel.
Before, only some algorithms used the MasterContext, the others used member variables in the Conductor (e.g. `_totalVerticesCount`, _`totalEdgesCount` or `_aggregators`) to keep track of the status of the computation. The MasterContext keeps also track of these variables, therefore it makes sense to unify the algorithms and use MasterContext for all algorithms.


jvolmer:
Is now integrated into https://github.com/arangodb/arangodb/pull/18277

PR #18261, opened 2023-03-07
Bug fix/reduce load for coverage

dothebart:
### Scope & Purpose

Coverage runs can't bear the load of these. Reduce the testset.


dothebart:
It seems we want to delete this test anyways.

PR #18262, opened 2023-03-07
[GORDO-1548] Add loading conductor state

jvolmer:
Adds loading conductor state: This state tells the workers to load the graph. After each worker sent back a GraphLoaded message, the state changes to Computing. The GraphLoaded message from each worker include the number of loaded vertices and edges: The Loading state accumulates them and creates with them a MasterContext - the computation state of the algorithm - which is given to the Computation state.
The worker handler does not really load the graph for now - see comment in Worker/Handler.h - here we need GORDO-1546 that gets rid of the feature dependency of the GraphStore.
This PR also changes the workers vector in the ConductorState into a set to make it easier to check if specific worker PIDs are included.


markuspf (approved):
LGTM

PR #18265, opened 2023-03-07
fix races in in-memory cache system (#18243)

jsteemann:
### Scope & Purpose

Partially address https://arangodb.atlassian.net/browse/BTS-1255:

Fix shutdown races (shutdown of a cache could overlap with spawned FreeMemory and Migrate tasks for the same cache, leading to assertion failures because the amount of allocated memory went below zero). This fixes errors we have seen in cache unit tests in devel recently, although the race conditions have been for longer.
In order to prevent the races, a new per-cache `_shutdownLock` is introduced, which is acquired in exclusive mode by the shutdown, and acquired in shared mode while FreeMemory or Migrate tasks execute. That way the shutdown is mutually exclusive with a cache's task execution.
Note: the `shutdown()` function of a cache will be called only when the cache manager is being shut down (shutdown of the entire cache subsystem) or when the cache manager's static `destroyCache(cache)` function is called for a particular cache. The latter happens during index destructors (for indexes with in-memory caches), after `truncate()` calls (when the caches can be wiped and recreated) and after `unload()` calls.

Also added more assertions for verifying some of the cache subsystem's memory usage invariants.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.10.5 (XXXX-XX-XX)
 --------------------
 
+* Fix potential memory under-accounting on cache shutdown for in-memory caches
+  for edge indexes.
+
 * Added new geo_s2 ArangoSearch analyzer (Enterprise Only).
 
 * GORDO-1554: Fixes invalid document insertion with invalid user-specified keys
--- a/arangod/Cache/Cache.cpp
+++ b/arangod/Cache/Cache.cpp
@@ -32,7 +32,6 @@
 
 #include "Basics/SpinLocker.h"
 #include "Basics/SpinUnlocker.h"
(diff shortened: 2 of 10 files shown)


MBkkt (approved):

neunhoef (approved):
LGTM

PR #18269, opened 2023-03-07
add more transparency for WAL file deletions

jsteemann:
### Scope & Purpose

Adds more log messages for log topic `engines` in DEBUG and TRACE mode to see what's going on w.r.t. to pruning obsolete WAL files. 

Also adds some code to remove WAL files from the archive directory at startup if the files are exactly 0 bytes small.
Finally removes an invalid `noexcept` specifier from a constructor which could throw an exception (when initializing its rocksdb::ManagedSnapshot member, which will call `new`).

Adds the following new metrics for tracking WAL files:
- `rocksdb_live_wal_files`: number of alive WAL files (not archived)
- `rocksdb_wal_released_tick_flush`: lower bound sequence number from which
  onwards WAL files will be kept (i.e. not deleted from the archive) because
  of external flushing needs. Candidates for these are arangosearch links
  and background index creation.
- `rocksdb_wal_released_tick_replication`: lower bound sequence number from
  which onwards WAL files will be kept because they may be needed by the
  replication.
- `arangodb_flush_subscriptions`: number of currently active flush
   subscriptions.

Gives names to FlushSubscriptions, so that if a FlushSubscription is blocking WAL file deletion, it can be figured out by turning on TRACE logging for the "flush" log topic.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18291
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18293
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBEngine.h
```diff
@@ -459,7 +461,7 @@ class RocksDBEngine final : public StorageEngine {
 
   class RocksDBSnapshot final : public StorageSnapshot {
    public:
-    explicit RocksDBSnapshot(rocksdb::DB& db) noexcept : _snapshot(&db) {}
+    explicit RocksDBSnapshot(rocksdb::DB& db) : _snapshot(&db) {}
```
MBkkt, line 464: Is our code ready to have exception from this ctor?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -3665,4 +3715,34 @@ std::shared_ptr<StorageSnapshot> RocksDBEngine::currentSnapshot() {
...
+      // file size is exactly 0 bytes
+      LOG_TOPIC("e79dd", DEBUG, Logger::ENGINES)
+          << "found empty WAL file in archive at startup: '" << f
+          << "', scheduling this file for later deletion";
+      _prunableWalFiles.emplace(basics::FileUtils::buildFilename("archive", f),
+                                TRI_microtime() + _pruneWaitTime);
+    }
+  }
+
+  _metricsPrunableWalFiles.operator=(_prunableWalFiles.size());
```
MBkkt, line 3745: Why operator=?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -3665,4 +3715,34 @@ std::shared_ptr<StorageSnapshot> RocksDBEngine::currentSnapshot() {
...
 }
 
+void RocksDBEngine::removeEmptyJournalFilesFromArchive() {
+  LOG_TOPIC("50812", DEBUG, Logger::ENGINES)
+      << "scanning WAL archive directory for empty files...";
+
+  std::string archiveDirectory =
+      basics::FileUtils::buildFilename(_dbOptions.wal_dir, "archive");
+
+  WRITE_LOCKER(lock, _walFileLock);
```
MBkkt, line 3725: lazy lock acquiring?

MBkkt (approved):

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Would be possible, but it is not important at this part of the code. The method is only called once during startup, and there are no concurrent readers or writers for `_walFileLock` at this point.
The lock is only here for consistency with other parts of the code.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Replaced with Gauge::store(). (this call and others in this file)

arangod/RocksDBEngine/RocksDBEngine.h
jsteemann: I think so. 
The only way to create `RocksDBSnapshot` is by calling the method `RocksDBEngine::currentSnapshot()`.
All of its 3 callers have wrapped the call into a try...catch.

Previously, if the ctor of `RocksDBSnapshot` would throw, the program would only std::terminate. Now it can continue, and the callers can handle the exception (looks like the callers expect exceptions anyway, because they already had try...catch blocks).

MBkkt: Good

arangod/RocksDBEngine/RocksDBEngine.cpp
MBkkt: Ok

Dronplane (approved):
LGTM

CHANGELOG
```diff
@@ -1,6 +1,16 @@
 devel
 -----
 
+* Added the following metrics for WAL file tracking:
+  - `rocksdb_live_wal_files`: number of alive WAL files (not archived)
+  - `rocksdb_wal_released_tick_flush`: lower bound sequence number from which
+    onwards WAL files will be kept (i.e. not deleted from the archivE) because
```
```suggestion
    onwards WAL files will be kept (i.e. not deleted from the archive) because
```

neunhoef (approved):
LGTM

PR #18274, opened 2023-03-08
[APM-283] Almost always use parallel gather

mpoeter:
### Scope & Purpose

Use parallel gather in almost all queries. The only case where we still cannot use parallel gather is when using traversals. All other queries should now be able to parallelize the gather node. This can not only speed up queries quite significantly, but also reduce the memory required by the coordinator for queries like `FOR i IN @n INSERT {} INTO collection`. The serial gather node processes one db server after the other, but since the new documents should be distributed evenly over all shards, the coordinator has to buffer documents that need to be sent to other servers.

Enterprise companion: https://github.com/arangodb/enterprise/pull/1229

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **integration tests**

#### Related Information

- [ ] Docs PR: 
- [x] Jira ticket:
	- https://arangodb.atlassian.net/browse/APM-283
	- https://arangodb.atlassian.net/browse/BTS-696


arangod/Aql/RestAqlHandler.cpp
```diff
@@ -382,6 +382,7 @@ RestStatus RestAqlHandler::useQuery(std::string const& operation,
 
   TRI_ASSERT(_engine != nullptr);
   TRI_ASSERT(std::to_string(_engine->engineId()) == idString);
+  QueryGuard guard(_engine);
```
MBkkt, line 385:
```suggestion
  std::lock_guard lock{engine->getQuery()};
```

arangod/Aql/QueryContext.h
```diff
@@ -103,6 +103,9 @@ class QueryContext {
 
   aql::Ast* ast();
 
+  void lock();
+  void unlock();
```
MBkkt, line 107: Maybe getLockGuard?

arangod/Aql/RestAqlHandler.h
```diff
@@ -136,6 +136,16 @@ class RestAqlHandler : public RestVocbaseBaseHandler {
   // our query registry
   QueryRegistry* _queryRegistry;
 
+  // Guard that keeps the lock on the QueryContext to serialize
+  // concurrent snippet execution
+  struct QueryGuard {
```
jsteemann, line 141: Probably it would be better to move the `QueryGuard` into QueryContext.h, and probably even allow creating one by calling a public method on the `QueryContext` class. Then we don't need to publicly expose `lock()` and `unlock()` methods in QueryContext at all, and the only way to get a lock would be by calling the one public method that returns a RAII `QueryGuard` object.

arangod/Aql/RestAqlHandler.h
```diff
@@ -136,6 +136,16 @@ class RestAqlHandler : public RestVocbaseBaseHandler {
...
+    explicit QueryGuard(aql::ExecutionEngine* engine);
```
jsteemann, line 142: I think we should disable copy construction and copy assignment of `QueryGuard` objects explicitly.
Plus we may want to explicitly define the behavior for move construction or move assignment.

tests/js/server/aql/aql-optimizer-rule-parallelize-gather-cluster.js
```diff
@@ -108,6 +104,10 @@ function optimizerRuleTestSuite () {
         "FOR doc IN " + cn + " SORT doc.value1 LIMIT 1000 RETURN doc",
         "FOR doc IN " + cn + " SORT doc.value1 LIMIT 1000, 1000 RETURN doc",
         "FOR i IN 1..1000 FOR doc IN " + cn + " FILTER doc.value == i RETURN doc",
+        "FOR doc IN " + cn + " LIMIT 10 UPDATE doc WITH {} IN " + cn,
+        "FOR i IN 1..1000 INSERT {} IN " + cn,
+        "FOR doc1 IN " + cn + " FOR doc2 IN " + cn + " FILTER doc1._key == doc2._key RETURN doc1",
+        "FOR doc1 IN " + cn + " FOR doc2 IN " + cn + " FOR doc3 IN " + cn + " FILTER doc1._key == doc2._key FILTER doc2._key == doc3._key RETURN doc1",
       ];
 
       if (require("internal").options()["query.parallelize-gather-writes"]) {
```
jsteemann, line 113: This `if` statement seems redundant nowadays, and should always evaluate to `true`. So we may as well remove it now.

CHANGELOG
```diff
@@ -1,6 +1,15 @@
 devel
 -----
 
+* APM-283: Use parallel gather in almost all queries. The only case were we
```
```suggestion
* APM-283: Use parallel gather in almost all queries. The only case where we
```

CHANGELOG
```diff
@@ -1,6 +1,15 @@
...
+  still cannot use parallel gather is when using traversals. All other queries
+  should now be able to parallelize the gather node. This can not only speed up
+  queries quite significantly, but also reduce the memory required by the
+  coordinator for queries like `FOR i IN @n INSERT {} INTO collection`. The
+  serial gather node processes one db server after the other, but since the new
```
jsteemann, line 9: The standard way of writing "db server" in our documentation seems to be "DB-Server":
```
$ grep -rio "db.*server" ./docs/3.11/ | grep -v "generated" | cut -d ":" -f 2 | sort | uniq -c | sort -n
...
      9 DBSERVER
     25 dbserver
     26 DB Server
     29 DBServer
     85 DB server
    305 DB-Server
```

```suggestion
  serial gather node processes one DB-Server after the other, but since the new
```

CHANGELOG
```diff
@@ -1,6 +1,15 @@
 devel
 -----
 
+* APM-283: Use parallel gather in almost all queries. The only case were we
+  still cannot use parallel gather is when using traversals. All other queries
```
jsteemann, line 5: This sounds rather pessimistic. According to the optimizer rule, we can actually parallelize the gather in some queries that use traversals.

Citing that code comment:
> we cannot parallelize the gather if we have a parallel traversal which
> itself depends again on a scatter/distribute node, because we are
> currently lacking synchronization for that scatter/distribute node.

We should probably use that some wording in the CHANGELOG to not sell the feature below its value.

arangod/Aql/QueryRegistry.cpp
```diff
@@ -182,146 +188,201 @@ void* QueryRegistry::openEngine(EngineId id, EngineType type) {
...
-        << "closing engine " << engineId
-        << ", query id: " << ei._queryInfo->_query->id()
-        << ", numOpen: " << ei._queryInfo->_numOpen;
-  } else {
-    LOG_TOPIC("ae981", TRACE, arangodb::Logger::AQL)
-        << "closing engine " << engineId << ", no query";
+  }
+
+  if (queryInfoToFinish) {
+    queryInfoToFinish->_promise.setValue(queryInfoToFinish->_query);
```
jsteemann, line 239: Is it safe to use the `_queryInfo` data here outside of the lock.
Can there be a race with the garbage collection of the QueryRegistry (haven't advanced this far down in this PR, but just wondering now when reading the changes from top to bottom).

arangod/Aql/QueryRegistry.h
```diff
@@ -142,6 +146,9 @@ class QueryRegistry {
 
     std::shared_ptr<ClusterQuery> _query;  // the actual query pointer
 
+    bool _finished = false;
```
jsteemann, line 149: Can we move this new Boolean member further down, directly in front of or after the other Boolean `_isTombstone` member?
This would make the struct not grow in memory usage, whereas placing the Boolean here will make it grow by 8 bytes.

arangod/Aql/QueryRegistry.cpp
```diff
@@ -182,146 +188,201 @@ void* QueryRegistry::openEngine(EngineId id, EngineType type) {
...
+  WRITE_LOCKER(writeLocker, _lock);
 
-  {
-    WRITE_LOCKER(writeLocker, _lock);
+  QueryInfoMapPerVocbase::iterator vocbaseQueriesIt;
+  QueryInfoMap::iterator queryMapIt;
+  if (!lookupQueryForFinalization(vocbase, id, errorCode, vocbaseQueriesIt,
+                                  queryMapIt)) {
+    return;
+  }
```
jsteemann, line 255: I think we should add assertions here that both `vocbaseQueriesIt` and `queryMapIt` are populated as expected and are valid iterators.

arangod/Aql/QueryRegistry.cpp
```diff
@@ -182,146 +188,201 @@ void* QueryRegistry::openEngine(EngineId id, EngineType type) {
...
+  QueryInfoMapPerVocbase::iterator vocbaseQueriesIt;
+  QueryInfoMap::iterator queryMapIt;
+  if (!lookupQueryForFinalization(vocbase, id, errorCode, vocbaseQueriesIt,
+                                  queryMapIt)) {
+    return std::shared_ptr<ClusterQuery>();
+  }
 
-    // remove query from the table of running queries
-    m->second.erase(q);
+  TRI_ASSERT(queryMapIt->second);
```
jsteemann, line 307: I think we should add assertions here that both `vocbaseQueriesIt` and `queryMapIt` are populated as expected and are valid iterators.

jsteemann (commented):
There are a few other tests that still reference the optimizer rule `parallelize-gather-writes` that this PR obsoletes.
We should adjust these tests as well, and remove any reference to the now-gone optimizer rule:
``` 
$ grep -r "parallelize-gather" tests/js/ enterprise/tests/
tests/js/server/aql/aql-optimizer-rule-optimize-cluster-single-document-operations.js:  var thisRuleEnabled  = { optimizer: { rules: [ "+all", "-move-filters-into-enumerate", "-parallelize-gather" ] } }; // we can only work after other rules
tests/js/server/aql/aql-optimizer-rule-optimize-cluster-single-document-operations.js:  var thisRuleDisabled = { optimizer: { rules: [ "+all", "-move-filters-into-enumerate", "-parallelize-gather", "-" + ruleName ] } };
tests/js/server/aql/aql-profiler-cluster.js:      const options = {optimizer: { rules: ["-parallelize-gather"] } };
tests/js/server/aql/aql-optimizer-rule-distribute-in-cluster.js:  const rulesAll         = { optimizer: { rules: [ "+all", "-reduce-extraction-to-projection", "-parallelize-gather" ] } };
tests/js/server/aql/aql-optimizer-rule-distribute-in-cluster.js:  const thisRuleDisabled = { optimizer: { rules: [ "+all", "-reduce-extraction-to-projection", "-parallelize-gather", "-" + ruleName ] } };
tests/js/server/aql/aql-skipping.js:      var queryOptions = {fullCount: true, optimizer: {"rules": ["-move-calculations-down", "-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {fullCount: true, optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {optimizer : { rules : [ "-sort-limit", "-parallelize-gather" ] }, fullCount: true};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {optimizer : { rules : [ "-parallelize-gather" ] }, fullCount: true};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {optimizer : { rules : [ "-parallelize-gather" ] }, fullCount: true};
tests/js/server/aql/aql-skipping.js:      var queryOptions = {optimizer : { rules : [ "-parallelize-gather" ] }, fullCount: true};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {fullCount: true, optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {fullCount: true, optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/server/aql/aql-skipping.js:      const queryOptions = {fullCount: true, optimizer: {"rules": ["-parallelize-gather"]}};
tests/js/client/shell/shell-aql-regression-cluster.js:        optimizer: {rules: ['-parallelize-gather']},
enterprise/tests/js/server/aql/aql-optimizer-rule-parallelize-gather-cluster-satellite-graphs.js:  const ruleName = "parallelize-gather";
```

CHANGELOG
mpoeter: Unfortunately the situation is a bit more complicated... Generally we cannot parallelize when we have traversals. The exception is when we have a disjoint SmartGraph where the traversal can be performed completely on the local server. However, there is another exception for that case where the traversal is parallelized and somewhere above the traversal node we have another scatter/distribute, then even for the disjoint SmartGraph we _cannot_ parallelize. Will try to communicate this a bit better in the CHANGELOG.

CHANGELOG
```diff
@@ -1,6 +1,17 @@
 devel
 -----
 
+* APM-283: Use parallel gather in almost all queries. The only case where we
+  cannot use parallel gather is when using traversals, although there are some
```
jsteemann, line 5: Instead of "traversals", shouldn't it be "traversals with parallelism" here?

CHANGELOG
```diff
@@ -1,6 +1,17 @@
...
+  exceptions for disjoint SmartGraphs where the traversal can run completely on
+  the local DB-server. All other queries should now be able to parallelize the
+  gather node. This can not only speed up queries quite significantly, but also
+  reduce the memory required by the coordinator for queries like
+  `FOR i IN @n INSERT {} INTO collection`. The serial gather node processes one
+  DB-server after the other, but since the new documents should be distributed
+  evenly over all shards, the coordinator has to buffer documents that need to
+  be sent to other servers.
```
jsteemann, line 13: Thinking about this sentence about the "serial gather node", I think this could be a bit difficult to understand.
Because we making things _parallel_, and not _serial_. 
IMO w should instead write something like "this also overcomes issues with the previous serial processing within gather nodes, which could lead to high memory usage on coordinators for buffering the documents for all shards, and to timeouts on some DB-Servers because of query parts being idle for too long".
Just a suggestion. The exact wording can probably still be improved.

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

arangod/Aql/OptimizerRules.cpp
```diff
@@ -8358,61 +8348,20 @@ void arangodb::aql::parallelizeGatherRule(Optimizer* opt,
...
-                                 .parallelizeGatherWrites();
-    GatherNode* gn = ExecutionNode::castTo<GatherNode*>(nodes[0]);
+  for (auto node : nodes) {
+    GatherNode* gn = ExecutionNode::castTo<GatherNode*>(node);
 
-    if (!gn->isInSubquery() && isParallelizable(gn, parallelizeWrites)) {
+    if (!gn->isInSubquery() && isParallelizable(gn)) {
       // find all graph nodes and make sure that they all are using satellite
-      nodes.clear();
+      containers::SmallVector<ExecutionNode*, 8> graphNodes;
```
MBkkt, line 8359: Create it with node and just call clear here?

arangod/Aql/QueryContext.cpp
```diff
@@ -131,6 +131,10 @@ void QueryContext::addDataSource(  // track DataSource
 
 aql::Ast* QueryContext::ast() { return _ast.get(); }
 
+std::lock_guard<std::mutex> QueryContext::acquireLockGuard() {
+  return std::lock_guard(_mutex);
```
MBkkt, line 135: NIT
```suggestion
  return {_mutex};
```

arangod/Aql/QueryContext.h
```diff
@@ -103,6 +103,10 @@ class QueryContext {
 
   aql::Ast* ast();
 
+  /// @brief Acquire a lock_guard on the mutex to serialize concurrent snippet
+  /// execution
+  std::lock_guard<std::mutex> acquireLockGuard();
```
MBkkt, line 108: NIT

if you move this to header can be written without mutex type, auto in return type, template deduction for return statement

arangod/Aql/QueryRegistry.cpp
```diff
@@ -182,146 +188,206 @@ void* QueryRegistry::openEngine(EngineId id, EngineType type) {
...
     }
-    LOG_TOPIC("5ecdc", TRACE, arangodb::Logger::AQL)
-        << "closing engine " << engineId
-        << ", query id: " << ei._queryInfo->_query->id()
-        << ", numOpen: " << ei._queryInfo->_numOpen;
-  } else {
-    LOG_TOPIC("ae981", TRACE, arangodb::Logger::AQL)
-        << "closing engine " << engineId << ", no query";
+
+    ei._isOpen = false;
```
MBkkt, line 215: Unnecessary assign.

MBkkt (approved):

arangod/Aql/QueryRegistry.cpp
mpoeter: Why do you think this is unnecessary? This way we mark the engine as unused, which is actually quite important.

mchacki: I think it actually is necessary.
We enter this code with _isOpen == true (otherwise we throw in the lines above)
And we need to leave it with _isOpen == false as we are giving back ownership to the Registry.

mchacki (approved):
LGTM

arangod/Aql/QueryRegistry.cpp
MBkkt: if (cond == false) throw
// Here cond is always true

PR #18299, opened 2023-03-10
Auto-flushing of WAL and column family data

jsteemann:
### Scope & Purpose

* Auto-flush RocksDB WAL files and in-memory column family data if the number of live WAL files exceeds a certain threshold. This is to make sure that WAL files are moved to the archive when there are a lot of live WAL files present (e.g. after a restart; in this case RocksDB does not count any previously existing WAL files when calculating the size of WAL files and comparing it `max_total_wal_size`. The feature can be configured via the following startup options:
  - `--rocksdb.auto-flush-min-live-wal-files`: minimum number of live WAL files that triggers an auto-flush. Defaults to `10`.
  - `--rocksdb.auto-flush-check-interval`: interval (in seconds) in which auto-flushes are executed. Defaults to `3600`. Note that an auto-flush is only executed if the number of live WAL files exceeds the configured threshold and the last auto-flush is longer ago than the configured auto-flush check interval. That way too frequent auto-flushes can be avoided.

This fixes a problem that existing WAL files are not counted by RocksDB after startup, and their size is not accounted for and compared against `max_total_wal_size`. This may prevent WAL files from being moved to the archive quickly.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,22 @@
 devel
 -----
 
+* Auto-flush RocksDB WAL files and in-memory column family data if the number
+  of live WAL files exceeds a certain threshold. This is to make sure that
+  WAL files are moved to the archive when there are a lot of live WAL files
+  present (e.g. after a restart; in this case RocksDB does not count any
+  previously existing WAL files when calculating the size of WAL files and 
+  comparing it `max_total_wal_size`.
+  The feature can be configured via the following startup options:
+  - `--rocksdb.auto-flush-min-live-wal-files`: minimum number of live WAL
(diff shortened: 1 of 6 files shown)


neunhoef (approved):

neunhoef:
LGTM

PR #18304, opened 2023-03-13
Auto-flushing of WAL and column family data

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/18299

* Auto-flush RocksDB WAL files and in-memory column family data if the number of live WAL files exceeds a certain threshold. This is to make sure that WAL files are moved to the archive when there are a lot of live WAL files present (e.g. after a restart; in this case RocksDB does not count any previously existing WAL files when calculating the size of WAL files and comparing it `max_total_wal_size`. The feature can be configured via the following startup options:
  - `--rocksdb.auto-flush-min-live-wal-files`: minimum number of live WAL files that triggers an auto-flush. Defaults to `10`.
  - `--rocksdb.auto-flush-check-interval`: interval (in seconds) in which auto-flushes are executed. Defaults to `3600`. Note that an auto-flush is only executed if the number of live WAL files exceeds the configured threshold and the last auto-flush is longer ago than the configured auto-flush check interval. That way too frequent auto-flushes can be avoided.

This fixes a problem that existing WAL files are not counted by RocksDB after startup, and their size is not accounted for and compared against `max_total_wal_size`. This may prevent WAL files from being moved to the archive quickly.

**Note**: this PR is currently based on `feature-3.9/rocksdb-wal-deletion-transparency`. This is necessary only temporarily. Once the other PR gets merged into 3.9, this PR can be rebased against 3.9 as well.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18300
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -238,6 +238,8 @@ RocksDBEngine::RocksDBEngine(application_features::ApplicationServer& server)
       _dbExisted(false),
       _runningRebuilds(0),
       _runningCompactions(0),
+      _autoFlushCheckInterval(60.0 * 60.0),
+      _autoFlushMinWalFiles(10),
       _metricsWalReleasedTickFlush(
           server.getFeature<arangodb::MetricsFeature>().add(
               rocksdb_wal_released_tick_flush{})),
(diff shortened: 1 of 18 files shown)


arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -338,6 +340,30 @@ void RocksDBEngine::shutdownRocksDBInstance() noexcept {
   _db = nullptr;
 }
 
+void RocksDBEngine::flushOpenFilesIfRequired() {
+  if (_metricsLiveWalFiles.load() < _autoFlushMinWalFiles) {
```
cpjulia, line 344: does the kind of memory order matter in this?

cpjulia (approved):
LGTM

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -238,6 +238,8 @@ RocksDBEngine::RocksDBEngine(application_features::ApplicationServer& server)
       _dbExisted(false),
       _runningRebuilds(0),
       _runningCompactions(0),
+      _autoFlushCheckInterval(60.0 * 60.0),
+      _autoFlushMinWalFiles(10),
```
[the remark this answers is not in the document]
jsteemann: Yes. We want to avoid too frequent flushes, and especially not flushes that do not provide any benefit.
If we have < 10 live journal files, these will not occupy a lot of disk storage.
The problem we are trying to solve with this PR is RocksDB having lots of live WAL files open from previous runs, which in total can consume GBs to TBs.

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: Not really. That's why I left it at the default. We could micro-optimize by using acquire and release here, or even relaxed. Then the only consequence would be that flushing may happen a bit earlier or later. But it is not important IMO.

mpoeter (approved):

neunhoef (approved):
LGTM

PR #18312, opened 2023-03-13
[GORDO-1549] Make worker context own aggregators

jvolmer:
This PR is similar to some things done in https://github.com/arangodb/arangodb/pull/18277:
-Adds a WorkerContext to each algorithm (before, only some algorithms had a worker context). The aggregates, that were previously owned by the Worker, are now owned by the WorkerContext, such that we don't have duplication of variables here.
-The algorithm can now also create a unique ptr to WorkerContext - opposed to a raw pointer - this is not used yet but will be used in our Worker actor

### DIFF
--- a/arangod/Pregel/Algorithm.h
+++ b/arangod/Pregel/Algorithm.h
@@ -59,7 +59,6 @@ struct IAlgorithm {
   [[nodiscard]] virtual auto masterContext(
       std::unique_ptr<AggregatorHandler> aggregators,
       arangodb::velocypack::Slice userParams) const -> MasterContext* = 0;
-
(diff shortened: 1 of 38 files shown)


markuspf (approved):
LGTM; I like the incremental approach, though I am slightly worried that we can now create raw pointers and unique pointers and its not clear why.

We have to make sure to remove the raw pointeryness soon.

jvolmer:
I deprecated the functions returning raw pointers.
I also added functions for message format and message combinator that return unique ptrs because I need them for my next PR.

markuspf:
Good idea using the deprecation flag! Thanks!

PR #18316, opened 2023-03-13
Fixing bumpTerm function. No it triggers a leader election.

maierlars:
### Scope & Purpose
Fix bumpTerm by triggering an actual leader election.


apetenchea (approved):
LGTM

PR #18328, opened 2023-03-14
[GORDO-1551] added aql result retrieval

hkernbach:
### Scope & Purpose

This PR adds the preparations for the AQL result retrieval from within the actor based architecture.

- [x] :pizza: New feature

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/GORDO-1551


arangod/Pregel/REST/RestPregelHandler.cpp
```diff
@@ -74,9 +76,42 @@ RestStatus RestPregelHandler::execute() {
...
+        }
+
+        // ActorID "0" being used here to initially spawn a new actor.
+        auto resultActorID = _pregel._actorRuntime->spawn<ResultActor>(
+            _vocbase.name(), std::make_unique<ResultState>(),
+            message::ResultMessages{message::ResultStart{}});
+        auto resultActorPID =
+            actor::ActorPID{.server = ServerState::instance()->getId(),
+                            .database = _vocbase.name(),
+                            .id = resultActorID};
```
jvolmer, line 105: For your intererest: I added ticket https://arangodb.atlassian.net/browse/GORDO-1576 to clean-up the actor, including that the spawn function should return an ActorPID instead of an ActorID, such that we don't have to create the PID here by ourselves.

arangod/Pregel/ResultActor.h
```diff
@@ -0,0 +1,118 @@
...
+
+  auto operator()(message::SaveResults start) -> std::unique_ptr<ResultState> {
+    this->state->results = {start.results};
+    this->state->finished = true;
+    return std::move(this->state);
+  }
+
+  auto operator()(message::AddResults start) -> std::unique_ptr<ResultState> {
+    if (this->state->results.fail()) {
+      this->state->finished = true;
```
jvolmer, line 58: This line is not actually required because finished will be set in line 64 when we get a message with an error inside.

arangod/Pregel/ResultActor.h
```diff
@@ -0,0 +1,118 @@
...
+      return std::move(this->state);
+    }
+
+    auto& builder = this->state->results.get().results;
+    {
+      VPackArrayBuilder arrayBuilder(&builder);
+      builder.add(VPackArrayIterator(start.results.get().results.slice()));
+    }
+    this->state->results =
+        ResultT<PregelResults>(PregelResults{std::move(builder)});
```
jvolmer, line 74: I'm not sure that this is correct: We now add a new array to our results-builder, but we want to add items to the already existing array in our results-builder. Not sure how to do that, but we could not mutate the results-builder but overwrite it like this:
```
    VPackBuilder newResultsBuilder;
    {
      VPackArrayBuilder ab(&newResultsBuilder);
      // Add existing results to new builder
      if (!results.get().results.isEmpty()) {
        newResultsBuilder.add(VPackArrayIterator(results.get().results.slice()));
      }
     // add new results from message to builder
      newResultsBuilder.add(VPackArrayIterator(start.results.get().results.slice()));
    }
    this->state->results = ResultT<PregelResults>(PregelResults(std::move(newAggregators)));
```

arangod/Pregel/ResultActor.h
```diff
@@ -0,0 +1,118 @@
...
+    return std::move(this->state);
+  }
+
+  auto operator()(message::SaveResults start) -> std::unique_ptr<ResultState> {
+    this->state->results = {start.results};
+    this->state->finished = true;
+    return std::move(this->state);
+  }
+
+  auto operator()(message::AddResults start) -> std::unique_ptr<ResultState> {
```
jvolmer, line 56: I just noticed that we are missing a check at the start of this function:
```
if (this->state->finished) {
  return std::move(this->state);
}
```
This is needed for the single server case: There, worker and conductor actors are on the same server and we only have one Result Actor. Now we get two messages in this actor: One SaveResults message and one AddResults message. With this check we make sure that the final results are just set when this actor gets the SaveResults message.

jvolmer (approved):

PR #18331, opened 2023-03-14
[GORDO-1549] Start computing state

jvolmer:
This PR partly implements the conductor computing state:
1. The conductor sends a `RunGlobalSuperStep` message to each worker
2. The worker gets this message but does not do anything yet with it (this will be done in the next PR).
3. The worker is able to receive `PregelMessage`s that are sent during the computation from other workers. The worker adds such a message either to its write cache (if the worker has the same global super step as the message) or to the `messagesForNextGss` buffer (if the message is for the next gss). (The case gss=0 is a bit weird: Before the computation starts, gss is initialized to zero, but the first gss is also 0, therefore some checking is done for this).

Some other changes I did during that:
- I changed the worker state to be closer to the old worker. Much can be improved here, but this makes right now the refactoring more easy, the worker state can the be refactored afterwards when the pregel runs successfully with actors.
- I slightly changed the execution state interface: The messages function does now return a map of messages that should be send to a specified actor. This makes the interface more flexible: Especially in this computing state, not all worker actors get the same message.


markuspf (approved):
LGTM

PR #18338, opened 2023-03-15
remove dead code/template parts for hash/skiplist in web UI

jsteemann:
### Scope & Purpose

Remove dead code/template parts for hash/skiplist in web UI.
It wasn't possible to create new hash/skiplist indexes in the web UI since at least ArangoDB 3.8.
The code for it was still present, but not used anymore.
This PR removes the code from the UI.
There should be no user-visible changes by this PR.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


js/apps/system/_admin/aardvark/APP/frontend/js/templates/indicesView.ejs
palashkaria, line 1: One suggestion/question: 

On [L74](https://github.com/arangodb/arangodb/pull/18338/files#diff-4fa746c4e5257ca715ecbab97f3c733d0117be9d032dbf75f3429f19984be508L74) there's a tooltip being created which says this: "Please note that for the RocksDB engine the index types hash, skiplist and persistent are identical, so that they are not offered seperately here"

Maybe we can remove this too?

jsteemann: IMO that tooltip still makes sense.
Because users that upgraded from older deployments (e.g. deployments that ran versions 3.6 or before), there may well be hash or skiplist indexes present. People may then wonder why they aren't able to create new hash or skiplist indexes in the UI.

palashkaria: Sounds good!

PR #18340, opened 2023-03-15
Mark mmaped-files commandline options obsolete

markuspf:
### Scope & Purpose

We do not support memory mapped files for Pregel anymore, so mark the command line options obsolete and remove the Pregel parameter in other places too.


arangod/Pregel/PregelFeature.cpp
jsteemann, line 1: It would be better to use method `addObsoleteOption(...)` here:
https://github.com/arangodb/arangodb/blob/devel/lib/ProgramOptions/ProgramOptions.h#L164

That has the advantage that no instance variable needs to be bound to the startup option, and the instance variables `_useMemoryMaps`, `_tempLocationType`, `_tempLocationCustomPath` can probably be removed altogether.

PR #18343, opened 2023-03-15
[CINFRA-687] Use larger int types in v8-replicated-logs

goedderz:
### Scope & Purpose

This mostly replaces uint32 types for LogIndexes, because these can be expected to grow large. Additionally, an assertion is introduced when errors can't be handled.

~Please review the latter part (in `lib/V8/v8-utils.cpp`) with extra care whether you agree that it's sensible.~ _Moved to https://github.com/arangodb/arangodb/pull/18364_


arangod/V8Server/v8-replicated-logs.cpp
```diff
@@ -87,6 +90,27 @@ static LogId UnwrapReplicatedLog(v8::Isolate* isolate,
       obj->GetInternalField(SLOT_CLASS)->Uint32Value(TRI_IGETC).ToChecked()};
 }
 
+namespace {
```
dothebart, line 93: I feel that this code rather belongs into `lib/V8/v8-conv.[cpp|h]`

dothebart: if its less bulky in the usage maybe have both?

goedderz: It's mostly different in that `TRI_ObjectToUInt64` returns `0` if it's not a number, while my code returns an error. I can achieve my main goals with both: Allow larger ints than 32bits, and don't introduce any security issues.

While I personally prefer errors rather than silent defaulting, its probably more JavaScript-y this way, and I don't think it's much of a problem here either way. So I chose to not add additional code but use the existing one.

maierlars (approved):

maierlars:
Formatting issues is unrelated to this PR and already in staging/replication-2.0.

PR #18349, opened 2023-03-15
[testing] use execution deadlines in chaos tests to avoid hard killing

dothebart:
### Scope & Purpose

better let arangosh itself run in deadlines, than being hard-killed from the outside


jsteemann (approved):
LGTM

PR #18351, opened 2023-03-16
Add FutureSharedLock

mpoeter:
### Scope & Purpose

This PR adds a SharedLock (RW lock) that does not block when trying to acquire a lock but instead returns a future that will be resolved once the lock has been acquired. The lock implementation itself is largely based on an MSC lock.

This lock will later be used to implement [APM-19](https://arangodb.atlassian.net/browse/APM-19).

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**

[APM-19]: https://arangodb.atlassian.net/browse/APM-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MBkkt (changes requested):
I don't see how it's related to ticket.

At the same time I don't see it as good design design.
If someone will use it will make at least 2 allocation per every lock.
(even in no contention case, also introduce possible contention in allocator)
It makes waiting async, but only if all our code will be written in callback style.

I think instead of future, it should return awaitable and in such case it's possible to make allocation only on caller coroutine.

Anyway I don't think such things should be merged silently and it should be discussed first.

MBkkt:
mcs btw

tests/Futures/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ target_link_libraries(arango_tests_futures
   PRIVATE
     arango_futures
     gtest
+    boost_boost
```
MBkkt, line 9: why?

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,244 @@
...
+            }
+          },
+          timeout);
+    }
+
+    Scheduler& _scheduler;
+    std::mutex _mutex;
+    // the list is ordered back to front, i.e., new items are added at the
+    // front; the last item (back) is the next to be scheduled
+    std::list<std::shared_ptr<Node>> _queue;
```
MBkkt, line 237: It can be intrusive list

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,244 @@
...
+           node = std::weak_ptr<Node>(*queueIterator),
+           queueIterator]() mutable {
+            auto me = self.lock();
+            if (me) {
+              std::lock_guard lock(me->_mutex);
+              auto nodePtr = node.lock();
+              if (nodePtr != nullptr && !nodePtr->scheduled) {
+                // if we could lock the node, the iterator must still be valid!
+                me->removeNode(queueIterator);
+                TRI_ASSERT(nodePtr.get() != nullptr) << (void*)nodePtr.get();
```
MBkkt, line 226:
```suggestion
                TRI_ASSERT(nodePtr.get() != nullptr);
```
it's null if assert failed

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,244 @@
...
+        std::chrono::milliseconds timeout) {
+      _scheduler.queueDelayed(
+          [self = this->weak_from_this(),
+           node = std::weak_ptr<Node>(*queueIterator),
+           queueIterator]() mutable {
+            auto me = self.lock();
+            if (me) {
+              std::lock_guard lock(me->_mutex);
+              auto nodePtr = node.lock();
+              if (nodePtr != nullptr && !nodePtr->scheduled) {
```
MBkkt, line 223: If we remove storing shared_ptr to scheduleNode callback
we can write something like this
```suggestion
              auto nodePtr = node.lock();
              if (nodePtr == nullptr) {
                return;  // most not expired will be here
              }
              std::lock_guard lock(me->_mutex);
              // we don't use use_count for sync, only as counter under mutex
              if (nodePtr.use_count() != 1) {
```

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,244 @@
...
+
+    void scheduleExclusive(Node* node) { scheduleNode(node); }
+
+    void scheduleShared(Node* leader) { scheduleNode(leader); }
+
+    void scheduleNode(std::shared_ptr<Node> node) {
+      TRI_ASSERT(node != nullptr);
+      node->scheduled = true;
+      _scheduler.queue(
+          [node, this]() { node->promise.setValue(LockGuard(this)); });
```
MBkkt, line 209: We can move promise from node to the functor.
In such case we don't need scheduled. because shared_ptr will die before

tests/Basics/FutureSharedLockTest.cpp
```diff
@@ -0,0 +1,656 @@
...
+
+  for (auto& t : threads) {
+    t.join();
+  }
+  EXPECT_EQ(0, numTasks.load() - lockTimeouts.load());
+  std::cout << "Found total " << totalFound.load() <<  //
+      "\nLock timeouts " << lockTimeouts.load() << std::endl;
+}
+
+}  // namespace
```
MBkkt, line 656:
```suggestion
}  // namespace

```

MBkkt (approved):

lib/Basics/FutureSharedLock.h
mpoeter: I am not sure I fully understand - the suggested change by itself would be incorrect. But this is the timeout handling, which is by definition not "popular", so it doesn't really make sense to optimize it.

MBkkt: This suggestion with correct with this one https://github.com/arangodb/arangodb/pull/18351#discussion_r1390941300

MBkkt: And it's kind of popular because it's for every trylock with timeout

MBkkt: > If we remove storing shared_ptr to scheduleNode callback

tests/Futures/CMakeLists.txt
mpoeter: Good question! I don't remember why I added this - but it can probably be removed now.

lib/Basics/FutureSharedLock.h
mpoeter: Yes, it could be an intrusive list, but for now it was simpler this way.

lib/Basics/FutureSharedLock.h
mpoeter: Ok, now I think I understand what you mean.

lib/Basics/FutureSharedLock.h
MBkkt: Ok

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,246 @@
...
+        // we are somewhere in the middle -> just remove ourselves
+        _queue.erase(queueIterator);
+      }
+    }
+
+    void scheduleExclusive(Node* node) { scheduleNode(node); }
+
+    void scheduleShared(Node* leader) { scheduleNode(leader); }
+
+    void scheduleNode(std::shared_ptr<Node> node) {
```
MBkkt, line 205: We don't need to copy shared_ptr here
```suggestion
    void scheduleNode(Node& node) {
```

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,246 @@
...
+      } else {
+        // we are somewhere in the middle -> just remove ourselves
+        _queue.erase(queueIterator);
+      }
+    }
+
+    void scheduleExclusive(Node* node) { scheduleNode(node); }
+
+    void scheduleShared(Node* leader) { scheduleNode(leader); }
+
```
MBkkt, line 204: It's unused
```suggestion
```

tests/Futures/CMakeLists.txt
MBkkt, line 9:
```suggestion
```

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,246 @@
...
+  void unlock() { _sharedState->unlock(); }
+
+ private:
+  struct Node {
+    explicit Node(bool exclusive) : exclusive(exclusive) {}
+
+    Promise<LockGuard> promise;
+    bool exclusive;
+    bool scheduled{false};
```
MBkkt, line 117: It's unused now
```suggestion
```

lib/Basics/FutureSharedLock.h
```diff
@@ -0,0 +1,246 @@
...
+        _queue.pop_back();
+
+        // if we are in shared mode, we can schedule all following shared
+        // nodes
+        if (!_exclusive) {
+          while (!_queue.empty() && !_queue.back()->exclusive) {
+            ++_lockCount;
+            scheduleNode(_queue.back());
+            _queue.pop_back();
+          }
```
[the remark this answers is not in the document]
mpoeter: What do you mean? In the "normal" RWLock implementation we always prefer writers. This lock is strictly fair, but we definitely do _not_ want to prefer readers, because this would be a fundamental change compared to the current RWLock implementation.

PR #18361, opened 2023-03-16
Log response bodies and HTTP headers on requests=trace.

neunhoef:
Currently, we lack information on outgoing response bodies and HTTP
headers (incoming and outgoing) in the case that we set the requests
log topic to "TRACE". This PR adds additional log messages for HTTP/1
and HTTP/2 and VST to this end.

### Scope & Purpose

- [*] :pizza: New feature
- [*] Better visibility

### Checklist

- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18362
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18363


dothebart:
hm, this has the potential to more than double the log volume of trace logs, right? could we i.e. send it to a separate facility so it can be enabled separately?

neunhoef:
@dothebart Yes, this potential is there, but there is also the command line option --log.request-parameters, which can be set to `false` to switch off these (and some other things). In any case, from my experience one either wants trace logging, but then it should be complete and not hide half of the stuff, or one does not need it at all, and then one can switch it to debug or whatever. I don't think this is now worthwhile to introduce another log topic for this.

arangod/GeneralServer/GeneralCommTask.cpp
jsteemann, line 1: `logRequestBody` and `logResponseBody` are almost identical, except a small part of the log message (`"request"` vs. `"response"`). The whole code to build the to-be-logged body can be put into a single funciton.

arangod/GeneralServer/VstCommTask.cpp
```diff
@@ -329,6 +329,17 @@ void VstCommTask<T>::processMessage(velocypack::Buffer<uint8_t> buffer,
           << VstRequest::translateMethod(req->requestType()) << "\",\""
           << url(req.get()) << "\"";
 
+      std::string_view body = req->rawPayload();
+      if (Logger::isEnabled(LogLevel::TRACE, Logger::REQUESTS) &&
+          Logger::logRequestParameters()) {
```
jsteemann, line 334: If `body` is not used outside of this `if`, let's move it inside.
```suggestion
      if (Logger::isEnabled(LogLevel::TRACE, Logger::REQUESTS) &&
          Logger::logRequestParameters()) {
        std::string_view body = req->rawPayload();        
```

arangod/GeneralServer/GeneralCommTask.cpp
jsteemann: I meant just the code to assemble the "body-as-a-string". This is identical in both functions right now.
The log messages are indeed slightly different, but then 15+ lines in front of it are identical in both.

jsteemann (approved):
LGTM

PR #18368, opened 2023-03-17
FE-142: inverted index UI

palashkaria:
### Scope & Purpose

Adds the inverted index addition screen

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manual tests

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-142


js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/IndexFormField.tsx
```diff
@@ -1,66 +1,102 @@
...
+          </FormLabel>
           <InputControl
             isDisabled={field.isDisabled}
             isRequired={field.isRequired}
             name={field.name}
-            inputProps={{ autoFocus: index === 0 }}
+            inputProps={{ autoFocus }}
           />
-          {field.tooltip && <InfoTooltip label={field.tooltip} />}
+          {field.tooltip ? <InfoTooltip label={field.tooltip} /> : <Box></Box>}
```
nikhil-varma, line 98: This is nice. I think we can make this a practice. Recently started noticing issues with returning `null` 😛

js/apps/system/_admin/aardvark/APP/react/src/views/views/forms/inputs/LinkPropertiesInput.tsx
```diff
@@ -31,140 +36,123 @@ const LinkPropertiesInput = ({
...
-              <FieldsDropdown
-                basePath={basePath}
-                isDisabled={!!disabled}
-                formState={formState}
+        <Stack>
+          <HStack alignItems="center">
+            <FormLabel margin="0" position="relative" top="1">
+              Fields
+            </FormLabel>
+            <InfoTooltip label="Add field names that you want to be indexed here. Click a field name to set field details" />
```
nikhil-varma, line 67: Does this make sense contextually ?
```suggestion
            <InfoTooltip label="Add field names that you want to be indexed here. Click  on a field name to set field details" />
```

js/apps/system/_admin/aardvark/APP/react/src/components/split/Split.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ChildrenPattern,
+  ComponentPattern,
+  RenderPattern,
+  SplitGridProps
+} from "react-split-grid";
+
+export const Split = (
+  props: SplitGridProps &
+    (ChildrenPattern | RenderPattern | ComponentPattern) & {
+      localStorageKey: string;
```
nikhil-varma, line 12: How about we rename this to `accesskey` ?

js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/invertedIndex/useCreateInvertedIndex.ts
```diff
@@ -0,0 +1,320 @@
...
+type PrimarySortFieldType = {
+  field: string;
+  direction: "asc" | "desc";
+};
+
+export type InvertedIndexFieldType = {
+  name: string;
+  analyzer?: string;
+  features?: AnalyzerFeatures[];
+  searchField?: boolean;
```
nikhil-varma, line 83:
```suggestion
  shouldSearchField?: boolean;
```

This kind of feels like the string of the search field currently.

nikhil-varma (changes requested):

js/apps/system/_admin/aardvark/APP/react/src/components/split/Split.tsx
palashkaria: used `storageKey` because we have this eslint rule, `accessKey` is keyword https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-access-key.md

js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/invertedIndex/useCreateInvertedIndex.ts
palashkaria: I agree, `searchField` sounds like it should be a `string`, but since this is the key on backend, we could change on just FE, but will need to convert back. I think we should leave it like this unless backend also wants to fix it. What do you think?

Also, cc: @Dronplane for feedback on the field name

nikhil-varma (approved):

KVS85:
- [x] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/27857/

PR #18370, opened 2023-03-17
[testing] process monitoring thread

dothebart:
### Scope & Purpose

if (parts) of the SUT crashes, we want to terminate tests in a timely manner. Hence we need to constantly monitor processes, and in case they disappear trigger the arangosh-self-shotgun.

- export `instanceManager` to tests, so they can compfortably halt / resume arangods
- move looped code from the `instanceManager` => `instance`
- adjust tests to use this
- add / remove processes to the monitoring thread if we intend to have them in running.
- several tests (i.e. recovery) will crash servers or manipulate instances without the instance manager, disable the montior for these.
- Improve crash / kill messages in several places

- [x] :pizza: New feature


js/client/bootstrap/modules/internal.js
```diff
@@ -64,6 +64,33 @@
...
+  }
+
+  // //////////////////////////////////////////////////////////////////////////////
+  // / @brief unmonitor a sub process
+  // //////////////////////////////////////////////////////////////////////////////
+
+  if (global.SYS_REMOVE_FROM_PID_MONITORING) {
+    exports.removePidFromMonitor = global.SYS_REMOVE_FROM_PID_MONITORING;
+    delete global.SYS_REMOVE_FROM_PID_MONITORING;
+  }
```
jsteemann, line 83: Just wondering why it is `SYS_ADD_TO_PID_MONITORING` and not `SYS_ADD_PID_TO_MONITORING`, and why it is `SYS_REMOVE_FROM_PID_MONITORING` and not `SYS_REMOVE_PID_FROM_MONITORING`?

lib/V8/processMonitor.cpp
jsteemann, line 1: Filename for this file should be `ProcessMonitor.cpp` or `v8-processmonitor.cpp`.
Same for the header file.

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <fcntl.h>
+
+#ifdef TRI_HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+
+#include "Basics/Mutex.h"
+#include "Basics/MutexLocker.h"
+#include "Basics/DeadlockDetector.h"
+#include "Basics/Thread.h"
```
jsteemann, line 80: Both some unnecessary here.

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <algorithm>
+#include <chrono>
```
jsteemann, line 31: Are **all** of these includes actually needed here?

js/client/modules/@arangodb/testutils/testing.js
```diff
@@ -122,6 +122,7 @@ let optionsDocumentation = [
   '   - `disableClusterMonitor`: if set to false, an arangosh is started that will send',
   '                              keepalive requests to all cluster instances, and report on error',
   '   - `disableMonitor`: if set to true on windows, procdump will not be attached.',
+  '   - `enableAliveMonitor`: checks whether spawned processe disappear in midnight and aborts.',
```
jsteemann, line 125: Why "in midnight"?

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+}  // namespace
+
+namespace arangodb {
+
+/// @brief if monitored & exited we keep them here:
+std::map<TRI_pid_t, ExternalProcessStatus> ExitedExternalProcessStatus;
+
+std::vector<ExternalId> monitoredProcesses;
+
+void addMonitorPID(ExternalId& pid) {
```
jsteemann, line 100: Why is `ExternalId` passed by _mutable_ reference here?
I would understand if it would be const reference, or by rvalue reference, but not like this.

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+  if (it == ExitedExternalProcessStatus.end()) {
+    return std::nullopt;
+  }
+  return std::optional<ExternalProcessStatus>{it->second};
+}
+
+class ProcessMonitorThread : public arangodb::Thread {
+ public:
+  ProcessMonitorThread(application_features::ApplicationServer& server)
+      : Thread(server, "ProcessMonitorThread") {}
```
jsteemann, line 128: There is only limited space for thread names. Normally suffixes such as "Thread" are omitted from the names, as they are not adding more insight or value.
```suggestion
      : Thread(server, "ProcessMonitor") {}
```

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+          triggerV8DeadlineNow(false);
+        }
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+  }
+};
+
+namespace {
+ProcessMonitorThread* mt = nullptr;
```
jsteemann, line 156: Please change to a unique_ptr, and move to the anonymous namespace section at the top of the file.

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+void launchMonitorThread(application_features::ApplicationServer& server) {
+  if (mt == nullptr) {
+    mt = new ProcessMonitorThread(server);
```
jsteemann, line 161: Please use `std::make_unique`.
Even better: instead of having these global functions and another global variable `mt`, why not move the entire management of the ProcessMonitor thread into its own ApplicationFeature?

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+    mt->start();
```
jsteemann, line 162: start() can fail. return value is not checked.

lib/V8/processMonitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+class ProcessMonitorThread : public arangodb::Thread {
+ public:
+  ProcessMonitorThread(application_features::ApplicationServer& server)
+      : Thread(server, "ProcessMonitorThread") {}
+  ~ProcessMonitorThread() { shutdown(); }
+
+ protected:
+  void run() override {
+    while (!isStopping()) {
+      std::vector<ExternalId> mp;
```
jsteemann, line 134: If any exception happens inside `run()`, the process monitoring will stop. Which is fine, but just saying.

lib/V8/v8-deadline.cpp
```diff
@@ -138,6 +147,13 @@ uint32_t correctTimeoutToExecutionDeadline(uint32_t timeoutMS) {
   return delta;
 }
 
+void triggerV8DeadlineNow(bool fromSignal) {
+  // Set the deadline to expired:
+  errorState = fromSignal ? errorExternalDeadline : errorProcessMonitor;
```
jsteemann, line 152: not thread-safe

js/client/bootstrap/modules/internal.js
dothebart: for me it it was `PidMonitoring` to which an entry was enlisted or removed; Grammer in variable names is non-existent ;)

js/client/modules/@arangodb/testutils/testing.js
dothebart: `in the middle of executed tests` better?

lib/V8/v8-deadline.cpp
```diff
@@ -166,12 +180,81 @@ static bool SignalHandler(DWORD eventType) {
...
+  triggerV8DeadlineNow(true);
 }
 
 #endif
 
+////////////////////////////////////////////////////////////////////////////////
+/// @brief enables monitoring for an external PID
+////////////////////////////////////////////////////////////////////////////////
+
+static void JS_addPidToMonitor(
```
jsteemann, line 192:
```suggestion
static void JS_AddPidToMonitor(
```

lib/V8/v8-deadline.cpp
```diff
@@ -192,7 +275,24 @@ static void JS_RegisterExecutionDeadlineInterruptHandler(
   TRI_V8_TRY_CATCH_END
 }
 
+static void JS_GetDeadlineString(
+    v8::FunctionCallbackInfo<v8::Value> const& args) {
+  TRI_V8_TRY_CATCH_BEGIN(isolate);
+  v8::HandleScope scope(isolate);
+  TRI_V8_RETURN_STRING(errorState);
```
jsteemann, line 282: Not thread-safe?

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "ApplicationFeatures/ApplicationServer.h"
+
+#include "V8/v8-deadline.h"
+#include "V8/V8SecurityFeature.h"
+
+#include "ProcessMonitoringFeature.h"
+
+using namespace arangodb::basics;
+using namespace arangodb::options;
+// using namespace arangodb::rest;
```
jsteemann, line 43:
```suggestion
```

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,147 @@
...
+            MUTEX_LOCKER(
+                mutexLocker,
+                _processMonitorFeature->_MonitoredExternalProcessesLock);
+            _processMonitorFeature->_ExitedExternalProcessStatus[pid._pid] =
+                status;
+          }
+          triggerV8DeadlineNow(false);
+        }
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
jsteemann, line 129: Wouldn't it be more efficient to let this sleep using a condition variable, and only signal it to wake up once something actually needs to be done?
Otherwise it will wake up 100 times a second, look a mutex, copy data etc.

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -0,0 +1,104 @@
...
+///
+/// @author Wilfried Goesgens
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#include <map>
+#include <optional>
+#include <vector>
+#include <memory>
```
jsteemann, line 29:
```suggestion
#include <memory>
#include <optional>
#include <vector>
```

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "ProcessMonitoringFeature.h"
+
+using namespace arangodb::basics;
+using namespace arangodb::options;
+// using namespace arangodb::rest;
+using namespace arangodb;
+
+namespace arangodb {
+
+void ProcessMonitoringFeature::addMonitorPID(ExternalId& pid) {
```
jsteemann, line 48: Why is `pid` passed by mutable reference here? It should either be by value, or by const reference.

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -0,0 +1,104 @@
...
+
+ public:
+  ProcessMonitorThread(application_features::ApplicationServer& server,
+                       ProcessMonitoringFeature* processMonitorFeature)
+      : Thread(server, "ProcessMonitor"),
+        _processMonitorFeature(processMonitorFeature) {}
+  ~ProcessMonitorThread() { shutdown(); }
+
+ protected:
+  ProcessMonitoringFeature* _processMonitorFeature;
```
jsteemann, line 49: We should use a reference here. That indicates to the reader of the code that there will always be a ProcessMonitorFeature around, and it cannot be a nullptr.
```suggestion
  ProcessMonitoringFeature& _processMonitorFeature;
```

client-tools/Shell/v8-deadline.cpp
```diff
@@ -83,8 +91,7 @@ bool isExecutionDeadlineReached(v8::Isolate* isolate) {
     return false;
   }
 
-  TRI_CreateErrorObject(isolate, TRI_ERROR_DISABLED,
-                        "Execution deadline reached!", true);
+  TRI_CreateErrorObject(isolate, TRI_ERROR_DISABLED, errorState, true);
```
jsteemann, line 94: Potentially not thread-safe: reading `errorState` without mutex here.

lib/V8/v8-deadline.h
```diff
@@ -25,6 +25,16 @@
 
 #include <chrono>
 #include <v8.h>
+#include <optional>
+#include "Basics/process-utils.h"
+
+#include "ApplicationFeatures/ApplicationFeature.h"
```
jsteemann, line 31: Is this include really needed in the header file?

jsteemann (commented):
Reviewed (only) the C++ part and am happy with it.

js/client/modules/@arangodb/testsuites/resilience.js
```diff
@@ -89,15 +91,15 @@ var _resilience = function(path) {
...
-const resilienceAnalyzers = (new _resilience('resilience_analyzers')).func;
+const resilienceMove = (new _resilience('resilience_move', false)).func;
+const resilienceMoveView = (new _resilience('resilience_move_view', true)).func;
+const resilienceRepair = (new _resilience('resilience_repair', true)).func;
+const resilienceFailover = (new _resilience('resilience_failover', false)).func;
+const resilienceFailoverFailure = (new _resilience('resilience_failover_failure', false)).func;
+const resilienceFailoverView = (new _resilience('resilience_failover_view', false)).func;
+const resilienceTransactions = (new _resilience('resilience_transactions', false)).func;
+const resilienceSharddist = (new _resilience('resilience_sharddist', true)).func;
+const resilienceAnalyzers = (new _resilience('resilience_analyzers', true)).func;
```
alexbakharew, line 102: Why `AliveMonitor` is disabled for some of them?

js/client/modules/@arangodb/testsuites/recovery_cluster.js
```diff
@@ -249,34 +249,36 @@ function recovery (options) {
       status: false
     };
   }
-  options.cluster = true;
+  let localOptions = _.clone(options);
+  localOptions.cluster = true;
+  localOptions.enableAliveMonitor = false;
```
alexbakharew, line 254: We don't wan't to track pids here, do we?

dothebart: yes. the test execution crashes servers from within server testpoints, and we can't properly track / disable that.

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,151 @@
...
+    if (!_monitorThread->start()) {
+      LOG_TOPIC("33333", FATAL, Logger::SYSCALL)
+          << "failed to launch monitoring background thread";
+      FATAL_ERROR_EXIT();
+    }
+  }
+}
+void ProcessMonitoringFeature::beginShutdown() {
+  if (_enabled) {
+    _monitorThread->shutdown();
```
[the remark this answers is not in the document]
dothebart: hm, this was heavily inspired by `CacheManagerFeature` - which actually even calls it one more time, which I will add here as well (`stop()` is missing)

js/client/modules/@arangodb/testsuites/resilience.js
dothebart: These tests SIG_STOP arangods from outside of the `instanceManager`, we can't handle that.

alexbakharew (approved):
LGTM

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "ProcessMonitoringFeature.h"
+
+using namespace arangodb::basics;
+using namespace arangodb::options;
+using namespace arangodb;
+
+namespace arangodb {
+
+void ProcessMonitoringFeature::addMonitorPID(ExternalId const& pid) {
+  MUTEX_LOCKER(mutexLocker, _MonitoredExternalProcessesLock);
```
MBkkt, line 48: _[A-Z] is not allowed for standard C++.

It's reserved for standard library.

Also it's not correspond to our codestyle

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "ApplicationFeatures/ApplicationServer.h"
+
+#include "V8/v8-deadline.h"
+#include "V8/V8SecurityFeature.h"
+
+#include "ProcessMonitoringFeature.h"
+
+using namespace arangodb::basics;
+using namespace arangodb::options;
+using namespace arangodb;
```
MBkkt, line 43: Why?

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -0,0 +1,101 @@
...
+  ////////////////////////////////////////////////////////////////////////////////
+
+  std::optional<ExternalProcessStatus> getHistoricStatus(TRI_pid_t pid);
+
+  ////////////////////////////////////////////////////////////////////////////////
+  /// @brief external process stati of processes exited while being monitored
+  ////////////////////////////////////////////////////////////////////////////////
+
+ protected:
+  arangodb::Mutex _MonitoredExternalProcessesLock;
```
MBkkt, line 88: Please use std::mutex

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,158 @@
...
+    for (auto it = _monitoredProcesses.begin(); it != _monitoredProcesses.end();
+         ++it) {
+      if (it->_pid == pid._pid) {
+        _monitoredProcesses.erase(it);
+        break;
+      }
+    }
+  }
+  // make sure its really not monitored anymore once we exit:
+  std::this_thread::sleep_for(std::chrono::milliseconds(110));
```
MBkkt, line 64: I don't understand this two sleeps. Please explain it.

Also we don't have realtime os.
So here thread can wakeup after 110ms and in other place thread can wakeup after 120ms.

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+void ProcessMonitorThread::run() {  // override
+  std::vector<ExternalId> mp;
+  mp.reserve(10);
+  while (!isStopping()) {
+    try {
+      {
+        MUTEX_LOCKER(mutexLocker,
+                     _processMonitorFeature._MonitoredExternalProcessesLock);
+        mp = _processMonitorFeature._monitoredProcesses;
```
MBkkt, line 121: you probably want swap here and clear after for.
Otherwise reserve is completely useless (we just alloc and dealloc this memory)

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -0,0 +1,101 @@
...
+#include <vector>
+
+#include "Basics/Thread.h"
+#include "Basics/process-utils.h"
+#include "Shell/arangosh.h"
+
+namespace arangodb {
+class ProcessMonitoringFeature;
+
+class ProcessMonitorThread : public arangodb::Thread {
```
MBkkt, line 38: It's not final?

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -0,0 +1,101 @@
...
+  friend ProcessMonitoringFeature;
+
+ public:
+  ProcessMonitorThread(application_features::ApplicationServer& server,
+                       ProcessMonitoringFeature& processMonitorFeature)
+      : Thread(server, "ProcessMonitor"),
+        _processMonitorFeature(processMonitorFeature) {}
+  ~ProcessMonitorThread() { shutdown(); }
+
+ protected:
```
MBkkt, line 48: Do we have some derived class?

MBkkt (commented):
I think this PR needs additional work and disscussion

PR #18380, opened 2023-03-20
[GORDO-1549] Execute one gss in worker

jvolmer:
This PR adds code in the worker actor to execute one gss - so everything between getting the message RunGlobalSuperStep from the conductor and sending GlobalSuperStepFinished message back to the conductor.

Additionally, some preparation is needed for this to work: The conductor CreateWorkers execution state now keeps track of which worker actors are responsible for which shards, gives this information to the Loading execution state which sends this information to all workers. This information is needed inside each worker because workers send messages around to other workers and in the actor framework they have to know to which actor they should send these messages to.

This PR also adds two actor-specific OutCaches: ArrayOutActorCache (for sending arrays of messages) and CombiningOutActorCache (for combining messages before sending them). These two Caches now send messages to specific actors instead of sending plain REST messages as the old OutCaches. 
When we have more time, we need to refactor the In- and OutCaches, they are very messy.

### DIFF
--- a/arangod/Pregel/Conductor/ExecutionStates/CollectionLookup.h
+++ b/arangod/Pregel/Conductor/ExecutionStates/CollectionLookup.h
@@ -24,8 +24,10 @@
 
 #include <string>
 #include <map>
+#include <unordered_map>
+#include <vector>
 
-#include "Cluster/ClusterTypes.h"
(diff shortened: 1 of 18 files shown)


markuspf (approved):
LGTM

PR #18381, opened 2023-03-20
Use storage manager in LogLeader.

maierlars:
### Scope & Purpose
The purpose of this PR is to reduce the amount of memory used on the leader. This is done by serving all iterator from disk instead of from RAM. 

The steps are:
1. use the storage manager in the log leader and local follower.
2. server iterators in the log leader from the storage manager.
3. remove the inMemoryLog from the log leader.


Simran-B (approved):
Metrics LGTM

PR #18382, opened 2023-03-20
Speed up tests

mpoeter:
### Scope & Purpose

Speed up js test execution

### Checklist

- [x] Tests
  - [x] **integration tests**


dothebart (approved):

jsteemann (approved):
I generally agree to the changes in this PR and think that they are useful to reduce the test durations.
I am not 100% sure however that due to the reduced sleep times, we may now run into _new_ trouble in very slow environments. I guess we simply have to see if some of the tests modified by this PR now start failing in some runs on very slow/overloaded CI servers.

jsteemann (approved):
LGTM

PR #18392, opened 2023-03-21
Chore: Add allProtocols flag and group some tests that run for all protocols

mpoeter:
### Scope & Purpose


scripts/test/test_launch_controller.py
```diff
@@ -67,10 +63,18 @@ def main():
         if args.cluster:
             extra_args += ["--cluster", "true", "--dumpAgencyOnError", "true"]
         extra_args += ["--testBuckets", args.testBuckets]
-        runner = TestingRunner(SiteConfig(Path(args.definitions).resolve()))
-        runner.scenarios.append(
-            TestConfig(runner.cfg, name, suite, [*extra_args], 1, 1, [])
-        )
+        config = SiteConfig(Path(args.definitions).resolve())
+        config.serialize_execution = True
```
dothebart, line 67: I think for better portability this flag should be set from the CLI as parameter.

dothebart (changes requested):
we need to have a CLI switch to python setting the serial execution so this can be used in oskar (optionally) as well, and to keep the code more similar. 
We also need to disable multi-execution on Mac since we need to cut down on testcoverage on these underpowered machines.

scripts/test/test_launch_controller.py
```diff
@@ -67,10 +63,18 @@ def main():
...
+        runner = TestingRunner(config)
+        if args.allProtocols:
+            for proto in ["http", "http2", "vst"]:
```
jsteemann, line 70: VST protocol is not supported anymore.
```suggestion
            for proto in ["http", "http2"]:
```

dothebart (approved):

PR #18397, opened 2023-03-21
[GORDO-1559] Add REST pregel option to run pregel with actors

jvolmer:
With this PR, pregel runs either as it was before without actors (the default) or with actors. In a REST call you can add `"actors":true` to run with actors, e.g.:
```
curl -X POST http://localhost:8529/_api/control_pregel -u root: -d '{"algorithm" : "wcc", "graphName" : "generatedGraph", "actors":true}'
```


markuspf (approved):
LGTM

PR #18414, opened 2023-03-23
[testing] Bug fix/fix shutdown killing

dothebart:
### Scope & Purpose

- process monitoring should move PID to the list of historic PIDs as fast as possible
- emergency child-killing needs to deregister PIDs as well.
- if we overlap in time, retry to get the real historic status for proper management


js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -793,7 +793,12 @@ class instance {
     if (this.pid === null) {
       return false;
     }
-    const res = statusExternal(this.pid, false);
+    let res = statusExternal(this.pid, false);
+    if (res.status === 'NOT-FOUND') {
+      print(`${Date()} ${this.name}: PID ${this.pid} missing on our list, retry?`);
+      time.sleep(1);
+      res = statusExternal(this.pid, false);
```
jsteemann, line 800: What is the rationale for sleeping 1s here, and then retrying? Will the process appear magically?
I am pretty sure there is some motivation for this adjustment, but from reading the code alone, it is not clear.
Please add a code comment for why you think this is a good idea.

dothebart: this is about the probable time of the process being removed from the list of PIDs we monitor in proces-utils.cpp, and re-appearing in the list of historic PIDs that exited in the monitoring thread. 
With the fixed non-sleeping between the removing from the monitored list and adding to the historic list the chance that the .js steps into this gap should be already way less likely to happen. But ies, 0.2s /should/ be enough.

client-tools/Shell/ProcessMonitoringFeature.h
```diff
@@ -73,6 +73,7 @@ class ProcessMonitoringFeature final : public ArangoshFeature {
   ////////////////////////////////////////////////////////////////////////////////
 
   void removeMonitorPID(ExternalId const& pid);
+  void removeMonitorPIDNoLock(ExternalId const& pid);
```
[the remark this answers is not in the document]
dothebart: made it protected, private doesn't work here.

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -49,16 +49,33 @@ void ProcessMonitoringFeature::addMonitorPID(ExternalId const& pid) {
...
+}
+
+void ProcessMonitoringFeature::moveMonitoringPIDToAttic(
+    ExternalId const& pid, ExternalProcessStatus const& exitStatus) {
+  MUTEX_LOCKER(mutexLocker, _MonitoredExternalProcessesLock);
+  removeMonitorPIDNoLock(pid);
+  _ExitedExternalProcessStatus[pid._pid] = exitStatus;
+}
+
+std::vector<ExternalId>& ProcessMonitoringFeature::getMonitoringVector() {
```
jsteemann, line 69: This is unsafe. We must return a _copy_ of the vector, not a reference.
If we return a reference, the caller can modify or peek into the vector without the mutex!

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -49,16 +49,33 @@ void ProcessMonitoringFeature::addMonitorPID(ExternalId const& pid) {
...
+  MUTEX_LOCKER(mutexLocker, _MonitoredExternalProcessesLock);
+  _monitoredProcessesThreadCopy = _monitoredProcesses;
```
jsteemann, line 71: Why is the copy made in an instance variable?
Why don't we just return a copy of the instance variable, i.e. `return _monitoredProcesses`?

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -111,29 +127,15 @@ void ProcessMonitoringFeature::stop() {
...
           // Its dead and gone - good
-          _processMonitorFeature.removeMonitorPID(pid);
-          {
-            MUTEX_LOCKER(
-                mutexLocker,
-                _processMonitorFeature._MonitoredExternalProcessesLock);
-            _processMonitorFeature._ExitedExternalProcessStatus[pid._pid] =
-                status;
-          }
+          { _processMonitorFeature.moveMonitoringPIDToAttic(pid, status); }
```
jsteemann, line 138:
```suggestion
          _processMonitorFeature.moveMonitoringPIDToAttic(pid, status);
```

jsteemann (approved):
LGTM

client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -49,16 +49,32 @@ void ProcessMonitoringFeature::addMonitorPID(ExternalId const& pid) {
   _monitoredProcesses.push_back(pid);
 }
 
+void ProcessMonitoringFeature::removeMonitorPIDNoLock(ExternalId const& pid) {
+  for (auto it = _monitoredProcesses.begin(); it != _monitoredProcesses.end();
+       ++it) {
+    if (it->_pid == pid._pid) {
+      _monitoredProcesses.erase(it);
```
MBkkt, line 56: Please use swap with last and pop_back instead of erase from random place of of vector

dothebart:
fuck git. https://github.com/arangodb/arangodb/pull/18443

PR #18419, opened 2023-03-23
[CINFRA-715] Supervision Deadlock Snapshot Missing

maierlars:
### Scope & Purpose

In some cases the supervision excluded a server from the quorum. If not all snapshots are available, this could create a deadlock.

Therefore we only consider excluding a server from the quorum, when after the exclusion there is still enough participants with a snapshot.
Additionally, if the flag `allowedInQuorum=false` is already in place, we undo it, _unless_ it was specified by the user in Target.


apetenchea (approved):
LGTM

PR #18420, opened 2023-03-23
SEARCH-300 (BTS-789) remove database directory for arangosearch

MBkkt:
### Scope & Purpose

SEARCH-300 remove database directory for arangosearch
and a lot of refactoring

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/FlushFeature.cpp
```diff
@@ -94,26 +94,18 @@ std::tuple<size_t, size_t, TRI_voc_tick_t> FlushFeature::releaseUnusedTicks() {
...
-    for (auto itr = _flushSubscriptions.begin();
-         itr != _flushSubscriptions.end();) {
-      auto entry = itr->lock();
-
-      if (!entry) {
-        // remove stale
-        itr = _flushSubscriptions.erase(itr);
-        ++stale;
-      } else {
-        LOG_TOPIC("5a4fb", TRACE, arangodb::Logger::FLUSH)
```
Dronplane, line 108: Please restore this logging. It is very useful for Debugging

arangod/RestServer/FlushFeature.cpp
```diff
@@ -94,26 +94,18 @@ std::tuple<size_t, size_t, TRI_voc_tick_t> FlushFeature::releaseUnusedTicks() {
...
-            << "found flush subscription: " << entry->name() << ", tick "
-            << entry->tick();
-
```
jsteemann, line 111: The PR removes this log message, but it shouldn't.
This log message helped us diagnose and prove that there are problems with stale links in the first place!

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1048,13 +1048,14 @@ void IResearchDataStore::shutdownDataStore() noexcept {
...
-  if (!irs::file_utils::exists_directory(exists, _dataStore._path.c_str()) ||
-      (exists && !irs::file_utils::remove(_dataStore._path.c_str()))) {
-    return {TRI_ERROR_INTERNAL,
-            absl::StrCat("failed to remove ArangoSearch index '",
-                         index().id().id(), "'")};
+  if (error) {
+    return {
+        TRI_ERROR_INTERNAL,
+        absl::StrCat("failed to remove ArangoSearch index '", index().id().id(),
+                     "' with error code '", error.message(), "'")};
```
jsteemann, line 1058: Technically, we are showing an error _message_ here and no error _code_. So we can as well simplify the error message construction.
```suggestion
                     "' with error code ", error.message())};
```

arangod/V8Server/V8DealerFeature.h
```diff
@@ -70,13 +70,24 @@ class V8DealerFeature final : public ArangodFeature {
...
+  void collectOptions(std::shared_ptr<options::ProgramOptions>) final;
+  void validateOptions(std::shared_ptr<options::ProgramOptions>) final;
+  void prepare() final;
+  void start() final;
+  void unprepare() final;
+
+  void verifyAppPaths();
+  ErrorCode createDatabase(std::string_view name, std::string_view id,
+                           bool removeExisting);
+  void removeDatabase(TRI_vocbase_t& database);
```
Dronplane, line 82:
```suggestion
  void cleanupDatabase(TRI_vocbase_t& database);
```

Dronplane (changes requested):

arangod/RestServer/FlushFeature.cpp
MBkkt: Yes, it's not intentional, will be fixed

arangod/VocBase/vocbase.h
```diff
@@ -308,9 +304,12 @@ struct TRI_vocbase_t {
   /// @brief drops a view
   arangodb::Result dropView(arangodb::DataSourceId cid, bool allowDropSystem);
 
-  /// @brief returns all known collections with their parameters
-  /// and optionally indexes
-  /// the result is sorted by type and name (vertices before edges)
+  /// @brief returns all known (document) collections with their parameters
```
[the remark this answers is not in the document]
MBkkt: It's copied from .cpp comment. Ok, will fix

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -146,82 +130,54 @@ void DatabaseManagerThread::run() {
...
-                      << "removing app directory '" << path << "' of database '"
-                      << database->name() << "'";
-
-                  TRI_RemoveDirectory(path.c_str());
-                }
-              }
-            }
-          }
+        TRI_ASSERT(!database->isSystem());
+        if (ServerState::instance()->isCoordinator()) {
```
jsteemann, line 134: The only thing the thread will do on coordinators is to physically remove the database objects that are in `_droppedDatabases` and that are not marked as "in use" anymore.

arangod/IResearch/IResearchDataStore.cpp
MBkkt: error message contains code and category. It's std:;error_code

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -146,82 +130,54 @@ void DatabaseManagerThread::run() {
...
-            TRI_ASSERT(newInstance == nullptr ||
-                       newInstance->id() != database->id());
-
-            if (newInstance == nullptr) {
+        search.removeDatabase(*database);
+
+        auto* queryRegistry = QueryRegistryFeature::registry();
+        if (dealer.isEnabled() || queryRegistry != nullptr) {
+          // TODO(MBkkt) Why shouldn't we remove database data
+          //  if exists database with same name?
```
jsteemann, line 150: Because this thread executes the deletion at some point _after_ the user issued a `db._dropDatabase()` call.
Consider the following situation:
- user executes `db._dropDatabase("foo");`
- user executes `db._createDatabase("foo");`
- DatabaseManagerThread now finds the deleted database "foo". Shall it remove its application directory now? We cannot be sure about it, because it may be the new database's application directory.

arangod/RestServer/DatabaseFeature.cpp
MBkkt: Understand. What about else (if we don't find anything on _droppedDatabases) branch? Is it necessary on coordinator.

Also what do you think about not start this thread on coordinator, and don't push anything to droppedDatabases.
Instead just call dtor + dealloc when we drop it

arangod/RestServer/DatabaseFeature.cpp
MBkkt: I understand, but we use id for new databases in path for v8 stuff (query register still use name :()

arangod/RestServer/FlushFeature.cpp
```diff
@@ -94,26 +94,18 @@ std::tuple<size_t, size_t, TRI_voc_tick_t> FlushFeature::releaseUnusedTicks() {
...
+      if (auto entry = e.lock(); entry) {
         minTick = std::min(minTick, entry->tick());
-        ++active;
-        ++itr;
+        return false;
       }
-    }
+      return true;
+    });
+    stale = static_cast<size_t>(end - it);
```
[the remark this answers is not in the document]
MBkkt: I completely disagree, if distance will do for each, I want instead of this code should be written ++stale in cycle above

arangod/VocBase/vocbase.cpp
```diff
@@ -470,84 +461,54 @@ struct arangodb::VocBaseLogManager {
...
-    if (_refCount.compare_exchange_weak(expected, updated,
-                                        std::memory_order_release,
-                                        std::memory_order_relaxed)) {
-      // compare-exchange worked. we're done
-      return true;
-    }
-    // compare-exchange failed. try again!
-    expected = _refCount.load(std::memory_order_relaxed);
-  }
+  } while (!_refCount.compare_exchange_weak(v, v + 2, std::memory_order_acquire,
```
jsteemann, line 470: Why is the memory order `acquire` here and not `release`?

MBkkt: Because we need to see markAsDropped, and our +2 not needed for synchronization code.

You can look on intrusive_ptr as an a good example of this pattern
https://www.boost.org/doc/libs/1_81_0/libs/atomic/doc/html/atomic/usage_examples.html#boost_atomic.usage_examples.example_reference_counters

arangod/VocBase/vocbase.cpp
```diff
@@ -470,84 +461,54 @@ struct arangodb::VocBaseLogManager {
...
+                                            std::memory_order_relaxed));
+  return true;
 }
 
-void TRI_vocbase_t::forceUse() { _refCount += 2; }
+void TRI_vocbase_t::forceUse() noexcept {
+  _refCount.fetch_add(2, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
MBkkt: Not

arangod/VocBase/vocbase.cpp
MBkkt: For -2 we need release because we want to synchronize with isDangling

arangod/VocBase/vocbase.cpp
```diff
@@ -470,84 +461,54 @@ struct arangodb::VocBaseLogManager {
...
-    return false;
-  }
-  auto refCount = _refCount.load();
-  // we are intentionally comparing with exactly 1 here, because a 1 means
-  // that noone else references the database but it has been marked as deleted
-  return (refCount == 1);
+bool TRI_vocbase_t::isDangling() const noexcept {
+  auto const v = _refCount.load(std::memory_order_acquire);
+  TRI_ASSERT((v & 1) == 0 || !isSystem());
+  return v == 1;
```
jsteemann, line 486: I think this could be a behavior change and may be unsafe.
It someone manages to set the drop bit on the system database, this function will now return `true`, but previously it always returned `false`. It could have unintended side-effects for example in the DatabaseManagerThread, which will remove the database then if this function here returns true.
ofc it would be useful to not allow to set the drop bit for the system database in the first place, but this function here previously provided some extra security.

arangod/RestServer/DatabaseFeature.cpp
jsteemann: Ideally, this thread should not be necessary at all on coordinators.
If we drop database objects immediately, it has a potential side-effect though: databases that are still in use elsewhere will then be deleted more quickly. So we may run into segfaults quickly if a database was looked up using `lookupDatabase(...)` and there is not protection (`useDatabase(...)`) in place. 
Previously, due to the potential delay that the DatabaseManagerThread has added post-deletion, these problems may have been hidden. I am not saying it's good to hide problems, just saying here that the change could lead to more crashes unless the other (potential) issues are fixed.

Dronplane (commented):
LGTM in general. See comments and this PR needs tests

arangod/VocBase/vocbase.cpp
MBkkt: No, please see assert in markAsDropped.
System database is never isDropped, so it's never isDangling

Dronplane: I guess ASSERT than should be our choice. As system database should never have delete bit set.

arangod/RestServer/DatabaseFeature.cpp
MBkkt: Not always, please check following code

```cpp
/// @brief return either the name of the database to be used as a folder name,
/// or its id if its name contains special characters and is not fully supported
/// in every OS
[[nodiscard]] static std::string_view getDatabaseDirName(std::string_view name,
                                                         std::string_view id) {
  bool const isOldStyleName = DatabaseNameValidator::isAllowedName(
      /*allowSystem=*/true, /*extendedNames=*/false, name);
  return (isOldStyleName || id.empty()) ? name : id;
}
```

arangod/RestServer/DatabaseFeature.cpp
MBkkt: >If we drop database objects immediately, it has a potential side-effect though

Yes you're right, we can dealloc only dangling database.
Callback on release can be solution, but I'm not insist.
And I don't do it in this PR

> So we may run into segfaults quickly if a database was looked up using lookupDatabase(...) and there is not protection (useDatabase(...)) in place.

But it's another issue, and probably should be fixed

MBkkt: What about else (if we don't find anything on _droppedDatabases) branch? Is it necessary on coordinator?

This question still actual

arangod/IResearch/IResearchFeature.cpp
```diff
@@ -1246,6 +1246,27 @@ void IResearchFeature::unprepare() {
...
 
+std::filesystem::path getPersistedPath(DatabasePathFeature const& dbPathFeature,
+                                       TRI_vocbase_t& database) {
+  std::filesystem::path path{dbPathFeature.directory()};
+  path /= "databases";
+  path /= absl::StrCat("database-", database.id());
+  return path;
+}
+
+void IResearchFeature::removeDatabase(TRI_vocbase_t& database) {
```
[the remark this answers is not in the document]
MBkkt: > Could we make this method static?

Done

> Or we could have DatabasePathFeature as argument

Not done

.clang-format
```diff
@@ -6,7 +6,9 @@ AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlines: Left
 AlignOperands:   true
-AlignTrailingComments: true
+AlignTrailingComments:
```
joerg84, line 9: Any reason why this is changed in this PR?

MBkkt: It's not intentionally, locally I have another version of clang format

MBkkt: Reverted

Dronplane (approved):
LGTM

arangod/RestServer/DatabaseFeature.cpp
jsteemann: Yes, the only code that the coordinator will execute is dropped all dangling databases in the `_droppedDatabases` list.
Everything else is skipped anyway on the coordinator.

jsteemann (approved):
LGTM

PR #18435, opened 2023-03-24
[GORDO-1582] Cancel pregel run

jvolmer:
With this PR you are able to cancel a pregel run. This does two things:
1. If pregel is still running: Cancel the run and cleanup the conductor (setting it to finished such that actor garbage collection can remove it). For that we now have now another map of execution number to conductor pid inside the pregel feature, such that we can find the conductor for the corresponding execution number.
2. Independent of pregel running or already finished: Cleanup all result actors (on conductor and workers). For this to work, the result actor on the coordinator has to know the actor pids of the result actors on the dbservers: In the RestPregelHandler this pid is send to the result actor on the coordinator after a result actor is created on the dbserver.

This PR adds the contains function to the actor runtime which checks if the runtime contains an actor with a given actor pid. This is needed because the map of execution number to conductor pid in the feature is not necessarily up-to-date: A conductor finishes itself after the pregel run finished and is then garbage collected, but this cannot be communicated to the feature. Therefore we always have to check if a conductor in this map still exists in the runtime.

### DIFF
--- a/arangod/Pregel/Actor/include/Actor/ActorList.h
+++ b/arangod/Pregel/Actor/include/Actor/ActorList.h
@@ -48,6 +48,11 @@ struct ActorList {
   ActorList(ActorMap map) : actors{std::move(map)} {}
   ActorList() = default;
 
+  auto contains(ActorID id) const -> bool {
+    return actors.doUnderLock(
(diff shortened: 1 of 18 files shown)


markuspf (approved):
LGTM

PR #18436, opened 2023-03-24
[CINFRA] Remove getCommittedLog finally.

maierlars:
### Scope & Purpose
Finaly step of evicting log entries from memory is to remove the function `getCommittedLog`. After that, it is no longer possible to get a _global_ view of the entire log at once, but instead the only option is to receive a log iterator. 

This iterator can be served from disk easily.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/Components/LogFollower.cpp
+++ b/arangod/Replication2/ReplicatedLog/Components/LogFollower.cpp
@@ -131,14 +131,15 @@ FollowerManager::~FollowerManager() {
 
 auto FollowerManager::getStatus() const -> LogStatus {
   auto commitIndex = commit->getCommitIndex();
-  auto log = storage->getCommittedLog();
+  auto mapping = storage->getTermIndexMapping();
   auto [releaseIndex, lowestIndexToKeep] = compaction->getIndexes();
   return LogStatus{FollowerStatus{
       .local =
           LogStatistics{
-              .spearHead = log.getLastTermIndexPair(),
+              .spearHead = mapping.getLastIndex().value_or(TermIndexPair{}),
               .commitIndex = commitIndex,
-              .firstIndex = log.getFirstIndex(),
+              .firstIndex =
+                  mapping.getFirstIndex().value_or(TermIndexPair{}).index,
               .releaseIndex = releaseIndex,
           },
       .leader = termInfo->leader,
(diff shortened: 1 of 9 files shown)


arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -324,7 +320,41 @@ auto StorageManager::getTermIndexMapping() const -> TermIndexMapping {
...
+
+  auto diskIter = guardedData.getLockedGuard()->methods->read(range.from);
+
+  struct Iterator : PersistedLogIterator {
+    explicit Iterator(LogRange range,
+                      std::unique_ptr<PersistedLogIterator> disk)
+        : _range(range), _disk(std::move(disk)) {}
+
+    auto next() -> std::optional<PersistingLogEntry> override {
+      while (true) {
```
apetenchea, line 341: NIT. It would not matter for the correctness of the code, but is there any reason why this is wrapped inside a `while (true)`?

arangod/Replication2/ReplicatedLog/Components/LogFollower.cpp
```diff
@@ -131,14 +131,15 @@ FollowerManager::~FollowerManager() {
 
 auto FollowerManager::getStatus() const -> LogStatus {
   auto commitIndex = commit->getCommitIndex();
-  auto log = storage->getCommittedLog();
+  auto mapping = storage->getTermIndexMapping();
```
apetenchea, line 134: This method returns `onDiskMapping`. As far as I can see from the code, it is the same as `spearheadMapping`, so it seems fine that it's used as the spearhead value below. I'm writing this comment as a double check.

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
maierlars: copy pasta! thanks

arangod/Replication2/ReplicatedLog/Components/LogFollower.cpp
maierlars: this is a bit misleading. The spearhead of a follower is the highest log index that is persisted on disk. However, in the storage manager we can have multiple storage transactions that depend on each other. The virtual result of the last transaction is also called spearhead. If one of the transactions fails, we abort them in a cascade.

PR #18438, opened 2023-03-24
[CircleCI] Detect if compiling changes checkout

markuspf:
### Scope & Purpose

Detect (and fail) if compiling the code changes the checked out code.


markuspf:
Need to test whether this works still;

PR #18439, opened 2023-03-24
[CINFRA] Remove InMemoryLogs from StorageManager

maierlars:
### Scope & Purpose

Remove all InMemoryLogs from StorageManager. On the leader on the uncommitted and not locally persisted log entries reside in memory. On the follower, only those not yet persisted.


mchacki (approved):
LGTM

apetenchea (approved):
LGTM

PR #18440, opened 2023-03-24
Added more restrictive handling of Collections

mchacki:
### Scope & Purpose

*This is a rudimentary fix for the WebUI to adhere to the more strict createCollection API.*

This fix ensures that only valid combinations are send over to ArangoDB.
However it would be great if someone with more UX experience could add a follow-up PR to make those restrictions a little more clearer for the user.

Basically:
DistributeShardsLike wins, if you set it, you give up control over shards/replicationFactor/writeConcern.
If you leave it out you can set the other 3.
If you are "oneShard" you are in the distributeShardsLike situation, you however cannot define the distributeShardsLike value yourself, that is injected by the OneShardDB.

Also please let me know if I need to check in the generated UI files, I am happy to add them.

- [X] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made (On purpose no entry, this bug is in no released version)
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
```diff
@@ -434,20 +434,27 @@
...
+              // if we are not using distribute shards like
+              // then we want to make use of the given shard information
+              tmpObj.shards = shards;
+              tmpObj.replicationFactor = replicationFactor === "satellite" ? replicationFactor : Number(replicationFactor);
+              tmpObj.writeConcern = Number(writeConcern);
+            } else {
+              // If we use distribute shards like on purpose do not add other
+              // sharding information. All of it will be deferred.
+              tmpObj.distributeShardsLike = distributeShardsLike;
+            }
```
jsteemann, line 457: It seems that this code is now executed for any !oneShardDB condition. Does it mean it is also executed in single server now? I think in single server we don't want to set any of these attributes. The previous code only executed the code in question when `window.App.isCluster` was set to true.

mchacki: Yes I did this on purpose.
On SingleServer the attributes are set and retains, but they will not have any effect.
This helps with dump/restore between single-server and clusters.

Happy to revert this, but in my opinion we should make the differences between single-server and cluster in terms of Meta-Data smaller.

js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
```diff
@@ -433,21 +433,32 @@
...
+                tmpObj.shards = shards;
+                tmpObj.replicationFactor = replicationFactor === "satellite" ? replicationFactor : Number(replicationFactor);
+                tmpObj.writeConcern = Number(writeConcern);
+              } else {
+                // If we use distribute shards like on purpose do not add other
+                // sharding information. All of it will be deferred.
+                tmpObj.distributeShardsLike = distributeShardsLike;
+              }
+            } else {
+              tmpObj.shards = shards;
```
jsteemann, line 460: Why are we assigning something to the `shards` attribute here in the non-cluster case?
It will populate the `numberOfShards` attribute in the end, but for what would that be needed when not running in the cluster?

jsteemann: Why should our backend care about `numberOfShards` in a single server? If it does, it would be a severe bug.

adityamukho: The backend API requires it as of now.

jsteemann: Why? A single server shard not care about `numberOfShards`.

jsteemann: I don't believe it.

adityamukho: Ah yes, I probably confused the error message for `shardKeys` with the `shards` parameter. Thanks for pointing this out.

CryptoNinjaGeek: @jsteemann So ok now?

jsteemann (approved):
LGTM

PR #18447, opened 2023-03-24
[GORDO-1581] Finish result actors after ttl

jvolmer:
This PR gives the result actors a time-to-live (TTL) that becomes active after the results are written into the actor. The actor sends periodically (every 20 seconds) a message to itself to check if the TTL expired and if so, it sets itself to finished.
This involves a modification in the actor framework: It adds a delay function to the scheduler, such that messages can be send delayed to the actor.

Additionally, this PR garbage collect the actors in the feature every 20 seconds:
- It calls the actor garbage collection which deletes all actors in the runtime that are finished
- It deletes all entries in the pregel-feature maps `_resultActor` and `_conductorActor` which do not exist in the runtime any more (because they were already garbage collected by the actor garbage collection).


markuspf (approved):
LGTM

PR #18449, opened 2023-03-25
Chore: Clean up eslint configuration for frontend a bit

mpoeter:
Scope & Purpose

Use separate eslintrc config for frontend, define gobals there and clean up source files.

This also updates the eslint version used in CircleCI to the much more recent v8.46.0.


palashkaria (approved):
LGTM! As discussed, we just need to fix the jenkins nightly pipeline to stop running this job as we move to circleCI!

PR #18454, opened 2023-03-27
testing: enable rocksdb logging

dothebart:
### Scope & Purpose

we found out startup may be slow inside of rocksdb, drill further down.


jsteemann (approved):
LGTM

PR #18455, opened 2023-03-27
[CINFRA] QuickFix for deadlock

maierlars:
### Scope & Purpose
On a former leader, there could still be some transactions open. If a snapshot transfer is required, the open transactions prevent dropping the shards.

This is a quick fix for this. In future the transaction handler should be stored in the Leader/Follower and not in the core, since transactions are explicitly not shared between terms.


apetenchea (approved):
LGTM

PR #18458, opened 2023-03-27
Add replication-version parameter to startLocalCluster.sh

maierlars:
### Scope & Purpose
Add `-r|--replication-version` parameter to the `startLocalCluster.sh` script. If not set, the default replication version is used.


apetenchea (approved):
LGTM

PR #18467, opened 2023-03-27
[GORDO-1594] Optimize internal pregel history API

hkernbach:
### Scope & Purpose

Follow up PR which basically removes an unnecessary complexity layer in our pregel history interface. 

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/GORDO-1594


markuspf (approved):
LGTM; we could, It think, remove the interface-indirection but this is not necessary right now.

PR #18470, opened 2023-03-27
[GORDO-1596] Query pregel actor status

jvolmer:
Adds a new status actor for every conductor.
If a pregel run ist started using actors, the pregel status is provided by the state this actor.

### DIFF
--- a/arangod/Pregel/REST/RestControlPregelHandler.cpp
+++ b/arangod/Pregel/REST/RestControlPregelHandler.cpp
@@ -34,6 +34,7 @@
 #include "Pregel/PregelFeature.h"
 #include "Pregel/REST/RestOptions.h"
 #include "Pregel/StatusWriter/CollectionStatusWriter.h"
+#include "Pregel/StatusActor.h"
 #include "Transaction/StandaloneContext.h"
 
 #include <velocypack/Builder.h>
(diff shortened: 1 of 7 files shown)


arangod/Pregel/REST/RestControlPregelHandler.cpp
```diff
@@ -171,6 +172,20 @@ void RestControlPregelHandler::handleGetRequest() {
     auto c = _pregel.conductor(executionNumber);
 
     if (nullptr == c) {
+      if (_pregel._statusActors.contains(executionNumber)) {
+        auto statusActor = _pregel._statusActors[executionNumber];
+        auto state =
+            _pregel._actorRuntime->getActorStateByID<pregel::StatusActor>(
+                statusActor.id);
```
markuspf, line 179: In an ideal world this would be a function `getStatusActorFor(executionNumber)`; I realise this is only used in this one place and I am nitpicking.

arangod/Pregel/StatusActor.h
```diff
@@ -0,0 +1,84 @@
...
+        "Status Actor: Error - network error {}", notFound.message);
+    return std::move(this->state);
+  }
+
+  auto operator()(auto&& rest) -> std::unique_ptr<StatusState> {
+    LOG_TOPIC("e9df2", INFO, Logger::PREGEL)
+        << "Status Actor: Got unhandled message";
+    return std::move(this->state);
+  }
+};
```
markuspf, line 72: One day we might want to figure out how to default handlers like this ;)

markuspf (changes requested):
Small comment about serialising the error message we should get from the inspector.

markuspf (approved):
LGTMN

jvolmer (commented):
nice

PR #18471, opened 2023-03-27
[GORDO-1597] send loading started update from conductor to status actor

jvolmer:
Adds status actor pid to conductor state, such that conductor is now able to send messages to the status actor.

Sends the message LoadingStarted from the conductor to the status update.

This PR shows an example where and how messages are send from the conductor to the status actor. More messages have to be send in the different actor execution states in this ticket (have a look at the ticket).


markuspf (approved):
Need to put down a guideline that requires full paths to be used for includes, I think.

PR #18474, opened 2023-03-28
[PREGEL bugfix] Move log before return

jvolmer:
Moves a log statement before a return statement, which resulted in spurious errors.


neunhoef (approved):
LGTM. This fixes compilation with gcc 11.2.0 for me. Thanks!

PR #18475, opened 2023-03-28
[CINFRA] Compaction Manager Deadlock

maierlars:
### Scope & Purpose
1. Fix a deadlock in the compaction manager. 
2. Move the iterator creation into a separate thread. This should reduce the appendEntries roundtrip time.


apetenchea (approved):
LGTM

goedderz (approved):

PR #18485, opened 2023-03-28
[CINFRA] Deadlock 456823

maierlars:
### Scope & Purpose
1. Sometimes a follower refuses to accept an append entries, because a previous one is still in flight. This is because the `requestGuard` was not reset explicitly at the end of the coroutine. It would then be reset once the coroutine is destroyed and (similar to futures) this can happen whenever the system feels like.
2. `recoverEntries` can deadlock with `updateCommitIndex`. However, `updateCommitIndex` is not used on the leader, therefore I removed the call (for now),
3. If we post the append entries on the scheduler, we have to make sure that the methods are still there. Otherwise we might get a nullptr access.

### DIFF
--- a/arangod/Replication2/ReplicatedLog/LogLeader.cpp
+++ b/arangod/Replication2/ReplicatedLog/LogLeader.cpp
@@ -750,7 +750,8 @@ auto replicated_log::LogLeader::GuardedLeaderData::updateCommitIndexLeader(
     _stateHandle->leadershipEstablished(std::make_unique<MethodsImpl>(_self));
   }
 
-  // Currently unused
+  // Currently unused, and could deadlock with recoverEntries, because that in
+  // return calls insert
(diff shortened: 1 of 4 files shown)


arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -750,7 +750,8 @@ auto replicated_log::LogLeader::GuardedLeaderData::updateCommitIndexLeader(
     _stateHandle->leadershipEstablished(std::make_unique<MethodsImpl>(_self));
   }
 
-  _stateHandle->updateCommitIndex(_commitIndex);
+  // Currently unused
```
goedderz, line 753:
```suggestion
  // Currently unused, and could deadlock with recoverEntries, because that in return calls insert
```

goedderz (approved):

PR #18492, opened 2023-03-29
[CINFRA] Added more metrics to the log leader.

maierlars:
### Scope & Purpose
This PR adds four new metrics:
1. Histogram of number of log entries per append-entries request
2. Histogram of append-entries request sizes
3. Number of log entries dropped by a follower before appending the log
4. Number of failed append-entries requests (only counting true errors, not NoPrevLogMatch)


apetenchea (approved):

nerpaula (approved):
LGTM

PR #18507, opened 2023-03-30
[CINFRA-714] Fix ResignLeadership

apetenchea:
### Scope & Purpose

The `ResignLeadership` job has an *undo* option, which has the effect of setting the old leader back once it is restarted. This option is not honored in the current replication2 implementation. 

There are differences between the way `ResignLeadership` works on replication 1 vs 2.
**Replication 1**
- we do a `MoveShard`, hence the undo operation is also a `MoveShard`

**Replication 2**
- we do a `ReconfigureReplicatedLog`, hence the undo operation is also a `ReconfigureReplicatedLog`, setting the old leader back
- this is on purpose, because leadership is bound to replicated logs instead of shards
- `MoveShard` must support undo, even for replication2

Just as the `ResignLeadership` job created a `MoveShard` with the *undo* option for replication1, it now can create a `ReconfigureReplicatedLog` with *undo* option for replication2. This process stays the same as before: once the job is finished, it writes the necessary parameters to *Target/ReturnLeadership* and leaves the scheduling of the undo job to the supervision.

The supervision reacts to such undo procedures while executing the `checkUndoLeaderChangeActions` method. In *Target/ReturnLeadership*, the key can be the *shardId* (replication1), or the *logId* (replication2), each case being handled in a separate job type.

For replication2, the need to undo an actual `MoveShard` job after a `ResignLeadership` seems to be gone. Nevertheless, the code to do so is there, and in theory it is possible to perform such a task. I have create a ticket for simplifying the code at a later stage: [CINFRA-729](https://arangodb.atlassian.net/browse/CINFRA-729).

The `moving-shards-cluster-data-R2.js` tests have been added to rlog.

*Additional Bugfix*
In order to decide if the replicated log has converged, the replicated logs supervision used to check the state of all participants in `Current`. However, due to potential stale entries, it should filter out those not present in `Plan`.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

[CINFRA-729]: https://arangodb.atlassian.net/browse/CINFRA-729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


arangod/Agency/MoveShard.cpp
```diff
@@ -1551,6 +1551,10 @@ void MoveShard::addUndoMoveShard(Builder& ops, Builder const& job) const {
     ops.add("timeStamp", VPackValue(now));
     ops.add("removeIfNotStartedBy", VPackValue(deadline));
     ops.add("rebootId", VPackValue(rebootId.value()));
-    ops.add("moveShard", job.slice());
+    ops.add("type", "moveShard");
+    ops.add("database", _database);
+    ops.add("collection", _collection);
+    ops.add("fromServer", _from);
+    ops.add("toServer", _to);
```
maierlars, line 1558: This change can break existing cluster that contain undo operations. We should not interfere with existing implementations.

maierlars (changes requested):

maierlars (approved):

PR #18508, opened 2023-03-30
one more place where we need to deregister the PID from the monitor.

dothebart:
### Scope & Purpose

If we intend to crash the server, we need to deregister its PID from the monitor.

### DIFF
--- a/js/client/modules/@arangodb/testutils/instance.js
+++ b/js/client/modules/@arangodb/testutils/instance.js
@@ -884,16 +884,15 @@ class instance {
     }
   }
   killWithCoreDump (message) {
-    internal.removePidFromMonitor(this.pid);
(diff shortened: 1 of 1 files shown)


js/client/modules/@arangodb/testutils/instance.js
```diff
@@ -884,6 +884,7 @@ class instance {
     }
   }
   killWithCoreDump (message) {
+    internal.removePidFromMonitor(this.pid);
```
alexbakharew, line 887: Should we check this condition before removing?
```
if (this.options.enableAliveMonitor) {
      internal.removePidFromMonitor(this.pid);
}
```

KVS85 (approved):

PR #18513, opened 2023-03-30
[GORDO-1598] Send status updates from worker

jvolmer:
Sends status updates from the worker actor to the status actor. There are 3 different message types for the different phases loading, computing (one message per global superstep) and storing.
The loading and storing messages are sent from the GraphLoader and GraphStorer, I adapted them a bit such you can now either give it an old status update callback (such that these two still work with the old pregel) or a callback to call the actor-dispatch function. (After we delete the old pregel without actors, we should think about how to not need to use a callback here.)
The status actor updates the status of each worker separately and aggregates the status of the workers to get a total value on all the variables.


markuspf (approved):
LGTM

PR #18516, opened 2023-03-30
Index progress visibility

kvahed:
### Scope & Purpose

**Index creation visibility** Allows one to check `_api/index` for `progress` 

- [ ] :hankey: Bugfix
- [X] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [X] Backport for 3.11: *https://github.com/arangodb/arangodb/pull/19369*

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1246
- [ ] Design document:


arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
...
+          network::sendRequestRetry(pool, "shard:" + shard.first,
+                                    fuerte::RestVerb::Get, url, body, reqOpts));
+    }
+    for (Future<network::Response>& f : futures) {
+      network::Response const& r = f.get();
+      if (r.fail()) {
+        LOG_TOPIC("afde4", INFO, Logger::CLUSTER)
+            << "Communication error while collecting figures for collection "
+            << _collection.name() + " from " + r.destination;
+      }
```
jsteemann, line 157: we probably should not go on here if `r` contains a failure.
```suggestion
        continue;
      }
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
...
+        LOG_TOPIC("agbe4", INFO, Logger::CLUSTER)
+            << "Result of collecting figures for collection "
+            << _collection.name() + " from " + r.destination << " is invalid";
+      }
+      if (resSlice.get(StaticStrings::Error).getBoolean()) {
+        LOG_TOPIC("a4beg", INFO, Logger::CLUSTER)
+            << "Failed to collect figures for collection "
+            << _collection.name() + " from " + r.destination;
+      }
+      if (resSlice.hasKey("progress") && resSlice.get("progress").isNumber()) {
```
jsteemann, line 170: Can be simplified to just
```suggestion
      if (resSlice.get("progress").isNumber()) {
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
...
+        progress += resSlice.get("progress").getNumber<uint64_t>();
+        success++;
+      } else {
+        LOG_TOPIC("aegb4", TRACE, Logger::CLUSTER)
+            << "No progress entry on index " << std::to_string(_iid.id())
```
jsteemann, line 175:
```suggestion
            << "No progress entry on index " << _iid.id()
```

arangod/ClusterEngine/ClusterIndex.cpp
jsteemann: Note: this and the following error checking sequence can be simplified by checking `r.combinedResult()`, which will already contain the error message in case something went wrong.

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +479,28 @@ ClusterIndex::coveredFields() const {
...
+
+bool ClusterIndex::inProgress() const {
+  auto const& vocbase = _collection.vocbase();
+  auto const& dbname = vocbase.name();
+  auto const cid = std::to_string(_collection.id().id());
+  auto const& agencyCache =
+      vocbase.server().getFeature<ClusterFeature>().agencyCache();
+  auto [acb, idx] =
+      agencyCache.read(std::vector<std::string>{AgencyCommHelper::path(
+          "Plan/Collections/" + dbname + "/" + cid + "/indexes")});
```
jsteemann, line 491: this must use URL-encoding
```suggestion
          "Plan/Collections/" + basics::StringUtils::urlEncode(dbname) + "/" + cid + "/indexes")});
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +479,28 @@ ClusterIndex::coveredFields() const {
...
+  auto slc = acb->slice()[0].get(std::vector<std::string>{
+      "arango", "Plan", "Collections", vocbase.name()});
+  if (slc.hasKey(std::vector<std::string>{cid, "indexes"})) {
+    slc = slc.get(std::vector<std::string>{cid, "indexes"});
+    for (auto const& index : VPackArrayIterator(slc)) {
+      if (index.get("id").copyString() == std::to_string(_iid.id())) {
```
jsteemann, line 497:
```suggestion
      if (index.get("id").stringView() == std::to_string(_iid.id())) {
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +479,28 @@ ClusterIndex::coveredFields() const {
...
+        if (index.hasKey("isBuilding")) {
+          LOG_TOPIC("fdae4", INFO, Logger::CLUSTER)  // kaveh remove
+              << " " << index.toJson();
+        }
+      }
+    }
+  }
+  return false;
```
jsteemann, line 505: function always returns false at the moment

arangod/Indexes/Index.cpp
```diff
@@ -199,7 +199,7 @@ Index::SortCosts Index::SortCosts::defaultCosts(size_t itemsInIndex) {
     Index::emptyCoveredFields{};
 
 // If the Index is on a coordinator instance the index may not access the
-// logical collection because it could be gone!
+// logical collection because it could be gone!Index::Index(
```
jsteemann, line 202:
```suggestion
// logical collection because it could be gone!
```

arangod/Indexes/Index.cpp
```diff
@@ -531,6 +531,9 @@ void Index::toVelocyPack(
               arangodb::velocypack::Value(oldtypeName(type())));
   builder.add(arangodb::StaticStrings::IndexName,
               arangodb::velocypack::Value(name()));
+  if (ServerState::instance()->isCoordinator()) {
+    builder.add("progress", VPackValue(66));
+  }
```
jsteemann, line 537:
```suggestion
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -27,6 +27,7 @@
 #include "Cluster/ClusterFeature.h"
 #include "Cluster/ClusterInfo.h"
 #include "Cluster/ServerState.h"
+#include "Logger/LogMacros.h"
```
jsteemann, line 30: Not necessary, logger not used in RestIndexHandler
```suggestion
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -152,6 +155,15 @@ Result fillIndexSingleThreaded(
     numDocsWritten++;
 
     if (numDocsWritten % 1024 == 0) {  // commit buffered writes
+      if (progress != nullptr && count > 0) {
+        auto pres = (*progress)(docsProcessed.load(std::memory_order_relaxed) *
+                                100 / count);
+        if (!pres.ok()) {
+          res.reset(TRI_ERROR_TRANSACTION_ABORTED, pres.errorMessage());
```
jsteemann, line 162:
```suggestion
          res = std::move(pres);
```

lib/Basics/sdt.h
jsteemann, line 1: This file should not be modified at all by this PR.
Please revert, as the changes will make it harder to update it in the future (it is a file not maintained by us).

lib/Basics/sdt.h
```diff
@@ -5,7 +5,7 @@
 */
 
 #ifndef _SYS_SDT_H
-#define _SYS_SDT_H    1
```
Dronplane, line 8: Are changes in this file intentional?

arangod/Cluster/ActionBase.h
```diff
@@ -112,6 +112,14 @@ class ActionBase {
   /// @brief finalize statistics
   void endStats();
 
+  /**
+   *  @brief    update progress by long running processes, while allowing to
+   *            abort the process by user.
+   *  @param  u percentage of work done
+   *  @return   abort if !ok(), true if ok(), with reason to abort.
+   */
+  virtual arangodb::Result setProgress(uint64_t u);
```
MBkkt, line 121: I don't understand why Result needed now.
Could you explain for me please?

arangod/Cluster/EnsureIndex.cpp
```diff
@@ -150,9 +155,11 @@ bool EnsureIndex::first() {
       // continue with the job normally
     }
 
+    auto lambda = std::make_shared<std::function<arangodb::Result(uint64_t u)>>(
+        [this](uint64_t u) { return setProgress(u); });
```
MBkkt, line 159: Why couldn't you pass just function?

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
     builder.add(StaticStrings::IndexEstimates, VPackValue(false));
   }
 
+  if (Index::hasFlag(flags, Index::Serialize::Figures)) {
```
MBkkt, line 133: I don't understand. Is it possible to call "collection.getIndexes" and receive some index which is not finished building?

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
...
+        _collection.vocbase().server().getFeature<NetworkFeature>().pool();
+    std::vector<Future<network::Response>> futures;
+    futures.reserve(shards->size());
+    std::string const prefix = "/_api/index/";
+    for (auto const& shard : *shards) {
+      network::RequestOptions reqOpts;
+      std::string const url =
+          prefix + shard.first + "/" + std::to_string(_iid.id());
+      futures.emplace_back(
+          network::sendRequestRetry(pool, "shard:" + shard.first,
```
MBkkt, line 148: Is this code can be called for index which is already finished creation?
If yes I think it should be fixed.

arangod/Cluster/ActionBase.h
MBkkt, line 121: Maybe double progress is better?

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,57 @@ void ClusterIndex::toVelocyPack(
...
+    auto const body = VPackBuffer<uint8_t>();
+    auto* pool =
+        _collection.vocbase().server().getFeature<NetworkFeature>().pool();
+    std::vector<Future<network::Response>> futures;
+    futures.reserve(shards->size());
+    std::string const prefix = "/_api/index/";
+    for (auto const& shard : *shards) {
+      network::RequestOptions reqOpts;
+      std::string const url =
+          prefix + shard.first + "/" + std::to_string(_iid.id());
```
MBkkt, line 146: NIT StrCat

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +479,28 @@ ClusterIndex::coveredFields() const {
       return _fields;
   }
 }
+
+bool ClusterIndex::inProgress() const {
```
MBkkt, line 483: Is this method called/used?

Also it doesn't look good, because if I will call it for N indexes it will be read Collection indexes N times

MBkkt (commented):
Is it work? I'm not sure but it's probably draft

gnusi:
We currently have a machinery for global cluster metrics in the code. IMHO this feature must be implemented on top of it.

arangod/Cluster/EnsureIndex.cpp
```diff
@@ -90,6 +90,12 @@ EnsureIndex::EnsureIndex(MaintenanceFeature& feature,
 
 EnsureIndex::~EnsureIndex() = default;
 
+// For local book keeping and reporting on /_admin/actions
+arangodb::Result EnsureIndex::setProgress(uint64_t u) {
```
jsteemann, line 94: As progress is reported as a double value inside the index itself, wouldn't it be useful to make it a double value here as well?

arangod/Cluster/EnsureIndex.cpp
```diff
@@ -90,6 +90,12 @@ EnsureIndex::EnsureIndex(MaintenanceFeature& feature,
...
+  _progress = u;
+  return arangodb::Result();
```
jsteemann, line 96: Nit: can be simplified
```suggestion
  return {};
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+    }
+    for (Future<network::Response>& f : futures) {
+      network::Response const& r = f.get();
+
+      // Only best effort accounting. If something breaks here, we just
+      // ignore the output. Account for what we can and move on.
+      if (r.fail()) {
+        LOG_TOPIC("afde4", INFO, Logger::CLUSTER)
+            << "Communication error while collecting figures for collection "
+            << _collection.name() + " from " + r.destination;
```
jsteemann, line 162:
```suggestion
            << _collection.name() << " from " << r.destination;
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+    auto* pool =
+        _collection.vocbase().server().getFeature<NetworkFeature>().pool();
+    std::vector<Future<network::Response>> futures;
+    futures.reserve(shards->size());
+    std::string const prefix = "/_api/index/";
+    for (auto const& shard : *shards) {
+      network::RequestOptions reqOpts;
+      reqOpts.param("withHidden", "true");
+      // best effort. only displaying progress
+      reqOpts.timeout = network::Timeout(10.0); // jsteeman, ideas?
```
jsteemann, line 147: RequestOptions can be created outside of the loop, as they are the same for every shard that is contacted.
```suggestion
    std::string const prefix = "/_api/index/";
    network::RequestOptions reqOpts;
    reqOpts.param("withHidden", "true");
    // best effort. only displaying progress
    reqOpts.timeout = network::Timeout(10.0);    
    for (auto const& shard : *shards) {
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+            << "Communication error while collecting figures for collection "
+            << _collection.name() + " from " + r.destination;
+        continue;
+      }
+      VPackSlice resSlice = r.slice();
+      if (!resSlice.isObject() || !resSlice.hasKey(StaticStrings::Error) ||
+          !resSlice.get(StaticStrings::Error).isBoolean()) {
+        LOG_TOPIC("agbe4", INFO, Logger::CLUSTER)
+            << "Result of collecting figures for collection "
+            << _collection.name() + " from " + r.destination << " is invalid";
```
jsteemann, line 170:
```suggestion
            << _collection.name() << " from " << r.destination << " is invalid";
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+      // ignore the output. Account for what we can and move on.
+      if (r.fail()) {
+        LOG_TOPIC("afde4", INFO, Logger::CLUSTER)
+            << "Communication error while collecting figures for collection "
+            << _collection.name() + " from " + r.destination;
+        continue;
+      }
+      VPackSlice resSlice = r.slice();
+      if (!resSlice.isObject() || !resSlice.hasKey(StaticStrings::Error) ||
+          !resSlice.get(StaticStrings::Error).isBoolean()) {
```
jsteemann, line 167:
```suggestion
      if (!resSlice.isObject() || !resSlice.get(StaticStrings::Error).isBoolean()) {
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+        LOG_TOPIC("agbe4", INFO, Logger::CLUSTER)
+            << "Result of collecting figures for collection "
+            << _collection.name() + " from " + r.destination << " is invalid";
+        continue;
+      }
+      if (resSlice.get(StaticStrings::Error).getBoolean()) {
+        LOG_TOPIC("a4beg", INFO, Logger::CLUSTER)
+            << "Failed to collect figures for collection "
+            << _collection.name() + " from " + r.destination;
```
jsteemann, line 176:
```suggestion
            << _collection.name() << " from " << r.destination;
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -123,6 +130,66 @@ void ClusterIndex::toVelocyPack(
...
+        continue;
+      }
+      if (resSlice.hasKey("progress") && resSlice.get("progress").isNumber()) {
+        progress += resSlice.get("progress").getNumber<double>();
+        success++;
+      } else {
+        LOG_TOPIC("aegb4", INFO, Logger::CLUSTER)
+            << "No progress entry on index " << std::to_string(_iid.id())
+            << "  from " + r.destination << ": " << resSlice.toJson();
```
jsteemann, line 185:
```suggestion
            << "  from " << r.destination << ": " << resSlice.toJson();
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +488,28 @@ ClusterIndex::coveredFields() const {
...
+          "Plan/Collections/" + dbname + "/" + cid + "/indexes")});
+  auto slc = acb->slice()[0].get(std::vector<std::string>{
+      "arango", "Plan", "Collections", vocbase.name()});
+  if (slc.hasKey(std::vector<std::string>{cid, "indexes"})) {
+    slc = slc.get(std::vector<std::string>{cid, "indexes"});
+    for (auto const& index : VPackArrayIterator(slc)) {
+      if (index.get("id").copyString() == std::to_string(_iid.id())) {
+        if (index.hasKey("isBuilding")) {
+          LOG_TOPIC("fdae4", INFO, Logger::CLUSTER)  // kaveh remove
+              << " " << index.toJson();
```
jsteemann, line 509: What should happen here instead of logging?

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +488,28 @@ ClusterIndex::coveredFields() const {
...
+  auto [acb, idx] =
+      agencyCache.read(std::vector<std::string>{AgencyCommHelper::path(
+          "Plan/Collections/" + dbname + "/" + cid + "/indexes")});
+  auto slc = acb->slice()[0].get(std::vector<std::string>{
+      "arango", "Plan", "Collections", vocbase.name()});
+  if (slc.hasKey(std::vector<std::string>{cid, "indexes"})) {
+    slc = slc.get(std::vector<std::string>{cid, "indexes"});
+    for (auto const& index : VPackArrayIterator(slc)) {
+      if (index.get("id").copyString() == std::to_string(_iid.id())) {
+        if (index.hasKey("isBuilding")) {
```
jsteemann, line 507:
```suggestion
        if (index.hasKey(StaticStrings::IndexIsBuilding)) {
```

lib/Basics/sdt.h
kvahed: weird. yes. no idea, where these changes came from.

arangod/ClusterEngine/ClusterIndex.cpp
kvahed: Yes, to see their progress. But of course not so they are used for queries.

arangod/ClusterEngine/ClusterIndex.cpp
kvahed: Again, only, when the figures are explicitly requested. This is a very verbose output. Due to time constraints, I would suggest a fix after my holidays.

arangod/Cluster/ActionBase.h
kvahed: Maybe. This is not new code. I have no hard feelings either way. Out of scope of this PR.

arangod/Cluster/EnsureIndex.cpp
kvahed: So, didn't change the base class, where this is coming from. Historic reasons.

arangod/Indexes/Index.cpp
kvahed: Reviewed too early. ;)

arangod/Cluster/ActionBase.h
kvahed: Again, this is not the scope of this PR. There is however a possibility, why such a prospective function impl might fail.

arangod/Cluster/EnsureIndex.cpp
kvahed: Only a matter of preference. There are no performance implications. Will keep it as is, if you allow.

arangod/ClusterEngine/ClusterIndex.cpp
kvahed: fixed

arangod/ClusterEngine/ClusterIndex.cpp
kvahed: fixed

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -421,3 +488,30 @@ ClusterIndex::coveredFields() const {
...
+        agencyCache.read(std::vector<std::string>{AgencyCommHelper::path(
+            "Plan/Collections/" + basics::StringUtils::urlEncode(dbname) + "/" +
+            cid + "/indexes")});
+    auto slc = acb->slice()[0].get(std::vector<std::string>{
+        "arango", "Plan", "Collections", vocbase.name()});
+    if (slc.hasKey(std::vector<std::string>{cid, "indexes"})) {
+      slc = slc.get(std::vector<std::string>{cid, "indexes"});
+      for (auto const& index : VPackArrayIterator(slc)) {
+        if (index.get("id").stringView() == std::to_string(_iid.id())) {
+          return index.hasKey("isBuilding");
```
jsteemann, line 510:
```suggestion
          return index.hasKey(StaticStrings::IndexIsBuilding);
```

jsteemann (approved):
LGTM

gnusi (changes requested):
Please add at least a shallow test.

MBkkt (changes requested):
I don't think make N (number of shards for index collection) requests per every index from every coordinator is good idea.

Also now it's read all collection indexes definition from agency cache for every index which is also not good

arangod/Cluster/ActionBase.h
```diff
@@ -112,6 +112,14 @@ class ActionBase {
   /// @brief finalize statistics
   void endStats();
 
+  /**
+   *  @brief    update progress by long running processes, while allowing to
+   *            abort the process by user.
+   *  @param  d percentage of work done
+   *  @return   abort if !ok(), true if ok(), with reason to abort.
```
neunhoef, line 119: I do not understand this comment and thus I do not understand the API.

arangod/Cluster/EnsureIndex.cpp
```diff
@@ -90,6 +90,12 @@ EnsureIndex::EnsureIndex(MaintenanceFeature& feature,
 
 EnsureIndex::~EnsureIndex() = default;
 
+// For local book keeping and reporting on /_admin/actions
+arangodb::Result EnsureIndex::setProgress(double d) {
+  _progress = static_cast<uint64_t>(d);
```
neunhoef, line 95: Why is `_progress` an `uint64_t`, if the API seems to suggest that `double` is used to express a percentage? Please clarify.

neunhoef: It seems it is in fact a `double`, so why casting it to `uint64_t` here?

neunhoef (changes requested):
In absence of a proper explanation or design doc I can only guess what is intended to implement here. The actual progress tracking of individual shard's indexes seems basically fine. This is reported back to the index object in batches, which seems to be a sane approach.
As far as I see from code study, the idea seems to be that in a cluster environment, the method which dumps the information about an index to velocypack is supposed to return progress information if the index is being built, which means that the `getIndexes` method will report them. We should probably ensure that indexes in progress are not dumped unless a specific option is given, but I cannot see this in the code so far (maybe I have overlooked something).
In general, fetching indexes should be a rare operation and is not performance critical. Therefore it is probably OK to query the AgencyCache about the index every time the API is called, and if it is in progress, we can query all the shards. I think this is the simplest solution and therefore I would support this.

arangod/Cluster/EnsureIndex.cpp
kvahed: For this I have no words. Simple fix. Obviously oversight. Left over from the bad idea in the `Maintenance` to use `uint64_t` as progress. One line change.

arangod/Cluster/ActionBase.h
kvahed: Removed the conflicting part.

neunhoef (approved):
LGTM.

PR #18517, opened 2023-03-30
[PREGEL] Fix nullptr issue in worker handler

jvolmer:
I discovered two issues in the worker handler code, that this PR fixes:
A) In the WorkerState constructor, we used the wrong message format and combiner to create the caches - we want to use the WorkerState properties message format and combiner and not the ones that are incoming and moved away (leading a call on a nullptr).
B) The dispatch method for the out cache is not created at place and not moved in which resulted in a bad_alloc in the scheduler.


markuspf (approved):
LGTM, though it scares me slightly that tests didn't catch it.

PR #18522, opened 2023-03-30
MDS-1001: Optimize Access to last element in Path.

mchacki:
### Scope & Purpose

*This PR adds a new optimizer rule: "optimize-traversal-last-element-access" that is targeted only to Traversals. We now look for an access pattern to the last element on the path `p.vertices[-1]` and replace it with `v`. Before we apply other Traversal optimizations. The above two statements are identical and reference the last vertex on the currently returned path. However the later variant has less internal overhead, and allows for some more optimizations. (e.g. pull the filter into the traversal in some cases)*

- [ ] :hankey: Bugfix
- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [X] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


js/common/bootstrap/errors.js
```diff
@@ -1,4 +1,5 @@
 /*jshint maxlen: 240 */
+/*global require */
```
mchacki, line 2:
```suggestion
```

CHANGELOG
```diff
@@ -1,6 +1,15 @@
 devel
 -----
 
+* MDS-1001: Performance improvement in AQL. If you are using a traversal like
+  `FOR v, e, p IN <....>` and later in the query access the last vertex on the path e.g.:
+  `FILTER p.vertices[-1].name == "ArangoDB"` it will now be transformed to
+  `FILTER v.name == "ArangoDB"` which is an equivalent statement. The later however
```
```suggestion
  `FILTER v.name == "ArangoDB"` which is an equivalent statement. The latter however
```

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "GraphOptimizerRules.h"
+
+#include "Aql/ExecutionNode.h"
+#include "Aql/ExecutionPlan.h"
+#include "Aql/Optimizer.h"
+#include "Aql/TraversalNode.h"
+#include "Containers/SmallVector.h"
+
+#include "Logger/LogMacros.h"
+
```
jsteemann, line 32: Seems unneeded
```suggestion
```

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+enum class PathAccessState {
+  // Search for access to any path variable
+  SEARCH_ACCESS_PATH,
+  // Check if we access edges or vertices on it
+  ACCESS_EDGES_OR_VERTICES,
+  // Check if we have an indexed access [x] (we find the x here)
+  SPECIFIC_DEPTH_ACCESS_DEPTH,
+  // CHeck if we have an indexed access [x] (we find [ ] here)
+  SPECIFIC_DEPTH_ACCESS_INDEXED_ACCESS,
```
jsteemann, line 46: According to our naming conventions, these should be named `kSearchAccessPath`, `kAccessEdgesOrVertices`, ...

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+        currentState = PathAccessState::SEARCH_ACCESS_PATH;
+        return node;
+      }
+    }
+  };
+
+  auto newCondition = Ast::traverseAndModify(condition, searchAccessPattern);
+  if (newCondition != condition) {
+    // Swap out everything.
+    return std::make_pair(appliedAChange, newCondition);
```
jsteemann, line 131: As far as I can see, `appliedAChange` should always be true here if we get here. 
Let's add an assertion here, i.e. `TRI_ASSERT(appliedAChange);`.

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  return std::make_pair(appliedAChange, nullptr);
```
jsteemann, line 134: As far as I can see, `appliedAChange` should always be false here if we get here. 
Let's add an assertion here, i.e. `TRI_ASSERT(!appliedAChange);`.

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+    return;
+  }
+
+  bool appliedAChange = false;
+  for (auto& n : tNodes) {
+    auto* calculation = ExecutionNode::castTo<CalculationNode*>(n);
+    auto [didApplyChange, replacementCondition] = swapOutLastElementAccesses(
+        plan->getAst(), calculation->expression()->nodeForModification(),
+        candidates);
+    // Get's true as soon as one of the swapOut calls returns true
```
jsteemann, line 197:
```suggestion
    // Gets true as soon as one of the swapOut calls returns true
```

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+    appliedAChange |= didApplyChange;
+    if (replacementCondition != nullptr) {
+      // This is the indicator that we have to replace the full expression here.
```
jsteemann, line 200: Let's also add `TRI_ASSERT(didApplyChange);` here.

arangod/Aql/GraphOptimizerRules.cpp
```diff
@@ -0,0 +1,206 @@
...
+      calculation->expression()->replaceNode(replacementCondition);
+    }
+  }
+  // Nothing done
```
jsteemann, line 204: Comment is not true. We may or may not have something done when we get here.
```suggestion
```

tests/js/server/aql/aql-optimizer-rule-last-access-on-path.js
```diff
@@ -0,0 +1,189 @@
...
+            if (typeID !== 35 || !(name === "vertices" || name === "edges") || subNodes.length !== 1 || !isReferenceToPath(subNodes[0])) {
+                return false;
+            }
+        }
+        return true;
+    };
+
+    const planContainsPathAccess = (nodes) => {
+        // Pick all Indexed Access nodes (`LET v = <var>[<index>]`
+        const indexedAccesses = nodes.filter(n => n.type === "CalculationNode").map(c => c.expression).filter(e => e.typeID === 37);
```
jsteemann, line 113: I am not really happy with using things such as `e.typeID === 37`.
Isn't there also a "type" attribute in each AstNode, which contains a string value of the node's type? That would be much better suited than using an enum value here.

tests/js/server/aql/aql-optimizer-rule-last-access-on-path.js
```diff
@@ -0,0 +1,189 @@
...
+    const isLastElementAccess = (indexedAccess) => {
+        if (indexedAccess.subNodes.length !== 2) {
+            // Index access needs exactly 2 members
+            return false;
+        }
+        const [path, index] = indexedAccess.subNodes;
+        {
+            // Test index
+            const {typeID, value} = index;
+            if (typeID !== 40 || value !== -1) {
```
jsteemann, line 96: Can we use the type name instead of the typeID? Seeing a typeID value `40` isn't really intuitive when looking at the test code.

tests/js/server/aql/aql-optimizer-rule-last-access-on-path.js
```diff
@@ -0,0 +1,189 @@
...
+                // Only type "value"(40) and access to last element "-1" can pass
+                return false;
+            }
+        }
+        {
+            // Test attribute access path
+            const {typeID, name, subNodes} = path;
+            if (typeID !== 35 || !(name === "vertices" || name === "edges") || subNodes.length !== 1 || !isReferenceToPath(subNodes[0])) {
```
jsteemann, line 104: Can we use the type name instead of the typeID? Seeing a typeID value `35` isn't really intuitive when looking at the test code.

tests/js/server/aql/aql-optimizer-rule-last-access-on-path.js
```diff
@@ -0,0 +1,189 @@
...
+        testShouldImpactTraversals: function () {
+            const queriesToTest = [
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" GRAPH "${graphName}" RETURN p.vertices[-1]`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" GRAPH "${graphName}" RETURN p.edges[-1]`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} RETURN p.vertices[-1]`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} RETURN p.edges[-1]`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} FILTER p.vertices[-1].name == "test" RETURN v`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} FILTER p.edges[-1].label == "foo" RETURN e`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} FILTER p.edges[-1].label == "foo" && p.vertices[-1].name == "test" RETURN v`,
+                `FOR v, e, p IN 1..2 OUTBOUND "${startVertex}" ${ec} SORT p.edges[-1].timestamp RETURN e`,
```
jsteemann, line 174: Additional test suggestion:
once we have done the replacement, e.g. replaced `p.vertices[-1]` with `v`, we should also check if projections are still correctly used. For example, `p.vertices[-1].timestamp` will be turned into `v.timestamp`, and in a different optimizer rule the optimizer should be smart enough to figure out that it only needs to extract the `timestamp` attribute from each vertex.
This is a separate optimizer rule, but the combination of the 2 rules should still work, so there should be a small test for that.

arangod/Aql/GraphOptimizerRules.cpp
mchacki: Right, I only needed it for my debug output, thanks for spotting

arangod/Aql/GraphOptimizerRules.cpp
jsteemann: How can it be the same pointer if it points to a different (i.e. modified) expression?

jsteemann: Ok, got it. You are calling the `traverseAndModify(...)` variant that internally uses `changeMember()` to modify the Ast in place.

jsteemann (approved):
LGTM

PR #18523, opened 2023-03-30
[CINFRA-700] Fix getResponsibleServer by relying on the supervision.

maierlars:
### Scope & Purpose
Currently the supervision for collection groups updates the shard map in plan. This shard map is copied by the leader into current. This will always contains the correct server as leader. Thus we can rely on the replication 1 way of looking up the shard.

This is much simpler than the previous code and it is correct! Once we got rid of the old shard map, we have touch this code anyways again.


apetenchea (approved):
LGTM

PR #18525, opened 2023-03-30
Bug fix/fix maintenance test error

maierlars:
### Scope & Purpose
Fix a small race in a test that caused sporadic jenkins failures.


jsteemann (approved):
LGTM

PR #18530, opened 2023-03-31
Fix for query execution stats info

cpjulia:
### Scope & Purpose

Fixed statistics values for writes executed and writes ignored when a query is executed using the rule `optimize-cluster-single-document-operations`. It was always increasing the amount of writes executed, even if the operation wasn't successful, and also never increasing the amount of writes ignored when needed.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18531

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


tests/js/server/aql/aql-optimizer-rule-optimize-cluster-single-document-operations.js
```diff
@@ -587,8 +587,104 @@ function optimizerClusterSingleDocumentTestSuite () {
...
+    },
 
+    testStatsWhenApplyingRuleOneAtATime: function () {
+      const collName = "testCollectionStats";
+      const doc = {value1: 1, value2: "a"};
+      const values = [1, 3];
+      values.forEach(numShards => {
+        try {
+          db._create(collName, {numberOfShards: numShards});
+          for (let i = 0; i < 10; ++i) {
```
jsteemann, line 599: What is the benefit of running all following operations 10 times? Shouldn't the results of all 10 iterations be exactly identical? I think the for loop is unnecessary here, unless I am missing something.

arangod/Aql/SingleRemoteModificationExecutor.cpp
```diff
@@ -198,7 +213,8 @@ auto SingleRemoteModificationExecutor<
     }
   }
 
-  stats.incrWritesExecuted(possibleWrites);
+  stats.incrWritesExecuted(writesExecuted);
+  stats.incrWritesIgnored(writesIgnored);
   stats.incrScannedIndex();
```
jsteemann, line 218: The `incrScannedIndex()` is not appropriate for write queries.
`incrScannedIndex()` should only be called if the optimizer rule is applied for a read query. Can you fix this and also add a test for this? Thanks!

jsteemann (changes requested):
See review comments

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #18533, opened 2023-03-31
[CINFRA-726] Expose snapshot status as part of the leader status.

maierlars:
### Scope & Purpose
The goal of this PR is to fix the display of servers in the webui for a given shard. We only want to report those servers in current that actually have received a snapshot from the leader (or already have one).


apetenchea (approved):
LGTM

PR #18536, opened 2023-03-31
testing: only enable the process monitoring in maintainer mode 

dothebart:
### Scope & Purpose

disable process monitoring thread in production arangosh'es.

### DIFF
--- a/client-tools/Shell/ProcessMonitoringFeature.cpp
+++ b/client-tools/Shell/ProcessMonitoringFeature.cpp
@@ -96,8 +96,12 @@ ProcessMonitoringFeature::~ProcessMonitoringFeature() = default;
 
 void ProcessMonitoringFeature::validateOptions(
     std::shared_ptr<options::ProgramOptions> /*options*/) {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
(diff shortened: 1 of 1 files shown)


client-tools/Shell/ProcessMonitoringFeature.cpp
```diff
@@ -96,8 +96,12 @@ ProcessMonitoringFeature::~ProcessMonitoringFeature() = default;
 
 void ProcessMonitoringFeature::validateOptions(
     std::shared_ptr<options::ProgramOptions> /*options*/) {
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
   _enabled =
       server().getFeature<V8SecurityFeature>().isAllowedToControlProcesses();
+#else
+  _enabled =
```
MBkkt, line 103: I except such ifdef for whole Feature in `client-tools/Shell/arangosh.h`.
In cmake also should be if to don't build this files in such case

dothebart: this would break the js API.

dothebart:
this effectively makes this a complete non-feature, hence closing this PR.

PR #18538, opened 2023-03-31
Add a function queueing a job and returning a future

markuspf:
### Scope & Purpose

Adds a function to Scheduler that queues a callable and returns a future that resolves when the callable has finished execution.


maierlars (approved):
LGTM

PR #18539, opened 2023-03-31
[CINFRA-719] Fix CollectionGroups test

maierlars:
### Scope & Purpose
Sometimes this test failed, because an empty collection group was still present, when the next test starts.


mchacki (approved):
LGTM

PR #18540, opened 2023-03-31
improve shutdown times (useful for testing)

jsteemann:
### Scope & Purpose

As suggested by @dothebart:

* Make REST API `/_admin/shutdown` sleep for only half a second until it initiates the server shutdown. Previously it slept for 2 seconds, but half a second should already be enough to send the server's response out.

The change should have a positive impact on the runtime of cluster tests that start and stop servers a lot (e.g. `server_parameters`).

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


dothebart (approved):

mpoeter:
Couldn't we instrad use a condition variable or an atomic with wait/notify to make this deterministic? This should usually be faster than half a second in most cases, but at the same time more stable in case of overload.

dothebart:
this is here so we know the server sends the reply before terminating.

arangod/RestHandler/RestShutdownHandler.cpp
```diff
@@ -127,8 +127,8 @@ RestStatus RestShutdownHandler::execute() {
   // don't block the response for workers waiting on this callback
   // this should allow workers to go into the IDLE state
   scheduler->queue(RequestLane::CLUSTER_INTERNAL, [self] {
```
MBkkt, line 129: Why don't use queue delayed?

CHANGELOG
```diff
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Make REST API `/_admin/shutdown` sleep for only half a second until it
+  initiates the server shutdown. Previously it slept for 2 seconds, but half
+  a second should already be enough to send the server's response out.
```
MBkkt, line 6: I agree with Manuel, it can be 0.001 enough and 2 not enough.

mpoeter:
@dothebart I am aware, but as @MBkkt has pointed out, 0.001s could be enough (and probably are in most cases), but sometimes even 2s might not be enough. So instead of sleeping for a fixed but arbitrary amount of time we should make this deterministic and sleep exactly as long as necessary.

jsteemann:
> @dothebart I am aware, but as @MBkkt has pointed out, 0.001s could be enough (and probably are in most cases), but sometimes even 2s might not be enough. So instead of sleeping for a fixed but arbitrary amount of time we should make this deterministic and sleep exactly as long as necessary.

How do we know what time is enough?
All we know is that we have called `generateResponse()`, but we don't know exactly when the response will be sent over the wire. We have no indication here if the client/caller has received the response except that if it sends another request via the same CommTask.
So what would be your suggestion to address this?

dothebart:
so could we determine when a result has been sent, and then continue the shutdown? Do we know when a job has sent its reply?

jsteemann:
We could track this in theory, but this would add unnecessary overhead for all responses.
The question is really if it is worth to do it just for the shutdown case (I don't think it is).

PR #18550, opened 2023-04-03
Make OutCache more independent of worker config

markuspf:
### Scope & Purpose

Make out caches more independent of worker config. 

The worker config is still needed because caches also take care of sending batches of messages across the network. This should be removed in a later step.

### DIFF
--- a/arangod/Pregel/OutgoingCache.cpp
+++ b/arangod/Pregel/OutgoingCache.cpp
@@ -52,15 +52,14 @@ using namespace arangodb::pregel::algos;
 
 template<typename M>
 OutCache<M>::OutCache(std::shared_ptr<WorkerConfig const> state,
+                      std::set<PregelShard> localShards,
                       MessageFormat<M> const* format)
-    : _config(state),
+    : _config(std::move(state)),
+      _localShards(std::move(localShards)),
       _format(format),
       _baseUrl(Utils::baseUrl(Utils::workerPrefix)) {}
 
 // ================= ArrayOutCache ==================
-
-template<typename M>
-ArrayOutCache<M>::~ArrayOutCache() = default;
 
 template<typename M>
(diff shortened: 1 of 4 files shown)


arangod/Pregel/Worker/Worker.cpp
```diff
@@ -157,7 +158,8 @@ void Worker<V, E, M>::_initializeMessageCaches() {
       auto incoming = std::make_unique<ArrayInCache<M>>(std::set<PregelShard>{},
                                                         _messageFormat.get());
       _inCaches.push_back(incoming.get());
-      _outCaches.push_back(new ArrayOutCache<M>(_config, _messageFormat.get()));
+      _outCaches.push_back(new ArrayOutCache<M>(
+          _config, _config->localPregelShardIDs(), _messageFormat.get()));
```
hkernbach, line 162: Minor: You could also use `_config->localPregelShardIDs()` in the constructor of ArrayOutCache and only pass 2 parameters instead of 3. 

Just ignore my comment in case you want to just use it to be more similar to the "ArrayInCache" constructor api.

hkernbach (approved):

arangod/Pregel/Worker/Worker.cpp
markuspf: It is the point of this PR to make `OutCache` more explicitly depend on things that are used from the "config"; in a further step in the coming days we will remove the dependency on `_config` entirely. 

I can hold off on merging this one until this is sorted, maybe, as to make this clear?

hkernbach: I don't mind. In case we have a ticket for this and do not forget - just merge it :)

jvolmer (approved):

markuspf:
This has been subsumed into the other PR atttached to the ticket.

PR #18562, opened 2023-04-03
[GORDO-1600] Multiple Quiver<V,E>s in a Worker

markuspf:
### Scope & Purpose

As a pre-step to re-parallelisation support multiple `Quiver<V, E>` in a Worker; they are summarised in a `Magazine<V,E>`.

Currently there is a 1-1 correspondence between `Quiver<V,E>`s and a shard of a vertex collection, so we can load vertex shards concurrently, and later process `Quiver<V,E>`s concurrently.

### DIFF
--- a/arangod/Pregel/GraphStore/GraphVPackBuilderStorer.cpp
+++ b/arangod/Pregel/GraphStore/GraphVPackBuilderStorer.cpp
@@ -59,7 +59,6 @@ auto GraphVPackBuilderStorer<V, E>::store(std::shared_ptr<Quiver<V, E>> quiver)
     -> void {
   std::string tmp;
 
-  result->openArray(/*unindexed*/ true);
   for (auto& vertex : *quiver) {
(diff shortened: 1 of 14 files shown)


arangod/Pregel/GraphStore/GraphVPackBuilderStorer.cpp
```diff
@@ -57,8 +57,6 @@ namespace arangodb::pregel {
 template<typename V, typename E>
 auto GraphVPackBuilderStorer<V, E>::store(std::shared_ptr<Quiver<V, E>> quiver)
     -> void {
-  result = std::make_unique<VPackBuilder>();
```
hkernbach, line 60: Is it safe here to skip this command? Could result therefore be uninitialized?

hkernbach (approved):
One minor comment. Apart from that LGTM - Prefer the new data structure in favor as our original attempt to not use any :P

PR #18564, opened 2023-04-04
[GORDO-1600] [GORDO-1585] Parallel vertex processing on worker

markuspf:
### Scope & Purpose

Re-parallelise processing of vertices on workers. Parallelisation is per-shard and future-based.

In preparation for parallelisation I made `_processVertices` more functional style and try not use global state.


jvolmer (approved):
LGTM

jvolmer:
We just have to keep in mind that we have to do the same thing inside the worker actor as well.

PR #18571, opened 2023-04-04
testing: need to actually invoke the code :/

dothebart:
### Scope & Purpose

It seems the code was there, but it wasn't actually exeuted

- [x] :hankey: Bugfix

### DIFF
--- a/js/client/modules/@arangodb/testutils/testrunner.js
+++ b/js/client/modules/@arangodb/testutils/testrunner.js
@@ -315,6 +315,7 @@ let failurePointsCheck = {
       };
       return false;
     }
+    return true;
   }
 };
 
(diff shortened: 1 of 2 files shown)


mpoeter (approved):

js/client/modules/@arangodb/testutils/testrunner.js
```diff
@@ -304,7 +304,7 @@ let failurePointsCheck = {
   name: 'failurepoints',
   setUp: function(obj, te) { return true; },
   runCheck: function(obj, te) {
-    let failurePoints = pu.checkServerFailurePoints(obj.instanceManager);
+    let failurePoints = obj.instanceManager.checkServerFailurePoints(obj.instanceManager);
```
KVS85, line 307: Why `obj.instanceManager.checkServerFailurePoints(obj.instanceManager)` can't just pick itself (`obj.instanceManager`) by default?

dothebart: the function may work with an instanceManager that has been de/serialized via the OS-environment

KVS85: Yes. But by default (empty value is passed) it can pick `this`.

PR #18577, opened 2023-04-04
[CINFRA] Add missing preconditions to collection groups supervision

maierlars:
### Scope & Purpose
There is not much to say about this: I was lazy and did not add preconditions to the agency transactions generated by the collection group supervision. This bit us and now I added them.

In almost all cases, it is required that the collection group in question is still in Target. If it is not, the only sensible action is to remove it everywhere else.

### DIFF
--- a/arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
+++ b/arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
@@ -500,14 +500,20 @@ struct TransactionBuilder {
   }
 
   void operator()(UpdateReplicatedLogConfig const& action) {
-    env = env.write()
-              .emplace_object(basics::StringUtils::concatT(
-                                  "/arango/Target/ReplicatedLogs/", database,
-                                  "/", action.logId, "/config"),
-                              [&](VPackBuilder& builder) {
-                                velocypack::serialize(builder, action.config);
-                              })
-              .end();
+    env =
+        env.write()
+            .emplace_object(basics::StringUtils::concatT(
+                                "/arango/Target/ReplicatedLogs/", database, "/",
+                                action.logId, "/config"),
+                            [&](VPackBuilder& builder) {
+                              velocypack::serialize(builder, action.config);
+                            })
+            .precs()
+            .isNotEmpty(basics::StringUtils::concatT(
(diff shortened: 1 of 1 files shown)


arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
```diff
@@ -529,6 +538,9 @@ struct TransactionBuilder {
                   "/arango/Plan/CollectionGroups/", database, "/", gid.id(),
                   "/collections/", action.cid))
               .inc("/arango/Plan/Version")
+              .precs()
+              .isNotEmpty(basics::StringUtils::concatT(
+                  "/arango/Target/CollectionGroups/", database, "/", gid.id()))
```
apetenchea, line 543: For the `DropCollectionPlan` action, I think this precondition might not be the best fit.

apetenchea: I find it an odd choice, since by now I'm so used to the fact the once something is "gone" from `Target`, we remove it from `Plan`. I would've expected the precondition to be more like `isEmpty`, because we want to remove the collection from `Plan` only if it is no longer in `Target`.
This is more of a question than a strong opinion, as I'm not very familiar with the supervision behind collection groups.
Other than that, I think the PR looks good!

apetenchea (approved):

arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
maierlars: I understand. If we decide to drop a collection from plan, it is already gone from Target. It is undefined behavior if we add a collection with the same ID to the target again. We might not notice it, we might have already deleted the data.

apetenchea: _If we decide to drop a collection from plan, it is already gone from Target._ - I agree.
But, the precondition:
```
.isNotEmpty(basics::StringUtils::concatT(
                  "/arango/Target/CollectionGroups/", database, "/", gid.id()))
```
Doesn't this translate to: "Drop the collection from plan, only when the Target entry is not empty"?

PR #18580, opened 2023-04-04
[GORDO-1519] Re-parallelise graph loading

markuspf:
### Scope & Purpose

Load graph in parallel shard by shard wrapped in a future.

### DIFF
--- a/arangod/Pregel/Worker/Worker.cpp
+++ b/arangod/Pregel/Worker/Worker.cpp
@@ -166,7 +166,21 @@ void Worker<V, E, M>::_initializeMessageCaches() {
 // @brief load the initial worker data, call conductor eventually
 template<typename V, typename E, typename M>
 void Worker<V, E, M>::setupWorker() {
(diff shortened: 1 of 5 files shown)


arangod/Pregel/Worker/Worker.cpp
```diff
@@ -166,7 +166,21 @@ void Worker<V, E, M>::_initializeMessageCaches() {
 // @brief load the initial worker data, call conductor eventually
 template<typename V, typename E, typename M>
 void Worker<V, E, M>::setupWorker() {
-  std::function<void()> finishedCallback = [self = shared_from_this(), this] {
+  // initialization of the graphstore might take an undefined amount
+  // of time. Therefore this is performed asynchronously
+  TRI_ASSERT(SchedulerFeature::SCHEDULER != nullptr);
```
jvolmer, line 171: I don't see where you use the scheduler here, so this is probably not needed.
Also, I don't see to which code the comment above relates.

jvolmer (approved):
Noticed two things, not sure if they are relevant:

PR #18581, opened 2023-04-04
[GORDO-1599] Re-parallelise graph storing

markuspf:
### Scope & Purpose

This PR moves the responsibility for storing a `Magazine` into a storer, as this seems to be the cleaner interface.

We then store concurrently per shard in the `Magazine`.


jvolmer (approved):
LGTM

PR #18583, opened 2023-04-05
testing: on the client we now only have the global instanceManager object anymore.

dothebart:
### Scope & Purpose

the runInClient implementation dropped to serialize the instance manager.
Hence tests must now use the `global.instanceManager` to work with instances.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #18584, opened 2023-04-05
disable this test on mac, the patterns won't match

dothebart:
### Scope & Purpose

this test will not work on mac.

- [x] :hankey: Bugfix


js/client/modules/@arangodb/testutils/test-utils.js
```diff
@@ -166,12 +166,17 @@ function filterTestcaseByOptions (testname, options, whichFilter) {
...
     return false;
   }
-  
+
   if (testname.indexOf('-nonwindows') !== -1 && platform.substr(0, 3) === 'win') {
     whichFilter.filter = 'non-windows';
     return false;
   }
 
+  if (testname.indexOf('-nonmac') !== -1 && platform.substr(0, 3) === 'win') {
```
jsteemann, line 175: why does the `nonmac` filter on platforms that are reporting to be Windows?
Should it be `&& platform.substr(0, 6) === 'darwin')` or similar?

js/client/modules/@arangodb/testutils/test-utils.js
```diff
@@ -166,12 +166,17 @@ function filterTestcaseByOptions (testname, options, whichFilter) {
...
+    whichFilter.filter = 'non-windows';
```
jsteemann, line 176: IMHO should rather be named "non-mac"
```suggestion
    whichFilter.filter = 'non-windows';
```

KVS85 (approved):

PR #18587, opened 2023-04-05
[GORDO-1647] Switch on actors per default in pregel

jvolmer:
With this PR, pregel tests will run using actors instead of the old pregel implementation. This PR is used to fix all the red tests when using actors.


MBkkt (changes requested):
merge devel to your branch probably wasn't correct

PR #18590, opened 2023-04-05
[testing]: renice the debugger so it doesn't steal resources from other parallel testruns

dothebart:
### Scope & Purpose

debuggers use lots of system resources. Give the OS-Scheduler the hint, that we don't want it to steal resources from other testruns.

- [x] :hankey: Bugfix


lib/Basics/process-utils.cpp
```diff
@@ -1565,3 +1567,29 @@ void TRI_ShutdownProcess() {
...
+  }
+#else
+  HANDLE processHandle = OpenProcess(PROCESS_SUSPEND_RESUME, FALSE, pid._pid);
+  DWORD dwPriorityClass;
+  if (prio == 0) {
+    dwPriorityClass = NORMAL_PRIORITY_CLASS;
+  } else {
+    dwPriorityClass = IDLE_PRIORITY_CLASS;
+  }
+  SetPriorityClass(hCurrentProcess, dwPriorityClass);
```
mpoeter, line 1592: In order to use SetPriorityClass you have to open the handle with `PROCESS_SET_INFORMATION` instead of `PROCESS_SUSPEND_RESUME`.

mpoeter (changes requested):

mpoeter (approved):

PR #18593, opened 2023-04-05
add possibility to provide Arguments to cluster instances similar as to the starter

dothebart:
### Scope & Purpose

in clusters one may want to divert log facilities into individual files. Use similar mechanism as the starter provides

- [x] :pizza: New feature


neunhoef (approved):
LGTM

PR #18607, opened 2023-04-06
Drain all syncer queues.

maierlars:
### Scope & Purpose
There are four different queues in the ClusterInfo for all four versions of `waitFor(Plan|Current)[Version]`. Only two of them are drained on shutdown, which could lead to shutdown hangers in threads that wait synchronously on those futures (for example a maintenance thread in synchronize shard).

This PR drains all four queues.


jsteemann (approved):
LGTM

Dronplane (approved):
LGTM

PR #18610, opened 2023-04-06
[PREGEL] Make pregel API again backwards compatible to 3.10

jvolmer:
We made the pregel API in devel more restrictive because some values did not at all make sense to use (i.e.g you should either give a graph OR vertex and edge collections where pregel should run on, not both).
But we don't want any changes in the API in 3.11, therefore this is reverted with this PR.


hkernbach (approved):
LGTM :')

markuspf (approved):
Small deletion of `LOG_DEVEL`; I am wondering whether one should pull this ugly parsing code into its own function and then put the suggestion in the comment; then again the function is itself not much more than the parsing :/

KVS85 (commented):
Tests must be added in a separate PR for this API in general and then backported to stable branches.

PR #18616, opened 2023-04-10
[GORDO-1619] Fix parallelism for processing vertices

markuspf:
### Scope & Purpose

The parallelism parameter was ignored for processing vertices and one future per shard is started per DBSServer. This could be considered a bug.

With this PR one `parallelism` many futures are started and work on shards one after the other.

As a pre-step to this, all code pertaining to processing vertices is moved out of the `Worker` struct into its own `VertexProcessor` struct.

### DIFF
--- a/arangod/Pregel/Iterators.h
+++ b/arangod/Pregel/Iterators.h
@@ -23,6 +23,8 @@
 
 #pragma once
 
+#include <cstddef>
+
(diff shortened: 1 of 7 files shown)


jvolmer (approved):
LGTM, this is so much better than before. You just wanted to add canceling the computation in between.

PR #18617, opened 2023-04-10
[GORDO-1621] Respect parallelism while storing results

markuspf:
### Scope & Purpose

Re-respect the parallelism parameter while storing results of a Pregel computation  back into the database.


jvolmer (approved):
LGTM

PR #18618, opened 2023-04-10
[GORDO-1620] Respect parallelism for graph loading

markuspf:
### Scope & Purpose

Re-respect the parallelism parameter when loading a graph from the database into the Pregel system.


jvolmer (approved):
LGTM

PR #18636, opened 2023-04-11
Using ASSERT_EQ in DurationTest

apetenchea:
### Scope & Purpose

Although with very low probability, the *DurationTest* test has been [seen failing on the mac](https://jenkins01.arangodb.biz/job/arangodb-ANY-mac-matrix.arm64/290/EDITION=enterprise,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=mac&&test&&arm64/artifact/testfailures.txt).
```
      "test" failed: /Users/jenkins/hw33-mac-m1/oskar/work/ArangoDB/tests/Pregel/DurationTest.cpp:46
Expected: (moreElapsed) > (elapsed), actual: 8-byte object <A3-F9 B2-A0 6F-8C 66-3E> vs 8-byte object <A3-F9 B2-A0 6F-8C 66-3E>
```

This looks like the test ran so fast that the time turned out to be the same when comparing `moreElapsed` to `elapsed`. ~~I added a few timeouts to prevent this from happening~~. I replaced `ASSERT_GT` with `ASSERT_GE`.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/tests/Pregel/DurationTest.cpp
+++ b/tests/Pregel/DurationTest.cpp
@@ -39,8 +39,12 @@ TEST(DurationTest, test_duration) {
   ASSERT_TRUE(d1.hasStarted());
   ASSERT_FALSE(d1.hasFinished());
 
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
   auto elapsed = d1.elapsedSeconds();
   ASSERT_GT(elapsed.count(), 0);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
 
   auto moreElapsed = d1.elapsedSeconds();
   ASSERT_GT(moreElapsed, elapsed);
@@ -49,9 +53,13 @@ TEST(DurationTest, test_duration) {
   ASSERT_TRUE(d1.hasStarted());
   ASSERT_TRUE(d1.hasFinished());
 
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
   auto evenMoreElapsed = d1.elapsedSeconds();
 
   ASSERT_GT(evenMoreElapsed, moreElapsed);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
 
   auto notMoreElapsed = d1.elapsedSeconds();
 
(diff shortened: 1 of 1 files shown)


jvolmer (commented):
I think it would make the test clearer to not add any of the timeouts but just exchange the ASSERT_GT in line 60 with an ASSERT_GE.

markuspf:
I am a bit surprised that this problem shows at all; for all my physics knowledge between two calls there should be some time that passes; so there might be some problems with my understanding of physics.

More likely there either is a problem with the precision of the conversion between the time-ticks and a double for the `Seconds` or optimisations that cause a steady clock to not in fact be (strictly) monotonous.

An exciting question (not that exciting) is thus whether clocks (or duration in seconds) can go backwards as well (as we would find out in another flaky instance of this test in which we use GE instead of GT...

apetenchea:
> I think it would make the test clearer (and faster) to not add any of the timeouts but just exchange the ASSERT_GT in line 60 with an ASSERT_GE.



> I think it would make the test clearer (and faster) to not add any of the timeouts but just exchange the ASSERT_GT in line 60 with an ASSERT_GE.

Thanks for the feedback!
If we remove the timeouts, should we replace all 3 assertions with `ASSERT_GE`? If the one in line 60 failed, then they all are probably prone to this kind of "time travel" on Mac (perhaps due to some weird optimization). Either way, if we choose to go with `ASSERT_GE`, this might not catch potential bugs in the future, for example, if `elapsedSeconds` would always return the same value.
I think we have 3 possibilities, each with pros and cons:
- use `ASSERT_GE` just for the line you mentioned, but I'm afraid that might not fix it completely
- use `ASSERT_GE` everywhere instead of `ASSERT_GT`, but this might make the test less robust
- use timeouts, but this makes the test slower and less clear
So, how should I proceed? Pregel is your domain, I will gladly honor your choice.

markuspf:
The only "correct" fix for this test as it is is to use `ASSERT_GE` and not use timeout: there is no guaranteed precision for the steady_clock, so it could well just tick once a second. 

One could try working out what the preciusion is *supposed* to be but then you set yourself up for pains if the implementation or hardware doesn't honor it.

Let's remove the timeouts and use `ASSERT_GE` and move on, ok?

jvolmer (approved):

PR #18642, opened 2023-04-12
Use proper gtest macros for more visibility.

maierlars:
### Scope & Purpose
Instead of just showing `false is not true`, please show what the _wrong_ value was. Maybe its just of by a little.


neunhoef (approved):
LGTM

PR #18640, opened 2023-04-11
use local foxx zip to reduce test flakyness

dothebart:
### Scope & Purpose

don't lean on internet download for this test.

- [x] :hankey: Bugfix

PR #18643, opened 2023-04-12
use local foxx zip to reduce test flakyness

dothebart:
### Scope & Purpose

don't lean on internet download for this test.

- [x] :hankey: Bugfix

replace https://github.com/arangodb/arangodb/pull/18640 which had more errnous commits


jsteemann (approved):
LGTM

PR #18644, opened 2023-04-12
Disable crash handler tests on mac

apetenchea:
### Scope & Purpose

One of the recovery tests fails constantly on macOS, in the [nightly builds](https://jenkins01.arangodb.biz/job/arangodb-ANY-mac-matrix.arm64/290/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=mac&&test&&arm64/artifact/testfailures.txt).

The feedback I got is that the test should not be ran on macOS in the first place, so I added `nonmac` in the file name.
While working on this test failure, I observed that we tried to de-reference a nullptr in order to produce a SIGSEGV. This is undefined behavior, and at least in theory is not guaranteed to produce a SIGSEGV. For example, on the macOS, it results in SIGABRT. As the test depends on a regex looking for SIGSEGV explicitly, I think it's better to just raise the desired signal ourselves.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


jsteemann (approved):
LGTM

PR #18645, opened 2023-04-12
Fix race in kill queries test

mpoeter:
### Scope & Purpose

The first server to simulate the query abortion reports back to the coordinator, which will then send abort requests to all db servers. These abort requests then race with the previous use requests. If the abort request is processed before the use request, the assertions we had in the code did not hold.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**


mchacki (approved):
LGTM

maierlars (approved):

PR #18630, opened 2023-04-11
[CINFRA] Fix replication2 cppcheck

apetenchea:
### Scope & Purpose
Fixing cppcheck issues

- https://jenkins01.arangodb.biz/job/arangodb-devel-cppcheck/lastBuild/cppcheck/folder.165655774/source.6b4e75e9-7581-4f08-8b41-40ffe47d9ed6/#262
- https://jenkins01.arangodb.biz/job/arangodb-devel-cppcheck/lastBuild/cppcheck/folder.288020395/source.fecf40e8-8b97-4545-8016-3cf561259fb3/#72

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

PR #18649, opened 2023-04-12
[CINFRA] Use after move

apetenchea:
### Scope & Purpose

The string `name` passed to `assertWasInserted` after being moved in the `try_emplace` above.

Provides one of the fixes from https://github.com/arangodb/arangodb/pull/18630 

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


goedderz (approved):

joerg84:
@apetenchea currently this is against your staging branch, should we have that directly to devel (i.e., 3.11 as well)

apetenchea:
> @apetenchea currently this is against your staging branch, should we have that directly to devel (i.e., 3.11 as well)

Yes, I did that. There is another PR to devel, which contains this change, among others: https://github.com/arangodb/arangodb/pull/18649#issuecomment-1505312753

PR #18657, opened 2023-04-12
Inserts that generate the key on the dbserver should not be retried.

maierlars:
### Scope & Purpose
If a database server fails and the connection is cancelled on the coordinator, the coordinator attempts a retry. In the case of an insert into a collection for which keys are generated on the dbserver (e.g. numberOfShards=1) a retried request can not be distinguish from a genuine new request.

This results in double documents. 

If the keys are generated on the coordinator the second request might fail with a unique key constraint violation.


apetenchea (approved):

jsteemann (approved):
LGTM

PR #18664, opened 2023-04-12
[PREGEL] Add pregel api tests for starting pregel

jvolmer:
Adds api tests for starting a pregel run, required by already merged PR https://github.com/arangodb/arangodb/pull/18610


KVS85 (changes requested):
Need detailed response assertion: only return code isn't enough.

KVS85 (approved):

PR #15385, opened 2021-12-20
Views Web UI

adityamukho:
**Note:** This PR is **large**, since during development of this feature, a lot of UI components that were common to both Analyzers and Views were created/updated. Unfortunately, this led to unavoidable changes to several files in the Analyzers Web UI section, that otherwise have no functional bearing on this feature.

The functionality of the Analyzers UI should remain unaffected, since from the perspective of that section, this PR only introduces functionally invariant code refactors.

### Scope & Purpose
This PR enhances the 'Views' section of the AARDVARK UI for ArangoDB, making it feature-complete, i.e. every field/property of an ArangoSearch View can now be configured through forms in the UI, in addition to the already existing method of editing the raw JSON representation. This includes the ability to configure _View Links_, with full support for their nested/recursive structures. Additionally, the JSON form has been improved with a schema validator, which informs the user of any errors in their JSON input.

- [ ] :hankey: Bugfix (requires CHANGELOG entry)
- [x] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification
- [ ] :book: CHANGELOG entry made

#### Backports:

- [ ] Backport for 3.9: *(Please link PR)*
- [ ] Backport for 3.8: *(Please link PR)*
- [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket number: https://arangodb.atlassian.net/browse/APM-116
- [x] Design document: (OUTDATED) https://docs.google.com/document/d/1YvKhbQlvZG22hPjuZpZx5MUGSfymrClrzZ_3P7DkElY/edit

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [x] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [ ] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [ ] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

[...]

PR #18670, opened 2023-04-13
FE-216: fix - make view patches async again

palashkaria:
### Scope & Purpose

This uses the jobs API to make view updates async again (which got broken in https://github.com/arangodb/arangodb/pull/15385)

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-216


js/apps/system/_admin/aardvark/APP/react/src/views/views/arangoSearchView/usePatchArangoSearchView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const result = await route.put(`/view/${encodedOldViewName}/rename`, {
+    name: view.name.normalize()
+  });
+
+  if (result.body.error) {
+    window.arangoHelper.arangoError(
+      "Failure",
+      `Got unexpected server response: ${result.body.errorMessage}`
+    );
+    error = true;
```
nikhil-varma, line 53: Out of curiosity, would returning the error itself make it more useful than just a `boolean` ? Maybe we could consider returning the error?

nikhil-varma: Something like this (_this is something that I recently started following_)

```
const {response, error} = await someAPICall();

if(error) {
 // TODO
}

```

nikhil-varma (approved):
Leaving a suggestion for future. Doesn't need it to be now. Thought it could be a useful pattern and if agreed upon, can be followed app-wide 🙂 
Otherwise all OK.

js/apps/system/_admin/aardvark/APP/react/src/views/views/arangoSearchView/usePatchArangoSearchView.tsx
palashkaria: we are using this boolean to stop the next API call (after rename). The error is not required anywhere as we are already showing it in a notification 

Also, this was existing code in search alias view, I just moved the code around; Can implement your suggestion across both views when we consolidate it! Will keep track

js/apps/system/_admin/aardvark/APP/react/src/components/hooks/useJobSync.ts
```diff
@@ -0,0 +1,63 @@
...
+            .then(data => readJob(data, job.id))
+            .catch(error => onJobError(error, job.id));
+        }
+      });
+    }
+  };
+  useEffect(() => {
+    window.arangoHelper.getAardvarkJobs(checkState);
+
+    let interval = window.setInterval(() => {
```
pluma, line 56: @palashkaria consider https://overreacted.io/making-setinterval-declarative-with-react-hooks/

pluma (approved):
LGTM but see the caveat about setInterval.

KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/28352/

PR #18677, opened 2023-04-13
Updated test www-authenticate

cpjulia:
### Scope & Purpose

This PR aims to fix a test that was failing on jenkins because it's expected to have code 401 returned instead of code 200.
It's attempted to test that the user is unauthorized after a password change by changing it ad making an HTTP request to `/_api/version`, which should return code 401 but doesn't because the password change didn't happen successfully before.
Test is `tests/js/client/server_parameters/www-authenticate.js`
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-windows/21074/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=cluster,limit=windows&&test&&!rta&&!release/artifact/testfailures.txt

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
maybe

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


tests/js/client/server_parameters/www-authenticate.js
```diff
@@ -62,33 +61,46 @@ function HttpAuthenticateSuite() {
...
-        users.update(user, "");
-        arango.reconnectWithNewPassword("");
+        let result = "";
+        // we put this in a try to make it sequential, because the request might be made when the password has not
+        // changed yet so it wouldn't get a 401
+        try {
+          // need to change password, otherwise, the request will return 200
+          result = users.update(user, "abc");
+        } finally {
+          assertEqual(result.code, 200, "update password returned with code " + result.code);
```
jsteemann, line 72: I don't think this is a good solution.
A `finally` block should never have any test assertions.
The purpose of a `finally` block is to clean up after success or failure, no matter what.
It does not make much sense to put assertions into it, because if they fail, the rest of the `finally` block isn't executed and thus the cleanup is not complete. It is like a destructor in C++, which you also don't want to throw its own exceptions.

jsteemann (changes requested):
There shouldn't be any assertions in `finally` blocks.

PR #18678, opened 2023-04-14
Reset log level to normal on agents after startup.

maierlars:
### Scope & Purpose
Set the log level to info after startup. It was set to debug to investigate the startup behavior of agents.


neunhoef (approved):
LGTM

jsteemann (approved):
LGTM

PR #18681, opened 2023-04-14
Asserting on doubles on macOS

apetenchea:
### Scope & Purpose

I have seen this test failing on macOS: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-mac-onlycatch.x86-64/24288/EDITION=enterprise,limit=mac&&test&&x86-64/artifact/testfailures.txt
This is not the first time we have had such problems with floating point precision on macOS.

I introduced `EXPECT_DOUBLE_EQ` for comparing doubles and replaced `ASSERT` with `EXPECT` where possible, such that the test doesn't completely stop on error.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/tests/Cache/PlainCache.cpp
+++ b/tests/Cache/PlainCache.cpp
@@ -361,10 +361,10 @@ TEST(CachePlainCacheTest, test_hit_rate_statistics_reporting) {
   {
     auto cacheStats = cacheHit->hitRates();
     auto managerStats = manager.globalHitRates();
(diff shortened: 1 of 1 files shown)


apetenchea:
Jenkins failure is unrelated (exclusive-collections-off.js)

jsteemann (approved):
LGTM

maierlars (approved):

PR #18500, opened 2023-03-29
Feature 3.10/sent time accounting

neunhoef:
### Scope & Purpose

This PR is used to diagnose some network problems on the ArangoGraph platform. It adds logging in case there is some abnormal delay in sending cluster internal requests. The following time points are stored for every request:

 - time when the request was submitted to fuerte
 - time when the request was pulled from the input queue and `asio::asyncWrite` was started
 - time when the request was sent away (`asio::asyncWrite` issued its callback)

The PR adds logging if there is any abnormal behaviour with respect to the above times, or if the sender had to wait too long after sending the request away until the response was received.

Furthermore, there are metrics showing bad behaviour.

- [*] :pizza: New feature

Finally, the timeouts for cluster internal commit and abort operations (el cheapo and streaming transactions) have been increased from 90s to 14 minutes for coordinators talking to leaders and to 13 minutes for dbservers talking to their followers.

### Checklist

- [*] :book: CHANGELOG entry made
- [*] :books: documentation written (release notes, API changes, ...)
- [*] Backports
  - [*] Backport for 3.10: This is the 3.10 port.

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [*] devel PR: https://github.com/arangodb/arangodb/pull/18683

PR #18683, opened 2023-04-14
Log delays in sending, adjust commit timeouts.

neunhoef:
### Scope & Purpose

This PR is used to diagnose some network problems on the ArangoGraph platform. It adds logging in case there is some abnormal delay in sending cluster internal requests. The following time points are stored for every request:

 - time when the request was submitted to fuerte
 - time when the request was pulled from the input queue and `asio::asyncWrite` was started
 - time when the request was sent away (`asio::asyncWrite` issued its callback)

The PR adds logging if there is any abnormal behaviour with respect to the above times, or if the sender had to wait too long after sending the request away until the response was received.

Furthermore, there are metrics showing bad behaviour.

- [*] :pizza: New feature

Finally, the timeouts for cluster internal commit and abort operations (el cheapo and streaming transactions) have been increased from 90s to 14 minutes for coordinators talking to leaders and to 13 minutes for dbservers talking to their followers.

### Checklist

- [*] Backports
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18500
  - [*] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18684


3rdParty/fuerte/src/GeneralConnection.h
```diff
@@ -103,6 +103,7 @@ class GeneralConnection : public fuerte::Connection {
   // Start an asynchronous request.
   void sendRequest(std::unique_ptr<Request> req, RequestCallback cb) override {
     // construct RequestItem
+    req->setTimeReceived();
```
jsteemann, line 106: Nit: although this is correct, the wording really confused me.
Because the request is not received here in the sense of Request -> Response (which could be received), but rather the request is received by the fuerte subsystem for processing.
I suggest to rename this to `setTimeProcessingStarted()` or `setTimeStarted()` or something.

arangod/Network/NetworkFeature.cpp
```diff
@@ -327,15 +353,116 @@ void NetworkFeature::sendRequest(network::ConnectionPool& pool,
...
+                     .count()
+              << " seconds, endpoint: " << endpoint
+              << ", request ptr: " << (void*)req.get()
+              << ", response ptr: " << (void*)res.get()
+              << ", error: " << uint16_t(err);
+          _delayedSends.count();
+        }
+        dur = std::chrono::steady_clock::now() - req->timeSent();
+        if (dur > std::chrono::seconds(61)) {
+          LOG_TOPIC("effc5", WARN, Logger::COMMUNICATION)
```
jsteemann, line 449: I think logging a warning here is too harsh, because it will log a warning for requests that are known to be taking long (e.g. requests for heavyweight AQL queries etc.). I think we should downgrade the log level here from warning to info, because it *can* happen.

maierlars (approved):

jsteemann (approved):
LGTM

PR #18685, opened 2023-04-14
Wait for the supervision to actually kill the job.

maierlars:
### Scope & Purpose
Test does not wait for the supervision to react. Sometimes its just to fast and the supervision had no chance to kill the job.


jsteemann (approved):
LGTM

PR #12023, opened 2020-06-29
Bug fix/ui css embedded document editor

hkernbach:
### Scope & Purpose

* Removed the edge id's hover styling in case of embedded document editor in 
  the Web UI as this functionality is disabled. This was misleading because
  the elements are not clickable (see screenshot below).

- [x] Bug-Fix for *devel-branch*
Backports needed for all supported versions.

#### Related Information

- [x] There is an internal planning ticket: https://arangodb.atlassian.net/browse/BTS-98

### Testing & Verification

This change is a trivial rework / code cleanup without any test coverage.

image

PR #18686, opened 2023-04-14
[PREGEL] Use number of cores as upper cut for pregel parallelism

jvolmer:
Fixes "Error: at assertion #12023: assertEqual: (6) is not equal to (8) (Error at tests/js/common/shell/shell-pregel-multicollection-edgecol.js:242:9 at Object.testMultiWCCParallelism (tests/js/common/shell/shell-pregel-multicollection-edgecol.js:234:22) at tests/js/common/shell/shell-pregel-multicollection-edgecol.js:365:9 .... at Object.shellClient [as shell_client] (/Users/jenkins/hw43-mac/oskar/work/ArangoDB/js/client/modules/@arangodb/testsuites/aql.js:208:79) - test failed"
Happening on mac, e.g.: "https://jenkins01.arangodb.biz/job/arangodb-ANY-mac-matrix.x86-64/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=mac&&test&&x86-64/4292/"

If the requested parallelism is larger than the number of cores on the running machine, pregel uses the number of cores for parallelism. Some Mac systems we run Jenkins on have only 6 cores, therefore the check on the parallelism of a pregel run that was started with 8 parallelism failed.
The test now executes a different assertion depending on the number of cores (which we get via the /_admin/metrics endpoint).

### DIFF
--- a/js/server/modules/@arangodb/test-helper.js
+++ b/js/server/modules/@arangodb/test-helper.js
@@ -135,8 +135,8 @@ exports.getMetric = function (endpoint, name) {
 };
 
 exports.getMetricSingle = function (name) {
-  let res = arango.GET_RAW("/_admin/metrics");
-  if (res.code !== 200) {
+  let res = request.get("/_admin/metrics");
+  if (res.status !== 200) {
     throw "error fetching metric";
   }
   return getMetricName(res.body, name);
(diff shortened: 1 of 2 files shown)


hkernbach (approved):
LGTM. 

Minor suggestion. Move the method `number_of_cores_on_machine` to a dedicated "helper file / class". Might be helpful for other tests as well.

KVS85 (changes requested):
This breaks `testMultiWCCParallelism` (see Jenkins run).

KVS85 (approved):

PR #18689, opened 2023-04-14
improve efficiency of agency poll operation

jsteemann:
### Scope & Purpose

instead of copying the to-be-returned values, write them into the result directly.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Agency/State.cpp
```diff
@@ -1864,13 +1894,15 @@ uint64_t State::toVelocyPack(index_t lastIndex, VPackBuilder& builder) const {
 /// @brief dump the entire in-memory state to velocypack.
 /// should be used for testing only
 void State::toVelocyPack(velocypack::Builder& builder) const {
-  std::lock_guard mutexLocker{_logLock};
-
   builder.openObject();
   builder.add("current", VPackValue(_cur));
```
MBkkt, line 1898: Is it safe to read `_cur` without lock?

arangod/Agency/Agent.cpp
```diff
@@ -1008,44 +1007,39 @@ std::tuple<futures::Future<query_t>, bool, std::string> Agent::poll(
...
-  std::lock_guard guard(_promLock);
+  if (builder != nullptr) {
+    return std::tuple<futures::Future<query_t>, bool, std::string>{
+        futures::makeFuture(std::move(builder)), true, std::string()};
+  }
+
   auto tp = steady_clock::now() +
             duration_cast<milliseconds>(duration<double>(timeout));
 
+  std::lock_guard guard(_promLock);
```
MBkkt, line 1041: Do you move get current time before lock intentionally?
Behavior is different

arangod/Agency/State.cpp
jsteemann: No, it's not safe. I overlooked the access to `_cur` here. Good catch!

arangod/Agency/Agent.cpp
jsteemann: Yes, it was intentional. The behavior can be different, but I don't think it makes a relevant difference if the time point is taken while under the lock or before. `_promises` is a multimap and thus sorted, and the only other place accessing `_promises` also looks as if it doesn't care about when the timepoint was exactly taken.

MBkkt (approved):

PR #18709, opened 2023-04-18
[PREGEL] Fix load-balancing-pregel-auth-cluster test

jvolmer:
Should fix https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-windows/21263/EDITION=community,STORAGE_ENGINE=rocksdb,TEST_SUITE=cluster,limit=windows&&test&&!rta&&!release/testReport/junit/EE_load_balancing_auth__tests_js_client_load-balancing_load-balancing-pregel-auth-cluster/js/EE_testPregelForwardingSameUser/

We now write the correct user into the historic pregel collection. Before, we wrote `ExecContext::current().user()` into the collection, which first can be null depending on which thread this is executed on (this null gave us the occasional bug) and second is not correct: We want to always write the user that started the pregel run.

Additionally, the test is improved: In the test we now make sure that the pregel run was created before we request its status, same for canceling the run.

### DIFF
--- a/tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
+++ b/tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
@@ -129,7 +129,6 @@ function PregelAuthSuite () {
     },
 
     testPregelForwardingSameUser: function() {
-      let url = baseUrl;
       const task = {
         algorithm: "pagerank",
         vertexCollections: ["vertices"],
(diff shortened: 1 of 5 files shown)


tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
```diff
@@ -138,39 +137,50 @@ function PregelAuthSuite () {
...
+      const postResponse = sendRequest(users[0], 'POST', baseUrl, task, true);
+      assertFalse(postResponse === undefined || postResponse === {});
+      assertEqual(postResponse.status, 200);
+      const pid = postResponse.body;
+      assertTrue(pid > 1);
+
+      // check that pregel status of run can be retrieved
+      // need to wait for short time until pregel run was created
+      const sleepIntervalSeconds = 0.2;
+      const maxWaitSeconds = 1;
```
mchacki, line 149: As this is a maximum for waiting do we want to be more generous here?
It is supposed to detect dead-locks/logic errors and will fail on very high timeouts.
Now we could get false negatives on slow machines.
Also on typical runs we should get nowhere near the timeout.

Therefore I would suggest a 10s max wait.
WDYT?

tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
```diff
@@ -179,33 +189,32 @@ function PregelAuthSuite () {
...
-      assertFalse(result.body.error);
-      assertEqual(result.status, 200);
+      const postResponse = sendRequest(runCreator, 'POST', baseUrl, task, true);
+      assertFalse(postResponse === undefined || postResponse === {});
+      assertEqual(postResponse.status, 200);
+      const pid = postResponse.body;
+      assertTrue(pid > 1);
+
+      // need to wait for short time until pregel run was created
+      require("internal").sleep(1);
```
mchacki, line 199: Should we do the same retry loop here?
On slow/full Jenkins we may still see false negatives here just because we could not get CPU time.

mchacki (approved):
LGTM.

I just have two suggestions to avoid sporadic test failures on very slow testing systems.

tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
mchacki: ah true
Mhm
Can we do the following:
We do the waiting loop on the correct user until it shows up (if this fails the test above needs to be red)
And then continue with the wrong user and expect it to be not found?

jvolmer: so true, thanks!

tests/js/client/load-balancing/load-balancing-pregel-auth-cluster.js
```diff
@@ -138,39 +137,50 @@ function PregelAuthSuite () {
...
+          assertTrue(false, "Pregel run was not canceled successfully but is in state " + statusResponse.body.state);
+          return;
+        }
+      } while (!pregelTestHelpers.runCanceled(statusResponse.body));
     },
 
     testPregelForwardingDifferentUser: function() {
-      let url = baseUrl;
+      const runCreator = users[0];
+      const differentUser = users[1];
```
hkernbach, line 183:
```suggestion
      const runUser = users[0];
      const differentUser = users[1];
```

Very very minor suggestion :P

hkernbach (approved):

PR #18721, opened 2023-04-19
avoid startup warnings on ARM

jsteemann:
### Scope & Purpose

In case /proc/cpu/alignment does not exist, there is no need to complain about it.
Inspired by the following log output from an ARM build:
```
2023-04-19T21:12:15Z [1] WARNING [b8a20] {memory} unable to detect CPU alignment settings. could not process file '/proc/cpu/alignment'. this may cause arangod to abort with SIGBUS. it may be necessary to set the value in '/proc/cpu/alignment' to 2
```
Now, if /proc/cpu/alignment is not present, we don't bother to log this warning.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RestServer/EnvironmentFeature.cpp
+++ b/arangod/RestServer/EnvironmentFeature.cpp
@@ -167,40 +167,45 @@ void EnvironmentFeature::prepare() {
 
     std::string const filename("/proc/cpu/alignment");
     try {
-      std::string const cpuAlignment =
-          arangodb::basics::FileUtils::slurp(filename);
-      auto start = cpuAlignment.find("User faults:");
+      if (basics::FileUtils::exists(filename)) {
+        std::string const cpuAlignment = basics::FileUtils::slurp(filename);
+        auto start = cpuAlignment.find("User faults:");
 
-      if (start != std::string::npos) {
-        start += strlen("User faults:");
-        size_t end = start;
-        while (end < cpuAlignment.size()) {
-          if (cpuAlignment[end] == ' ' || cpuAlignment[end] == '\t') {
-            ++end;
-          } else {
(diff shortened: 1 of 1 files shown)


dothebart:
can this alignment info potentially be found in the cpuinfo file as well? or is it safe that its disabled if the file isn't there?

jsteemann:
> can this alignment info potentially be found in the cpuinfo file as well? or is it safe that its disabled if the file isn't there? Maybe the file was moved in later kernel versions and a second path should be tried?

You mean in `/proc/cpuinfo`? I don't think so, because it has a different purpose (showing CPU capabilities rather than kernel configuration).
The purpose of this PR is to avoid logging of unnecessary warning messages if /proc/cpu/alignment is not present.
If I understand correctly, this file is not present on arm64 (https://github.com/raspberrypi/linux/issues/2283).

Suppressing an unnecessary warning sounds reasonable, especially because a user can do nothing to fix the issue.
Should the fix be there for whatever reason, it will still be picked up and processed as normal.

cpjulia (approved):
LGTM

dothebart:
`alignment trap is not available on any arm64 kernel.` sounds like we could delete it all together, since we don't support 32 bit systems?

PR #18723, opened 2023-04-20
FE-238 - auto-login support on core web UI

palashkaria:
### Scope & Purpose

This adds measures to handle auto-login support on the web UI:
1. Logout button is hidden is auto-login via ArangoGraph is enabled
3. We were using `jwtUser` set in sessionStorage during login. Since login is now skipped, we have switched to using `/whoAmI` response to add that value

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-238


palashkaria:
Moving back to draft as there might be some messaging changes

js/apps/system/_admin/aardvark/APP/frontend/js/templates/userBarView.ejs
```diff
@@ -16,35 +18,15 @@
...
+    <% if (!ssoEnabled) {%>
+      <li class="dropdown-footer">
+        <button id="userLogout" class="button-primary btn-small pull-right">Logout</button>
       </li>
-      <li class="dropdown-item">
-        <a id="userLogout" class="tab">Logout</a>
+    <% }%>
+    <% if (ssoEnabled) {%>
+      <li class="dropdown-footer" style="max-width: 200px; line-height: 1.4; padding: 0 10px !important;">
+        To log out of this DB, log out of the ArangoGraph.
```
ansoboleva, line 28: Please change the message to "This login is automatically managed by ArangoGraph."

ansoboleva (changes requested):
please see my comment

js/apps/system/_admin/aardvark/APP/frontend/js/views/userBarView.js
```diff
@@ -104,11 +104,13 @@
               }
 
               self.$el = $('#userBar');
+              var autoLoginEnabled = arangoHelper.getAutoLoginEnabled(autoLoginEnabled);
```
nikhil-varma, line 107: We can use `const` ?

nikhil-varma (approved):

KVS85 (changes requested):
@ansoboleva @palashkaria

What will happen if there is a non-AG deployment with this header being set but not JWT token is provided?

What will happen to a a non-AG deployment with this header being set but no authentication enabled?

palashkaria:
> @ansoboleva @palashkaria
> 
> What will happen if there is a non-AG deployment with this header being set but not JWT token is provided?
> 
> What will happen to a a non-AG deployment with this header being set but no authentication enabled?


@KVS85 

- If this header is set, and you are logged in, you will see this:
image (3)

- If the header is set, and you are not logged in, you will be asked to log in.

This header is not dealing with authentication, it is only used to show/hide the logout button. The authentication (& setting of this header) is handled by the auth middleware on the AG side.

KVS85:
- [ ] https://jenkins.arangodb.biz/job/arangodb-ANY-jslint/28594/

PR #16471, opened 2022-06-29
Feature/parallelize moar traversals

hkernbach:
### Scope & Purpose

This PRs adds parallel traversal execution in cluster mode.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [x] Backports (no backports)

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/957
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #18729, opened 2023-04-20
[BTS-1325] Don't go to sleep forever

goedderz:
## Scope & Purpose

https://arangodb.atlassian.net/browse/BTS-1325

Fix spurious nightly failures in `aql-satellite-graph-traversal-generic.js` and `aql-disjoint-smart-graph-traversal-generic.js`.

The constellation running into the bug is two parallel traversals, plus a remote node, on a DBServer snippet. This can be conceptually reduced to two AsyncNodes (stacked, not parallel) plus the Remote.

This is a regression that seems to have been introduced ~only~ by ~the combination of https://github.com/arangodb/arangodb/pull/18274 and~ https://github.com/arangodb/arangodb/pull/16471.

The problem occurs when the upper async node (in a scheduled thread) stores a result it got from the remote, then wakes up the main query. The lower async is currently in progress, putting the main query back to sleep. When the lower async node’s thread finishes, it wakes the main query up again. However, if the lower node has a “WAITING” result stored, the query will now go to sleep forever.

### Simplified example

Assume [at least] the following nodes on a DBServer snippet:

```
 10 RemoteNode
 ...
 20 AsyncNode
 ...
 30 AsyncNode
 ...
RestHandler
```

where “RestHandler” is just a reminder that it’s the place that will start executing, or resuming, the query.

Here's an example sequence, leading to the query snippet going to sleep forever. Pretend the blue-ish box (labeled "boolean fix") isn't there for now, an explanation of that follows below.

```mermaid
sequenceDiagram

    participant RH as RestHandler
    participant A1 as AsyncNode[30]
    participant T1 as Thread1
    participant A2 as AsyncNode[20]
    participant T2 as Thread2
    participant R as RemoteNode[10]
    participant N as Network

    RH  ->>+ A1: call    
    A1 ->>+ T1: spawn
    Note over A1: state = InProgress
    A1 -->>- RH: WAITING (scheduled)
    T1  ->>+ A2: call

    A2 ->>+ T2: spawn
    Note over A2: state = InProgress
    T2  ->>+ R: call
    R ->> N: request
    
    R -->>- T2: WAITING (network)
    Note over A2,T2: store WAITING
    Note over A2: state = GotResult

    N -->>+ R: response
    R  -->>- RH: wake up
    
    activate RH
    RH  ->>+ A1: call
    A1 -->>- RH: WAITING (in progress)
    deactivate RH
    T2 -->>- RH: wake up
    activate RH

    RH  ->>+ A1: call
    A1 -->>- RH: WAITING (in progress)
    deactivate RH
    A2 -->>- T1: WAITING (scheduled)
    Note over A1,T1: store WAITING
    Note over A1: state = GotResult

    T1 -->>- RH: wake up
    activate RH


    rect rgb(191, 223, 255)
    Note over RH, T2: "boolean fix" would introduce this
    RH  ->>+ A1: call
    Note over A1: ignore stored WAITING
    A1 ->>+ T1: spawn
    Note over A1: state = InProgress
    A1 -->>- RH: WAITING (scheduled)
    deactivate RH
    T1 ->>+ A2: call
    A2 -->>- T1: WAITING (stored)
    Note over A1,T1: store WAITING
    Note over A1: state = GotResult
    T1 -->>- RH: wake up
    activate RH
    end

    RH  ->>+ A1: call
    A1 -->>- RH: WAITING (stored)
    deactivate RH
    Note over RH: #128164; forever #128164;
```

### Discussion

Each node from which a WAITING can originate (i.e. RemoteNode or AsyncNode) will wake up the RestHandler once for each WAITING (that was not just passed through). Wake-ups are synchronized with the "main" thread (starting at the RestHandler), and won't get lost.

This would guarantee liveness, but there's one exception: If a RemoteNode or AsyncNode gets a wake-up while in progress (either network request or spawned thread, respectively), WAITING is returned, but no additional wake-up is scheduled. This may seem OK, because there's still at least one wake-up to come.

However, an AsyncNode can store WAITINGs to return later. When such a stored WAITING is returned, there's no guarantee there'll be a wake-up later.

This combination needs to be avoided by any fix: either by queuing additional wake-up calls, or ignoring a (to-be) stored WAITING to account for the previously ignored wake-up.

It is not sufficient to do this only once; i.e. make a note that (at least) one call has arrived while the async node was in progress, and ignore a possibly stored WAITING after. This makes this occur less often, but doesn't prevent it from happening: The blue-ish box above labeled with "boolean fix" extends the above scenario with this patch, but also showing that it still results in the query never waking up again.

So the implemented fix counts all wake-ups happening on AsyncNodes while in progress. It is not necessary for RemoteNodes, as they don't store WAITING, but will return a result after their wake-up call is triggered. If an AsyncNode would previously have stored WAITING, it will immediately repeat its call to upstream instead, once for each call that happened while in progress.

### Transfer to real-world scenarios and tests

The example above does not exist in a real setup. Most notably, an AsyncNode is always surrounded by a MutexNode/GatherNode pair, and it always has at least one sibling. This, in combination with the fact that currently only TraversalNodes are allowed between Async- and MutexNodes, makes real-world examples more complicated to describe and follow than the one above. Additionally, the "boolean fix" as sketched above would actually be sufficient for these real-world scenarios: But it can still break if e.g. different nodes than Traversal are allowed in the future.

While a more specialized integration test for a regression test is possible (e.g. by starting with `testSmallCircleBFSParallelismSubqueryOuterLoop_smallCircle_Graph_satellite_4` from `aql-satellite-graph-traversal-generic.js` and making it more likely to fail, e.g. by introducing selective sleeps), it is complicated and will not be able to detect certain other kinds of errors (e.g. those that might be introduced by allowing parallelism for different nodes than traversal).

Therefore we opted for C++ tests instead, which sketch the simplified example as laid out above. There are two tests: One that deterministically fails for the original code, as well as with the "boolean fix". And another that has a randomized execution order, which fails for the original code in about 80% of the time, or 10% with the "boolean fix", but with the hope of being able to expose more kinds of errors as well.

The following patch was used to run the test selectively with the original behavior, or the boolean fix:

```diff
diff --git arangod/Aql/AsyncExecutor.cpp arangod/Aql/AsyncExecutor.cpp
index 9abc0ede6ed..d75902928f1 100644
--- arangod/Aql/AsyncExecutor.cpp
+++ arangod/Aql/AsyncExecutor.cpp
@@ -67,6 +67,11 @@ ExecutionBlockImpl<AsyncExecutor>::execute(AqlCallStack const& stack) {
   return res;
 }

+// t == 0: original code
+// t == 1: boolean fix
+// t == 2: fix with loop
+#define t 0
+
 std::tuple<ExecutionState, SkipResult, SharedAqlItemBlockPtr>
 ExecutionBlockImpl<AsyncExecutor>::executeWithoutTrace(
     AqlCallStack const& stack) {
@@ -97,7 +102,17 @@ ExecutionBlockImpl<AsyncExecutor>::executeWithoutTrace(
       // we may not return WAITING if upstream returned DONE
       _internalState = AsyncState::Empty;
     }
+#if t == 1
+    if (_numWakeupsQueued > 0 && _returnState == ExecutionState::WAITING) {
+      _numWakeupsQueued = 0;
+      // fall through
+    } else {
+      _numWakeupsQueued = 0;
+      return {_returnState, std::move(_returnSkip), std::move(_returnBlock)};
+    }
+#else
     return {_returnState, std::move(_returnSkip), std::move(_returnBlock)};
+#endif
   } else if (_internalState == AsyncState::GotException) {
     TRI_ASSERT(_returnException != nullptr);
     std::rethrow_exception(_returnException);
@@ -128,6 +143,7 @@ ExecutionBlockImpl<AsyncExecutor>::executeWithoutTrace(
 #endif
           }

+#if t == 2
           // If we got woken up while in progress, wake up our dependency now.
           // This is necessary so the query will always wake up from sleep. See
           // https://arangodb.atlassian.net/browse/BTS-1325
@@ -160,6 +176,7 @@ ExecutionBlockImpl<AsyncExecutor>::executeWithoutTrace(
 #endif
             }
           }
+#endif

           _returnState = state;
           _returnSkip = std::move(skip);
```

- [X] :hankey: Bugfix

## Checklist

- [X] Tests
  - [X] **Regression tests**
  - [X] C++ **Unit tests**
- [X] Backports
  - [X] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19177
  - [X] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19182


arangod/Aql/AsyncExecutor.cpp
```diff
@@ -101,11 +108,18 @@ ExecutionBlockImpl<AsyncExecutor>::executeWithoutTrace(
...
+      _sharedState->asyncExecuteAndWakeup([this, stack](bool const isAsync) {
         std::unique_lock<std::mutex> guard(_mutex, std::defer_lock);
 
         try {
           auto [state, skip, block] = _dependencies[0]->execute(stack);
+
+#ifdef ARANGODB_USE_GOOGLE_TESTS
+          if (_failureCallback) {
+            _failureCallback();
+          }
```
mpoeter, line 120: `_failureCallback` isn't really a matching name anymore. How about `postAsyncExecuteCallback` or something along those lines?

tests/Aql/AsyncExecutorTest.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  asyncBlock0->setPostAsyncExecuteCallback([&] {
+    while (!scheduler.queueEmpty()) {
+      scheduler.runOnce();
+    }
+    while (wakeupsQueued > 0) {
+      --wakeupsQueued;
+      testHelper.executeOnce();
+    }
+  });
```
mpoeter, line 81: I think this deserves a comment what we are doing here and why. Something like "the postAsyncExecuteCallback is called after the AsyncNode has called execute on its dependency, but before the results of that call is stored, so the node is still "inProgress". By making sure all all tasks an wakeups are processed before leaving the callback we simulate the thread being slow."

mpoeter (commented):
I think one more code comment would be nice, otherwise LGTM.

mpoeter (approved):

markuspf (approved):
LGTM

PR #18740, opened 2023-04-21
Do not candidate if we became follower just before.

maierlars:
### Scope & Purpose
Sometimes the agent decided to candidate, but waits on a mutex. During that time the agent is converted into a follower. This can have the strange effect that a new follower immediately candidates.


neunhoef (approved):
LGTM

PR #18747, opened 2023-04-24
Fix createCollection/loadPlan deadlock

mchacki:
### Scope & Purpose

*This is a bug-fix for an extremely rare deadlocking race, when trying to create a new Collection (or a new database as the specific code-part is shared)
This is more of a theoretical bug-fix as we so far have been unable to reproduce the exact order of events that led here, hence there is no additional test.

**UPDATE** The race condition fixed here is described in: https://arangodb.atlassian.net/browse/BTS-1335

The situation we are fixing:
1) Agencycallbacks are used to keep the Coordinator in sync with the Agency
2) If an AgencyCallback blocks, e.g. by waiting on a lock, the updating from latest Plan is delayed.
3) If above agency callback would never the the lock, the plan would never be updated again

In CreateCollectionCode this dead-lock could potentially happen:
1) The main code waits on a loadPlan to reach the version when collection creation is triggered, while holding a shared mutex.
2) The AgencyCallbacks in Current (where shard creation success is reported) acquire the shared mutex to update local report overview.
3) If for some reason the AgencyCallback waits on the Mutex, while the main code is already waiting on loadPlan, the loadPlan will never succeed, as it would only be updated by it's own internal callback.

*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cluster/ClusterInfo.cpp
```diff
@@ -122,6 +122,13 @@ struct ShardStatistics {
 
 namespace {
 
+struct CreateCollectionReport {
+  std::optional<ErrorCode> dbServerResult{std::nullopt};
+  size_t nrDone{0};
+  std::string errMsg{StaticStrings::Empty};
```
jsteemann, line 128: Why not combine `dbServerResult` and `errMsg` into an `arangodb::Result`?

MBkkt: Also it's not necessary to init `std::string` by `StaticStrings::Empty`, it's empty by default

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3346,7 +3350,8 @@ Result ClusterInfo::createCollectionsCoordinator(
 
     if (info.state == ClusterCollectionCreationState::DONE) {
       // This is possible in Enterprise / Smart Collection situation
-      (*nrDone)++;
+      report->doUnderLock(
+          [](CreateCollectionReport& report) { report.nrDone++; });
```
MBkkt, line 3354: It was atomic, why do you decide to change it to mutex?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3287,17 +3294,14 @@ Result ClusterInfo::createCollectionsCoordinator(
...
-      std::make_shared<std::atomic<std::optional<ErrorCode>>>(std::nullopt);
-  auto nrDone = std::make_shared<std::atomic<size_t>>(0);
-  auto errMsg = std::make_shared<std::string>();
-  auto cacheMutex = std::make_shared<std::mutex>();
-  auto cacheMutexOwner = std::make_shared<std::atomic<std::thread::id>>();
-  auto isCleaned = std::make_shared<bool>(false);
+  // Also note: We will use this guarded structure to protect access into state
+  // entries inside the infos list, as tehy can be modified concurrently.
+  auto report = std::make_shared<Guarded<CreateCollectionReport>>(
+      CreateCollectionReport{});
```
MBkkt, line 3304: It will call default ctor
```suggestion
  auto report = std::make_shared<Guarded<CreateCollectionReport>>();
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3813,7 +3836,11 @@ Result ClusterInfo::createCollectionsCoordinator(
       }
     }
 
-    if (nrDone->load(std::memory_order_acquire) == infos.size() &&
+    auto nrDone =
+        report->doUnderLock([](CreateCollectionReport const& report) -> size_t {
+          return report.nrDone;
+        });
```
MBkkt, line 3842: It was atomic, why do you decide to change it to mutex?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3427,15 +3437,20 @@ Result ClusterInfo::createCollectionsCoordinator(
...
-                    TRI_ERROR_CLUSTER_COULD_NOT_CREATE_COLLECTION,
-                    std::memory_order_release);
-                TRI_ASSERT(info.state != ClusterCollectionCreationState::DONE);
-                info.state = ClusterCollectionCreationState::FAILED;
+                report->doUnderLock([&p,
+                                     &info](CreateCollectionReport& report) {
+                  report.errMsg =
+                      "Error in creation of collection: " + p.key.copyString() +
+                      ". Collection already dropped. " + __FILE__ + ":" +
+                      std::to_string(__LINE__);
```
MBkkt, line 3445: You can create string before lock acquire

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3477,20 +3492,26 @@ Result ClusterInfo::createCollectionsCoordinator(
...
           // We can have multiple calls to this callback, one per leader and one
           // per follower As soon as all leaders are done we are either FAILED
           // or DONE, this cannot be altered later.
-          TRI_ASSERT(info.state != ClusterCollectionCreationState::FAILED);
-          info.state = ClusterCollectionCreationState::DONE;
-          (*nrDone)++;
+          report->doUnderLock([&info](CreateCollectionReport& report) {
+            TRI_ASSERT(info.state != ClusterCollectionCreationState::FAILED);
+            info.state = ClusterCollectionCreationState::DONE;
+            report.nrDone++;
```
MBkkt, line 3513: It was atomic, why do you decide to change it to mutex?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3813,7 +3836,11 @@ Result ClusterInfo::createCollectionsCoordinator(
       }
     }
 
-    if (nrDone->load(std::memory_order_acquire) == infos.size() &&
+    auto nrDone =
+        report->doUnderLock([](CreateCollectionReport const& report) -> size_t {
+          return report.nrDone;
+        });
+    if (nrDone == infos.size() &&
```
MBkkt, line 3843: If nrDone equal infos.size() it means info will not change.
So it's safe to touch it without lock after this condition

MBkkt (changes requested):

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3427,15 +3437,20 @@ Result ClusterInfo::createCollectionsCoordinator(
                     << "Did not find shard in _shardServers: "
```
neunhoef, line 3437: Why do we need the `_planProt` lock here anyway? As far as I see we do not access `ClusterInfo` structures here anyway. Rather, we use `shardServers`, which is local to the closure. Therefore, the error log is also wrong...

arangod/Cluster/ClusterInfo.cpp
neunhoef: Simplicity.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3729,7 +3750,6 @@ Result ClusterInfo::createCollectionsCoordinator(
       // using loadPlan, this is necessary for the callback closure to
```
neunhoef, line 3750: We should adjust this comment.

neunhoef (changes requested):
LGTM. Please add a CHANGELOG entry such that we can see when this was fixed.
I only found some outdated code comments, which needs adjusting.

arangod/Cluster/ClusterInfo.cpp
MBkkt: It will simpler

arangod/Cluster/ClusterInfo.cpp
MBkkt: In such case this code is incorrect, so if we will see data race here, we will found bug and will try to understand why it's repeated

arangod/Cluster/ClusterInfo.cpp
MBkkt: I don't think here exist difference in terms of simplicity

arangod/Cluster/ClusterInfo.cpp
mpoeter, line 3842:
```suggestion
    auto nrDone = report->getLockedGuard()->nrDone;
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3314,8 +3318,8 @@ Result ClusterInfo::createCollectionsCoordinator(
       // owned by the callback d) info might be deleted, so we cannot use it. e)
       // If the callback is ongoing during cleanup, the callback will
       //    hold the Mutex and delay the cleanup.
-      RECURSIVE_MUTEX_LOCKER(*cacheMutex, *cacheMutexOwner);
-      *isCleaned = true;
+      report->doUnderLock(
+          [](CreateCollectionReport& report) { report.isCleaned = true; });
```
mpoeter, line 3322:
```suggestion
      report->getLockedGuard()->isCleaned = true;
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3789,7 +3809,10 @@ Result ClusterInfo::createCollectionsCoordinator(
   });
 
   do {
-    auto tmpRes = dbServerResult->load(std::memory_order_acquire);
+    auto tmpRes = report->doUnderLock(
+        [](CreateCollectionReport const& report) -> std::optional<ErrorCode> {
+          return report.dbServerResult;
+        });
```
mpoeter, line 3815:
```suggestion
    auto tmpRes = report->getLockedGuard()->dbServerResult;
```

arangod/Cluster/ClusterInfo.cpp
mpoeter, line 3842:
```suggestion
    auto nrDone = report->getLockedGuard()->nrDone;
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3899,7 +3926,10 @@ Result ClusterInfo::createCollectionsCoordinator(
       // Report if this operation worked, if it failed collections will be
       // cleaned up by deleteCollectionGuard.
       for (auto const& info : infos) {
-        TRI_ASSERT(info.state == ClusterCollectionCreationState::DONE);
+        TRI_ASSERT(report->doUnderLock(
+                       [&info](auto const&) -> ClusterCollectionCreationState {
+                         return info.state;
+                       }) == ClusterCollectionCreationState::DONE);
```
mpoeter, line 3932:
```suggestion
        TRI_ASSERT(report->getLockedGuard()->state == ClusterCollectionCreationState::DONE);
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3932,7 +3962,11 @@ Result ClusterInfo::createCollectionsCoordinator(
           << " isNewDatabase: " << isNewDatabase
           << " first collection name: " << infos[0].name
           << " result: " << *tmpRes;
-      return {*tmpRes, *errMsg};
+      auto errMsg = report->doUnderLock(
+          [](CreateCollectionReport const& report) -> std::string {
+            return report.errMsg;
+          });
```
mpoeter, line 3968:
```suggestion
      auto errMsg = report->getLockedGuard()->errMsg;
```

arangod/Cluster/ClusterInfo.cpp
mchacki: My thought here was mainly to keep everything that was updated under the same lock before also update und the same lock, therefore I placed all atomics into one guarded structure.
I agree that is potentially less performant than before, but we are waiting on all-server round-trips here, so mutex performant should not fall into account at all.

arangod/Cluster/ClusterInfo.cpp
mchacki: I want to keep impact on this fix as minimal as possible, so everything stays as it was before.

arangod/Cluster/ClusterInfo.cpp
mchacki: I do not understand the comment.
info.size() is constant and will not change (we do not suddenly create more collections then asked for)
But nrDone will change, everytime one of the collections shows up in current.
=> I need synchronization on nrDone, which achieved over the Guarded structure.
info.size() is indeed accessed without a lock.

arangod/Cluster/ClusterInfo.cpp
mchacki: I agree this would be better.
I just wanted to keep the change as small as possible and tried to "just" move the lock-protected entries into one data-structure.
Given that we most likely want to backport this change.

Will try locally how much I need to adapt for this change.

mchacki: Applied change

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -122,6 +122,13 @@ struct ShardStatistics {
 
 namespace {
 
+struct CreateCollectionReport {
+  std::optional<ErrorCode> dbServerResult{std::nullopt};
```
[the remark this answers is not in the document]
mchacki: Done, replaced by Result

arangod/Cluster/ClusterInfo.cpp
MBkkt: It's about 3918

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -3314,8 +3318,8 @@ Result ClusterInfo::createCollectionsCoordinator(
       // owned by the callback d) info might be deleted, so we cannot use it. e)
```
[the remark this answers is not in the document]
mchacki: Thank you for adding this.
=> I found a potential use-after free while following the instructions here introduced with my current version. So need to fix this. :+1:

neunhoef (approved):
LGTM

PR #18752, opened 2023-04-25
Increase container size for shell_server_aql tests to avoid OOM kills

mpoeter:
### Scope & Purpose

This is just a quick fix. Long term we should identify the tests that cause the memory usage to grow so much and either optimize them or move them to a separate "large" suite.

- [x] :hankey: Bugfix


markuspf (approved):

CryptoNinjaGeek:
Size increase approved for the trial period

PR #18759, opened 2023-04-25
[CINFRA-747] Skip explicit commits

apetenchea:
### Scope & Purpose

For simple transactions (that is, not ElCheapo nor AQL), we can skip the explicit commit. This way, the follower applies transactions immediately, without waiting for the commit marker.
This serves as an experiment to see whether it has any impact on performance.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


apetenchea:
We concluded that this optimization currently has no effect on performance, as we found other bottlenecks, such as the storage engine. However, we might reconsider it in the future, as it at least reduces the amount of network requests.

PR #18781, opened 2023-04-27
[GORDO-1449] Worker State Machine

adityamukho:
### Scope & Purpose

This PR implements the Pregel worker actor as a state machine, similar to the conductor, as described in https://arangodb.atlassian.net/browse/GORDO-1449.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

https://github.com/arangodb/arangodb/pull/18615

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/GORDO-1449
- [ ] Design document:


arangod/Pregel/Worker/Handler.h
```diff
@@ -49,18 +50,22 @@ struct VerticesProcessed {
 
 template<typename V, typename E, typename M, typename Runtime>
 struct WorkerHandler : actor::HandlerBase<Runtime, WorkerState<V, E, M>> {
+  DispatchStatus const& dispatchStatus =
+      this->template dispatch<pregel::message::StatusMessages>;
+  DispatchMetrics const& dispatchMetrics =
+      this->template dispatch<metrics::message::MetricsMessages>;
+  DispatchConductor const& dispatchConductor =
+      this->template dispatch<conductor::message::ConductorMessages>;
```
jvolmer, line 58: Just a thought: We could also only give the dispatchStatus to the receive function (instead of giving 3 dispatch function to it) because this is the only function that needs to be called during a state. The other dispatching - to metrics and conductor actor and also the calls to the status actor at the end of a state - could still be made inside the handler.
Pro: We just give one dispatch function to the state and therefore have less callbacks which can make the code more readable. The remaining sending is transparent in the handler. 
Con: Two different ways of dispatching from the worker - perhaps this is more confusing...
I'm tending a bit more to the approach with less callbacks, but are not sure what is the best approach, just wanted to mention it here.

arangod/Pregel/Worker/Handler.h
```diff
@@ -49,18 +50,22 @@ struct VerticesProcessed {
...
-    this->template dispatch<conductor::message::ConductorMessages>(
-        this->state->conductor, ResultT<conductor::message::WorkerCreated>{});
 
-    this->template dispatch<metrics::message::MetricsMessages>(
-        this->state->metricsActor,
-        arangodb::pregel::metrics::message::WorkerStarted{});
+    auto newState =
+        this->executionState->receive(this->sender, start, dispatchStatus,
+                                      dispatchMetrics, dispatchConductor);
+    changeState(newState);
```
jvolmer, line 68: If you stay with the approach that you give all dispatch functions to the execution function receive method, you could actually change the execution state just inside the execution state because it knows the worker - then the receive function does not need to return anything.

arangod/Pregel/Worker/Handler.h
adityamukho: OK, let's see what @markuspf has to say.

arangod/Pregel/Worker/Handler.h
adityamukho: Agreed. Will change this, contingent on Markus letting us get away with 3 callbacks in the `receive` method.

adityamukho: Bundled all dispatch callbacks (there were more) into a `Dispatcher` struct.

arangod/Pregel/Worker/ExecutionStates/InitialState.cpp
```diff
@@ -0,0 +1,105 @@
...
+template<typename V, typename E, typename M>
+Initial<V, E, M>::Initial(actor::ActorPID self, WorkerState<V, E, M>& worker)
+    : self(std::move(self)), worker{worker} {}
+
+template<typename V, typename E, typename M>
+auto Initial<V, E, M>::receive(actor::ActorPID const& sender,
+                               worker::message::WorkerMessages const& message,
+                               Dispatcher dispatcher)
+    -> std::unique_ptr<ExecutionState> {
+  if (std::holds_alternative<worker::message::PregelMessage>(message)) {
```
jvolmer, line 55: You should never get a PregelMessage in the Initial state. A PregelMessage should only be received in the Computing state and the LoadedState (the latter one happens if another worker already switched to the Computing state and sends a PregelMessage and this worker is in the progress of changing from the Loaded to the Computing state)

arangod/Pregel/Worker/ExecutionStates/LoadingState.cpp
```diff
@@ -0,0 +1,130 @@
...
+template<typename V, typename E, typename M>
+Loading<V, E, M>::Loading(actor::ActorPID self, WorkerState<V, E, M>& worker)
+    : self(std::move(self)), worker{worker} {}
+
+template<typename V, typename E, typename M>
+auto Loading<V, E, M>::receive(actor::ActorPID const& sender,
+                               worker::message::WorkerMessages const& message,
+                               Dispatcher dispatcher)
+    -> std::unique_ptr<ExecutionState> {
+  if (std::holds_alternative<worker::message::PregelMessage>(message)) {
```
jvolmer, line 54: Also in this state we don't expect any PregelMessage.

arangod/Pregel/Worker/ExecutionStates/FatalErrorState.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include "State.h"
+
+namespace arangodb::pregel::worker {
+template<typename V, typename E, typename M>
+struct WorkerState;
+
+struct FatalError : ExecutionState {
+  // TODO: Tell the conductor that this worker has entered an error state.
```
jvolmer, line 33: this todo still needs to be done

arangod/Pregel/Worker/ExecutionStates/LoadedState.cpp
```diff
@@ -0,0 +1,95 @@
...
+template<typename V, typename E, typename M>
+Loaded<V, E, M>::Loaded(actor::ActorPID self, WorkerState<V, E, M>& worker)
+    : self(std::move(self)), worker{worker} {}
+
+template<typename V, typename E, typename M>
+auto Loaded<V, E, M>::receive(actor::ActorPID const& sender,
+                              worker::message::WorkerMessages const& message,
+                              Dispatcher dispatcher)
+    -> std::unique_ptr<ExecutionState> {
+  if (std::holds_alternative<worker::message::PregelMessage>(message)) {
```
jvolmer, line 53: this is correct: In this state we could already receive a PregelMessage from another actor that is already in the Computing state

arangod/Pregel/Worker/Handler.h
```diff
@@ -283,60 +130,9 @@ struct WorkerHandler : actor::HandlerBase<Runtime, WorkerState<V, E, M>> {
         arangodb::pregel::metrics::message::WorkerGssStarted{.threadsAdded =
                                                                  1});
```
jvolmer, line 131: These lines should be deleted, this message is already sent inside the Computing state

arangod/Pregel/Worker/ExecutionStates/StoredState.cpp
```diff
@@ -0,0 +1,90 @@
...
+template<typename V, typename E, typename M>
+Stored<V, E, M>::Stored(actor::ActorPID self, WorkerState<V, E, M>& worker)
+    : self(std::move(self)), worker{worker} {}
+
+template<typename V, typename E, typename M>
+auto Stored<V, E, M>::receive(actor::ActorPID const& sender,
+                              worker::message::WorkerMessages const& message,
+                              Dispatcher dispatcher)
+    -> std::unique_ptr<ExecutionState> {
+  if (std::holds_alternative<worker::message::Cleanup>(message)) {
```
jvolmer, line 54: yes, you can receive a Cleanup message here, but you should actually be able to receive a Cleanup message in every execution state and do the cleanup, because a pregel run can be canceled at any time by the user.

arangod/Pregel/Worker/Handler.h
```diff
@@ -350,118 +146,50 @@ struct WorkerHandler : actor::HandlerBase<Runtime, WorkerState<V, E, M>> {
...
-    this->template dispatch<pregel::message::SpawnMessages>(
-        this->state->spawnActor, pregel::message::SpawnCleanup{});
-    this->template dispatch<pregel::conductor::message::ConductorMessages>(
-        this->state->conductor, pregel::conductor::message::CleanupFinished{});
-    this->template dispatch<metrics::message::MetricsMessages>(
-        this->state->metricsActor,
-        arangodb::pregel::metrics::message::WorkerFinished{});
+    auto newState =
+        this->state->executionState->receive(this->sender, message, dispatcher);
+    changeState(std::move(newState));
```
jvolmer, line 192: You have to be able to receive a Cleanup message in every exeution state and cleanup, so you can also think about just reverting the changes in this function and just leave the execution state as it was (or change for better clarity to the Done state I mentioned in ResultsProducedState.cpp) because calling `this->finish()` means that this worker is garbage collected in the near future anyway. Then you would also not need both CleaningUp and CleanedUp execution state.

arangod/Pregel/Worker/ExecutionStates/ResultsProducedState.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "Pregel/Algos/SLPA/SLPAValue.h"
+#include "Pregel/Algos/ColorPropagation/ColorPropagationValue.h"
+#include "Pregel/Algos/DMID/DMIDMessage.h"
+
+using namespace arangodb;
+using namespace arangodb::pregel;
+using namespace arangodb::pregel::worker;
+
+template<typename V, typename E, typename M>
+ResultsProduced<V, E, M>::ResultsProduced(actor::ActorPID self,
```
jvolmer, line 46: This could also be a generic Done state. Then you would also not need the Stored state but change from Storing to Done state.

arangod/Pregel/Worker/State.h
```diff
@@ -64,6 +68,7 @@ struct WorkerState {
         resultActor(std::move(resultActor)),
         statusActor(std::move(statusActor)),
         metricsActor(std::move(metricsActor)) {
+    executionState = std::make_unique<Initial<V, E, M>>(self, *this);
     config->updateConfig(specifications);
 
     if (messageCombiner) {
```
jvolmer, line 74: this whole if statement could possibly be done in the constructor of the Computing state.

arangod/Pregel/Worker/State.h
```diff
@@ -95,6 +100,7 @@ struct WorkerState {
   uint32_t messageBatchSize = 500;
```
jvolmer, line 100: It would be really helpful to extract as many of these variables into the Computing state as possible, my guess would be that you can move isWaitingForAllMessagesSince, readCache, writeCache and messageBatchSize into the Computing state.

jvolmer (commented):
I really like that everything interesting is happening now in the execution states and that the handler is so empty!

arangod/Pregel/Worker/State.h
adityamukho: Cannot move `readCache` and `writeCache` since they're initialized in the `WorkerState` constructor, based on the `messageCombiner` field.

That leaves just `isWaitingForAllMessagesSince` and `messageBatchSize` as candidates for moving into the `Computing` state. Not sure if the refactor makes sense for just two fields.

arangod/Pregel/Worker/ExecutionStates/StoredState.cpp
adityamukho: Fixed in 41b3ac94a31112e68185e219bf6af0c9e10ffd4c

markuspf (approved):
I think this looks good, it only needs to be rebased on `devel` again and then we should merge it and get to the PR that activates the actor code and make all test green *and test the performance*

PR #18792, opened 2023-04-27
Bug fix/test failure backup dump collections not ready

cpjulia:
### Scope & Purpose

This PR aims to fix the following error which repeats for 3 different tests cases in `tests/js/server/dump/dump-test.inc`: testKeygenAutoInc_cluster, testKeygenPadded_cluster and testKeygenUuid_cluster
in the error, one cannot insert documents into the collection because it's in read-only state, which, deductively, would mean that the shards are not in sync yet, making the collection not ready to be used in transactions.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


tests/js/server/dump/dump-test.inc
```diff
@@ -51,6 +51,34 @@
...
 
+  function waitForCollInSync(collName) {
+    let res;
+    let numSecs = 0.5;
+    let collDistribution = {};
+    let collDistributionPlan = {};
+    let collDistributionCurrent = {};
+    while (true) {
+      const dbName = db._name();
+      const res = db._connection.GET("/_db/" + dbName + "/_admin/cluster/shardDistribution?returnAsJSON=true");
```
jsteemann, line 62: `returnAsJSON` does nothing for the called REST endpoint.
```suggestion
      const res = db._connection.GET("/_db/" + dbName + "/_admin/cluster/shardDistribution");
```

tests/js/server/dump/dump-test.inc
```diff
@@ -51,6 +51,34 @@
...
+      const colls = res["results"];
+      for (const key in colls) {
+        if (key.includes(collName)) {
```
jsteemann, line 65: why is this an `includes` here? Shouldn't it be `===`?

tests/js/server/dump/dump-test.inc
```diff
@@ -51,6 +51,34 @@
...
+          collDistribution = colls[key];
```
jsteemann, line 66: we can break after we have found the correct `collDistribution`

tests/js/server/dump/dump-test.inc
```diff
@@ -51,6 +51,34 @@
...
+        }
+      }
+      assertFalse(_.isEmpty(collDistribution));
+      collDistributionPlan = collDistribution["Plan"];
+      collDistributionCurrent = collDistribution["Current"];
+      assertFalse(_.isEmpty(collDistributionPlan));
+      assertFalse(_.isEmpty(collDistributionCurrent));
+      if (_.isEqual(collDistributionPlan, collDistributionCurrent) || numSecs >= 16) {
```
jsteemann, line 74: if we reach `numSecs >= 16`, should we continue the test anyway? it will very likely fail then.
Why not assert here that numSecs is < 16, and if the assertion fails, use a descriptive error message so that developers can see that the shards did not get in sync. This will be better than random hot_backup test failures down the line that are hard to understand and debug.

jsteemann (approved):
LGTM

maierlars (approved):

PR #18797, opened 2023-04-28
FE-255: add validation for collection, document, graph, view and database names

palashkaria:
### Scope & Purpose

- add extended & traiditonal validation  when adding a collection, document, graph, view and database names
- add traditional validation when editing a arangosearch view – to be improved later

For more details see [FE-255](https://arangodb.atlassian.net/browse/FE-255)
 
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested
- [x] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/18862

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-255

[FE-255]: https://arangodb.atlassian.net/browse/FE-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/28867/

js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
```diff
@@ -508,6 +528,11 @@
               '',
               true,
               [
+                ...(!extendedNames ? traditionalNameValidation : extendedValidation),
+                {
+                  rule: Joi.string().max(256, 'utf8'),
+                  msg: 'Collection name max length is 256.'
```
jsteemann, line 534: Max collection name length is 256 bytes, not 256 UTF-8 characters (which could use more than 256 bytes storage to represent them)

js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
```diff
@@ -498,7 +498,27 @@
...
+            msg: 'Collection name must always start with a letter.'
+          },
+          {
+            rule: Joi.string().regex(/^[a-zA-Z0-9\-_]*$/),
+            msg: 'Only symbols, "_" and "-" are allowed.'
+          }];
+          var extendedValidation = [{
+            rule: Joi.string().regex(/^(?![0-9])/),
+            msg: 'Collection name cannot start with a number.'
+          },
```
jsteemann, line 513: There is a rule missing here: collection names must not start with a `.` (dot), and, non-system collection names must not start with a `_` (underscore).

jsteemann: Additionally, any control characters below `\u0020` are disallowed in collection names.

js/apps/system/_admin/aardvark/APP/frontend/js/views/databaseView.js
```diff
@@ -386,7 +386,34 @@
...
+      }, {
+        rule: Joi.string().regex(/^(?!.*[/:])/),
+        msg: 'Database name cannot contain a forward slash (/) or a colon (:)'
+      }, {
+        rule: Joi.string().regex(/^\S(.*\S)?$/),
+        msg: 'Database name cannot contain leading/trailing spaces.'
+      },
+      {
+        rule: Joi.string().max(128, 'utf8'),
+        msg: 'Database name max length is 128.'
```
jsteemann, line 415: 128 bytes, not characters. And it is 128 bytes after NFC normalization.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
```diff
@@ -671,6 +671,20 @@
               if (!this.collection.getIsEnterpriseEdge()) {
                 // We do show the _key globally for every type of edge collection,
                 // except in case we do have an "EnterpriseSmart" edge.
+                var documentNameValidations = [
+                  {
+                    rule: Joi.string().regex(/^[a-zA-Z0-9_\-:\.@()\+,=;$!*\%'%]+$/),
```
jsteemann, line 676: At the end of the sequence I see `\%'%]`. Why is the percent character listed twice, and escaped once? This does not look right.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
```diff
@@ -671,6 +671,20 @@
...
+                    msg: "Only these characters are allowed: a-z, A-Z, 0-9 and  _ - : . @ ( ) + , = ; $ ! * ' %."
+                  },
+                  {
+                    rule: Joi.string().max(254, 'utf8'),
+                    msg: 'Document key max length is 254.'
+                  },
+                  {
+                    rule: Joi.string().allow('').optional(),
```
jsteemann, line 684: Not sure what this rule means, but an empty document key is not allowed.

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
```diff
@@ -746,20 +755,24 @@
...
-
+      var documentNameValidations = [
+        {
+          rule: Joi.string().regex(/^[a-zA-Z0-9_\-:\.@()\+,=;$!*\%'%]+$/),
+          msg: "Only these characters are allowed: a-z, A-Z, 0-9 and  _ - : . @ ( ) + , = ; $ ! * ' %."
+        },
+        {
+          rule: Joi.string().allow('').optional(),
+          msg: ''
+        }
```
jsteemann, line 766: Same comments apply as above.

If the same rules appear repeatedly in the same file, I suggest putting them only in one place in this file, e.g. into a single function that returns them. That will reduce code maintenance long-term.

jsteemann: Validation for key length (254 bytes max) is missing here, but is present in the other location in this file.

js/apps/system/_admin/aardvark/APP/frontend/js/views/graphManagementView.js
```diff
@@ -1253,15 +1253,29 @@
 
         // create graph section
         title = 'Create Graph';
-
+        var nameValidation = [
+          {
+            rule: Joi.string().regex(/^[a-zA-Z0-9_\-:\.@()\+,=;$!*\%'%]+$/),
```
jsteemann, line 1258: Third time this validation rule is used in the code.
It should really be placed into a function that is used from all 3 places.

js/apps/system/_admin/aardvark/APP/frontend/js/views/settingsView.js
```diff
@@ -146,7 +146,27 @@
 
           var buttons = [];
           var tableContent = [];
-
+          var extendedNames = window.frontendConfig.extendedNames;
+          var traditionalNameValidation = [{
+            rule: Joi.string().regex(/^[a-zA-Z]/),
```
jsteemann, line 151: This is a repetition of the validation rules used in collectionsView.js. 
Instead of repeating the rules here and having to change them in multiple places, they should be put into functions and shared between call sites.

js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/IndexFieldsHelper.ts
```diff
@@ -26,7 +26,20 @@ export const commonFieldsMap = {
...
 
+const traditionalNameSchema = Yup.string()
+  .max(256, "Index name max length is 256.")
+  .matches(/^[a-zA-Z]/, "Index name must always start with a letter.")
+  .matches(/^[a-zA-Z0-9\-_]*$/, 'Only symbols, "_" and "-" are allowed.');
+
+const extendedNameSchema = Yup.string()
+  .max(256, "Index name max length is 256.")
+  .matches(/^(?![0-9])/, "Index name cannot start with a number.")
+  .matches(/^\S(.*\S)?$/, "Index name cannot contain leading/trailing spaces.");
```
jsteemann, line 37: An index name must also not contain `/` (forward slash).

js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/IndexFieldsHelper.ts
```diff
@@ -26,7 +26,20 @@ export const commonFieldsMap = {
   }
 };
 
+const traditionalNameSchema = Yup.string()
+  .max(256, "Index name max length is 256.")
+  .matches(/^[a-zA-Z]/, "Index name must always start with a letter.")
+  .matches(/^[a-zA-Z0-9\-_]*$/, 'Only symbols, "_" and "-" are allowed.');
+
+const extendedNameSchema = Yup.string()
+  .max(256, "Index name max length is 256.")
```
jsteemann, line 35: 256 bytes after UTF-8 NFC normalization

jsteemann (changes requested):
I think there are a few small deviations in the validations used by the frontend (via this PR) and the backend. These should be fixed.
I also suggest to put the validation rules into functions, so they can be reused from multiple call sites. For example, I have seen the same validation rule for document keys being declared 3 times in this PR. This will likely lead to inconsistencies in the future, or at least increase our code maintenance effort.
But TBH I would prefer the frontend to not implement the validation rules on its own, because if at some point the validation is changed in the backend, the frontend will not know about it and we will have inconsistencies.
What would be much better IMHO would be an API call such as `/_api/validate/<type>` that the frontend can use to check if a given name is valid or not. That would allow us to keep the validation rules in a single place, as it should be.

js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
palashkaria: @jsteemann should we also exclude `\u007F` (127 - DEL key)?

js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
palashkaria: Yes, you are right. I verified this
1. For Joi: since we provided the encoding, Joi does use the [byte length](https://github.com/hapijs/joi/blob/master/lib/types/string.js#L471), ([docs](https://github.com/hapijs/joi/blame/master/lib/types/string.js#L471), [issue](https://github.com/hapijs/joi/blame/master/lib/types/string.js#L471)). We have to add normalize() too
2. For Yup, I am adding a custom validation method which will do this
3. For JSONs (i.e. edits), we are currently not doing length validation

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
palashkaria: they key is omitted, this was already the case, no change made there

js/apps/system/_admin/aardvark/APP/frontend/js/views/documentsView.js
palashkaria: moved validations to common helper

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
```diff
@@ -0,0 +1,143 @@
+/* jshint unused: false */
+(function () {
+  "use strict";
+
+  window.arangoValidationHelper = {
+    getControlCharactersRegex: () => {
+      return /[^\u0000-\u001F]/;
```
jsteemann, line 7: Is this regex actually correct?
IMO it matches already if a _single_ character is outside the control characters class, but it doesn't check the entire string.
Example:
```
> "foobar\tab".match(/[^\u0000-\u001F]/)
[ 'f', index: 0, input: 'foobar\tab', groups: undefined ]

> "foobar\tab".match(/^[^\u0000-\u001F]+$/)
null
```

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
```diff
@@ -0,0 +1,143 @@
...
+    },
+    getDocumentKeySpecialCharactersValidation: () => {
+      var keySpecialCharactersValidation = {
+        rule: Joi.string().regex(/^[a-zA-Z0-9_\-:\.@()\+,=;$!*\%']+$/),
```
jsteemann, line 11: nit: percent character does not need to be escaped
```suggestion
        rule: Joi.string().regex(/^[a-zA-Z0-9_\-:\.@()\+,=;$!*%']+$/),
```

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
```diff
@@ -0,0 +1,143 @@
...
+        rule: Joi.string().required(),
+        msg: "No database name given.",
+      });
+      return databaseNameValidations;
+    },
+    getDocumentNameValidations: () => {
+      var keySpecialCharactersValidation =
+        window.arangoValidationHelper.getDocumentKeySpecialCharactersValidation();
+      return [
+        keySpecialCharactersValidation,
```
jsteemann, line 70: Should there be a validation for the maximum key length (254) here?

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
```diff
@@ -0,0 +1,143 @@
...
+      return collectionNameValidations;
+    },
+    getGraphNameValidations: () => {
+      var keySpecialCharactersValidation =
+        window.arangoValidationHelper.getDocumentKeySpecialCharactersValidation();
+
+      var graphNameValidations = [
+        keySpecialCharactersValidation,
+        {
+          rule: Joi.string().normalize().max(254, "utf8"),
```
jsteemann, line 132: nit: graph name does not need normalization, as no UTF-8 characters are allowed in it.
```suggestion
          rule: Joi.string().max(254, "utf8"),
```

js/apps/system/_admin/aardvark/APP/react/src/views/collections/indices/addIndex/invertedIndex/useInvertedIndexJSONSchema.tsx
```diff
@@ -12,7 +14,8 @@ const invertedIndexJSONSchema: JSONSchemaType<InvertedIndexValuesType> = {
     },
     name: {
       nullable: true,
-      type: "string"
+      type: "string",
+      pattern: extendedNames ? "" : "^[a-zA-Z][a-zA-Z0-9-_]*$"
```
jsteemann, line 18: nit: inside character classes I prefer `-` literals to be escaped for better readability.
```suggestion
      pattern: extendedNames ? "" : "^[a-zA-Z][a-zA-Z0-9\-_]*$"
```

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
palashkaria: yes correct, I missed it for documents, adding it

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arangoValidationHelper.js
```diff
@@ -0,0 +1,147 @@
...
+      return databaseNameValidations;
+    },
+    getDocumentNameValidations: () => {
+      var keySpecialCharactersValidation =
+        window.arangoValidationHelper.getDocumentKeySpecialCharactersValidation();
+      return [
+        keySpecialCharactersValidation,
+        {
+          rule: Joi.string().max(256, "utf8"),
+          msg: "Document name max length is 256.",
```
jsteemann, line 73:
```suggestion
          msg: "Document key max length is 256.",
```

jsteemann (approved):
LGTM

KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/28923/

KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/28932/

PR #18808, opened 2023-04-28
FE-251: Prohibit multiple expansions of the same node

ajurdak:
### Scope & Purpose

It was possible to expand the same node in the graph viewer several times what led to multiple "(expanded)" entries in the node label. With this PR the IDs of the expanded nodes are collected in an array. If the right-clicked node has been expanded before the "Expand Node" menu item won't be displayed anymore.

- [x] :hankey: Bugfix

### Checklist

- [x] Manually tested
- [ ] Backports
  - [x] Backport for 3.11: #18831
  - [ ] Backport for 3.10: ToDo
  - [ ] Backport for 3.9: ?
  - [ ] Backport for 3.8: ?

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-251


js/apps/system/_admin/aardvark/APP/react/src/views/graphV2/graphNetwork/NodeRightClickMenu.tsx
```diff
@@ -44,7 +45,9 @@ export const NodeRightClickMenu = forwardRef(
           >
             Edit Node
           </MenuItem>
-          <ExpandNodeButton foundNode={foundNode} />
+          {
+            expandedNodes?.includes(rightClickedEntity.nodeId) ? null : <ExpandNodeButton foundNode={foundNode} />
```
palashkaria, line 49: why don't we just use `foundNode.label?.includes(" (expanded)")` ? We don't have to clutter the context with more stuff

palashkaria: Also, instead of removing the item, we can simply disable it. Otherwise it may feel broken


let's do that inside `ExpandNodeButton` L136 itself

```
return <MenuItem isDisabled={foundNode.label?.includes(" (expanded)")} onClick={expandNode}>Expand Node</MenuItem>;
```

palashkaria (changes requested):

js/apps/system/_admin/aardvark/APP/react/src/views/graphV2/graphNetwork/NodeRightClickMenu.tsx
ajurdak: I didn't do a string comparison as I see the danger of someone else changing the "(expanded)" string what then would break the condition. And to collect the IDs of the expanded nodes will help us to save some history of the users actions in the future.

ajurdak: In 4984792ccf801082e69b43867853733136811d16 I changed from hiding the "Expand Node" menu item to disable it.

palashkaria: I understand the concerns, but I’d still go with not tracking another state, as this code needs a lot of refactoring, and another state will just add to the overhead. As we are already tracking “(fixed)” using the string, it makes sense to follow the same pattern here, as during refactoring it will be simpler to follow patterns. I don’t want two different ways of doing this as it will be an overhead for refactoring. We should be maintaining state per node. 

The idea is: Let’s optimize for refactoring and add as little code as possible as of now. This is a bugfix, let’s limit the surface area. 

Let me know if you still think otherwise @pluma4345 , otherwise I’d recommend making the change I suggested @ajurdak

palashkaria (approved):

js/apps/system/_admin/aardvark/APP/react/src/views/graphV2/graphNetwork/NodeRightClickMenu.tsx
palashkaria: The reasoning is quite straighforward – we want to refactor this into better code. Keeping consistent, localized patterns helps us do that.

KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/28757/

PR #18812, opened 2023-04-28
Fix locally outdated information from Current

apetenchea:
### Scope & Purpose

**Brief**
This PR fixes a race in the Maintenance, impacting replication1 leaders that have just been resigned. It was detected in the _resilience-move_ suite, most often reproduced during the _testCleanOutMultipleCollections_ test. It may also cause instability during cluster upgrades.

**Problem**
A _CleanOutServer_ job works by spawning multiple _MoveShard_ sub-jobs. Is is considered finished only when all sub-jobs are finished successfully. If one or more sub-jobs fail, the job is considered failed.  
In a nutshell, a _MoveShard_ job for a shard _S_, from server _A_ (leader of _S_) to server _B_ (not a leader nor follower of _S_) has to perform the following steps:
1. Add _B_ as a follower of _S_
2. When _B_ is reported as a follower in _Current_, the job considers _B_ to be in-sync with _A_
3. Resign _A_ - this is signaled appending an underscore to it's ID in _Plan_

If after step (2), _B_ is removed from _Current_, the job considers it to be out-of-sync and aborts. This is exactly how the test was seen failing before.  
The problem is triggered after step (3). The Maintenance detects that _A_ is supposed to resign and executes a _ResignLeadership_ action. After that, _A_ is no longer leader, but the followers are not yet aware of it. For that, is has to report back to current. There is a special case for just-resigned leader in the `reportInCurrent` method, executed in `phaseTwo` of the maintenance. When this gets executed, _A_ writes updates the list of servers in _Current_ for shard _S_. However, as opposed to the case when it is actually a leader, the buggy version uses _A's_ understanding of _Current_, which is not guaranteed to be up-to-date. If that is the case, we might end up with an outdated list of servers, not containing _B_. This will abort the _MoveShard_ job, and consequently, the _CleanOutServer_ parent job.  
This is not a total failure, as there cluster is able to recover after the jobs are aborted. But, it creates instability and makes the _CleanOutServer_ procedure unreliable.  
Explanation

**Solution**
The proposed solution is to update _Current_ from the local information, just as if _A_ was still leader, as this information is up-to-date. Only _A_ is aware of the fact that it has resigned, so it can't possibly overwrite the changes of another leader, since no other server had yet any chance to take over the leadership of the shard.

- [x] :hankey: Bugfix

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports


tests/js/server/resilience/move/moving-shards-cluster.inc
```diff
@@ -409,8 +409,12 @@ exports.MovingShardsSuite = function({useData, replVersion}) {
     while (true) {
       var job = queryAgencyJob(result.json.id);
       console.info("Status of cleanOutServer job:", job.status);
-      if (job.error === false && job.status === "Finished") {
-        return result;
+      if (job.error === false) {
+        if (job.status === "Finished") {
+          return result;
+        } else if (job.status === "Failed") {
```
jsteemann, line 415: Is it possible for a job to have `job.error === true && job.status === "Failed"?
Or should the `job.status === "Failed"` check be moved out of the outer `job.error === false` condition here?

apetenchea: No, `error === true` is used to indicate that the job was not found. In such case, `job` is `null`.

neunhoef (approved):
LGTM

jsteemann (approved):
LGTM

PR #18826, opened 2023-05-02
Bug fix 3.11/test failure backup dump collections not ready

cpjulia:
### Scope & Purpose

Backport for https://github.com/arangodb/arangodb/pull/18792.

This PR aims to fix the following error which repeats for 3 different tests cases in `tests/js/server/dump/dump-test.inc`: testKeygenAutoInc_cluster, testKeygenPadded_cluster and testKeygenUuid_cluster
in the error, one cannot insert documents into the collection because it's in read-only state, which, deductively, would mean that the shards are not in sync yet, making the collection not ready to be used in transactions.

### DIFF
--- a/tests/js/server/dump/dump-test.inc
+++ b/tests/js/server/dump/dump-test.inc
@@ -51,6 +51,37 @@
     return c;
   }
 
+  function waitForCollInSync(collName) {
+    let res;
+    let numSecs = 0.5;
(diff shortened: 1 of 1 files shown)


jsteemann (approved):
LGTM

maierlars (approved):

PR #18832, opened 2023-05-02
[FE-249] Remove frontend build artifacts

pluma4345:
### Scope & Purpose

This PR removes the frontend build artifacts and npm dependencies from the repository.

The goal of this change is to simplify the code review and merge process for the frontend team and reduce the repository size. Dependency versions and changes will still be tracked via the committed lockfile to continue allowing reproducible builds.

As the frontend is an integral part of the ArangoDB product (i.e. accidentally not building the frontend when you don't know it has to be built explicitly would be bad), the CMake frontend task is now enabled by default. To allow backend developers to opt out of this, this PR adds a `SKIP_FRONTEND` flag which also skips the checks for Node 16 and `yarn`.

As discussed with @markuspf, an improvement on this PR would be to find a way to track changes in frontend source files so CMake knows not to rebuild the frontend every time `make` is invoked. This would primarily affect backend developers as frontend developers mostly use the dev server to work on frontend code.

This PR requires two changes to our Circle CI configuration going forward:
1. The default CMake flags should disable the frontend build.
2. PRs modifying frontend file should run the frontend build (as well as any acceptance tests in the future).

In its current state, this PR works as intended and is ready for review. The possible improvements to the CMake dirty tracking are beyond my capabilities so any feedback on how these could be implemented is welcome. Otherwise we should probably consider them out of scope for this PR. The frontend build currently takes around 2 minutes on my laptop, so these optimizations are primarily about reducing rebuild time, not so much reducing the time for a full build.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: FE-249
- [ ] Design document:


cmake/frontend/aardvark.cmake
```diff
@@ -1,13 +1,195 @@
-
 ################################################################################
 ## web interface build
 ################################################################################
-add_custom_target(frontend
+
+find_package(Nodejs 16 REQUIRED)
+find_package(Yarn REQUIRED)
+
+set(FRONTEND_SRC
```
markuspf, line 8: I know this looks weird, but is a "best practice" in CMake projects to explicitly list all dependencies; I am not sure whether it is just the `.js` files that constitute frontend sources or whether more files should be added here.

CMake's dependency tracking will trigger a rebuild of the frontend whenever a file in this list is touched, and cmake will complain if a file in this list is missing.

Of course it will not complain if there are excess files, i.e. a developer forgot to add a file here.

mpoeter: TBH I don't think it makes sense to track all the js files in cmake. CMake only needs to create the `frontend` target for `make` which will then trigger `webpack` for the actual build which will automatically pick up all the required js files anyway.

So I think we should just remove this.

palashkaria: @pluma4345 can we remove this as discussed? I think after that we are good to go for this

js/apps/system/_admin/aardvark/APP/react/src/views/query/editor/ImportQueryModal.tsx
```diff
@@ -66,20 +66,11 @@ export const ImportQueryModal = ({
...
-        }
+        if (typeof data !== "string") return;
         const queries = JSON.parse(data);
-        const sanitizedQueries = queries.map((query: QueryType) => {
-          return {
-            name: query.name,
-            value: query.value,
-            parameter: query.parameter
-          };
-        });
```
palashkaria, line 79: @pluma4345 this probably sneaked in while resolving merge conflicts? This should be reverted

js/apps/system/_admin/aardvark/APP/react/src/views/query/queryResults/QueryExecuteResult.tsx
```diff
@@ -220,10 +220,6 @@ export const getAllowCSVDownload = (queryResult: QueryResultType) => {
   }
   let allowCSVDownload = true;
   queryResult.result.forEach((row: any) => {
-    if (!row) {
-      allowCSVDownload = false;
-      return;
-    }
     if (typeof row === "object") {
```
palashkaria, line 223: @pluma4345 this is also a bug fix, please revert this

js/apps/system/_admin/aardvark/APP/react/src/views/collections/addCollection/AddCollectionModal.tsx
```diff
@@ -25,7 +25,7 @@ const INITIAL_VALUES = {
   distributeShardsLike: "",
   isSatellite: false,
   smartJoinAttribute: ""
-} as Partial<CreateCollectionOptions> & Record<string, any>;
```
palashkaria, line 28: `staisfies` operator throws typescript errors on build for me. Please revert this

js/apps/system/_admin/aardvark/APP/react/src/views/collections/addCollection/AddCollectionModal.tsx
```diff
@@ -136,7 +136,7 @@ export const AddCollectionModal = ({
       isSatellite,
       ...extra
     } = values;
-    const clusterOptions = window.App
+    const clusterOptions: Partial<CreateCollectionOptions> = window.App
```
palashkaria, line 139: I don't think this was intentional here. Let's revert any unnecessary changes in this PR

palashkaria (changes requested):

.circleci/base_config.yml
```diff
@@ -9,7 +9,7 @@ parameters:
     default: "devel"
   build-targets:
     type: string
-    default: "arangod arangosh arangoimport arangoexport arangodump arangorestore arangobench"
+    default: "arangod arangosh arangoimport arangoexport arangodump arangorestore arangobench frontend"
```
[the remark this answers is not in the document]
KVS85: @pluma4345 we don't need this frontend here because CircleCI tests don't use UI.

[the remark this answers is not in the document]
KVS85: On the other hand it's interesting that still Swagger is eliminated as well and this frontend target is technically dependent.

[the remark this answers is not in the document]
markuspf: @pluma4345 to add these tools to the build environment they need to be added to the Dockerfile `scripts/docker/build/arangodb-build.Dockerfile` and the docker image has to be re-built (which at this point is still a manual process), and the docker image has to be added with the correct tag to the circleci config.

it looks like nodejs is already installed currently though (because the linting step uses eslint)

I can make these changes tomorrow if noone beats me to it.

palashkaria:
@pluma4345 I wanted to check - are we planning to port these changes to 3.10/3.9 too? Is that even possible? (I don't remember what was discussed/agreed clearly)

palashkaria (approved):
LGTM!

KVS85 (approved):

mpoeter (approved):
Frankly, I would just remove the whole make/cmake stuff and have people call yarn directly. But that is just my opinion...

Either way - change LGTM.

KVS85:
Unrelated functional failure only (not related to frontend build).

PR #18854, opened 2023-05-04
[BTS-1340]

cpjulia:
### Scope & Purpose

Fixed telemetrics api making shell hang when logging into it and leaving it too quickly so the telemetrics API doesn't have time to send the data to the warehouse. The thread could be hanging on SSL_connect() until the connection timeout if the socket is blocking, so the socket was made non-blocking for the thread to leave SSL_connect() after the connection is interrupted when leaving the shell.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: https://github.com/arangodb/arangodb/pull/18912

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1340
- [ ] Design document:


lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,6 +305,10 @@ bool SslClientConnection::connectSocket() {
 
   _errorDetails.clear();
   _socket = _endpoint->connect(_connectTimeout, _requestTimeout);
+  long flags = fcntl(_socket.fileDescriptor, F_GETFL, 0);
```
jsteemann, line 308: We should not call `fcntl` in case we don't need to. This will likely run a system call otherwise, which we would like to avoid if not needed.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,6 +305,10 @@ bool SslClientConnection::connectSocket() {
...
+  if (_isTelemetrics) {
+    fcntl(_socket.fileDescriptor, F_SETFL, flags | O_NONBLOCK);
```
jsteemann, line 310: Do we want to handle return values from `fcntl` here? It can return an error, at least.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +350,33 @@ bool SslClientConnection::connectSocket() {
...
+      auto end = steady_clock::now();
+      if ((errorDetail != SSL_ERROR_WANT_READ &&
+           errorDetail != SSL_ERROR_WANT_WRITE) ||
+          duration_cast<seconds>(end - start).count() >= _connectTimeout) {
+        break;
+      }
+    }
+  } else {
+    ret = SSL_connect(_ssl);
+    errorDetail = SSL_get_error(_ssl, ret);
```
jsteemann, line 372: I guess we should not call `SSL_get_error` unconditionally here. What if no error has happened at all? Is it then still ok to call `SSL_get_error` here?

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +350,33 @@ bool SslClientConnection::connectSocket() {
...
+      errorDetail = SSL_get_error(_ssl, ret);
+      if (_isInterrupted) {
+        break;
+      }
+      auto end = steady_clock::now();
+      if ((errorDetail != SSL_ERROR_WANT_READ &&
+           errorDetail != SSL_ERROR_WANT_WRITE) ||
+          duration_cast<seconds>(end - start).count() >= _connectTimeout) {
+        break;
+      }
```
jsteemann, line 368: This is a busy loop. Do we want to add at least some sleep here?

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +350,33 @@ bool SslClientConnection::connectSocket() {
...
   if (ret != 1) {
     _errorDetails.append("SSL: during SSL_connect: ");
 
-    int errorDetail;
     long certError;
 
-    errorDetail = SSL_get_error(_ssl, ret);
     if ((errorDetail == SSL_ERROR_WANT_READ) ||
         (errorDetail == SSL_ERROR_WANT_WRITE)) {
       return true;
```
jsteemann, line 382: If we drop out here and `_isTelemetrics` is `true`, we don't clean up the socket flags.
If there are other premature exits from this function, then they may also miss the cleanup.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -446,48 +473,57 @@ bool SslClientConnection::writeClientConnection(void const* buffer,
     return false;
   }
 
-  int written = SSL_write(_ssl, buffer, (int)length);
-  int err = SSL_get_error(_ssl, written);
-  switch (err) {
-    case SSL_ERROR_NONE:
-      *bytesWritten = written;
+  int written = -1;
+  auto start = steady_clock::now();
```
jsteemann, line 477: Is any modification to `writeClientConnection()` needed, given that the socket flags are only modified upon connection?
It would be better to minimize the changes in this PR so that only the connection part is changed.

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -300,6 +300,7 @@ TelemetricsHandler::buildHttpClient(std::string& url) const {
     std::unique_ptr<httpclient::GeneralClientConnection> connection(
         httpclient::GeneralClientConnection::factory(
             cf.getCommFeaturePhase(), newEndpoint, 30, 60, 3, sslProtocol));
+    connection->setSocketNonBlocking(true);
```
jsteemann, line 303: This is not safe, because the returned `connection` object can be a nullptr.
Please move the line into the following `if` statement.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,9 +307,30 @@ bool SslClientConnection::connectSocket() {
...
+      return false;
+    }
+    if (fcntl(_socket.fileDescriptor, F_SETFL, _socketFlags | O_NONBLOCK) ==
+        -1) {
+      _errorDetails = "Attempt to create non-blocking socket generated error " +
+                      std::to_string(errno);
+      _isConnected = false;
+      return false;
+    }
+  }
```
jsteemann, line 325: All this code should only be executed after the check for `TRI_isvalidsocket(_socket)`. Because why should we execute it on a socket that we already does not work?

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,9 +307,30 @@ bool SslClientConnection::connectSocket() {
...
 
   if (!TRI_isvalidsocket(_socket) || _ctx == nullptr) {
     _errorDetails = _endpoint->_errorMessage;
+    if (_isSocketNonBlocking) {
+      // we don't care about the return value here because we were already
+      // unsuccessful in the connection attempt
+      cleanUpSocketFlags();
+    }
```
jsteemann, line 333: Should not be necessary if we do the whole fcntl stuff _after_ this check.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -72,6 +74,7 @@
 using namespace arangodb;
 using namespace arangodb::basics;
 using namespace arangodb::httpclient;
+using namespace std::chrono;
```
jsteemann, line 77: std::chrono is used in just 3 places in this file. Does it really justify a `using namespace` directive?

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,9 +307,30 @@ bool SslClientConnection::connectSocket() {
 
   _errorDetails.clear();
   _socket = _endpoint->connect(_connectTimeout, _requestTimeout);
+  if (_isSocketNonBlocking) {
+    _socketFlags = fcntl(_socket.fileDescriptor, F_GETFL, 0);
+    if (_socketFlags == -1) {
+      _errorDetails = "Socket file descriptor read returned with error " +
+                      std::to_string(errno);
+      _isConnected = false;
```
jsteemann, line 315: Call to `disconnectSocket()` is missing.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -302,9 +307,30 @@ bool SslClientConnection::connectSocket() {
...
+      return false;
+    }
+    if (fcntl(_socket.fileDescriptor, F_SETFL, _socketFlags | O_NONBLOCK) ==
+        -1) {
+      _errorDetails = "Attempt to create non-blocking socket generated error " +
+                      std::to_string(errno);
+      _isConnected = false;
+      return false;
```
jsteemann, line 323: Call to `disconnectSocket()` is missing.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +379,45 @@ bool SslClientConnection::connectSocket() {
...
+      if (_isInterrupted) {
+        break;
+      }
+      auto end = steady_clock::now();
+      if ((errorDetail != SSL_ERROR_WANT_READ &&
+           errorDetail != SSL_ERROR_WANT_WRITE) ||
+          duration_cast<seconds>(end - start).count() >= _connectTimeout) {
+        break;
+      }
+      std::this_thread::sleep_for(seconds(1));
```
jsteemann, line 398: Sleeping for 1s is too long, because we want to abort the loop relatively quickly if necessary.
What about 50ms instead?

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -467,6 +472,42 @@ function telemetricsShellReconnectGraphTestsuite() {
...
+      let options = internal.options();
+      let endpoint = arango.getEndpoint().replace(/\+vpp/, '').replace(/^http:/, 'tcp:').replace(/^https:/, 'ssl:').replace(/^vst:/, 'tcp:').replace(/^h2:/, 'tcp:');
+      const args = {
+        'javascript.startup-directory': options['javascript.startup-directory'],
+        'server.endpoint': endpoint,
+        'server.username': arango.connectedUser(),
+        'server.password': '',
+        'javascript.execute': file,
+        'client.failure-points': 'startTelemetricsForTest',
+        'log.output': "file://telemetrics.log",
```
jsteemann, line 488: Is this logfile actually needed? What directory is the file written to?

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +376,46 @@ bool SslClientConnection::connectSocket() {
...
+    auto start = std::chrono::steady_clock::now();
+    while ((ret = SSL_connect(_ssl)) == -1) {
+      errorDetail = SSL_get_error(_ssl, ret);
+      if (_isInterrupted) {
+        break;
+      }
+      auto end = std::chrono::steady_clock::now();
+      if ((errorDetail != SSL_ERROR_WANT_READ &&
+           errorDetail != SSL_ERROR_WANT_WRITE) ||
+          duration_cast<std::chrono::seconds>(end - start).count() >=
```
jsteemann, line 392:
```suggestion
          std::chrono::duration_cast<std::chrono::seconds>(end - start).count() >=
```

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -590,3 +654,7 @@ bool SslClientConnection::readable() {
 
   return false;
 }
+
+long SslClientConnection::cleanUpSocketFlags() {
```
jsteemann, line 658: my local version of `fcntl` returns an int (`man 2 fcntl`):
```
NAME
       fcntl - manipulate file descriptor

SYNOPSIS
       #include <unistd.h>
       #include <fcntl.h>

       int fcntl(int fd, int cmd, ... /* arg */ );
```

tests/js/client/server_parameters/test-telemetrics-true-auth.js
```diff
@@ -467,6 +472,44 @@ function telemetricsShellReconnectGraphTestsuite() {
...
+      let status = {};
+      while (true) {
+        status = internal.statusExternal(result.pid);
+        if (status.status === "TERMINATED" || numSecs >= 16) {
+          break;
+        }
+        internal.sleep(numSecs);
+        numSecs *= 2;
+      }
+      assertEqual(status.status, "TERMINATED", "couldn't leave shell immediately after executing script");
```
jsteemann, line 511: Readers of this code may wonder what this test actually checks.
It is unclear from this code that here we want the start sub-shell to close quickly **despite** any telemetrics requests or connection setups being in flight. There is also no mention here that the connection setup could take up to 30 seconds and we want to be able to interrupt it ASAP if the shell is otherwise ready to close.

jsteemann (approved):
LGTM

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -309,11 +313,35 @@ bool SslClientConnection::connectSocket() {
     return false;
   }
 
+  if (_isSocketNonBlocking) {
```
neunhoef, line 316: How is this supposed to work on Windows? Windows does not have `fcntl`. Therefore the jenkins build on Windows does not work. Does Windows even have non-blocking sockets?

neunhoef (changes requested):
We need a solution for Windows. Or at least put #ifdefs around it.

lib/SimpleHttpClient/SslClientConnection.cpp
Dronplane: That can be done on windows by 
`unsigned long mode = 1; ioctlsocket(fd, FIONBIO, &mode);`

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +377,44 @@ bool SslClientConnection::connectSocket() {
 
   ERR_clear_error();
 
-  int ret = SSL_connect(_ssl);
+  int ret = -1;
+  int errorDetail = -1;
+
+  if (_isSocketNonBlocking) {
+    auto start = std::chrono::steady_clock::now();
+    while ((ret = SSL_connect(_ssl)) == -1) {
```
Dronplane, line 385: NIT according to https://www.openssl.org/docs/man1.1.1/man3/SSL_connect.html the values should be "<0" not necessary "-1"

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -343,15 +377,44 @@ bool SslClientConnection::connectSocket() {
...
+  }
 
   if (ret != 1) {
     _errorDetails.append("SSL: during SSL_connect: ");
 
-    int errorDetail;
     long certError;
 
-    errorDetail = SSL_get_error(_ssl, ret);
     if ((errorDetail == SSL_ERROR_WANT_READ) ||
```
Dronplane, line 418: If in non-blocking mode we will reach this line by the timeout - errorDetail would be exactly SSL_ERROR_WANT_READ/WRITE  and we will return true. But connection might be still not established. Is this expected?

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -119,6 +111,18 @@ void TelemetricsHandler::fetchTelemetricsFromServer() {
       _httpClient = clientManager.getConnectedClient(true, false, false, 0);
 
       if (_httpClient != nullptr && _httpClient->isConnected()) {
+        std::string role;
+        Result result;
+        std::tie(result, role) =
+            clientManager.getArangoIsCluster(*(_httpClient.get()));
```
jsteemann, line 117:
```suggestion
        auto [result, role ] =
            clientManager.getArangoIsCluster(*(_httpClient.get()));
```

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -309,11 +315,34 @@ bool SslClientConnection::connectSocket() {
     return false;
   }
 
+  if (_isSocketNonBlocking) {
+#if defined(__linux__) || defined(__APPLE__)
```
jsteemann, line 319: It feels really weird that a part of the work is now done in `setSocketToNonBlocking()` and a part is done here.
It would be a lot more sensible to move this part into `setSocketToNonBlocking()` as well.

lib/SimpleHttpClient/SslClientConnection.cpp
```diff
@@ -590,3 +653,40 @@ bool SslClientConnection::readable() {
...
+    _errorDetails = "Attempt to create non-blocking socket generated error " +
+                    std::to_string(WSAGetLastError());
+    return false;
+  }
+#endif
+  return true;
+}
+
+bool SslClientConnection::cleanUpSocketFlags() {
+#if defined(__linux__) || defined(__APPLE__)
```
jsteemann, line 676: Let's add an assertion here, so that we know it is only called if we are actually using a non-blocking socket.
```suggestion
  TRI_ASSERT(_isSocketNonBlocking);
#if defined(__linux__) || defined(__APPLE__)
```

jsteemann (commented):
LGTM

client-tools/Shell/TelemetricsHandler.cpp
```diff
@@ -335,18 +337,15 @@ void TelemetricsHandler::getTelemetricsInfo(velocypack::Builder& builder) {
 
 void TelemetricsHandler::arrangeTelemetrics() {
   try {
+#ifndef ARANGODB_ENABLE_MAINTAINER_MODE
```
Dronplane, line 340: Is it needed here? That makes check on line 348 a bit useless as it is always true in non-maintainer mode.

Dronplane (approved):
LGTM

client-tools/Shell/TelemetricsHandler.cpp
cpjulia: It's not needed, but actually because the value is set to true already in the constructor for non-maintainer mode, but the function that fetches temeletrics restricts is to only being true if it's single server or it's a coordinator, so the value might have to change there
I removed it

PR #18829, opened 2023-05-02
Increase chaos client timeout

Dronplane:
### Scope & Purpose

As we see sometimes chaos test clients run into "Execution deadline reached" error. This leads to inconsistent chaos tests results. 10s is not enough to finish executions, so this PR increases it to 60s. 

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #18857, opened 2023-05-04
Fix chaos timeout

MBkkt:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/18829

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


js/client/modules/@arangodb/test-helper.js
```diff
@@ -268,7 +268,8 @@ const buildCode = function(key, command, cn, duration) {
   let file = fs.getTempFile() + "-" + key;
   fs.write(file, `
 (function() {
-require('internal').SetGlobalExecutionDeadlineTo((${duration} + 10) * 1000);
```
dothebart, line 271: if you do this, you need to change the grace period of the watching arangosh as well.

MBkkt: It's backport

dothebart: please create a subsequent PR then. the result aggregation behaviour is probably better if the subshells terminate on their own.

Dronplane: @dothebart  Could you please elaborate a bit more:
1. What and where should we change more?
2. Is it related to this backport only or we need to do something in devel too?
Or if it is not much effort, could you please just push necessary changes here? (And open a devel PR if necessary). 
As for now these changes seems to be working in devel and we don't see targeted errors anymore. But maybe we are missing something.

dothebart: If I remember this correctly the deadline of the spawned arangoshs is calculated against a timeout in the testing.js arangosh by the specified `duration`. If you increase the sluggishness of the deadline to the duration, you need to adjust that in the calling suite as well.

Dronplane: As I see, after clients started
`internal.sleep(duration);` is done e.g. no adjustments (+10s) are made. So I assume still nothing is needed. Or I`m missing something?

Dronplane: duration - is expected arangosh instance run.
Deadline - is this duration + some extent to accommodate possible delay

dothebart: ok, so if the time is too short, why then not increase the digit on the outside, but somewhere inside?

KVS85 (approved):

PR #18867, opened 2023-05-05
[GORDO] Add status details in every conductor state

jvolmer:


markuspf (approved):
This particular change is OK, but it showed up a race condition between the creation of the status document in the historic pregel collection and the client asking for status via the execution number created in `startExecution`; the ticket to fix this is GORDO-1643

PR #18878, opened 2023-05-08
[CINFRA] Wait for follower to resign

apetenchea:
### Scope & Purpose

The *follower_resigning_while_acquiring_snapshot_concurrently* test works by starting a snapshot transfer in a separate thread, while resigning the follower concurrently from the main test thread. In the event of [high system load](https://jenkins01.arangodb.biz/job/arangodb-matrix-pr-mac-onlycatch.x86-64/24681/EDITION=community,limit=mac&&test&&x86-64/artifact/testfailures.txt), it is possible that the main thread is starved and never resigns the follower, even though the thread doing the snapshot transfer was started and called `notify_one`. This can lead to a stack overflow.  
The current PR introduces a limit on the number of batches, such that the snapshot can't go on indefinitely without waiting for the main thread to resign the follower.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/RestHandler/RestDocumentStateHandler.cpp
+++ b/arangod/RestHandler/RestDocumentStateHandler.cpp
@@ -115,7 +115,7 @@ RestStatus RestDocumentStateHandler::handleGetRequest(
 
   auto const& verb = suffixes[1];
   if (verb == "snapshot") {
-    if (suffixes.size() != 3) {
+    if (suffixes.size() < 3) {
       generateError(
           rest::ResponseCode::BAD, TRI_ERROR_HTTP_BAD_PARAMETER,
           "expect GET /_api/document-state/<state-id>/snapshot/<action>");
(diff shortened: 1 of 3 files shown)


mchacki (approved):
LGTM.
Before merge I would like to discuss if we need any kind of Production protection against this or if this only happens in the test scenario.

apetenchea:
> LGTM. Before merge I would like to discuss if we need any kind of Production protection against this or if this only happens in the test scenario.

**In theory**, it could happen in production, but only in theory. To understand why is that, I'll give a summary of how snapshot transfer currently works:

1. The follower initiates the transfer: that is a call to `DocumentFollowerState::handleSnapshotTransfer`. This returns a `Future`.
2. Inside this method, the follower has to fetch the snapshot in batches. As you probably guessed, fetching a batch involves another network request, therefore it returns a `Future`. 
3. The futures returned by batches are chained, something that looks like: `std::move(snapshotFuture).then(/* fetch next batch */)`.

If the `snapshotFuture` is already resolved, the `then` lambda is executed immediately. In other words, if the network runs "lightning fast", such that batches are fetched instantaneously, we could run into a stack overflow (considering we get more then 2000 such continuous batches).  
In the C++ test scenario above, the problem appeared because batches were returned instantaneously. In production, there is always network in between.

PR #18883, opened 2023-05-09
[PREGEL] Fix performance regression

markuspf:
### Scope & Purpose

This should fix a 50% performance regression in Pregel between 3.10 and 3.11/devel.


neunhoef (approved):
LGTM

hkernbach (approved):

PR #18887, opened 2023-05-09
[3.11] Fix handling of premature follower trx abortion

mpoeter:
### Scope & Purpose

When we have a query that results in a failure this will cause the leaders to abort the transaction on the followers. However, if the followers have transactions that are shared with leaders of other shards, and if one of those leaders has not yet seen the error, then it will happily continue to replicate to that follower. But if the follower has already aborted the transaction, then it will reject the replication request. Previously this would cause the follower to be dropped, but now this should be handled gracefully.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] Backports
  - [x] Backport for devel #18881

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,7 +14,7 @@ v3.11.0 (XXXX-XX-XX)
 
 v3.11.0-beta.1 (2023-05-07)
 ---------------------------
-    
+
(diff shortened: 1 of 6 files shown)


neunhoef (approved):
LGTM

PR #18861, opened 2023-05-05
[BTS-1343]

cpjulia:
### Scope & Purpose

This PR removes an error message of endpoint being invalid when attempting to fetch telemetrics from the servers by having the invalid endpoint as a starting point. As telemetrics is transparent to the end user, the error message interferes with the user experience if keeps appearing. 

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1343
- [ ] Design document:

PR #18908, opened 2023-05-09
3.11 backport for BTS-1343

cpjulia:
### Scope & Purpose

Backport to https://github.com/arangodb/arangodb/pull/18861.

This PR removes an error message of endpoint being invalid when attempting to fetch telemetrics from the servers by having the invalid endpoint as a starting point. As telemetrics is transparent to the end user, the error message interferes with the user experience if keeps appearing.


jsteemann (approved):
LGTM

apetenchea (approved):

PR #16244, opened 2022-05-10
[BTS-680] Try to fix data race

jvolmer:
### Scope & Purpose

This PR tries to fix a data race that possibly causes BTS-680 by
- checking the trigger with the same mutex with which data is added to the vector
- exiting wait of condition variable when data is empty

Currently it is not clear whether this fixes the bug.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.9: *(Please link PR)*
  - [ ] Backport for 3.8: *(Please link PR)*
  - [ ] Backport for 3.7: *(Please link PR)*

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-680
- [ ] Design document:

PR #13207, opened 2020-12-11
Bug fix/bts 213.2

mchacki:
### Scope & Purpose

In all of our Transaction starters in the Coordinator, (AQL, JS, HttpAPI) there is a chance that on an
exclusive lock on a shard, we can get into a distributed dead-lock situation, which is resolved by rather long timeouts.
This PR tries to remove the DeadLock situation by fixing the common cases.
So now if you are using exclusive locks on Collections with more then one Shard it is muss less likely that a Lock Timeout (ErrorNum 18 "lock timeout") is reported to the client code.
It is not impossible to get the lock-timeout, as in case of an exclusive lock, you can still have other ongoing operations that can hold a write or exclusive lock you need to wait for, that simply take a large amount of time. But the chance to get into such a situation with "fast" exclusive lock operations is highly reduced.

Side-Effect:
In AQL we can now do a fan-out parallel lock instead of sequential locking, which may speed up non-blocked queries with many servers involved.

- [x] :hankey: Bugfix (requires CHANGELOG entry)
- [ ] :pizza: New feature (requires CHANGELOG entry, feature documentation and release notes)
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification
- [x] :book: CHANGELOG entry made

#### Backports:

- [ ] No backports required
- [ ] Backports required for: *(Please specify versions and link PRs)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket number: BTS-213
- [ ] Design document: 

### Testing & Verification

*(Please pick either of the following options)*

- [ ] This change is a trivial rework / code cleanup without any test coverage.
- [ ] The behavior in this PR was *manually tested*
- [ ] This change is already covered by existing tests, such as *(please describe tests)*.
- [x] This PR adds tests that were used to verify all changes:
  - [ ] Added new C++ **Unit tests**
  - [x] Added new **integration tests** (e.g. in shell_server / shell_server_aql)
  - [ ] Added new **resilience tests** (only if the feature is impacted by failovers)
- [ ] There are tests in an external testing repository:
- [ ] I ensured this code runs with ASan / TSan or other static verification tools

Link to Jenkins PR run:

http://jenkins.arangodb.biz:8080/job/arangodb-matrix-pr/13277/

### Documentation

> All new features should be accompanied by corresponding documentation. 
> Bugs and features should furthermore be documented in the CHANGELOG so that
[...]

PR #18914, opened 2023-05-10
[BTS-1373] fix error in test code

goedderz:
### Scope & Purpose

Fix a possible crash in a failure test, when the race controller is reset unexpectedly early.

This strictly only affects failure test code, which is not part of production executables. It also shouldn't occur when the tests run as expected, without "external" failures.

Previous PRs this is based on:
- https://github.com/arangodb/arangodb/pull/13207
- https://github.com/arangodb/arangodb/pull/16244

### DIFF
--- a/lib/Basics/DebugRaceController.cpp
+++ b/lib/Basics/DebugRaceController.cpp
@@ -24,7 +24,9 @@
 #ifdef ARANGODB_ENABLE_MAINTAINER_MODE
 
 #include "DebugRaceController.h"
+
 #include "ApplicationFeatures/ApplicationServer.h"
+#include "Basics/ScopeGuard.h"
 
 using namespace arangodb;
 using namespace arangodb::basics;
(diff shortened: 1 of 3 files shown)


goedderz:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/29629/

lib/Basics/DebugRaceController.cpp
```diff
@@ -24,47 +24,59 @@
...
 auto DebugRaceController::waitForOthers(
     size_t numberOfThreadsToWaitFor, std::any myData,
-    arangodb::application_features::ApplicationServer const& server) -> bool {
-  std::unique_lock<std::mutex> guard(_mutex);
-  if (!_didTrigger) {
+    application_features::ApplicationServer const& server)
+    -> std::optional<std::vector<std::any>> {
+  auto notifyGuard =
+      scopeGuard([this]() noexcept { _condVariable.notify_all(); });
+  {
```
jsteemann, line 60: nit: extra block scope here looks unnecessary, but shouldn't do any harm.

lib/Basics/DebugRaceController.cpp
```diff
@@ -24,47 +24,59 @@
...
+    std::unique_lock<std::mutex> guard(_mutex);
+
+    if (didTrigger(guard, numberOfThreadsToWaitFor)) {
+      return std::nullopt;
```
jsteemann, line 64: Why are we returning `std::nullopt` here in case didTrigger() returns true, but return `_data` below if didTrigger() returns true there? This may be correct, but I don't understand it.

lib/Basics/DebugRaceController.cpp
goedderz: Yes, I added it just to emphasize that the mutex is unlocked before the scope guard is executed.

lib/Basics/DebugRaceController.cpp
goedderz: The goal is to get exactly `numberOfThreadsToWaitFor` to get into the code below, add their data, and then continue. Any thread above that number will run into the early return here.

The check below is for a different reason: If the wait finishes, usually `didTrigger` is true, i.e. `_data.size() == numberOfThreadsToWaitFor`, and we return the data. But if the wait is aborted early, e.g. due to a reset (which clears `_data`), then we return `nullopt` instead.

jsteemann (approved):
LGTM

PR #18928, opened 2023-05-10
Hotbackup Deadlock and Consistency fixes.

maierlars:
### Scope & Purpose

Before this PR, when the user requested a HotBackup, the ArangoDB cluster waited for all in-progress transactions to be committed. This lead to deadlocks.
Also, the HotBackup snapshot of the RocksDB database did only include some portion of the WAL and was not guaranteed to include enough of the WAL to recover MerkleTrees. This would lead to MerkleTrees being inconsistent with the database and potential data loss from replication. 

After this PR, requesting a HotBackup merely prevents transactions from being committed.

The HotBackup process has to take two snapshots: One of the RocksDB store, and one of the ArangoSearch store.

Consistency is achieved by taking a snapshot of RocksDB with a large enough portion of the WAL so that the recovery process after the DBServer restart can replay all necessary operations on ArangoSearch and MerkleTrees.

Unfortunately, to achieve taking a snapshot of RocksDB including the WAL files, a small patch is applied to our RocksDB fork. Future work can include upstreaming that patch.

Enterprise: https://github.com/arangodb/enterprise/pull/1267


arangod/IResearch/IResearchDataStore.h
```diff
@@ -446,6 +457,8 @@ class IResearchDataStore {
 
   void recoveryCommit(uint64_t tick);
 
+  IResearchFeature* getIResearchFeature() const { return _asyncFeature; }
+
```
MBkkt, line 461: features could be accessed from server
```suggestion
```

arangod/IResearch/IResearchDataStore.h
```diff
@@ -105,6 +107,11 @@ class IResearchDataStore {
       // for now we require that each index has its own snapshot
       TRI_ASSERT(_snapshot);
     }
+
+    [[nodiscard]] TRI_voc_tick_t tick() const noexcept {
+      return _snapshot->tick();
+    }
+
```
MBkkt, line 114: _snapshot is public here, is it really necessary?

```suggestion
```

3rdParty/rocksdb
MBkkt, line 1: It looks like all archived files will be copied
Is it impossible to get max or min tick for archived WAL file?

arangod/IResearch/IResearchDataStore.h
maierlars: yes, I like it more that way.

3rdParty/rocksdb
maierlars: and then do what with that?

maierlars: there is more than just iresearch.

MBkkt: I don't understand your point.
WAL files with tick smaller than hotbackup flush subscription tick could be already removed when we trying to copy all rocksdb wal files.
It depends on our luck, so I don't understand point to always copy them instead of always don't copy them.

Maybe I misunderstood something, so could you explain it please?

Dronplane: To account everyone we could iterate over all flush subscriptions and determine min-needed-tick just like it is done when we release ticks. That could be useful as we have `--rocksdb.wal-file-timeout` default to 10 seconds and during heavy load  we potentially could copy significant amount of not needed files.

maierlars: There is no copy. It is hard linked. Only the active log file is modified by rocksdb.

What are your concerns?

MBkkt: > There is no copy. It is hard linked. Only the active log file is modified by rocksdb.

According to doxygen function documentation it's copied.
```
  // When a checkpoint is created,
  // (1) SST and blob files are hard linked if the output directory is on the
  // same filesystem as the database, and copied otherwise.
  // (2) other required files (like MANIFEST) are always copied.
```
But maybe it's outdated and ofc it sounds more logical that archived files should be hard linked and only active should be copied.
But could you please give reference to source code that do this or test that check this.

> What are your concerns?

Anyway if we back to our discussion, I don't understand for what reason we want to make hotbackup bigger than needed (yes, hard links will be shared, but without hotbackup this files would be removed in the future).
The only valid reason for me that it's difficult to implement such approach. Is it?

Dronplane: >There is no copy. It is hard linked

I was under impression that we now include WAL files in the hotbackup. So eventually they would be copied anyway. As there is not much sense to keep "backup" on the same "volume" as the database

maierlars: > The only valid reason for me that it's difficult to implement such approach. Is it?

How about not worth it?

neunhoef: In typical applications WAL size is negligible in comparison to data size, so "larger backup" is not an argument. If we keep too many WAL files, then this is a bug elsewhere.
Simplicity is what trumps everything else here. The fact that we can now include WAL files in hotbackup and rely on recovery to get consistency between RocksDB and search is a **phantastic achievement** w.r.t. simplicity. Let's end this fruitless discussion here and now.

MBkkt: @neunhoef I agreed it's nice solution, but I don't understand how it's related to discussion.
And as I already wrote difficulty of implementation this suggestion is good argument to don't do it at least now.
But for now no one write: "yes it's difficult"

arangod/RocksDBEngine/Methods/RocksDBTrxMethods.cpp
```diff
@@ -31,6 +31,8 @@
 #include "RocksDBEngine/RocksDBTransactionState.h"
 #include "Statistics/ServerStatistics.h"
 #include "Metrics/Counter.h"
+#include "Transaction/Manager.h"
+#include "Transaction/ManagerFeature.h"
```
jsteemann, line 35: nit: uncommon order of include files. Are the additional includes actually still needed? I guess they are leftovers from some previous state of the PR.

arangod/StorageEngine/HotBackup.cpp
```diff
@@ -33,6 +33,7 @@
 #ifdef USE_ENTERPRISE
 #include "Enterprise/RocksDBEngine/RocksDBHotBackup.h"
 #include "Enterprise/StorageEngine/HotBackupFeature.h"
+#include "Logger/LogMacros.h"
```
jsteemann, line 36: nit: potentially not necessary

arangod/Transaction/Manager.h
```diff
@@ -35,6 +35,7 @@
 #include "VocBase/AccessMode.h"
 #include "VocBase/Identifiers/TransactionId.h"
 #include "VocBase/voc-types.h"
+#include "Basics/ReadLocker.h"
```
jsteemann, line 38: nit: uncommon include order

arangod/Transaction/Manager.cpp
```diff
@@ -1020,6 +1018,7 @@ Result Manager::statusChangeWithTimeout(TransactionId tid,
 
 Result Manager::commitManagedTrx(TransactionId tid,
                                  std::string const& database) {
+  READ_LOCKER(guard, _hotbackupCommitLock);
```
jsteemann, line 1021: Just wondering if other places than `transaction::Manager::commitManagedTrx(...)` should also prevent transactions from committing. At least this is not the only place that will commit a transaction, but is only used to commit a streaming transaction.

As far as I can see, the `_hotbackupCommitLock` lock is currently acquired by `transaction::Manager::holdTransactions()` in write mode. That will block the `transaction::Manager::commitManagedTrx()` from making progress and committing anything. Fine.
But as mentioned `commitManagedTrx` is used only to commit streaming transactions. It is not used to commit any other types of transactions, for example standalone AQL queries. 
So if I am not wrong, AQL queries will not be prevented from committing even if the write-lock on `_hotbackupCommitLock` is acquired by `holdTransactions()`.
This may or may not be intentional, just wanted to clarify it.

arangod/Transaction/Manager.cpp
```diff
@@ -108,7 +108,6 @@ void Manager::registerTransaction(TransactionId transactionId,
   if (!isReadOnlyTransaction && !isFollowerTransaction) {
     LOG_TOPIC("ccdea", TRACE, Logger::TRANSACTIONS)
         << "Acquiring read lock for tid " << transactionId.id();
-    _rwLock.lockRead();
     _nrReadLocked.fetch_add(1, std::memory_order_relaxed);
```
jsteemann, line 111: Only slightly related but may be addressed anyway by this PR:
`_nrReadLocked` is only counted up and down, but serves no other purpose than its value being reported in `TRACE` log messages. I guess that this variable was originally added to debug problems during hot backup (and its deadlocks). I would consider it superfluous by now, so IMHO we can as well remove it.
But I am happy if the variable is kept for now if it is still considered useful.

arangod/Transaction/Manager.cpp
maierlars: Interesting, is this different on dbservers? This code was previously used on dbservers to stop all transactions. Maybe it is flawed as well there .

maierlars: I see. Is this different on dbservers? Is registerTransaction called for all transactions? Maybe it is already broken

arangod/Transaction/Manager.cpp
maierlars: Gone.

arangod/Transaction/Manager.cpp
maierlars: Should be fixed now.

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -4531,7 +4534,7 @@ arangodb::Result hotBackupCoordinator(ClusterFeature& feature,
       auto releaseLocks = scopeGuard([&]() noexcept {
         try {
           hotbackupCancelAsyncLocks(pool, lockJobIds, lockedServers);
-          unlockDBServerTransactions(pool, backupId, lockedServers);
+          unlockServersTrxCommit(pool, backupId, lockedServers);
```
neunhoef, line 4537: Same argument as above.

arangod/RocksDBEngine/Methods/RocksDBTrxMethods.cpp
```diff
@@ -203,6 +203,8 @@ Result RocksDBTrxMethods::triggerIntermediateCommit() {
   LOG_TOPIC("0fe63", DEBUG, Logger::ENGINES) << "executing intermediate commit";
 #endif
 
+  // We have to make sure that no intermediate commit happens while we
```
neunhoef, line 206: Is this comment still appropriate?

arangod/Transaction/Manager.cpp
```diff
@@ -123,14 +119,11 @@ void Manager::unregisterTransaction(TransactionId transactionId,
                                     bool isReadOnlyTransaction,
                                     bool isFollowerTransaction) {
   // always perform an unlock when we leave this function
-  auto guard = scopeGuard([this, transactionId, &isReadOnlyTransaction,
+  auto guard = scopeGuard([transactionId, &isReadOnlyTransaction,
```
neunhoef, line 122: This guard seems unnecessary, since it does not do anything except TRACE logging something which is no longer relevant. Remove?

neunhoef (changes requested):
Review not yet complete, but already some comments, publishing for earlier visibility.

Dronplane (commented):
Maybe I've missed that, but do we test here that ArangoSearch view documents are all present (and nothing duplicated) after restoring hotbackup?

neunhoef (approved):
Apart from the things I have so far reported: LGTM. I will now move on to the enterprise part.

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -4483,19 +4485,20 @@ arangodb::Result hotBackupCoordinator(ClusterFeature& feature,
...
     // 15, 30 and 60 to try before the default timeout of 120s has been reached.
     double lockWait(15.0);
     while (steady_clock::now() < end && !feature.server().isStopping()) {
-      result = lockDBServerTransactions(pool, backupId, dbServers, lockWait,
-                                        lockedServers);
+      result = lockServersTrxCommit(pool, backupId, serversToBeLocked, lockWait,
+                                    lockedServers);
       if (!result.ok()) {
-        unlockDBServerTransactions(pool, backupId, lockedServers);
+        unlockServersTrxCommit(pool, backupId, lockedServers);
```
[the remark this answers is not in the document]
maierlars: I'm against changing this in this PR:
1. This PR is already big enough and it is not a new issue.
2. The code in question is a huge mess. It is more likely for me to break it than fix anything useful. It has to be reimplemented from ground up.
3. The existing logic is good enough for dbservers, I would argue it is good enough for coordinators. There was probably a reason for why to do it like that.

3rdParty/rocksdb
neunhoef: Let me try to explain the reasoning here in a bit more detail: In the process of this project we found out (which I at least did not know beforehand) that RocksDB does not even tell us the **exact** sequence number the hotbackup will "contain" in the sense of what state we see once RocksDB recovery has been run after a restore. Therefore it is essentially **inevitable** that we include at least **some** WAL files (and fix the bug in restore which we so far had to not restore those WAL files).
We have various mechanisms with which we track which WAL files need to be kept (alive or in `archive`) during normal operations (flush subscriptions, and other places which keep a sequence number). Therefore, the easiest, simplest and almost certainly most reliable way to achieve hotbackup consistency (with search and other subsystems like Merkle trees) appears to be to include the **full** WAL at the time when the hotbackup is taken.
"Simplicity" is here the main argument, "complexity" is our largest enemy. Hotbackup and its relationship to all other parts of the system is already (somewhat inherently) complex. Anything we can do to reduce complexity is very welcome. And this solution to include the full WAL is exactly of this flavour: Simplicity first!
As mentioned above, under normal circumstances, this will not incur any non-negligible cost, since the WAL will in general be tiny in comparison to the actual data.
If this assumption is not the case, we have a bug elsewhere (which we did indeed have in the past and might still have). But in this case this bug needs to be fixed.

arangod/Cluster/ClusterMethods.cpp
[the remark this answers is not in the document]
joerg84: Can we adress this in a BTS follow-up?

joerg84 (approved):
LGTM, the open point(s) can imo be addressed later

markuspf (approved):

joerg84:
> Maybe I've missed that, but do we test here that ArangoSearch view documents are all present (and nothing duplicated) after restoring hotbackup?

Afaik it was

PR #18932, opened 2023-05-10
upgrade libunwind to version 1.7

jsteemann:
### Scope & Purpose

Upgrade bundled version of libunwind to latest stable, i.e. 1.7-rc

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


markuspf:
I think libunwind now comes with it's own `CMakeLists.txt`, so it should be sufficient to `add_subdirectory` instead of the `ExternalProject`.

Further silly question: The last time I looked at upgrading libunwind, I was wondering whether it couldn't be just removed from our codebase and use whichever version that comes with the Docker build container. Is there anything standing in the way of such a move?

jsteemann:
> Further silly question: The last time I looked at upgrading libunwind, I was wondering whether it couldn't be just removed from our codebase and use whichever version that comes with the Docker build container. Is there anything standing in the way of such a move?

That may work when building with the Docker build container, but what about building arangod locally, outside the Docker build container? How would stacktraces be created then?

jsteemann:
> does libmusl have similar support for all this stack printing business as glibc has?

No

jsteemann:
> I think libunwind now comes with it's own `CMakeLists.txt`, so it should be sufficient to `add_subdirectory` instead of the `ExternalProject`.

We can't.
The CMakeLists.txt file in libunwind can be used for Windows only. Running it with other compilers than Visual Studio will make it error out: https://github.com/libunwind/libunwind/blob/v1.7-stable/CMakeLists.txt#L97
Apart from that, it doesn't seem to be well-maintained, e.g. not even the version number in it is correct.

markuspf:
> > I think libunwind now comes with it's own `CMakeLists.txt`, so it should be sufficient to `add_subdirectory` instead of the `ExternalProject`.
> 
> We can't. The CMakeLists.txt file in libunwind can be used for Windows only. Running it with other compilers than Visual Studio will make it error out: https://github.com/libunwind/libunwind/blob/v1.7-stable/CMakeLists.txt#L97 Apart from that, it doesn't seem to be well-maintained, e.g. not even the version number in it is correct.

Jeez, ok nevermind then. Thanks for checking.

jsteemann:
> Shouldn't libunwind be available as a package for your build environment?

Yes, there is one.
But I was actually after a simple library upgrade and not after a CMake change that may also lead to us using different versions of libunwind during development and releases.
Feel free to implement the change you are suggesting if you prefer it. But my goal for this PR is just to unblock ourselves quickly.

markuspf (approved):
LGTM

PR #18911, opened 2023-05-10
[BTS-1350]

cpjulia:
### Scope & Purpose

 Fixed imprecision in index info in telemetrics object for cluster. When the collection also had an arangosearch view, telemetrics showed imprecise index values because the view was being considered as an index of type arangosearch, and its object wouldn't contain expected fields that would be found in other indexes' objects.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: https://github.com/arangodb/arangodb/pull/18933

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1350
- [ ] Design document:

PR #18933, opened 2023-05-11
3.11 backport for BTS-1350

cpjulia:
### Scope & Purpose

Backport for https://github.com/arangodb/arangodb/pull/18911.
Fixed imprecision in index info in telemetrics object for cluster. When the collection also had an arangosearch view, telemetrics showed imprecise index values because the view was being considered as an index of type arangosearch, and its object wouldn't contain expected fields that would be found in other indexes' objects.


neunhoef (approved):
LGTM

gnusi (changes requested):
We need the information about links as well.

gnusi (approved):

jsteemann (approved):
LGTM

PR #18943, opened 2023-05-11
[BTS-1181] [3.11] Fix data race on the collection list in TransactionState

mpoeter:
### Scope & Purpose

This race could happen in the cluster when a collection is sharded by a different attribute than `_key` and a document lookup by _key is performed. This requires the lookup to be sent to each shard, since we cannot deduce the shard based on the `_key` value. If this lookup is done as part of a stream transaction, and if the collection has not been declared in the transaction's `read` list, then the collection is added lazily. Previously this would result in a race if multiple shards are located on the same server.

For transactions with `allowDirtyReads` set to true, read collections are always added lazily, which made this race more likely.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] Backports
  - [x] Backport for devel: #18941
  - [x] Backport for 3.10: #18944
  - [x] Backport for 3.9: #18945


MBkkt (approved):

jsteemann (approved):
The new lock is only used in some situations when `_collections` is accessed, but it is not used in others.
For example, the method TransactionState::allCollections(...), implemented in this header file, does not use the lock.
As `_collections` is a protected member of TransactionState, we also need to make sure that no derived classes use `_collections` without the lock, or, preferrably, make `_collections` private.

**Update**: discussed with @mpoeter and we agreed that it is safe and good to go with just this fix. He will also look into cleaning up access to `_collections` in a follow-up PR.

Issue #18919, opened 2023-05-10
Log file on Windows becomes very large because of `TestFileIOHeartbeat`


larsborn:
## My Environment

* __ArangoDB Version__: 3.9.2
* __Deployment Mode__: Single Server
* __Deployment Strategy__:  Manual Start
* __Configuration__: -
* __Infrastructure__: own
* __Operating System__: Windows 10
* __Total RAM in your machine__: 32Gb
* __Disks in use__: SSD
* __Used Package__: Windows Installer

## Component, Query & Data

__Affected feature__: Logging

## Steps to reproduce

1. terminate the ArangoDb service when it created the `TestFileIOHeartbeat` file but didn't delete it yet
2. Next time, the server will try to create it, it'll fail and produce a log line
3. This blew up log file size to a couple of Gb within a few days

__Problem__:
```
CreateFileW(filename, GENERIC_READ | GENERIC_WRITE, FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, CREATE_NEW, 0, NULL);
```
Accoding to MSDN, the parameter `CREATE_NEW` leads to failure when the file already exists:

> Creates a new file, only if it does not already exist.
> If the specified file exists, the function fails and the last-error code is set to ERROR_FILE_EXISTS (80).

On a side node: it was necessary to reproduce the error with a separate program because arangod hid the error code:

```c
#include <stdio.h>
#include <windows.h>

int main()
{
    HANDLE fp;
    fp = CreateFileW(L"C:\\ProgramData\\ArangoDB\\TestFileIOHeartbeat", GENERIC_READ | GENERIC_WRITE, FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, CREATE_NEW, 0, NULL);
    if (fp == INVALID_HANDLE_VALUE) {
        unsigned int error = GetLastError();
        printf("Error: 0x%x", error);
    } else {
        printf("Success");
    }

    return 0;
}
```

__Expected result__: 
1. emission of the error code of `80` in the logs.
2. graceful handling of the file being present

apetenchea:
The pull request addressing this issue has been merged https://github.com/arangodb/arangodb/pull/18958

Thanks again for your help!

PR #18958, opened 2023-05-11
Add the functionality for truncating files on Windows

apetenchea:
### Scope & Purpose

This PR addresses an issue caused by the ever growing log files on Windows, due to errors reported from the `IOHeartbeatThread`. The `IOHeartbeatThread` feature periodically checks that we are able to write to disk. It does so by creating a file, writing data to it, and finally removing it. In case the file already exists (for example, due to the process being restarted), the `IOHeartbeatThread` is unable to create the file again on Windows and it gets stuck. Every other second, it prints the following message:

`IOHeartbeat: exception when writing test file: write failed for file 'TestFileIOHeartbeat': No error`

As you can see, there are two problems:

1) Reporting the error of the last system call does not work on Windows. Hence, it defaults to `No error`, which only confuses the user. This PR adds the necessary Windows code for reporting the last system call error. This is an effort to improve the overall error reporting of the system.
2) The thread gets stuck in this state and produces a log every other second, thus yielding huge log files as time goes by. That was due to the way `TRI_createFile` used to work, being limited to either creating a non-existing file (_CREATE_NEW_) or opening an already existing file (_OPEN_ALWAYS_), but the code for overwriting an existing file was never added (_CREATE_ALWAYS_). You can think of it as the equivalent of _O_TRUNC_ on Linux, which is exactly how the `IOHeartbeatThread` creates the file on Linux.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/18963
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/18964
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/18966

#### Related Information

Reported in issue: https://github.com/arangodb/arangodb/issues/18919

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue: https://github.com/arangodb/arangodb/issues/18919
- [ ] Design document:


lib/Basics/error.cpp
```diff
@@ -49,7 +58,19 @@ std::string_view TRI_last_error() {
   ErrorCode err = LastError._number;
 
   if (err == TRI_ERROR_SYS_ERROR) {
+#if defined(_WIN32)
+    if (LastError._sys == 0) {
+      return {};
+    }
+
+    static CHAR messageBuffer[sizeof(TCHAR) * 1024] = {0};
```
jsteemann, line 66: It looks like using a static buffer is not thread-safe, in case `TRI_last_error()` is called from multiple threads in parallel, and they all write to the same buffer.
I guess we should remove the `static` declaration here.

apetenchea: Good catch! I changed the parameters of `FormatMessageA` such that it allocates the buffer itself. I had to change the return type of `TRI_last_error` to `std::string` in order to support the Windows counterpart. As we discussed, it should not impact performance, as it is used in case an error occurs.

neunhoef (approved):
LGTM

lib/Basics/error.cpp
```diff
@@ -45,22 +54,44 @@ thread_local ErrorContainer LastError;
...
+    std::size_t size = FormatMessageA(
+        FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM |
+            FORMAT_MESSAGE_IGNORE_INSERTS,
+        nullptr, LastError._sys, 0, (LPSTR)&messageBuffer, 0, nullptr);
+    if (size == 0) {
+      return "failed to get error message, the error code is " +
+             std::to_string(LastError._sys);
+    }
+    std::string message(messageBuffer, size);
+    LocalFree(messageBuffer);
```
jsteemann, line 75: This is not exception-safe, i.e. if `std::string(messageBuffer, size);` throws due to `std::bad_alloc`, the messageBuffer allocation is leaked. Would be better to free it using a scopeGuard or another RAII mechanism that is guaranteed to be executed.

apetenchea: Thanks for pointing that out! I have revised the code to include a scope guard. We should be safe now, even if an exception is thrown. Please take a look at the updated changes and let me know if you have further feedback or suggestions.

jsteemann (approved):
LGTM

PR #18971, opened 2023-05-12
[GORDO-1642] [3.11] Try and address some obvious timewasters

markuspf:
### Scope & Purpose

Backport of #18970

### DIFF
--- a/arangod/Pregel/IncomingCache.cpp
+++ b/arangod/Pregel/IncomingCache.cpp
@@ -105,9 +105,9 @@ void InCache<M>::storeMessage(PregelShard shard, std::string_view vertexId,
 template<typename M>
 ArrayInCache<M>::ArrayInCache(containers::FlatHashSet<PregelShard> localShards,
                               MessageFormat<M> const* format)
-    : InCache<M>(format), _localShards(localShards) {
+    : InCache<M>(format), _localShards(std::move(localShards)) {
   // one mutex per shard, we will see how this scales
-  for (PregelShard pregelShard : _localShards) {
+  for (PregelShard const& pregelShard : _localShards) {
     this->_bucketLocker[pregelShard];
     _shardMap[pregelShard];
   }
@@ -218,9 +218,11 @@ template<typename M>
 CombiningInCache<M>::CombiningInCache(
     containers::FlatHashSet<PregelShard> localShards,
     MessageFormat<M> const* format, MessageCombiner<M> const* combiner)
-    : InCache<M>(format), _combiner(combiner), _localShards(localShards) {
+    : InCache<M>(format),
+      _combiner(combiner),
+      _localShards(std::move(localShards)) {
   // one mutex per shard, we will see how this scales
-  for (PregelShard pregelShard : localShards) {
+  for (PregelShard const& pregelShard : localShards) {
     this->_bucketLocker[pregelShard];
     _shardMap[pregelShard];
   }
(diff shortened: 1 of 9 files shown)


hkernbach (approved):

arangod/Pregel/IncomingCache.cpp
```diff
@@ -103,7 +103,7 @@ void InCache<M>::storeMessage(PregelShard shard, std::string_view vertexId,
 // ================== ArrayIncomingCache ==================
 
 template<typename M>
-ArrayInCache<M>::ArrayInCache(std::set<PregelShard> localShards,
+ArrayInCache<M>::ArrayInCache(containers::FlatHashSet<PregelShard> localShards,
                               MessageFormat<M> const* format)
     : InCache<M>(format), _localShards(localShards) {
```
jsteemann, line 108:
```suggestion
    : InCache<M>(format), _localShards(std::move(localShards)) {
```

arangod/Pregel/IncomingCache.cpp
```diff
@@ -103,7 +103,7 @@ void InCache<M>::storeMessage(PregelShard shard, std::string_view vertexId,
...
   // one mutex per shard, we will see how this scales
```
jsteemann, line 109: In the following loop, each `PregelShard` is copied. This is unnecessary. The range-based for loop should use a const reference to PregelShard.

arangod/Pregel/IncomingCache.cpp
jsteemann, line 1: I can only make comments in sections of this file that are modified by this PR, but I see a few recurring patterns through all these files:
- sets and associative containers are passed by copy and are not moved. this may not be a problem now, because the containers may be empty in most cases, but it may change in the future.
- in range-based for loops, the values are accessed by value (copying them) and not by const reference. This causes unnecessary overhead.

arangod/Pregel/IncomingCache.cpp
```diff
@@ -215,12 +215,14 @@ void ArrayInCache<M>::forEach(
...
-    : InCache<M>(format), _combiner(combiner), _localShards(localShards) {
+CombiningInCache<M>::CombiningInCache(
+    containers::FlatHashSet<PregelShard> localShards,
+    MessageFormat<M> const* format, MessageCombiner<M> const* combiner)
+    : InCache<M>(format),
+      _combiner(combiner),
+      _localShards(std::move(localShards)) {
   // one mutex per shard, we will see how this scales
-  for (PregelShard pregelShard : localShards) {
+  for (PregelShard const& pregelShard : localShards) {
```
jsteemann, line 225: Must be changed here, because we moved out of `localShards` above.
```suggestion
  for (PregelShard const& pregelShard : _localShards) {
```

jsteemann (approved):
LGTM

arangod/Pregel/IncomingCache.cpp
markuspf: And really at least the compiler should have told me about this :roll_eyes:

markuspf:
This PR can be merged as  they don't make anything worse; we don't have confirmation yet whether they have a good influence on @neunhoef 's testcase

arangod/Pregel/IncomingCache.cpp
Simran-B: FWIW, GitHub recently added the ability to comment on files as a whole: https://github.blog/changelog/2023-04-11-commenting-on-files-in-a-pull-request-is-now-generally-available/

PR #18976, opened 2023-05-12
Feature/cursor api next batch no retriable

cpjulia:
### Scope & Purpose

In batched query results, when executing requests for  `/_api/cursor/<cursorId>/<batchId>`, removed the restriction that the user would only be able to fetch the next batch using the next batch id in `<batchId>` if the query option `allowRetry` was set to true, but maintained the restriction that the user can only retrieve the latest batch if the query option `allowRetry` is set to true.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19009

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: not it, but related to https://arangodb.atlassian.net/browse/APM-592
- [ ] Design document:

### DIFF
--- a/arangod/RestHandler/RestCursorHandler.cpp
+++ b/arangod/RestHandler/RestCursorHandler.cpp
@@ -43,6 +43,8 @@
 #include <velocypack/Iterator.h>
 #include <velocypack/Value.h>
 
+#include "Logger/LogMacros.h"
+
 using namespace arangodb;
 using namespace arangodb::basics;
 using namespace arangodb::rest;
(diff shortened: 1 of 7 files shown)


arangod/RestHandler/RestCursorHandler.cpp
```diff
@@ -775,7 +775,8 @@ void RestCursorHandler::lookupCursor(std::string_view id,
 
   TRI_ASSERT(_cursor != nullptr);
 
-  if (mustBeRetriable && !_cursor->isRetriable()) {
+  if (batchId.has_value() && _cursor->isCurrentBatchId(batchId.value()) &&
```
jsteemann, line 778: Why only fail if the batchId is identical to the current batch id?
What if a batchId is specified that is totally out of range, but not identical to the current batch id? For example, if the batch id to fetch is 23 and you are specifying a batchId value of 1 or 99?

jsteemann (approved):
LGTM

dothebart (approved):

PR #18978, opened 2023-05-12
[BTS-1400] [3.11] Fix query not found issue with parallel-gather

mpoeter:
### Scope & Purpose

Previously an upstream operation that took longer than 30s would cause the request to abort with an error "query ID xxx not found"

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**
- [x] Backports
  - [x] Backport for devel: #18994


jsteemann (approved):
LGTM

mchacki (approved):
So I agree this PR is good and complete.
But we need an additional one to reactivate the original feature as it was disabled for 3.11.0 right?
_Edit:_
Ok forget the comment, I just see the reference :+1:

PR #18990, opened 2023-05-15
Fix issues in junit xml generator

mpoeter:
### Scope & Purpose

- use correct EE/CE prefix
- only include `time` attribute if we have valid data
- set the `failed` attribute to the number of failed tests instead of to some message

- [x] :hankey: Bugfix


dothebart (approved):
lgtm, Thanks!

markuspf (approved):

PR #18994, opened 2023-05-15
[BTS-1400] Fix query not found issue with parallel-gather

mpoeter:
### Scope & Purpose

Previously an upstream operation that took longer than 30s would cause the request to abort with an error "query ID xxx not found"

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**
- [x] Backports
  - [x] Backport for 3.11: #18978

### DIFF
--- a/arangod/Aql/QueryRegistry.h
+++ b/arangod/Aql/QueryRegistry.h
@@ -86,13 +86,7 @@ class QueryRegistry {
     return static_cast<ExecutionEngine*>(res.get());
   }
 
-  traverser::BaseEngine* openGraphEngine(EngineId eid) {
-    auto res = openEngine(eid, EngineType::Graph, {});
-    if (res.fail()) {
-      return nullptr;
-    }
-    return static_cast<traverser::BaseEngine*>(res.get());
-  }
+  traverser::BaseEngine* openGraphEngine(EngineId eid);
 
   /// @brief close, return a query to the registry, if the query is not found,
   /// an exception is thrown. If the ttl is negative (the default is), the
--- a/arangod/Aql/QueryRegistry.h
+++ b/arangod/Aql/QueryRegistry.h
@@ -26,10 +26,12 @@
 #include "Aql/types.h"
 #include "Basics/Common.h"
 #include "Basics/ErrorCode.h"
+#include "Basics/ResultT.h"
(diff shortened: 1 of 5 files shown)


tests/js/client/shell/shell-aql-regression-cluster.js
```diff
@@ -105,5 +105,28 @@ function ShellAqlRegressionSuite () {
...
+
+    tearDownAll: function () {
+      db._drop(colName);
+    },
+
+    testLongRunningUpstream: function () {
+      // Regression test for https://arangodb.atlassian.net/browse/BTS-1400
+      // Previously an upstream operation that took longer than 30s would cause
+      // the request to abort with an error "query ID xxx not found"
+      db._query("FOR d in @@col  LET x = SLEEP(35) INSERT {} INTO @@col", { "@col": colName });
```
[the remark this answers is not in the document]
jsteemann: Ah ok, I didn't realize that this query produces 2 DB server snippets for each shard. Then it makes sense.
But it was unclear to me when just looking at the test code or the code comment.

arangod/Aql/QueryRegistry.h
```diff
@@ -71,16 +75,18 @@ class QueryRegistry {
   /// Note that an open query will not expire, so users should please
   /// protect against leaks. If an already open query is found, an exception
   /// is thrown.
-  void* openEngine(EngineId eid, EngineType type);
-  ExecutionEngine* openExecutionEngine(EngineId eid) {
-    return static_cast<ExecutionEngine*>(
-        openEngine(eid, EngineType::Execution));
+  ResultT<void*> openEngine(EngineId eid, EngineType type,
```
mchacki, line 78: Not in this PR.
But as far as I see openEngine really wants to be a template
openEngnie<EngineType>
so it guarantees to return the correct type and not void*.
We have the wrappers anyhow.

mchacki (approved):
LGTM, good fix!

arangod/Aql/QueryRegistry.h
mpoeter: Yes, but then the behavior of `openGraphEngine` must be consistent (it currently isn't). I was planning to do this when investigating how to use parallel gather with traversals (which is currently disabled).

PR #18998, opened 2023-05-15
[BTS-1402] Don't apply parallelize gather rule for one-shard DBs

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/1270

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests** - see enterprise companion
- [ ] Backports

#### Related Information

- Jira ticket: https://arangodb.atlassian.net/browse/BTS-1402

### DIFF
--- a/arangod/Aql/OptimizerRules.cpp
+++ b/arangod/Aql/OptimizerRules.cpp
@@ -8348,6 +8348,11 @@ void arangodb::aql::parallelizeGatherRule(Optimizer* opt,
                                           OptimizerRule const& rule) {
   TRI_ASSERT(ServerState::instance()->isCoordinator());
 
+  if (plan->getAst()->query().vocbase().isOneShard()) {
+    opt->addPlan(std::move(plan), rule, false);
+    return;
+  }
+
   bool modified = false;
 
   containers::SmallVector<ExecutionNode*, 8> nodes;


jsteemann (approved):
LGTM

arangod/Aql/OptimizerRules.cpp
```diff
@@ -8348,6 +8348,11 @@ void arangodb::aql::parallelizeGatherRule(Optimizer* opt,
                                           OptimizerRule const& rule) {
   TRI_ASSERT(ServerState::instance()->isCoordinator());
 
+  if (plan->getAst()->query().vocbase().isOneShard()) {
```
mchacki, line 8351: Does it hurt if it is applied? Or is this "just" unnecessary work?

If yes:
If we  are in a sharded database "by accident" have collections with only 1 shard, that are all on the same server we may apply a plan that looks similar to a oneShardDB plan.

mchacki (approved):
In general LGTM.
Im unsure if this is enough or if we need a protection for "accidental" one shards.

mchacki (changes requested):
Wait a second.
Can we have a test similar to the example in the BTS ticket?
Otherwise I cannot be sure that we actually fix something, I am not fully convinced that parallel gather causes the BTS query to "fail"

mpoeter:
As it turned out the referenced BTS issue has a different root cause - thus closing this.

PR #19009, opened 2023-05-16
3.11 backport for cursor api fetch next batch

cpjulia:
### Scope & Purpose

Backport to https://github.com/arangodb/arangodb/pull/18976. 
In batched query results, when executing requests for /_api/cursor/<cursorId>/<batchId>, removed the restriction that the user would only be able to fetch the next batch using the next batch id in <batchId> if the query option allowRetry was set to true, but maintained the restriction that the user can only retrieve the latest batch if the query option allowRetry is set to true.


jsteemann (approved):
LGTM

PR #19011, opened 2023-05-16
Fixup and cleanup checkLogIds

markuspf:
### Scope & Purpose

The `checkLogIds.py` script looked like it was written for a purpose different from what it was used for.

In particular, it was renamed from `generateLogIds.py` in the past, and the checking logIds for uniqueness was hardcoded, so the script itself was not usable for any other purpose.

For the purpose of checking logIds it was also broken: it did not check whether a logId it found was valid leading to code passing CircleCI checks and failing Jenkins checks.

All the unused code is now deleted and replaced by code that explicitly does the following:
* collect all places that have invalid logIds. A logId consists of precisely 5 characters from the set `[a-f0-9]`.
* collect logIds that are used more than once
* print this information in a vaguely useful way.

Note that the check that is run on Jenkins is *different* from the one performed in `checkLogIds`, and has a separate issue in that it does not detect `LOG_PREGEL` macros.

### DIFF
--- a/utils/checkLogIds.py
+++ b/utils/checkLogIds.py
@@ -17,7 +17,7 @@ def check_file(fullpath, project_path, id_database, invalid_logid_database):
             match = g_log_topic_pattern.search(line)
 
             if match:
-                location_string = "{}:{}".format(project_path, cnt)
+                location_string = "{}:{}".format(project_path, cnt + 1)
                 macro = match.group('macro')
                 rest = match.group('rest')
                 param = match.group('param').strip()
--- a/utils/checkLogIds.py
+++ b/utils/checkLogIds.py
@@ -3,23 +3,27 @@
 import sys
 import os
 
-from enum import Enum
-
 import re
-g_log_topic_pattern = re.compile(r'(?P<macro>LOG_(TOPIC(_IF)?|TRX|CTX(_IF)?|QUERY|PREGEL))\((?P<param>[^),]*),(?P<rest>.*)')
+g_log_topic_pattern = re.compile( \
+    r'(?P<macro>LOG_(TOPIC(_IF)?|TRX|CTX(_IF)?|QUERY|PREGEL))\((?P<param>[^),]*),(?P<rest>.*)')
 
 # a LogId is exactly 5 hexadecimal digits
 g_log_id_pattern = re.compile(r'"[0-9a-f]{5}"')
 
-def check_file(fullpath, project_path, id_database, invalid_logid_database):
-    with open(fullpath) as source_file_handle:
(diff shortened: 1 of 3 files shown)


utils/checkLogIds.py
```diff
@@ -80,117 +26,54 @@ def do_operation(fullpath, project_path, target_file_handle, id_database, levels
...
-                if fullpath in exclude:
-                    continue
-                if (filename.endswith((".cpp",".h"))):
-                    status = handle_file(fullpath, project_root, id_database, operation)
+        for filename in files:
+            fullpath = root + os.path.sep + filename
+            if fullpath in exclude:
+                continue
+            if (filename.endswith((".cpp", ".h", ".tpp"))):
+                status = check_file(fullpath, root_directory, id_database, invalid_logid_database)
```
apetenchea, line 74: We now simply ignore the `status`. In fact, it's always going to be `None`, since `check_file` does not return anything.

utils/checkLogIds.py
```diff
@@ -5,71 +5,17 @@
...
-    MODIFY = 2
-
 class Status(Enum):
     OK = 0
-    OK_REPLACED = 1
-    FAIL = 2
-
-def is_good(status):
-    return status in (Status.OK, Status.OK_REPLACED)
+    FAIL = 1
```
apetenchea, line 10: `Status` is no longer relevant in the new code. Not that it was more relevant in the old one anyway.

apetenchea (changes requested):
Thanks for taking the time to give a much needed first-aid on this script!

Although this comes from the old code, the value of `location_string` should be changed: `location_string = "{}:{}".format(fullpath, cnt + 1)`.
`fullpath` indicates the path of the file, and `cnt + 1` is a minor correction to account for the fact that lines start from 1 in the text editor. Otherwise, the script prints the path to the project directory, and always the line above the actual log ID.

The rest of my comments are style-related remarks, feel free to ignore these.

utils/checkLogIds.py
markuspf: Yeah, good spot; I just removed any mention of status.

apetenchea:
Looks good, but my first comment still holds, as the error messages look something like this:
`"063x1" at /home/circleci/project:766`
Where we calculate the location of the problem, at line 20, it should be `location_string = "{}:{}".format(fullpath, cnt + 1).`

markuspf:
Eris, so sorry that I overlooked that; fixed now (also one of the more important fixes!)

dothebart (approved):
please make sure to use pylint on checkLogIds.py

PR #19013, opened 2023-05-16
[BTS-615] Invalid Key Babies error

maierlars:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1273

Previously a invalid key in a batch insert operation aborted the whole batch operation. After this PR the invalid key is reported as an babies error and all valid documents are inserted. This also fixes a discrepancy between cluster and single server.


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #19032, opened 2023-05-16
Fix network response metric. Remove wrong divisor.

maierlars:
### Scope & Purpose
We messed up with calculating the metrics value for `arangodb_network_response_duration`. We did the conversion from nanoseconds to seconds twice, once using `std::chrono` and then a second time using an explicit division by `1e-9`. Other metrics are not affected.


jsteemann (approved):
LGTM

neunhoef (approved):
LGTM. This will need backports.

PR #19038, opened 2023-05-17
Fix junit.xml generation

mpoeter:


dothebart (approved):

dothebart (approved):

dothebart (approved):

dothebart (approved):

dothebart (approved):

dothebart (approved):

dothebart:
tbh I have doubts that the change to the argv handling is good.

js/common/bootstrap/modules/internal.js
```diff
@@ -814,7 +814,7 @@ global.DEFINE_MODULE('internal', (function () {
         if ((thisString.length > 2) &&
           (thisString.slice(0, 2) === '--')) {
           let option = thisString.slice(2, thisString.length);
-          if ((argv.length > i) &&
+          if ((argv.length > i + 1) &&
             (argv[i + 1].slice(0, 1) !== '-')) {
```
mpoeter, line 818: @dothebart If `i == argv.length - 1`, then the old code produced the not so very helpful error `failed to parse the options: Cannot read property 'slice' of undefined`

dothebart: room for improvement fur sure. However:
if you want to support
`--a=b` this must map to `--a b` and implies that most probably Flags like `--a` vs. `--no-a` should be also accepted and produce `--a true` vs. `--a false` internally.

dothebart (approved):

PR #19039, opened 2023-05-17
[3.11] Fix junit.xml generation

mpoeter:
Backport for #19038 and #18990


dothebart (approved):

js/client/modules/@arangodb/testutils/result-processing.js
```diff
@@ -308,24 +314,16 @@ function saveToJunitXML(options, results) {
       }
       let elm = {
         errors: errors,
-        failures: msg,
```
jsteemann, line 311: This changes renders the local variable `msg` useless. Is that intentional?

js/client/modules/@arangodb/testutils/result-processing.js
```diff
@@ -350,18 +348,11 @@ function saveToJunitXML(options, results) {
...
-          }, true);
-        } else {
-          state.xml.elem('testcase', {
-            name: 'all_tests_in_' + state.xmlName,
-            failures: testSuite.failuresFound,
-            time: 0 + testSuite.duration
-          }, true);
-        }
+        const elem = addOptionalDuration({ name: 'all_tests_in_' + state.xmlName }, testSuite);
+        if (testSuite.failed !== 0 || testSuite.failed !== undefined) {
```
jsteemann, line 352: Condition looks strange, should it rather be
```suggestion
        if (testSuite.failed !== 0 && testSuite.failed !== undefined) {
```
Reason: if `testSuite.failed === undefined` it is also `!== 0`.

js/client/modules/@arangodb/testutils/result-processing.js
```diff
@@ -1114,6 +1105,37 @@ function getFailedTestCases(options) {
...
+function getGTestResults(fileName, defaultResults) {
+  let results = defaultResults;
+  if (!fs.exists(fileName)) {
+    defaultResults.failed += 1;
+    print(RED + "No testresult file found at: " + fileName + RESET);
+    return defaultResults;
+  }
+  let gTestResults = JSON.parse(fs.read(fileName));
+  results.failed = gTestResults.failures + gTestResults.errors;
+  results.status = (gTestResults.errors === 0) || (gTestResults.failures === 0);
```
jsteemann, line 1117: I don't understand this. It seems that we have both `errors` and `failures`, and that both can be set independently.
Shouldn't it be
```suggestion
  results.status = (gTestResults.errors === 0) && (gTestResults.failures === 0);
```
then?

js/client/modules/@arangodb/testutils/result-processing.js
```diff
@@ -1114,6 +1105,37 @@ function getFailedTestCases(options) {
...
+      failed: testSuite.failures + testSuite.errors,
+      status: (testSuite.failures + testSuite.errors) === 0,
+      duration: parseFloat(testSuite.time) * 1000 // gtest writes sec, internally we have ms
+    };
+    if (testSuite.failures !== 0) {
+      let message = "";
+      testSuite.testsuite.forEach(function (suite) {
+        if (suite.hasOwnProperty('failures')) {
+          suite.failures.forEach(function (fail) {
+            message += fail.failure;
```
jsteemann, line 1129: If there are multiple failures, they will all be concatenated without any separator. Should there be at least a "\n" as a separator here?

js/client/modules/@arangodb/testutils/result-processing.js
mpoeter: Good catch! msg can be removed now.

js/client/modules/@arangodb/testutils/result-processing.js
mpoeter: Good catch! These are errors that have been in the code base for who knows how long.. 🙄

dothebart (approved):

jsteemann (approved):
LGTM

PR #19047, opened 2023-05-17
track index selectivity estimates memory usage

jsteemann:
### Scope & Purpose

This PR tracks the approximate memory usage by index selectivity estimates that use the RocksDBCuckooIndexEstimator.
Memory is tracked for the base allocation and the queued insert and remove operations. Currently, the memory is tracked by index, but the values can later be added to a global metric.
The PR also releases the memory used by the estimator early when the collection is dropped. Previously, the memory was not released until the server shutdown. this was bad when a lot of collections were created and dropped.
Still missing in this PR is making the total memory usage of estimators available as a metric.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBCuckooIndexEstimator.cpp
+++ b/arangod/RocksDBEngine/RocksDBCuckooIndexEstimator.cpp
@@ -24,6 +24,7 @@
 
 #include "Basics/Exceptions.h"
 #include "Basics/ReadLocker.h"
+#include "Basics/ScopeGuard.h"
 #include "Basics/WriteLocker.h"
 #include "RocksDBEngine/RocksDBCuckooIndexEstimator.h"
 #include "RocksDBEngine/RocksDBFormat.h"
@@ -50,7 +51,8 @@ RocksDBCuckooIndexEstimator<Key>::RocksDBCuckooIndexEstimator(uint64_t size)
       _nrCuckood(0),
       _nrTotal(0),
       _appliedSeq(0),
-      _needToPersist(false) {
+      _needToPersist(false),
+      _memoryUsage(0) {
   // Inflate size so that we have some padding to avoid failure
   size *= 2;
   size = (size >= 1024) ? size : 1024;  // want 256 buckets minimum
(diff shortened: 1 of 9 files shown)


Documentation/Metrics/arangodb_internal_index_estimates_memory.yaml
```diff
@@ -0,0 +1,22 @@
...
+category: Statistics
+complexity: medium
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  Total memory usage of all index selectivity estimates.
+  The metric includes memory used for buffering index write 
+  operations that still have to applied.
```
nerpaula, line 16:
```suggestion
  operations that still have to be applied.
```

arangod/RocksDBEngine/RocksDBCuckooIndexEstimator.cpp
```diff
@@ -707,32 +732,126 @@ void RocksDBCuckooIndexEstimator<Key>::initializeDefault() {
...
-  _slotAllocSize = _size * kSlotSize * kSlotsPerBucket + 64;
+  _allocSize = bytesForSlots + bytesForCounters + 64;
 
-  _slotBase = new char[_slotAllocSize];
+  _allocBase = new char[_allocSize];
 
   _base = reinterpret_cast<char*>(
-      (reinterpret_cast<uintptr_t>(_slotBase) + 63) &
+      (reinterpret_cast<uintptr_t>(_allocBase) + 63) &
       ~((uintptr_t)0x3fu));  // to actually implement the 64-byte alignment,
```
maierlars, line 773: Just FYI there is [std::aligned_alloc](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc). If you want to use `new` you can use
```
auto* p = new(std::align_val_t(64)) char[_allocSize];
```

However, some people on the internet claim that invoking `new` in that way, did not work as expected, although the standard mandates this.

Also `_allocBase` could be stored in a `std::unique_ptr` to make sure the data is always free eventually.

Feel free to completely ignore this comment.

maierlars (commented):
Looks good so far. Afaik you wanted to add tests. Thus not yet approved.

arangod/RocksDBEngine/RocksDBCuckooIndexEstimator.cpp
jsteemann: I moved the allocation to use `std::make_unique<char[]>(allocSize);` instead of `new`.
If there are concerns about `std::aligned_alloc`, then I think won't make that change in this PR.

PR #19050, opened 2023-05-19
Bug fix 3.10/bts 1261

cpjulia:
### Scope & Purpose

Backport for https://github.com/arangodb/arangodb/pull/19060.

For some named graphs in cluster, when creating a debugDump for a traversal query that would use the graph's name, the graph wouldn't be able to be recreated because the info gathered in the process of creating a debugDump was broken. It uses the result form an aql explain to get the graph info and use in the debugDump, but this wouldn't be available because, instead of having the name of the graph as key in the graph object of the explain, there would be an array with edge collection names. It was changed for the case when there's a named graph, but maintained when there's no access to the graph's name.


CHANGELOG
```diff
@@ -1,6 +1,19 @@
...
+  to be recreated because the info gathered in the process of creating a 
+  debugDump was broken. It uses the result form an aql explain to get the graph 
+  info and use in the debugDump, but this wouldn't be available because, 
+  instead of having the name of the graph as key in the graph object of the 
+  explain, there would be an array with edge collection names. It was changed 
+  for the case when there's a named graph, but maintained when there's no 
+  access to the graph's name.
+
+* FE-266: Use the method 'fromGeoJson()' instead of 'geometry.coordinates'
+  for GeoJSON.
```
mchacki, line 15: I do not think this is intentional, please check for hidden merge conflict on changelog.

```suggestion
```

mchacki (approved):
Please check CHANGELOG before we merge.
I think there is a hidden merge conflict on an unrelated feature.
Code looks good to me

jsteemann (approved):
LGTM, but please see @mchacki 's comment.

PR #19060, opened 2023-05-22
[BTS-1261]

cpjulia:
### Scope & Purpose

For some named graphs in cluster, when creating a debugDump for a traversal query that would use the graph's name, the graph wouldn't be able to be recreated because the info gathered in the process of creating a debugDump was broken. It uses the result form an aql explain to get the graph info and use in the debugDump, but this wouldn't be available because, instead of having the name of the graph as key in the graph object of the explain, there would be an array with edge collection names. It was changed for the case when there's a named graph, but maintained when there's no access to the graph's name.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [x] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19062
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19050
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/19061

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1261
- [ ] Design document:

PR #19061, opened 2023-05-22
3.9 backport for BTS-1261

cpjulia:
### Scope & Purpose

Backport for https://github.com/arangodb/arangodb/pull/19060.

For some named graphs in cluster, when creating a debugDump for a traversal query that would use the graph's name, the graph wouldn't be able to be recreated because the info gathered in the process of creating a debugDump was broken. It uses the result form an aql explain to get the graph info and use in the debugDump, but this wouldn't be available because, instead of having the name of the graph as key in the graph object of the explain, there would be an array with edge collection names. It was changed for the case when there's a named graph, but maintained when there's no access to the graph's name.


mchacki (approved):
LGTM

jsteemann (approved):
LGTM

PR #19081, opened 2023-05-23
Removing replication-version parameter

apetenchea:
### Scope & Purpose

The `--database.default-replication-version` should not be available in the 3.11 release.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


jsteemann (approved):
LGTM, but very likely the change will only hit 3.11.1. So the parameter will be available in 3.11.0 and removed in 3.11.1. Fine for me, just mentioning that.

goedderz (approved):

PR #19085, opened 2023-05-24
Add a more verbose message to sporadically failing assert

markuspf:
### Scope & Purpose

For some reason the changed assert sometimes fails with `result.code` being `undefined`; add output of the result to be able to tell slightly better what's going on once this test fails agian.

### DIFF
--- a/tests/js/client/server_parameters/test-return-queue-time-header-on.js
+++ b/tests/js/client/server_parameters/test-return-queue-time-header-on.js
@@ -162,7 +162,7 @@ function testSuite() {
       waitForPending(pendingAtStart, 1, n - 30);
 
       let result = arango.GET("/_api/version", { "x-arango-queue-time-seconds": "0.00001" });
-      assertEqual(412, result.code);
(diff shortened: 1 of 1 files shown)


tests/js/client/server_parameters/test-return-queue-time-header-on.js
```diff
@@ -162,7 +162,7 @@ function testSuite() {
       waitForPending(pendingAtStart, 1, n - 30);
 
       let result = arango.GET("/_api/version", { "x-arango-queue-time-seconds": "0.00001" });
-      assertEqual(412, result.code);
+      assertEqual(412, result.code, JSON.stringify(result));
```
jsteemann, line 165: Nit (does not need to be changed): using `JSON.stringify()` here is fine, but not necessary. If the assertion goes wrong the `result` value would be pretty-printed anyway by `assertEqual`. At least it would not emit `[Object object]` or such.

jsteemann (approved):
LGTM

PR #19089, opened 2023-05-24
[CINFRA-741] Don't call apply entries without a snapshot

goedderz:
### Scope & Purpose

`applyEntries` must only be called after a snapshot is available.

- [X] :hankey: Bugfix

### Checklist

- [X] Tests
  - [X] C++ **Unit tests**


apetenchea (approved):
LGTM

PR #19092, opened 2023-05-24
[APM-674] ClusterInfo memory accounting

maierlars:
### Scope & Purpose
Let's see if we can do memory accounting using PMR allocators.


lib/Containers/FlatHashMap.h
```diff
@@ -60,4 +64,24 @@ template<class K, class V,
...
+template<class K, class V,
+         class Hash = typename absl::flat_hash_map<K, V>::hasher,
+         class Eq = typename absl::flat_hash_map<K, V, Hash>::key_equal>
+using FlatHashMap = arangodb::containers::FlatHashMap<
+    K, V, Hash, Eq, std::pmr::polymorphic_allocator<std::pair<const K, V>>>;
+}
+
 }  // namespace arangodb::containers
+
+// THIS IS JUST TEMPORARY!
```
MBkkt, line 77: https://github.com/abseil/abseil-cpp/blob/master/absl/hash/internal/hash.h#LL532C1-L537C2

arangod/Cluster/ClusterTypes.h
```diff
@@ -42,6 +42,15 @@ typedef std::string ViewID;           // ID of a view
 typedef std::string ShardID;          // ID of a shard
 typedef uint32_t ServerShortID;       // Short ID of a server
 typedef std::string ServerShortName;  // Short name of a server
+namespace pmr {
+typedef std::pmr::string ServerID;         // ID of a server
```
MBkkt, line 46: ServerID is fixed size string (PRMR/COORD/...-uuid), I think better to use array<char, ...>

arangod/Cluster/ClusterInfo.h
```diff
@@ -1187,22 +1198,25 @@ class ClusterInfo final {
   static constexpr double checkAnalyzersPreconditionTimeout = 10.0;
 
   mutable std::mutex _failedServersMutex;
-  containers::FlatHashSet<ServerID> _failedServers;
+  std::pmr::unordered_set<pmr::ServerID> _failedServers;
```
MBkkt, line 1201: Why do you decide to change type?

arangod/Cluster/ClusterTypes.h
```diff
@@ -42,6 +42,15 @@ typedef std::string ViewID;           // ID of a view
 typedef std::string ShardID;          // ID of a shard
 typedef uint32_t ServerShortID;       // Short ID of a server
 typedef std::string ServerShortName;  // Short name of a server
+namespace pmr {
```
MBkkt, line 45: I think it should be ClusterInfo local

lib/Containers/NodeHashMap.h
```diff
@@ -33,4 +37,12 @@ template<class K, class V,
...
 using NodeHashMap = absl::node_hash_map<K, V, Hash, Eq, Allocator>;
 
+namespace pmr {
+template<class K, class V,
+         class Hash = typename absl::node_hash_map<K, V>::hasher,
+         class Eq = typename absl::node_hash_map<K, V, Hash>::key_equal>
+using NodeHashMap =
+    absl::node_hash_map<K, V, Hash, Eq,
+                        std::pmr::polymorphic_allocator<std::pair<const K, V>>>;
+}
```
MBkkt, line 47: Why do we need this using globally?
Same about other changes in the Containers directory

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -546,10 +547,43 @@ DECLARE_HISTOGRAM(arangodb_load_current_runtime, ClusterInfoScale,
...
+namespace {
+struct CountingMemoryResource : std::pmr::memory_resource {
+  explicit CountingMemoryResource(memory_resource* base,
+                                  metrics::Gauge<std::size_t>& metric)
+      : base(base), metric(metric) {}
+
+ private:
+  void* do_allocate(size_t bytes, size_t alignment) override {
+    metric += bytes;
+    return base->allocate(bytes, alignment);
```
gnusi, line 562: Since `memory_resource::allocate` may throw, we have to change the order
```suggestion
    auto* p = base->allocate(bytes, alignment);
    metric += bytes;
    return p;
```

gnusi (changes requested):

lib/Containers/NodeHashMap.h
maierlars: why not? There are other places where we can make use of pmr version of NodeHashMap.

lib/Containers/FlatHashMap.h
maierlars: I have no idea what this should tell me.

arangod/Cluster/ClusterTypes.h
maierlars: out of scope.

arangod/Cluster/ClusterTypes.h
maierlars: why?

arangod/Cluster/ClusterInfo.cpp
maierlars: Thanks! Good point!

lib/Containers/FlatHashMap.h
MBkkt: absl hash allow to makes overloads for your own values

arangod/Cluster/ClusterTypes.h
MBkkt: I don't think so, you already changed the type, why don't change to simpler and smaller one?

arangod/Cluster/ClusterTypes.h
MBkkt: Because it's used in ClusterInfo only

lib/Containers/NodeHashMap.h
MBkkt: For what task?

lib/Containers/FlatHashMap.h
MBkkt: I saw that you write the current container_internal specialization is temporary, so I pointed how it could be done

arangod/Cluster/ClusterTypes.h
maierlars: I think so. ServerIDs are not fixed size. In fact the server name can be dictated from the outside, as it is done for kubernetes.

arangod/Cluster/ClusterInfo.h
maierlars: `pmr::ServerID` is too big for the flat hashset.

arangod/Cluster/ClusterTypes.h
MBkkt: > ServerIDs are not fixed size.

Really? :(

> In fact the server name can be dictated from the outside, as it is done for kubernetes.

Could you explain for me where ServerID set to not restricted size string?
I found only two places and both looks like role + uuid

```cpp
  std::string id;
  bool hadPersistedId = hasPersistedId();
  if (!hadPersistedId) {
    id = generatePersistedId(role);

    LOG_TOPIC("0d924", INFO, Logger::CLUSTER)
        << "Fresh start. Persisting new UUID " << id;
  } else {
    id = getPersistedId();
    LOG_TOPIC("db3ce", DEBUG, Logger::CLUSTER)
        << "Restarting with persisted UUID " << id;
  }
  setId(id);
```

```cpp
    // If we have persisted id, we use that. Otherwise we check, if we were
    // given a disaster recovery id that wins then before a new one is generated
    // and that choice persisted.
    std::string uuid;
    if (ServerState::instance()->hasPersistedId()) {
      uuid = ServerState::instance()->getPersistedId();
    } else {
      std::string recoveryId = _agent->config().recoveryId();
      if (recoveryId.empty()) {
        uuid = ServerState::instance()->generatePersistedId(
            ServerState::ROLE_AGENT);
      } else {
        uuid = recoveryId;
        ServerState::instance()->writePersistedId(recoveryId);
      }
    }
    _agent->id(uuid);
    ServerState::instance()->setId(uuid);
```


Maybe this option?
```cpp
  options->addOption("--agency.disaster-recovery-id",
                     "Specify the ID for this agent. WARNING: This is a "
                     "dangerous option, for disaster recover only!"
```

maierlars: > Or we just put something not uuid like to the uuid file?
Yes.

arangod/Cluster/ClusterInfo.h
maierlars: As I did?

MBkkt: I saw and resolved

arangod/Cluster/ClusterTypes.h
MBkkt: Ok

lib/Containers/FlatHashMap.h
maierlars: but how to make it transparent?

MBkkt: Define your own Hash/Eq. So you don't want to write it now and just make internal overload, ok

Documentation/Metrics/arangodb_internal_cluster_info_memory.yaml
```diff
@@ -0,0 +1,18 @@
+name: arangodb_internal_cluster_info_memory
+introducedIn: "3.12"
+help: |
+  Amount of memory spent in ClusterInfo.
+unit: byte
+type: gauge
+# all other categories are to specific
```
jsteemann, line 7:
```suggestion
# all other categories are too specific
```

Documentation/Metrics/arangodb_internal_cluster_info_memory.yaml
```diff
@@ -0,0 +1,18 @@
+name: arangodb_internal_cluster_info_memory
```
jsteemann, line 1: I suggest to rename this metric to "_memory_usage", as it then would be in line with several other metrics.
```suggestion
name: arangodb_internal_cluster_info_memory_usage
```

Documentation/Metrics/arangodb_internal_cluster_info_memory.yaml
```diff
@@ -0,0 +1,18 @@
...
+  Amount of memory spent in ClusterInfo.
+unit: byte
+type: gauge
+# all other categories are to specific
+category: Statistics
+complexity: high
+exposedBy:
+  - coordinator
+  - dbserver
+  - agent
```
jsteemann, line 13: IMHO there is no ClusterInfo in the agency.
```suggestion
```

arangod/Cluster/ClusterTypes.h
```diff
@@ -41,6 +42,17 @@ typedef std::string ViewID;           // ID of a view
...
+using ManagedString = std::basic_string<char, std::char_traits<char>,
+                                        ResourceUsageAllocator<char>>;
+typedef ManagedString ServerID;         // ID of a server
+typedef ManagedString DatabaseID;       // ID/name of a database
+typedef ManagedString CollectionID;     // ID of a collection
+typedef ManagedString ViewID;           // ID of a view
+typedef ManagedString ShardID;          // ID of a shard
+typedef uint32_t ServerShortID;         // Short ID of a server
```
jsteemann, line 53: Do we need this extra typedef? It seems it does not need a special allocator for anything.
```suggestion
typedef uint32_t ServerShortID;         // Short ID of a server
```

arangod/Cluster/ClusterInfo.h
```diff
@@ -1200,7 +1252,13 @@ class ClusterInfo final {
...
 
   mutable std::mutex _failedServersMutex;
-  containers::FlatHashSet<ServerID> _failedServers;
+  template<typename K>
+  using NodeHashSet =
+      containers::NodeHashSet<K, typename containers::NodeHashSet<K>::hasher,
+                              typename containers::NodeHashSet<K>::key_equal,
+                              ResourceUsageAllocator<K>>;
+
+  NodeHashSet<pmr::ServerID> _failedServers;
```
jsteemann, line 1261: Previously `_failedServers` was a FlatHashSet. Now it is a NodeHashSet. Which is fine, but is there a specific reason for this change?

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -547,20 +550,55 @@ DECLARE_HISTOGRAM(arangodb_load_current_runtime, ClusterInfoScale,
 DECLARE_HISTOGRAM(arangodb_load_plan_runtime, ClusterInfoScale,
                   "Plan loading runtimes [ms]");
 
+DECLARE_GAUGE(arangodb_internal_cluster_info_memory, std::size_t,
```
jsteemann, line 553: I suggest to rename the metric so it is in line with many other memory usage metrics.
```suggestion
DECLARE_GAUGE(arangodb_internal_cluster_info_memory_usage, std::size_t,
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1511,16 +1555,28 @@ void ClusterInfo::loadPlan() {
...
-  decltype(_collectionGroups) newCollectionGroups;
+  decltype(_replicatedLogs) newReplicatedLogs{_resourceMonitor};
+  decltype(_collectionGroups) newCollectionGroups{_resourceMonitor};
+
+  static_assert(std::uses_allocator_v<
+                ClusterInfo::NewStuffByDatabase,
+                ResourceUsageAllocator<ClusterInfo::NewStuffByDatabase>>);
+
+  std::make_obj_using_allocator<NewStuffByDatabase>(
+      ResourceUsageAllocator<int>(_resourceMonitor));
```
jsteemann, line 1566: My understanding of this statement is that it creases a temporary `NewStuffByDatabase` object on the stack, which, because it is anonymous, is not referred to later.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6600,53 +6670,87 @@ ClusterInfo::getCurrent(uint64_t& index,
...
 }
 
 void ClusterInfo::setFailedServers(
     containers::FlatHashSet<ServerID> failedServers) {
   std::lock_guard lock{_failedServersMutex};
-  _failedServers = std::move(failedServers);
+  _failedServers.clear();
+  for (auto const& server : failedServers) {
+    _failedServers.emplace(server);
+  }
```
jsteemann, line 6687: nit: this method gets its argument by value, so we could now move the values into the container.
```suggestion
  for (auto& server : failedServers) {
    _failedServers.emplace(std::move(server));
  }
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6600,53 +6670,87 @@ ClusterInfo::getCurrent(uint64_t& index,
...
 }
 
 #ifdef ARANGODB_USE_GOOGLE_TESTS
 void ClusterInfo::setServers(
     containers::FlatHashMap<ServerID, std::string> servers) {
   WRITE_LOCKER(writeLocker, _serversProt.lock);
-  _servers = std::move(servers);
+  _servers.clear();
+  for (auto const& [k, v] : servers) {
+    _servers.emplace(k, v);
```
jsteemann, line 6696: same

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6600,53 +6670,87 @@ ClusterInfo::getCurrent(uint64_t& index,
...
+  }
 }
 
 void ClusterInfo::setServerAliases(
     containers::FlatHashMap<ServerID, std::string> aliases) {
   WRITE_LOCKER(writeLocker, _serversProt.lock);
-  _serverAliases = std::move(aliases);
+  _serverAliases.clear();
+  for (auto const& [k, v] : aliases) {
+    _serverAliases.emplace(k, v);
```
jsteemann, line 6705: same (also for the following methods).

arangod/Cluster/ClusterInfo.cpp
maierlars: left over, I will remove this.

lib/Basics/debugging.h
```diff
@@ -146,6 +146,11 @@ struct remove_cvref {
 template<class T>
 using remove_cvref_t = typename remove_cvref<T>::type;
 
+template<typename>
+struct is_basic_string : std::false_type {};
+template<typename C, typename T, typename A>
+struct is_basic_string<std::basic_string<C, T, A>> : std::true_type {};
```
MBkkt, line 152: Maybe we should check only char string?

lib/Basics/debugging.h
```diff
@@ -156,7 +161,7 @@ struct is_container
               !std::is_same_v<unsigned char*, typename std::decay<T>::type> &&
               !std::is_same_v<unsigned char const*,
                               typename std::decay<T>::type> &&
-              !std::is_same_v<T, std::string> &&
+              !is_basic_string<T>::value &&
               !std::is_same_v<T, std::string_view> &&
               !std::is_same_v<T, const std::string>,
```
MBkkt, line 166: Could be removed if we remove const from T above

arangod/Cluster/ClusterInfo.h
MBkkt: It's maybe because increased size of entry, in arangodb we have static asserts to prevent such usage of absl hash tables.
40 bytes now.

I think speed is pretty same in such case according to benchmark, so we probably could increase the limit, or just use node hash set here as Lars wrote

arangod/Cluster/ClusterInfo.h
```diff
@@ -150,19 +154,61 @@ class ClusterInfo final {
     uint64_t hash;
     std::shared_ptr<LogicalCollection> collection;
   };
+
+  struct Hasher {
```
MBkkt, line 158: Please keep using absl::Hash, it's important for absl containers

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -6600,53 +6670,87 @@ ClusterInfo::getCurrent(uint64_t& index,
...
+  }
+  return result;
 }
 
 void ClusterInfo::setFailedServers(
     containers::FlatHashSet<ServerID> failedServers) {
   std::lock_guard lock{_failedServersMutex};
-  _failedServers = std::move(failedServers);
+  _failedServers.clear();
+  for (auto const& server : failedServers) {
```
MBkkt, line 6685: Could you call reserve before inserting please?

arangod/Cluster/ClusterInfo.h
MBkkt: Personally me prefer increase limit to 40, 32 was written because string is 32.
But for maps we use bigger pair sizeof, so I think it's ok to just change this to 40

arangod/Cluster/ClusterInfo.cpp
maierlars: no we can't, because the string inside the container uses a different allocator.

arangod/Cluster/ClusterInfo.cpp
maierlars: still not possible.

arangod/Cluster/ClusterInfo.cpp
maierlars: only used when `ARANGODB_USE_GOOGLE_TESTS` is defined. So probably not even relevant for production.

arangod/Cluster/ClusterInfo.h
maierlars: why is it important?

MBkkt: Because absl hashtables performance more dependant of quality of hash functions.

And it already was and it's not difficult to write `absl::HashOf(value)` instead of `std::hash<T>{}(value)`, so why not?

lib/Basics/debugging.h
jsteemann: Discussed with @MBkkt: can be ignored for now.

arangod/Cluster/ClusterInfo.h
jsteemann: Discussed with @MBkkt : can be ignored for now.

lib/Basics/debugging.h
jsteemann: Discussed with @MBkkt : can be ignored for now.

jsteemann (approved):
LGTM

MBkkt (commented):
LGTM.
But maybe add new js test for new metric?

MBkkt (approved):

PR #19096, opened 2023-05-24
remove outdated defines from the codebase

jsteemann:
### Scope & Purpose

This PR removes the following defines from the codebase:
* `TRI_PADDING_32`: was set to 1 if we were on a 32 bit platform. As we don't support 32 bit platforms for many years, it is time now to retire this define. It was only used in a single code component, where it would have added padding on 32 bit systems but none on 64 bit.
* `TRI_UNALIGNED_ACCESS`: was set to 1 on x86_64/amd64, but not on aarch64. when set to 1, the fasthash code potentially did unaligned data reads of its input buffer. This works on x86_64/amd64 but is not portable. When set to 0, the fasthash code would do proper 64 bit-aligned reads, which is portable but may have a slightly negative performance impact. As the define is now removed, we always go with aligned reads, which is portable.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cache/PlainBucket.h
+++ b/arangod/Cache/PlainBucket.h
@@ -54,11 +54,6 @@ struct PlainBucket {
   std::uint32_t _cachedHashes[slotsData];
   CachedValue* _cachedData[slotsData];
 
-  // padding, if necessary?
-#ifdef TRI_PADDING_32
-  uint32_t _padding[slotsData];
-#endif
-
   //////////////////////////////////////////////////////////////////////////////
   /// @brief Initialize an empty bucket.
   //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


MBkkt (changes requested):
unaligned read can be faster especially on new x86 processors
Could you check before this change?

jsteemann:
disliked by @MBkkt

MBkkt:
As I wrote I don't have something against remove these global defines.
But if we want to change fashash function code, we need benchmark or alternative solution is just using memcpy instead of direct load

PR #19097, opened 2023-05-25
Simpler Scheduler?

maierlars:
### Scope & Purpose
The supervised scheduler starts and stops thread using heuristics. We are not sure if this complexity is actually worth this. Thus we added a simple fixed-sized thread pool alternative.

### DIFF
--- a/arangod/Scheduler/ThreadPoolScheduler.h
+++ b/arangod/Scheduler/ThreadPoolScheduler.h
@@ -38,20 +38,19 @@
 namespace arangodb {
 class NetworkFeature;
 class SharedPRNGFeature;
-class SupervisedSchedulerWorkerThread;
-class SupervisedSchedulerManagerThread;
+class ThreadPoolSchedulerWorkerThread;
 
-class SupervisedScheduler final : public Scheduler {
+class ThreadPoolScheduler final : public Scheduler {
  public:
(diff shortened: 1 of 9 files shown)


arangod/Scheduler/ThreadPoolScheduler.h
```diff
@@ -0,0 +1,225 @@
...
+  std::atomic<bool> _stopping;
+  std::atomic<bool> _acceptingNewJobs;
+
+  // Since the lockfree queue can only handle PODs, one has to wrap lambdas
+  // in a container class and store pointers. -- Maybe there is a better way?
+  struct {
+    /// @brief the number of items that have been enqueued via tryBoundedQueue
+    /// Items that are added via an unbounded queue operation are not counted!
+    std::atomic<uint64_t> numCountedItems{0};
+    boost::lockfree::queue<WorkItemBase*> queue;
```
MBkkt, line 168: boost lockfree queue performance is bad compared to other lockfree queue or even intrusive single linked list with mutex. also it contains data races
 Could you change this too?

maierlars: no. out of scope.

maierlars:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/30155/ check for devel

PR #19106, opened 2023-05-26
APM-663: revision tree memory tracking

jsteemann:
### Scope & Purpose

Implements https://arangodb.atlassian.net/browse/APM-663

Tracks memory used for buffered updates to all revision trees in a new metric `arangodb_revision_tree_buffered_memory_usage`.

This PR also tries to optimize a problem with constant hibernation and resurrection patterns for the same tree. This was been found in reality by observing the hibernation and resurrection patterns of collections that is seldomly written into. For example, the `_statistics` system collection is written into only every ~10 seconds. Previously hibernation could happen randomly after a write to the collection, and then the next write would resurrect it. 
To break out of recurring unnecessary hibernation/resurrection cycles for the same tree, we now track the time of the last hibernation attempt for each revision tree, and do not attempt hibernation if the last attempt was less than a minute ago.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/APM-663
- [ ] Design document:


nerpaula (approved):
LGTM

markuspf (changes requested):
Maybe the description should also mention that you introduce the `lastCompressAttempt` (and what for).

jsteemann:
> Maybe the description should also mention that you introduce the `lastCompressAttempt` (and what for).

True. I added that later and forgot to mention it. Will add it.

Documentation/Metrics/arangodb_revision_tree_buffered_memory_usage.yaml
```diff
@@ -0,0 +1,24 @@
+name: arangodb_revision_tree_buffered_memory_usage
+introducedIn: "3.12.0"
+help: |
+  Total memory usage of buffered updates for all revision trees.
+unit: bytes
+type: gauge
+category: Replication
+complexity: simple
```
maierlars, line 8: Is this really simple, i.e. it should be visible on the most simple dashboard? I would go for `advanced`.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -1261,17 +1267,24 @@ void RocksDBEngine::trackRevisionTreeResurrection() noexcept {
 
 void RocksDBEngine::trackRevisionTreeMemoryIncrease(
     std::uint64_t value) noexcept {
-  if (value != 0) {
```
markuspf, line 1264: Do you by chance remember why this was here?

arangod/RocksDBEngine/RocksDBMetaCollection.h
```diff
@@ -165,6 +170,15 @@ class RocksDBMetaCollection : public PhysicalCollection {
           std::unique_ptr<containers::RevisionTree>,
           std::unique_lock<std::mutex>& lock)> const& callback);
 
+  // used for calculating memory usage in _revisionsBufferedMemoryUsage
+  // approximate memory usage for top-level items
+  static constexpr uint64_t bufferedEntrySize() { return 40; }
```
markuspf, line 175: How did this `40` get chosen?

tests/js/client/shell/shell-revision-tree-tracking-noncluster.js
```diff
@@ -0,0 +1,116 @@
...
+      const n = 10000;
+
+      let c = db._create(cn);
+      for (let i = 0; i < n; ++i) {
+        c.insert({value: i});
+      }
+
+      // must have more memory allocated for the documents
+      let metric = getMetric();
+      assertTrue(metric >= initial + n * (8 + 32), { metric, initial });
```
markuspf, line 106: Is this the same 40 as above?

markuspf (approved):
From what I can tell this looks good to me;

I am queasy about how to ascertain that all places that allocate/deallocate memory are accounted for: since I don't know the code in depth, the only way for me to tell is to learn this whole subsystem. That'd probably take too much time. Maybe we could have a quick zoom so I can get a high level view of what's happening?

arangod/RocksDBEngine/RocksDBEngine.cpp
jsteemann: It was there only to save an atomic write in case it wouldn't be necessary.

PR #19107, opened 2023-05-26
[BTS-1407] Cancel Requests in retry

maierlars:
### Scope & Purpose
Previously if a request decided to be retried, it would post that on the scheduler but keep the work item for itself. During shutdown, the work item wouldn't be cancelled. Now the NetworkFeature keeps track of those requests.

https://arangodb.atlassian.net/browse/BTS-1407


jsteemann (approved):
LGTM

PR #19113, opened 2023-05-26
Remove ttl and observe support in Agency

maierlars:
### Scope & Purpose
Remove unused "features" from the agency and the node store.

This PR also contains some other fixes:
- There was a `using namespace arangodb::velocypack` in a header file. This using was already used by a few cpp files. Thus, when I removed it, I had to add it into the individual compilation units.
- `Node.h` was included transitivily in a lot of files. I reduced that by introducing a pointer indirection. (This indirection will be used in a follow up PR)


MBkkt (approved):

neunhoef (commented):
LGTM. I think TTL is unused as we speak. But I fear `arangosync` is still using "observe" to find out about changes in the Plan (shard movements, failovers, new collections, dropped collections, etc.). We might need to change things there before we can go ahead with this.

maierlars:
This has all been discussed with @informalict already. They are going to remove it.

informalict (commented):
What would happen if TTL was removed:
One ArangoSync master registers leadership with TTL 30 seconds, so if it is gone then no other ArangoSync master can take over leadership.

What would happen when “observe” was removed:
On the source cluster tracking inventory changes will not be triggered, so eventually as a last resort it will be applied after 5 minutes. For the time being it happens only when callback is not received from agency.
Observing /arangodb/arangosync/masters will not work properly. ArangoSync masters will not know about each other for 5 minutes when it changes.
/arango/Readonly will not be tracked anymore, so checking in-sync process will be delayed. Eventually it should be checked every 30 seconds.

neunhoef (approved):
LGTM

PR #19117, opened 2023-05-29
Fixing a shutdown issue.

maierlars:
### Scope & Purpose
Previously during shutdown there was the following race: 
1. the resume lambda was posted on the scheduler and executed.
2. The lambda was able to obtain the `self` pointer.
3. The shutdown code was executed and called `_retryRequests.clear()`
4. The lambda attempted to erase a missing key. Then resumed the already cancelled request.


jsteemann (approved):
LGTM

PR #19119, opened 2023-05-30
FE-287: fix number validation for replicationFactor and writeConcern

palashkaria:
### Scope & Purpose

Number ending in 0 were not being accepted for replicateFactor & writeConcern fields
Affects:
1. graph addition,
2. collection addition -> advanced (for clusters)

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] Manually tested
- [x] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19120
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19121
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/19122

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-287


js/apps/system/_admin/aardvark/APP/frontend/js/views/collectionsView.js
```diff
@@ -650,7 +650,7 @@
                 false,
                 [
                   {
-                    rule: Joi.string().allow('').optional().regex(/^[1-9]*$/),
+                    rule: Joi.string().allow('').optional().regex(/^[1-9][0-9]*$/),
```
jsteemann, line 653: Nit:
Just mentioning here that there is also a maximum value for writeConcern and for replicationFactor.
The current hard-coded maximum value is 10, although the effective maximum value for replicationFactor is the minimum of the number of available DB servers and 10.
The effective maximum value for writeConcern is the effective maximum value of replicationFactor.
The current validation will allow entering much higher values, which will then be rejected by the server.

I don't think it is a good idea to hard-code all these validations in the frontend anyway, as any change in the server will leave the frontend validations untouched and silently get out of sync.
I don't suggest doing any change here, but probably we should not put too much effort into the frontend validations now until we have a better approach of keeping the frontend and server side validation in sync.

ajurdak (approved):
LGTM.

PR #19124, opened 2023-05-30
[CINFRA-773] Fix sequence number tracking in syncWal()

goedderz:
### Scope & Purpose

`RocksDBSyncThread::_lastSequenceNumber` is updated in `RocksDBSyncThread::syncWal()` too early (before the sync call), and it's updated even if the sync failed.

This is currently of no consequence (it is unused), but it will be used for the `waitForSync=false` implementation of Replication 2.

- [X] :hankey: Bugfix

### DIFF
--- a/arangod/RocksDBEngine/RocksDBSyncThread.cpp
+++ b/arangod/RocksDBEngine/RocksDBSyncThread.cpp
@@ -55,7 +55,13 @@ Result RocksDBSyncThread::syncWal() {
 
   // set time of last syncing under the lock
   auto const now = std::chrono::steady_clock::now();
-  {
+
+  // actual syncing is done without holding the lock
+  auto const result = sync(db);
+
+  if (result.ok()) {
+    auto lastSequenceNumber = db->GetLatestSequenceNumber();
+
     std::lock_guard guard{_condition.mutex};
 
     if (now > _lastSyncTime) {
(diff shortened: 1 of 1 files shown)


arangod/RocksDBEngine/RocksDBSyncThread.cpp
```diff
@@ -55,24 +55,27 @@ Result RocksDBSyncThread::syncWal() {
...
   auto const now = std::chrono::steady_clock::now();
-  {
+
+  // actual syncing is done without holding the lock
+  auto const result = sync(db);
+
+  if (result.ok()) {
+    auto lastSequenceNumber = db->GetLatestSequenceNumber();
+
     std::lock_guard guard{_condition.mutex};
```
MBkkt, line 65: if `GetLatestSequenceNumber` will be after `lock` condition `if (lastSequenceNumber > _lastSequenceNumber) {` is probably should be assert

goedderz: Yes, though I don't quite understand what you're going for. But now that you point here I notice that I still called `GetLatestSequenceNumber` after the sync call 🤦‍♂️, which is precisely what I _didn't_ want.

goedderz: To clarify: `_lastSequenceNumber` should always be a sequence number that has already been synced to disk (so a lower bound on the largest sequence number that has been synced).

And regarding the assert: I could add e.g.
```TRI_ASSERT(db->GetLatestSequenceNumber() >= _lastSequenceNumber);```
under the lock, but I don't see much use for that. Or did you mean something else?

MBkkt: I don't think it's required, I just thought assert is better because we check invariant.
I don't insist

MBkkt: Hmm, I see last commit, understand now

jsteemann (approved):
LGTM

PR #19132, opened 2023-05-31
micro-optimizations for cache performance

jsteemann:
### Scope & Purpose

Micro-optimizations for cache performance
* track number of used slots, so we don't need to unnecessarily iterate over empty slots on find/evict etc.
* optimization for eviction: previously, eviction worked by calling `evictionCandidate()` and then `evict()`. this did a scan over the slots to find the eviction candidate, and then another scan over the slots to find the value to evict. two scans are unnecessary here, and they are now combined into a single one.
* move from `arangodb::Result` to `::ErrorCode` for cache insert/remove/ banish/find operations. this avoids dynamic memory allocations in case the insert operation fails because all slots are full or the cache has hit its size limit.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cache/BucketState.h
```diff
@@ -47,25 +46,20 @@ struct BucketState {
...
   /// @brief Flags which can be queried or toggled to reflect state.
   ///
-  /// Each flag must have exactly one bit set, fit in uint32_t. The 'locked'
+  /// Each flag must have exactly one bit set, fit in uint16_t. The 'locked'
   /// flag is special and should remain the least-significant bit. When other
   /// flags are added,they should be kept in alphabetical order for readability,
   /// and all flag values should be adjusted to keep bit-significance in
   /// ascending order.
   //////////////////////////////////////////////////////////////////////////////
-  enum class Flag : std::uint32_t {
```
jsteemann, line 56: note: the removed enum values were not in use in the existing code, so they are removed in this PR.
the change from uint32_t to uint16_t was necessary so we are able to additionally squeeze a uint16_t `_slotsUsed` counter into the 128 bytes that a PlainBucket or TransactionalBucket occupies, without making the struct any bigger.

arangod/Cache/BucketState.cpp
```diff
@@ -45,18 +45,17 @@ BucketState& BucketState::operator=(BucketState const& other) noexcept {
...
   while (attempt < maxTries) {
     // expect unlocked, but need to preserve migrating status
-    std::uint32_t current = _state.load(std::memory_order_relaxed);
-    std::uint32_t expected =
-        current & (~static_cast<std::uint32_t>(Flag::locked));
+    FlagType current = _state.load(std::memory_order_relaxed);
+    FlagType expected = current & (~static_cast<FlagType>(Flag::locked));
     if (current == expected) {
-      uint32_t desired = expected | static_cast<std::uint32_t>(Flag::locked);
+      uint32_t desired = expected | static_cast<FlagType>(Flag::locked);
```
MBkkt, line 58: uint32_t is better because everything smaller implicitly casts to int

jsteemann: What does that mean exactly? 
The stored `_state` value is supposed to fit into a uint16_t, because if we use a uint32_t for it, the struct would become to big.
So not exactly sure what you are suggesting exactly. Thanks for clarifying it.

jsteemann, line 58:
```suggestion
      FlagType desired = expected | static_cast<FlagType>(Flag::locked);
```

arangod/Cache/PlainBucket.cpp
```diff
@@ -95,57 +86,96 @@ CachedValue* PlainBucket::find(std::uint32_t hash, void const* key,
...
+    std::uint64_t size = _cachedData[slot]->size();
+    // evict value. we checked that it is freeable
+    delete _cachedData[slot];
+    _cachedHashes[slot] = _cachedHashes[_slotsUsed - 1];
+    _cachedData[slot] = _cachedData[_slotsUsed - 1];
+    _cachedHashes[_slotsUsed - 1] = 0;
+    _cachedData[_slotsUsed - 1] = nullptr;
+    TRI_ASSERT(_slotsUsed > 0);
+    --_slotsUsed;
+    checkInvariants();
```
mpoeter, line 146: This is the same code as in remove. Can we split this into a separate helper function?

arangod/Cache/PlainBucket.cpp
```diff
@@ -155,36 +185,50 @@ void PlainBucket::clear() noexcept {
...
     _cachedData[i] = nullptr;
   }
+  _slotsUsed = 0;
+  checkInvariants();
 
   _state.unlock();
 }
 
-void PlainBucket::moveSlot(std::size_t slot, bool moveToFront) noexcept {
+void PlainBucket::moveSlot(std::size_t slot) noexcept {
```
mpoeter, line 198: Since this now always moves to front, can we actually call the function `moveSlotToFront`?

arangod/Cache/TransactionalBucket.cpp
```diff
@@ -193,60 +183,57 @@ bool TransactionalBucket::isBanished(std::uint32_t hash) const noexcept {
...
+    std::uint64_t size = _cachedData[slot]->size();
+    // evict value. we checked that it is freeable
+    delete _cachedData[slot];
+    _cachedHashes[slot] = _cachedHashes[_slotsUsed - 1];
+    _cachedData[slot] = _cachedData[_slotsUsed - 1];
+    _cachedHashes[_slotsUsed - 1] = 0;
+    _cachedData[_slotsUsed - 1] = nullptr;
+    TRI_ASSERT(_slotsUsed > 0);
+    --_slotsUsed;
+    checkInvariants();
```
mpoeter, line 204: Same as above - please split into separate function.

arangod/Cache/TransactionalBucket.cpp
```diff
@@ -193,60 +183,57 @@ bool TransactionalBucket::isBanished(std::uint32_t hash) const noexcept {
...
-      _cachedHashes[slot] = 0;
-      _cachedData[slot] = nullptr;
-      moveSlot(slot, optimizeForInsertion);
+      _cachedHashes[i] = _cachedHashes[_slotsUsed - 1];
+      _cachedData[i] = _cachedData[_slotsUsed - 1];
+      _cachedHashes[_slotsUsed - 1] = 0;
+      _cachedData[_slotsUsed - 1] = nullptr;
+      TRI_ASSERT(_slotsUsed > 0);
+      --_slotsUsed;
+      checkInvariants();
```
mpoeter, line 236: Same duplication as above.

arangod/Cache/TransactionalBucket.cpp
```diff
@@ -274,28 +265,21 @@ void TransactionalBucket::updateBanishTerm(std::uint64_t term) noexcept {
...
-    memset(_banishHashes, 0, (slotsBanish * sizeof(std::uint32_t)));
+    for (std::size_t i = 0; i < kSlotsBanish; ++i) {
+      _banishHashes[i] = 0;
+    }
   }
 }
 
-void TransactionalBucket::moveSlot(std::size_t slot,
-                                   bool moveToFront) noexcept {
+void TransactionalBucket::moveSlot(std::size_t slot) noexcept {
```
mpoeter, line 274: Same as above - rename to `moveSlotToFront`.

MBkkt (approved):

arangod/Cache/TransactionalBucket.cpp
```diff
@@ -274,31 +254,33 @@ void TransactionalBucket::updateBanishTerm(std::uint64_t term) noexcept {
       _state.toggleFlag(BucketState::Flag::banished);
     }
 
-    memset(_banishHashes, 0, (slotsBanish * sizeof(std::uint32_t)));
+    for (std::size_t slot = 0; slot < kSlotsBanish; ++slot) {
+      _banishHashes[slot] = 0;
+    }
```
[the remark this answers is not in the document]
jsteemann: Seems to compile to identical code with g++-12 for x86_64 and -O3:
https://godbolt.org/z/PaeK1Mja1

arangod/Cache/BucketState.h
```diff
@@ -47,25 +46,20 @@ struct BucketState {
...
-    banished = 0x00000002,
-    disabled = 0x00000004,
-    evictions = 0x00000008,
-    migrated = 0x00000010,
-    migrating = 0x00000020,
-    rebalancing = 0x00000040,
-    resizing = 0x00000080,
-    shutdown = 0x00000100,
-    shuttingDown = 0x00000200,
+  enum class Flag : std::uint16_t {
```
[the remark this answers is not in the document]
jsteemann: Could be changed to uint8_t as well, but this would not help use save any storage space later. So we can as well go with a uint16_t here.

arangod/Cache/TransactionalBucket.cpp
```diff
@@ -255,14 +231,18 @@ void TransactionalBucket::evict(CachedValue* value,
...
+    _banishHashes[slot] = 0;
   }
   _banishTerm = 0;
-  for (std::size_t i = 0; i < slotsData; ++i) {
-    _cachedHashes[i] = 0;
-    _cachedData[i] = nullptr;
+  for (std::size_t slot = 0; slot < kSlotsData; ++slot) {
+    _cachedHashes[slot] = 0;
+    _cachedData[slot] = nullptr;
   }
```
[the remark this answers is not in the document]
jsteemann: I intentionally did not want to use memset here.
It is true that memsetting the entire struct with 0 would also work and would probably be most efficient, but for readability I would prefer using simple loops and assignments here.
It turns out that clearing two arrays from the same loops produces less optimal code than writing two separate loops one after the other, so I changed that.

arangod/Cache/BucketState.h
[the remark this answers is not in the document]
MBkkt: Ok

arangod/Cache/TransactionalBucket.cpp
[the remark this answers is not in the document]
MBkkt: Ok

arangod/Cache/TransactionalBucket.cpp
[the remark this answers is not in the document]
MBkkt: Good

PR #19133, opened 2023-05-31
Bug fix/bts 846

cpjulia:
### Scope & Purpose

Relaxed setup cost for calculation of default sort cost. There was a hardcoded value of 100 being added to the sort cost that would be dominant when there was a small amount of documents in the collection. Even if an index would be the best candidate, if it doesn't cover a sort node present in a query, it might end up not being chosen because the default sort cost would be minimum 100 because of the former hardcoded value.
Example:
```
c = db._create("c")
c.insert({field:1})
c.ensureIndex({type:"persistent",unique:true, fields:["field"]})
db._explain(`FOR d IN c FILTER d.field == 1 SORT d._key RETURN d`)
```
Here, we want the persistent index to be chosen instead of the primary index.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-846
- [ ] Design document:


tests/js/server/aql/aql-persistent-index.js
```diff
@@ -1552,8 +1596,137 @@ function PersistentIndexOverlappingSuite () {
...
+    testNonUniqueSimple: function () {
+      testandValidate(query1, false, false, false, false, false, 1);
+    },
+
+    testNonUniqueSimple2: function () {
+      testandValidate(query2, false, false, false, false, false, 1);
+    },
+
+    testUniqueSimpleManyDocs: function () {
+      if (isCluster) {
```
jsteemann, line 1632: the `isCluster` checks here and following are not necessary, because the entire suite is only run if `!isCluster`.

tests/js/server/aql/aql-persistent-index.js
```diff
@@ -30,20 +30,64 @@ const internal = require("internal");
...
+const idxName = "testIdx";
+const query1 = `FOR d IN ${cn} FILTER d.value == 1 SORT d._key RETURN d`;
+const query2 = `FOR d IN ${cn} FILTER d.value == 2 SORT d._key RETURN d`;
+
+
 let explain = function (query, params) {
   return helper.removeClusterNodes(helper.getCompactPlan(AQL_EXPLAIN(query, params, { optimizer: { rules: [ "-all", "+use-indexes" ] } })).map(function(node) { return node.type; }));
 };
 
+let testandValidate = function (query, isUnique, insertMoreDocs, insertIndexBeforeDocs, sameValue, isPrimary, selectivity) {
```
jsteemann, line 45:
```suggestion
let testAndValidate = function (query, isUnique, insertMoreDocs, insertIndexBeforeDocs, sameValue, isPrimary, selectivity) {
```

tests/js/server/aql/aql-persistent-index.js
```diff
@@ -30,20 +30,64 @@ const internal = require("internal");
...
+  let ensureIndex = () => {
+    internal.db[cn].ensureIndex({type: "persistent", name: idxName, unique: isUnique, fields: ["value"]});
+  };
+
+  if (insertMoreDocs) {
+    if (insertIndexBeforeDocs) {
```
jsteemann, line 51: Do we actually need to test the before/after index creation in our tests here?
I don't see the benefit of testing it here, because we will be waiting for all estimates to be synced anyway.
To simplify, we could just create the index in a way that we don't need to wait for the estimates to settle.

tests/js/server/aql/aql-persistent-index.js
```diff
@@ -30,20 +30,64 @@ const internal = require("internal");
...
+  for (const key in nodes) {
+    if (nodes[key].type === "IndexNode") {
+      const indexNode = nodes[key];
+      assertEqual(indexNode.indexes.length, 1);
+      const usedIndex = indexNode.indexes[0];
+      assertEqual(usedIndex.type, (isPrimary ? "primary" : "persistent"));
+      assertEqual(usedIndex.name, (isPrimary ? "primary" : idxName));
+      assertEqual(usedIndex.fields.length, 1);
+      assertEqual(usedIndex.fields[0], (isPrimary ? "_key" : "value"));
+      assertEqual(usedIndex.selectivityEstimate, selectivity);
```
jsteemann, line 79: If there is no IndexNode in the plan, none of the checks will be executed and the test will still pass.

jsteemann (approved):
LGTM

PR #19138, opened 2023-05-31
Fix TimeStampTransformer

markuspf:
Previously the implementation of `TimeStampTransformer`, which takes a `std::chrono::system_clock::time_point` and transforms it into a string used the ArangoDB library function `timepointToString`.

This function in turn converted the `time_point` into a `time_t` and then uses `TRI_gmtime` to format the `time_point`.

If one uses std::chrono::system_clock::time_point::min() as input value to timepointToString, this yields an invalid time_t on some platforms (in particular here Windows), leading to an error in gmtime.

The ArangoDB library function ignores that error and continues.

This leads to invalid characters being left in the serialisation of the timepoint.

This PR changes the `TimeStampTransformer` to use formatting and parsing functions from the `date.h` library to perform (de)serialisation of timestamps.

It also adds some tests.

### Scope & Purpose

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

### DIFF
--- a/lib/Inspection/include/Inspection/Transformers.h
+++ b/lib/Inspection/include/Inspection/Transformers.h
@@ -36,7 +36,7 @@ namespace arangodb::inspection {
 struct TimeStampTransformer {
   using SerializedType = std::string;
   using clock = std::chrono::system_clock;
-  static constexpr std::string_view formatString = "%Y-%m-%d %H:%M:%S";
+  static constexpr std::string_view formatString = "%FT%TZ";
   auto toSerialized(clock::time_point source, std::string& target) const
       -> inspection::Status {
     target = date::format(formatString.data(), source);
--- a/lib/Inspection/include/Inspection/Transformers.h
+++ b/lib/Inspection/include/Inspection/Transformers.h
@@ -39,7 +39,8 @@ struct TimeStampTransformer {
   static constexpr std::string_view formatString = "%FT%TZ";
   auto toSerialized(clock::time_point source, std::string& target) const
       -> inspection::Status {
-    target = date::format(formatString.data(), source);
+    target =
+        date::format(formatString.data(), floor<std::chrono::seconds>(source));
     return {};
   }
   auto fromSerialized(std::string const& source,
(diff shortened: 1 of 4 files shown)


lib/Inspection/include/Inspection/Transformers.h
```diff
@@ -26,21 +26,34 @@
...
+
+#include "date/date.h"
+#include "fmt/core.h"
 
 namespace arangodb::inspection {
 
 struct TimeStampTransformer {
   using SerializedType = std::string;
   using clock = std::chrono::system_clock;
+  static constexpr std::string_view formatString = "%Y-%m-%d %H:%M:%S";
```
mpoeter, line 39: Shouldn't we format this as UTC ISO8601? In that case the format string should look like this:
```suggestion
  static constexpr std::string_view formatString = "%Y-%m-%dT%H:%M:%SZ";
```

mpoeter (approved):

lib/Inspection/include/Inspection/Transformers.h
```diff
@@ -26,21 +26,35 @@
...
 
 struct TimeStampTransformer {
   using SerializedType = std::string;
   using clock = std::chrono::system_clock;
+  static constexpr std::string_view formatString = "%FT%TZ";
   auto toSerialized(clock::time_point source, std::string& target) const
       -> inspection::Status {
-    target = timepointToString(source);
+    target =
+        date::format(formatString.data(), floor<std::chrono::seconds>(source));
```
jsteemann, line 43: Is this safe? We are now passing a pointer into an `std::string_view`'s data.
AFAIK this pointer is not guaranteed to point to a null-terminated string: https://en.cppreference.com/w/cpp/string/basic_string_view/data
So how can `date::format(pointer, ...)` know where the format string ends?

PR #19139, opened 2023-06-01
experimental: enable SstFileWriter for index creation

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1288

Highly experimental: enable RocksDB SstFileWriter for index creation. Using the SstFileWrite to create index entries for an existing collection has several benefits over the traditional method of using RocksDB WriteBatches:
- the WAL is fully bypassed, thus we save the write to the WAL
- writes to memtables are bypassed
- writes do not need to bubble up the LSM tree as usual, thus we save all write-amplification that is present with normal RocksDB writes

The code that uses the SstFileWriter has never run in any larger scale tests and has been dormant for about one year. Before merging this PR it will need thorough testing.
It will also be good to check the status of https://github.com/facebook/rocksdb/issues/10007 before merging this PR. The RocksDB issue is about potential incompatibility between using `IngestExternalFile()` (which this PR starts using) and `GetUpdatesSince()` (which we are using for WAL tailing and replication).
The issue in RocksDB should be addressed first, so that `IngestExternalFile()` and `GetUpdatesSince()` become compatible with each other.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1288
- [ ] Design document:


neunhoef:
Nice idea, but not feasible for now. And we observed quite a few problems with direct sst file ingestion.

PR #19142, opened 2023-06-01
testing: test Filename to auto evaluation

dothebart:
### Scope & Purpose

if we don't find a testsuite by `NAME`, and `NAME` is an existing file, divert to interpret it as `auto --test NAME`

This can be used to instead of having to sort out the `auto` comand:
```
./scripts/unittest auto --test tests/js/client/shell/shell-admin-cluster-rebalance.js
```
directly invoke the test using the auto mechanism:
```
./scripts/unittest  tests/js/client/shell/shell-admin-cluster-rebalance.js 
```

- [x] :pizza: New feature

### DIFF
--- a/js/client/modules/@arangodb/testutils/testing.js
+++ b/js/client/modules/@arangodb/testutils/testing.js
@@ -451,7 +451,7 @@ function loadTestSuites () {
   testFuncs['auto'] = autoTest;
 }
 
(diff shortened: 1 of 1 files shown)


neunhoef (approved):
LGTM

apetenchea (approved):

PR #19144, opened 2023-06-01
Python toolbox

hkernbach:
### Scope & Purpose

Python3 based project to bundle tools for developers.
See: `scripts/toolbox/README.md`

- [x] :pizza: New feature

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/GORDO-1649


scripts/toolbox/modules/Feed.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+
+import subprocess
+
+
+def start(cfg, operations):
+    # Feed needs to have a local file as input
+    fileName = "work/feedInput.txt"
+    with open(fileName, "w") as text_file:
```
apetenchea, line 9: It is generally preferred to stick to either snake case or camel case. This is just a stylistic recommendation.

scripts/toolbox/modules/TestExecutor.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+import tests.RocksDBCacheSizeTest as RocksDBCacheSizeTest
```
apetenchea, line 3:
```suggestion
from tests import RocksDBCacheSizeTest
```

scripts/toolbox/tests/RocksDBCacheSizeTest.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def module_from_file(module_name, file_path):
+    spec = importlib.util.spec_from_file_location(module_name, file_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+Feed = module_from_file("Feed", "modules/Feed.py")
```
apetenchea, line 13: It is actually easier to import functionality from modules. This is all you need:
```suggestion
from modules import Feed
```

scripts/toolbox/config.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
apetenchea, line 1: I noticed that you prefer to use modules directly to organize the code. I usually use classes for this, but it's more of a matter of preference. The way you did it works perfectly fine.
I just want to point out that with this approach every module is like a "singleton" class. Modifying imported functionality affects it for the entire scope of the import.

```python
import config

def foo():
    print(config.arangodb.host)

foo()  # localhost
config.arangodb['host'] = 'abcd'
foo()  # abcd
```

Using a separate `config.json` and providing an ArangoDB class, for example, would force the users to work with instances of that class. It also has the benefit of aiding the autocomplete functionality of the IDE, if you provide these keys as class properties.

That being said, I know it may seem like a lot of work now, for just one test. This is not a request for change, but rather a note for the future.

scripts/toolbox/README.md
```diff
@@ -0,0 +1,39 @@
...
+### Installing
+
+* Currently, no specific installation is needed. Just clone the repository and run the scripts.
+
+### Executing program
+
+* Example of how to run the program
+
+```
+python3 main.py --mode=single --init=true --startupParameters='{"rocksdb.block-cache-size": "133713371337"}'
```
apetenchea, line 25: While trying this command on my machine, I get in _arangodb.log_:
`FATAL [d54b7] {general} cannot convert username 'arangodb' to numeric uid`

The output is:
```
Started ArangoDB in single mode, logfile: work/arangodb.log, port: 8529 (silently!)
--> Started test: "RocksDBCacheSizeTest". Output file is: "work/feed.json"
exit status 4
=> All tests finished
Stopping ArangoDB...
```
There is no feed.json output. Perhaps I am missing some local configuration. I don't mind it, but this is just to let you know that for me it doesn't work out-of-the-box.

scripts/toolbox/modules/Feed.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+
+import subprocess
+
+
+def start(cfg, operations):
```
apetenchea, line 6: I think we should create the _work_ folder here if it does not exist. I got an exception first time I ran it. Worked fine after I created it manually.

apetenchea (approved):
Congrats on what seems like a great initiative!

Feel free to treat all my comments as **recommendations**. The code looks correct and in-line with the provided functionality, hence I already approved the PR.

.gitmodules
```diff
@@ -23,3 +23,6 @@
 [submodule "3rdParty/rta-makedata"]
 	path = 3rdParty/rta-makedata
 	url = https://github.com/arangodb/rta-makedata
+[submodule "scripts/toolbox/submodules/feed"]
```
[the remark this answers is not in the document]
hkernbach: Thank you, done :)

scripts/toolbox/modules/Feed.py
hkernbach: Applied

scripts/toolbox/modules/TestExecutor.py
hkernbach: This actually works. For whatever reasons, my IDE says this is wrong: 
image

scripts/toolbox/tests/RocksDBCacheSizeTest.py
hkernbach: Indeed .. awesome. IDE also confused me on this one: 

image

hkernbach: image

scripts/toolbox/modules/Feed.py
hkernbach: I see, will add it!

scripts/toolbox/tests/RocksDBCacheSizeTest.py
hkernbach: ^-- Issues has been my local IDE. just as a note.

scripts/toolbox/modules/Feed.py
hkernbach: Added

scripts/toolbox/README.md
hkernbach: Resolved :)

PR #19154, opened 2023-06-02
Fix IN lookup

jsteemann:
### Scope & Purpose

https://arangodb.atlassian.net/browse/OASIS-25262

* OASIS-25262: Fixed undefined behavior in IN lookup in unique indexes when the lookup array had to be rebuilt in memory.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19153
  - [ ] Backport for 3.10: not required
  - [ ] Backport for 3.9: not required

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/OASIS-25262
- [ ] Design document:


markuspf:
Can you state what constituted the UB? Just the fact that `_current` was left uninitialized?

jsteemann:
> Can you state what constituted the UB? Just the fact that `_current` was left uninitialized?

The IN lookup builder has a vpack Builder `_searchValues` and a vpack ArrayIterator `_current`. The invariant is that `_current` always points into memory owned by `_searchValues`.
This is also actually the case, with the one exception fixed in this PR: when the lookup condition needed to be reformatted (for internal reasons), then `_searchValues` was first emptied and then some data was added to it.
That adding of data could cause a problem without the fix, because it could lead to `_searchValues` making a memory allocation and uses a different memory addresses for its owned data. If that happened, then `_searchValues` owned a new memory region, and `_current` still pointed into the old region, which was already disowned.
The fix is to update `_current` to always point into the region owned by `_searchValues` after `_searchValues` is modified.
Note: a memory allocation is not always necessary. It only happens if the buffer already owned by `_searchValues` is not big enough to hold the new data. That's why this case slipped through the cracks.

MBkkt (approved):

neunhoef (approved):
LGTM

markuspf (approved):

PR #19156, opened 2023-06-02
Use libunwind in jemalloc heap profiling

neunhoef:
Use libunwind in jemalloc heap profiling if available.

This is necessary to get jemalloc heap profiling to work with libmusl
builds and static binaries.

This implements the following:

If USE_LIBUNWIND is on, then we add a dependency in the build process
such that jemalloc_build can rely on the fact that libunwind is already
built. Furthermore, the configure script of jemalloc gets the newly
built libunwind.a and the included headers and can then be switched to
libunwind profiling mode.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Backports
  - [*] Backport for 3.11: 
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19155

There is no enterprise part.


CMakeLists.txt
```diff
@@ -757,7 +757,7 @@ endfunction()
 # JEMALLOC
 # ------------------------------------------------------------------------------
 
-option(USE_JEMALLOC_PROF "use jemalloc profiler" OFF)
+option(USE_JEMALLOC_PROF "use jemalloc profiler" ON)
```
jsteemann, line 760: Do we really want to turn on profiling by default?

jsteemann (approved):
LGTM if this PR does not enable the profiling code by default.
I can't estimate the effects of including the profiling code on runtime performance.
If we want to turn on the profiling code by default, we should first run a side-by-side performance comparison of standard `devel` with a branch that turns the profiling code on, and then see if the performance results are "ok enough" to use that.

3rdParty/jemalloc/CMakeLists.txt
```diff
@@ -31,8 +31,14 @@ if (LINUX OR DARWIN)
...
+  SET(JEMALLOC_C_FLAGS ${CMAKE_C_FLAGS})
   if (USE_JEMALLOC_PROF)
-    SET(JEMALLOC_PROF "--enable-prof")
+      if (USE_LIBUNWIND)
+          SET(JEMALLOC_PROF "--enable-prof" "--enable-prof-libunwind" "--with-static-libunwind=${LIBUNWIND_LIB}")
+          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -I${LIBUNWIND_HOME}/include")
+      else ()
+          SET(JEMALLOC_PROF "--enable-prof")
+      endif()
   endif ()
```
MBkkt, line 42: I think set to the CMAKE_C_FLAGS will be ok, because we're in jemalloc cmakelists.txt.
But it doesn't really matter

MBkkt (approved):

neunhoef:
Here is some evidence that compares a build without profiling with one with profiling (inactive) and with profiling (active) from our simple performance tests:
[comparison.txt](https://github.com/arangodb/arangodb/files/11675266/comparison.txt)

And here is some data from a program which does only allocations and deallocations in many threads:

```
jemalloc, no prof, libmusl, static:
  4.27  4.29  4.30  4.29  4.28  4.29  4.32  4.27  4.31  4.32
Median 4.29

jemalloc, prof compiled in, switched off, libmusl, static:
  4.29  4.26  4.32  4.31  4.31  4.32  4.28  4.30  4.29  4.36
Median 4.30

jemalloc, prof compiled in, switched on, libmusl, static:
  4.93  4.96  4.98  4.94  4.92  4.98  4.98  5.06  4.92  4.99
Median 4.96
```

neunhoef:
Conclusion: Compiling in profiling support does not cost any measurable performance, not even in a program which does lots of allocations in many threads. Activating it can cost some performance, but even in a program which only allocates and frees the overhead is some 15%.
Therefore this PR switches USE_JEMALLOC_PROF by default to 1.

markuspf (approved):
LGTM and I agree that there is no discernible performance penalty when compiling profiling in as a default.

There are big upsides in being able to profile without having to recompile!

PR #19158, opened 2023-06-02
Use libunwind in jemalloc heap profiling

neunhoef:
Use libunwind in jemalloc heap profiling if available.

This is necessary to get jemalloc heap profiling to work with libmusl
builds and static binaries.

This implements the following:

If USE_LIBUNWIND is on, then we add a dependency in the build process
such that jemalloc_build can rely on the fact that libunwind is already
built. Furthermore, the configure script of jemalloc gets the newly
built libunwind.a and the included headers and can then be switched to
libunwind profiling mode.

Original PR: https://github.com/arangodb/arangodb/pull/19156

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Backports


markuspf (approved):
LGTM

maierlars (approved):

PR #19173, opened 2023-06-04
[SEARCH-480] [APM-766] Speedup ArangoSearch recovery

MBkkt:
### Scope & Purpose

* Remove arangodb transaction abstractions from ArangoSearch recovery
* Add caching for get links/indexes in PutCF/DeleteCF
* Add batching for removes in recovery (for inserts too, but for them almost no difference)
* Also make OutOfSync more optimistic:
  Before these changes, even if link doesn't need recovery it can be marked as OutOfSync

For single link and 1e6 replaces (remove/insert) in recovery this PR improves time in 2 times:
30 sec instead of 1 min

For multiple links improvement could be in a few times better.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1418,12 +1418,8 @@ Result IResearchDataStore::initDataStore(
         }
 
         // Make last batch Commit if necessary
-        if (auto& removes = linkLock->_recoveryRemoves; removes != nullptr) {
-          if (!removes->empty()) {
-            linkLock->_recoveryTrx.Remove(std::move(removes));
-          }
-          removes.reset();
-          linkLock->_recoveryTrx.Commit(recoveryTick);
+        if (linkLock->_recoveryRemoves != nullptr) {
+          linkLock->recoveryCommit(recoveryTick);
         }
 
         // register flush subscription
@@ -1836,6 +1832,15 @@ std::tuple<uint64_t, uint64_t, uint64_t> IResearchDataStore::avgTime() const {
 }
 
 #endif
+
+void IResearchDataStore::recoveryCommit(uint64_t tick) {
+  TRI_ASSERT(_recoveryRemoves);
+  if (!_recoveryRemoves->empty()) {
+    _recoveryTrx.Remove(std::move(_recoveryRemoves));
+  }
+  _recoveryRemoves.reset();
+  _recoveryTrx.Commit(tick);
+}
 
 void IResearchDataStore::initClusterMetrics() const {
   TRI_ASSERT(ServerState::instance()->isCoordinator());
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1504,7 +1504,6 @@ void IResearchDataStore::properties(LinkLock linkLock,
 }
 
 IResearchTrxState* IResearchDataStore::getContext(TransactionState& state) {
-  TRI_ASSERT(!isOutOfSync());
   void const* key = this;
   auto* context = basics::downCast<IResearchTrxState>(state.cookie(key));
   if (ADB_LIKELY(context)) {
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1069,6 +1069,8 @@ void IResearchDataStore::shutdownDataStore() noexcept {
         << "caught something while removeMetrics arangosearch data store '"
         << index().id().id() << "'";
   }
+  _recoveryRemoves.reset();
+  _recoveryTrx.Abort();
   _dataStore.resetDataStore();
 }
 
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1645,7 +1645,7 @@ void IResearchDataStore::recoveryInsert(uint64_t recoveryTick,
   try {
     std::ignore = insertDocument<FieldIteratorType>(*this, _recoveryTrx, doc,
                                                     documentId, meta);
-  } catch (std::bad_alloc const& e) {
+  } catch (std::bad_alloc const&) {
     throw;
   } catch (...) {
   }
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1691,6 +1691,7 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
   TRI_ASSERT(_dataStore);  // must be valid if _asyncSelf->get() is valid
 
   if (trx != nullptr) {
+    TRI_ASSERT(_recoveryRemoves == nullptr);
     auto* key = this;
 
     auto& state = *(trx->state());
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -227,9 +227,9 @@ auto getIndexFeatures() {
 template<typename FieldIteratorType, typename MetaType>
 Result insertDocument(IResearchDataStore const& dataStore,
                       irs::IndexWriter::Transaction& ctx,
-                      transaction::Methods const& trx,
                       velocypack::Slice document, LocalDocumentId documentId,
-                      MetaType const& meta, IndexId id) {
+                      MetaType const& meta) {
+  auto const id = dataStore.index().id();
   auto const collectionNameString = [&] {
     if (!ServerState::instance()->isSingleServer()) {
       return std::string{};
@@ -243,7 +243,7 @@ Result insertDocument(IResearchDataStore const& dataStore,
   std::string_view collection = collectionNameString.empty()
                                     ? meta.collectionName()
                                     : collectionNameString;
-  FieldIteratorType body{collection, dataStore.index().id()};
+  FieldIteratorType body{collection, id};
   body.reset(document, meta);  // reset reusable container to doc
 
   if (!body.valid()) {
@@ -280,7 +280,7 @@ Result insertDocument(IResearchDataStore const& dataStore,
 
   // Sorted field
   {
-    SortedValue field{collection, dataStore.index().id(), document};
+    SortedValue field{collection, id, document};
     for (auto& sortField : meta._sort.fields()) {
       field.slice = get(document, sortField, VPackSlice::nullSlice());
       doc.template Insert<irs::Action::STORE_SORTED>(field);
@@ -289,7 +289,7 @@ Result insertDocument(IResearchDataStore const& dataStore,
 
   // Stored value field
   {
-    StoredValue field{collection, dataStore.index().id(), document};
+    StoredValue field{collection, id, document};
     for (auto const& column : meta._storedValues.columns()) {
       field.fieldName = column.name;
       field.fields = &column.fields;
@@ -339,6 +339,15 @@ auto makeAfterCommitCallback(void* key) {
     }
   };
 }
+
+std::string MakeMessage(std::string_view begin, Index const& index,
+                        TransactionState& state, LocalDocumentId documentId,
+                        auto&&... args) {
+  return absl::StrCat(begin, " ArangoSearch index: ", index.id().id(),
+                      ", tid: ", state.id().id(),
+                      ", documentId: ", documentId.id(),
+                      std::forward<decltype(args)>(args)...);
+};
 
 static std::atomic_bool kHasClusterMetrics{false};
 
@@ -635,7 +644,6 @@ IResearchDataStore::IResearchDataStore(
     : _asyncFeature(&server.getFeature<IResearchFeature>()),
       // mark as data store not initialized
       _asyncSelf(std::make_shared<AsyncLinkHandle>(nullptr)),
-      _error(DataStoreError::kNoError),
       _maintenanceState(std::make_shared<MaintenanceState>()) {
   // initialize transaction callback
 #ifdef USE_ENTERPRISE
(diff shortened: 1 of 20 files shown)


arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1632,107 +1585,86 @@ bool IResearchDataStore::exists(IResearchDataStore::Snapshot const& snapshot,
...
+  }
+  if (!_recoveryTrx.FlushRequired()) {
+    return;
+  }
+  if (!_recoveryRemoves->empty()) {
+    _recoveryTrx.Remove(std::move(_recoveryRemoves));
+    _recoveryRemoves = makePrimaryKeysFilter(_hasNestedFields);
+  }
+  _recoveryTrx.Commit(recoveryTick);
+  // TODO(MBkkt) IndexWriter::Commit? Probably makes sense only if were removes
```
MBkkt, line 1665: Probably makes sense even without removes.
Because if we will make commit, even if recovery crashes, already committed part won't be executed second time

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -65,316 +65,270 @@
...
-      // Intentional NOOP. Index is committed upon creation.
-      // So if this marker was written  - index was persisted already.
+    case RocksDBLogType::DatabaseDrop:
+    case RocksDBLogType::CollectionDrop:
+      // TODO(MBkkt) Do we really need clear for these?
+    case RocksDBLogType::DatabaseCreate:
+    case RocksDBLogType::CollectionCreate:
+    case RocksDBLogType::IndexCreate:
+    case RocksDBLogType::IndexDrop:
+      // TODO(MBkkt) Clear only necessary entries
```
MBkkt, line 232: I mean we could get objectId and remove it from _ranges.
```cpp
auto const objectId = RocksDBLogValue::objectId(blob);
_ranges.erase(objectId);
```
But I don't sure will something like this correct, or it could return something invalid

arangod/IResearch/IResearchPrimaryKeyFilter.cpp
```diff
@@ -58,9 +58,9 @@ bool PrimaryKeysFilter<Nested>::next() {
     auto const pkRef =
         irs::numeric_utils::numeric_traits<LocalDocumentId::BaseType>::raw_ref(
             pk);
-    auto doc = irs::doc_limits::eof();
+    auto doc = irs::doc_limits::invalid();
```
MBkkt, line 61: eof also shouldn't be in segment
But this code is more correct and also condition with `0x0` instead of `0xffff`

MBkkt:
Tests passed except known cluster sanitizers failures

arangod/IResearch/IResearchRocksDBLink.h
```diff
@@ -49,25 +49,17 @@ class IResearchRocksDBLink final : public RocksDBIndex, public IResearchLink {
...
-  }
-
-  Result removeInRecovery(transaction::Methods& trx,
-                          LocalDocumentId const& documentId,
-                          rocksdb::SequenceNumber tick) {
-    return IResearchDataStore::remove(trx, documentId, meta().hasNested(),
-                                      &tick);
+  void recoveryInsert(uint64_t tick, LocalDocumentId documentId,
+                      VPackSlice doc) {
+    return IResearchDataStore::recoveryInsert<FieldIterator<FieldMeta>,
```
Dronplane, line 60: return not needed

arangod/IResearch/IResearchDataStore.h
```diff
@@ -505,6 +508,8 @@ class IResearchDataStore {
   std::mutex _commitMutex;
 
   // for insert(...)/remove(...)
+  irs::IndexWriter::Transaction _recoveryTrx;
```
Dronplane, line 511: at least this transaction should be aborted explicitly in postRecoveryCallback. But iy would be better to not create it at all if not in recovery.

MBkkt: Fixed

markuspf (commented):
It is really hard for an outside person like me to review and gauge correctness of a PR like this.

One reason for this is that it conflates (as far as I understand) multiple things:


> * Remove arangodb transaction abstractions from ArangoSearch recovery

It looks like this was unused, and this could have been a self-contained, easy-to-review PR, with a description why it is removed.

> *   Add caching for get links/indexes in PutCF/DeleteCF
> *   Add batching for removes in recovery (for inserts too, but for them almost no difference)
> *  Also make OutOfSync more optimistic

Could these three things have been done separately? It is hard to tell in a review which change relates to which of these points. I can just about tell where `outOfSync` is handled.
If it needs to be one PR for some reason, it would be nice to have a commit history that reflects the separate steps that are mentioned in the list above, so one can tell by reading the commits which changes are related.

> For single link and 1e6 replaces (remove/insert) in recovery this PR improves time in 2 times:
> 30 sec instead of 1 min

How is this benchmarked and how can I reproduce this? Is it a thing that can be benchmarked as part of CI/Nightlies/Benchmarks?

There are also changes in the 3rdParty iresearch library which I cannot currently review or judge. At this point in time I can only treat the iresearch-toolkit as a black box with functional guarantees on which I can rely, and I will also have to learn at least those guarantees.

For that it is necessary for them to be explained/documented. I am grateful to be pointed towards documentation, and even more grateful for explicit direct explanations.

Other than the above I found a typo in a variable name, but could find any problems with the code.

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -65,316 +65,268 @@
...
-void IResearchRocksDBRecoveryHelper::PutCF(uint32_t column_family_id,
-                                           const rocksdb::Slice& key,
-                                           const rocksdb::Slice& value,
-                                           rocksdb::SequenceNumber tick) {
-  if (column_family_id != _documentCF) {
-    return;
+template<bool Force>
+void IResearchRocksDBRecoveryHelper::clear() noexcept {
+  if constexpr (!Force) {
+    if (_indexes.size() < kMaxSize || _links.size() < kMaxSize ||
```
Dronplane, line 104:
```suggestion
    if (_indexes.size() < kMaxSize && _links.size() < kMaxSize &&
```

MBkkt: Good catch!

arangod/IResearch/IResearchRocksDBLink.h
MBkkt: Fixed

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
MBkkt: Fixed

MBkkt:
@markuspf 

> It looks like this was unused, and this could have been a self-contained, easy-to-review PR, with a description why it is removed.

It's the main part of fix (according to profiler) and this change caused most other changes

> Could these three things have been done separately? It is hard to tell in a review which change relates to which of these points. I can just about tell where outOfSync is handled.

Because of the first point I needed to write it from scratch.
So I decided to write better version.

> If it needs to be one PR for some reason, it would be nice to have a commit history that reflects the separate steps that are mentioned in the list above, so one can tell by reading the commits which changes are related.

I understand that it's better, but unfortunately I mostly don't write code iterative, so I couldn't.

> How is this benchmarked and how can I reproduce this? Is it a thing that can be benchmarked as part of CI/Nightlies/Benchmarks?

We just disabled commit to the link (could be in the code or changes view defition commitIntervalMsec=0), made 1e6 replaces and restart server.
We used release build without maintainer mode.
Also we made same second time with profiler, to understand the issue better.

> There are also changes in the 3rdParty iresearch library which I cannot currently review or judge. At this point in time I can only treat the iresearch-toolkit as a black box with functional guarantees on which I can rely, and I will also have to learn at least those guarantees.
For that it is necessary for them to be explained/documented. I am grateful to be pointed towards documentation, and even more grateful for explicit direct explanations.

It will be good, but currently iresearch code has only poor doxygen docs, same as arangodb.
Feel free to improve this!

> Other than the above I found a typo in a variable name, but could find any problems with the code.

Thanks, fixed.

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
Dronplane: Looks like only `case RocksDBLogType::IndexCreate:` makes sense

MBkkt: Done

Dronplane (changes requested):

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -65,316 +65,260 @@
...
-
-      LinkContainer links;
-      auto mustReplay = lookupLinks(links, *coll);
+      // TODO(MBkkt) Truncate could recover index from OutOfSync state
+      auto const objectId = RocksDBLogValue::objectId(blob);
 
-      if (links.empty()) {
-        // no links found. nothing to do
-        TRI_ASSERT(!mustReplay);
+      auto ranges = getRanges(objectId);
```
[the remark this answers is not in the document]
MBkkt: Done

arangod/IResearch/IResearchDocument.cpp
```diff
@@ -169,7 +169,7 @@ arangodb::iresearch::FieldMeta const* findMeta(
   return meta != context->_fields.end() ? &meta->second : context;
 }
 
-bool inObjectFiltered(std::string& buffer,
+bool inObjectFiltered(std::string& buffer, size_t buffer_hash,
```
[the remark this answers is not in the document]
MBkkt: Yep

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -216,7 +216,8 @@ void IResearchRocksDBRecoveryHelper::LogData(rocksdb::Slice const& blob,
   auto const type = RocksDBLogValue::type(blob);
 
   switch (type) {
-    case RocksDBLogType::IndexCreate: {
+    case RocksDBLogType::IndexCreate:
+    case RocksDBLogType::IndexDrop: {
```
Dronplane, line 220: Why Drop?

MBkkt: Because it invalidates pointer in the range (on practice everything will be ok because we're holding reference in the post recovery callback, but we will just execute unnecessary inserts/removes)

Dronplane:
Also fixes https://arangodb.atlassian.net/browse/SEARCH-476 (as iresearch update from upstream)

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1500,109 +1505,55 @@ void IResearchDataStore::properties(LinkLock linkLock,
...
-  if (recoveryTick == nullptr) {
-    // skip recovery check
-  } else if (*recoveryTick <= _dataStore._recoveryTickLow) {
-    LOG_TOPIC("6e128", TRACE, TOPIC)
-        << "skipping 'exists', operation tick '" << *recoveryTick
-        << "', recovery tick low '" << _dataStore._recoveryTickLow << "'";
+void IResearchDataStore::recoveryRemove(LocalDocumentId documentId) {
+  TRI_ASSERT(!isOutOfSync());
+  TRI_ASSERT(_recoveryRemoves);
+  _recoveryRemoves->emplace(documentId);
```
jsteemann, line 1550: This buffers all removes for the recovery in main memory.
What if there are a lot of remove operations to recover? Wouldn't this use a large amount of main memory then, and could lead to OOM?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1612,16 +1563,15 @@ bool IResearchDataStore::exists(IResearchDataStore::Snapshot const& snapshot,
       irs::numeric_utils::numeric_traits<LocalDocumentId::BaseType>::raw_ref(
           encoded);
 
-  for (auto const& segment : snapshot.getDirectoryReader()) {
+  for (auto const& segment : reader) {
     auto const* pkField = segment.field(DocumentPrimaryKey::PK());
 
     if (ADB_UNLIKELY(!pkField)) {
       continue;
```
jsteemann, line 1570: Is it actually possible to get here?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1773,6 +1703,9 @@ void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
       // clear
       state.cookie(key, nullptr);
     }
+  } else if (_recoveryRemoves != nullptr) {  // TODO(MBkkt) should be assert?
+    _recoveryRemoves->clear();
+    _recoveryTrx.Abort();
```
jsteemann, line 1708: Does calling `_recoveryTrx.Abort()` allow to put more data into the transaction after the abort?

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1632,107 +1582,86 @@ bool IResearchDataStore::exists(IResearchDataStore::Snapshot const& snapshot,
...
+                                                    documentId, meta);
+  } catch (std::bad_alloc const&) {
+    throw;
+  } catch (...) {
+  }
+  if (!_recoveryTrx.FlushRequired()) {
+    return;
+  }
+  if (!_recoveryRemoves->empty()) {
+    _recoveryTrx.Remove(std::move(_recoveryRemoves));
```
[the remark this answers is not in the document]
MBkkt: Only bad_alloc

arangod/IResearch/IResearchDataStore.cpp
MBkkt: Yes. It will be new Transaction.
like
```cpp
vector trx1;
... do somethign with trx1
trx1 = {};
... do somethign with trx1 (trx2)
```

MBkkt: Simplified IndexWriter::Transaction is just a batch of operations.
They could be committed via Commit and dropped via Abort.

arangod/IResearch/IResearchDataStore.cpp
MBkkt: I don't think so, probably should be assert, but I kept old behavior

arangod/IResearch/IResearchDataStore.cpp
MBkkt: > This buffers all removes for the recovery in main memory.
What if there are a lot of remove operations to recover? Wouldn't this use a large amount of main memory then, and could lead to OOM?

Yes it's possible, but it was possible (but was easy in the 10~ times) -- they just kept in the IndexWriter instead of filter before commit in the postRecoveryCallback
These changes first step to fix this.
Then in general we need to enable IndexWriter::Commit in the background before recovery ends.
And then we need to have settings to restrict how many removes could be in the batch.

PR #19177, opened 2023-06-05
[BTS-1325] [3.11] Don't go to sleep forever

goedderz:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/18729; see there for details.

### DIFF
--- a/arangod/Aql/AsyncExecutor.h
+++ b/arangod/Aql/AsyncExecutor.h
@@ -42,7 +42,18 @@ class SharedQueryState;
 
 // The RemoteBlock is actually implemented by specializing ExecutionBlockImpl,
 // so this class only exists to identify the specialization.
-class AsyncExecutor final {};
+class AsyncExecutor final {
+ public:
+  struct Properties {
+    static constexpr bool preservesOrder = true;
+    static constexpr BlockPassthrough allowsBlockPassthrough =
+        BlockPassthrough::Enable;
+    static constexpr bool inputSizeRestrictsOutputSize = false;
+  };
+  // using Fetcher = std::monostate;
+  using Infos = std::monostate;
+  // using Stats = NoStats;
+};
 
 /**
  * @brief See ExecutionBlockImpl.h for documentation.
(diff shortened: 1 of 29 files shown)


mpoeter (approved):

arangod/Aql/AsyncExecutor.h
```diff
@@ -42,7 +42,18 @@ class SharedQueryState;
 
 // The RemoteBlock is actually implemented by specializing ExecutionBlockImpl,
 // so this class only exists to identify the specialization.
-class AsyncExecutor final {};
+class AsyncExecutor final {
```
MBkkt, line 45: All this structure looks unused except monostate, but it already was. Did I miss something?

arangod/Aql/ExecutionState.cpp
```diff
@@ -23,35 +23,45 @@
...
     case ExecutionState::WAITING:
-      ostream << "WAITING";
-      break;
+      return "WAITING";
   }
-  return ostream;
+  ADB_PROD_ASSERT(false)
+      << "Unhandled state "
+      << static_cast<std::underlying_type_t<decltype(state)>>(state);
+  std::abort();
```
MBkkt, line 44: Why? Isn't assert(false) no return?

goedderz: It is, but not all compilers believe/prove that - iirc it was MSVC that complains with a warning about a missing return in a non-void function.

MBkkt: Msvc needs `__assume(false)` for unreachable "tag".
Btw absl has cross platform unreachable marker

MBkkt: Also it can be valid detection.
We saw that with msvc after triggered assert code continue works
(Probably out assert written incorrect)

MBkkt: Anyway resolved

arangod/Aql/AsyncExecutor.h
MBkkt: I just don't see their usage in the new diff

MBkkt: Maybe I just missed something

goedderz: Yes, it's far from obvious. The `AsyncExecutor` was not used in the C++ tests until now. It now is, and uses the existing `ExecutorTestHelper` (e.g.
```cpp
  auto testHelper = makeExecutorTestHelper();
  testHelper
      .addDependency<AsyncExecutor>(registerInfos, {}, ExecutionNode::ASYNC)
      ...
```
). This needs the `Infos`, for example in `addDependency`:
```cpp
  template<typename E>
  auto addDependency(
      RegisterInfos registerInfos, typename E::Infos executorInfos,
      ExecutionNode::NodeType nodeType = ExecutionNode::SINGLETON)
      -> ExecutorTestHelper&;
```

And as I said, the Properties aren't currently necessary, but I think they're useful documentation of a block's behavior.

PR #19181, opened 2023-06-05
Fix typos in hotbackup error messages

markuspf:
### Scope & Purpose

If `arangobackup` exits with an exception it used to show `arangodump` as the command that failed. This might be confusing so it might well be fixed right away. It is unclear to me how one could even trigger this kind of failure, though.


neunhoef (approved):
LGTM

Issue #19175, opened 2023-06-05
AQL: cannot and-combine normalized condition


aquamatthias:
I updated my ArangoDB installation to 3.11 and now I see queries failing with this message:

```
Error: AQLQueryExecuteError
Message: [HTTP 500][ERR 4] AQL: cannot and-combine normalized condition (exception location: /Users/jenkins/hw22-mac-m1/oskar/work/ArangoDB/arangod/Aql/Condition.cpp:668). Please report this error to arangodb.com (while optimizing plan) (exception location: /Users/jenkins/hw22-mac-m1/oskar/work/ArangoDB/arangod/RestHandler/RestCursorHandler.cpp:279). Please report this error to arangodb.com
```

The same queries worked in older versions of ArangoDB so I consider this a regression.

## My Environment

* __ArangoDB Version__:        3.11.0
* __Deployment Mode__:      Single Server
* __Deployment Strategy__:  Manual Start
* __Operating System__:        MacOS 13.4
* __Total RAM in your machine__:        16GB
* __Disks in use__:  SSD

## Component, Query & Data

__Affected feature__:
AQL

__AQL query (if applicable)__:
```
LET filter0 = (FOR m0 in `resoto` FILTER @b0 IN m0.kinds  RETURN m0)LET m1 =( FOR c0_0 in filter0 FOR c0_1 IN 0..1 OUTBOUND c0_0 resoto_default OPTIONS { bfs: true, uniqueVertices: 'global' } FILTER (c0_0._key==c0_1._key) or ((@b1 IN c0_1.kinds) and (c0_1.trail_status.is_logging == @b2)) COLLECT l0_l0_res=c0_0 WITH COUNT INTO counter1 FILTER counter1==0  RETURN l0_l0_res)  FOR result in m1 RETURN UNSET(result, ["flat"])
```
with following bind vars:
```json
{
        "b0": "aws_region",
        "b1": "aws_cloud_trail",
        "b2": true
}
```

__AQL explain and/or profile (if applicable)__:

```
Error: AQLQueryExecuteError
[...]

PR #19188, opened 2023-06-06
[BTS-1443] Fix Github issue #19175

hkernbach:
### Scope & Purpose

Fix an issue and added a regression test for issue #19175, found by a community user.

The issue was caused by the following:

The expected usage of an AQL `Condition` object is to append things to it via `Condition::andCombine(...)` any number of times, and then at the end call `Condition::normalize(...)` exactly once on it. And then not call `Condition::andCombine(...)` on the already normalized object anymore. This is because `andCombine(...)` will throw if it is invoked on an already normalized condition.

However, there was one glitch: up to including 3.10, an AQL `Condition` object would not set its `_isNormalized` boolean flag when calling `Conndition::normalize(...)`, so the fact that the condition was normalized was not properly recoded. This looked unintuitive and wrong. So one could normalize a condition and then end up with the Condition’s `_isNormalized` flag still being false. And any follow-up calls to `andCombine(...)` would not throw but succeed.

And exactly that behavior was changed in 3.11: whenever `Condition::normalize(...)` was called, we would now set `_isNormalized = true`. Any following calls to `andCombine(...)` would then fail, as in the issue reported by the user.

The change was believed to be ok, because all existing tests passed with the change in place.

It seems that the issue reported by the user was triggered by some traversal optimization code that tried to pull filter conditions into the traversal. And in this code, the condition was first normalized and then `andCombine(...)` was called, which violated the above assumptions. However, that traversal optimization code relies on `normalize(...)` being executed first, so we can’t make `andCombine(...)` fail here.

The solution I picked in my fix is to all invocations of `andCombine(...)` even for normalized conditions. In fact, this was already allowed before, because calling `normalize(...)` on a condition previously did not set `_isNormalized = true`. So we can now officially allow it. We can still set `_isNormalized = true` in `normalize(...)`. Additionally, after invoking `andCombine(...)` on a condition, we need to reset its `_isNormalized` flag.

This change makes the new test pass.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19264
  - [ ] Backport for 3.10: not needed
  - [ ] Backport for 3.9: not needed

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] Jira ticket: https://arangodb.atlassian.net/browse/BTS-1443
- [x] GitHub issue: https://github.com/arangodb/arangodb/issues/19175
- [ ] Design document:


jsteemann (approved):
LGTM

arangod/Aql/Condition.cpp
```diff
@@ -678,6 +672,12 @@ void Condition::andCombine(AstNode const* node) {
   }
 
   TRI_ASSERT(_root != nullptr);
+
+  // note: it seems that andCombine() is sometimes called even for
+  // conditions that have been normalized already. in this case, we
+  // clear the normalization flag again after we have modified the
+  // condition.
+  _isNormalized = false;
```
markuspf, line 680: OOI: How expensive is rerunning the normalisation?

markuspf (approved):
LGTM; one question left.

PR #19194, opened 2023-06-06
Fixing Node API.

maierlars:
### Scope & Purpose
Lets get rid of the `std::optional<std::reference_wrapper<T>>` abomination, which was caused by a brain fart of mine. Now, I'm aware of the fact, that the agency will crash with a nullptr access, instead of just logging `std::bad_optional_access` over and over again. I think it is better now, because you see _where_ the nullptr access happened.


MBkkt (approved):

MBkkt (commented):
I think it should be merged even in the current state.

But it looks like adding `TRI_ASSERT(pointer != nullptr)` to the place where it dereferences will be useful

neunhoef (approved):
LGTM

PR #19200, opened 2023-06-07
[BTS-1454] cppcheck fixes

maierlars:
### Scope & Purpose

Fix some cpp check issues.


arangod/RocksDBEngine/RocksDBTransactionCollection.cpp
```diff
@@ -63,6 +63,7 @@ RocksDBTransactionCollection::RocksDBTransactionCollection(
 
 RocksDBTransactionCollection::~RocksDBTransactionCollection() {
   try {
+    // cppcheck-suppress virtualCallInConstructor
     releaseUsage();
```
goedderz, line 67: I guess you could also do
```suggestion
RocksDBTransactionCollection::releaseUsage();
```
instead of the suppression. I don't care, though.

goedderz (approved):

arangod/Cluster/HeartbeatThread.cpp
```diff
@@ -362,7 +362,7 @@ void HeartbeatThread::run() {
       runSingleServer();
     }
   } else if (ServerState::instance()->isAgent(role)) {
-    runAgent();
```
markuspf, line 365: Why is itcorrect to just remove this?

arangod/RocksDBEngine/RocksDBTransactionCollection.cpp
```diff
@@ -63,6 +63,7 @@ RocksDBTransactionCollection::RocksDBTransactionCollection(
 
 RocksDBTransactionCollection::~RocksDBTransactionCollection() {
   try {
+    // cppcheck-suppress virtualCallInConstructor
```
MBkkt, line 66: Is cppcheck warning about this even for final functions?

MBkkt (approved):

arangod/Cluster/HeartbeatThread.cpp
maierlars: because on the agency we don't need a heartbeat thread and the implementation literally does _nothing_.

PR #19207, opened 2023-06-07
[APM-764] Fast arangodump

maierlars:
### Scope & Purpose
Improve the dump performance of arangodump by using a new dump api and parallelise writing and receiving of data.
https://arangodb.atlassian.net/browse/APM-764

- [x] add tests for `_api/dump`.
- [ ] revisit the implementation of `BoundedChannel<T>`.


arangod/RocksDBEngine/RocksDBDumpContext.h
```diff
@@ -0,0 +1,238 @@
...
+  // return id of the context. will not change during the lifetime of the
+  // context.
+  std::string const& id() const noexcept;
+
+  // return database name used by the context. will not change during the
+  // lifetime of the context.
+  std::string const& database() const noexcept;
+  // return name of the user that created the context. can be used for access
+  // permissions checking. will not change during the lifetime of the context.
+  std::string const& user() const noexcept;
```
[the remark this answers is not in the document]
maierlars: to safe allocation when a string is needed.

client-tools/Dump/DumpFeature.h
```diff
@@ -135,15 +141,69 @@ class DumpFeature final : public ArangoDumpFeature {
...
+    std::string const server;
+    std::atomic<std::uint64_t> _batchCounter{0};
+    std::string dumpId;
+
+    enum BlockAt { kLocalQueue = 0, kRemoteQueue = 1 };
+
+    void countBlocker(BlockAt where, int64_t c);
+    void printBlockStats();
+
+    std::atomic<std::int64_t> blockCounter[2];
```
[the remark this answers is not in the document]
maierlars: noise

client-tools/Dump/DumpFeature.h
```diff
@@ -135,15 +141,69 @@ class DumpFeature final : public ArangoDumpFeature {
...
+    std::unique_ptr<httpclient::SimpleHttpResult> receiveNextBatch(
+        httpclient::SimpleHttpClient&, std::uint64_t batchId,
+        std::optional<std::uint64_t> lastBatch);
+
+    void createDumpContext(httpclient::SimpleHttpClient& client);
+    void finishDumpContext(httpclient::SimpleHttpClient& client);
+
+    ClientManager& clientManager;
+    std::shared_ptr<DumpFileProvider> const fileProvider;
+    std::unordered_map<std::string, ShardInfo> const shards;
```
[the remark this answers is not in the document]
maierlars: Yes reason because is doesn't matter.

lib/Basics/BoundedChannel.h
```diff
@@ -0,0 +1,143 @@
...
+#include <memory>
+#include <mutex>
+#include <vector>
+
+namespace arangodb {
+
+// TODO there will be people crying because this is not ideal. Like calling
+//  notify, while holding the mutex. I don't care right now and probably not
+//  in the future. This is used in IO heavy workloads, where the latency impact
+//  can't even be measured.
```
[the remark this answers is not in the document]
maierlars: clearly states that calling notify while holding the mutex is not ideal. Also you probably have read the PR description of this draft, which clearly states that the bounded channel is to be revisisted.

lib/Basics/BoundedChannel.h
```diff
@@ -0,0 +1,143 @@
...
+template<typename T>
+struct BoundedChannel {
```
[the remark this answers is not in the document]
maierlars: where else should I put it?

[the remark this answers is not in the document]
markuspf: I think `lib/Basics` is a far better place for such abstractions than it being hidden inside the arangodump code.

There could surely be a conversation whether `lib/Basics` is too much of a kitchen sink and should be better categorised, but that conversation should not happen on this pull request.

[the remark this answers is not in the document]
maierlars: Yeah but it's also used in the arangod part, so ...?

arangod/RocksDBEngine/RocksDBDumpContext.h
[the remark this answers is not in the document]
MBkkt: Same could be done for string_view. Just copy construction

client-tools/Dump/DumpFeature.h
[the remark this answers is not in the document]
MBkkt: So instead of option, with same difficulty, smaller allocation count and memory footprint, faster search and insert.
You choose this, because your code spend more time in the IO.

Do I understand you right?

lib/Basics/BoundedChannel.h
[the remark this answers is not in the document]
MBkkt: Maybe just lib/Dump?

Just separate common directory for all code that you or anyone else need to share between arangod and client-tools part of dump

client-tools/Dump/DumpFeature.h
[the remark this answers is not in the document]
maierlars: Sadly you don't.

lib/Basics/BoundedChannel.h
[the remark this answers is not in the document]
maierlars: This is not a concern to me right now. You can do it if you want to spend your time on this.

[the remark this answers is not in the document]
maierlars: And when I want to use it in arangorestore as well, I move it from the `lib/Dump` folder to then `lib/DumpAndRestore` folder?

client-tools/Dump/DumpFeature.h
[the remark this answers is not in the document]
MBkkt: Ok, explain please

lib/Basics/BoundedChannel.h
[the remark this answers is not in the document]
MBkkt: Ok

arangod/RocksDBEngine/RocksDBDumpManager.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+using namespace arangodb;
+
+RocksDBDumpManager::RocksDBDumpManager(RocksDBEngine& engine)
+    : _engine(engine) {}
+
+RocksDBDumpManager::~RocksDBDumpManager() {
+  std::lock_guard mutexLocker{_lock};
+  _contexts.clear();
+}
```
mpoeter, line 45: Using locks in the destructor is a code smell. In almost all cases they are either unnecessary or they actually don't protect you from a race after all. In this case the lock is unnecessary.

arangod/RocksDBEngine/RocksDBDumpManager.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // as a collection of cursors for shard dumping.
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_CURSOR_NOT_FOUND,
+                                   "requested dump context not found");
+  }
+
+  auto& context = (*it).second;
+  if (!context->canAccess(database, user)) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_FORBIDDEN,
+                                   "insufficient permissions");
+  }
```
mpoeter, line 125: This is a duplication of what we have in `find`. Can we factor this out into a separate function?

arangod/RocksDBEngine/RocksDBDumpManager.cpp
```diff
@@ -0,0 +1,175 @@
...
+  for (auto it = _contexts.begin(); it != _contexts.end();
+       /* no hoisting */) {
+    auto& context = it->second;
+
+    if (force || context->expires() < now) {
+      it = _contexts.erase(it);
+    } else {
+      ++it;
+    }
+  }
```
mpoeter, line 163: This looks like a perfect use case for `std::remove_if`.
```suggestion
  if (force) {
    _contexts.clear();
    return;
  }
  std::remove_if(_contexts.begin(), _contexts.end(), [now](auto& v) { return v.second->expires() < now; }):
  }
```

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,269 @@
...
+        (len == 2 && suffixes[0] != "next") || len < 1 || len > 2) {
+      return ResultT<std::pair<std::string, bool>>::error(
+          TRI_ERROR_BAD_PARAMETER,
+          "expecting POST /_api/dump/start or /_api/dump/next/<id>");
+    }
+  } else {
+    // invalid HTTP method
+    return ResultT<std::pair<std::string, bool>>::error(
+        TRI_ERROR_HTTP_METHOD_NOT_ALLOWED);
+  }
```
mpoeter, line 132: This contains quite some duplication with `execute()`. Can we somehow avoid this?

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  return {std::make_pair(StaticStrings::Empty, false)};
+}
+
+void RestDumpHandler::handleCommandDumpStart() {
+  std::string database = _request->databaseName();
+  std::string user = _request->user();
+
+  // TODO: check permissions
```
mpoeter, line 152: I guess this TODO still needs to be addressed.

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,269 @@
...
+      return s.getNumber<uint64_t>();
+    }
+    return ::defaultParallelism;
+  }();
+  opts.ttl = [&body]() {
+    if (auto s = body.get("ttl"); s.isNumber()) {
+      return s.getNumber<double>();
+    }
+    return ::defaultTtl;
+  }();
```
mpoeter, line 189: What if any of those parameters is present but not a number? It looks like we will then silently ignore it. Shouldn't we reject the request with an error instead?

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return;
+  } else {
+    for (auto it : VPackArrayIterator(s)) {
+      opts.shards.emplace_back(it.copyString());
+    }
+  }
+
+  auto& engine =
+      server().getFeature<EngineSelectorFeature>().engine<RocksDBEngine>();
+  auto* manager = engine.dumpManager();
```
mpoeter, line 203: Not a fan of this whole `getFeature` stuff. Can we at least move this to the constructor and store the manager as a member, pretending that all dependencies have been passed in (constructor injection) instead of fetching them dynamically?

arangod/RocksDBEngine/RocksDBDumpContext.h
```diff
@@ -0,0 +1,238 @@
...
+    std::vector<WorkItem> _work;
+    bool _completed{false};
+    size_t _waitingWorkers{0};
+    std::size_t const _numWorkers;
+    Result _result;
+  };
+
+  // Returns the next batch and assigned it batchId. If lastBatch is not nullopt
+  // frees the batch with the given id. This function might block, if no batch
+  // is available. It returns nullptr is there is no batch left.
```
mpoeter, line 162:
```suggestion
  // is available. It returns nullptr if there is no batch left.
```

arangod/RocksDBEngine/RocksDBDumpManager.cpp
maierlars: Yes, I agree!

arangod/RocksDBEngine/RocksDBDumpContext.h
```diff
@@ -0,0 +1,238 @@
...
+  // when it is returned to the manager.
+  // timestamp is in seconds since 1970/1/1
+  std::atomic<double> _expires;
+
+  // a guard object that protects the underlying database from being deleted
+  // while the dump is ongoing. will be populated in the constructor and then
+  // be static.
+  std::unique_ptr<DatabaseGuard> _databaseGuard;
+
+  // collection access objects that project the underlying collections/shards
```
mpoeter, line 195:
```suggestion
  // collection access objects that protect the underlying collections/shards
```

arangod/RocksDBEngine/RocksDBDumpGuard.cpp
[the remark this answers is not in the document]
maierlars: pretty sure you ment `RocksDBDumpContextGuard`?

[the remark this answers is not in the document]
maierlars: wait, what is this file doing in here anyways?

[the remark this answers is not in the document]
maierlars: It is completely unused.

arangod/RestHandler/RestDumpHandler.cpp
maierlars: Oh wow, I wasn't aware of that. My plan was to add a permissions test as well. Would have found out later 😆

maierlars: Permissions + Test added.

tests/js/client/shell/api/dump-cluster.js
```diff
@@ -0,0 +1,210 @@
...
+// /
+// / Unless required by applicable law or agreed to in writing, software
+// / distributed under the License is distributed on an "AS IS" BASIS,
+// / WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// / See the License for the specific language governing permissions and
+// / limitations under the License.
+// /
+// / Copyright holder is ArangoDB GmbH, Cologne, Germany
+// /
+// / @author
```
neunhoef, line 26: There should be an author here.

tests/js/client/authentication/dump-api.js
```diff
@@ -0,0 +1,135 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is triAGENS GmbH, Cologne, Germany
+///
+/// @author Lars Maier
+/// @author Copyright 2013, triAGENS GmbH, Cologne, Germany
```
neunhoef, line 26: Copyright notice outdated.

tests/js/client/authentication/dump-api.js
```diff
@@ -0,0 +1,135 @@
+/*jshint globalstrict:false, strict:false */
+/*global assertTrue, assertEqual, assertNotUndefined */
+
+////////////////////////////////////////////////////////////////////////////////
+/// @brief test the authentication
+///
+/// DISCLAIMER
+///
+/// Copyright 2010-2012 triagens GmbH, Cologne, Germany
```
neunhoef, line 9: Copyright notice outdated.

client-tools/Dump/DumpFeature.cpp
```diff
@@ -1364,4 +1470,349 @@ void DumpFeature::start() {
...
+    std::unique_ptr<arangodb::httpclient::SimpleHttpResult> response(
+        client.request(arangodb::rest::RequestType::POST, url, nullptr, 0, {}));
+    auto check = ::arangodb::HttpResponseChecker::check(
+        client.getErrorMessage(), response.get());
+    if (check.fail()) {
+      LOG_TOPIC("ad972", ERR, arangodb::Logger::DUMP)
+          << "An error occurred while dumping from server '" << server
+          << "': " << check.errorMessage();
+
+      // TODO complete this list!
```
neunhoef, line 1646: How about the error that happens when a connection is broken? This sounds important to me in the cloud business...

neunhoef (approved):
This basically LGTM. I have a few minor things, which should be addressed, see comments. For the sake of quickness, I approve, but please look at the comments.

client-tools/Dump/DumpFeature.cpp
neunhoef: All others are trivial. This one should at least handle the connection broken case, in particular since it is simple to implement, because the retry logic is already there.

maierlars: Sadly the SimpleHttpClient has a very intransparent error handling. Many errors result in a `DEAD` connection, i.e. the response is `nullptr`. Connection refused, connection broken or DNS resolution failure. Fixing all these things would require more work.

client-tools/Dump/DumpFeature.cpp
```diff
@@ -1364,4 +1470,349 @@ void DumpFeature::start() {
...
+    builder.add("parallelism", VPackValue(options.dbserverWorkerThreads));
+    {
+      VPackArrayBuilder ab(&builder, "shards");
+      for (auto const& [shard, info] : shards) {
+        builder.add(VPackValue(shard));
+      }
+    }
+  }
+
+  auto bodystr = builder.toJson();
```
mpoeter, line 1489: Could also be simplified with an inspector.

mpoeter (changes requested):
Most of the comments are simplifications/deduplications. I would very much appreciate if those could be addressed to improve the overall code quality, but as we have agreed those can be considered optional.

The only remaining point that IMO definitely needs to be addressed is the lifetime management in `RocksDBDumpContextGuard`.

client-tools/Dump/DumpFeature.cpp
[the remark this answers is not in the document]
maierlars: All other classes are nested in there as well. This is how the code is organized in arangodump. Of course we can spend time on refactoring it.

arangod/RocksDBEngine/RocksDBDumpContextGuard.cpp
```diff
@@ -0,0 +1,55 @@
...
+    : _manager(manager), _ctx(std::move(ctx)) {
+  TRI_ASSERT(_ctx != nullptr);
+}
+
+RocksDBDumpContextGuard::RocksDBDumpContextGuard(
+    RocksDBDumpContextGuard&& other) noexcept
+    : _manager(other._manager), _ctx(std::move(other._ctx)) {
+  TRI_ASSERT(_ctx != nullptr);
+  _ctx->extendLifetime();
+}
```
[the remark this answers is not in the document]
maierlars: Not sure what the point of this guard is anyways. It's essentially just a wrapper of a `shared_ptr`. Maybe @jsteemann can comment on this?

client-tools/Dump/DumpFeature.cpp
neunhoef: OK, what a sad state of affairs!

neunhoef (approved):
LGTM. This is all OK and can be merged for now as is. However, we must address something later: As it is now, it is possible via the API to ask for arbitrary RAM usage on the dbservers. Think of a cost-aware customer on ArangoGraph, who orders a cluster with three dbservers which have 3 GB of RAM and 0.75 vCPUs. If that customer stores 64 GB of data on each dbserver and then asks for a parallel dump with a batch size of 64 MB and a prefetch of 256 batches, we immediately go OOM without any further protection.
That is, with this we have built yet another subsystem which can explode if used improperly. We should add memory accounting to this subsystem and give it a configurable global limit on each dbserver and then enforce it and if we run into the limit, handle it gracefully. It would be allowed for the dump to fail, but there must not be a crash or over-allocation of RAM.
But this is future work (probably for 3.12).

arangod/RocksDBEngine/RocksDBDumpManager.cpp
jsteemann: Addressed.

arangod/RocksDBEngine/RocksDBDumpManager.cpp
jsteemann: https://stackoverflow.com/a/9210110/3042070

arangod/RocksDBEngine/RocksDBDumpContextGuard.cpp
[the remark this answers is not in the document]
jsteemann: Originally I intended the guard to have a pointer to the manager, and call something like `Manager::returnLease(context)`, which would extend the lifetime of the context etc.
However, all of this turned out to be unnecessary, and we can get away with returning just a normal shared_ptr from `Manager::find()`. The `extendLifetime()` call is now moved to the call site as well, and there is no need for an extra guard object. I have also removed the corresponding code.

arangod/RestHandler/RestDumpHandler.cpp
jsteemann: I tried your suggestion in commit 0cd69ed8770d6845d35b10e722944cc9a6bc00fa, but it turned out that it didn't work.
It seems to be problematic to access the storage engine's `dumpManager()` function in the RestHandler's constructor, simply because the manager seemed to be a nullptr at that point.
This is probably related to some hot backup starting/restarting servers at will, but I don't know for sure.
The problem of the manager being a nullptr can be avoided by using the existing code, which only calls `dumpManager()` during execute().

arangod/RestHandler/RestDumpHandler.cpp
jsteemann: Pushed a change into the PR that should improve the situation, although I am still not very happy with it.

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,328 @@
...
+  } else {
+    for (auto it : VPackArrayIterator(s)) {
+      if (!it.isString()) {
+        THROW_ARANGO_EXCEPTION_MESSAGE(
+            TRI_ERROR_BAD_PARAMETER,
+            "invalid 'shards' value in request - expected array or strings");
+      }
+      opts.shards.emplace_back(it.copyString());
+    }
+  }
```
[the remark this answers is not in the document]
jsteemann: I added an Inspector to RocksDBDumpContextOptions as requested.

neunhoef (approved):
LGTM

arangod/RestHandler/RestDumpHandler.cpp
mpoeter: Ok, this is probably related to our lack of dependency management and the resulting initialization order. I would still prefer if we could at least move the getFeature call into the constructor and keep the reference to the engine as member and only defer the `dumpManager` call.

arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,288 @@
...
+  _response->setHeaderNC(StaticStrings::DumpBlockCounts,
+                         std::to_string(counts));
+  _response->setContentType(rest::ContentType::DUMP);
+  _response->addRawPayload(batch->content);
+  _response->setGenerateBody(true);
+  _response->setResponseCode(rest::ResponseCode::OK);
+
+  // prolong lifetime of context, so that it is still there for follow-up
+  // requests.
+  context->extendLifetime();
```
mpoeter, line 206: We already have a call to `extendLifetime()` in line 185. Is this intentional that it is called again?

markuspf (approved):
LGTM

arangod/RocksDBEngine/RocksDBDumpManager.cpp
mpoeter: Good point! But with C++20 we now have [`erase_if`](https://en.cppreference.com/w/cpp/container/unordered_map/erase_if) which is even nicer since it takes the container instead of the iterators.

client-tools/Dump/DumpFeature.cpp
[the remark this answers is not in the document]
mpoeter: I would appreciate that, but I would understand if you don't want to spend time on it.

arangod/RocksDBEngine/RocksDBDumpContext.h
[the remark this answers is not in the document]
jsteemann: I think the current approach of returning an std::string by reference is fine. The caller can get it by reference or by value, depending on needs.
Returning an std::string_view would also work here, but would IMO not provide any benefit, but would require the already existing code.

arangod/RestHandler/RestDumpHandler.cpp
jsteemann: It actually was intentional, but @maierlars seems to have removed it now...
The rationale was to extend the lifetime as soon as the context is leased from the registry, so it is not garbage-collected while it is in use.
And then to extend its lifetime once again when it is "returned" to the registry, because the time a context is used can be pretty long (in case of slow IO or server overload).

maierlars: yeah, I was thinking about that. I was unsure if _that_ was the reason or if it was just an oversight. Feel free to add it again.

neunhoef (approved):
LGTM

dothebart:
please make sure that this new behaviour works well once I/O saturation appears: https://www.arangodb.com/2019/01/using-the-linux-kernel-and-cgroups-to-simulate-starvation/ can help simulate this.

maierlars:
> please make sure that this new behavior works well once I/O saturation appears: https://www.arangodb.com/2019/01/using-the-linux-kernel-and-cgroups-to-simulate-starvation/ can help simulate this.

This has already been tested in the real world. We used different machines and hit multiple IO bottlenecks. One time we hit the disk io limit, another time it was the network.

arangod/RocksDBEngine/RocksDBDumpManager.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+void RocksDBDumpManager::garbageCollect(bool force) {
+  auto const now = TRI_microtime();
+
+  std::lock_guard mutexLocker{_lock};
+
+  std::erase_if(_contexts, [&](auto const& x) {
+    return force || x.second->expires() < now;
+  });
```
mpoeter, line 122: I find this easier to read, but feel free to keep it the way it is
```suggestion
  std::lock_guard mutexLocker{_lock};

  if (force) {
    _contexts.clear();
    return;
  }
  auto const now = TRI_microtime();
  std::erase_if(_contexts, [&](auto const& x) { return x.second->expires() < now; });
```

mpoeter (approved):

PR #19209, opened 2023-06-07
Update replication2

apetenchea:
### Scope & Purpose

This PR updates replication2 in devel.

Enterprise counterpart: https://github.com/arangodb/enterprise/pull/1292

- [x] :pizza: New feature


lib/Assertions/ProdAssert.h
```diff
@@ -38,6 +38,7 @@
 #if defined(ARANGODB_ENABLE_MAINTAINER_MODE)
 
 #include "AssertionConditionalLogger.h"
+#include "AssertionLogger.h"
```
MBkkt, line 41: Looks unnecessary

lib/V8/v8-globals.h
```diff
@@ -539,9 +539,6 @@ struct TRI_v8_global_t {
   /// @brief replicated log template
   v8::Persistent<v8::ObjectTemplate> VocbaseReplicatedLogTempl;
```
MBkkt, line 540: Is it used?

lib/Inspection/include/Inspection/VPack.h
```diff
@@ -45,7 +45,26 @@ void serialize(Builder& builder, T& value) {
...
+}
+
+template<class T, class Context>
+void serializeWithContext(Builder& builder, T& value, Context const& context) {
+  inspection::VPackSaveInspector<Context> inspector(builder, context);
+  if (auto res = inspector.apply(value); !res.ok()) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(
+        TRI_ERROR_INTERNAL,
+        std::string{"Error while serializing to VelocyPack: "} + res.error() +
+            "\nPath: " + res.path());
```
MBkkt, line 59: StrCat, concatT

arangod/Replication2/coro-helper.h
MBkkt, line 1: Do we allow using coroutine in the arangodb codebase?
They currently have a lot of unresolved issues.
Like
* https://github.com/llvm/llvm-project/issues/62842
* https://discourse.llvm.org/t/address-thread-identification-problems-with-coroutine/62015
* no one except clang doesn't correctly support symmetric transfer (which is important to avoid stack overflow)
* msvc has hidden crash in case of symmetric transfer in the final_suspend awaiter
* ...

So I think it should be publicly discussed and shouldn't be a part of this PR.

MBkkt, line 1: Also such implementation makes unnecessary allocation:
Future for coroutine itself and Future for awaiter for every coroutine.
And this spends two times more memory:
T in the Future and T in the optional.
And it doesn't use symmetric transfer to avoid stack overflow and less stack usage, if symmetric transfer is possible

This could be improved but it's not a single line suggestion, so I happy to discuss it separately.

MBkkt (changes requested):

lib/Assertions/Assert.h
```diff
@@ -36,13 +36,17 @@
...
-      : ::arangodb::debug::AssertionLogger{__FILE__, __LINE__,                \
-                                           ARANGODB_PRETTY_FUNCTION, #expr} & \
+#define TRI_ASSERT(expr) /*GCOVR_EXCL_LINE*/                                   \
+  (ADB_LIKELY(expr))                                                           \
+      ? (void)nullptr                                                          \
+      : ::arangodb::debug::AssertionLogger{ADB_HERE, ARANGODB_PRETTY_FUNCTION, \
+                                           #expr} &                            \
             ::arangodb::debug::AssertionLogger::assertionStringStream
 
+#define ADB_MAINTAINER_CRASH()                                             \
```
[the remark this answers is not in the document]
mchacki: Not used anymore will remove thanks for finding.

lib/V8/v8-globals.h
mchacki: Yes in our tests, we expose access to those logs through V8.

mchacki: In : arangod/V8Server/v8-replicated-logs.cpp

MBkkt: Thanks, I just tried to find it in devel, but it's probably new usage

lib/Assertions/ProdAssert.h
mchacki: Yes it was, removed now

lib/Inspection/include/Inspection/VPackSaveInspector.h
```diff
@@ -42,7 +42,7 @@ namespace arangodb::inspection {
 template<class Context = NoContext>
 struct VPackSaveInspector
     : SaveInspectorBase<VPackSaveInspector<Context>, Context> {
- protected:
+ public:
   using Base = SaveInspectorBase<VPackSaveInspector, Context>;
 
  public:
```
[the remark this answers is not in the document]
mchacki: Yes removed

lib/Inspection/include/Inspection/VPack.h
mchacki: replaced with your suggestion

PR #19213, opened 2023-06-08
Update replication2 staging

apetenchea:
### Scope & Purpose

The purpose of this PR is to bring the staging/replication-2.0 branch in sync with the latest changes from devel.

Additionally, the `testDistributeShardsLikeOnIdenticalServers_R2` test from the `shard-distribution.js` suite has been detected as flaky. The JS-only fix is provided together with this PR. In case of replication2 databases, we might have to wait for all shards to be visible in `Current`.

Do not merge this PR into staging directly from GitHub, as we try to avoid squash commits.

Enterprise counterpart: https://github.com/arangodb/enterprise/pull/1293

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM

PR #19224, opened 2023-06-12
[3.10.6] Fast ArangoDump

maierlars:
### Scope & Purpose

Cherry-picked version for 3.10.6.


arangod/RestHandler/RestDumpHandler.cpp
```diff
@@ -0,0 +1,328 @@
...
+      bool parseSuccess = false;
+      VPackSlice body = this->parseVPackBody(parseSuccess);
+      if (!parseSuccess) {  // error message generated in parseVPackBody
+        return {TRI_ERROR_BAD_PARAMETER};
+      }
+
+      ExecContext& exec =
+          *static_cast<ExecContext*>(_request->requestContext());
+      LOG_DEVEL << "exec user = " << exec.user()
+                << " req user = " << _request->user();
```
MBkkt, line 141: Should it be LOG_DEVEL?

MBkkt (changes requested):

neunhoef (approved):
LGTM. This is all OK and can be merged for now as is. However, we must address something later: As it is now, it is possible via the API to ask for arbitrary RAM usage on the dbservers. Think of a cost-aware customer on ArangoGraph, who orders a cluster with three dbservers which have 3 GB of RAM and 0.75 vCPUs. If that customer stores 64 GB of data on each dbserver and then asks for a parallel dump with a batch size of 64 MB and a prefetch of 256 batches, we immediately go OOM without any further protection.
That is, with this we have built yet another subsystem which can explode if used improperly. We should add memory accounting to this subsystem and give it a configurable global limit on each dbserver and then enforce it and if we run into the limit, handle it gracefully. It would be allowed for the dump to fail, but there must not be a crash or over-allocation of RAM.
But this is future work (probably for 3.12).

MBkkt (commented):
Try to remove changes requested

PR #19227, opened 2023-06-13
Implement a crude bash completion for ./scripts/unittest

goedderz:
### Scope & Purpose

Add a script that generates a crude bash completion for `./scripts/unittest`. Minimal support is added to `./scripts/unittest` (aka `testing.js`) itself.


jsteemann (approved):
LGTM

PR #19232, opened 2023-06-13
Bug fix 3.11/telemetrics force string persisted

cpjulia:
### Scope & Purpose

Backport to https://github.com/arangodb/arangodb/pull/19233.
This PR forces the value put in the velocypack object respective to persisted id to be a string when the server has no persisted id value and its id is used instead to prevent this error in the endpoint's parser:
```
Saving messages in BiqQuery failed: '400 Provided Schema does not match Table telemetrics-project:telemetrics_dataset.telemetrics_table. Field deployment.persisted_id has changed type from STRING to INTEGER``
```


jwierzbo (approved):

jsteemann (approved):
LGTM

naushniki (approved):

PR #19237, opened 2023-06-14
Allow to specify a factory for non default-constructible types

goedderz:
### Scope & Purpose

Allow to inspect, and in particular deserialize, types that aren't default constructible. A static factory can be specialized to allow the Inspection library to create values.

- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [X] C++ **Unit tests**

### DIFF
--- a/lib/Inspection/include/Inspection/Factory.h
+++ b/lib/Inspection/include/Inspection/Factory.h
@@ -50,7 +50,7 @@ struct Factory {
   }
 };
 
-template<typename T, typename Derived>
+template<typename T, typename Derived = Factory<T>>
 struct BaseFactory {
   static auto make_unique() -> std::unique_ptr<T> {
     // If you get a compile error here because T has deleted copy and move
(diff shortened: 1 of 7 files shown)


lib/Inspection/include/Inspection/Factory.h
```diff
@@ -0,0 +1,67 @@
...
+  }
+  static auto make_shared() -> std::shared_ptr<T> {
+    // If you get a compile error here because your T is not default
+    // constructible, see comment above.
+    return std::make_shared<T>();
+  }
+};
+
+template<typename T, typename Derived>
+struct BaseFactory {
```
mpoeter, line 54: Usually any class deriving from this should be a specialization of `Factory`, so we could even default this so that the base class can be simplified.
```suggestion
template<typename T, typename Derived = Factory<T>>
struct BaseFactory {
```

goedderz: Good idea!

PR #19249, opened 2023-06-15
[CINFRA-781] Have leaders track their participants' RebootIds

goedderz:
### Scope & Purpose

https://arangodb.atlassian.net/browse/CINFRA-781

The leader has to track its participant’s RebootIds and report them to Current.

In order to do that, on startup it has to:

* copy the map participantId => rebootId from the agency cache (before establishing leadership)
* report the map to current after leadership has been established

Later, if any participant’s RebootId changes (register callback in the RebootTracker):

* get the RebootId of the participant from the agency cache
* insert an empty log entry
* after the log entry has been committed, report the map to current

- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [X] **integration tests**

### DIFF
--- a/arangod/Cluster/Maintenance.cpp
+++ b/arangod/Cluster/Maintenance.cpp
@@ -1536,6 +1536,59 @@ static void writeUpdateReplicatedLogLeader(
   }
 }
 
+static auto reportCurrentReplicatedLogSafeRebootIds(
+    replication2::replicated_log::QuickLogStatus const& status,
+    std::optional<replication2::agency::LogCurrent> const& current)
+    -> std::optional<
+        std::unordered_map<replication2::ParticipantId, RebootId>> {
+  auto statusSafeRebootIds = status.safeRebootIds != nullptr
+                                 ? std::optional{*status.safeRebootIds}
+                                 : std::nullopt;
+  if (!current.has_value() or current->safeRebootIds != statusSafeRebootIds) {
+    // Note that this may still be std::nullopt, which means no update.
+    return statusSafeRebootIds;
+  }
+  return std::nullopt;
+}
+
+static void writeUpdateReplicatedLogSafeRebootIds(
+    VPackBuilder& report, replication2::LogId id, DatabaseID const& dbName,
+    replication2::LogTerm& localTerm,
+    std::unordered_map<replication2::ParticipantId, RebootId> const&
+        safeRebootIds) {
+  // update Current/ReplicatedLogs/<dbname>/<logId>/safeRebootIds
+  // with precondition
+  // Plan/ReplicatedLogs/<dbname>/<logId>/term/term == currentTerm
+  using namespace cluster::paths;
+  auto reportPath = aliases::current()
+                        ->replicatedLogs()
+                        ->database(dbName)
+                        ->log(to_string(id))
+                        ->safeRebootIds()
+                        ->str(SkipComponents(
+                            1) /* skip first path component, i.e. 'arango' */);
+  auto preconditionPath =
+      aliases::plan()
+          ->replicatedLogs()
+          ->database(dbName)
+          ->log(to_string(id))
+          ->currentTerm()
+          ->term()
+          ->str(
+              SkipComponents(1) /* skip first path component, i.e. 'arango' */);
+  report.add(VPackValue(reportPath));
+  {
+    VPackObjectBuilder o(&report);
+    report.add(OP, VP_SET);
+    report.add(VPackValue("payload"));
+    velocypack::serialize(report, safeRebootIds);
+    {
+      VPackObjectBuilder preconditionBuilder(&report, "precondition");
+      report.add(preconditionPath, VPackValue(localTerm));
+    }
+  }
+}
+
 static void writeUpdateReplicatedLogLocal(
     VPackBuilder& report, replication2::LogId id, DatabaseID const& dbName,
     ServerID const& serverId, replication2::LogTerm localTerm,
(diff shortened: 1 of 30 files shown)


arangod/Cluster/Maintenance.cpp
```diff
@@ -1536,6 +1536,59 @@ static void writeUpdateReplicatedLogLeader(
...
+}
+
+static void writeUpdateReplicatedLogSafeRebootIds(
+    VPackBuilder& report, replication2::LogId id, DatabaseID const& dbName,
+    replication2::LogTerm& localTerm,
+    std::unordered_map<replication2::ParticipantId, RebootId> const&
+        safeRebootIds) {
+  // update Current/ReplicatedLogs/<dbname>/<logId>/safeRebootIds
+  // with precondition
+  // Plan/ReplicatedLogs/<dbname>/<logId>/term/term == currentTerm
```
mchacki, line 1561:
```suggestion
  // Plan/ReplicatedLogs/<dbname>/<logId>/term/ == currentTerm
```

mchacki: or put `<term>` at the end.
I got a bit confused by the double term at the end.

arangod/Replication2/ReplicatedLog/DefaultRebootIdCache.cpp
```diff
@@ -0,0 +1,78 @@
...
+    auto serverStateIt = rebootIds.find(participant);
+    if (serverStateIt != rebootIds.end()) {
+      result.try_emplace(participant, serverStateIt->second.rebootId);
+    } else {
+      // We need to report a RebootId for each participant. Reporting 0 is
+      // always safe, as it is the most pessimistic assumption.
+      result.try_emplace(participant, RebootId(0));
+      // All participants should always be available in the lists of known
+      // servers (I think).
+      TRI_ASSERT(false) << fmt::format(
```
mchacki, line 61: For our internal tests I agree. So keep this in!
Is this assumption still true after a full reboot of the cluster? Or after a doom event?

I am just wondering if we want to take out a server from the cluster it needs to be erased from knownServers list eventually, is it guaranteed that we cleared up all traces of it in participants?

Also nothing to fix in this PR, just want to make sure we do not forget this case.

arangod/Replication2/ReplicatedLog/LogLeader.h
```diff
@@ -301,19 +326,20 @@ class LogLeader : public std::enable_shared_from_this<LogLeader>,
...
     // active - that is currently used to check for committed entries
-    std::shared_ptr<agency::ParticipantsConfig const> activeParticipantsConfig;
-    LogIndex activeParticipantsConfigLogIndex;
+    std::shared_ptr<InnerTermConfig const> activeInnerTermConfig;
+    LogIndex activeInnerConfigLogIndex;
     // committed - latest active config that has committed at least one entry
     // Note that this will be nullptr until leadership is established!
-    std::shared_ptr<agency::ParticipantsConfig const>
-        committedParticipantsConfig;
+    // std::shared_ptr<InnerTermConfig const> committedInnerTermConfig;
```
mchacki, line 333:
```suggestion
```

mchacki: Same line without comment below

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -63,12 +63,14 @@
...
 #include "Replication2/ReplicatedLog/Algorithms.h"
 #include "Replication2/ReplicatedLog/InMemoryLog.h"
+#include "Replication2/ReplicatedLog/IRebootIdCache.h"
 #include "Replication2/ReplicatedLog/LogStatus.h"
 #include "Replication2/ReplicatedLog/ReplicatedLogIterator.h"
 #include "Replication2/ReplicatedLog/ReplicatedLogMetrics.h"
 #include "Replication2/IScheduler.h"
 #include "Replication2/ReplicatedLog/Components/StorageManager.h"
 #include "Replication2/ReplicatedLog/Components/CompactionManager.h"
+#include "Replication2/ReplicatedLog/IRebootIdCache.h"
```
mchacki, line 73:
```suggestion
```
Double includes are always double as secure :grin:

mchacki (approved):
LGTM.

I have some minor things in the comments, nothing serious

arangod/Cluster/Maintenance.cpp
goedderz, line 1561: Thanks for noticing!
```suggestion
  // Plan/ReplicatedLogs/<dbname>/<logId>/currentTerm/term == localTerm
```

goedderz: https://github.com/arangodb/arangodb/pull/19249/files/92745149d7bf035a59d5a2e41f4ed4151063c11d#r1236726153

arangod/Replication2/ReplicatedLog/DefaultRebootIdCache.cpp
goedderz: > Is this assumption still true after a full reboot of the cluster? Or after a doom event?

My assumptions are:
* a server is always added to the known servers before it has a chance to be added to a log (in Target/).
* a server that's removed from the cluster is first freed from all responsibilities (like being a participant in a log), and only then removed from the list of known servers.

It may well be possible that the latter part isn't true for replication 2.0 (yet), but that just means we have to fix `/_admin/cluster/removeServer`. I can open a ticket that we make sure this works and add some tests.

goedderz: https://arangodb.atlassian.net/browse/CINFRA-788

goedderz:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/30554/

mchacki (approved):

PR #19273, opened 2023-06-20
Add grace period for disconnected leaders in active failover

jsteemann:
### Scope & Purpose

Potential fix for https://arangodb.atlassian.net/browse/BTS-1488

Backport of https://github.com/arangodb/arangodb/pull/19339 

* Added startup option `--replication.active-failover-leader-grace-period`. This startup option can be used to set the amount of time (in seconds) for which the current leader in an active failover setup will continue to assume its leadership even if it lost connection to the agency.

  In case the leader cannot contact the agency anymore, the agency will elect a new leader after the supervision grace period has elapsed. In order to avoid a split-brain situation with multiple servers assuming leadership, this option can be used to make a disconnected leader refuse any incoming write operations after the grace period controled by this option has elapsed. Ideally the startup option should be given a value greater than the value of the supervision grace period, in order to avoid a temporarily disconnected leader giving up leadership too early and unnecessarily.

  The default value is 120 seconds. Setting the option to a value of 0 will keep the existing behavior, in which a disconnected leader will not refuse incoming write operations.

  The rationale of not reusing the `--agency.supervision-grace-period` startup option here is that by using a separate option, users can opt in or out of the behavior change. Using the agency option also could be risky because it may not be set on the other servers in an active failover deployment.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19337
  - [x] Backport for 3.9: https://github.com/arangodb/arangodb/pull/19340

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1488
- [ ] Design document:


neunhoef (approved):
LGTM

CryptoNinjaGeek:
Green Jenkins: https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/30674/

maierlars (approved):

PR #19285, opened 2023-06-22
[CINFRA-776] Add syncIndex to participants

apetenchea:
### Scope & Purpose

This PR adds `syncIndex` to log participants, populating it as described in the [design document](https://arangodb.atlassian.net/wiki/spaces/CInfra/pages/2061369370/Concept+Conservative+leader+election#Sync-Index-implementation).

Briefly, we need the `syncIndex` in order to prevent any entries that have not been yet synced to disk from being compacted.
After we write a batch of entries to the WAL, we get back a sequence number. Having this number, we can correlate it to the highest log index from the written batch (the `lastIndex`). We'll know that the `lastIndex` (along with all the other entries that go before it) has been synced to disk when the given sequence number has been synced. Therefore, we register a listener on the `RocksDBSyncThread`, being updated whenever the `lastSequenceNumber` is updated.
Note that `onSync` does everything in a newly scheduled task, so the syncer thread is not blocked. The code around this could be further optimized, but for now there is only one use-case for such listeners. As we try to move to a new WAL implementation, my focus was to keep RocksDB related changes to a minimum.

Additional unit tests have been added for the `StorageManager`.

[--rocksdbsync-interval](https://www.arangodb.com/docs/stable/programs-arangod-options.html#--rocksdbsync-interval)
This represents the interval for automatic, non-requested disk syncs (in milliseconds, 0 = turn automatic syncing off). Turning it off means there's no longer a dedicated thread for that, thus no `RocksDBSyncThread`. Other parts of the project rely on calling the `static RocksDBSyncThread::sync(rocksdb::DB* db)` when needed, but due to the static nature of this method, it does not play well with our current `syncIndex` implementation. We issue a warning about this in `RocksDBEngine.cpp`. As we move to a new WAL implementation, the problem will get fixed.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**


arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -97,10 +97,58 @@ struct comp::StorageManagerTransaction : IStorageTransaction {
...
+            // We're done writing to the WAL. That doesn't necessarily
+            // mean the data is synced to disk.
+            if (auto mngr = weakManager.lock(); res.ok() && mngr) {
+              // Methods are available as long as the manager is not
+              // destroyed.
+              mngr->scheduler->queue(
+                  [lastIndex, weakManager = std::move(weakManager),
+                   wfs = methods.waitForSync(res.get())]() mutable {
+                    auto syncRes = wfs.get();
+                    if (auto mngr = weakManager.lock()) {
```
maierlars, line 133: What's this stunt supposed to do? We create a future with `waitForSync` and then put in a lambda, post that on the scheduler and then synchronously wait for the future in a new thread. Is there a problem with `thenValue`?

arangod/Replication2/ReplicatedLog/Components/IStorageManager.h
```diff
@@ -57,7 +57,7 @@ struct IStorageTransaction {
       -> futures::Future<Result> = 0;
   virtual auto removeBack(LogIndex start) noexcept
       -> futures::Future<Result> = 0;
-  virtual auto appendEntries(InMemoryLog) noexcept
+  virtual auto appendEntries(InMemoryLog, bool waitForSync) noexcept
```
maierlars, line 60: I'm totally fine with having the `bool` parameter in here. But arangodb code tends to have 4 or 5 of these parameters on functions to make them _just behave a little different_.

This is why I added the `WriteOptions` to the `IStorageEngineMethods` interface. This allows to write code that looks similar to python code with key word arguments. In particular you don't have to specify the first 4 if you want to change the default value of the fifth one.

arangod/RocksDBEngine/RocksDBSyncThread.cpp
```diff
@@ -184,3 +201,10 @@ void RocksDBSyncThread::run() {
     }
   }
 }
+
+void RocksDBSyncThread::notifySyncListeners(
+    rocksdb::SequenceNumber seq) noexcept {
+  for (auto& listener : _syncListeners) {
+    listener->onSync(seq);
+  }
```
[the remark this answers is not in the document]
apetenchea: I know it may look like bad design - it leaves the door open for starving the SyncThread.
I wanted to go with the simplest approach here, as my goal is to keep RocksDB changes to a minimum. I expect these changes will no longer be needed (and reverted) once we move to a new WAL.
Having the SyncThread notify listeners in a scheduled task would require it to have access to the SCHEDULER. That's not necessarily something bad, but unless there are any other use-cases for it, I would rather avoid it.

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
apetenchea: The new version is using `thenFinal`.

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -97,10 +97,58 @@ struct comp::StorageManagerTransaction : IStorageTransaction {
...
+                      mngr->syncIndex.doUnderLock(
+                          [lastIndex](auto&& syncIndex) {
+                            if (syncIndex < lastIndex) {
+                              syncIndex = lastIndex;
+                            }
+                          });
+                    }
+                  });
+            }
+            return res.result();
```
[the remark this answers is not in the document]
apetenchea: Moved it to `updateSyncIndex`.

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -97,10 +97,58 @@ struct comp::StorageManagerTransaction : IStorageTransaction {
...
+              methods.insert(std::move(iter), {.waitForSync = waitForSync});
+
+          if (waitForSync) {
+            return std::move(fut).thenValue(
+                [lastIndex, weakManager = std::move(weakManager)](auto&& res) {
+                  if (auto mngr = weakManager.lock(); res.ok() && mngr) {
+                    mngr->syncIndex.doUnderLock([lastIndex](auto&& syncIndex) {
+                      if (syncIndex < lastIndex) {
+                        syncIndex = lastIndex;
+                      }
```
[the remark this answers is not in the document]
apetenchea: Moved to `updateSyncIndex`.

arangod/Replication2/ReplicatedLog/Components/IStorageManager.h
apetenchea: I don't have a strong preference on this, but I'm happy to make it look more similar to Python :)

maierlars (approved):

mchacki (approved):
LGTM

PR #19292, opened 2023-06-23
BTS-1495 Inverted index cause incorrect plan optimization

MBkkt:
### Scope & Purpose

The aql generate plan without all filter variables, so inverted index incorrectly consider it handles all filter conditions in such case

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Condition.cpp
+++ b/arangod/Aql/Condition.cpp
@@ -1093,7 +1093,8 @@ AstNode* Condition::removeCondition(ExecutionPlan const* plan,
 }
 
 /// @brief remove (now) invalid variables from the condition
-bool Condition::removeInvalidVariables(VarSet const& validVars) {
+bool Condition::removeInvalidVariables(VarSet const& validVars,
+                                       bool& isAllCoveredByIndex) {
   if (_root == nullptr) {
     return false;
   }
@@ -1134,6 +1135,7 @@ bool Condition::removeInvalidVariables(VarSet const& validVars) {
       }
 
       if (invalid) {
+        isAllCoveredByIndex = false;
         andNode->removeMemberUncheckedUnordered(j);
         // repeat with some member index
         TRI_ASSERT(nAnd > 0);
--- a/arangod/Aql/Condition.cpp
+++ b/arangod/Aql/Condition.cpp
@@ -1094,7 +1094,7 @@ AstNode* Condition::removeCondition(ExecutionPlan const* plan,
 
 /// @brief remove (now) invalid variables from the condition
 bool Condition::removeInvalidVariables(VarSet const& validVars,
-                                       bool& isAllCoveredByIndex) {
+                                       bool& noRemoves) {
   if (_root == nullptr) {
     return false;
   }
@@ -1135,7 +1135,7 @@ bool Condition::removeInvalidVariables(VarSet const& validVars,
       }
 
       if (invalid) {
-        isAllCoveredByIndex = false;
+        noRemoves = false;
         andNode->removeMemberUncheckedUnordered(j);
         // repeat with some member index
         TRI_ASSERT(nAnd > 0);
--- a/arangod/Aql/Condition.h
+++ b/arangod/Aql/Condition.h
@@ -166,7 +166,7 @@ class Condition {
                                     AstNode*, bool isPathCondition);
 
   /// @brief remove (now) invalid variables from the condition
-  bool removeInvalidVariables(VarSet const&);
+  bool removeInvalidVariables(VarSet const&, bool& isAllCoveredByIndex);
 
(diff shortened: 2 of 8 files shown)


markuspf (changes requested):
Two obvious issues:
 * I do not understand the problem description in the PR,
 * Can you add a regression test that shows the wrong behaviour which is made green by this patch?

MBkkt:
Yes, test will be added

MBkkt:
I don't understand what you don't understand in the issue description, I think it's pretty clear

Dronplane:
@markuspf  as stated in PR title it is related to https://arangodb.atlassian.net/browse/BTS-1495 
Ticket is quite verbose (which is good) so PR does not contain full copy from it, just a short summary of a cause.

arangod/Aql/Condition.h
```diff
@@ -166,7 +166,7 @@ class Condition {
                                     AstNode*, bool isPathCondition);
 
   /// @brief remove (now) invalid variables from the condition
-  bool removeInvalidVariables(VarSet const&);
+  bool removeInvalidVariables(VarSet const&, bool& isAllCoveredByIndex);
```
Dronplane, line 169:
```suggestion
  bool removeInvalidVariables(VarSet const&, bool& wasRemoves);
```
That name describes the meaning much better. As there is no index to cover something in  Condition

MBkkt: But in such case need to be revert this bool value

MBkkt: What about `noRemoves`?

MBkkt: Done

Dronplane (approved):
LGTM

arangod/Aql/Condition.cpp
```diff
@@ -1093,7 +1093,8 @@ AstNode* Condition::removeCondition(ExecutionPlan const* plan,
 }
 
 /// @brief remove (now) invalid variables from the condition
-bool Condition::removeInvalidVariables(VarSet const& validVars) {
+bool Condition::removeInvalidVariables(VarSet const& validVars,
+                                       bool& noRemoves) {
```
markuspf, line 1097: Even after staring at the whole code (not just the diff) for multiple minutes it is not clear to me what the relevance of this out-variable (?) is and what the optimizer rule's intent is.

Can you explain that to me?

markuspf (changes requested):

tests/js/common/aql/aql-arangosearch-inverted-index-optimization.js
```diff
@@ -0,0 +1,126 @@
...
+      FOR contribution IN InventoryBelowMin_aggregated_contributions
+      OPTIONS { indexHint: "inverted_InventoryBelowMinLevelPlanEffect", forceIndexHint: true }
+        FOR d IN dates
+          FILTER contribution.main_dimension_key == "pui"
+          FILTER IN_RANGE(d, contribution.FromDT, contribution.ToDT, true, false)
+          RETURN [contribution, d]`;
+      let docs = db._query(query).toArray();
+      assertEqual(docs.length, 5);
+
+      if (isServer) {
```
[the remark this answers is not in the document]
MBkkt: This not only server tests

arangod/Aql/Condition.cpp
Dronplane: @markuspf as this is a bugfix  and we want it to be merged soon. Have you any more  comments to the PR?

tests/js/common/aql/aql-arangosearch-inverted-index-optimization.js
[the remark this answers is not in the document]
markuspf: What does it test that is not fine just on the server?

markuspf (approved):
I understand the fix now, and I think it is correct;

I am rather unhappy with how hard to understand the fix is; 

Mind that I don't think this is not your fault, I just would like to think about ways to improve the situation.

tests/js/common/aql/aql-arangosearch-inverted-index-optimization.js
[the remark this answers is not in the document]
MBkkt: In general I want more client test and less server, I used server only code only because js tests doesn't have necessary helpers for client

Dronplane:
Tests blue except unrelated failures

PR #19309, opened 2023-06-23
[APM-662] Enhanced memory tracking for the AQL Optimizer

adityamukho:
### Scope & Purpose

Use PMR allocators for memory tracking inside the AQL optimizer. This PR introduces PMR allocators and as a start, allows for memory tracking and resource usage monitoring of a RollingVector.

After much trial and error, it was decided to use `boost::container::pmr allocators` in place of the `std::pmr` allocators so as to allow seamless compilation on all platforms (libstdc++ and libc++). This has the added advantage that these allocators work with boost containers as well as std containers.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/APM-662
- [ ] Design document:


arangod/Aql/CountingMemoryResource.h
```diff
@@ -0,0 +1,59 @@
...
+  CountingMemoryResource(memory_resource* base,
+                         ResourceMonitor& resourceMonitor)
+      : base(base), _resourceMonitor(resourceMonitor) {}
+
+ private:
+  void* do_allocate(size_t bytes, size_t alignment) override {
+    void* mem = base->allocate(bytes, alignment);
+    _resourceMonitor.increaseMemoryUsage(bytes);
+
+    return mem;
```
jsteemann, line 41: What if `increaseMemoryUsage(...)` throws here?
Wouldn't that leak the memory acquired by the `allocate(...)` call?

adityamukho: As far as I can tell, I could do one of two things to address this:

1. Swap the order of statements so as to update the resource monitor first and then perform the `allocate`. However, `allocate` could also throw at this point and I'm not really sure what kind of cleanup is required if it does.
2. I could wrap the resource monitor update in a try-catch block, and deallocate in the catch block (followed by a rethrow).

Not sure which, if either, is a good fix.

jsteemann: I think we need to wrap the second statement in a try catch, so if it goes wrong we can roll back the first one.
Only then it will be safe.

arangod/Aql/QueryContext.h
```diff
@@ -36,6 +36,12 @@
 #include "VocBase/voc-types.h"
 #include <velocypack/Builder.h>
 
+#ifdef __APPLE__
+#include <experimental/memory_resource>
+#else
+#include <memory_resource>
+#endif
```
MBkkt, line 43: Could you write this and typedef single time in the single header?
Also it's not correct at least for libc++ and linux

lib/Containers/RollingVector.h
```diff
@@ -38,12 +45,36 @@ namespace containers {
...
+struct pmr_vector_t {
+  typedef std::experimental::pmr::vector<T> type;
+};
+#else
+typedef std::pmr::memory_resource memory_resource_t;
+
+template<typename T>
+struct pmr_vector_t {
+  typedef std::pmr::vector<T> type;
+};
```
MBkkt, line 62: Why not just using?

MBkkt (changes requested):
At least `ifdef __APPLE__` is not enough.
In general you probably more interested in `_LIBCPP_VERSION`

arangod/Aql/CountingMemoryResource.h
```diff
@@ -0,0 +1,88 @@
...
+  void* do_allocate(size_t bytes, size_t alignment) override {
+    void* mem = base->allocate(bytes, alignment);
+
+    try {
+      _resourceMonitor.increaseMemoryUsage(bytes);
+    } catch (...) {
+      base->deallocate(mem, bytes, alignment);
+
+      throw;
+    }
```
MBkkt, line 68: This could be done better via raii

MBkkt, line 68: Also I think to important is call increaseMemoryUsage **before** real allocation.
Because otherwise it's impossible for _resourceMonitor prevent allocation.

arangod/Aql/CountingMemoryResource.h
```diff
@@ -0,0 +1,88 @@
...
+
+    return mem;
+  }
+
+  void do_deallocate(void* p, size_t bytes, size_t alignment) override {
```
MBkkt, line 73: noexcept

arangod/Aql/CountingMemoryResource.h
```diff
@@ -0,0 +1,88 @@
...
+  static memory_resource_t* new_delete_resource() {
+    return std::pmr::new_delete_resource();
+  }
+};
+#endif
+
+struct CountingMemoryResource : memory_resource_t {
+  CountingMemoryResource(memory_resource* base,
+                         ResourceMonitor& resourceMonitor)
+      : base(base), _resourceMonitor(resourceMonitor) {}
```
MBkkt, line 56: noexcept

arangod/Aql/QueryContext.h
```diff
@@ -155,6 +168,8 @@ class QueryContext {
   constexpr static std::size_t baseMemoryUsage = 8192;
 
  protected:
+  std::unique_ptr<memory_resource_t> _memoryResource;
```
MBkkt, line 171: Why unique_ptr needed? You could just put here CountingMemoryResourse.
Probably even better you could put ResourceMonitor to it

lib/Containers/RollingVector.h
hkernbach: Yes will adjust, way better 👍🏼

arangod/Aql/QueryContext.h
MBkkt: I still don't understand why this PR needs pmr and why not just own allocator?

arangod/Aql/QueryContext.h
```diff
@@ -61,6 +62,7 @@ class Ast;
 
 /// @brief an AQL query basic interface
 class QueryContext {
+
```
MBkkt, line 65: unintentionally?
```suggestion
```

3rdParty/boost/CMakeLists.txt
```diff
@@ -125,6 +125,20 @@ target_link_libraries(boost_program_options PUBLIC boost_boost) #this will add r
...
+    ${boost_src}/libs/container/src/alloc_lib.c
+    ${boost_src}/libs/container/src/dlmalloc.cpp
+    ${boost_src}/libs/container/src/global_resource.cpp
+    ${boost_src}/libs/container/src/monotonic_buffer_resource.cpp
+    ${boost_src}/libs/container/src/pool_resource.cpp
+    ${boost_src}/libs/container/src/synchronized_pool_resource.cpp
+    ${boost_src}/libs/container/src/unsynchronized_pool_resource.cpp
+)
+target_link_libraries(boost_container PUBLIC boost_boost) #this will add required includes
+target_compile_definitions(boost_container PRIVATE "BOOST_SYSTEM_STATIC_LINK=1")
```
MBkkt, line 139: it's definitely not boost system :)
Please add it correct way.
At the same time looks like now you're using only single simple header

lib/Basics/Memory/CountingMemoryResource.h
```diff
@@ -0,0 +1,66 @@
...
+
+    try {
+      _resourceMonitor.increaseMemoryUsage(bytes);
+    } catch (...) {
+      base->deallocate(mem, bytes, alignment);
+
+      throw;
+    }
+
+    return mem;
```
MBkkt, line 48: Now it's impossible to prevent big allocation via increaseMemoryUsage
```suggestion
      _resourceMonitor.increaseMemoryUsage(bytes);
      ScopeGuard guard{[&] {
        _resourceMonitor.decreaseMemoryUsage(bytes); 
      }};
      void* mem = base->allocate(bytes, alignment);
      guard.cancel();
      return mem;
```

MBkkt (changes requested):

jsteemann:
Closing this PR as obsolete, because we ultimately went with a different solution that is not based on PMRs.

PR #19312, opened 2023-06-26
Agency Node Refactor

maierlars:
### Scope & Purpose
Refactor agency nodes and prepare them for memory accounting. Instead of mutating nodes in place, we use an immutable representation of the agency. This reduced the memory usage: previously we stored the data three times:
1. Spearhead
2. ReadDB
3. Supervision Snapshot

After this change, all three places point to the same node (in the stable case).
Futhermore, the supervision no longer copies the state every run, but just copies a shared pointer.
Finally, read db no longer applies the transactions a second time but instead copies a shared pointer.
And one other thing: for reading you first copy a shared pointer and then immediately release the lock. Previously the lock was held during read.

**The part with shared spearhead and readDB will happen in another PR. This PR is already big enough and adding that feature requires more work.**


arangod/Agency/FailedFollower.cpp
```diff
@@ -369,7 +370,7 @@ bool FailedFollower::start(bool& aborts) {
 
   slice = result.get(std::vector<std::string>(
       {agencyPrefix, "Supervision", "Health", _to, "Status"}));
-  if (!slice.isString() ||
+  if (slice.isString() &&
```
neunhoef, line 373: This changes the logic: If the slice is not a string (for example, does not exist), we used to see the log message, but now we would no longer see it. Is this intentional?

neunhoef (commented):
Just one comment so far. Will continue the review...

arangod/Agency/Node.h
```diff
@@ -286,26 +176,13 @@ class Node final {
...
   /// @return  returns nullptr if not found or type doesn't match
-  Node const* hasAsNode(std::string const&) const noexcept;
-
-  /// @brief accessor to Node object
-  Node& hasAsWritableNode(std::string const&);
-
-  /// @brief accessor to Node's type
-  /// @return  returns nullopt if not found or type doesn't match
-  std::optional<NodeType> hasAsType(std::string const&) const noexcept;
+  std::shared_ptr<Node const> hasAsNode(std::string const&) const noexcept;
```
neunhoef, line 181: This is the same as `get` with a string argument, we should remove this one here and use `get` directly instead.

arangod/Agency/Node.h
```diff
@@ -286,26 +176,13 @@ class Node final {
...
 
   /// @brief accessor to Node's Slice value
   /// @return  returns nullopt if not found or type doesn't match
-  std::optional<velocypack::Slice> hasAsSlice(
+  std::optional<velocypack::String> hasAsSlice(
```
neunhoef, line 185: This has been and is dangerous and is used only a few times in the code base. Let's remove it and replace calls with other methods.

arangod/Agency/Node.h
```diff
@@ -67,189 +80,66 @@ enum Operation {
...
   /// @brief Get node specified by path
-  Node const* get(
-      std::shared_ptr<cluster::paths::Path const> const& path) const;
-
-  /// @brief Get root node
-  Node const& root() const;
-
-  /// @brief Get root node
-  Node& root();
+  NodePtr get(std::shared_ptr<cluster::paths::Path const> const& path) const;
```
neunhoef, line 122: There is another `get` method further down, let's move it here and explain, that `get` with an argument moves down the tree and returns a `NodePtr` or `nullptr` if the path does not exist.

arangod/Agency/Node.h
```diff
@@ -335,16 +212,10 @@ class Node final {
...
   /// @brief accessor to Node's Array
   /// @return  returns nullopt if not found or type doesn't match
-  std::optional<velocypack::Slice> hasAsArray(std::string const&) const;
-
-  // These two operator() functions could be "protected" once
-  //  unit tests updated.
-  //
-  /// @brief Get node specified by path string
-  Node& getOrCreate(std::string_view path);
+  Array const* hasAsArray(std::string const&) const;
```
neunhoef, line 215: Let's add a code comment here explaining that all `hasAs...` methods walk down the tree and then insist that a certain type is at the end. If not (or if the path does not exist), they return `nullptr` or a `None` option.

arangod/Agency/Node.h
```diff
@@ -335,16 +212,10 @@ class Node final {
...
 
   /// @brief Get node specified by path string
-  Node const* get(std::string_view path) const;
+  NodePtr get(std::string_view path) const;
```
neunhoef, line 218: Let's move this method up to the other `get` methods and add a comment here that the `getTYPE` methods check that the current node has a certain type and return it, or `None` if the type does not match. This gives then some clartity for the user about the API.

neunhoef (commented):
Some comments for `Node.h` as a starter. To be continued...

arangod/Agency/Store.cpp
```diff
@@ -614,51 +607,64 @@ bool Store::applies(arangodb::velocypack::Slice const& transaction) {
...
+      }
+      node = std::move(ret).get();
+      if (node == nullptr) {  // if root node is deleted, keep an empty node
+        node = Node::create();
+      }
       // check for triggers here
       callTriggers(abskeys.at(i.first), op, value);
     } else {
-      success &= _node->hasAsWritableNode(abskeys.at(i.first)).applies(value);
+      node = node->applies(abskeys.at(i.first), value);
```
neunhoef, line 637: I know this has been like this before, but triggers will not be checked in this branch of the if. That is, one can avoid triggers by using the shortcut without `"op"`, I think we should fix this right here.

neunhoef (commented):
I seem to have found an old bug. Commenting before the review is done, to allow quick fixing.

neunhoef (approved):
LVGTM. This PR gives us a substantial improvement in Agency code and also the AgencyCache. The (internal) `Node`/`Store` API has been cleaned up considerably (without breaking things!). But even more substantial are the internal changes in the data structures of the Agency data. The move to immutable data structures and the corresponding improvements will not only save a huge amount of RAM but also speed up operations considerably. This is a cleanup which was long overdue!
I have not found any serious problems and nearly all minor things have already been addressed.
I did find one bug in the agency triggers which had been there before, and which will be fixed in this PR as well.

PR #19313, opened 2023-06-26
RocksDB transactions memory tracking

cpjulia:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1311

* Track memory usage for write operations in AQL queries.
  The memory usage will be reported in the query's `peakMemoryUsage` 
  stats return value, and also be tracked against the global metric
  `arangodb_aql_global_memory_usage`. The memory usage of write 
  operations in AQL queries was previously not tracked and not accounted
  for in AQL queries' own memory usage. The memory used for write operations
  also was not accounted for when setting a memory limit for AQL queries. 
  The memory used for write operations is now properly tracked, including in-memory
  write batches and locked keys in RocksDB (for conflict checking).
  With that change, the memory usage reported for AQL queries that perform
  write operations may be significantly higher than before, but the values
  now will reflect the reality a lot better. This has the consequence that
  previously working AQL write queries can now fail because their actual memory
  usage hits the configured per-query or global memory limit threshold.
  In these cases it is useful to either revisit the per-query and/or global
  memory limit settings, turn on intermediate commits for the query, or adjust
  the query so that it uses less memory.

* Track memory usage for write transactions that are not AQL queries. 
  The memory usage of non-AQL write operations will be tracked in the new
  metrics `arangodb_transactions_rest_memory_usage` and
  `arangodb_transactions_internal_memory_usage`.
  The former metric contains the memory usage of user-initiated write operations, 
  e.g. single or multi-document writes via the document API, JavaScript transactions,
  streaming transactions or index creation. The latter metric contains the 
  memory usage of internal operations, such as background statistics gathering, 
  TTL index maintenance, replication etc.
  Memory usage updates will be reported to these metrics whenever a write
  operation increases or decreases its memory usage by 8kb or more. We do
  this to save lots of small updates of the metrics values, which would be very
  contended otherwise.

* Track memory usage of index creation operations. Writes to indexes are 
  performed in in-memory write batches until they are written to storage. The
  memory usage of these in-memory write batches was previously untracked.

* To make changes more testable, each write operation in the code is tagged with
  a `transaction::OperationOrigin` object. This categorizes a write operation into
  either AQL, REST, or internal. The `OperationOrigin` object is created whenever
  a `transaction::Context` is created first. The context will be responsible for returning
  the original `OperationOrigin` even for nested operations, e.g. if an AQL query 
  (OperationOrigin: AQL) is started inside a streaming transaction (OperationOrigin: REST),
  the returned origin for the nested AQL query will still be REST.
  There is an internal API (not publicly announced nor supported) for fetching the
  meta data of the most recent write operations, e.g. their origins and their memory
  usage. This API is used for testing only. It is only available in maintainer mode and
  deactivated in release mode. There is no plan to make this API available as a feature.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1311
- [ ] Design document:


arangod/Aql/ClusterQuery.cpp
```diff
@@ -136,9 +140,22 @@ void ClusterQuery::prepareClusterQuery(
...
+        if (coll.getCollection()->system()) {
+          isSystem = true;
+          return false;  // returns false for early aborting and not continue
+                         // the traversal, as it's already enough
+        }
+        return true;
+      });
+  if (!isSystem) {
+    _trx->state()->setResourceMonitor(_resourceMonitor);
+  }
```
jsteemann, line 158: Still undefined behavior and not correct
```suggestion
```

arangod/Aql/RestAqlHandler.cpp
```diff
@@ -96,6 +96,7 @@ RestAqlHandler::RestAqlHandler(ArangodServer& server, GeneralRequest* request,
 //    variables: [ <variables> ]
 //  }
 void RestAqlHandler::setupClusterQuery() {
+  LOG_DEVEL << "prepareClusterQuery";
```
jsteemann, line 99:
```suggestion
```

arangod/Replication/TailingSyncer.cpp
```diff
@@ -836,6 +837,7 @@ Result TailingSyncer::changeCollection(VPackSlice const& slice) {
 /// @brief truncate a collections. Assumes no trx are running
 Result TailingSyncer::truncateCollection(
     arangodb::velocypack::Slice const& slice) {
+  LOG_DEVEL << "TailingSyncer::truncateCollection";
```
jsteemann, line 840:
```suggestion
```

arangod/V8Server/v8-vocbase.cpp
```diff
@@ -769,8 +771,9 @@ static void JS_ExecuteAqlJson(v8::FunctionCallbackInfo<v8::Value> const& args) {
     queryId = TRI_NewServerSpecificTick();
   }
 
+  auto origin = transaction::OperationOriginREST{"executing query from JSON"};
```
neunhoef, line 774: Why is this `OriginREST` rather than `OriginAql` as the other AQL calls in this file?

arangod/VocBase/Methods/AqlUserFunctions.cpp
```diff
@@ -44,13 +44,20 @@
...
+
 #include <regex>
+#include <string>
+#include <string_view>
 
 using namespace arangodb;
 
 namespace {
+constexpr std::string_view moduleName = "AQL user functions administration";
+constexpr transaction::OperationOriginREST operationOrigin{moduleName};
```
neunhoef, line 59: Why REST and not AQL?

arangod/V8Server/v8-vocbase.cpp
jsteemann: Fixed

CHANGELOG
```diff
@@ -47,8 +71,6 @@ devel
   The timeouts have now been unified, so that the request timeout is 60 seconds
   and that the HeartbeatThread waits for at most 65 seconds (as before).
 
-* BTS-1598: fix race in agency
```
neunhoef, line 50: Something seems to have been lost with this merge here.

neunhoef (commented):
LGTM so far. See one more comment.
I will now do another experiment and compare the new metrics with the memory profiling.

neunhoef (approved):
LGTM.
I ran examples of large and many ongoing transactions and found the following:

When I compared the actual memory usage (heap profiling) to the one reported by the new transaction memory use metric, there were a few discrepancies:

1. The exact memory accounting is still a bit off (not surprising, because we rely on estimates about the exact behaviour of the RocksDB internal data structures). In particular, the memory used by the PointLockManager is still underestimated.

2. Furthermore, I found that the PointLockManager does not return all its memory after a huge amount of points were locked (for some large transactions). This is, because an `absl::flat_hash_map` is used which does not automatically shrink, even if all locks are released. We should fix this behaviour, since it can use to a large overuse of RAM.

However, I suggest that we fine tune this in a separate PR to get this one out of the way.
The rest is only fine tuning.

neunhoef:
Latest changes: LGTM.

PR #19315, opened 2023-06-26
Optimise runtime lock performance

kvahed:
### Scope & Purpose

*Optimise runtime lock performance*

* Fix lock code when a timeout occurs. In this case it was possible that a queued writer was not notified.
* Disable multi-insert optimization in AQL when it is executed as part of a streaming transaction. This is necessary because the multi-insert optimization will start its own sub-transaction, and does not know about if it is invoked from an outer streaming transaction. If it is invoked from inside a streaming transaction that holds write locks and then starts its own transaction that tries to acquire the same locks, it will deadlock.
* (unimportant) Adjust a test so that it does not acquire a lock recursively on the same collection 3 times.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: *https://github.com/arangodb/arangodb/pull/19346*
  - [x] Backport for 3.10: *https://github.com/arangodb/arangodb/pull/19345*

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1661,63 +1661,38 @@ Result RocksDBMetaCollection::applyUpdatesForTransaction(
...
+    TRI_ASSERT(false);
+    return TRI_ERROR_INTERNAL;
+  }
 
-      startTime = now;
-      waitTime = 1;
-    } else {
-      TRI_ASSERT(startTime > 0.0);
+  if (gotLock) {
+    // keep the lock and exit the loop
```
jsteemann, line 1683: We don't have a loop anymore
```suggestion
    // keep the lock and exit
```

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1661,63 +1661,38 @@ Result RocksDBMetaCollection::applyUpdatesForTransaction(
...
+    return TRI_ERROR_NO_ERROR;
+  }
 
-      if (now > startTime + timeout) {
-        LOG_TOPIC("d1e52", TRACE, arangodb::Logger::ENGINES)
-            << "timed out after " << timeout << " s waiting for "
-            << AccessMode::typeString(mode) << " lock on collection '"
-            << _logicalCollection.name() << "'";
+  TRI_ASSERT(startTime > 0.0);
```
jsteemann, line 1687: `startTime` will always be 0.0 here, so the assertion will be violated.
We can as well remove the `startTime` variable entirely, as it does not serve any purpose anymore.

arangod/RocksDBEngine/RocksDBMetaCollection.h
```diff
@@ -192,7 +191,7 @@ class RocksDBMetaCollection : public PhysicalCollection {
  protected:
   RocksDBMetadata _meta;  /// collection metadata
   /// @brief collection lock used for write access
-  mutable basics::ReadWriteLock _exclusiveLock;
+  mutable std::shared_timed_mutex _exclusiveLock;
```
jsteemann, line 194: If we are using `shared_timed_mutex` now, we must also include `shared_mutex` in the header.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
kvahed: agree. removing

arangod/RocksDBEngine/RocksDBMetaCollection.h
kvahed: sure.

CHANGELOG
```diff
@@ -1,7 +1,9 @@
 devel
 -----
 
-* Attempt to avoid busy looping when locking collections with a timeout
+* Optimise runtime performance of exclusive locks.
```
```suggestion
* Optimize runtime performance of exclusive locks.
```

jsteemann (approved):
LGTM

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1661,63 +1659,35 @@ Result RocksDBMetaCollection::applyUpdatesForTransaction(
...
+  bool gotLock = false;
+  if (mode == AccessMode::Type::WRITE) {
+    gotLock =
+        _exclusiveLock.try_lock_for(std::chrono::duration<double>(timeout));
+  } else if (mode == AccessMode::Type::READ) {
+    gotLock = _exclusiveLock.try_lock_shared_for(
+        std::chrono::duration<double>(timeout));
+  } else {
+    // we should never get here
+    TRI_ASSERT(false);
```
MBkkt, line 1676: Same assert as in the 1665 line. Could you remove one of them?

CHANGELOG
```diff
@@ -37,6 +39,7 @@ devel
 * BTS-1490: query can use a lot more memory when using COLLECT WITH COUNT.
 
 * Attempt to avoid busy looping when locking collections with a timeout
+>>>>>>> 38f46a7e32a48ba83467c529f5bab2ae0350db65
```
mpoeter, line 42:
```suggestion
>>>>>>> 38f46a7e32a48ba83467c529f5bab2ae0350db65
```

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1661,63 +1659,37 @@ Result RocksDBMetaCollection::applyUpdatesForTransaction(
...
 /// @brief lock a collection, with a timeout
 ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
-  uint64_t waitTime = 0;  // indicates that time is uninitialized
-  double startTime = 0.0;
-
   // user read operations don't require any lock in RocksDB, so we won't get
   // here. user write operations will acquire the R/W lock in read mode, and
   // user exclusive operations will acquire the R/W lock in write mode.
   TRI_ASSERT(mode == AccessMode::Type::READ || mode == AccessMode::Type::WRITE);
+  auto tous = std::chrono::duration_cast<std::chrono::microseconds>(
```
mpoeter, line 1666: Can we please rename this to `timeout_us` or just `timeout`? I know the first one does not match our styleguide, but `tous` sounds really nonsensical to me (is this french?).

mpoeter:
@MBkkt Yes, we need a RWLock that favors writers over readers. That's why we use our own implementation.

CHANGELOG
kvahed: already fixed

lib/Basics/ReadWriteLock.cpp
```diff
@@ -130,6 +130,20 @@ void ReadWriteLock::lockRead() {
...
+  }
+  std::unique_lock<std::mutex> guard(_reader_mutex);
+  while (true) {
+    if (tryLockRead()) {
+      return true;
+    }
+    _readers_bell.wait_until(guard, end);
+  }
+  return false;
+}
```
MBkkt, line 146: Is it expected that return false is impossible?

MBkkt (changes requested):

CHANGELOG
mpoeter: Cannot confirm

mpoeter (changes requested):

lib/Basics/ReadWriteLock.cpp
kvahed: Yes

MBkkt: It's a bug.

MBkkt: It will be busy loop if timeout will be reached

mpoeter: Valery is right. The `while (true)` does not honor the timeout. You have to use a similar approach as in `tryLockWriteFor` with the `cv_status`

mpoeter (changes requested):

lib/Basics/ReadWriteLock.cpp
```diff
@@ -130,6 +129,22 @@ void ReadWriteLock::lockRead() {
   }
 }
 
+bool ReadWriteLock::tryLockReadFor(std::chrono::microseconds timeout) {
+  auto end_time = std::chrono::steady_clock::now() + timeout;
+  if (tryLockRead()) {
+    return true;
+  }
```
jsteemann, line 136: We can avoid calculating `end_time` if we reverse the order of these operations:
```suggestion
  if (tryLockRead()) {
    return true;
  }
  auto end_time = std::chrono::steady_clock::now() + timeout;  
```

lib/Basics/ReadWriteLock.cpp
```diff
@@ -130,6 +129,22 @@ void ReadWriteLock::lockRead() {
...
+  std::unique_lock<std::mutex> guard(_reader_mutex);
+  std::cv_status status(std::cv_status::no_timeout);
+  while (std::cv_status::no_timeout == status) {
+    if (tryLockRead()) {
+      return true;
+    }
+    status = _readers_bell.wait_until(guard, end_time);
+  }
```
jsteemann, line 144:
```suggestion
  std::cv_status status(std::cv_status::no_timeout);
  std::unique_lock<std::mutex> guard(_reader_mutex);
  do {
    if (tryLockRead()) {
      return true;
    }
    status = _readers_bell.wait_until(guard, end_time);
  } while (std::cv_status::no_timeout == status);
```

mpoeter (approved):

MBkkt (approved):

kvahed (commented):
LGTM

PR #19318, opened 2023-06-26
experimental PR to improve concurrency control

jsteemann:
### Scope & Purpose

Experimental PR to skip concurrency control in RocksDB for the following RocksDB keys when inserting new documents:
- the document entry in the Documents column family. this entry has a unique, server-generated id, so there should be no conflicts on this key.
- any entries for non-unique secondary indexes. these entries also have the document id included, which is unique and server-generated. there should be no conflicts on them.
The primary index entry and other unique secondary index entries are still properly concurrency-controlled.

For exclusive transactions we don't need to change anything, as in exclusive mode the entire transaction skips concurrency control.
Currently only insert operations are wired, but we may be able to support more operations here.

Skipping RocksDB's concurrency control has the following advantages:
- the keys won't be inserted into the map which holds the transaction's currently locked keys. this saves memory and allocations.
- no locks are acquired for inserting the keys into this map, and later for removing the entries from the map. this could lead to better concurrency in the (striped but shared) lock maps for all transactions.

A simple performance test such as
```
db._drop("x"); 
c = db._create("x"); 
c.ensureIndex({ fields: ["value"], type: "persistent" }); 
docs = []; 
for (i = 0; i < 5000; ++i) {
  docs.push({ value: i });
}
time = require("internal").time; 
s = time(); 
for (i = 0; i < 2 * 1000 * 1000; i += docs.length) {
  c.insert(docs); 
}
print(time() - s);
```
shows a speedup of 10 to 20%.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM. But I do not fully understand the details. Does this prevent the use of the point lock manager? So no locks are held for these keys? Is it allowed to switch this on and off during one transaction? Does this change prevent conflict detection with range deletes?

arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -140,4 +142,30 @@ struct IndexingEnabler {
...
+
+  void setConcurrencyControl(bool value) {
+    if (!_exclusive) {
+      _meth->SetSkipConcurrencyControl(!value);
+    }
+  }
+
+ private:
+  RocksDBMethods* _meth;
+  bool const _exclusive;
```
mpoeter, line 168: We should either check that `_meth` is not null or use a reference instead. Also, can we call it `_methods` instead of `_meth? ;)

jsteemann: I renamed the instance variable from `_meth` to `_methods` in 3 structs in that header.
I did not change its type from pointer to reference because in 2 of the 3 structs `_methods` can legitimately be a nullptr if the initial condition was false. Thus I added assertions to the constructor that the passed in constructor argument `methods` is always a non-nullptr.

arangod/RocksDBEngine/Methods/RocksDBTrxBaseMethods.h
```diff
@@ -47,6 +47,8 @@ class RocksDBTrxBaseMethods : public RocksDBTransactionMethods {
     return _indexingDisabled;
   }
 
+  void SetSkipConcurrencyControl(bool value) final override;
```
MBkkt, line 50: **NIT**
```suggestion
  void SetSkipConcurrencyControl(bool value) final;
```

arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -77,67 +78,98 @@ class RocksDBMethods {
...
+
+  ~ConcurrencyControlToggler() {
+    // nothing to do for exclusive transactions, as they already have the
+    // concurrency control disabled for the entire transaction
+    if (!_exclusive) {
+      _methods->SetSkipConcurrencyControl(false);
+    }
+  }
+
+  void setConcurrencyControl(bool value) {
```
MBkkt, line 162: **NIT**
Why do you decide to change `SetSkipConcurrencyControl` to opposite `setConcurrencyControl`?
Anyway I think `skipConcurrencyControl` is better option.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1399,6 +1399,9 @@ Result RocksDBCollection::insertDocument(transaction::Methods* trx,
     return res.reset(TRI_ERROR_DEBUG);
   }
 
+  ConcurrencyControlToggler cct(mthds, state->isOnlyExclusiveTransaction());
+  cct.setConcurrencyControl(false);
```
MBkkt, line 1403: I think will be nice to move it the ctor, because re-enabling in the dtor

arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -77,67 +78,98 @@ class RocksDBMethods {
...
+      : _methods(methods), _exclusive(isExclusive) {
+    TRI_ASSERT(methods != nullptr);
+  }
+
+  ~ConcurrencyControlToggler() {
+    // nothing to do for exclusive transactions, as they already have the
+    // concurrency control disabled for the entire transaction
+    if (!_exclusive) {
+      _methods->SetSkipConcurrencyControl(false);
+    }
```
MBkkt, line 159:
```suggestion
    setConcurrencyControl(true);
```

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1443,11 +1446,14 @@ Result RocksDBCollection::insertDocument(transaction::Methods* trx,
...
       }
       auto& rIdx = basics::downCast<RocksDBIndex>(*it->get());
+
+      cct.setConcurrencyControl(rIdx.unique());
       // if we already performed the preflight checks,
       // there is no need to repeat the checks once again here
       res = rIdx.insert(*trx, mthds, documentId, doc, options,
                         /*performChecks*/ !performPreflightChecks);
       if (!res.ok()) {
+        cct.setConcurrencyControl(true);
```
MBkkt, line 1456: Don't you want to put `cct.setConcurrencyControl(rIdx.unique());` to the `reverseIdxOps` lambda?

arangod/RocksDBEngine/RocksDBMethods.h
jsteemann: The name of the member inside RocksDB's PessimisticTransaction is also `skip_concurrency_control_`.
And its value defaults to `false`, so it is only called in exceptional situations. So the thing here is not to enable the concurrency control, but rather to disable it (temporarily).

jsteemann: Anyway, I made changes to the PR as a whole now.

arangod/RocksDBEngine/RocksDBCollection.cpp
jsteemann: Done

arangod/RocksDBEngine/RocksDBCollection.cpp
jsteemann: IMHO it is not important to optimize the reverse operation.
The reverse operation is only called in case something went wrong, and the insert is rolled back. This should be a rather exceptional situation for which we do not need to optimize performance.

MBkkt: Agreed, but it doesn't make code more difficult I think, so why not?

jsteemann: It is not difficult, but every further iteration for micro-optimizations cost time & money.

MBkkt: Yeah copy and paste single line is cost a lot of money, especially compared to something like absolutely unnecessary renaming mthds to methods.

Ok, I don't think it's important, so I resolve it

jsteemann:
CI testing uncovered some issues with follow-up requests for the synchronous replication inside streaming transactions, when the collections are locked exclusively. The original locking mode (write/exclusive) for the collection is then somewhat lost and automatically turned into write. That causes some confusion inside the ongoing transaction, which can make RocksDB operations fail.
Shelving the PR for now because of this.

jsteemann:
Obsolete

PR #19321, opened 2023-06-26
[CINFRA] Waiting for rebootId to change

apetenchea:
### Scope & Purpose

The _testSnapshotDiscardedOnRebootIdChange_ does the following:
1. Start a snapshot for one of the replication2 followers
2. Stop the follower
3. Expect the snapshot to have been canceled

Step (3) depends on the follower's rebootId to be increased as a consequence of step (2). However, if the test runs fast enough, there's no guarantee of a rebootID change between (2) and (3).

This PR moves a misplaced `waitFor` right after the follower is stopped.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


hkernbach (approved):

jsteemann (approved):
LGTM

PR #19324, opened 2023-06-26
add LZ4 compression for the edge cache

jsteemann:
### Scope & Purpose

* Added transparent LZ4 compression for values in the in-memory edge cache if their size exceeds a configurable threshold.

  LZ4 compression of edge index cache values allows to store more data in main memory than without compression, so the available memory can be used more efficiently. The compression is transparent and does not require any change to queries or applications. The compression can add CPU overhead for compressing values when storing them in the cache, and for decompressing values when fetching them from the cache.

  The new startup option `--cache.min-value-size-for-edge-compression` can be used to set a threshold value size for compression edge index cache payload values. The default value is `1GB`, which will effectively disable compression until the value is changed. Setting the startup option to a lower value (e.g. `100`) will compress all payloads whose size exceeds that value.

  The new startup option `--cache.acceleration-factor-for-edge-compression` can be used to fine-tune the compression. It controls the LZ4-internal "acceleration" factor used for the compression. The default value is `1`. Higher values typically mean less compression but faster speeds.

  The following new metrics can be used to determine the usefulness of compression:

  - `rocksdb_cache_edge_uncompressed_entries_size`: will return the total number of bytes of all entries that were ever stored in the in-memory edge cache, before compression was applied.
  - `rocksdb_cache_edge_effective_entries_size`: will return the total number of bytes of all entries that were stored in the in-memory edge cache, after compression was applied.
  - `rocksdb_cache_edge_compression_ratio`: overall compression ratio for all entries ever stored in the in-memory edge cache.

  Note that these metrics will be increased upon every insertion into the edge cache, but not decreased when data gets evicted from the cache.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19371
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -355,7 +362,28 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
...
-            VPackSlice cachedData(finding.value()->value());
+            uint8_t const* data = finding.value()->value();
+            if (data[0] == 0xffU) {
+              TRI_ASSERT(finding.value()->size() >= 5);
+              // Custom type. this means the data is lz4-compressed
+              uint32_t uncompressedSize;
+              memcpy(&uncompressedSize, data + 1, sizeof(uncompressedSize));
+              uncompressedSize = basics::bigToHost<uint32_t>(uncompressedSize);
+              if (_lz4CompressBuffer.size() < uncompressedSize) {
+                _lz4CompressBuffer.resize(uncompressedSize);
```
neunhoef, line 373: Is it somehow clear that this code is only used from a single thread? The buffer is local to the object, so this is a question about multithreaded-use of this class.

neunhoef (approved):
LGTM. See my one comment with a question about multii-threaded use.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: Yes, an IndexIterator is supposed to be used by only a single thread.
This is true for all other IndexIterator implementations as well. They are all not thread-safe (because there is no need to make them thread-safe right now).

3rdParty/velocypack
jsteemann, line 1: Note: this PR also updates velocypack to the latest upstream version, because the following change is required for this PR:
```
commit 8915b1dccebb9698e2291607a2e14bd7323bf23d (HEAD -> main, origin/main, origin/HEAD)
Author: Jan <jsteemann@users.noreply.github.com>
Date:   Thu Jun 29 13:53:35 2023 +0200

    make previously private resetTo() and advance() methods public (#161)
```

maierlars (approved):

PR #19333, opened 2023-06-27
[CINFRA-791] Fix supervision flaky test

apetenchea:
### Scope & Purpose

This is a test-only fix targeting `testChangeLeaderToNewFollower`, which seems to be flaky.
The test replaces the leader of a replicated log with a new participant.

Previously, there was a specific order in which the supervision actions where expected to occur:
1) Remove the old leader
```
'UpdateParticipantFlagsAction',  // remove force of leader (previous old follower)
'UpdateParticipantFlagsAction',  // disallow old leader from quorum
'RemoveParticipantFromPlanAction',  // remove old leader
```
2) Prepare the new server for taking over leadership
```
'UpdateParticipantFlagsAction',  // allow new participant as a leader
'UpdateParticipantFlagsAction',  // force new participant
```

Actually, (2) may occur before (1), without affecting the correctness of the system. I have adapted so this scenario is accepted as well.

### DIFF
--- a/tests/js/server/replication2/replication2-supervision-replicated-log-target-cluster.js
+++ b/tests/js/server/replication2/replication2-supervision-replicated-log-target-cluster.js
@@ -821,30 +821,44 @@ const replicatedLogSuite = function () {
 
       const actions = helper.getSupervisionActionTypes(database, logId);
       const expected = [
-        // - create log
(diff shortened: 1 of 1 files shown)


jsteemann (approved):
LGTM

PR #19334, opened 2023-06-27
testing: try to acquire an agency dump even if we are in force shutdown

dothebart:
### Scope & Purpose

Since we mess with SUT by stopping processes now and then, we should not loose hope to acquire an agency dump even if we are in force shutdown.

- [x] :pizza: New feature


mchacki (approved):
LGTM.
Thank you for improving this :+1:

PR #19349, opened 2023-06-29
Memory Accounting Request Bodies

maierlars:
### Scope & Purpose
Adding memory accounting and tracking for request body sizes.


Documentation/Metrics/arangodb_request_body_size.yaml
```diff
@@ -0,0 +1,15 @@
+name: arangodb_request_body_size
```
jsteemann, line 1: We should rename this to `arangodb_request_body_memory_usage`, in order to avoid confusion with the already existing metrics `arangodb_request_body_size_vst`, `arangodb_request_body_size_http1`, `arangodb_request_body_size_http2`.

Documentation/Metrics/arangodb_request_body_size.yaml
```diff
@@ -0,0 +1,15 @@
...
+introducedIn: "3.12.0"
+help: |
+  Memory consumed by incoming request bodies.
+unit: bytes
+type: histogram
```
jsteemann, line 6: I think this metric should be a "gauge", not a "histogram".

arangod/Metrics/GaugeCounterGuard.h
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  void sub(T delta) noexcept {
+    if (_metric) {
+      _metric->fetch_sub(delta);
+      _totalValue -= delta;
+    }
+  }
+
+  void fire() noexcept { reset(); }
```
jsteemann, line 65: nit: IMO having both `fire()` and `reset()` methods is superfluous. We could just use `reset()` for everything.

arangod/Metrics/GaugeCounterGuard.h
```diff
@@ -0,0 +1,80 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+#pragma once
+
+#include "Gauge.h"
+#include "Logger/LogMacros.h"
```
jsteemann, line 24:
```suggestion
```

arangod/GeneralServer/RestHandler.cpp
```diff
@@ -66,7 +67,14 @@ RestHandler::RestHandler(ArangodServer& server, GeneralRequest* request,
               .with<structuredParams::UrlName>(_request->fullUrl())
               .with<structuredParams::UserName>(_request->user())
               .share()),
-      _canceled(false) {}
+      _canceled(false) {
+  if (server.hasFeature<GeneralServerFeature>() &&
+      server.isEnabled<GeneralServerFeature>()) {
+    _requestBodySizeTracker = metrics::GaugeCounterGuard<std::uint64_t>{
+        server.getFeature<GeneralServerFeature>()._requestBodySize,
+        _request->rawPayload().size()};
```
jsteemann, line 75: if the request has Content-Type: application/json, the request may consume additional memory for its internal `_vpackBuilder` if it is an HTTP request. We should also count this memory usage.

arangod/RestHandler/RestBaseHandler.cpp
```diff
@@ -48,7 +48,11 @@ RestBaseHandler::RestBaseHandler(ArangodServer& server, GeneralRequest* request,
 arangodb::velocypack::Slice RestBaseHandler::parseVPackBody(bool& success) {
   try {
     success = true;
-    return _request->payload(true);
+    auto slice = _request->payload(true);
+    if (_request->contentType() != rest::ContentType::VPACK) {
```
jsteemann, line 52: nit: if we are calling `_request->payload(...)` multiple times for the same request, then we will account the memory multiple times here. This is not ideal, because `_request->payload(...)` will allocate memory only for the first time if the Content-Type is application/json, but not for subsequent calls.

PR #19351, opened 2023-06-29
[CINFRA-792] Fix flaky test

apetenchea:
### Scope & Purpose

The `testParticipantFailedOnInsert` has been marked as flaky. Briefly, it does the following:
1) Create a replicated log and wait for leadership to be established
2) Insert an entry into the log and expect it to be committed with a quorum size of 3 (equal to the replication factor).

**Background**
_(1) is where the race happens and it works as follows_
a) Write the log configuration to Target
b) The supervision notices that we want to create a new replicated log and writes an entry in Plan
c) DBServers notice the entry in plan and they start creating the replicated log locally. As they create their logs, they report them to Current.
d) In parallel with (c), the supervision notices servers coming up and adapts the `effectiveWriteConcern`. In this case, it reduces it to 2, which is the minimum necessary for an election to happen. As soon as 2 servers out of 3 have reported to current, a leader is elected and the log is ready to go. This is how we can end up with a quorum of 2, as the replicated log simply does not have to wait for the 3rd server to finish setup.
e) As all the servers finish reporting to Current, the supervision adapts the `effectiveWriteConcern` again to 3. For this to take effect, the Maintenance code of the leader has to notice the change and adapt the `activeParticipantsConfig.effectiveWriteConcern` locally. Only then, a quorum will be guaranteed to have 3 participants.

_(2) is where the test fails spuriously, due to the quorum size being 2 instead of 3._

**Solution**
We want to check that the `effectiveWriteConcern` eventually becomes equal to the replication factor, as all servers are operational. We do this in 3 steps:
- Wait for all servers to report in Current.
- Wait for the supervision to realize that all servers have reported and adjust the `effectiveWriteConcern` in Plan.
- Wait for the leader to react to this change and adjust its internal `effectiveWriteConcern`. As this is not reported to the Agency, our only chance of observing this state is by forcing the leader to update its `committedParticipantsConfig`, which is reported. We achieve this by pinging the log periodically.

**Additional tests**
In order to illustrate step (1.d), I have created an additional test `testLogStartWithMissingFollower`. In this scenario, given 3 servers out of which only 2 are healthy, we should be able to initialize and operate a replicated log. Note that this test currently passes only for `waitForSync=true`. I explain why in the [CINFRA-793 issue](https://arangodb.atlassian.net/browse/CINFRA-793).

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**


tests/js/server/replication2/replication2-soft-write-concern-replicated-log-cluster.js
```diff
@@ -28,48 +28,87 @@ const {assertEqual} = jsunity.jsUnity.assertions;
...
 
+      // Due to the writeConcern of 2, the leader may have established leadership when
+      // only one of the followers was ready. While the committedParticipantsConfig's
+      // effectiveWriteConcern becomes 2, the leadership has been established without
+      // a 3rd participant. It is therefore possible that the next insert will be
+      // committed with a quorum of 2.
+      waitFor(() => {
+        let {plan} = helper.readReplicatedLogAgency(database, logId);
+        if (plan.participantsConfig.config.effectiveWriteConcern !== replicationFactor) {
+          return Error(`effectiveWriteConcern not reached, found ${JSON.stringify(plan)}`);
```
maierlars, line 88:
```suggestion
          return Error(`effectiveWriteConcern not reached, found ${JSON.stringify(plan)}, expected ${replicationFactor}`);
```

maierlars (approved):
LGTM

tests/js/server/replication2/replication2-soft-write-concern-replicated-log-cluster.js
apetenchea: Applied in https://github.com/arangodb/arangodb/pull/19375

PR #19360, opened 2023-07-03
Fix SEARCH-465

MBkkt:
### Scope & Purpose

During collection truncate, after commit it to rocksdb, but before afterTruncate, could iresearch index commit happens.
In such case tick will be released and we may lose info about truncate if crash/error happened in/before afterTruncate.

Test already was there.

Fix is pretty simple, we just lock iresearch index commit mutex, before actual commit to rocksdb happens.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -903,7 +903,6 @@ Result IResearchDataStore::commitUnsafeImpl(
     }();
 #endif
     auto engineSnapshot = _engine->currentSnapshot();
-    TRI_IF_FAILURE("ArangoSearch::DisableMoveTickInCommit") { return {}; }
```
MBkkt, line 906: Because this failure point, test works before this change

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -112,13 +112,17 @@ bool IResearchRocksDBRecoveryHelper::skip(Impl& impl) {
...
 void IResearchRocksDBRecoveryHelper::clear() noexcept {
-  if constexpr (!Force) {
+  if constexpr (Force) {
+    if (!_ranges.empty()) {
+      _ranges.clear();
+    }
+  } else {
     if (_indexes.size() < kMaxSize && _links.size() < kMaxSize &&
         _ranges.size() < kMaxSize) {
       return;
```
jsteemann, line 122: I guess it is intentional that we don't clear `_ranges` here anymore, but please confirm.

MBkkt: we clear

MBkkt: image
line 124

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -112,13 +112,17 @@ bool IResearchRocksDBRecoveryHelper::skip(Impl& impl) {
 
 template<bool Force>
 void IResearchRocksDBRecoveryHelper::clear() noexcept {
-  if constexpr (!Force) {
+  if constexpr (Force) {
+    if (!_ranges.empty()) {
+      _ranges.clear();
+    }
```
MBkkt, line 118: A little optimization for case when few IndexCreate/IndexDrop happens in row

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1666,8 +1665,9 @@ void IResearchDataStore::recoveryInsert(uint64_t recoveryTick,
   // TODO(MBkkt) IndexWriter::Commit? Probably makes sense only if were removes
 }
 
-void IResearchDataStore::afterTruncate(TRI_voc_tick_t tick,
-                                       transaction::Methods* trx) {
+void IResearchDataStore::truncateCommit(TruncateGuard guard,
```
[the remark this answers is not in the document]
MBkkt: I think it worse, because it will be rvalue reference to unique_ptr

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
```diff
@@ -112,13 +112,17 @@ bool IResearchRocksDBRecoveryHelper::skip(Impl& impl) {
 
 template<bool Force>
 void IResearchRocksDBRecoveryHelper::clear() noexcept {
-  if constexpr (!Force) {
+  if constexpr (Force) {
+    if (!_ranges.empty()) {
```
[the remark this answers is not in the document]
MBkkt: Not really needed?
Only hash set clear is not O(1)
But you right I can write it a little better, thanks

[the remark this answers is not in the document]
MBkkt: Done

arangod/RocksDBEngine/RocksDBIndex.cpp
```diff
@@ -239,8 +239,19 @@ Result RocksDBIndex::drop() {
   return r;
 }
 
-void RocksDBIndex::afterTruncate(TRI_voc_tick_t,
-                                 arangodb::transaction::Methods*) {
+ResultT<TruncateGuard> RocksDBIndex::truncateBegin(rocksdb::WriteBatch& batch) {
+  auto bounds = getBounds();
+  auto s =
+      batch.DeleteRange(bounds.columnFamily(), bounds.start(), bounds.end());
+  auto r = rocksutils::convertStatus(s);
```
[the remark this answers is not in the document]
MBkkt: Because assert in ResultT(Result&&) ctor :(

arangod/IResearch/IResearchRocksDBRecoveryHelper.cpp
jsteemann: Ok, but with the PR's changes we are clearing `_ranges` under different conditions as before. 
Which may be fine, but I don't know.

jsteemann: If `Force` is true, we are now clearing unconditionally. Previously we only cleared `_ranges` if the other if condition was true.

MBkkt: Because if force is true it could be empty, in such case clear is unnecessary, because it's already empty.
But if force is false it's never empty (after not force check) so we don't need additional check

jsteemann (approved):
LGTM

arangod/IResearch/IResearchDataStore.cpp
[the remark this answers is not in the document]
MBkkt: Done, because guard need to be alive before noexcept trigger crash

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -787,25 +789,29 @@ Result RocksDBCollection::truncateWithRangeDelete(transaction::Methods& trx) {
...
     return rocksutils::convertStatus(s);
   }
+  // we need crash server if any failure in order to trigger recovery,
+  // as rocksdb truncate already committed here
+  [&]() noexcept {
+    // post commit sequence
+    rocksdb::SequenceNumber seq = db->GetLatestSequenceNumber() - 1;
+
+    uint64_t numDocs = _meta.numberDocuments();
+    _meta.adjustNumberDocuments(seq,
```
jsteemann, line 799: `RocksDBMetadata::adjustNumberDocuments` performs a `try_emplace` into a container and thus can throw. It is not `noexcept` but called from a lambda marked `noexcept`. Is this intentional?

MBkkt: Yep, I made comment which explains why.
If exception happens we have two options:
1. crash => recovery
2. make rollback for truncate on already truncated indexes and collection

2 iss impossible I think, so decided to use 1

Dronplane: @jsteemann  as I'm going to merge this. Have you any concerns/suggestions with this PR?

PR #19363, opened 2023-07-03
[DUMP] Retry on write, read and connect errors

maierlars:
### Scope & Purpose
Also retry on write, read and connect errors. Initially I thought there was no way of knowing why the request failed, but I was wrong. Additionally retry creating dump contexts.

The following things have also been changes:
1. The server enforces a empty response body if the response code is `204 No Content`. This already almost always true. The problematic exception was the new dump api.
2. The `SimpleHttpClient` will drop the body of all http responses that have response code `204`. This is done by dropping the connection if the `Content-Length` is not zero and the response code is `204`.
3. The dump api now explicitly creates an empty body in case 204.

### DIFF
--- a/arangod/GeneralServer/HttpCommTask.cpp
+++ b/arangod/GeneralServer/HttpCommTask.cpp
@@ -600,6 +600,13 @@ void HttpCommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> baseRes,
   // will add CORS headers if necessary
   this->finishExecution(*baseRes, _origin);
 
+  // handle response code 204 No Content
+  if (response.responseCode() == rest::ResponseCode::NO_CONTENT) {
+    TRI_ASSERT(response.isResponseEmpty())
+        << "response code 204 requires body to be empty";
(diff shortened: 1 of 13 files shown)


jsteemann (approved):
LGTM

kvahed (approved):
LGTM

arangod/GeneralServer/HttpCommTask.cpp
```diff
@@ -600,6 +600,11 @@ void HttpCommTask<T>::sendResponse(std::unique_ptr<GeneralResponse> baseRes,
   // will add CORS headers if necessary
   this->finishExecution(*baseRes, _origin);
 
+  // handle response code 204 No Content
+  if (response.responseCode() == rest::ResponseCode::NO_CONTENT) {
+    response.clearBody();  // in non maintainer build clear the body
```
jsteemann, line 605: I guess we are clearing the body in both maintainer and non-maintainer mode. Or am I missing something?
```suggestion
    response.clearBody();
```

jsteemann (approved):
LGTM

PR #19371, opened 2023-07-04
add LZ4 compression for the edge cache

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19324

* Added transparent LZ4 compression for values in the in-memory edge cache if their size exceeds a configurable threshold.

  LZ4 compression of edge index cache values allows to store more data in main memory than without compression, so the available memory can be used more efficiently. The compression is transparent and does not require any change to queries or applications. The compression can add CPU overhead for compressing values when storing them in the cache, and for decompressing values when fetching them from the cache.

  The new startup option `--rocksdb.min-value-size-for-edge-compression` can be used to set a threshold value size for compression edge index cache payload values. The default value is `100`, i.e. values with a payload size of at least 100 bytes will be LZ4-compressed. To turn compression off, it is enough to set the option to a very large value.

  The following new metrics can be used to determine the usefulness of compression:

  - `arangodb_edge_cache_uncompressed_entries_size`: will return the total number of bytes of all entries that were ever stored in the in-memory edge cache, before compression was applied.
  - `arangodb_edge_cache_effective_entries_size`: will return the total number of bytes of all entries that were stored in the in-memory edge cache, after compression was applied.

  Note that these metrics will be increased upon every insertion into the
  edge cache, but not decreased when data gets evicted from the cache.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -169,6 +169,12 @@ DECLARE_COUNTER(arangodb_revision_tree_hibernations_total,
                 "Number of revision tree hibernations");
 DECLARE_COUNTER(arangodb_revision_tree_resurrections_total,
                 "Number of revision tree resurrections");
+DECLARE_GAUGE(arangodb_edge_cache_uncompressed_entries_size, uint64_t,
+              "Total gross memory size of all edge cache entries ever stored "
+              "in the cache");
+DECLARE_GAUGE(arangodb_edge_cache_effective_entries_size, uint64_t,
+              "Total effective memory size of all edge cache entries ever "
+              "stored in the cache");
 
 // global flag to cancel all compactions. will be flipped to true on shutdown
 static std::atomic<bool> cancelCompactions{false};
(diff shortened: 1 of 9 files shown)


kvahed (approved):
LGTM

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -755,6 +772,43 @@ replication doing a resync, however.)");
...
+option to a value that is lower than medium-to-large average payload sizes.
+It is normally not that useful to compress values that are smaller than 100 bytes.)")
+      .setIntroducedIn(31102);
+
+  options
+      ->addOption(
+          "--cache.acceleration-factor-for-edge-compression",
+          "The acceleration factor for the LZ4 compression of in-memory "
+          "edge cache entries.",
+          new UInt32Parameter(&_accelerationFactorForEdgeCompression, 1, 1, 10),
```
MBkkt, line 800: Is 10 for this function mean max value? In such case it's incorrect, exist valid and useful examples with 50.
https://github.com/inikep/lzbench/blob/master/lz4/lz4.h#L190 correct max

Also I think important to note that 1 step increase compression speed commonly on 3%, and could increase a little decompress speed.
Or just point that lz4_compress_fast parameter

MBkkt (approved):

PR #19353, opened 2023-06-30
[CINFRA-777] Report syncIndex to the leader

apetenchea:
### Scope & Purpose

Followers report their `syncIndex` to the leader (in the `AppendEntriesResponse`), and let the leader track it for all participants.

This is the continuation of https://github.com/arangodb/arangodb/pull/19285

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**

PR #19375, opened 2023-07-04
[CINFRA-778] Use syncIndex for compaction

apetenchea:
### Scope & Purpose

This PR is a follow-up to https://github.com/arangodb/arangodb/pull/19353 and concludes the [Make compaction work with waitForSync=false](https://arangodb.atlassian.net/browse/CINFRA-769) story.

The main modification has some impact on the `checkCommitIndex` method, where inside `collectFollowerStates`, we use the `syncIndex` to calculate the `largestCommonIndex`, instead of the spearhead, as it was before.
In order to aid with debugging and testing, I added the `syncIndex` field to `LocalStatistics`, such that it can be retrieved when querying the status of a log.

**Note** that I enabled the `waitForSync` option in `InMemoryLogManager.cpp`, where it was previously hard-coded to true. This is needed for the tests to work as expected.

`syncCommitIndex` has been added to `LeaderStatus`. This shows what the `commitIndex` would be if only synced entries would be taken into account. It is not used internally, but only for better observability of the system.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] :books: Updated [internal documentation](https://arangodb.atlassian.net/wiki/spaces/CInfra/pages/1780973607/Concept+Log+Compaction+Release-Index+and+LITK) and [Glossary](https://arangodb.atlassian.net/wiki/spaces/CInfra/pages/1817509948/Glossary)


tests/js/server/replication2/replication2-replicated-log-wait-for-sync.js
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      // For wait-for-sync true, we expect the syncIndex to be equal to the spearhead
+      let status = log.status();
+      for (const [k, v] of Object.entries(status.participants)) {
+        assertEqual(v.response.local.syncIndex, v.response.local.spearhead.index,
+          `syncIndex should be equal to spearhead for ${k}, status: ${JSON.stringify(status)}`);
+        assertEqual(v.response.lowestIndexToKeep, 101,
+          `lowestIndexToKeep should be 101 for ${k}, status: ${JSON.stringify(status)}`);
+      }
```
maierlars, line 84: I was initially confused about this. The reason for my confusion was that I thought, the spearhead might be higher than the commit index. But `syncIndex` and `commitIndex` are not directly related. There is something like a `syncCommitIndex`, which is essentially calculating the commit index but using the `syncIndex` for each follower instead.

In `waitForSync=true` case they it is the same, however.

[This comment contains not action for you, please ignore it.]

maierlars (approved):

tests/js/server/replication2/replication2-replicated-log-wait-for-sync.js
apetenchea: Added `syncCommitLeader` to `LeaderStatus`.

maierlars (approved):

PR #19390, opened 2023-07-06
Add python code to control a cluster and HotBackup test script

markuspf:
### Scope & Purpose

Adds a bit of python code to the toolbox to start a cluster which does the same job as `startLocalCluster.sh`, except that it is programmable in python.

Based on that code adds a test for HotBackups; this test hangs on `devel` and requires the work in #18928 to become green. It is only in here to demonstrate how to use it.


maierlars (approved):

apetenchea (approved):
Really nice 💪

PR #19395, opened 2023-07-07
[CINFRA] Initializing syncIndex to 0

apetenchea:
### Scope & Purpose

**Changes**
The `syncIndex` is initialized to 0 at the start of a term and updated to its real value when the first append-entries request if successfully synced.

**Why?**
We are not initializing it with the current spearhead anymore, because in a different term the leader might replace parts of the log (using `removeBack`) - which might make it seem, at least temporarily, as if the `syncIndex` is greater than the spearhead. The safest thing to do is to initialize it to 0 - worst case, compaction may be delayed for a bit, but that is not critical.
Note that, with this change, the `syncIndex` may be decreased - it can get to 0 during a term change. This implies that `syncCommitIndex` may be decreased as well.
I have added a couple of new assertions, making sure that `syncIndex <= spearhead` always holds.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
+++ b/arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
@@ -160,11 +160,7 @@ StorageManager::StorageManager(std::unique_ptr<IStorageEngineMethods> methods,
     : guardedData(std::move(methods)),
       loggerContext(
           loggerContext.with<logContextKeyLogComponent>("storage-manager")),
-      scheduler(std::move(scheduler)),
-      syncIndex(getTermIndexMapping()
-                    .getLastIndex()
-                    .value_or(TermIndexPair{})
-                    .index) {}
+      scheduler(std::move(scheduler)) {}
 
 auto StorageManager::resign() noexcept
     -> std::unique_ptr<IStorageEngineMethods> {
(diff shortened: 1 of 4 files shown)


arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
```diff
@@ -447,11 +443,16 @@ auto StorageManager::getCommittedLogIterator(
...
 }
 
-void StorageManager::updateSyncIndex(LogIndex index) {
+void StorageManager::updateSyncIndex(TermIndexPair index) {
   syncIndex.doUnderLock([index](auto& currentIndex) {
+    TRI_ASSERT(index.term >= currentIndex.term)
+        << "trying to update sync "
+           "index with a lower term "
+           "than the current one "
+        << currentIndex << " vs " << index;
```
maierlars, line 455: This can and will fail in production. It is totally valid for the leader to send log entries that have a lower term than the current term. Those log entries might be entirely new to the follower and when they are synced, this assertion will simply pop.

maierlars (changes requested):

arangod/Replication2/ReplicatedLog/Components/StorageManager.cpp
apetenchea: I have reverted this

maierlars (approved):

PR #19407, opened 2023-07-10
BTS-1512: testing: if we have a crashdump, test state must be failed

dothebart:
- forward port of https://github.com/arangodb/oskar/pull/526/files to circle ci
- don't create coredumps on failure points

- [x] Backports
 - [x] 3.11 https://github.com/arangodb/arangodb/pull/19411
 - [x] 3.10 https://github.com/arangodb/arangodb/pull/19412
 - [x] 3.9 https://github.com/arangodb/arangodb/pull/19413

### DIFF
--- a/arangod/IResearch/IResearchDataStore.cpp
+++ b/arangod/IResearch/IResearchDataStore.cpp
@@ -1684,6 +1684,7 @@ void IResearchDataStore::truncateCommit(TruncateGuard&& guard,
   };
 
   TRI_IF_FAILURE("ArangoSearchTruncateFailure") {
+    CrashHandler::setHardKill();
     THROW_ARANGO_EXCEPTION(TRI_ERROR_DEBUG);
   }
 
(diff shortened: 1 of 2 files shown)


mpoeter (approved):

arangod/IResearch/IResearchDataStore.cpp
```diff
@@ -1684,6 +1684,7 @@ void IResearchDataStore::truncateCommit(TruncateGuard&& guard,
   };
 
   TRI_IF_FAILURE("ArangoSearchTruncateFailure") {
+    CrashHandler::setHardKill();
```
MBkkt, line 1687: I don't sure. I'm ok with such fix.

But in general the most important question here is
How process should exit in case of some exception was thrown after truncate written to rocksdb wal,
but before truncate applied to all indexes.

Is it ok terminate in such case? For users, not for tests!
I asked that last time and there are no concerns about that.
But I think I need to explicitly ask it again.

dothebart: should that question rather be in a BTS since its not the main topic of the current issue?

mpoeter (approved):

MBkkt (approved):
Approve to unblock other prs

PR #19410, opened 2023-07-10
[CINFRA] Fix flaky test

apetenchea:
### Scope & Purpose

The following assertion failed
```
assertEqual(status.participants[followers[0]].response.local.syncIndex, fixedFollowerSyncIndex)
```
Between the time we take a "snapshot" of the sync indices and the time we set the failure point, there could be some `syncIndex` updated, due to some delayed log pings. Moving the status. Moving the `log.status()` call after the `debugSetFailAt` guarantees there's going to be no more `syncIndex` updates on the follower after the status is retrieved.
Additionally, we're now waiting for all the followers to catch up with the fixed sync index, before fixing the lowest index to keep. Before this patch, in case on of the followers stayed behind, the lowest index to keep could have still increased until the delayed follower caught up.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM

mchacki:
CircleCI tests are green:
Relevant Jenkins as well.
Other issues are related to current Jenkins situation.

PR #19423, opened 2023-07-13
[BTS-1428] Explaining the reson for suppression

apetenchea:
### Scope & Purpose

A lock-order inversion inside the replication2 code has been suppressed. My initial response was that it's a false positive. I have just double checked together with @goedderz and we confirmed it.
This PR adds an explanation in the TSAN suppressions file.


mpoeter (approved):

mpoeter:
Thanks for adding a detailed explanation about _why_ this is a false positive! No build necessary since this only adds comments.

PR #19424, opened 2023-07-13
Reuse spearhead nodes as read db.

maierlars:
### Scope & Purpose
Final step in the refactoring of the agency nodes. Instead of applying transactions twice and storing the agency state twice, we keep the old spearhead pointers around and reuse them as readDB once the index is committed.


neunhoef (approved):
LGTM

maierlars:
Agency code is way to fragile for this change. We need larger refactorings and cleanups before we can do something like this.

PR #19426, opened 2023-07-13
3.11 Test-Only - Added a more sophisticated test for createCollection API

mchacki:
### Scope & Purpose

*This PR adds a test suite containing tests around the edge-cases of cerateCollection API. We plan to rearrange some internals of this API with 3.12 and wewant to use the test file as reference to not accidentially break behavior.*

**Correction:**
This PR actually fixes 2 items:
1) Activate DC2DC special handling if Authentication is disabled (I think this is no production use case)
2) Fixed an assertion we could trigger, but which is a valid call to the API with a valid return statement. (See Enterprise)

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1317
- [ ] Design document:


tests/js/common/shell/shell-collection-api.js
```diff
@@ -0,0 +1,1691 @@
...
+  ERROR_HTTP_CORRUPTED_JSON
+} = require("internal").errors;
+
+const filterClusterOnlyAttributes = (attributes) => {
+  // The following attributes are not returned by the Collection Properties API
+  // However the user can still send them to the API. Erase them here as they should
+  // not be returned by the server.
+  if (!isCluster) {
+    const clusterOnlyAttributes = ["numberOfShards", "shardKeys", "replicationFactor"];
+    /* , , "writeConcern" */
```
jsteemann, line 58: why is `writeConcern` not part of this list? It should be a cluster-only attribute. If it's not, please leave at least a comment in the test file why it is not included in the list.
Wondering also if we should exclude `smartJoinAttribute` here.

jsteemann: We may also still return the `minReplicationFactor` attribute, which was the predecessor of `writeConcern`. Should it be excluded here too?

jsteemann: Same for `shardingStrategy`, which should be cluster-only.

tests/js/common/shell/shell-collection-api.js
```diff
@@ -0,0 +1,1691 @@
...
+      }
+    },
+
+    testCreateEdge: function () {
+      const res = tryCreate({name: collname, type: 3});
+      try {
+        assertTrue(res.result, `Result: ${JSON.stringify(res)}`);
+        if (isServer) {
+          // Type in options is ignored, need third parameter
+          validateProperties({}, collname, 2);
```
jsteemann, line 301:
```suggestion
          validateProperties({}, collname, 3);
```

tests/js/common/shell/shell-collection-api.js
mchacki: 1) writeConcern is actually exposed on SingleServer, therefore it is not on the list.
Will add a comment.
2) minReplicationFactor is sometimes exposed on SingleServer, namely when this filtering is disabled
2a) Enterprise Edition using Enterprise Features like satellites and SmartGraphs disable this filtering, they expose the given list of ClusterOnlyAttributes even on SingleServer.
3) shardingStrategy and distributeShardsLike seem to be always ignored.

I try to come up with a way to make this a bit easier to understand on the Test side.

tests/js/common/shell/shell-collection-api.js
```diff
@@ -0,0 +1,1691 @@
...
+      const res = tryCreate({name: collname, isSmart: true});
+      try {
+        if (!isEnterprise) {
+          assertTrue(res.result, `Result: ${JSON.stringify(res)}`);
+          // isSmart is just ignored, we keep default: false
+          validateProperties({}, collname, 2);
+          validateDeprecationLogEntryWritten();
+          if (!isCluster) {
+            // SingleSever does not expose is Smart at all
+            validatePropertiesDoNotExist(collname, ["isSmart"]);
```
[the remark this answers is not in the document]
mchacki: Yes we do,
and if you create a real SmartVertexCollection (see: testCreateSmartGraph) the value isSmart: true is exposed in Enterprise mode on Single and Cluster.
It seems that on SingleServer the value can only be "true" or not exist.
On Cluster it can be only be "false" or "true".

At least that is what Is exposed via the properties API.

jsteemann (approved):
LGTM

goedderz (approved):

PR #19434, opened 2023-07-17
Fixing Hotbackup Test

maierlars:
### Scope & Purpose
1. After a restore the cluster is not immediately ready for action. (For example AQL instantiation fails with missing shard). Hence we added a small grace period.
2. The streaming transaction test (el cheapo) did not use transactions at all. Obviously it failed.


jsteemann (approved):
LGTM

PR #19442, opened 2023-07-18
Memory Accounting Agency Nodes

maierlars:
### Scope & Purpose
Memory accounting for agency nodes. Some members are not yet accounted for, because that would change the API to much. (For example the keys of the `Children` map are still `std::string`s.)


jsteemann (approved):
LGTM

PR #19443, opened 2023-07-18
[CircleCI] Add non-maintainer build job

mpoeter:


.circleci/base_config.yml
```diff
@@ -313,6 +327,16 @@ workflows:
           preset: community-pr
           edition: x64-community
           enterprise: false
+          targets: "arangodbtests arangod arangosh arangoimport arangoexport arangodump arangorestore arangobench fuertetest"
```
mchacki, line 330: Maybe we can find a better solution to duplicating this list in the future, but this is such a minor improvement that we do not need to block this change

mchacki (approved):
LGTM!

PR #19428, opened 2023-07-14
Make post recovery callbacks parallel

MBkkt:
### Scope & Purpose

It should helps for case when we have more than single heavy link.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19344, opened 2023-06-28
Try to make cluster tests tsan run more useful

MBkkt:
### Scope & Purpose

* Suppress very popular and easy to reproduce replication2 lock order inversions
  These errors make all report unreadable because they spam it
* Fix 3 simple data races
* Remove already unnecessary suppressions

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19444, opened 2023-07-18
Backports to 3.11

MBkkt:
### Scope & Purpose

* Fix unstable test (#19281)
* Try to make cluster tests tsan run more useful (#19344) 
* Fix SEARCH-465 (#19360)
* Make post recovery callbacks parallel (#19428)

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1323
- [ ] Design document:


mpoeter (dismissed):
Completely missing any description of what this PR is about.

Dronplane (approved):
LGTM

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -586,9 +587,19 @@ void DatabaseFeature::recoveryDone() {
...
   // '_pendingRecoveryCallbacks' will not change because
   // !StorageEngine.inRecovery()
+  // It's single active thread before recovery done,
+  // so we could use general purpose thread pool for this
+  std::vector<futures::Future<Result>> futures;
+  futures.reserve(_pendingRecoveryCallbacks.size());
   for (auto& entry : _pendingRecoveryCallbacks) {
-    auto result = entry();
-
+    futures.emplace_back(SchedulerFeature::SCHEDULER->queueWithFuture(
```
jsteemann, line 595: I don't know exactly what's in a recovery callback object, but as we are running them concurrently now, are they all fully thread-safe?

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -787,25 +789,29 @@ Result RocksDBCollection::truncateWithRangeDelete(transaction::Methods& trx) {
...
     return rocksutils::convertStatus(s);
   }
+  // we need crash server if any failure in order to trigger recovery,
+  // as rocksdb truncate already committed here
+  [&]() noexcept {
+    // post commit sequence
+    rocksdb::SequenceNumber seq = db->GetLatestSequenceNumber() - 1;
+
+    uint64_t numDocs = _meta.numberDocuments();
+    _meta.adjustNumberDocuments(seq,
```
jsteemann, line 799: `RocksDBMetadata::adjustNumberDocuments()` can throw an exception in an emplace operation goes wrong (e.g. when there is no memory available).
In this case, the exception will bubble up from there into this function, which is marked `noexcept`. This will make the server crash with `std::terminate`.
If this behavior is intended, it is all fine, but it is not clear to me rn.

arangod/RestServer/DatabaseFeature.cpp
MBkkt: For now yes, they're thread-safe. We also run sanitizers.
In future, if someone decide to add new postrecovery callback he will need to account this.
So doxygen was changed.

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: Yes it's intentionally because if something throw after we put data in rocksdb we will state in the inconsistent state.

jsteemann: Thanks, I already forgot about it.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -787,25 +789,29 @@ Result RocksDBCollection::truncateWithRangeDelete(transaction::Methods& trx) {
   if (!s.ok()) {
     return rocksutils::convertStatus(s);
   }
+  // we need crash server if any failure in order to trigger recovery,
+  // as rocksdb truncate already committed here
```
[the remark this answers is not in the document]
jsteemann: Because it is not explicitly mentioned that the way to crash the server is to mark the function as `noexcept`. One has to reverse-engineer the motivation from the code, not from the comment.
A clearer comment would have been something like
"We intentionally mark this lambda noexcept, so that when any exception occurs inside the lambda (can happen in case of out of memory), we crash the server instead of continuing with an inconsistent state. The post-crash recovery will then get everything into a consistent state again.".

jsteemann (approved):
LGTM

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt, line 793:
```suggestion
  // We intentionally mark this lambda noexcept, so that when any exception
  // occurs inside the lambda (can happen in case of out of memory for an example),
  // we crash the server instead of continuing with an inconsistent state. 
  // The post-crash recovery will then get everything into a consistent state again.
```

MBkkt: :(

PR #19446, opened 2023-07-19
avoid recursive locking in agency

jsteemann:
### Scope & Purpose

when loading configuration with an AQL query (which acquires the collection's status lock) and then starting a transaction on the same collection (acquires the status lock once more) while the query object is still in scope. this leads to the same thread locking the same rw lock in read mode twice, which can be dangerous if a writer is queued for the lock. then the two threads will deadlock.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19448
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19447
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19447, opened 2023-07-19
avoid recursive locking agency

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19446

when loading configuration with an AQL query (which acquires the collection's status lock) and then starting a transaction on the same collection (acquires the status lock once more) while the query object is still in scope. this leads to the same thread locking the same rw lock in read mode twice, which can be dangerous if a writer is queued for the lock. then the two threads will deadlock.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19448
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Agency/State.cpp
+++ b/arangod/Agency/State.cpp
@@ -1002,16 +1002,24 @@ index_t State::loadCompacted() {
 
 /// Load persisted configuration
 bool State::loadOrPersistConfiguration() {
-  std::string const aql(
-      "FOR c in configuration FILTER c._key==\"0\" RETURN c.cfg");
+  // this must be in a lambda so the query lifetime does not overlap
+  // with the lifetime of the other transaction in this function.
+  // otherwise, they may both try to acquire the status lock of the
+  // same collection at the same time, which would be recursive locking.
+  auto loadConfiguration = [&]() -> aql::QueryResult {
+    std::string const aql(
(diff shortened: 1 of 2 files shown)


arangod/Agency/State.cpp
```diff
@@ -1002,16 +1002,24 @@ index_t State::loadCompacted() {
 
 /// Load persisted configuration
 bool State::loadOrPersistConfiguration() {
-  std::string const aql(
-      "FOR c in configuration FILTER c._key==\"0\" RETURN c.cfg");
+  // this must be in a lambda so the query lifetime does not overlap
```
MBkkt, line 1005: well lambda is not required it's possible to do same without it.
we can just write query.reset();

But I think it doesn't matter

MBkkt (approved):

mchacki (approved):
LGTM

PR #19449, opened 2023-07-19
[BTS-1511] Fix AqlValues with 7-byte integers

goedderz:
### Scope & Purpose

AqlValues save and access integers of differents sizes differently. Those that fit into up to 48bits are saved as a native 64bit integer (either signed or unsigned), beside the originating slice. There's not enough space for that for larger integers, so the slice is saved so that the contained integer is aligned, and this is then interpreted directly as an integer (plus possible little endian to host endianness conversion). This affects integers of 7 or 8 bytes in size.

If the originating slice contained a 7 byte integer, the 8th byte was not initialized.

These optimizations (including this bug) were introduced in https://github.com/arangodb/arangodb/pull/13569.

Note that the final byte cannot simply be initialized with zero, but must be sign-extended - i.e. must be 0x00 for nonnegative integers, but 0xff for negative.

- [X] :hankey: Bugfix

### Checklist

- [X] Tests
  - [X] **Regression tests**
  - [X] C++ **Unit tests**
- [X] Backports
  - [X] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19457
  - [X] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19458
  - [X] Backport for 3.9: https://github.com/arangodb/arangodb/pull/19459

#### Related Information

- Jira ticket: https://arangodb.atlassian.net/browse/BTS-1511
- Introduced in: https://github.com/arangodb/arangodb/pull/13569
- Related bugfix: https://github.com/arangodb/arangodb/pull/13579


arangod/Aql/AqlValue.cpp
```diff
@@ -1674,6 +1675,27 @@ void AqlValue::initFromSlice(arangodb::velocypack::Slice slice,
...
+        // Set the remaining byte (if any):
+        std::memset(_data.longNumberMeta.data.slice.slice + length, filler,
+                    remainingBytes);
+        TRI_ASSERT(
+            slice.isUInt()
+                ? slice.getUInt() > std::numeric_limits<std::int64_t>().max() ||
+                      slice.getUInt() == static_cast<uint64_t>(this->toInt64())
+                : slice.getInt() == this->toInt64())
+            << (slice.isUInt() ? "uint " : "int ") << slice.toJson()
+            << " != " << this->toInt64();
```
MBkkt, line 1698:
```suggestion
        auto& last_byte = _data.longNumberMeta.data.slice.slice[8];
        last_byte = 0x00;
        memcpy(_data.longNumberMeta.data.slice.slice, slice.begin(),
               static_cast<size_t>(length));
        // If length == 9, we're done;
        // If length == 8, there's one byte left to fill if it's signed int;
        if (length == 8 && !slice.isUint() && slice.begin()[7] > 0x7F) {
          last_byte = 0xFF;
        }
        TRI_ASSERT(
            slice.isUInt()
                ? slice.getUInt() > std::numeric_limits<std::int64_t>().max() ||
                      slice.getUInt() == static_cast<uint64_t>(this->toInt64())
                : slice.getInt() == this->toInt64())
            << (slice.isUInt() ? "uint " : "int ") << slice.toJson()
            << " != " << this->toInt64();
```

signed shift for negative values is implementation defined behavior.
Also any arguments smaller than int implicitly casted to the int.
why not simplify it?

Also it looks like `!slice.isUint()` in code above unnecessary.

mpoeter: Actually, this was fixed in C++20, so signed shifts are now well defined.

MBkkt: Good, but still calling memset with runtime size doesn't look good for me

goedderz: > Also any arguments smaller than int implicitly casted to the int.
> why not simplify it?

It's written this way for the reader, so the intent is clear. The compiler doesn't care either way.

> Also it looks like !slice.isUint() in code above unnecessary.

I don't know what you mean. There's no `!slice.isUint()` in my PR, only in your suggestion. And there it's strictly necessary.

> Good, but still calling memset with runtime (for compiler) size doesn't look good for me

Sure, I've removed the call to memset.

MBkkt: > I don't know what you mean. There's no !slice.isUint() in my PR, only in your suggestion. And there it's strictly necessary

Yeah you're right.

MBkkt (approved):
I really don't like shift for signed char, but it works

LGTM

arangod/Aql/AqlValue.cpp
MBkkt: > It's written this way for the reader, so the intent is clear. The compiler doesn't care either way.

Well my concerns about it's pretty tricky.

That we have int(0x00) in unsigned part of condition, and int(0x00) or int(0xFFFFFFFF), not 0xFF, in other

jsteemann (approved):
LGTM

arangod/Aql/AqlValue.cpp
goedderz: I understand. I've changed it to a branch.

KVS85:
```
D:\win02-colo2\oskar\work\ArangoDB\arangod\Aql\AqlValue.cpp(1694,1): error C2220: the following warning is treated as an error [D:\win02-colo2\oskar\work\ArangoDB\build\arangod\Aql\arango_aql.vcxproj]
D:\win02-colo2\oskar\work\ArangoDB\arangod\Aql\AqlValue.cpp(1694,1): warning C4018: '>': signed/unsigned mismatch [D:\win02-colo2\oskar\work\ArangoDB\build\arangod\Aql\arango_aql.vcxproj]
```

mpoeter (approved):

PR #19450, opened 2023-07-20
Remove the special case for a single agent instance.

maierlars:
### Scope & Purpose
Previously there was special handling in place when there is only a single leader in the cluster. Back then this was necessary because of reasons. Those reasons are no longer valid today and the `n > 1` case generalizes to `n = 1` as well.

(I also added a changelog entry for removing ttl and observe. This was missing)

However, by doing that I found that our `agency-restart` tests only ever tested the `n = 1` special case and never tested the production case. Of course there was a bug and I fixed it as well: when the `log` collection is missing a test asserted that everything will be reset to the empty state. But when starting the agent looked at the latest index it found in the `compact` collection. This was 4500. And then waited for the commitIndex (which could only ever be 0, because the log was empty) to reach 4500.

Furthermore, the agency-restart tests now have to wait for the agency to be ready. Sadly the test framework didn't offer an easy solution, so I had to add this to all restart tests.

### DIFF
--- a/arangod/Agency/Constituent.cpp
+++ b/arangod/Agency/Constituent.cpp
@@ -657,10 +657,11 @@ void Constituent::run() {
       int64_t randTimeout = RandomGenerator::interval(a, b);
       int64_t randWait = randTimeout;
 
-      // Speedup for size() == 1 - do not wait at all. Candidate immediately.
+      // Speedup for size() == 1 - do not wait at all. Start election
+      // immediately.
       if (size() == 1) {
         LOG_TOPIC("31ed7", INFO, Logger::AGENCY)
-            << "Immediately candidate because of single agency";
+            << "Immediately start election because of single agency";
         randWait = 0;
       }
 


jsteemann (approved):
LGTM

arangod/Agency/Constituent.cpp
```diff
@@ -640,164 +634,162 @@ void Constituent::run() {
...
+                                       _agent->config().timeoutMult());
+      int64_t b = static_cast<int64_t>(M * _agent->config().maxPing() *
+                                       _agent->config().timeoutMult());
+      int64_t randTimeout = RandomGenerator::interval(a, b);
+      int64_t randWait = randTimeout;
+
+      // Speedup for size() == 1 - do not wait at all. Candidate immediately.
+      if (size() == 1) {
+        LOG_TOPIC("31ed7", INFO, Logger::AGENCY)
+            << "Immediately candidate because of single agency";
```
jsteemann, line 663: nit: I guess "to candidate" is a rather literal translation. Should rather be "stand(ing) as candidate"

jsteemann (approved):
LGTM

arangod/Agency/Constituent.cpp
maierlars:
```suggestion
      // Speedup for size() == 1 - do not wait at all. Start election immediately.
      if (size() == 1) {
        LOG_TOPIC("31ed7", INFO, Logger::AGENCY)
            << "Immediately start election because of single agency";
```

PR #19465, opened 2023-07-24
Improvements for in-memory cache subsystem

jsteemann:
### Scope & Purpose

* Changed default values for the following options related to the in-memory
  cache subsystem:

  - `--cache.ideal-lower-fill-ratio` changed from 0.04 to 0.08
  - `--cache.ideal-upper-fill-ratio` changed from 0.25 to 0.33

  This allows slightly better hash table utilization and earlier memory
  reclamation.

* Added metric `rocksdb_cache_edge_empty_inserts_total` to count the number
  of inserts into the edge cache for non-connected edges.

* Renamed two edge-cache related metrics to improve naming consistency:
  - `rocksdb_cache_edge_effective_entries_size` was renamed to 
    `rocksdb_cache_edge_inserts_effective_entries_size_total` and was changed
    from a gauge to a counter.
  - `rocksdb_cache_edge_uncompressed_entries_size` was renamed to 
    `rocksdb_cache_edge_inserts_uncompressed_entries_size_total` and is now
    also a counter instead of a gauge.

* Do not auto-reload entries into in-memory edge cache for edge index entries
  that were previously not contained in the in-memory edge cache. This change
  will help to keep the hot set in memory rather than evicting it in favor of
  "random" other index entries.

* Properly track memory usage for allocated objects in the in-memory cache
  (e.g. the edge cache or the in-memory cache for other persistent indexes).
  Previously the memory used for the underlying hash tables was accounted for
  correctly, but the sizes of the cache payloads (keys and values) were not
  accounted for under all circumstances (at least for the initial entries in
  the caches).
  This change leads to more accurate memory usage tracking and reporting by the 
  in-memory cache subsystem, and to the cache subsystem not exceeding its 
  configured memory usage limit. 
  The cache subsystem was also changed so that it can use as much memory as
  configured by the global cache memory limit (configurable via startup option
  `--cache.size`). Previously the cache subsystem was freeing memory as soon
  as it hit 56% of the configured limit.
  Overall, the effective memory usage of the cache subsystem will be very
  different to the memory usage in previous versions. In previous versions the
  configured memory usage limit could be temporarily exceeded, but in most
  cases the cache used considerably less memory than allowed by the limit.
  Now the cache will try to use up to as much memory as allowed by the
  configured memory usage limit.
  If that leads to an overall overallocation of resources (i.e. too high total
  memory of the system), it may be necessary to lower the value of the startup
  option `--cache.size`, so that the cache memory usage is decreased.

* Reduce memory usage for empty in-memory edge caches by ~40%. This is achieved
  by allocating each cache's statistics objects only lazily, when actually 
  needed.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,34 @@
...
+
+* Properly track memory usage for allocated objects in the in-memory cache
+  (e.g. the edge cache or the in-memory cache for other persistent indexes).
+  Previously only the memory used for the underlying hash tables was accounted
+  for, but not the value of the cache payloads (keys and values).
+  This change leads to more accurate memory usage tracking and reporting by the 
+  in-memory cache subsystem, and the subsystem not exceeding its configured
+  limit. However, as memory usage by the in-memory cache subsystem was 
+  previously (inaccurately) underreported, it also means that the limit will 
+  now behit more quickly. It may be necessary to increase the value of the
```
jsteemann, line 25:
```suggestion
  now be hit more quickly. It may be necessary to increase the value of the
```

arangod/Cache/Cache.cpp
```diff
@@ -41,6 +41,7 @@
 #include "Cache/PlainCache.h"
 #include "Cache/Table.h"
 #include "Cache/TransactionalCache.h"
+#include "Logger/LogMacros.h"
```
jsteemann, line 44:
```suggestion
```

Documentation/Metrics/rocksdb_cache_edge_inserts_effective_entries_size_total.yaml
```diff
@@ -1,20 +1,22 @@
...
   - dbserver
   - agent
   - single
 description: |
-  Total effective memory size of all edge cache data that were ever stored in
-  the edge cache of any collection/shard. The size is calculated after any
-  potential compression, so the compression efficiency can be estimated by 
-  this metric by the size of the uncompressed edge cache entries.
+  Total effective memory size of all edge cache data that were stored in the
+  the edge cache of any collection/shard since the server start.
```
maierlars, line 16: NIT counters are not reset (with the help of prometheus) so this is actually _since the server initially started_.

maierlars (approved):

arangod/RocksDBEngine/RocksDBIndex.cpp
```diff
@@ -194,11 +189,15 @@ void RocksDBIndex::setupCache() {
 
 void RocksDBIndex::destroyCache() {
```
MBkkt, line 190: NIT
this function noexcept now

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
```diff
@@ -647,6 +651,9 @@ class RocksDBEdgeIndexLookupIterator final : public IndexIterator {
     _totalCachedSizeEffective +=
         cache::kCachedValueHeaderSize + key.size() + size;
     ++_totalInserts;
+    if (isEmpty) {
+      ++_totalEmptyInserts;
+    }
```
MBkkt, line 656:
```suggestion
    _totalEmptyInserts += static_cast<uint64_t>(isEmpty);
```
[godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:8,positionColumn:2,positionLineNumber:8,selectionStartColumn:2,selectionStartLineNumber:8,startColumn:2,startLineNumber:8),source:'%23include+%3Calgorithm%3E%0A%23include+%3Ccstdint%3E%0A%23include+%3Ccstring%3E%0A%0Avoid+foo1(bool+isEmpty,+uint64_t%26+_totalEmptyInserts)+%7B%0A++++if+(isEmpty)+%7B%0A++++++++%2B%2B_totalEmptyInserts%3B%0A++++%7D%0A%7D%0A%0Avoid+foo2(bool+isEmpty,+uint64_t%26+_totalEmptyInserts)+%7B%0A++++_totalEmptyInserts+%2B%3D+static_cast%3Cuint64_t%3E(isEmpty)%3B%0A%7D%0A%0Aint+main()+%7B%7D'),l:'5',n:'0',o:'C%2B%2B+source+%232',t:'0')),header:(),k:46.206174777603366,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g131,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3+-std%3Dc%2B%2B20+-DNDEBUG+-std%3Dc%2B%2B20++-Wall+-fsized-deallocation+-mfxsr+-mmmx+-msse+-msse2+-mcx16+-msahf+-mpopcnt+-msse3+-msse4.1+-msse4.2+-mssse3+-mpclmul+-mavx+-mxsave+-maes',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:2),l:'5',n:'0',o:'+x86-64+gcc+13.1+(Editor+%232)',t:'0')),header:(),k:39.49420079618994,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.1',editorid:2,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+13.1+(Compiler+%231)',t:'0')),k:14.299624426206703,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

arangod/Cache/Cache.h
```diff
@@ -251,21 +266,52 @@ class Cache : public std::enable_shared_from_this<Cache> {
...
+  // are created
+  std::atomic<bool> _haveFindStats;
+
+  bool const _enableWindowedStats;
+
+  // used to create eviction stats under the lock (so that only a single
+  // thread creates them)
+  std::mutex _findStatsLock;
+  // this struct is allocated lazily, when the _findStats are first written to
+  std::unique_ptr<FindStats> _findStats;
```
MBkkt, line 310: https://github.com/abseil/abseil-cpp/blob/master/absl/base/call_once.h
We could use this to decrease sizeof and speedup it.

arangod/Cache/Cache.h
```diff
@@ -274,18 +320,30 @@ class Cache : public std::enable_shared_from_this<Cache> {
...
+
+  // this will only flip from false to true, when the eviction stats
+  // are created
+  std::atomic<bool> _haveEvictionStats;
+  // used to create eviction stats under the lock (so that only a single
+  // thread creates them)
+  std::mutex _evictionStatsLock;
+  // this struct is allocated lazily, when the _evictionStats are first written
+  // to
+  std::unique_ptr<EvictionStats> _evictionStats;
```
MBkkt, line 336: https://github.com/abseil/abseil-cpp/blob/master/absl/base/call_once.h
We could use this to decrease sizeof and speedup it.

arangod/Cache/FrequencyBuffer.h
```diff
@@ -122,18 +109,21 @@ class FrequencyBuffer {
...
     }
-    std::sort(data.begin(), data.end(),
-              [](std::pair<T, std::uint64_t> const& left,
-                 std::pair<T, std::size_t> const& right) {
-                return left.second < right.second;
-              });
+    std::sort(
+        data.begin(), data.end(),
+        [](std::pair<T, std::uint64_t> const& left,
+           std::pair<T, std::size_t> const& right) {
```
MBkkt, line 115: Well, It's definitely doesn't look good. It probably should be
```suggestion
        [](std::pair<T, std::uint64_t> const& left,
           std::pair<T, std::uint64_t> const& right) {
```

arangod/Cache/FrequencyBuffer.h
```diff
@@ -122,18 +109,21 @@ class FrequencyBuffer {
     for (auto f : frequencies) {
       data.emplace_back(std::pair<T, std::size_t>(f.first, f.second));
```
MBkkt, line 110: size_t incorrect here
```suggestion
      data.emplace_back(f.first, f.second);
```

arangod/Cache/FrequencyBuffer.h
```diff
@@ -36,82 +36,69 @@
...
   }
 
-  //////////////////////////////////////////////////////////////////////////////
   /// @brief Return a list of (event, count) pairs for each recorded event in
   /// ascending order.
-  //////////////////////////////////////////////////////////////////////////////
   typename FrequencyBuffer::stats_t getFrequencies() const {
+    T const empty{};
     // calculate frequencies
     std::unordered_map<T, std::size_t, Hasher, Comparator> frequencies;
```
MBkkt, line 98:
```suggestion
    std::unordered_map<T, std::uint64_t, Hasher, Comparator> frequencies;
```

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1934,17 +1934,17 @@ void RocksDBCollection::destroyCache() const {
 void RocksDBCollection::invalidateCacheEntry(RocksDBKey const& k) const {
   if (useCache()) {
     TRI_ASSERT(_cache != nullptr);
-    bool banished = false;
-    while (!banished) {
+    do {
```
MBkkt, line 1937: Why did you use `do while` instead of just `while`?

MBkkt (changes requested):

arangod/Cache/Cache.h
jsteemann: I understand that absl::call_once() would help us in the majority of cases, but it seems it does not provide a way to query the status of the `once_flag` from the outside.

This would be necessary in `Cache::hitRates()`, where we check if we have created findStats already, but do **not** want to create them if not. IMO `absl::call_once()` does not allow this.

arangod/RocksDBEngine/RocksDBCollection.cpp
jsteemann: * to get rid of the `banished` variable
* to make the intent more clear: we always want to execute `banish()` at least once

arangod/RocksDBEngine/RocksDBIndex.cpp
jsteemann: Added.

arangod/RocksDBEngine/RocksDBEdgeIndex.cpp
jsteemann: Thanks for the suggestion. Added.

arangod/Cache/Cache.h
jsteemann: See above.

arangod/Cache/FrequencyBuffer.h
jsteemann: Good catch. I actually did not change this part, but I also didn't see it.
Fixed.

arangod/Cache/FrequencyBuffer.h
jsteemann: Changed.

arangod/Cache/FrequencyBuffer.h
jsteemann: Changed.

arangod/Cache/Cache.h
MBkkt: Well exists easy way to fix it.
Add additional atomic_bool  flag which is set in initial callback.
And make atomic pointer instead of unique_ptr

Second is less convenient. Anyway, both provide better sizeof and better performance than current version

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: You write do while true, so my question is why don't use just while true?

MBkkt: Also I forgot to put NIT, it's not really important for me, so you could just resolve it

It just looks like you start with "do while some condition", and then removed condition, so it looks obsolete for me.
But in general it doesn't matter

arangod/Cache/Cache.h
jsteemann: I doubt that there is a major performance difference when moving to any of the versions you suggest. Simply because in the general case we only check the value of an atomic variable, which is only once flipped from `false` to `true`. The cost of lazily creating the stats object is there anyway, and it should not matter if we have an extra mutex here or an extra check inside the mutex.
The sizeof values will be better though, but we are talking about a few bytes of out several hundred.

The atomic pointer solution would work, but it has the disadvantage that when multiple threads discover that the pointer is not set, only one of them should create the object. So it needs extra synchronization there.

MBkkt: Well suggestion with bool simpler, a little but faster, and have smaller sizeof.

So the question why do we need solution with mutexes?

About atomic pointer I meant different.
Pointer could be used instead of separate bool flag in any solution, so it's just about smaller sizeof and a little more complicated code.


About sizeof, you first who start to optimize it in this PR, also will be good to reorder fields in this class, to make less padding bytes

arangod/Cache/Cache.cpp
```diff
@@ -82,15 +80,15 @@ Cache::~Cache() {
...
     TRI_ASSERT(_findStats != nullptr);
     memoryUsage += sizeof(decltype(_findStats)::element_type);
     if (_findStats->findStats) {
       memoryUsage += _findStats->findStats->memoryUsage();
     }
   }
 
-  if (_haveEvictionStats.load(std::memory_order_relaxed)) {
+  if (_evictionStatsCreated.load(std::memory_order_relaxed)) {
     TRI_ASSERT(_evictionStats != nullptr);
```
MBkkt: Needs to remove assert or make load(acq)

arangod/Cache/Cache.cpp
```diff
@@ -82,15 +80,15 @@ Cache::~Cache() {
   // eviction stats
   std::uint64_t memoryUsage = 0;
 
-  if (_haveFindStats.load(std::memory_order_relaxed)) {
```
MBkkt: should be load(acq) because access to the data

arangod/Cache/Cache.cpp
```diff
@@ -185,7 +183,9 @@ std::pair<double, double> Cache::hitRates() {
   double lifetimeRate = std::nan("");
   double windowedRate = std::nan("");
 
-  if (_haveFindStats.load(std::memory_order_relaxed)) {
+  if (_findStatsCreated.load(std::memory_order_relaxed)) {
```
MBkkt, line 186: should be load(acq) to access stats

arangod/Cache/Cache.cpp
```diff
@@ -401,45 +405,36 @@ bool Cache::reportInsert(bool hadEviction) {
...
+    if (_enableWindowedStats) {
+      _findStats->findStats = std::make_unique<StatBuffer>(
+          _manager->sharedPRNG(), Manager::kFindStatsCapacity);
     }
-  }
+    _manager->adjustGlobalAllocation(
+        sizeof(decltype(_findStats)::element_type) +
+        (_enableWindowedStats ? _findStats->findStats->memoryUsage() : 0));
+
+    _findStatsCreated.store(true);
```
MBkkt, line 421: release is enough.
I prefer release other seq_cst, because it better for x86.
For aarch64 is no difference

arangod/Cache/Cache.cpp
```diff
@@ -401,45 +405,36 @@ bool Cache::reportInsert(bool hadEviction) {
...
-  }
+  absl::call_once(_evictionStatsOnceFlag, [this]() {
+    TRI_ASSERT(!_evictionStatsCreated.load(std::memory_order_relaxed));
+    TRI_ASSERT(_evictionStats == nullptr);
+
+    _evictionStats = std::make_unique<EvictionStats>();
+    _manager->adjustGlobalAllocation(
+        sizeof(decltype(_evictionStats)::element_type));
+
+    _evictionStatsCreated.store(true);
```
MBkkt, line 435: relaxed is enough if you will remove assert.
Otherwise release is enough

MBkkt (changes requested):

arangod/Cache/Cache.h
jsteemann: I was trying to optimize things that matter most, which is not the sizeof the Cache object, but the size of the dynamic memory it allocates.

MBkkt: > Reduce memory usage for empty in-memory edge caches by ~40%. This is achieved
by allocating each cache's statistics objects only lazily, when actually
needed.

MBkkt: Well, in such case I couldn't understand this :(

jsteemann:
Sanitizer run also showed no newly introduced errors:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/31027/
There were 2 test failures in the single server TSan run, but these are timing-related and totally unrelated to the changes in this PR.

PR #19470, opened 2023-07-26
FE-313: upgrade arangojs in web UI to 8.x

palashkaria:
### Scope & Purpose

This upgrades the `arangojs` version used by the web UI to 8.x

- Fixed imports
- Fixed type issues
- Used `db.route()` instead of `db.request()`
- Followed the [migration guide](https://arangodb.github.io/arangojs/MIGRATING)

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-313

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -116,7 +116,6 @@ devel
   the maximum number of file descriptors the ArangoDB process can open. 
   Thus the option should only be enabled on deployments with a limited number
   of collections/shards/indexes.
->>>>>>> devel
 
 * BTS-1511: AQL: Fixed access of integers in the ranges
   [-36028797018963968, -281474976710657] and


js/apps/system/_admin/aardvark/APP/react/src/views/analyzers/addAnalyzer/AddAnalyzerForm.tsx
```diff
@@ -17,12 +17,16 @@ import { useAnalyzersContext } from "../AnalyzersContext";
...
-const ANALYZER_TYPE_OPTIONS = Object.keys(TYPE_TO_LABEL_MAP).map(type => {
-  return {
-    value: type,
-    label: TYPE_TO_LABEL_MAP[type as AnalyzerTypes]
-  };
-});
+const ANALYZER_TYPE_OPTIONS = Object.keys(TYPE_TO_LABEL_MAP)
+  .map(type => {
+    const excludedTypes = ["minhash"];
+    if (excludedTypes.includes(type)) return null;
```
palashkaria, line 23: excluding for now, will add support in the next PR, there's also a separate task for the same: [FE-314](https://arangodb.atlassian.net/browse/FE-314)

palashkaria:
> * The PR includes some changes (types etc) related to minhash being added. Did this slip in by accident? This isn't a showstopper as it doesn't change anything and as you said another PR adds the actual functionality but I wasn't sure if this is intentional.

Yes, types changes are intentional to fix some TypeScript issues. It can be done by marking as `any` in a few places too, but I just went ahead and did the correct changes - the minHash PR builds on top of these.

> * We should move away from `db.route.request` to the actual API implementations where possible. This is out of scope for the upgrade itself but it would allow us to more tightly dogfood the driver by making sure the type signatures are what we expect.

Agreed, I think the current implementation is this way due to some issues in the driver (which may be fixed already). We can take up moving to the drive impl as a subsequent task

palashkaria:
Thanks @pluma4345 


> * The PR includes some changes (types etc) related to minhash being added. Did this slip in by accident? This isn't a showstopper as it doesn't change anything and as you said another PR adds the actual functionality but I wasn't sure if this is intentional.

Yes, the type changes are intentional to fix some TypeScript issues. It can be done by marking as `any` in a few places too, but I just went ahead and did the correct changes - the minHash PR builds on top of these.

> * We should move away from `db.route.request` to the actual API implementations where possible. This is out of scope for the upgrade itself but it would allow us to more tightly dogfood the driver by making sure the type signatures are what we expect.

Agreed, I think the current implementation is this way due to some issues in the driver (which may be fixed already). We can take up moving to the drive impl as a subsequent task

palashkaria:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-ANY-jslint/30593/

PR #19472, opened 2023-07-27
optionally allow partitioning of .sst files

jsteemann:
### Scope & Purpose

Optionally allow partitioning of .sst files for different collections/shards and indexes.
This is a potential performance improvement which allows RocksDB compactions to be more efficient.
We don't have performance test results yet, so it is only a _potential_ improvement that we need to try out.

* Added experimental startup options for RocksDB .sst file partitioning:

  - `--rocksdb.partition-files-for-documents`
  - `--rocksdb.partition-files-for-primary-index`
  - `--rocksdb.partition-files-for-edge-index`
  - `--rocksdb.partition-files-for-persistent-index`

  Enabling any of these options will make RocksDB's compaction write the 
  data for different collections/shards/indexes into different .sst files. 
  Otherwise the document data from different collections/shards/indexes 
  can be mixed and written into the same .sst files.

  Enabling these options usually has the benefit of making the RocksDB
  compaction more efficient when a lot of different collections/shards/indexes
  are written to in parallel.
  The disavantage of enabling this option is that there can be more .sst
  files than when the option is turned off, and the disk space used by
  these .sst files can be higher than if there are fewer .sst files (this
  is because there is some per-.sst file overhead).
  In particular on deployments with many collections/shards/indexes this can 
  lead to a very high number of .sst files, with the potential of outgrowing 
  the maximum number of file descriptors the ArangoDB process can open. 
  Thus the option should only be enabled on deployments with a limited number
  of collections/shards/indexes.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19476
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,31 @@
 devel
 -----
 
+* Added experimental startup options for RocksDB .sst file partitioning:
+
+  - `--rocksdb.partition-files-for-documents`
+  - `--rocksdb.partition-files-for-primary-index`
+  - `--rocksdb.partition-files-for-edge-index`
+  - `--rocksdb.partition-files-for-persistent-index`
+
+  Enabling any of these options will make RocksDB's compaction write the 
+  data for different collections/shards/indexes into different .sst files. 
+  Otherwise the document data from different collections/shards/indexes 
+  can be mixed and written into the same .sst files.
+
+  Enabling these options usually has the benefit of making the RocksDB
+  compaction more efficient when a lot of different collections/shards/indexes
+  are written to in parallel.
+  The disavantage of enabling this option is that there can be more .sst
+  files than when the option is turned off, and the disk space used by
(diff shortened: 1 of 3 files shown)


MBkkt (approved):
Interesting, it makes it more inverted index style.

It should make hotbackup potentially a lot of smaller because only files that changed will be duplicated in the future

markuspf (approved):
LGTM, let's find out what the performance implications are!

PR #19485, opened 2023-07-28
add diagnostic information when incosistency is found

jsteemann:
### Scope & Purpose

Add diagnostic information when incosistency is found between primary index and documents cf. Also validate document existence and lock document keys properly when isRestore=true and we are in maintainer mode.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBCollection.cpp
+++ b/arangod/RocksDBEngine/RocksDBCollection.cpp
@@ -1612,6 +1612,9 @@ Result RocksDBCollection::modifyDocument(
     velocypack::Slice oldDoc, LocalDocumentId newDocumentId,
     velocypack::Slice newDoc, RevisionId oldRevisionId,
     RevisionId newRevisionId, OperationOptions const& options) const {
+  TRI_ASSERT(oldDoc.get(StaticStrings::KeyString).stringView() ==
+             newDoc.get(StaticStrings::KeyString).stringView());
+
   savepoint.prepareOperation(newRevisionId);
 
   // Coordinator doesn't know index internals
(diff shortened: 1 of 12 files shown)


arangod/RocksDBEngine/RocksDBKey.cpp
```diff
@@ -29,6 +29,8 @@
 #include "RocksDBEngine/RocksDBFormat.h"
 #include "RocksDBEngine/RocksDBTypes.h"
 
+#include <iostream>
```
neunhoef, line 32: Needed?

arangod/RocksDBEngine/RocksDBKey.h
```diff
@@ -31,6 +31,7 @@
 #include "VocBase/voc-types.h"
 #include "Zkd/ZkdHelper.h"
 
+#include <iosfwd>
```
neunhoef, line 34: Needed?

neunhoef (approved):
LGTM. See two comments, two include statements seem to be superfluous.

arangod/RocksDBEngine/RocksDBKey.h
jsteemann: I think so, as the file is implementing `operator<<` to append to an `std::ostream`.
It already did so before, but the required headers were missing.

arangod/RocksDBEngine/RocksDBKey.cpp
jsteemann: I think so, as the file is implementing `operator<<` to append to an `std::ostream`.
It already did so before, but the required headers were missing.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1405,11 +1417,27 @@ Result RocksDBCollection::insertDocument(transaction::Methods* trx,
...
+  rocksdb::Status s;
+#ifdef ARANGODB_ENABLE_MAINTAINER_MODE
+  // use Put() instead of PutUntracked()
+  if (options.isRestore) {
+    s = mthds->Put(RocksDBColumnFamilyManager::get(
+                       RocksDBColumnFamilyManager::Family::Documents),
+                   key.ref(),
+                   rocksdb::Slice(doc.startAs<char>(),
+                                  static_cast<size_t>(doc.byteSize())),
+                   /*assume_tracked*/ false);
```
mpoeter, line 1429: `Put` will only check for conflicts with concurrent operations, but would still silently overwrite any already existing key. We should instead use a `GetForUpdate` to check for an existing key and lock it. Then we can even stick with `PutUntracked`.

mpoeter (changes requested):

arangod/RocksDBEngine/RocksDBCollection.cpp
jsteemann: Addressed.

mpoeter (approved):

PR #19491, opened 2023-07-29
FE-94: Support for query options and optimizer rules on web UI

palashkaria:
### Scope & Purpose

Adds support for passing query options & disabling optimizer rules from the Query UI

- query call happens directly, via the `cursor` endpoint 
- explain/profile calls happen via `aardvark.js` wrapper APIs, where `options` are passed on to the db
- option JSON contains disabled rules, and can be modified directly

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-94


js/apps/system/_admin/aardvark/APP/react/src/views/query/editor/QueryOptionsTab.tsx
```diff
@@ -0,0 +1,317 @@
...
+          if (rule.flags.clusterOnly && !isCluster) {
+            return false;
+          }
+          if (!rule.flags.canBeDisabled) {
+            return false;
+          }
+          if (rule.flags.hidden) {
+            return false;
+          }
+          return true;
```
shd8, line 284: It could be done in a single sentence:

`!(rule.flags.enterpriseOnly && !isEnterprise || rule.flags.clusterOnly && !isCluster || !rule.flags.canBeDisabled || rule.flags.hidden)`

The only inconvenient is that loses some readability. Just as a suggestion, up to you to change it.

js/apps/system/_admin/aardvark/APP/react/src/views/query/editor/QueryOptionsTab.tsx
```diff
@@ -0,0 +1,317 @@
...
+  return (
+    <Grid templateColumns="repeat(2, 1fr)" gap="2" alignItems="center">
+      <FormLabel>Disabled Rules</FormLabel>
+      <MultiSelect
+        options={ruleOptions}
+        value={ruleOptions.filter(option => {
+          return disabledRules.includes(option.value);
+        })}
+        onChange={value => {
+          console.log(value);
```
shd8, line 310: Is this `console.log` needed?

shd8 (approved):
LGTM after addressing a couple little comments.

js/apps/system/_admin/aardvark/APP/react/src/views/query/editor/QueryOptionsTab.tsx
palashkaria: I'd suggest always choosing readability in cases like these, instead of doing it in a single line. Would prefer to keep it this way, and I have further made it more readable by pulling out the flags early, so we don't have to keep doing `rules.flags.{flag}`

PR #19498, opened 2023-07-31
Fix HotBackupConsistencyTest.py 

markuspf:
### Scope & Purpose

Import `ServerConnectionError` exception and add a catch-all exception branch


maierlars (approved):

jsteemann (approved):
LGTM

PR #19499, opened 2023-07-31
[BTS-1264] Report a warning if an unknown ellipsoid is used.

maierlars:
### Scope & Purpose
Before this patch we silently fell back to `sphere`. Now we generate an aql error.

### DIFF
--- a/arangod/Aql/Functions.cpp
+++ b/arangod/Aql/Functions.cpp
@@ -6022,7 +6022,7 @@ geo::Ellipsoid const* detEllipsoid(ExpressionContext* expressionContext,
                    "invalid ellipsoid {} specified, falling back to \"sphere\"",
                    slice.stringView())));
   } else {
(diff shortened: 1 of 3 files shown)


arangod/Aql/Functions.cpp
```diff
@@ -6007,6 +6007,30 @@ AqlValue functions::Distance(ExpressionContext* expressionContext,
...
+      return ell;
+    }
+    registerWarning(
+        expressionContext, functionName,
+        Result(TRI_ERROR_BAD_PARAMETER,
+               fmt::format(
+                   "invalid ellipsoid {} specified, falling back to \"sphere\"",
+                   slice.stringView())));
+  } else {
+    registerWarning(expressionContext, "GEO_AREA",
```
jsteemann, line 6025:
```suggestion
    registerWarning(expressionContext, functionName,
```

jsteemann (approved):
LGTM

PR #19502, opened 2023-08-01
Fix nightly issue merkel trees

maierlars:
### Scope & Purpose

Merkeltrees can be equal while still having a different range of keys.

### DIFF
--- a/tests/js/client/chaos/test-chaos-load-common.inc
+++ b/tests/js/client/chaos/test-chaos-load-common.inc
@@ -60,6 +60,11 @@ const fetchRevisionTree = (serverUrl, shardId) => {
   return JSON.parse(result.body);
 };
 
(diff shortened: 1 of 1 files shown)


tests/js/client/chaos/test-chaos-load-common.inc
```diff
@@ -91,7 +96,7 @@ const checkCollectionConsistency = (cn) => {
           failed = true;
         }
 
-        if (!_.isEqual(leaderTree, followerTree)) {
+        if (!compareTree(leaderTree.computed, followerTree.computed)) {
```
maierlars, line 99: Trees are already tested to be locally equal above.

mpoeter (approved):

jsteemann (approved):
LGTM

PR #19511, opened 2023-08-01
[BTS-1328] Send server state asynchronously.

maierlars:
### Scope & Purpose

The heartbeat thread used synchronous code to send the heartbeat to the agency. This code used a timeout of 20s. What was suggested in the ticket is to send an async request every 5s, potentially on a new connection. The number of concurrent requests is thus limited to 5.


arangod/Agency/AsyncAgencyComm.cpp
```diff
@@ -676,6 +677,38 @@ AsyncAgencyComm::FutureResult AsyncAgencyComm::deleteKey(
...
+    }
+  }
+
+  RequestMeta meta;
+  meta.requestId = _manager.nextRequestId();
+  meta.method = RestVerb::Post;
+  meta.type = RequestType::CUSTOM;
+  meta.url = AGENCY_URL_TRANSIENT;
+  meta.timeout = timeout;
+  meta.skipScheduler = false;
```
jsteemann, line 706: Do we want to use `skipScheduler = false` here? It seems to be the default value too, but do we really want to post the result on the scheduler for resolving?

jsteemann (approved):
LGTM

arangod/Agency/AsyncAgencyComm.cpp
maierlars: probably more work to post it, than actually executing the code 🤔 I will change it.

maierlars: wait a minute. Although this function is currently used in only one place, we don't know in general what is executed when the future is resolved. From that point of view I would argue for posting it on the scheduler.

neunhoef: I see both points. How about the following: We could make `skipScheduler` an explicit argument of the method. Then one **has** to specify it at the call site and there one can judge if the attached `then` closures merit executing on the scheduler?

neunhoef (approved):
LGTM. I think this is a sensible change which should improve reliability.

neunhoef (approved):
LGTM

PR #19514, opened 2023-08-02
[3.10] Backport HotBackup Improvements

markuspf:
### Scope & Purpose

Backport of HotBackup Improvements to 3.10.


arangod/Transaction/Manager.cpp
```diff
@@ -105,11 +104,8 @@ void Manager::registerTransaction(TransactionId transactionId,
   if (!isReadOnlyTransaction && !isFollowerTransaction) {
     LOG_TOPIC("ccdea", TRACE, Logger::TRANSACTIONS)
         << "Acquiring read lock for tid " << transactionId.id();
-    _rwLock.lockRead();
-    _nrReadLocked.fetch_add(1, std::memory_order_relaxed);
     LOG_TOPIC("ccdeb", TRACE, Logger::TRANSACTIONS)
-        << "Got read lock for tid " << transactionId.id()
-        << " nrReadLocked: " << _nrReadLocked.load(std::memory_order_relaxed);
+        << "Got read lock for tid " << transactionId.id();
```
jsteemann, line 108: This code is identical to the code in devel (which should be the source of truth), but effectively the `if` condition now only guards two log messages. There is no lock acquired here anymore, so the log messages are IMO misleading and can be removed.

KVS85:
- [ ] https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/31094/

PR #19521, opened 2023-08-03
BTS-1539: backport a change from devel

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19515

https://arangodb.atlassian.net/browse/BTS-1539

Backport a change from devel that relaxes a comparison made in the cluster maintenance for computed values. 
Not having computed values is stored as `null` normally, but under some condition not having computed values can also be stored as an empty array in devel. 
Both values should be treated equal by the maintenance. If the maintenance treats `null` and `[]` as different, it may initiate a busy loop, trying forever to adjust the `computedValues` property of the shard.

Although we believe that in 3.11 having no computed values is always stoted as `null`, this backport is made to avoid potential issues during upgrades or in case other changes are backported from devel. the added test should run into an issue in case the problem from devel occurs.

So this is not a bugfix, but a precautionary backport. Thus intentionally no CHANGELOG entry.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1539
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/Maintenance.cpp
+++ b/arangod/Cluster/Maintenance.cpp
@@ -117,7 +117,6 @@ static std::shared_ptr<VPackBuilder> compareRelevantProps(
       if (planned.isNone()) {
         continue;
       }
-
       bool isSame = true;
       // Register any change
       if (property == StaticStrings::Schema) {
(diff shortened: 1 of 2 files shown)


maierlars (approved):

neunhoef (approved):
LGTM

PR #19530, opened 2023-08-07
fix wrong underlying parameter type

jsteemann:
### Scope & Purpose

for `--rocksdb.recycle-log-file-num`. The option does not make any difference to use with the default WAL recovery mode we are using. RocksDB automatically sanitizes the option to a value of `0`.

Previously we exposed this startup option as a boolean value, but it should rather be a size_t. Interestingly, the RocksDB code also uses the option with a wrong type in at least one place of their code.

This change should not modify the behavior in any way, except if the option is now set to a non-zero value and non-default recovery mode is selected.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBOptionFeature.cpp
+++ b/arangod/RocksDBEngine/RocksDBOptionFeature.cpp
@@ -273,6 +273,7 @@ RocksDBOptionFeature::RocksDBOptionFeature(Server& server)
       // note: this is a default value from RocksDB (db/column_family.cc,
       // kAdjustedTtl):
       _periodicCompactionTtl(30 * 24 * 60 * 60),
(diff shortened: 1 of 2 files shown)


Dronplane (approved):
LGTM. But should we at least mention such change in the CHANGELOG as this looks like a breaking change for someone who was using non-default recovery mode?

neunhoef (approved):
LGTM

PR #19531, opened 2023-08-07
Index Background Creation Bug

maierlars:
### Scope & Purpose
Do not skip documents that have been inserted, when removing in same transaction.

When indexes are created in background a temporary IndexBuilder index is inserted as placeholder for the actual index. This index records which documents are inserted and removed during index creation. However, we limited this recording to a single event per document per transaction.

If the same document is inserted and removed within the same transaction during an index background creation, the document will eventually show up in the index.

Furthermore, the comment "don't track documents more than once" is a bit misleading. Since we use the local document id to track documents, any such id will show up in insert at most once.

### DIFF
--- a/arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
+++ b/arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
@@ -67,11 +67,6 @@ using namespace arangodb::rocksutils;
 
 namespace {
 
-struct BuilderCookie : public arangodb::TransactionState::Cookie {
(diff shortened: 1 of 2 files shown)


MBkkt:
Should this be backported?

MBkkt:
@maierlars 
Should this be backported?

maierlars:
First I would wait a few weeks to see if this does not have any unintended side effects.

Additionally, you have write the same document twice in the same transaction. That is an unlikely use case.

PR #19538, opened 2023-08-08
Clarify log message when registering transaction

markuspf:
### Scope & Purpose

Clarify a log message that had gotten useless with recent HotBackup changes.


jsteemann:
We can either use this PR to change to log message, or remove the log message altogether, as it provides little value.
I am fine with either solution.

jsteemann (approved):

markuspf:
I considered removing the log message altogether, too, and then just thought that it might be useful to know that this `registerTransaction` method was called.

seeing as the `unregisterTransaction` method doesn't have a log message in it either, maybe removing it altogether is the better option.

PR #19539, opened 2023-08-08
Replace VPackDeserializer with inspector

mpoeter:
### Scope & Purpose

Replace VPackDeserializer with inspector and completely remove VPackDeserializer.

Enterprise companion: https://github.com/arangodb/enterprise/pull/1346


arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -75,112 +72,96 @@ constexpr const char RETURN_TYPE_PARAM_NAME[] = "returnType";
...
+using namespace arangodb::aql;
 
 bool parse_options_slice(VPackSlice const& slice,
                          arangodb::iresearch::AqlAnalyzer::Options& options) {
-  auto const res = deserialize<ValidatingOptionsDeserializer,
-                               hints::hint_list<hints::ignore_unknown>>(slice);
+  auto const res = arangodb::velocypack::deserializeWithStatus(slice, options);
+
+  // auto const res = deserialize<ValidatingOptionsDeserializer,
+  //                              hints::hint_list<hints::ignore_unknown>>(slice);
```
markuspf, line 157: TODO: Remove

MBkkt (approved):

arangod/IResearch/IResearchAqlAnalyzer.cpp
```diff
@@ -75,112 +72,95 @@ constexpr const char RETURN_TYPE_PARAM_NAME[] = "returnType";
...
+              return {absl::StrCat("Value of '", BATCH_SIZE_PARAM_NAME,
+                                   "' should be greater than 0")};
+            }
+            if (v > MAX_BATCH_SIZE) {
+              return {absl::StrCat("Value of '", BATCH_SIZE_PARAM_NAME,
+                                   "' should be less or equal to ",
+                                   MAX_BATCH_SIZE)};
+            }
+            return {};
+          }),
```
MBkkt, line 118: I know it was, but will be nice to have another interface for deserialization

Something like 

```suggestion
      f.field(BATCH_SIZE_PARAM_NAME, o.batchSize)
          .fallback(10u)
          .booleanInvariant([](uint32_t v) -> bool {
            return v != 0 && v <= MAX_BATCH_SIZE;
          })
          .error([] () -> std::string {
            return absl::StrCat("Value of '", BATCH_SIZE_PARAM_NAME,
                                   "' should be greater than 0 and less or equal to ",
                                   MAX_BATCH_SIZE);
          },
```

And use this error not only in case of incorrect invariant, but also for not uint32_t values.

MBkkt: Well, your last example not same. My main point is have custom message for case not uint32_t values.

MBkkt: > the invariant from the actual error message, because you loose all the context.

For this place it's ok. here I don't need context to provide information to user what value should be specified

mpoeter: If the value in the vpack is not a uint32 then you will automatically get a proper error message. That is one of the reasons why I implemented the inspector in the first place, so that deserializing from json produces better error messages. However, it will only contain the information which type was expected, which type we found, and which field (incl. path) we tried to deserialize. It is not possible to customize the error. Theoretically this could be added, but I would not want to do that in this PR.

MBkkt: I don't think it's proper.
Proper write what you need to specify, this is impossible without custom error messqge

MBkkt: I don't expect this will be changed in this PR, just some suggestions to the future.

mpoeter: Changed to always initialize with 1.0

PR #19540, opened 2023-08-08
BTS-1546: Fix buffer overflow in crashdump

neunhoef:
### Scope Purpose

This fixes a buffer overflow in the crash dumper, which was observed in
the wild and prevented us from getting stack traces at all.

3.11 backport: https://github.com/arangodb/arangodb/pull/19546
3.10 backport: https://github.com/arangodb/arangodb/pull/19547


dothebart (approved):
LGTM.
imho
```
void appendNullTerminatedString(std::string_view src, char*& dst) {		
```
should be deleted.

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -407,14 +410,18 @@ void logBacktrace() try {
...
             // pad frame id to 2 digits length
             *p++ = ' ';
           }
-          p += arangodb::basics::StringUtils::itoa(uint64_t(frame), p);
+          if (BYTES_LEFT(buffer, p) > 19) {
+            p += arangodb::basics::StringUtils::itoa(uint64_t(frame), p);
+          }
 
-          appendAddress(pc, base, p);
+          if (BYTES_LEFT(buffer, p) > 16) {
```
jsteemann, line 422: IMHO `appendAddress()` can append up to 23 bytes.

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -179,28 +185,37 @@ void appendHexValue(unsigned char const* src, size_t len, char*& dst,
...
-void appendAddress(unw_word_t pc, long base, char*& dst) {
+void appendAddress(unw_word_t pc, long base, size_t bytesLeft, char*& dst) {
+  char* begin = dst;
   if (base == 0) {
     // absolute address of pc
-    appendNullTerminatedString(" [$0x", dst);
-    appendHexValue(reinterpret_cast<unsigned char const*>(&pc),
-                   sizeof(decltype(pc)), dst, false);
+    appendNullTerminatedString(" [$0x", bytesLeft, dst);
+    if (bytesLeft - (dst - begin) >= 2 * sizeof(decltype(pc))) {
```
jsteemann, line 205: The left side of the comparison can underflow if `bytesLeft < (dst - begin)`, and then it will always evaluate to `true`.

jsteemann: It should rather be `if (bytesLeft >= (dst - begin) + 2 * sizeof(decltype(pc)))`

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -430,21 +469,25 @@ void logBacktrace() try {
...
               // in this case, append mangled version. still better than nothing
-              appendNullTerminatedString(mangled, p);
+              appendNullTerminatedString(mangled, BYTES_LEFT(buffer, p), p);
             }
             // print offset into function
-            appendNullTerminatedString(" (+0x", p);
-            appendHexValue(reinterpret_cast<unsigned char const*>(&offset),
-                           sizeof(decltype(offset)), p, true);
-            appendNullTerminatedString(")", p);
+            appendNullTerminatedString(" (+0x", BYTES_LEFT(buffer, p), p);
```
dothebart, line 480: imho 0x should as well be in the if.

lib/CrashHandler/CrashHandler.cpp
neunhoef: Since we used `bytesLeft` in the first call to `appendNullTerminatedString`, `(dst - begin)` can never be larger than `bytesLeft`, therefore the difference cannot underflow.

lib/CrashHandler/CrashHandler.cpp
neunhoef: Fixed.

lib/CrashHandler/CrashHandler.cpp
neunhoef: Fixed in the meantime.

lib/CrashHandler/CrashHandler.cpp
neunhoef: Fixed to make it more understandable.

dothebart (approved):

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -430,21 +473,25 @@ void logBacktrace() try {
...
               // in this case, append mangled version. still better than nothing
-              appendNullTerminatedString(mangled, p);
+              appendNullTerminatedString(mangled, BYTES_LEFT(buffer, p), p);
             }
             // print offset into function
-            appendNullTerminatedString(" (+0x", p);
-            appendHexValue(reinterpret_cast<unsigned char const*>(&offset),
-                           sizeof(decltype(offset)), p, true);
-            appendNullTerminatedString(")", p);
+            if (BYTES_LEFT(buffer, p) >= 20) {
```
jsteemann, line 484: appendHexValue will append 16 bytes, and the string before can append 5 bytes (` (+0x`).
```suggestion
            if (BYTES_LEFT(buffer, p) >= 21) {
```

mpoeter (changes requested):
I _really_ don't like this code! We are adding a lot of additional complexity all over the place. Can we please use a proper stream/buffer implementation that handles handles all this internally?

jsteemann (approved):
LGTM

neunhoef:
I am totally fine with refactoring this more. This all started because I got a core dump from Willi today and found that the stack frame in the crashing thread had been overwritten by the buffer overflow. I first tried to do a very small fix of only a few lines which would have caught the actual issue at hand. But then reviewers - rightfully - wanted to protect once and for all, so I did the necessary changes to protect us in all places where this critical function was used.
I agree that it is now not very tidy. The fundamental problem here is that we want to avoid memory allocations on the heap when we are in the crash handling situation. Therefore we need a special solution and cannot work with the usual approaches.
Jan has agreed to look into this and implement a small special purpose class to handle all this more cleanly. So let's try this next.

jsteemann:
@markuspf @mpoeter : I created a small buffer utility class and made the CrashHandler use it now.
The buffer utility class is protected against buffer overflows and has a separate unit test.
Please re-review.

dothebart (approved):
much cleaner now. Nice. 
`mangled` would not benefit from this new class?

lib/Basics/SizeLimitedString.h
```diff
@@ -0,0 +1,160 @@
...
+        recalculateState();
+      }
+    }
+    return *this;
+  }
+
+  // append a hex-encoded value.
+  // will append up to as many bytes as there is remaining space in the buffer.
+  template<typename T>
+  SizeLimitedString& appendHexValue(T value, bool stripLeadingZeros) noexcept {
```
neunhoef, line 117: On a big endian platform this will write out the hex value in the wrong order. Should we bother?

neunhoef (commented):
LGTM, with the small exception of the endianess, should we bother?

neunhoef (commented):
LGTM.

lib/Basics/SizeLimitedString.h
```diff
@@ -0,0 +1,181 @@
...
+    recalculateState();
+  }
+
+  SizeLimitedString& append(std::string_view data) noexcept {
+    return append(data.data(), data.size());
+  }
+
+  SizeLimitedString& append(char const* data) noexcept {
+    return append(data, strlen(data));
+  }
```
mpoeter, line 87: This can be removed. `std::string_view` has an implicit conversion constructor from `char*`. In case of string literals this is even performed at compile time.

lib/Basics/SizeLimitedString.h
```diff
@@ -0,0 +1,181 @@
...
+          push_back(chars[c & 0xfU]);
+          stripLeadingZeros = false;
+        }
+        ++p;
+      }
+      if (stripLeadingZeros) {
+        push_back('0');
+      }
+    }
+    return *this;
```
mpoeter, line 163: This can be simplified as follows:
```cpp
    std::array<unsigned char, sizeof(T)> buffer[sizeof(T)];
    memcpy(buffer, &value, sizeof(T));
    if constexpr (basics::isLittleEndian()) {
      std::reverse(buffer.begin(), buffer.end());
    }

    // stringify value. if string is already full, this may do some
    // iterations, but not append to the buffer. we are not optimizing for
    // this case.
    constexpr char chars[] = "0123456789abcdef";
    for (auto c : buffer) {
      if (!stripLeadingZeros || (c >> 4U) != 0) {
        push_back(chars[c >> 4U]);
        stripLeadingZeros = false;
      }
      if (!stripLeadingZeros || (c & 0xfU) != 0) {
        push_back(chars[c & 0xfU]);
        stripLeadingZeros = false;
      }
    }
    if (stripLeadingZeros) {
      push_back('0');
    }

    return *this;
```

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -142,156 +145,84 @@ std::atomic<char const*> stateString = nullptr;
...
     // dump address that was accessed when the failure occurred (this is
     // somewhat likely a nullptr)
-    appendNullTerminatedString(" accessing address 0x", p);
-    unsigned char const* x =
-        reinterpret_cast<unsigned char const*>(info->si_addr);
-    unsigned char const* s = reinterpret_cast<unsigned char const*>(&x);
-    appendHexValue(s, sizeof(unsigned char const*), p, false);
+    buffer.append(" accessing address 0x");
+    auto const* x = reinterpret_cast<void const*>(info->si_addr);
+    buffer.appendHexValue(x, false);
```
mpoeter, line 199: Wouldn't this suffice?
```suggestion
    buffer.appendHexValue(info->si_addr, false);
```

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -142,156 +145,84 @@ std::atomic<char const*> stateString = nullptr;
...
     // AT_PHDR points to the program header, which is located after the ELF
     // header. This allows us to calculate the base address of the executable.
     auto baseAddr = getauxval(AT_PHDR) - sizeof(Elf64_Ehdr);
-    appendNullTerminatedString(" - image base address: 0x", p);
-    unsigned char const* x = reinterpret_cast<unsigned char const*>(baseAddr);
-    unsigned char const* s = reinterpret_cast<unsigned char const*>(&x);
-    appendHexValue(s, sizeof(unsigned char const*), p, false);
+    buffer.append(" - image base address: 0x");
+    auto const* x = reinterpret_cast<void const*>(baseAddr);
+    buffer.appendHexValue(x, false);
```
mpoeter, line 212:
```suggestion
    buffer.appendHexValue(baseAddr, false);
```

lib/CrashHandler/CrashHandler.cpp
```diff
@@ -142,156 +145,84 @@ std::atomic<char const*> stateString = nullptr;
...
-      appendNullTerminatedString(prefix, p);
-      unsigned char const* s =
-          reinterpret_cast<unsigned char const*>(&ctx->uc_mcontext.gregs[reg]);
-      appendHexValue(s, sizeof(greg_t), p, false);
+  if (auto ctx = static_cast<ucontext_t*>(ucontext); ctx) {
+    auto appendRegister = [ctx, &buffer](char const* prefix, int reg) {
+      buffer.append(prefix);
+      auto const* s =
+          reinterpret_cast<greg_t const*>(&ctx->uc_mcontext.gregs[reg]);
+      buffer.appendHexValue(*s, false);
```
mpoeter, line 223:
```suggestion
      buffer.appendHexValue(ctx->uc_mcontext.gregs[reg], false);
```

mpoeter (approved):
This is _much_ better than what we had before!

I think there are still a view things that can be simplified a bit, but otherwise LGTM.

mpoeter (approved):

markuspf (approved):

PR #19542, opened 2023-08-08
[BTS-1544][DEVEL] _system db not reporting its sharding property correctly

hkernbach:
### Scope & Purpose

The _system database has not been reporting its sharding value properly.
This PR fixes the behavior. 

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] Backports
  - [x] Backport for 3.11: *(Please link PR)*

#### Related Information

- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1347
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1544


arangod/VocBase/VocbaseInfo.cpp
```diff
@@ -346,6 +346,16 @@ VocbaseOptions getVocbaseOptions(ArangodServer& server, VPackSlice options,
...
 
   bool haveCluster = server.hasFeature<ClusterFeature>();
+  {
+    // We need to also take care of the `cluster.force-one-shard` option here.
+    // It set, the entire cluster is forced to be a OneShard deployment.
+    if (haveCluster) {
+      if (server.getFeature<ClusterFeature>().forceOneShard()) {
+        vocbaseOptions.sharding = "single";
+      }
+    }
```
jsteemann, line 356: Is this actually correct?
What if the cluster was created initially without `--cluster.force-one-shard`, then the `_system` database was set up without OneShard option, and then the cluster got restarted with `--cluster.force-one-shard true`?
Then we would report being a OneShard database here although we aren't.

hkernbach: @jsteemann Is this specific case then not fetched during start up time?

jsteemann: The problem is that the `_system` database is created in a very different way in an initally empty cluster.
It is created via `Upgrade::createDB(...)`, and this method does not set the `{ sharding: "single" }` attribute if `--cluster.force-one-shard` it set.
We should fix the issue _there_ and not fake the result of some function and return a _wrong_ value for an existing `_system` database which is not set up as a OneShard database.

markuspf (changes requested):
Looks mostly ok to me, I'd prefer to do the unrelated removal of the static strings in a separate PR though.

markuspf (approved):

PR #19543, opened 2023-08-08
Clean up and add assertions

jsteemann:
### Scope & Purpose

This PR does multiple things at the same time:
* adds assertions for OutputAqlItemRow state in IndexExecutor.cpp and EnumerateCollectionExecutor.cpp. Also adds assertions for the state of the various produceRows/skip implementations.
* adds assertions for OutputAqlItemRow state in DocumentProducingHelper.cpp
* replaces `a.reset(new ...)` with `a = std::make_(shared|unique)` when `a` was a shared_ptr or unique_ptr.
* replaces calls to `ExecutionPlan::registerNode(new T(args))` with `ExecutionPlan::createNode<T>(Args&&... args)`
* fixes a potential issue with deferred query killing, when a vocbase pointer was captured in a lambda, and the lambda could be executed by the scheduler at any later point with the vocbase being already deleted and the pointer being invalid.
* adds application server state information (in startup, running, in shutdown etc.) to CrashHandler output.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cluster/MaintenanceRestHandler.cpp
```diff
@@ -185,48 +185,44 @@ void MaintenanceRestHandler::putAction() {
...
       priority = static_cast<int>(value.getInt());
-    } else if (key.isString() && (key.stringView() == "forced") &&
+    } else if (key.isString() && key.stringView() == "forced" &&
                value.isBool()) {
       forced = value.isTrue();
     } else {
-      good = false;
-    }  // else
-  }    // for
+      return false;
```
mchacki, line 217: Is this a silent bug-fix here?
As far as I understand in original variant always an action was created.
As we could encounter an unexpected key/value pair at any position in the pinpoint the generated Action may actually be "undefined".
I prefer this variant!

mchacki (approved):
LGTM.

PR #19565, opened 2023-08-10
Make Pregel work on OneShard databases without issues

markuspf:
### Scope & Purpose

Pregel can work without issues on a OneShard database, so don't error out in that case.

Enterprise part with a test: https://github.com/arangodb/enterprise/pull/1350

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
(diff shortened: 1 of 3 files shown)


markuspf:
For this PR to pass tests, a fix for OneShard databases developed by @hkernbach is currently necessary

hkernbach (approved):

PR #19572, opened 2023-08-11
Replace multiple maps with a single map and use a follower struct.

maierlars:
### Scope & Purpose
There were multiple maps mapping follower ids to different information. This PR creates a struct that holds all this information and reduces everything to a single map.


maierlars:
Sanitizer run: https://jenkins01.arangodb.biz/view/PR/job/arangodb-matrix-pr/31150/

PR #19583, opened 2023-08-11
[3.11] Fix devel CI build

markuspf:
### Scope & Purpose

The "sudden" failure is caused by the fact that the CircleCI lint action just installs eslint of the day. This likelly got updated and is now complaining.

This PR fixes the immediate redness, a better fix will be to fix the eslint version for linting jobs.

### DIFF
--- a/js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
+++ b/js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
@@ -1,5 +1,5 @@
 /* jshint unused: false */
-/* global Noty, Blob, window, Joi, sigma, $, tippy, document, _, arangoHelper, frontendConfig, sessionStorage, localStorage, XMLHttpRequest */
+/* global Noty, window, Joi, sigma, $, tippy, document, _, arangoHelper, frontendConfig, sessionStorage, localStorage, XMLHttpRequest */
 
 (function () {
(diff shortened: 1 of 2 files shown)


js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
```diff
@@ -1,5 +1,5 @@
 /* jshint unused: false */
-/* global Noty, Blob, window, Joi, sigma, $, tippy, document, _, arangoHelper, frontendConfig, sessionStorage, localStorage, XMLHttpRequest */
```
mchacki, line 2: This Blob is actually required at least on Jenkins still.
And I would say it is correct, as Blob is used later in this file.

mchacki (changes requested):
I think this should now break Jenkins JSLint job causing our nightlies to be red.

hkernbach (approved):

mchacki (approved):
LGTM

js/apps/system/_admin/aardvark/APP/frontend/js/arango/arango.js
mpoeter: In the browser environment Blob is already predefined, so we don't need to define it yet again. This redefinition is exactly what the newer eslint complains about.

PR #19585, opened 2023-08-11
[CINFRA-871] ModifyShard operation

apetenchea:
### Scope & Purpose

**General explanation**
Together with the `CreateShard` and `DropShard` operations, the `ModifyShard` has to follow a similar code-path:
1. The maintenance first sees there is a difference between Plan and local.
2. In case of replication2, the `UPDATE_COLLECTION` action is triggered only for the leader. Basically, the maintenance should have no effect on followers at this stage.
3. The `UPDATE_COLLECTION` action calls into a special method for the replication2 leader, which ensures that first the `ModifyShard` operation is committed in the log, then the leader creates it locally.

**Remarks**
- We have decided to keep `servers` and `failoverCandidates` in sync by honoring the `followersToDropString` that comes with the `UPDATE_COLLECTION` action, such that replication 1 doesn't go crazy. The `followersToDropString` holds no real meaning in the context of replication2 and is eventually going to be removed. Followers completely ignore it.
- This is a bit of an ugly hack, but in order to first execute the piece of code that's specific to replication2 leader, we differentiate it from the regular maintenance action by passing `{"from", "maintenance"}` in the action description.
- For the integration tests, I had to use a workaround in order to modify the properties of a collection. Currently, doing so through the JS API does not work for replication2.

**Additional fixes**
- The `MaintenanceActionExecutor` now returns the real result value of the action.
- The `replicatedStateFollowerSuiteV1` was accidentally left to use a replication2 database, but it was intended as a replication1 test (now fixed).
- There was a race between the maintenance and the replicated state. While a vocbase was being dropped, we used to call `shutdownReplicatedLogs()` after the vocbase was made unavailable. The already existing references to it were still valid, but new references would fail to be created. Therefore, if the maintenance would trigger a database drop while a follower would execute `appendEntries`, the database drop would fail to acquire to follower's `_guardedData` lock (needed for `resign()`) and just wait for `appendEntries` to finish. However, since some of the document operations left in the log might require a valid vocbase reference (for example, `ModifyShard`), these operation would fail, leading to a confusing scenario: the document state has not officially resigned, yet it failed to apply an entry that it received from the leader. In most cases, this is fatal, but when dropping the database, this is fine.
Two modifications were made for this fix: the `shutdownReplicatedLogs()` was moved before the point where the vocbase is made unavailable, and the `_resigning` boolean was introduced in order to make the state machine aware that it is about to resign soon, without having to acquire any locks (this is just a speedup).

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**


tests/js/server/replication2/replication2-replicated-state-document-store.js
```diff
@@ -583,20 +583,51 @@ const replicatedStateRecoverySuite = function () {
...
+      collection = db._create(collectionName, {"numberOfShards": 2, "writeConcern": 2, "replicationFactor": 3});
       let plan = lh.readAgencyValueAt(`Plan/ReplicatedLogs/${dbNameR1}`);
       assertEqual(plan, undefined, `Expected no replicated logs in agency, got ${JSON.stringify(plan)}`);
     },
+
+    testChangeCollectionProperties: function () {
+      // This test does not work with replication version 2,
+      // because the collection properties must be written to Target.
+      // I'm leaving this test here for now, because it is a good example.
+      // We should eventually port it to replication2.
```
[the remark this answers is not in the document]
apetenchea: I don't know whether some of @mchacki 's collection related tickets already includes this. To be sure, I created [CINFRA-822](https://arangodb.atlassian.net/browse/CINFRA-822) and added it to the [Leftovers of CreateCollection](https://arangodb.atlassian.net/browse/CINFRA-626) epic.

tests/js/server/replication2/replication2-replicated-state-document-store.js
```diff
@@ -940,25 +978,91 @@ const replicatedStateDocumentShardsSuite = function () {
...
+        checkCollectionShards(collection2);
+      } finally {
+        db._drop("other-collection");
+      }
     },
+
+    testModifyShard: function () {
+      const collection = db._create(collectionName, {numberOfShards: 1, writeConcern: 2, replicationFactor: 3});
+
+      // TODO currently we cannot modify collection.properties for replication2, because the API must write to `Target`.
```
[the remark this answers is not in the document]
apetenchea: [CINFRA-822](https://arangodb.atlassian.net/browse/CINFRA-822)

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
```diff
@@ -82,6 +83,31 @@ auto MaintenanceActionExecutor::executeDropCollectionAction(
...
+          {maintenance::DATABASE, _gid.database},
+          {maintenance::COLLECTION, collection},
+          {maintenance::SHARD, shard},
+          {maintenance::SERVER_ID, _server},
+          {maintenance::FOLLOWERS_TO_DROP, followersToDrop}},
+      maintenance::HIGHER_PRIORITY, true, std::move(properties));
+
+  maintenance::UpdateCollection updateCollectionAction(_maintenanceFeature,
+                                                       actionDescription);
+  bool work = updateCollectionAction.first();
```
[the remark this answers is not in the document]
apetenchea: In this case, checking just the replication version would not be enough. We also need to differentiate between replication2 leader and follower. The `updateReplication2Shard` is supposed to be only executed on the leader, among with the first part of `doUpdateCollection()` related to `followersToDrop`. It probably wouldn't end up as simple as a single if statement, but I'm sure there must be a nicer way of achieving this, without having to rely on "from".

Am I interested in a better way of doing this?
Totally! I don't feel very pleased myself with the current code. The main reason I chose this approach was to keep it consistent with it's siblings (CreateShard and DropShard), which follow the same template. From the amount of code written point of view, this is the most straightforward way, and I already had two working examples in the Maintenance.

I did not want to "fix" it for `ModifyShard` in this PR, as I'm hoping to adapt all 3 of these operations. I have created a refactoring sub-task [CINFRA-823](https://arangodb.atlassian.net/browse/CINFRA-823), and I suggest taking care of it in a separate PR. Does that sound like a good compromise to you?

arangod/Cluster/UpdateCollection.cpp
```diff
@@ -159,3 +172,23 @@ void UpdateCollection::setState(ActionState state) {
   }
   ActionBase::setState(state);
 }
+
+bool UpdateCollection::updateReplication2Shard(
```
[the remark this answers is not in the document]
apetenchea: The motivation for choosing this name was somewhat similar to the above. We have already two similar actions:
- CreateCollection -> createReplication2Shard
- DropCollection -> dropReplication2Shard
`updateReplication2Shard` follows the same template.
We don't have to be stuck with that name though. `replicateModifyShardOperation` or `initiateModifyShard` might make it seems as if we're doing half the job in there, whereas in the context of the leader this also triggers the actual shard modification (basically does everything). This is not great, and it shows here again how "hidden" this important difference between leader and follower is in fact :(
Combining both replication1 and 2 in the same maintenance code, while keeping it obvious for both what events are currently going on, is not a trivial change in this case. Shall we also move this to the refactoring sub-task?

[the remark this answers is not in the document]
goedderz: I didn't know about the similarity with the other actions, that makes sense.

goedderz (commented):
@apetenchea One additional observation. It looks to me as the leader will insert one `ModifyShard` operations per maintenance run until the first one actually succeeds, or am I missing something? That'd be undesirable, and we should fix it soon. Doesn't have to be in this PR.

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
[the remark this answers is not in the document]
apetenchea: Yes, the maintenance instantiates the Action only for leaders. The maintenance itself only ever triggers `UpdateCollection::first()` on leaders. On followers, `UpdateCollection::first()` is triggered from the `MaintenanceActionExecutor`, when they apply the log entry locally.
You're right, `ignoreFollowersToDrop` would handle the difference between leader and followers in `doUpdateCollection()`.

apetenchea:
> @apetenchea One additional observation. It looks to me as the leader will insert one `ModifyShard` operations per maintenance run until the first one actually succeeds, or am I missing something? That'd be undesirable, and we should fix it soon. Doesn't have to be in this PR.

Oh no, I think you're right. My intention was as follows:

```
auto&& applyEntryRes = data.transactionHandler->applyEntry(std::move(op));
if (applyEntryRes.fail()) {
  LOG_CTX("b5e46", FATAL, self->loggerContext)
      << "ModifyShard operation failed on the leader, after being "
         "replicated to followers: "
      << applyEntryRes;
  FATAL_ERROR_EXIT();
}
```
Basically, if the `UpdateCollection::first()` call fails on the leader, it produces a crash. We would not be able to replicate it again. However, now that I'm looking again, does it ever return `true`? I assumed `true` was used to indicate failure. But no, apparently it does not.
Do you know what is the best way to check that an action succeeded? I think it should be something like:
```
bool work = updateCollectionAction.first();
if (work) {
  return {TRI_ERROR_INTERNAL};
}
return updateCollectionAction.result();
```
This seems like a trivial change, I'm going to do it it this PR.

apetenchea:
@goedderz all 3 actions (create, drop and modify) in `MaintenanceActionExecutor` now return the action's result. This way, if the leader successfully replicates it and the entry gets committed, but the action fails locally, the leader crashes.
The maintenance action is retried next round only if replication fails.

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
```diff
@@ -79,7 +80,32 @@ auto MaintenanceActionExecutor::executeDropCollectionAction(
...
+  if (work) {
+    return {TRI_ERROR_INTERNAL};
+  }
```
goedderz, line 107:
```suggestion
  updateCollectionAction.first();
```
I think you want to ignore the return value, right?

apetenchea: I would be curious to know what the return value means. But, since we got the `result()`, I guess we can safely ignore that.

goedderz: Afair this has to do with how the Maintenance handles re-running an action.

apetenchea:
After modifying the `MaintenanceActionExecutor` to take the action's `result()` into account, some other things fell apart. I added explanations in the **Additional fixes** section.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -201,9 +208,15 @@ auto DocumentLeaderState::recoverEntries(std::unique_ptr<EntryIterator> ptr)
     }
 
     auto abortAllResult = data.transactionHandler->applyEntry(abortAll);
-    TRI_ASSERT(abortAllResult.ok());
+    TRI_ASSERT(abortAllResult.ok()) << abortAllRes.result();
```
goedderz, line 211: This checks the result of the local `transactionHandler->applyEntry`, but prints the result of `replicateOperation`. Is that intentional? The variable names are a little ambiguous here.

arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -804,6 +804,10 @@ ErrorCode DatabaseFeature::dropDatabase(std::string_view name) {
 
     TRI_vocbase_t* vocbase = it->second;
 
+    // Signal replicated logs to stop here, while they still have access to the
+    // database.
+    vocbase->shutdownReplicatedLogs();
+
```
goedderz, line 810: Doing it this early will actually introduce a race: New replicated logs could still be added, which are then not shutdown at all. The call to shutdown logs must only happen after the vocbase was marked as dropped, which prevents new logs from being created.

goedderz (changes requested):

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: Thanks for noticing that! Fixed in https://github.com/arangodb/arangodb/pull/19585/commits/fb8206f18f0b2c8139b0bc78212c3766d51c810d

arangod/RestServer/DatabaseFeature.cpp
apetenchea: To my current understanding, no replicated states (nor logs) may be created after `VocbaseLogManager::resignAll()` is called. `shutdownReplicatedLogs()` just calls into `_logManager->resignAll()`.
The `resignAll()` starts with the following lines:
```
auto guard = _guardedData.getLockedGuard();
guard->resignAllWasCalled = true;
``` 
The `resignAllWasCalled` atomic is used to prevent new instances from being created.
The one point of creation (that I am aware of) for replicated logs and states is `VocBaseLogManager::GuardedData::buildReplicatedStateWithMethods()`. This does the following check in the beginning:
```
if (resignAllWasCalled) {
  return {TRI_ERROR_SHUTTING_DOWN};
}
```
I think this is enough to prevent the race you're mentioning.

goedderz: Ah, good! Thanks for the explanation. The error message is probably wrong when a database is being deleted, rather than the process being shut down, but that doesn't have to be fixed here.

goedderz (approved):

PR #19591, opened 2023-08-14
[CI] Remove outdated custom eslint

markuspf:
### Scope & Purpose

Attempts to remove all custom and outdated eslint code (that masquerades as jslint code)

This removes 2200 files and 22MiB from the repository. For comparison: all sources in `arangod`, `lib`, and `Enterprise` are 35MB.


jsteemann:
If this is removed, please also remove the `--jslint` option from the arangosh, which is based on the now-removed functionality. Additionally, the removed functionality is mentioned in `CONTRIBUTING.md`, and should be removed from there too.
And because it is a downwards-incompatible change, a CHANGELOG entry would be useful and the removed functionality should be mentioned in the documentation. Thanks!

dothebart (approved):

jsteemann (approved):
LGTM, but let's also add a CHANGELOG entry and open a documentation PR that mentions this downwards-incompatible change. This is important so we can say "it's documented" should someone still use the functionality and insist on it missing being a bug.

mpoeter (approved):

PR #19610, opened 2023-08-15
add build-id in startup message and diagnostics

jsteemann:
### Scope & Purpose

This change will make arangod log its own build-id on startup and in crash messages, provided that it was compiled on Linux. The build-id will also be exposed when running `/path/to/arangod --version`.
The build-id should be identical to the build-id revealed by `readelf -n /path/to/arangod`.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


dothebart (approved):

jsteemann:
Looks like that the approach does **not** work with static executables built using libmusl.
It works fine for local builds, but probably the benefit of having a build-id there is relatively low.

MBkkt (dismissed):
Why?

MBkkt:
@jsteemann I read thread in the slack and still don't understand the purpose of this number in the log/whatever

lib/Basics/BuildId.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    _size = statbuf.st_size;
+    return true;
+  }
+
+  auto get() const -> const char* {
+    return reinterpret_cast<const char*>(_memory);
+  }
+
+  auto valid() const -> bool { return _memory != nullptr; }
```
markuspf, line 87:
```suggestion
  auto valid() const -> bool { return _memory != MAP_FAILED; }
```

Or remove as this method is not used atm

tests/Basics/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 target_sources(arangodbtests
   PRIVATE
+  BuildIdTest.cpp
```
jsteemann, line 3: I guess this test file is currently missing from the PR, so building the tests will likely fail.

lib/Basics/BuildId.ld
jsteemann, line 1: Good idea. I also read about this, but did not want to go that direction myself because of potentially portability issues.
Also wondering how the linker will pick up the linker script. Is it somehow auto-detected based on filenames, or does it need to be explicitly invoked? I can't see any place in this PR that would invoke it explicitly, at least.

tests/Basics/BuildIdTest.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <gtest/gtest.h>
+
+#include "Basics/BuildId.h"
+
+using namespace arangodb::build_id;
+
+TEST(BuildIdTest, successfully_obtains_build_id) {
+  if constexpr (arangodb::build_id::supportsBuildId()) {
+    auto buildId = getBuildId();
+    ASSERT_EQ(buildId.size(), 20);
```
jsteemann, line 32: Note that the build-id size seems to be variable in reality.
When I built arangod locally using clang, I got build-ids that were 8 bytes long (16 bytes in hex), whereas when I built with Oskar, I got build-ids that were 10 bytes long (20 bytes in hex).
The actual length seems to depend on the build-id "style".
From `man ld`:
```
       --build-id
       --build-id=style
           Request the creation of a ".note.gnu.build-id" ELF note section or
           a ".buildid" COFF section.  The contents of the note are unique
           bits identifying this linked file.  style can be "uuid" to use 128
           random bits, "sha1" to use a 160-bit SHA1 hash on the normative
           parts of the output contents, "md5" to use a 128-bit MD5 hash on
           the normative parts of the output contents, or "0xhexstring" to use
           a chosen bit string specified as an even number of hexadecimal
           digits ("-" and ":" characters between digit pairs are ignored).
           If style is omitted, "sha1" is used.
```

lib/Basics/BuildId.ld
markuspf: Can you specify which portability issues you see?

I went through 2 iterations for this code:
 * open/mmap `/proc/self/exe` and parse just enough of the elf header to find the build-id. This works, and is probably ok-ish robustness-wise. There's not much that can break other than maybe reading all of the arangodb executable into the build-id.
 * having the linker do the dirty work of telling me where to find the build id. I just realised this morning that this is actually not as straightforward as I thought yesterday: the build-id is part of the binary (arangod in this instance); nothing really compels the linker to tell the loader to put it into memory. (I think this is why the `__attribute__(section(...))` approach does not work). But as far as I understand from the documentation this approach works and is quite robust. In particular in this PR, if there's a problem with the code one can use `-DUSE_BUILD_ID=Off` and everything will "just work".

From the documentation (for the linker), the `BuildId.ld` file is looked at and interpreted as an implicit linker script when passed as a file during the linking phase. It's a bit awkward to have to add `BuildId.ld` to every executable that uses the library that has the `build-id` code in it, but I did not just want to blanket-edit the linker script for every executable built with the build system.

As a sidenote for myself: There are two issues which I conflated in this PR's changes to the build system: "put a build id in the executable" and the "access the build id from inside the exectuable". While one can always still add `-Wl,--build-id` by hand, I think it should be the default to add the build-id and there should be a verification step (using readelf) in the tests that the build id is actually there.

tests/Basics/BuildIdTest.cpp
markuspf: Fair enough; the 20 was just a potshot to see whether there is *anything* I can test that has meaning.

markuspf:
It's probably better to move the build-id code into a small library of its own and just link that to `arangod` and maybe `arangodbtests`; that makes the whole linker-spiel slightly less annoying.

MBkkt:
I read thread in the slack and still don't understand the purpose of this number in the log/version string/whatever

markuspf:
> I read thread in the slack and still don't understand the purpose of this number in the log/version string/whatever

The use-case is as follows: Someone is asked for help with a problem with ArangoDB, but only gets a screenshot of a crash, or only logs, or can only access an arangodb instance via the REST interface. With this patch the `build-id` is more easily available.

With this information in hand one can try and find the binary that was used and investigate further using *that* binary. 

In a more advanced setup, the `build-id` of every shipped binary is recorded in a database (https://sourceware.org/elfutils/Debuginfod.html) so our lives become easier.

(This is not hypothetical, we have chased up binaries using build-ids)

MBkkt:
Do we really have different binaries for same commit+time?

PR #19612, opened 2023-08-16
fix multiple issues in scheduler

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1354

Fix multiple issues in scheduler.

- sendRequestRetry in NetworkFeature could add a request to the scheduler even when the server was already stopping. the request was canceled correctly, but still posted to the scheduler unnecessarily. this is now fixed so that the request is not posted to the scheduler anymore if the server is shutting down
- this also fixes an issue with retry requests being posted to the scheduler after the shutdown was started. such retry requests led to assertion failures (`item->isDisabled()`) in the shutdown of the scheduler.
- in case the requested delay for `Scheduler::queueDelayed()` was less than 1ms, the task was queued directly and a nullptr was returned. this potential nullptr return value was not handled correctly in the NetworkFeature and could have led to nullptr dereferences. this is now fixed by removing the special case in `queueDelayed()` for tasks with a delay < 1ms.
- `Scheduler::queueDelayed()` was, probably intentionally, marked noexcept, although the method could throw in at least 2 places in OOM situations. now, the method is really noexcept and returns a nullptr on failure. this is not a problem because most call sites of the function handle nullptr return values correctly. the only problematic call site in `sendRequestRetry()` is now also adjusted.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1354
- [ ] Design document:


mchacki (approved):
LGTM

arangod/Network/NetworkFeature.cpp
```diff
@@ -528,15 +539,14 @@ void NetworkFeature::retryRequest(
       }
 
       _retryRequests.emplace(std::move(req), std::move(item));
-      return;
+      // the request successfully made it into _retryRequests.
+      // now abandon the automatic cancelation.
+      cancelGuard.cancel();
```
MBkkt, line 544: It's not really necessary if you're really moved req

PR #19614, opened 2023-08-16
[BTS-724][DEVEL] Added test for opt-rule: remove-redundant-path-var

hkernbach:
### Scope & Purpose

Add missing test for optimizer rule: `remove-redundant-path-var`

- [x] :hammer: Added missing test

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :book: CHANGELOG not required, only added a new test

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-724


js/server/modules/@arangodb/testutils/aql-graph-traversal-generic-tests.js
```diff
@@ -1294,6 +1294,43 @@ function testOpenDiamondDfsLabelVariableForwarding(testGraph) {
...
+  }
+
+  for (const query of queryList) {
+    const optPlans = AQL_EXPLAIN(query, {}, { optimizer: { rules: optimizerRuleToDeactivate } }).plan;
+    assertTrue(optPlans.rules.includes("remove-redundant-path-var"));
+  }
+
+  // Now return p, to make sure the rule is not being executed.
+  const optPlans = AQL_EXPLAIN(buildQuery('[p]'), {}, { optimizer: { rules: optimizerRuleToDeactivate } }).plan;
+  assertEqual(optPlans.rules, []);
```
jsteemann, line 1327: nit: this makes the test depend on if there are other, unrelated optimizer rules than the one we are looking for.
This may break the test when new optimizer rules are added.
Would prefer a more tailored solution that checks if the optimizer rule in question is not there, but will be happen with other, unrelated optimizer rules.
```suggestion
  assertFalse(optPlans.rules.includes("remove-redundant-path-var"));
```

jsteemann (approved):
LGTM

js/server/modules/@arangodb/testutils/aql-graph-traversal-generic-tests.js
hkernbach: Yes, my implementation is very "strict". Let me think about this. I think I might like the approach that a new optimizer rule may break my test - to be eventually able to identify an unwanted change. Changing it to your suggestion works, but would not identify any potentially unwanted optimizer rule changes for that simply query...

hkernbach: Ok, CI proved that your approach is required. Will change.

hkernbach:
All tests green except one unrelated windows failure.

PR #19627, opened 2023-08-18
Allow content negotation for metrics API responses

jsteemann:
### Scope & Purpose

Allow compression content-negotation in metrics API, so that responses from the metrics API at `/_admin/metrics` can be sent compressed if the client supports it.

The metrics returned by the `/_admin/metrics` API are returned as a big text string, using only a very limited alphabet. The data should be very well compressible. As metrics are queried from all servers pretty often, we could probably save a lot of traffic at least on idle servers. For busy servers the traffic reduction is probably not significant.

According to the Prometheus docs, it supports content-negotiation and the "gzip" content encoding:
https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

Local experiments show that enabling gzip compression for metrics reduces the size of the metrics API response by about 90%.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

PR #19631, opened 2023-08-21
[CI] ARM Builds/Tests

markuspf:
### Scope & Purpose

Make most of the build/test setup parameterised over the platform (arm or x86) and run builds and tests on both.


.circleci/generate_config.py
```diff
@@ -221,8 +221,20 @@ def filter_tests(args, tests, enterprise):
...
 
+def get_size(size, arch):
+    aarch64_sizes = {
+        "small": "arm.medium",
+        "medium": "arm.medium",
+        "medium+": "arm.large",
+        "large": "arm.large",
+        "xlarge": "arm.xlarge"
+    }
+    if arch == "aarch64":
```
dothebart, line 232: why param this in? shouldn't 
```
IS_ARM = platform.processor() == "arm" or platform.processor() == "aarch64"
```
be enough here?

CryptoNinjaGeek (approved):

.circleci/generate_config.py
mpoeter: No, this script runs just once on some system, but generates the jobs for all architectures.

PR #19632, opened 2023-08-21
[CINFRA-825] Remove external release api from replicated log

goedderz:
### Scope & Purpose

Remove the public release API of replicated logs: It must only ever be used by the corresponding replicated state.

In order to keep js-tests working, implement `release` via the `black-hole` state machine (only).

Improved runtime of `testRelease` in `shell-replicated-logs-cluster.js`.

### DIFF
--- a/arangod/Replication2/Methods.cpp
+++ b/arangod/Replication2/Methods.cpp
@@ -47,6 +47,7 @@
 #include "Replication2/ReplicatedLog/LogStatus.h"
 #include "Replication2/ReplicatedLog/ReplicatedLog.h"
 #include "Replication2/ReplicatedState/ReplicatedState.h"
+#include "Replication2/StateMachines/BlackHole/BlackHoleStateMachine.h"
 #include "VocBase/vocbase.h"
 
 #include "Agency/AgencyPaths.h"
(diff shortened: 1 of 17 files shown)


arangod/Replication2/Methods.cpp
```diff
@@ -220,8 +221,31 @@ struct ReplicatedLogMethodsDBServer final
...
+                replicated_state::black_hole::BlackHoleState>>(stateBase.get());
+        state != nullptr) {
+      if (auto leaderState = state->getLeader(); leaderState != nullptr) {
+        return leaderState->release(index);
+      } else {
+        return Result(
+            TRI_ERROR_HTTP_FORBIDDEN,
+            fmt::format("/release API is only allowed on leaders; while "
+                        "trying to release state machine {} on {}.",
+                        id, ServerState::instance()->getId()));
```
apetenchea, line 239: I was under the impression that `/release` can be called on followers as well. Each state machine (be it leader or follower) can call release independently.
Will the `leaderState->release(index)` call propagate to followers?
Not that restricting the `/release` like this would do any harm, since it is only used for testing. I'm just curios to learn why.

apetenchea (approved):
LGTM

arangod/Replication2/Methods.cpp
goedderz: We could allow it on the followers. But the API is now restricted to the black hole state machine already, just to support the tests. I would have removed it completely if it weren't for the tests; and I think we would write the tests differently today, so we wouldn't need it (i.e. unit tests).

PR #19633, opened 2023-08-21
[CINFRA-826] Fix replication2 TSAN

apetenchea:
### Scope & Purpose

This PR attempts to fix the current TSAN issues that were introduced in https://github.com/arangodb/arangodb/pull/19285
- protect the `_syncListeners` with the `_syncListenersMutex`
- `std::move()` the additional `logPersistor` copy
 
Here is the [tsan.log](https://jenkins.arangodb.biz/job/arangodb-ANY-linux-san.x86-64/2056/EDITION=enterprise,SAN_MODE=TSan,STORAGE_ENGINE=rocksdb,TEST_SUITE=single,limit=linux&&test&&x86-64/artifact/tsan.log.arangod.47013.log) file

Jenkins: https://jenkins01.arangodb.biz/job/arangodb-ANY-linux-san.x86-64/2063/

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/RocksDBEngine/RocksDBEngine.cpp
+++ b/arangod/RocksDBEngine/RocksDBEngine.cpp
@@ -1277,7 +1277,7 @@ void RocksDBEngine::start() {
   _logPersistor = logPersistor;
 
   if (auto* syncer = syncThread(); syncer != nullptr) {
-    syncer->registerSyncListener(logPersistor);
+    syncer->registerSyncListener(std::move(logPersistor));
   } else {
     LOG_TOPIC("0a5df", WARN, Logger::REPLICATION2)
         << "In replication2 databases, setting waitForSync to false will not "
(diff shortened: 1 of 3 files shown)


jsteemann (approved):
LGTM, but couldn't we avoid adding an RW lock completely by creating the listeners a bit earlier in the code?
I.e. register all sync listeners first, and only then call notify on them when the sync listeners won't change anymore.

apetenchea:
> LGTM, but couldn't we avoid adding an RW lock completely by creating the listeners a bit earlier in the code? I.e. register all sync listeners first, and only then call notify on them when the sync listeners won't change anymore.

Totally! In fact, this is how it worked before. The only such listener is the `logPersistor`, that is instantiated once in `RocksDBEngine.cpp:1271`. This means that `registerSyncListener()` is only called once.
But, TSAN generated about 8 warnings, I just wanted to silence it. I was weary to add a new warning suppression because this involves RocksDB related code and did not want to take the risk of hiding something that might show up in the future. Instead, I chose to over-engineer this already over engineered sync-listener stunt. Once our custom WAL implementation is ready for use, I am going to remove this from the syncer thread.
If you believe a suppression would've been better, I'm happy to change it back.

MBkkt (approved):
> If you believe a suppression would've been better, I'm happy to change it back.

I don't think so, because this race could crash process

PR #19635, opened 2023-08-21
Fix an include with clang, and issue an error when mold is used with the build id reader

goedderz:
### Scope & Purpose

The linker scripts used for the build id reader, introduced in https://github.com/arangodb/arangodb/pull/19610, aren't supported by `mold`. Make our `CMakeLists.txt` issue an error when they're used together.

Additionally, the PR https://github.com/arangodb/arangodb/pull/19220 (which replaced jemalloc's libunwind include argument `-I` with `-isystem`) stopped my clang from finding `libunwind.h`. I've now added both arguments, so it hopefully works everywhere.


markuspf (approved):
I do wonder whether one should turn that linker list into a positive list of supported linkers and warn when a linker is used that is not "supported".

I will have another think about whether one should maybe use the code that reads the build-id from the binary (as that should not lead to difficulties linking at least)

3rdParty/jemalloc/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ if (LINUX OR DARWIN)
   if (USE_JEMALLOC_PROF)
       if (USE_LIBUNWIND)
           SET(JEMALLOC_PROF "--enable-prof" "--enable-prof-libunwind" "--with-static-libunwind=${LIBUNWIND_LIB}")
-          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -isystem ${LIBUNWIND_HOME}/include")
+          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -isystem ${LIBUNWIND_HOME}/include -I${LIBUNWIND_HOME}/include")
```
MBkkt, line 38: I don't think it's good idea, because -isystem was used to avoid warnings produced by jemalloc headers

What clang doesn't like?

mchacki:
Did not do a formal review, just tried if it worked on my setup, where I have seen the libunwind issue.
So I can confirm this PR builds on Ubuntu + clang-14 not hitting libunwind issue again.

dothebart (approved):

markuspf (approved):

PR #19641, opened 2023-08-22
expose endianness

jsteemann:
### Scope & Purpose

To improve debuggability,

* Remove attribute `dfdb` from response of storage engine API (GET `/_api/engine`).

* The storage engine API (GET `/_api/engine`) now returns an `endianness` attribute for the RocksDB storage engine on single servers and database servers.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [x] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19647, opened 2023-08-22
Assert same size of uint64_t and double for bitcast

markuspf:
### Scope & Purpose

Just two slight nitpicks I noticed while viewing a diff too late in #19641.


markuspf:
One could also use `std::bit_cast`.

jsteemann (approved):
LGTM

PR #19661, opened 2023-08-24
Suppressing cppcheck warning

apetenchea:
### Scope & Purpose

cppcheck failed during nighly: https://jenkins01.arangodb.biz/job/arangodb-devel-cppcheck/848/  
The warning was related to the `verifyHMAC` call: "Using object that points to local variable 'message' that is out of scope.. Pointer to container is created here.".

cppcheck is overly cautious in this context.  There's nothing indicating that message goes out of scope within the function. The only usage of `validateJwtHMAC256Signature` occurs here:
```cpp
std::string const message = header + "." + body;
if (!validateJwtHMAC256Signature(message, signature)) {
  //...
}
```
The underlying data referenced by the `string_view` remains valid for its entire use. Therefore, I suppressed this warning.

cppcheck: https://jenkins01.arangodb.biz/job/arangodb-ANY-cppcheck/980/

### DIFF
--- a/arangod/Auth/TokenCache.cpp
+++ b/arangod/Auth/TokenCache.cpp
@@ -396,6 +396,7 @@ bool auth::TokenCache::validateJwtHMAC256Signature(
   absl::WebSafeBase64Unescape(signatureWebBase64, &signature);
 
   READ_LOCKER(guard, _jwtSecretLock);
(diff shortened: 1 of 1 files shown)


MBkkt (approved):

jsteemann (approved):
LGTM

PR #19670, opened 2023-08-29
[CINFRA-828] Read from nondeclared collection

apetenchea:
### Scope & Purpose

This is an issue that, as far as I know until now, impacted only replication2. It caused the `testReadFromNonDeclaredCollection_V2` to become flaky: [ArangoError: collection '6010197' not found in transaction state](https://app.circleci.com/pipelines/github/arangodb/arangodb/14256/workflows/20d88381-d4ce-42e7-ae5f-31825cd22b11/jobs/324503).

The test scenario is the following:
1. Create 3 collections, out of which at least one uses a different sharding key.
2. Insert documents in each collection.
3. Open a transaction and try to read a document from each collection, without declaring the collections. The document lookup is done by `_key`.

The collection that uses a special sharding attribute is going to query all shards when searching a document whose only given identifier is `_key`. Since there are 3 dbservers and 4 shards, a dbserver is going to be responsible for 2 shards. This causes the coordinator to send two requests, each looking up the document in a another shard. Here is were the concurrency problem begins.

A `TransactionState` uses the `_collections` property to keep track of the collections that are necessary for responding to requests. _In case of read-only transactions, collections can be added lazily._ The code for adding a collection at runtime is in `TransactionState::addCollectionInternal` and looks like this:
```cpp
_collections.insert(_collections.begin() + position, trxColl);
```
In order to keep `_collections` a sorted vector, all the collections with a greater id are shifter to towards the end, making room for the new one.
In the meantime, `TransactionState::findCollection` is called from another thread. This iterates through `_collections`:
```cpp
for (auto* trxCollection : _collections)
```
Suppose we have the following collection ids in the list: `[1, 3]`. One thread calls `TransactionState::findCollection` looking for 3. After the for loop has already started, another thread (corresponding to a different shard), adds 2, hence the list becomes `[1, 2, 3]`. 3 is shifted past the visible end of the current for loop, which leads to the collection with id 3 being incorrectly labeled as "not found".

### Solution

In order to prevent such races from happening, I have changed the implementation to use recursive read-write locks and added locking wherever needed.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
```diff
@@ -173,6 +175,7 @@ std::lock_guard<std::mutex> ReplicatedRocksDBTransactionState::lockCommit() {
 
 /// @brief abort and rollback a transaction
 Result ReplicatedRocksDBTransactionState::doAbort() {
+  std::shared_lock lock{_collectionsLock};
```
jsteemann, line 178: Suggestion: move the lock acquisition into the `if (!mustBeReplicated())` branch.
This avoids acquiring the lock unnecessarily, and also makes it clearer for the reader that the lock only protects `_collections`, but not the mutation of `_hasActiveTrx`.

arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
```diff
@@ -67,6 +68,7 @@ Result ReplicatedRocksDBTransactionState::beginTransaction(
 
 /// @brief commit a transaction
 futures::Future<Result> ReplicatedRocksDBTransactionState::doCommit() {
+  std::shared_lock lock{_collectionsLock};
```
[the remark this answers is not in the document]
jsteemann: Additionally, this method can call `allCollections(...)`, which acquires the same lock again.
This can lead to deadlock.

arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
jsteemann: Same potential deadlock problem exists here as above.

apetenchea:
Thanks! I think I went a bit overboard with the locking.
I removed some locks based on the following:
- replication1 is not affected
- aql transactions are not affected
- the described problem does not occur when a transaction is created or finished, but only during collection reads
The code is a bit hard to figure correctly. I have seen how these methods call into each other, and locking is now a sensitive topic. If we really need to protect every access to `_collections`, I guess in the end our only hope will be `lib/Basics/RecursiveLocker.h`.
But for now, I stripped this PR down to a minimalistic approach.

MBkkt (approved):

arangod/RocksDBEngine/ReplicatedRocksDBTransactionState.cpp
```diff
@@ -71,6 +71,8 @@ futures::Future<Result> ReplicatedRocksDBTransactionState::doCommit() {
 
   if (!mustBeReplicated()) {
     Result res;
+    // We are aborting the transaction, the list of collections is complete - no
+    // need to acquire the lock.
     for (auto& col : _collections) {
```
mchacki, line 76: I would actually be in favor to acquire a read lock here, very close around the _collections usage, not for the full method.

With the APIs we use here we rely on the outside world to "know" when calling a method here is save.
E.g. we rely on the callers to protect es against a concurrent: addCollection() and commit() call.
For this example it may be ok. But for abort in my opinion it is not clear enough. (e.g. I could send many parallel requests, where the first throws an error triggering an abort to other participants, which than races with the initial search request)

I would suggest we better be save than sorry here.

arangod/StorageEngine/TransactionState.h
```diff
@@ -392,7 +397,7 @@ class TransactionState : public std::enable_shared_from_this<TransactionState> {
 
   // in case of read-only transactions collections can be added lazily.
   // this can happen concurrently, so for this we need to protect the list
-  mutable std::mutex _collectionsLock;
+  mutable std::shared_mutex _collectionsLock;
```
mchacki, line 400: Not sure about this, but I would have gone for a `basics::ReadWriteLock`
We also have RECURSIVE_READ_LOCKER and RECURSIVE_WRITE_LOCKER helpers
In case those methods stack and could acquire locks recursively.
As an alternative: We can make recursively called functions in a private/protected "NoLock" variant, that simply "assumes" a lock is held, all public methods however needs to setup proper locking.
A

mchacki (changes requested):
I would prefer to have this Class filled with proper read/write locking around _collections.
This way we can be sure it does not break again in the future.

apetenchea:
@mchacki I changed the implementation to use recursive read-write locks.

A small additional change that I did now was to address a //TODO left in `TransactionState::findCollectionOrPos`, regarding the use of binary search. Shouldn't make much of a difference, since this is a small vector anyways.

arangod/StorageEngine/TransactionState.cpp
```diff
@@ -324,29 +324,23 @@ TransactionCollection* TransactionState::findCollection(
...
+  if (_collections.empty()) {
+    return CollectionNotFound{0};
+  }
 
-    if (cid < trxCollection->id()) {
-      // collection not found
-      break;
-    }
-    // next
+  auto it = std::lower_bound(_collections.begin(), _collections.end(), cid,
```
mchacki, line 332: Nice :+1:

mchacki (approved):
LGTM, thank you for picking my suggestion, I think it is more save now

arangod/StorageEngine/TransactionState.cpp
```diff
@@ -82,9 +84,7 @@ TransactionCollection* TransactionState::collection(
   TRI_ASSERT(_status == transaction::Status::CREATED ||
              _status == transaction::Status::RUNNING);
 
-  std::lock_guard lock(_collectionsLock);
   auto collectionOrPos = findCollectionOrPos(cid);
```
jsteemann, line 87: We should rather acquire the read lock here, instead of acquring it inside `findCollectionOrPos()`. `findCollectionOrPos(...)` is protected (but could even be made private).
That probably then only leaves the ReplicatedRocksDBTranactionState::doCommit()` method requiring the lock recursively, which could also easily be fixed. And then we wouldn't need a recursive lock anymore.

apetenchea: Unfortunately, we need to use a recursive lock in `TransactionState::addCollectionInternal` regardless of the `findCollectionOrPos` implementation, because somewhere down the line it gets into `RocksDBTransactionState::isOnlyExclusiveTransaction`, which again has to iterate over the list of collections. So there would be at least one other method which needs the recursive lock.
I have tried a minimalistic approach in https://github.com/arangodb/arangodb/pull/19670/commits/b8f3b9fd638cd22d6db099b5dc68cf99dd0e70d2, but then I followed with a "better safe than sorry" approach.

PR #19671, opened 2023-08-29
[BTS-842] Attempt to somewhat improve error message

markuspf:
Improve error message when parsing of edge document fails.

This is not an ideal solution, but the error reporting facilities currently do not allow for an error message that reports exactly what the parse error is.

### DIFF
--- a/js/common/bootstrap/errors.js
+++ b/js/common/bootstrap/errors.js
@@ -94,7 +94,7 @@
     "ERROR_ARANGO_DATABASE_NAME_INVALID" : { "code" : 1229, "message" : "database name invalid" },
     "ERROR_ARANGO_USE_SYSTEM_DATABASE" : { "code" : 1230, "message" : "operation only allowed in system database" },
     "ERROR_ARANGO_INVALID_KEY_GENERATOR" : { "code" : 1232, "message" : "invalid key generator" },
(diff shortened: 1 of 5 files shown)


js/common/bootstrap/errors.js
```diff
@@ -94,7 +94,7 @@
     "ERROR_ARANGO_DATABASE_NAME_INVALID" : { "code" : 1229, "message" : "database name invalid" },
     "ERROR_ARANGO_USE_SYSTEM_DATABASE" : { "code" : 1230, "message" : "operation only allowed in system database" },
     "ERROR_ARANGO_INVALID_KEY_GENERATOR" : { "code" : 1232, "message" : "invalid key generator" },
-    "ERROR_ARANGO_INVALID_EDGE_ATTRIBUTE" : { "code" : 1233, "message" : "edge attribute missing or invalid" },
+    "ERROR_ARANGO_INVALID_EDGE_ATTRIBUTE" : { "code" : 1233, "message" : "Expecting both `_from` and `_to` to be defined in edge document and have format `collectionName/vertexKey` " },
```
Simran-B, line 97:
```suggestion
    "ERROR_ARANGO_INVALID_EDGE_ATTRIBUTE" : { "code" : 1233, "message" : "Expecting both `_from` and `_to` attributes to be defined in the edge document and have the format `collectionName/vertexKey` " },
```

PR #19675, opened 2023-08-29
[SEARCH-330] Use rocksdb MultiGet for materialize from search views

MBkkt:
### Scope & Purpose

~~* Benchmarks run~~
According to local benchmarks (ranges depends on query set size):
* MultiGet on same hot set faster than devel without cache in 1.5-2 times
* MultiGet on same hot set slower than devel with cache in 1.5-2 times
* MultiGet on cold set with enabled io_uring faster than devel in 2 times
* readFromSnapshot without cache faster in 1.5 times than readFromSnapshot with cache if was cache miss (probably mostly because populating cache)

~~Also by default collection cache are disabled, so I think best option doesn't touch cache in multiGet at all or touch only for get~~

TODO

* Maybe inverted index? Better in the separate PR
* Enable io_uring for rocksdb (probably needs patch for init and correct work with linux kernel older than 5.1)
* Remove ping-pong between iresearch skiplists and columnstore for common and merge view executors
* Remove ping-pong between different iresearch columns

Single thread execution could be like this:
For Heap/Merge:
1. take all needed search did + reader offset in sorted order
2. materialize for segments: primary key column, then stored field column i, ...
3. materialize documents from rocksdb via MultiGet

For just view, opposite to the current code, suggested to also take all needed documents from all segments, and only then materialize

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBMethods.h
```diff
@@ -55,7 +55,19 @@ class RocksDBMethods {
                                           rocksdb::Slice const&,
                                           rocksdb::PinnableSlice*,
                                           ReadOwnWrites,
-                                          rocksdb::Snapshot const*) = 0;
+                                          rocksdb::Snapshot const*) {
+    THROW_ARANGO_EXCEPTION_MESSAGE(TRI_ERROR_INTERNAL,
+                                   "RocksDBMethods does not provide MultiGet");
```
jsteemann, line 60:
```suggestion
                                   "RocksDBMethods does not provide GetFromSnapshot");
```

gnusi (approved):

arangod/Aql/MultiGet.h
```diff
@@ -0,0 +1,117 @@
...
+#include <rocksdb/status.h>
+
+#include <string>
+#include <vector>
+
+namespace arangodb::aql {
+
+#if defined(ARANGODB_ENABLE_MAINTAINER_MODE) && defined(__linux__) && \
+    defined(__aarch64__)
+inline constexpr bool kEnableMultiGet = false;
```
jsteemann, line 41: It looks strange that MultiGet is only disabled in maintainer mode on Linux on ARM, but not elsewhere.
There is probably a reason for this, but it should be mentioned here in a code comment.

MBkkt: Forgot to comment this, idea is to test old behavior at least on some configuration

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -1047,7 +1047,7 @@ Result RocksDBCollection::readFromSnapshot(
   }
 
   return lookupDocumentVPack(
-      trx, token, cb, /*withCache*/ true, readOwnWrites,
+      trx, token, cb, /*withCache*/ false, readOwnWrites,
```
jsteemann, line 1050: What was the motivation for changing the value of the `withCache` parameter here?
Not questioning that it is correct, but would like to understand why it was changed.

arangod/RocksDBEngine/RocksDBMethods.h
MBkkt: Fixed

arangod/Aql/MultiGet.h
MBkkt: Fixed

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: It feels strange for me that we populate cache with probably outdated results from snapshot.
It used only in inverted index now and in case when MultiGet disabled

arangod/Aql/MultiGet.h
jsteemann: Why don't we want to enable MultiGet globally?
Assuming that the MultiGet version works, is there a benefit in keeping the non-MultiGet version, but not making it available via configuration/opt-in somehow?

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -153,6 +157,40 @@ MaterializeExecutor<T, localDocumentId>::produceRows(
...
+      auto doc = it++;
+      auto const storage = (*doc)->storage;
+      if (ADB_UNLIKELY(!storage.isSet())) {
+        return MultiGetState::kNext;
+      }
+
+      auto& logical = *(*doc)->segment->collection;
+      _context.serialize(_trx, logical, storage);
+
+      (*doc)->segment = nullptr;
```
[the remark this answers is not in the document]
Dronplane: BTW this could be necessary if we would like to avoid multiget in case of single document.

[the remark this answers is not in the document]
MBkkt: > Is there any path that will make document read by single get even if kEnableMultiGet is set?

Now isn't. But implementation allow this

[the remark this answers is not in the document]
Dronplane: When benchmarking could you please check case like 
`FOR d IN 1..1000  FOR c IN someView SEARCH c.unique_field == d RETURN c` e.g. nested loop with single document as result of innner loop. If it is comparable with old implementation, then we don't need to bother with optimizing single document case.

arangod/Aql/MultiGet.h
MBkkt: Removed

arangod/Aql/MultiGet.h
```diff
@@ -0,0 +1,119 @@
...
+    auto& physical = static_cast<RocksDBCollection&>(*logical.getPhysical());
+    if (ADB_UNLIKELY(!methods)) {
+      methods = RocksDBTransactionState::toMethods(&trx, logical.id());
+    }
+
+    auto const objectId = physical.objectId();
+    auto const start = keysBuffer.size();
+    rocksutils::uint64ToPersistent(keysBuffer, objectId);
+    rocksutils::uint64ToPersistent(keysBuffer, id.id());
+    TRI_ASSERT(keysBuffer.size() - start == sizeof(uint64_t) * 2);
```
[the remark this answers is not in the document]
MBkkt: Done

arangod/Aql/MaterializeExecutor.cpp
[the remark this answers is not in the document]
MBkkt: Now such case will use SingleGet

arangod/Aql/MultiGet.h
```diff
@@ -0,0 +1,119 @@
...
+
+    keys.emplace_back(keysBuffer.data() + start, sizeof(uint64_t) * 2);
+  }
+};
+
+enum class MultiGetState : uint8_t {
```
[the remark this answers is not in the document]
MBkkt: Done

arangod/Aql/MultiGet.h
```diff
@@ -0,0 +1,119 @@
...
+inline constexpr bool kEnableMultiGet = true;
+#endif
+
+struct MultiGetContext {
+  RocksDBTransactionMethods* methods{};
+  StorageSnapshot const* snapshot{};
+  std::string keysBuffer;
+  std::vector<rocksdb::Slice> keys;
+  std::vector<rocksdb::Status> statuses;
+  std::vector<rocksdb::PinnableSlice> values;
```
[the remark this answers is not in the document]
MBkkt: Fixed, now it's fixed memory

arangod/Aql/IResearchViewExecutor.tpp
```diff
@@ -824,6 +721,55 @@ IResearchViewExecutorBase<Impl, ExecutionTraits>::skipRowsRange(
   return {reportedState, stats, call.getSkipCount(), upstreamCall};
 }
 
+template<typename Impl, typename ExecutionTraits>
+void IResearchViewExecutorBase<Impl, ExecutionTraits>::materialize() {
+  // TODO(MBkkt) unify it more with MaterializeExecutor
+  auto pkReadingOrder = _indexReadBuffer.getMaterializeRange(0);
```
Dronplane, line 727: Could we avoid using this method here? And directly only do `iota` for merge executors? Custom materialize range makes sense only for heap sort. So hardcoded call with 0  skip is really overkill. For simple executor that is also an overkill as entire buffer could be processed as is and `it` could be just a counter.

MBkkt: We can avoid this allocation only for simple view executor (not heap or merge).
(Because only this executor have order of produced documents that corresponding to snapshot)

But it needs to make quite a lot constexpr code.
So I think about make it (todo comment in getMaterializeRange) after fix possible ping pong between rocksdb and search.
And stabilize code more

Dronplane: BTW For merge sort we can keep this vector as a member to avoid reallocations multiple times. As most likely all batches except last one would have same size

arangod/Aql/MultiGet.h
```diff
@@ -0,0 +1,161 @@
...
+    auto it = end - _local;
+    for (size_t c = 0; it != end; ++it, ++c) {
+      TRI_ASSERT(*it == nullptr);
+      if (!_statuses[c].ok()) {
+        continue;
+      }
+      // TODO(MBkkt) Maybe we can avoid call byteSize() and rely on size()?
+      auto const* data = _values[c].data();
+      VPackSlice slice{reinterpret_cast<uint8_t const*>(data)};
+      const auto size = slice.byteSize();
```
MBkkt, line 153:
```suggestion
      auto const size = slice.byteSize();
```

arangod/Aql/IResearchViewExecutor.tpp
MBkkt: For heap we also can, just needs to create another buffer.

I think it should be in the MultiGet context:
Something like this:
```
union InputOutput {
  void* input;         // pointer to the localdocid + segment
  uint8_t* output;  // allocated output data, could be null
};

std::vector<InputOutput> _values;

~MultiGetContext() {
  for (size_t i = 0; i != _global; ++i) delete[] _values[i].output;
}

std::unique_ptr<uint8_t[]> extract(size_t i) {
  return {std::exchange(_values[i], nullptr)};
}
```

MBkkt: In such case we never need additional allocation

MBkkt: As I mentioned I'm aware about it but want to do it in the next step of materialize improvements

Dronplane: Ok. Let's merge this PR then in this state. And make more performance improvements in the next steps.

PR #19690, opened 2023-08-30
Show warning on startup when an old on-disk format is used.

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19689

Show warning on startup when an old on-disk format is used.
"Old on-disk format" here refers to our little-endian storage format for the RocksDB storage engine that we used in ArangoDB 3.2 and 3.3. Any new deployments that were set up since 3.4 will automatically have the better, big-endian storage format.
Using the little-endian storage format can have a severe impact on write performance, so it is not recommended to be used anymore.
The recommended way to change the storage format is to perform a logical dump of the deployment with arangodump, and restoring the dump into a fresh deployment using arangorestore.
Taking a hot backup and restoring it will not be sufficient, as this will keep the old storage format.

We expect that 0 users on ArangoGraph will be using the old storage format, since ArangoGraph only started after 3.4 was released. It may affect on-prem users though, provided that they are using a database instance that was created with a version older than 3.4.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: this PR
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

MBkkt (approved):
LGTM.

Not really corresponding to this PR, but I don't really understand why do we need runtime switch for endianness?
Because we don't want to make two binaries?

PR #19659, opened 2023-08-23
Bug fix/bts 1548

mchacki:
### Scope & Purpose

*Include a leading Collection in the new CollectionGroups. This is not required at all for Replication2 to work. However it is the easiest to make it a drop-in replacement for Replication 1 in the beginning. We reintroduce the limitation that we cannot drop a leading collection, but we gain that the behavior is the same, and we are protected against unexpected changes of "distributeShardsLike" responses. We may lift this restriction sometime later if we had time to adapt all our tools.*

One of the "tools" that used distributeShardsLike and relied on it to be a non-modifiable value is SmartGraphs, this caused: https://arangodb.atlassian.net/browse/BTS-1548

With this restriction test in above ticket is green again.

I did not add additional tests on purpose this behavior is covered by existing tests, which only run for replication 1.0 right now.

Side-Querst: This PR removes some functions that are not called anymore.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*
  - [ ] Backport for 3.9: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1548
- [ ] Design document:

PR #19695, opened 2023-08-31
[BTS-1589] Fix read after free

goedderz:
### Scope & Purpose

TSan found a data race by a write during `~NewStuffByDatabase`, when destroying a `shared_ptr<CollectionGroupPlanSpecification>`, and a read in `ClusterInfo::dropCollectionCoordinator`. While not 100% clear from the stack trace, the only candidate seems to be a `std::string const&` returned by `LogicalCollection::distributeShardsLike`, which is taken from `ClusterInfo::_collectionGroups` for replication v2 since https://github.com/arangodb/arangodb/pull/19659. Note that `_newStuffByDatabase` shares its `shared_ptr<CollectionGroupPlanSpecification>` with `_collectionGroups`.

Having `distributeShardsLike` return a copy of that string instead of a reference should solve that problem.

- [X] :hankey: Bugfix

#### Related Information

Introduced by https://github.com/arangodb/arangodb/pull/19659. No need for backports.

### DIFF
--- a/arangod/VocBase/LogicalCollection.cpp
+++ b/arangod/VocBase/LogicalCollection.cpp
@@ -356,7 +356,7 @@ replication::Version LogicalCollection::replicationVersion() const noexcept {
   return vocbase().replicationVersion();
 }
 
-std::string const& LogicalCollection::distributeShardsLike() const noexcept {
+std::string LogicalCollection::distributeShardsLike() const noexcept {
   if (ServerState::instance()->isCoordinator() &&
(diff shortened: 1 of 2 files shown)


goedderz:
https://jenkins01.arangodb.biz/job/arangodb-ANY-linux-san.x86-64/2080/

MBkkt (approved):

jsteemann (approved):
LGTM

PR #19703, opened 2023-09-01
Fix bugs around legacy little endian key encoding format

neunhoef:
### Scope & Purpose

This addresses https://arangodb.atlassian.net/browse/BTS-1541

Overview: In 3.2 and 3.3 we have created the ArangoDB/RocksDB instance
with little endian key encoding, which was a very bad idea for
performance. Newer versions use big endian key encoding, which is much
better. However, legacy databases created in 3.2 and 3.3 continue to use
the old encoding and new versions should be compatible with this.

This PR fixes some compatibility problems and ensures that newer code,
which relies on the old encoding cannot be executed any more, if the
underlying format is still little endian key encoding.

More specifically:

 - A method in RocksDBKeyBounds.cpp is fixed which computes a wrong
   key range for persistent indexes in the little endian key encoding
   case.
 - Another method which computes a range for the documents column family
   is disabled for the little endian key encoding case and now reports
   an error.
 - The parallel index creation (introduced in 3.10) will automatically
   reduce the number of threads used to 1 for little endian key
   encoding. This is necessary, since the parallel index creation relies
   on the fact that ranges in `_rev` values (`uint64_t`) correspond to
   RocksDB key ranges, which is not true for little endian key encoding.
 - The new arangodump protocol (introduced in 3.10) will no longer work
   and only error out for the little endian key encoding. Again, the
   algorithms used by that protocol will subtly (and silently) break
   for little endian key encoding because ranges in `_rev` values
   (`uint64_t`) do not correspond to ranges in RocksDB keys.
 - A new hidden command line option
   `--rocksdb.enforce-legacy-little-endian-keys` is added to be able
   to generate legacy databases. This is only for testing and should
   otherwise not be used.

- [*] :hankey: Bugfix

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11:
  - [ ] Backport for 3.10:

#### Related Information

- [*] Enterprise PR: there is no enterprise part to this
- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1541
- [*] Original issue: https://arangodb.atlassian.net/browse/ES-1697


CHANGELOG
```diff
@@ -1,6 +1,11 @@
 devel
 -----
 
+* Fixed BTS-1541: Old legacy little endian key format for RocksDB database
+  (created in ArangoDB 3.2 and 3.3) showed wrong behaviour on newer
+  versions. This fixes a persistent index corruption bug with the Old
```
```suggestion
  (created in ArangoDB 3.2 and 3.3) showed wrong behavior on newer
  versions. This fixes a persistent index corruption bug with the old
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -805,6 +806,25 @@ when disk size is very constrained and no replication is used.)");
               arangodb::options::Flags::OnSingle))
       .setIntroducedIn(31005);
 
+  options
+      ->addOption("--rocksdb.force-legacy-little-endian-keys",
+                  "If enabled and a new RocksDB database is generated, "
+                  "the old legacy little endian key encoding will be "
+                  "used. This is bad for performance and disables a "
+                  "few features like parallel index generation and is "
+                  "thus only sensible for testing purposes. Don't use!",
```
jsteemann, line 815: This description is repeated in the long description below. I suggest shortening this description to
```suggestion
                  "Force usage of legacy little endian key encoding when creating "
                  "a new RocksDB database directory. DO NOT USE IN PRODUCTION.",
```

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -805,6 +806,25 @@ when disk size is very constrained and no replication is used.)");
               arangodb::options::Flags::OnSingle))
       .setIntroducedIn(31005);
 
+  options
```
jsteemann, line 809: This option should not be made available in regular releases. I suggest we restrict it to either maintainer mode or failure tests, e.g. `#ifdef ARANGODB_ENABLE_MAINTAINER_MODE` or `#ifdef ARANGODB_ENABLE_FAILURE_TESTS`.

arangod/RocksDBEngine/RocksDBEngine.cpp
Simran-B: In hindsight, a better name for `.setLongDescription()` would be `.setDetails()` I suppose. Should we change this to avoid misunderstandings?

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -805,6 +806,27 @@ when disk size is very constrained and no replication is used.)");
...
+                  arangodb::options::makeFlags(
+                      arangodb::options::Flags::DefaultNoComponents,
+                      arangodb::options::Flags::Uncommon,
+                      arangodb::options::Flags::Experimental,
+                      arangodb::options::Flags::OnAgent,
+                      arangodb::options::Flags::OnDBServer,
+                      arangodb::options::Flags::OnSingle))
+      .setIntroducedIn(31200)
+      .setLongDescription(R"(If enabled and a new RocksDB database 
+is generated, the old legacy little endian key encoding will be used. This is bad for performance and disables a few features like parallel index generation and is thus only sensible for testing purposes. Don't use!)");
```
Simran-B, line 827:
```suggestion
      .setLongDescription(R"(If enabled and a new RocksDB database
is generated, the legacy little endian key encoding is used.

Only use this option for testing purposes! It is bad for performance and
disables a few features like parallel index generation!)");
```

arangod/RocksDBEngine/RocksDBEngine.cpp
neunhoef: Done.

arangod/RocksDBEngine/RocksDBEngine.cpp
neunhoef: Would be OK for me, but it does not seem to be urgent to me. What difference in meaning would be from `LongDescription` to `Details`?

neunhoef: Description changed.

PR #19705, opened 2023-09-01
bump file descriptors in dump/restore

jsteemann:
### Scope & Purpose

Bump file descriptors limit in arangodump & arangorestore.

* arangodump and arangorestore will now try to bump up the limit for the number of open file descriptors at startup, on Linux and macOS. This way the tools are allowed to open more files concurrently, which can be useful when dump/restore with many threads.

* Added startup option `--descriptors-minimum` to arangodump and arangorestore. This option can be used to set a lower bound value for the open file descriptors limit that is required to invoke arangodump/arangorestore. Setting the option to a value of 0 makes arangodump/arangorestore start regardless of the actual limit for the number of open file descriptors.

  This option is available on Linux and macOS only.

The file descriptors bumping code is now shared between arangod, arangodump, and arangorestore.
All 3 use the new `BumpFileDescriptorsFeature`, which resides in `lib/ApplicationFeatures`. Previously, the file descriptors bumping happened in `arangod/RestServer/FileDescriptorsFeature.cpp`, and was recently moved to `lib/Basics/FileDescriptors.cpp` so it could be shared across the executables.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


lib/ApplicationFeatures/BumpFileDescriptorsFeature.h
```diff
@@ -0,0 +1,67 @@
...
+
+#ifdef TRI_HAVE_GETRLIMIT
+namespace arangodb {
+class LoggerFeature;
+
+namespace application_features {
+class GreetingsFeaturePhase;
+}
+
+class BumpFileDescriptorsFeature
```
MBkkt, line 39: FileDescriptorsFeature looks very similar
Could we reuse it?

arangod/RestServer/FileDescriptorsFeature.cpp
```diff
@@ -77,20 +73,13 @@ FileDescriptorsFeature::FileDescriptorsFeature(Server& server)
...
 void FileDescriptorsFeature::collectOptions(
     std::shared_ptr<ProgramOptions> options) {
-  options->addOption(
-      "--server.descriptors-minimum",
-      "The minimum number of file descriptors needed to start (0 = no minimum)",
-      new UInt64Parameter(&_descriptorsMinimum),
-      arangodb::options::makeFlags(arangodb::options::Flags::DefaultNoOs,
-                                   arangodb::options::Flags::OsLinux,
-                                   arangodb::options::Flags::OsMac));
-
```
MBkkt, line 93: Ok, I see, you separated them

client-tools/Dump/arangodump.cpp
```diff
@@ -80,6 +81,10 @@ int main(int argc, char* argv[]) {
         []<typename T>(auto& server, TypeTag<T>) {
           return std::make_unique<T>(server);
         },
+        [](ArangoDumpServer& server, TypeTag<BumpFileDescriptorsFeature>) {
+          return std::make_unique<BumpFileDescriptorsFeature>(
+              server, "--descriptors-minimum");
```
MBkkt, line 86: What is the point to have ability to specify different name for same option for different binaries?

MBkkt (approved):

PR #19707, opened 2023-09-01
minor internal improvements for arangodump

jsteemann:
### Scope & Purpose

Small internal improvements for arangodump:

- add trace log messages to see what it is doing
- save collection name in shard => file map to avoid extra lookup later
- rename a function to make its purpose more clear
- make sure started threads are properly shut down even if an exception happens

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/client-tools/Dump/DumpFeature.cpp
+++ b/client-tools/Dump/DumpFeature.cpp
@@ -55,6 +55,7 @@
 
 #include <chrono>
 #include <thread>
+#include <unordered_map>
 
 #include <absl/strings/str_cat.h>
 #include <velocypack/Collection.h>
@@ -83,7 +84,7 @@ std::string serverLabel(std::string const& server) {
   return absl::StrCat(" on server '", server, "'");
 }
 
-constexpr std::string_view getSuffix(bool useVPack) noexcept {
+constexpr std::string_view getDatafileSuffix(bool useVPack) noexcept {
   std::string_view suffix("json");
   if (useVPack) {
     suffix = "vpack";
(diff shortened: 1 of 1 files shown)


client-tools/Dump/DumpFeature.cpp
```diff
@@ -1854,9 +1867,10 @@ void DumpFeature::ParallelDumpServer::runWriterThread() {
         FATAL_ERROR_EXIT();
       }
 
-      auto file = fileProvider->getFile(it2->second.collectionName);
-      filesByShard.emplace(shardId, file);
-      return file;
+      std::string collectionName = it2->second.collectionName;
```
MBkkt, line 1870: Maybe string_view? It looks like shards map destroyed not early than filesByShard

jsteemann: I also thought about using an `std::string_view` initially, and I think it could be done. 
But it makes things harder to reason about. For example, the `collectionName` is handed to `DumpFileProvider::getFiles(std::string const&)`. So this method would need to change its signature.  And then this method does a lookup using that filename in an associative container (`_collectionInfo`), which is an `std::map<std::string, ...>`. So I would need to change this too.
Additionally returning an std::string_view from a function/lambda as we would do here if we change the return type of the lambda makes it harder for the caller to reason about until when the std::string_view will remain valid, and it is possible to introduce ownership problems should the code change in the future. Returning an std::string is self-explanatory and there are no ownership issues. So I prefered simplicity over performance here, which should be ok, as this is not really a hot function compared to what else is happening when dumping collections.

MBkkt (approved):

PR #19717, opened 2023-09-04
Refuse server startup with outdated little-endian on-disk format.

jsteemann:
### Scope & Purpose

Docs PR: https://github.com/arangodb/docs/pull/1465

* Refuse server startup with outdated little-endian on-disk format. The little-endian on-disk format was used for deployments that were created with either ArangoDB 3.2 or 3.3 when using the RocksDB storage engine. Since ArangoDB 3.4, a big-endian on-disk format was used for the RocksDB storage engine, which is more performant. Deployments that were set up with the RocksDB storage engine using ArangoDB 3.2 or 3.3 and that have been upgraded since then will still use the old format. This should not affect many users, because the default storage engine in ArangoDB 3.2 and 3.3 was the MMFiles storage engine. Furthermore, deployments that have been recreated from an arangodump since ArangoDB 3.4 will not be affected, because restoring an arangodump into a fresh deployment will also make ArangoDB use the big-endian on-disk format.

  When the outdated little-endian on-disk format is detected at startup, the server will bail out with a descriptive error message. The recommend way to migrate the database from the little-endian storage format to the big-endian storage format is to

  1. create a full logical backup of the database using arangodump
  2. stop the database servers in the deployment
  3. wipe the existing database directories
  4. restart the servers in the deployment
  5. restore the logical dump into the deployment

  It will not be sufficient to take a hot backup of a little-endian dataset
  and restore it, because when restoring a hot backup, the original database
  format will be restored as it was at time of the backup.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs/pull/1465
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,34 @@
 devel
 -----
 
+* Refuse server startup with outdated little-endian on-disk format.
+  The little-endian on-disk format was used for deployments that were created
+  with either ArangoDB 3.2 or 3.3 when using the RocksDB storage engine.
+  Since ArangoDB 3.4, a big-endian on-disk format was used for the RocksDB
+  storage engine, which is more performant.
+  Deployments that were set up with the RocksDB storage engine using ArangoDB
+  3.2 or 3.3 and that have been upgraded since then will still use the old
+  format. This should not affect many users, because the default storage
+  engine in ArangoDB 3.2 and 3.3 was the MMFiles storage engine. Furthermore,
+  deployments that have been recreated from an arangodump since ArangoDB 3.4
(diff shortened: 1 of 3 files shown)


neunhoef (approved):
LGTM

MBkkt (approved):

PR #19719, opened 2023-09-04
Fix bugs around legacy little endian key encoding format

neunhoef:
### Scope & Purpose

This addresses https://arangodb.atlassian.net/browse/BTS-1541

Overview: In 3.2 and 3.3 we have created the ArangoDB/RocksDB instance
with little endian key encoding, which was a very bad idea for
performance. Newer versions use big endian key encoding, which is much
better. However, legacy databases created in 3.2 and 3.3 continue to use
the old encoding and new versions should be compatible with this.

This PR fixes some compatibility problems and ensures that newer code,
which relies on the old encoding cannot be executed any more, if the
underlying format is still little endian key encoding.

More specifically:

 - A method in RocksDBKeyBounds.cpp is fixed which computes a wrong
   key range for persistent indexes in the little endian key encoding
   case.
 - Another method which computes a range for the documents column family
   is disabled for the little endian key encoding case and now reports
   an error.
 - The parallel index creation (introduced in 3.10) will automatically
   reduce the number of threads used to 1 for little endian key
   encoding. This is necessary, since the parallel index creation relies
   on the fact that ranges in `_rev` values (`uint64_t`) correspond to
   RocksDB key ranges, which is not true for little endian key encoding.
 - The new arangodump protocol (introduced in 3.10) will no longer work
   and only error out for the little endian key encoding. Again, the
   algorithms used by that protocol will subtly (and silently) break
   for little endian key encoding because ranges in `_rev` values
   (`uint64_t`) do not correspond to ranges in RocksDB keys.
 - A new hidden command line option
   `--rocksdb.enforce-legacy-little-endian-keys` is added to be able
   to generate legacy databases. This is only for testing and should
   otherwise not be used.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports

#### Related Information

- [*] Enterprise PR: there is no enterprise part to this
- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1541
- [*] Original issue: https://arangodb.atlassian.net/browse/ES-1697
- [*] Devel PR: https://github.com/arangodb/arangodb/pull/19703

### DIFF
--- a/arangod/RocksDBEngine/RocksDBKeyBounds.cpp
+++ b/arangod/RocksDBEngine/RocksDBKeyBounds.cpp
@@ -415,7 +415,7 @@ RocksDBKeyBounds::RocksDBKeyBounds(RocksDBEntryType type, uint64_t first,
       uint8_t const maxSlice[] = {0x02, 0x03, 0x1f};
       VPackSlice max(maxSlice);
 
-      _internals.reserve(2 * sizeof(uint64_t) + (second ? max.byteSize() : 0));
+      _internals.reserve(2 * sizeof(uint64_t) + max.byteSize());
       uint64ToPersistent(_internals.buffer(), first);
       _internals.separate();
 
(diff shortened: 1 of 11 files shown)


arangod/RocksDBEngine/RocksDBKeyBounds.cpp
```diff
@@ -427,9 +427,18 @@ RocksDBKeyBounds::RocksDBKeyBounds(RocksDBEntryType type, uint64_t first,
...
-        // termination case, as it will be in the next prefix
-        uint64ToPersistent(_internals.buffer(), first + 1);
+        // in case of forward iteration, we can use the next objectId as
+        // a quick termination case, as it will be in the next prefix,
+        // however, this trick only works when we have the big endian
+        // data format:
+        if (rocksDBEndianness == RocksDBEndianness::Big) {
+          uint64ToPersistent(_internals.buffer(), first + 1);
+        } else {
+          // little endian:
```
hkernbach, line 437: We could add an assert here, in case there will be a third type some when in the future,
e.g. "TRI_ASSERT(rocksutils::rocksDBEndianness == RocksDBEndianness::Little);"

arangod/RocksDBEngine/RocksDBUpgrade.cpp
```diff
@@ -119,8 +120,10 @@ void arangodb::rocksdbStartupVersionCheck(ArangodServer& server,
       }
     }
   } else {
-    // new DBs are always created with Big endian data-format
-    endianess = RocksDBEndianness::Big;
+    // new DBs are always created with Big endian data-format, Unless
+    // forced by a command line option:
```
hkernbach, line 124: This is only true in maintainer mode, isn't it? Should this code path not also get a "#ifdef ARANGODB_ENABLE_MAINTAINER_MODE" flag?

hkernbach (approved):
LGTM, only left two minor comments.

PR #19725, opened 2023-09-04
fix issues with Pregel garbage-collection

jsteemann:
### Scope & Purpose

Fix issues with Pregel garbage-collection

When dropping a database, the Pregel conductor will still hold a ref count for the already-dropped database, until the conductor is garbage-collected.
Garbage-collection for conductors only happens every ~20s, and by default only after a timeout that is too long for being useful in integration tests.
This PR changes the conductor garbage collection so that conductors of dropped databases can be cleaned up even before their TTL has expired.

The PR also adds an extra round of garbage collection in the stop phase of the PregelFeature, because when performing the garbage collection in the unprepare phase only, assertion failures can occur. The problem is that garbage-collecting conductors can post to the scheduler, which is already forbidden during the unprepare step. The hope is that when garbage-collecting in the stop phase, we can eliminate all conductors before we enter the unprepare phase, so we can work around the assertion failure.

Garbage collection also turned up an issue, because the garbage collection first tries to cancel all GC-able conductors. Conductor::cancel() can throw however, which previously led to no conductors being garbage collected. The PR changes the logic so that when cancel() throws for a single conductor, the garbage collection still goes on and cancels all other GC-able conductors and also removes them from the internal inventory.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Improve Pregel garbage collection for dropped databases.
+
 * Refuse server startup with outdated little-endian on-disk format.
   The little-endian on-disk format was used for deployments that were created
   with either ArangoDB 3.2 or 3.3 when using the RocksDB storage engine.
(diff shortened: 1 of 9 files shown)


neunhoef (approved):
LGTM

PR #19728, opened 2023-09-05
change container type back to std::unordered_map

jsteemann:
### Scope & Purpose

Change container type back to std::unordered_map because it has more desirable iteration order properties than absl::FlatHashMap for our internal testing.
Using a FlatHashMap would be ok, but we seem to have several tests that rely on two snapshots of variables being returned in the same order. This may not be true when using a FlatHashMap.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


MBkkt (changes requested):
Why couldn't we fix tests?
unordered_map also doesn't have such property, it's property of implementation

MBkkt (commented):
> This is something that std::unordered_map does not do.

It's not property of standard.
It's property of implementation.
I think we can merge it to quickly fix tests, but maybe we at least create bts ticket for these tests?

mpoeter (approved):

PR #19729, opened 2023-09-05
Improve memory accounting for RocksDB transactions

jsteemann:
### Scope & Purpose

This PR tries to improve the memory tracking for keys locked in RocksDB transactions, by improving the calculation formula for key memory usage.
Locked keys are stored as std::string inside RocksDB's lock map data structure, and we previously counted the memory usage as `max(key size, sizeof(std::string))`. This can lead to underreporting of memory usage for longer strings, which exceed std::string's SSO buffer. Typically the SSO buffer is only 15 bytes long, but the size of std::string is 40. In the worst case, we counted a key of size 32 with a memory usage of only 32, whereas in reality it would rather be 32 (size of std::string) + 32 (size of the value) + 1 (NUL terminator) + dynamic memory allocation overhead.
This is now taken into account.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RocksDBEngine/RocksDBMethods.cpp
```diff
@@ -0,0 +1,94 @@
...
+// and is not guaranteed to be 15. we checked the SSO buffer
+// size for all current relevant implementation though:
+// - g++-11: 15
+// - g++-12: 15
+// - MSVC: 15
+// - clang++-14: 15
+// TODO: replace this constant with the expression
+//   std::string{}.capacity();
+// once all compilers provide constexpr std::string::capacity
+// implementations (clang 14 does, gcc 12 does, but gcc 11 doesn't).
```
MBkkt, line 43: // gnu libstdc++ -- 15
// msvc stl -- 15
// llvm libc++ -- 22

The interesting moment it's easy to detect what implementation are used by their sizeof
sizeof(std::string)
// gnu libstdc++ -- 32
// msvc stl -- 32
// llvm libc++ -- 24

MBkkt (approved):
Comment about sso, isn't really correct, but for now it doesn't really matter.
So I think we can merge it

PR #19730, opened 2023-09-06
Fix MemoryAccounting Agency Node Array

maierlars:
### Scope & Purpose
Due to an oversight, velocypack data in arrays was not accounted for.

Should improve memory accounting and fix an issue reported in https://arangodb.atlassian.net/browse/BTS-1584


jsteemann (approved):
Thanks.
Also verified manually that it works as expected and fixes the issue described in https://arangodb.atlassian.net/browse/BTS-1584.

PR #19741, opened 2023-09-08
Chore Cinfra: Bug-Fixes on Replication2 only

mchacki:
### Scope & Purpose

* This is a collection of some accumulated items:
1) Moved Index create/drop from ClusterInfo into a Static Class unmodified. ClusterInfo was original designed to just be a cache, but by now does actual work. With this change we split concerns. It is also preparation to add targeted implementation for Replication2 in this code.
2) Include CollectionGroup Properties into Shard Properties on Maintenance action. This way a shard will respond correctly to questions like "waitForSync". The actual work on WaitForSync is implemented somewhere else and already worked, this change mostly effects return codes in HTTP API
3) Detect and apply property changes of Collections in Target (replication2) Now schema and computedValues.
*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1372
- [ ] Design document:


arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
```diff
@@ -601,6 +603,27 @@ struct TransactionBuilder {
...
+    auto tmp = env.write();
+    auto allProps = velocypack::serialize(action.spec);
+    ADB_PROD_ASSERT(allProps.isObject());
+    for (auto const& it : VPackObjectIterator(allProps.slice())) {
+      tmp = std::move(tmp).emplace_object(
+          basics::StringUtils::concatT("/arango/Plan/Collections/", database,
+                                       "/", action.cid, "/",
+                                       it.key.stringView()),
+          [&](VPackBuilder& builder) {
+            velocypack::serialize(builder, it.value);
```
apetenchea, line 616: Question: if the value is null, does that write a `null` into the agency, or removes the property? Do we care about it, or it's not a problem?

apetenchea (approved):
LGTM

arangod/Replication2/Supervision/CollectionGroupSupervision.cpp
mchacki: That is a good spot!

If the value "null" is added for schema, then schema will not be serialized (as null here is equal to non-existent with the std::optional)
This way the update actually does not remove the schema (or set to null) with this method.
Will fix

mchacki: Fixed: a6260d57ff6

apetenchea (approved):

PR #19746, opened 2023-09-08
JoinNode +  Executor

maierlars:
### Scope & Purpose
This PR implements the bare minimum to have a join node for single server deployments. It only has a single generic join strategy. 

The optimizer rule currently only detects queries that look like this:
```
for doc1 in col1
   sort doc1.x
   for doc2 in col2
        filter doc1.x == doc2.x
        return [doc1, doc2]
```

The optimizer rule is disabled by default.
Projections are not yet supported. More complicated conditions are not yet supported.


MBkkt (commented):
Interesting.

But is nested_loop_join is nested loop join algorithm?
It all looks like different implementation of merge join.
I don't see O(n*m) for any algorithm here

Could you explain for me is it expected, please?

tests/Aql/IndexMergeJoin.cpp
```diff
@@ -0,0 +1,337 @@
...
+  }
+};
+
+struct EvenOddPattern : PatternBase {
+  void operator()(RocksDBInstance& db) {
+    generateData(db, scale, 1, [x = 0]() mutable { return 2 * ++x; });
+    generateData(db, scale, 2, [x = 0]() mutable { return 2 * ++x + 1; });
+  }
+};
+
```
MBkkt, line 161:
```
struct BigGapsPattern : PatternBase {
  void operator()(RocksDBInstance& db) {
    generateData(db, scale, 1, [x = uint64_t{0}]() mutable { return ++x; });
    generateData(db, scale / 100, 2, [x = uint64_t{0}]() mutable { return ++x * 1000; });
  }
};
```

maierlars: To make this fair, the first one has to be the smaller collection. The optimizer will always try to use the smaller collection first.

MBkkt: Understand. The idea was to check case when you have to pretty sparse match group.

Btw if this PR compare seek vs next approach, and seek overhead is big, maybe something hybrid will be good? Do Next some n times before Seek?
If "swap" or "equal" happened reset the counter.

maierlars:
Regarding you question: what is called `nested_loops` is what our product management called the thing arangodb does currently. It is not O(n * m). In fact it is a binary search in the bigger list for every element of the smaller list. Which is O(n * log m). The difference to a merge join is that we only advance the first iterator by one.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -2858,4 +2859,155 @@ void RocksDBVPackIndex::warmupInternal(transaction::Methods* trx) {
...
+        break;
+      }
+      into[idx] = k;
+    }
+  }
+
+  bool seek(std::span<VPackSlice> span) override {
+    loadKey(span);
+    RocksDBKey rkey;
+    rkey.constructUniqueVPackIndexValue(_index->objectId(), _builder.slice());
```
jsteemann, line 2934: Potential performance optimization: we could make the RocksDBKey value an instance member, so it can be recycled to create multiple keys.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -2858,4 +2859,155 @@ void RocksDBVPackIndex::warmupInternal(transaction::Methods* trx) {
...
+    return position(span);
+  }
+};
+
+}  // namespace
+
+std::unique_ptr<AqlIndexStreamIterator> RocksDBVPackIndex::streamForCondition(
+    transaction::Methods* trx, IndexStreamOptions const& opts) {
+  if (!supportsStreamInterface(opts)) {
+    THROW_ARANGO_EXCEPTION(TRI_ERROR_BAD_PARAMETER);
```
jsteemann, line 2976: Let's use TRI_ERROR_INTERNAL here, because it is not really an error the user could fix. We should also use `THROW_ARANGO_EXCEPTION_MESSAGE(code, msg)` with a more descriptive error message. Only getting "Bad parameter" will otherwise be hard to debug if this is ever called.
```suggestion
    THROW_ARANGO_EXCEPTION(TRI_ERROR_BAD_PARAMETER);
```

js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -1169,7 +1164,7 @@ function processQuery(query, explain, planIndex) {
     var boundValue = bound.isConstant ? value(JSON.stringify(bound.bound)) : buildExpression(bound.bound);
     return attribute(attr) + ' ' + operators[bound.include ? 1 : 0] + ' ' + boundValue;
   };
-
+/*
```
jsteemann, line 1167: We should remove the commented-out code for good, but we can do this in a separate PR if you prefer.

js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -2143,11 +2161,36 @@ function processQuery(query, explain, planIndex) {
...
+      node.indexInfos.forEach((info) => {
+        let line = nodePrefix(node);
+        if (profileMode) {
+          // TODO
+          /*
+          line += pad(1 + maxCallsLen - String(node.calls).length) + value(node.calls) + '   ' +
+            pad(1 + maxItemsLen - String(node.items).length) + value(node.items) + '   ' +
+            pad(1 + maxFilteredLen - String(node.filtered).length) + value(node.filtered) + '   ' +
+            pad(1 + maxRuntimeLen - runtime.length) + value(runtime) + '   ';
+            */
```
jsteemann, line 2176: Profiling output doesn't work yet. I think it is not hard to fix this. As our aim is to run performance tests with the JoinNode, I would be for adding support for profiling output to this PR.

tests/js/server/aql/aql-index-join-noncluster.js
```diff
@@ -0,0 +1,201 @@
...
+              FILTER doc1.x == doc2.x
+              RETURN [doc1, doc2]
+      `);
+
+      assertEqual(result.length, 100);
+      for (const [a, b] of result) {
+        assertEqual(a.x, b.x);
+      }
+    },
+/*
```
jsteemann, line 177: Is this test intentionally commented out? Can it be activated for this PR already?

maierlars: The problem here is that the optimizer somehow does not convert the second for + sort into an index scan. I haven't looked into details but it appears to be a more general problem.

jsteemann (approved):
LGTM

PR #19750, opened 2023-09-11
APM-780: limits for arangodump

jsteemann:
### Scope & Purpose

https://arangodb.atlassian.net/browse/APM-780

APM-780: limits for (parallel) arangodump

* Added the following server startup options that can be used to limit
  the resource usage of parallel arangodump invocations:

  - `--dump.max-memory-usage`: Maximum memory usage (in bytes) to be 
    used by the server-side parts of all ongoing arangodump invocations.
    This option can be used to limit the amount of memory for prefetching
    and keeping results on the server side when arangodump is invoked
    with the `--parallel-dump` option. It does not have an effect for
    arangodump invocations that did not use the `--parallel-dump` option.
    Note that the memory usage limit is not exact and that it can be 
    slightly exceeded in some situations to guarantee progress.
  - `--dump.max-docs-per-batch`: Maximum number of documents per batch 
    that can be used in a dump. If an arangodump invocation requests
    higher values than configured here, the value will automatically be
    capped to this value. Will only be followed for arangodump invocations
    that use the `--parallel-dump` option.
  - `--dump.max-batch-size`: Maximum batch size value (in bytes) that 
    can be used in a dump. If an arangodump invocation requests larger
    batch sizes than configured here, the actual batch sizes will be capped
    to this value. Will only be followed for arangodump invocations that
    use the `--parallel-dump` option.
  - `--dump.max-parallelism`: Maximum parallelism (number of server-side
    threads) that can be used in a dump. If an arangodump invocation requests
    a higher number of prefetch threads than configured here, the actual 
    number of server-side prefetch threads will be capped to this value. 
    Will only be followed for arangodump invocations that use the 
    `--parallel-dump` option.

* Added the following metrics to observe the behavior of parallel
  arangodump on the server:

  - `arangodb_dump_memory_usage`: Current memory usage of all ongoing 
    arangodump operations on the server.
  - `arangodb_dump_ongoing`: Number of currently ongoing arangodump 
    operations on the server.
  - `arangodb_dump_threads_blocked_total`: Number of times a server-side 
    dump thread was blocked because it honored the server-side memory 
    limit for dumps.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/APM-780
- [ ] Design document:


arangod/RocksDBEngine/RocksDBDumpContext.cpp
```diff
@@ -312,6 +367,9 @@ std::shared_ptr<RocksDBDumpContext::Batch const> RocksDBDumpContext::next(
     return it->second;
   }
 
+  // release lock temporarily because _channel.pop() can block
+  guard.unlock();
+
```
maierlars, line 372: Releasing this lock can cause a retried request to associate another batch with a already used batchId.

For example, assume on thread blocked on a call `next(5, 4)` on `_channel.pop`. Now, because of some network issues, arangodump retries the same request. We get another thread that enters `next(5, 4)`.

Now we have two threads, each will pop a new batch (say `A` and `B`) from the channel, but both associate this batch with id `5`. Assume that `B` is associated with `5` and the response for `A` is lost. Then first we get an error message that we have a duplicate batch id (although everything went fine) and batch `A` can never be received.

maierlars (changes requested):

arangod/RestServer/DumpLimitsFeature.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // if we have at least 4GB of RAM, the default size is (RAM - 2GB) * 0.2
+    return static_cast<uint64_t>(
+        (PhysicalMemory::getValue() - (static_cast<uint64_t>(2) << 30)) * 0.2);
+  }
+  if (PhysicalMemory::getValue() >= (static_cast<uint64_t>(2) << 30)) {
+    // if we have at least 2GB of RAM, the default size is 256MB
+    return (static_cast<uint64_t>(256) << 20);
+  }
+  // for everything else the default size is 64MB
+  return (static_cast<uint64_t>(64) << 20);
```
MBkkt, line 49: Why do we need 3 cases? I understand two: one larger than 4gb, another is smaller.
But is it common that we between 2 and 4 gb of memory?

arangod/RocksDBEngine/RocksDBDumpContext.cpp
```diff
@@ -238,14 +274,23 @@ RocksDBDumpContext::RocksDBDumpContext(RocksDBEngine& engine,
...
               }
 
-              handleWorkItem(std::move(workItem));
+              do {
+                try {
+                  handleWorkItem(workItem);
+                  break;
+                } catch (basics::Exception const& ex) {
+                  if (ex.code() != TRI_ERROR_RESOURCE_LIMIT) {
+                    throw ex;
```
MBkkt, line 290:
```suggestion
                    throw;
```

arangod/RocksDBEngine/RocksDBDumpContext.cpp
```diff
@@ -238,14 +274,23 @@ RocksDBDumpContext::RocksDBDumpContext(RocksDBEngine& engine,
...
+                  }
+                }
+              } while (!stopped());
```
MBkkt, line 293: maybe some limit for iteration count here?

arangod/RocksDBEngine/RocksDBDumpContext.cpp
jsteemann: Pushed a change that removes the unlock.

arangod/RestServer/DumpLimitsFeature.cpp
jsteemann: Removed one of the cases.

arangod/RocksDBEngine/RocksDBDumpContext.cpp
jsteemann: Changed.

arangod/RocksDBEngine/RocksDBDumpContext.cpp
jsteemann: You are right, there is currently no limit on the number of iterations.
However, the iterations will effectively be bounded because of the `!stopped()` condition.
Each dump context has a TTL, and if no new requests come in for a dump context (e.g. because all threads are stalled), then it will eventually time out and `stopped()` will become true.
Then all threads that cannot make any progress will be stopped.

maierlars (approved):

PR #19752, opened 2023-09-12
[BTS-1226] Added minMaxDepth to path validator.

maierlars:
### Scope & Purpose
BTS: https://arangodb.atlassian.net/browse/BTS-1226

Given the following query:
```
  for i in a
    for v, e, p in 2 OUTBOUND i e, f
        filter v._key == "3"
        return k
```
The path validator will load all vertices, even those that are not at depth 2, to check the post condition. Thus it might access a collection that is not listed in the cluster in the `with` statement. 

This PR changes that behavior and only tests for the post condition on vertices that are at min depth. (and thus only loads those documents)


maierlars:
This PR does not solve the underlying issue and makes things needlessly complicated.

PR #19753, opened 2023-09-12
Added cppcheck run on pr runs

CryptoNinjaGeek:
### Scope & Purpose

Added cppcheck to run on pr run for each build type to ensure all code gets covered.
cppcheck is now using the project created before compilation to only cover required files and test based on compile options and flags.


.circleci/base_config.yml
```diff
@@ -358,6 +412,10 @@ workflows:
           name: build-ce-x64
           preset: community-pr
           enterprise: false
+      - run-cppcheck:
+          name: cppcheck
+          requires:
+            - build-ce-x64
```
KVS85, line 418: EE should be used.

KVS85 (changes requested):

.circleci/base_config.yml
CryptoNinjaGeek: @KVS85 this is still a draft PR

CryptoNinjaGeek: We run for both as we now use project files, which can potential have different files compiled because of community and enterprise differences.

KVS85: There is always CE + EE for every single CE change (with the same base branch name in case it's not set explicitly because we ship both). If a cppcheck run for a CE branch doesn't show a problem and not stop a PR that doesn't mean EE cppcheck won't have it when an EE nightly or a package is built.

KVS85 (changes requested):

.circleci/base_config.yml
```diff
@@ -331,6 +332,59 @@ jobs:
...
+            cmake . 
+            make -j4 FILESDIR=/usr install 
+            ln -s /root/workspace/build/ /root/project/
+      - run:
+          name: Install cppcheck junit formator
+          command: |
+            cd /root 
+            wget https://bootstrap.pypa.io/get-pip.py
+            python3 get-pip.py
+            pip install cppcheck-junit
```
KVS85, line 369: Why do we install cppcheck every single run (which requires CI time) instead of persisting it within build container?

Also, build containers have to have specific suffix now according to https://github.com/arangodb/arangodb/tree/devel/scripts/docker.

KVS85 (changes requested):

.circleci/base_config.yml
CryptoNinjaGeek: Moved to base image in separate pr, waiting with review/merge of this until after that is approved and new images build.

KVS85 (approved):

.circleci/base_config.yml
```diff
@@ -377,6 +418,10 @@ workflows:
           enterprise: true
           publish-artifacts: false
           build-tests: false
+      - run-cppcheck:
+          name: cppcheck
+          requires:
+            - build-ee-x64
```
mpoeter, line 424: I think this should rather run as part of the lint workflow. I don't see why we need to depend on the build job - we could simply run the cmake configure step ourselves.

mpoeter (approved):
Small suggestion, but otherwise LGTM.

I would also be fine to leave it as is and maybe change this later.

PR #19754, opened 2023-09-12
BTS-1601

maierlars:
### Scope & Purpose
BTS: https://arangodb.atlassian.net/browse/BTS-1601
adhov performance tests - old version
```
http://localhost:8530@_system> for (num of [100, 1000, 2000, 4000, 8000, 16000, 32000, 64000]){ let start = Date.now(); db._query(`return MERGE(for i in 1..${num} return {[i]: i})`); let end = Date.now(); let time = end - start; print(num, time); }
100 3
1000 66
2000 228
4000 838
8000 2995
16000 17443
32000 68708
64000 234479
```
new version:
```
http://localhost:8530@_system> for (num of [100, 1000, 2000, 4000, 8000, 16000, 32000, 64000]){ let start = Date.now(); db._query(`return MERGE(for i in 1..${num} return {[i]: i})`); let end = Date.now(); let time = end - start; print(num, time); }
100 2
1000 4
2000 7
4000 13
8000 23
16000 51
32000 83
64000 103

```

This *only* improves the version of the **array non-recursive merge**. All other merge variants are untouched.

### DIFF
--- a/arangod/Aql/Functions.cpp
+++ b/arangod/Aql/Functions.cpp
@@ -944,18 +944,45 @@ AqlValue mergeParameters(ExpressionContext* expressionContext,
   if (initial.isArray() && n == 1) {
     // special case: a single array parameter
     // Create an empty document as start point
-    builder.openObject();
-    builder.close();
-    // merge in all other arguments
-    for (VPackSlice it : VPackArrayIterator(initialSlice)) {
-      if (!it.isObject()) {
-        registerInvalidArgumentWarning(expressionContext, funcName);
-        return AqlValue(AqlValueHintNull());
+    if (!recursive) {
+      // Fast path for large arrays
+      containers::FlatHashMap<std::string_view, VPackSlice> attributes;
+
+      // first we construct a map holding the latest value for a key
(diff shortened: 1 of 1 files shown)


arangod/Aql/Functions.cpp
```diff
@@ -944,18 +944,45 @@ AqlValue mergeParameters(ExpressionContext* expressionContext,
...
+      // Fast path for large arrays
+      containers::FlatHashMap<std::string_view, VPackSlice> attributes;
+
+      // first we construct a map holding the latest value for a key
+      for (VPackSlice it : VPackArrayIterator(initialSlice)) {
+        if (!it.isObject()) {
+          registerInvalidArgumentWarning(expressionContext, funcName);
+          return AqlValue(AqlValueHintNull());
+        }
+        for (auto const& [key, value] : VPackObjectIterator(it)) {
```
jsteemann, line 957: Could even be optimized to 
```
for (auto const& [key, value] : VPackObjectIterator(it, /*useSequentialIteration*/ true)) {
```
so it iterates the attributes in sequential order, and not in index table order. Iteration in sequential order is probably better for pretching data and will lead to less jumping around "randomly" in memory.

PR #19759, opened 2023-09-13
Added cppcheck to base images for circleci

CryptoNinjaGeek:
### Scope & Purpose

Install cppcheck in our base circleci images

### DIFF
--- a/scripts/docker/build/arangodb-build.Dockerfile
+++ b/scripts/docker/build/arangodb-build.Dockerfile
@@ -14,7 +14,7 @@ RUN [ "/tools/install-openldap.sh", "3.1.2" ]
 RUN apk --no-cache del perl
 
 COPY install-cppcheck.sh /tools/
-RUN [ "/tools/install-cppcheck.sh", "2.11" , "2.11.0" ]
+RUN if [ "$(apk --print-arch)" = "x86_64" ] ; then /tools/install-cppcheck.sh 2.11 2.11.0 ; fi
 
 RUN ln /usr/bin/sccache /usr/local/bin/gcc && \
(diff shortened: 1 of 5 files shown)


scripts/docker/build/arangodb-build.Dockerfile
```diff
@@ -13,6 +13,9 @@ COPY install-openldap.sh /tools/
 RUN [ "/tools/install-openldap.sh", "3.1.2" ]
 RUN apk --no-cache del perl
 
+COPY install-cppcheck.sh /tools/
+RUN [ "/tools/install-cppcheck.sh", "2.11" , "2.11.0" ]
```
KVS85, line 17:
```
COPY install-cppcheck.sh /tools/
RUN if [ "$(apk --print-arch)" = "x86_64" ] ; then /tools/install-cppcheck.sh 2.11 2.11.0 ; fi
```

KVS85 (changes requested):
We probably don't need cppcheck in ARM build image.

scripts/docker/build/arangodb-build.Dockerfile
CryptoNinjaGeek: @KVS85 Ok, I removed it from arm images
The reason why I didn't do a check before, was to have a consistent set of tools on both architectures

KVS85 (approved):

KVS85:
https://hub.docker.com/layers/arangodb/build-alpine/3.16-gcc11.2-openssl3.1.2-e5e06cdaa21/images/sha256-51c78b6223111bbe7919d634df77439e1daf8a1fbca7ff95509a0d25f2f3428f

PR #19699, opened 2023-08-31
[CINFRA-818] Index creation should go through the replicated log

apetenchea:
### Scope & Purpose

This PR introduces the _CreateIndex_ operation in replication2 databases. It allows for index creation to be controlled by the replicated log leader and passed to the followers subsequently.

**Remarks**
- For unique indexes, the leader first builds the index itself and only then replicates it. Other indexes don't have this limitation.
- The maintenance has been patched such that it only affects replication2 leaders. Followers are ignored, they only become aware of new indexes through the replicated log.

**Todos**
- Creating indexes during snapshot transfer - this is a bigger topic, will be handled in different PR.
- I added basic integration tests. In the near future I hope to add some extra unit tests.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*

PR #19760, opened 2023-09-13
[CINFRA-819] DropIndex operation

apetenchea:
### Scope & Purpose

DropIndex implementation for replication2. Followup for https://github.com/arangodb/arangodb/pull/19699

Remarks
- When dropping an index, it is not necessary to treat unique indexes separately. Dropping an unique index on the leader makes the leader's policy looser than the follower's. Any documents that are inserted during the drop might still be rejected by the leader. Any document that exists in the log, after the drop-index operation, will not be rejected by the follower due to unique-constraint violation.
- Some methods have been refactored in order to reduce code duplication.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**


arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -473,6 +473,18 @@ auto DocumentFollowerState::handleSnapshotTransfer(
       });
 }
 
+template<class T>
+auto DocumentFollowerState::GuardedData::applyAndRelease(T const& op,
```
mchacki, line 477: Like it :+1:

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -650,14 +650,65 @@ auto DocumentLeaderState::createIndex(
           }
 
           if (auto releaseRes = self->release(logIndex); releaseRes.fail()) {
-            LOG_CTX("a03c7", ERR, self->loggerContext)
+            LOG_CTX("26355", ERR, self->loggerContext)
```
mchacki, line 653: Why this LogID change?
Not that it is critical as the log cannot be written yet in production

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
```diff
@@ -145,19 +145,31 @@ auto DocumentStateShardHandler::getShardMap() -> ShardMap {
...
+    -> Result {
+  auto res = basics::catchToResult([this, shard = std::move(shard),
+                                    index =
+                                        std::move(index)]() mutable -> Result {
+    return _maintenance->executeDropIndex(std::move(shard), std::move(index));
   });
 
-  if (res.fail()) {
-    res = Result{res.errorNumber(),
-                 fmt::format("Error: {} Failed to ensure index on shard {}: {}",
```
mchacki, line 155: Don't we loose this additional log information now?

I think this is important as the general Index error message is generic, and this adds the context on the shard.

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
```diff
@@ -105,12 +106,13 @@ auto MaintenanceActionExecutor::executeModifyCollectionAction(
 auto MaintenanceActionExecutor::executeCreateIndex(
     ShardID shard, std::shared_ptr<VPackBuilder> const& properties,
     std::shared_ptr<methods::Indexes::ProgressTracker> progress) -> Result {
-  DatabaseGuard guard(_vocbase);
```
mchacki, line 108: You are loosing the Guard here.
This guard makes sure the vocbase pointer is not dropped.

Is there anything else that guarantees this guard?

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
```diff
@@ -122,6 +124,25 @@ auto MaintenanceActionExecutor::executeCreateIndex(
   return res;
 }
 
+auto MaintenanceActionExecutor::executeDropIndex(ShardID shard,
+                                                 velocypack::SharedSlice index)
+    -> Result {
+  auto col = _vocbase.lookupCollection(shard);
```
mchacki, line 130: If my comment on the DataBaseGuard is correct then we need the guard here as well.

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.cpp
mchacki: Come to think about it, should the guard actually be a member of the class that is only erased after the action is completed?

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.h
```diff
@@ -70,12 +72,17 @@ class MaintenanceActionExecutor : public IMaintenanceActionExecutor {
...
+      -> Result override;
   void addDirty() override;
 
  private:
   GlobalLogIdentifier _gid;
   MaintenanceFeature& _maintenanceFeature;
   ServerID _server;
+
+  // The vocbase reference remains valid for the lifetime of the executor.
+  // Replicated logs are stopped before the vocbase is marked as dropped.
```
mchacki, line 85: Ah so this means someone outside holds the DatabaseGuard?
if that is true ignore my above comments on the DatabaseGuard.

tests/js/server/replication2/replication2-replicated-state-document-store.js
```diff
@@ -1001,6 +1000,39 @@ const replicatedStateDocumentShardsSuite = function () {
           }
         }
       }
+
+      // Drop the index.
+      assertTrue(collection.dropIndex(index));
```
mchacki, line 1005: Good test :+1:

mchacki (changes requested):
I mostly have questions, if they get good answers happy to move this to approved.

The general Target of this PR is good, just want to double check that all assumptions are okay.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
apetenchea: I've had a conflict with another LOG_CTX and I just patched one of them, I forgot which one was first. Are you ok with this change, or should I revert?

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.h
apetenchea: There is no need for a `DatabaseGuard`. I have added it "out of reflex" before, but we made sure that, when the database is being dropped or the system is shutting down, the vocbase reference only becomes invalid after the replicated logs have been shut down.
This happens in `DatabaseFeature::dropDatabase()`. There, `shutdownReplicatedLogs()` is called before the vocbase is marked as dropped.

apetenchea: I am not against using one though, just for the safety of it, but I would rather do it for all the helpers then, in a way which does not interfere with unit tests. I can do this in a later refactoring.

arangod/Cluster/DropIndex.cpp
```diff
@@ -111,3 +119,21 @@ bool DropIndex::first() {
...
+  if (auto state = vocbase->getReplicatedStateById(col.replicatedStateId());
+      state.ok()) {
+    auto leaderState = std::dynamic_pointer_cast<
+        replication2::replicated_state::document::DocumentLeaderState>(
+        state.get()->getLeader());
+    if (leaderState != nullptr) {
+      return leaderState->dropIndex(col, std::move(index)).get();
+    } else {
+      // TODO prevent busy loop and wait for log to become ready (CINFRA-831)
+      std::this_thread::sleep_for(std::chrono::milliseconds{50});
```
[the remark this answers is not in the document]
apetenchea: Yes, and your suggestion would work as well. There are other maintenance actions with the same problem, and I would like to fix them in a single PR ([CINFRA-831](https://arangodb.atlassian.net/browse/CINFRA-831). Ideally I want a method like `waitForLogReady(timeout)`.

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
apetenchea: Addressed in https://github.com/arangodb/arangodb/pull/19760/commits/2493f257db51548e5dc338b416c957773ca59478

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
mchacki: I am fine with the change :+1:

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
mchacki: :+1:

arangod/Replication2/StateMachines/Document/MaintenanceActionExecutor.h
mchacki: Okay, I agree we are protected against dropping, so no Guard required.

mchacki (approved):
LGTM.

All my questions are addressed.

PR #19770, opened 2023-09-14
improve performance of LimitExecutor

jsteemann:
### Scope & Purpose

Improve performance of LimitExecutor. For simple queries such as
```
FOR doc IN <collection>
  LIMIT 0, @limit
  RETURN doc
```
the overhead of LIMIT is quite significant.
This is because the LimitExecutor works individually on all the rows it passes through.

Instead, we can simply fast-forward the input range. This reduces the overhead of LIMIT significantly in my own tests for higher LIMIT values.

For example, here are some performance tests results for query
```
FOR doc IN <collection> LIMIT 0, $limit RETURN doc
```
for a collection with 11.2m documents on single server, for varying $limit values. Each query was run 5 times and the average query runtime was printed.
Queries were run on single server, and result documents were not returned but immediately discarded.

```
before (without changes from this PR):
--------------------------------------
LIMIT 1000     => average: 0.0005 s
LIMIT 10000    => average: 0.0048 s
LIMIT 100000   => average: 0.0333 s
LIMIT 1000000  => average: 0.3406 s
LIMIT 10000000 => average: 3.3854 s
LIMIT 20000000 => average: 3.7982 s
LIMIT 984332   => average: 0.3335 s
LIMIT 39983    => average: 0.0138 s
LIMIT 1234567  => average: 0.4174 s

after (with changes from this PR):
-----------------------------------
LIMIT 1000     => average: 0.0005 s
LIMIT 10000    => average: 0.0028 s
LIMIT 100000   => average: 0.0270 s
LIMIT 1000000  => average: 0.2728 s
LIMIT 10000000 => average: 2.7450 s
LIMIT 20000000 => average: 3.0863 s
LIMIT 984332   => average: 0.2742 s
LIMIT 39983    => average: 0.0115 s
LIMIT 1234567  => average: 0.3440 s
```

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


MBkkt:
@jsteemann 

Could you post your results for such queries?
I just want to understand how big overhead from limit node
1.
```
FOR doc IN <collection>
  RETURN doc
```
2.
```
FOR doc IN <collection>
  LIMIT 1, 1000000000
  RETURN doc
```
3.
```
FOR doc IN <collection>
  LIMIT 1, 1000000000
  RETURN doc
```

goedderz:
Note that the rows weren't actually copied before; still, this PR removes some (constant) per-line overhead.

markuspf (approved):

MBkkt:
@goedderz 
But why is overhead so big? Can we do something in aql to reduce it?

goedderz:
> @goedderz But why is overhead so big? Can we do something in aql to reduce it?

@MBkkt I didn't have the time to profile it this time. It might be generic overhead of the AQL state machine in ExecutionBlockImpl. It might be the leftover per-row overhead in `countAndSkipAllRemainingDataRows`; it's there to find the next shadow row (this could be improved by making use of the `_shadowRows` data structure in AqlItemBlock to find the next shadow row with less steps).

But let's also compare this with other runtimes. `FOR i IN 1..1000000 LIMIT 1000000 RETURN i` has this profile (without this patch):

```
Execution plan:
 Id   NodeType            Calls     Items   Filtered   Runtime [s]   Comment
  1   SingletonNode           1         1          0       0.00002   * ROOT
  2   CalculationNode         1         1          0       0.00003     - LET #1 = 1 .. 1000000   /* range */   /* simple expression */
  3   EnumerateListNode    1000   1000000          0       0.15436     - FOR i IN #1   /* list iteration */
  4   LimitNode            1000   1000000          0       0.02014       - LIMIT 0, 1000000
  5   ReturnNode           1000   1000000          0       0.02006       - RETURN i
```

While `0.02s` isn't completely negligible in comparison with the `0.15s` of the list iteration, it pretty much shows it'll probably be negligible in all real-world scenarios, for two reasons: 1) The cost per row of any reasonable query will be significantly higher than that of a list iteration, and 2) limit is seldomly used with such large line numbers.

So while further optimizing it might be possible and interesting, I don't think its currently worth the effort.

PR #19778, opened 2023-09-18
[BTS-1610] Fix shadowrow forwarding

goedderz:
### Scope & Purpose

Non-relevant shadow rows (which mark the end of an iteration of an _outer_ subquery) must make an execution block return the output it has accumulated so far. When an aql item block starts with a non-relevant shadow row (i.e. it ends the iteration of an _outer_ subquery), this previously did not happen correctly, and nodes would collect more output into the same block. This would then lead pass-through nodes to return early (when encountering the non-relevant shadow row), thereby dropping some rows. In maintainer mode, this would lead to an assertion.

What was missing is that in such a scenario, a node must pop calls from all call lists with a depth lower than the encountered shadow row; i.e., from subqueries more innermost than the shadow row. Without that, a specific call on a subquery level without a default would be repeated, when it mustn't.

- [X] :hankey: Bugfix

### Checklist

- [X] Tests
  - [X] **Regression tests**
  - [ ] C++ **Unit tests**
- [ ] Backports
  - [ ] Backport for 3.11: **TODO**
  - [ ] Backport for 3.10: **TODO**

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/browse/BTS-1610


arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -2088,8 +2108,14 @@ ExecutionBlockImpl<Executor>::executeWithoutTrace(
...
-    // We can never return less rows then what we got!
-    TRI_ASSERT(_outputItemRow == nullptr || _outputItemRow->numRowsLeft() == 0);
+    // We can never return less rows than what we got!
+    TRI_ASSERT(_outputItemRow == nullptr || _outputItemRow->numRowsLeft() == 0)
+        << printBlockInfo() << " Passthrough block didn't process all rows. "
+        << (_outputItemRow == nullptr
+                ? fmt::format("output == nullptr")
+                : fmt::format("rows left = {}, rows written = {}",
+                              _outputItemRow->numRowsLeft(),
+                              _outputItemRow->numRowsWritten()));
```
mchacki, line 2118: Very good :+1:

arangod/Pregel/Algos/ShortestPath/ShortestPath.cpp
```diff
@@ -22,7 +22,6 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "Pregel/Algos/ShortestPath/ShortestPath.h"
-#include "Aql/ExecutionBlockImpl.tpp"
```
mchacki, line 25: 🥳

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -1237,6 +1252,8 @@ auto ExecutionBlockImpl<Executor>::shadowRowForwarding(AqlCallStack& stack)
       return ExecState::NEXTSUBQUERY;
     }
 
+    [[maybe_unused]] bool const hasDoneNothing =
```
mchacki, line 1255: Why is this maybe_unused and the others not?
It is only used in one of the If cases but so are the ones above?

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -1286,6 +1303,10 @@ auto ExecutionBlockImpl<Executor>::shadowRowForwarding(AqlCallStack& stack)
       // we need to forward them
       return ExecState::SHADOWROWS;
     } else {
+      if (hasDoneNothing and not shadowRow.isRelevant()) {
```
mchacki, line 1306: Why is this handling not Needed in the other NEXTSUBQUERY case? (5 lines above, cannot comment there...)

Happy to discuss this offline, I have gut feeling that the above place is equally in danger.

mchacki (commented):
In general Looks good to me.

I am not 100% sure if all cases are now covered, see comments. Then this can go to approved state.

arangod/Aql/ExecutionBlockImpl.tpp
goedderz: an accident.
```suggestion
    bool const hasDoneNothing =
```

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -1214,6 +1225,10 @@ auto ExecutionBlockImpl<Executor>::shadowRowForwardingSubqueryEnd(
     // Need to return!
     return ExecState::DONE;
   } else {
+    if (hasDoneNothing and not shadowRow.isRelevant()) {
```
MBkkt, line 1228: Why do we need to different style of logical condition in single file :(?

MBkkt (approved):

arangod/Aql/AqlCallStack.cpp
```diff
@@ -69,6 +69,16 @@ auto AqlCallStack::popCall() -> AqlCallList {
   return call;
 }
 
+void AqlCallStack::popDepthsLowerThan(size_t depth) {
+  TRI_ASSERT(!_operations.empty());
+  for (auto i = _operations.size() - depth; i < _operations.size(); ++i) {
```
jsteemann, line 74: Can `depth` be larger here than `_operations.size()`? If yes, then `i` could underflow.
If not, can we at least add an assertion that checks for this?

arangod/Aql/AqlCallStack.cpp
```diff
@@ -69,6 +69,16 @@ auto AqlCallStack::popCall() -> AqlCallList {
...
+    auto& operation = _operations[i];
+    if (operation.hasMoreCalls()) {
+      std::ignore = operation.popNextCall();
+    }
```
jsteemann, line 78: Can an operation in theory have more than one call? If so, do we need to repeat the `if (hasMoreCalls()) pop()` logic here until there are no more calls left?

ubsan_arangodb_suppressions.txt
```diff
@@ -15,6 +15,7 @@ null:3rdParty/rocksdb/file/random_access_file_reader.cc
 null:3rdParty/rocksdb/file/writable_file_writer.cc
 null:3rdParty/rocksdb/monitoring/perf_step_timer.h
 null:3rdParty/rocksdb/util/user_comparator_wrapper.h
+alignment:3rdParty/rocksdb/util/crc32c.cc
```
jsteemann, line 18: Is this change intentional? It seems to have nothing to do with the actual bugfix.

arangod/Aql/AqlCallStack.cpp
mchacki: Good point!

It cannot be larger, but this API does not protect us.

jsteemann (approved):
LGTM

PR #19780, opened 2023-09-18
fix UB in NetworkFeature

jsteemann:
### Scope & Purpose

Fix UB in NetworkFeature
in case a specific failure point was set for the NetworkFeature, this could lead to two threads racing for modifying the same Promise. this could cause random failures, the most prominent one seems to have been the exception "Promise already satisfied.".
This does not affect any released version.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Network/NetworkFeature.cpp
+++ b/arangod/Network/NetworkFeature.cpp
@@ -285,7 +285,7 @@ void NetworkFeature::prepare() {
 
 void NetworkFeature::start() {
   Scheduler* scheduler = SchedulerFeature::SCHEDULER;
-  if (scheduler != nullptr) {  // is nullptr in catch tests
+  if (scheduler != nullptr) {  // is nullptr in unit tests
     auto off = std::chrono::seconds(1);
     ::queueGarbageCollection(_workItemMutex, _workItem, _gcfunc, off);
   }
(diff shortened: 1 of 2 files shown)


arangod/Network/NetworkFeature.cpp
```diff
@@ -528,6 +528,10 @@ void NetworkFeature::retryRequest(
     }
   };
 
+  // we need the mutex during `queueDelayed` because the lambda might be
+  // executed faster than we can add the work item to the _retryRequests map.
+  std::unique_lock guard(_workItemMutex);
+
   // this will automatically cancel the request when we leave this
   // method, unless we are canceling this scopeGuard explicitly.
   // note that the mutex lock will be unlocked before the scopeGuard
```
neunhoef, line 537: This comment is no longer valid, we should change it.

neunhoef (changes requested):
This is good and I think it will fix the problem. See one comment, where I note that a source code comment is no longer valid.
While we are at it, I suggest to add a source code comment in the header for `_workItemMutex` saying that it also protects the `_retryRequests` map. This is not obvious in the header file and confused me.

neunhoef (approved):
LGTM

PR #19781, opened 2023-09-18
parallel execution for AQL

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1397

Parallel execution of ExecutionBlocks inside the same query, based on prior art by @mpoeter.
When a dependency returns `HASMORE`, the engine can schedule an async fetch request for more data onto the scheduler. So that the caller can process the just-returned results and the dependency can produce more results, concurrently.

This is currently supported for the following executor types:
- `ENUMERATE_COLLECTION`
- `ENUMERATE_LIST`
- `ENUMERATE_VIEW`
- `INDEX`
- `FILTER`
- `CALCULATION`
- `COLLECT`
- `SORT`
- `JOIN`

Parallel execution is turned off for the following executors deliberately:
- `SINGLETON`: as there will be no dependencies
- `NORESULTS`: as a NoResults node will not return any data and thus is always quick
- `RETURN`: in most cases Return will be a passthough action and won't take a long time
- `LIMIT`: when enabling async prefetching for Limit, we currently run into assertion failures, thus turned off for now. This restriction may be lifted in the future.
- `INSERT`/`UPDATE`/`REPLACE`/`REMOVE`/`UPSERT`: the original version turned itself off deliberately for modification nodes. We are currently keeping this behavior, but the restriction may be lifted in the future.
- `TRAVERSAL`, `SHORTEST_PATH`, `ENUMERATE_PATHS`: currently we turn off the async prefetching for any graph nodes, because the previous version also did so. The restriction may be lifted in the future.

Additionally, async prefetching is currently turned off inside subqueries and for any direct dependency of a `REMOTE` node. Additionally, all query snippets that include a `REMOTE` node will currently not use async prefetching. These restrictions may be lifted in the future.

The explain output was adjusted so that an additional column "Par" shows whether a node can run concurrently with other nodes.
The profiling output was adjusted so that the number of parallel calls is shown in the additional "Par" column.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1397
- [ ] Design document:


maierlars (approved):
LGTM JoinNode

MBkkt (approved):

arangod/Aql/OptimizerRulesFeature.cpp
```diff
@@ -819,13 +819,10 @@ in case the indexes qualify for it.)");
   // should only be activated for experimental usage at one's own risk.
```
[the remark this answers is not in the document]
jsteemann: Fixed

arangod/Aql/DocumentProducingNode.cpp
```diff
@@ -172,6 +172,13 @@ bool DocumentProducingNode::doCount() const noexcept {
   return _count && !hasFilter();
 }
 
+AsyncPrefetchEligibility DocumentProducingNode::canUseAsyncPrefetching()
+    const noexcept {
+  return (!hasFilter() || (_filter->isDeterministic() && !_filter->willUseV8()))
+             ? AsyncPrefetchEligibility::kEnableForNode
+             : AsyncPrefetchEligibility::kDisableForNode;
```
mpoeter, line 179: Why do we have these constraints on the filter?

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -465,6 +467,33 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
...
+  });
+  if (_profileLevel >= ProfileLevel::Blocks) {
+    // only if profiling is turned on, get current time
+    start = currentSteadyClockValue();
+  } else {
+    // if profiling is turned off, don't get current time, but instead
+    // cancel the scopeGuard so we don't unnecessarily call timing
+    // functions on exit
+    profilingGuard.cancel();
+  }
```
mpoeter, line 495: Shouldn't this be done in the `traceExecuteBegin`/`traceExecuteEnd` functions?

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -513,30 +533,28 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
...
+    setFailure({ex.code(),
+                absl::StrCat(ex.what(), " [node #", getPlanNode()->id().id(),
+                             ": ", getPlanNode()->getTypeString(), "]")});
     LOG_QUERY("7289a", DEBUG)
         << printBlockInfo()
         << " local statemachine failed with exception: " << ex.what();
-    throw;
+    if (_prefetchTask && !_prefetchTask->isConsumed()) {
+      _prefetchTask->waitFor();
+    }
```
mpoeter, line 544: Why do we need to wait for the prefetchTask here?

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -513,30 +533,28 @@ ExecutionBlockImpl<Executor>::execute(AqlCallStack const& stack) {
...
-    _firstFailure.reset(TRI_ERROR_INTERNAL, std::move(msg));
+    setFailure({TRI_ERROR_INTERNAL,
+                absl::StrCat(ex.what(), " [node #", getPlanNode()->id().id(),
+                             ": ", getPlanNode()->getTypeString(), "]")});
     LOG_QUERY("2bbd5", DEBUG)
         << printBlockInfo()
         << " local statemachine failed with exception: " << ex.what();
+    if (_prefetchTask && !_prefetchTask->isConsumed()) {
+      _prefetchTask->waitFor();
+    }
```
mpoeter, line 557: Why do we need to wait for the prefetchTask here?

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -899,33 +931,42 @@ auto ExecutionBlockImpl<Executor>::executeFetcher(ExecutionContext& ctx,
...
     });
 
-    // TODO - we should also consider the limit here, but unfortunately the
-    // softLimit is currently also used for the batchSize
-    if (std::get<ExecutionState>(result) == ExecutionState::HASMORE &&
-        _exeNode->isAsyncPrefetchEnabled()) {
+    // note: SCHEDULER is a nullptr in unit tests
+    if (SchedulerFeature::SCHEDULER != nullptr &&
+        std::get<ExecutionState>(result) == ExecutionState::HASMORE &&
+        _exeNode->isAsyncPrefetchEnabled() && !ctx.clientCall.hasLimit()) {
```
mpoeter, line 957: Maybe add a comment why we cannot use async prefetching if we have a limit

arangod/Aql/ExecutionBlockImpl.tpp
```diff
@@ -899,33 +931,42 @@ auto ExecutionBlockImpl<Executor>::executeFetcher(ExecutionContext& ctx,
...
-      _prefetchTask->reset();
 
       // TODO - we should avoid flooding the queue with too many tasks as that
       // can significantly delay processing of user REST requests.
 
       // we can safely ignore the result here, because we will try to
       // claim the task ourselves anyway.
-
-      SchedulerFeature::SCHEDULER->queue(
+      bool queued = SchedulerFeature::SCHEDULER->tryBoundedQueue(
```
mpoeter, line 969: Comment is outdated since you now handle the result.

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2278,6 +2320,13 @@ CostEstimate CalculationNode::estimateCost() const {
   return estimate;
 }
 
+AsyncPrefetchEligibility CalculationNode::canUseAsyncPrefetching()
+    const noexcept {
+  return expression()->isDeterministic() && !expression()->willUseV8()
+             ? AsyncPrefetchEligibility::kEnableForNode
+             : AsyncPrefetchEligibility::kDisableForNode;
```
mpoeter, line 2327: Why do we need these constraints?

arangod/Aql/ExecutionNode.cpp
```diff
@@ -2755,6 +2808,12 @@ CostEstimate ReturnNode::estimateCost() const {
   return estimate;
 }
 
+AsyncPrefetchEligibility ReturnNode::canUseAsyncPrefetching() const noexcept {
+  // ReturnNodes could make use of async prefetching, but the gain
+  // is too little to justify it.
+  return AsyncPrefetchEligibility::kDisableForNode;
```
mpoeter, line 2814: Actually, for streaming cursors prefetching from the return node is probably exactly what we want.

arangod/Aql/IndexNode.cpp
```diff
@@ -525,6 +525,16 @@ CostEstimate IndexNode::estimateCost() const {
   return estimate;
 }
 
+AsyncPrefetchEligibility IndexNode::canUseAsyncPrefetching() const noexcept {
+  if (_condition != nullptr && _condition->root() != nullptr &&
+      (!_condition->root()->isDeterministic() ||
+       _condition->root()->willUseV8())) {
+    // non-deterministic or V8 index lookup condition
```
mpoeter, line 532: Why do we need these constraints?

arangod/Aql/JoinNode.cpp
```diff
@@ -348,10 +346,27 @@ CostEstimate JoinNode::estimateCost() const {
...
+        (!it.filter->isDeterministic() || it.filter->willUseV8())) {
+      // non-deterministic or V8 filter condition
+      return AsyncPrefetchEligibility::kDisableForNode;
+    }
+    if (it.condition != nullptr && it.condition->root() != nullptr &&
+        (!it.condition->root()->isDeterministic() ||
+         it.condition->root()->willUseV8())) {
+      // non-deterministic or V8 index lookup condition
+      return AsyncPrefetchEligibility::kDisableForNode;
+    }
```
mpoeter, line 361: Why do we need these constraints?

arangod/Aql/DocumentProducingNode.cpp
jsteemann: The constraint for V8 is because the `Query` object only has a single V8 context, which it can enter and exit.
If multiple threads try to enter/exit the V8 context of the query concurrently, this would not be thread-safe.
The constraint for determinism is there because executing non-deterministic calculations could cause different query results if they are prefetched.
For example, a streaming query such as 
```
FOR i IN 1..1001
  LET x = ASSERT(i <= 1000)
  RETURN x
```
With the prefetching enabled, the query may fail before the first batch with the 1000 valid results is returned.

arangod/Aql/ExecutionBlock.cpp
```diff
@@ -136,40 +133,44 @@ void ExecutionBlock::collectExecStats(ExecutionStats& stats) {
...
-      _execNodeStats.runtime -= currentSteadyClockValue();
-      TRI_ASSERT(_execNodeStats.runtime < 0.0);
-    }
+    // only if profiling is turned on, get current time
+    start = currentSteadyClockValue();
+  } else {
+    // if profiling is turned off, don't get current time, but instead
+    // cancel the scopeGuard so we don't unnecessarily call timing
+    // functions on exit
+    profilingGuard.cancel();
```
mpoeter, line 151: That won't work like that - we cannot use a scopeGuard here, but do the update in `traceExecuteEnd`.

mpoeter (changes requested):

mpoeter (approved):

PR #19787, opened 2023-09-19
[FEATURE] Remove http2 from fuerte.

maierlars:
### Scope & Purpose
Remove http2 from fuerte http client. It is unused, almost untested and known to be broken.

- [x] :hankey: Bugfix
- [x] :pizza: New feature


maierlars:
Can't be deleted because otherwise we have no way to test our http2 server side implementation. (Which we want to keep?)

PR #19789, opened 2023-09-19
Make our waitpid implementation interupteable by the monitoring thread

dothebart:
### Scope & Purpose

We have process monitoring & deadline handling. 
If we spawn & wait for subprocesses this would not work.
Add that for the linux case. 

- [x] :pizza: New feature


jsteemann (approved):
LGTM

PR #19792, opened 2023-09-20
FE-307: Move services screen to new UI

palashkaria:
### Scope & Purpose

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-307


js/apps/system/_admin/aardvark/APP/react/src/components/select/SelectBase.tsx
```diff
@@ -15,13 +16,23 @@ export type OptionType = {
 const Option = <IsMulti extends boolean>(
   props: OptionProps<OptionType, IsMulti>
 ) => {
+  const finalChildren =
+    typeof props.children === "boolean" ? `${props.children}` : props.children;
```
nikhil-varma, line 20: Is this condition required? We can use `String()` to do the same.

js/apps/system/_admin/aardvark/APP/react/src/components/select/SelectBase.tsx
```diff
@@ -15,13 +16,23 @@ export type OptionType = {
...
+      <components.Option {...props} children={finalChildren} />
     </div>
   );
 };
 
+const SingleValue = <IsMulti extends boolean>(
+  props: SingleValueProps<OptionType, IsMulti>
+) => {
+  const finalChildren =
+    typeof props.children === "boolean" ? `${props.children}` : props.children;
```
nikhil-varma, line 32: Same question.

js/apps/system/_admin/aardvark/APP/react/src/views/services/ServicesReactView.jsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { ServicesViewWrap } from "./ServicesViewWrap";
+
+const ServicesReactView = () => {
+  return <ServicesViewWrap />;
+};
```
nikhil-varma, line 6: Wrapper required?

js/apps/system/_admin/aardvark/APP/react/src/views/services/useFetchServices.ts
```diff
@@ -0,0 +1,27 @@
...
+        mount: service.mount
+      });
+      return {
+        ...service,
+        config: config.body,
+        deps: deps.body
+      };
+    });
+    const finalServices = await Promise.all(services);
+    return finalServices as ServiceDescription[];
```
nikhil-varma, line 24: Any particular need of assertion?

js/apps/system/_admin/aardvark/APP/react/src/components/select/SelectBase.tsx
palashkaria: but what if children is not something which should be converted to a string?

js/apps/system/_admin/aardvark/APP/react/src/views/services/ServicesReactView.jsx
palashkaria: yes, a JSX wrapper is required to avoid some TS issues; can be fixed, but that's the pattern we follow everywhere

js/apps/system/_admin/aardvark/APP/react/src/views/services/useFetchServices.ts
palashkaria: yes, without this the type is `any[]`

nikhil-varma: How about using `Promise.all<Type>(services)`.. Still the same issue?

nikhil-varma (approved):
Just minor queries.. All good otherwise

js/apps/system/_admin/aardvark/APP/react/src/views/services/useFetchServices.ts
palashkaria: yes, that will work too!

js/apps/system/_admin/aardvark/APP/react/src/views/services/ServicesReactView.jsx
palashkaria: Yes at some point, but currently it's `require`d in `App.js`, might need to deal with webpack too somewhere

PR #19796, opened 2023-09-20
fix error handling to have proper brackets.

dothebart:
### Scope & Purpose

Javascript must not use `not` in this way.

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #19820, opened 2023-09-26
Avoid ReadWriteLock overflow

mpoeter:
### Scope & Purpose

Extend the internal state of ReadWriteLocks from 32-bit to 64-bit to prevent potential overflows. Previously creating more than 64k streaming transactions at the same time could produce an overflow that would effectively result in undefined behavior.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**
  - [x] C++ **Unit tests**
- [ ] Backports


tests/Basics/ReadWriteLockTest.cpp
```diff
@@ -616,3 +616,18 @@ TEST(ReadWriteLockTest, testLockWriteAttempted) {
   ASSERT_FALSE(lock.isLockedRead());
   ASSERT_FALSE(lock.isLockedWrite());
 }
+
+TEST(ReadWriteLockTest, readerOverflow) {
+  // this is a regression test for the old version where we only used 16-bits
```
kvahed, line 621:
```suggestion
  // this is a regression test for the old version where we only used 16 bits
```

kvahed (approved):
minor remark

jsteemann (approved):
LGTM

MBkkt (approved):

PR #19840, opened 2023-09-27
[APM-697] Improve collect, aggregate, etc performance

MBkkt:
### Scope & Purpose

Mostly via avoiding copies big values (byte size > 15) from rocksdb
Part of APM-697 task

TODO list for next PRs:
1. Replace malloc/free via new/delete for vpack buffer
2. Remove special case for int48
3. Fix projection behavior
4. ...

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1385
- [ ] Design document:


arangod/Graph/Cache/RefactoredTraverserCache.cpp
```diff
@@ -120,75 +120,66 @@ bool RefactoredTraverserCache::appendEdge(EdgeDocumentToken const& idToken,
...
+            transaction::BuilderLeaser builder(_trx);
+            {
+              VPackObjectBuilder guard(builder.get());
+              _edgeProjections.toVelocyPackFromDocument(*builder, edge, _trx);
+            }
+            result = aql::AqlValue(builder->slice());
+          } else {
+            result = aql::AqlValue(edge);
+          }
+          result = aql::AqlValue(edge);
```
jsteemann, line 170: I see that this is identical to the original code, but it seems that we are unconditionally clobbering `result` in line 170, even if it was set to something else in lines 166 or 168. This looks inefficient and wrong. Especially it looks like that the projections currently cause _extra_ work, and the full edge is copied out later anyway.

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -2021,24 +1876,30 @@ Result RocksDBCollection::lookupDocumentVPack(
...
     // write entry back to cache
     cache::Cache::SimpleInserter<DocumentCacheType>{
         static_cast<DocumentCacheType&>(*cache), key->string().data(),
         static_cast<uint32_t>(key->string().size()), ps.data(),
         static_cast<uint64_t>(ps.size())};
   }
 
+  auto data = buffer.release();
+  cb(token, std::move(data));
+  buffer.acquire(std::move(data));
```
jsteemann, line 1901: `data` is moved twice here. While this may work (if `cb` does not steal the value from `data`), it looks strange to the reader of the code.

arangod/RocksDBEngine/RocksDBIncrementalSync.cpp
```diff
@@ -342,9 +345,9 @@ Result syncChunkRocksDB(DatabaseInitialSyncer& syncer,
           auto [documentId, revisionId] = lookupResult;
 
           tempBuilder->clear();
-          r = physical->lookupDocument(*trx, documentId, *tempBuilder,
-                                       /*readCache*/ true, /*fillCache*/ false,
-                                       ReadOwnWrites::yes);
+          auto r =
```
jsteemann, line 348: This assigns the lookup result to a new in-scope variable `r`, whereas the old code assigned the lookup result to a variable `r` from an outer scope.
Is this change intentional?

arangod/RocksDBEngine/RocksDBIncrementalSync.cpp
```diff
@@ -889,9 +887,10 @@ Result handleSyncKeysRocksDB(DatabaseInitialSyncer& syncer,
               auto [documentId, revisionId] = lookupResult;
 
               tempBuilder.clear();
-              r = physical->lookupDocument(
-                  *trx, documentId, tempBuilder, /*readCache*/ true,
-                  /*fillCache*/ false, ReadOwnWrites::yes);
+              r = physical->lookup(trx.get(), documentId, callback,
+                                   {.readCache = true,
```
jsteemann, line 891: nit: `.readCache = true` seems to be the default and doesn't need to be mentioned explicitly here.
All previous callers of `lookup(...)` at least did not specify it explicity, so this place here is a little inconsistent with the previous occurrences.

arangod/Aql/AqlValue.cpp
```diff
@@ -81,302 +83,209 @@ static inline uint8_t intLength(int64_t value) noexcept {
...
-    case RANGE: {
-      break;
-    }
+    case VPACK_INLINE:
+      return VPackSlice{_data.inlineSliceMeta.slice}.isNumber();
+    case VPACK_SLICE_POINTER:
+      return VPackSlice{_data.slicePointerMeta.pointer}.isNumber();
+    case VPACK_MANAGED_SLICE:
+      return VPackSlice{_data.managedSliceMeta.pointer}.isNumber();
+    default:
```
jsteemann, line 188: Is it somehow possible that we ever attempt to store a number in a `VPACK_MANAGED_STRING` AqlValue? Then this case would need to be handled here too.

arangod/RocksDBEngine/RocksDBIncrementalSync.cpp
MBkkt: No, it's not, thanks!

arangod/RocksDBEngine/RocksDBCollection.cpp
MBkkt: Changed

arangod/Aql/AqlValue.cpp
MBkkt: No, it's not, you can see it in ctor

arangod/Graph/Cache/RefactoredTraverserCache.cpp
MBkkt: I just comment out code above, if you know how to fix it, feel free to suggest

arangod/Aql/AqlValue.cpp
```diff
@@ -1339,64 +1068,52 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
-AqlValue::AqlValue() noexcept {
-  // construct a slice of type None
-  // we will simply zero-initialize the two 64 bit words
-  _data.words[0] = 0;
-  _data.words[1] = 0;
-
-  // VPACK_INLINE must have a value of 0, and VPackSlice::None must be equal
-  // to a NUL byte too
-  static_assert(AqlValueType::VPACK_INLINE == 0,
-                "invalid value for VPACK_INLINE");
```
jsteemann, line 1356: Is this `static_assert` kept somewhere (probably in the header file)?
It is important that a default constructed AqlValue has type VPACK_INLINE and contains the velocypack None type. This is ensure by making all bits in the AqlValue struct zero. Would be good to keep the assertion somewhere.

arangod/Aql/AqlValue.cpp
```diff
@@ -1339,64 +1068,52 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
+  setType(t);
+  switch (t) {
     case VPACK_MANAGED_SLICE:
       _data.managedSliceMeta.lengthOrigin =
           other._data.managedSliceMeta.lengthOrigin;
-      _data.managedSliceMeta.managedPointer = static_cast<uint8_t*>(data);
+      _data.managedSliceMeta.pointer =
+          const_cast<uint8_t*>(static_cast<uint8_t const*>(data));
       break;
-    case VPACK_SLICE_POINTER:
```
jsteemann, line 1388: Is this case removed intentionally? Why?

arangod/Aql/AqlValue.cpp
```diff
@@ -1339,64 +1068,52 @@ int AqlValue::Compare(velocypack::Options const* options, AqlValue const& left,
...
+  if (size < sizeof(AqlValue)) {
+    initFromSlice(slice, size);
+  } else {
+    setType(AqlValueType::VPACK_MANAGED_STRING);
+    _data.managedStringMeta.pointer = data.release();
+  }
 }
 
 AqlValue::AqlValue(uint8_t const* pointer) noexcept {
+  // TODO(MBkkt) Is external possible here?
```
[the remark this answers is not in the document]
MBkkt: It wasn't in your PR, and it's affect only this ctor, so I added todo to check it later

arangod/Aql/AqlValue.cpp
MBkkt: Because this function should be called only for destructible AqlValue.
So it looks like "typo" in old code

arangod/Aql/AqlValue.cpp
MBkkt: It's kept in erase function, which called from dtor now

arangod/Aql/AqlValue.cpp
jsteemann: ok

arangod/Aql/AqlValue.cpp
```diff
@@ -1633,46 +1346,45 @@ Range const* AqlValue::range() const {
...
-    case VPACK_INLINE_INT48:
-    case VPACK_INLINE_INT64:
-    case VPACK_INLINE_UINT64:
-    case VPACK_INLINE_DOUBLE:
-    case VPACK_SLICE_POINTER:
-      return 0;
     case VPACK_MANAGED_SLICE:
       return _data.managedSliceMeta.getLength();
+    case VPACK_MANAGED_STRING:
+      return sizeof(std::string) + _data.managedStringMeta.getLength();
```
jsteemann, line 1368: Nit: this will slightly underreport the memory usage, as we are at least missing the one byte for the null terminator. Instead of adding the string's `size()` here, we could also use its `capacity()`, because the string allocation may be larger than what it stores.

MBkkt: Fixed

arangod/Aql/AqlValue.cpp
MBkkt: You can also see assert(false) in default case, so it also will be checked by tests.

arangod/Graph/Cache/RefactoredTraverserCache.cpp
jsteemann: We can leave it as you did it now in this PR, and fix it later when this PR is merged.

arangod/Aql/AqlValue.cpp
MBkkt: Same for `AqlValue::data()`

jsteemann (approved):
LGTM

PR #19842, opened 2023-09-27
further improve performance of `freeMemoryWhile`

jsteemann:
### Scope & Purpose

Further improve performance of `freeMemoryWhile`, by not invoking the callback function if nothing was reclaimed.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19843
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19846

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19846, opened 2023-09-27
further improve performance of `freeMemoryWhile`

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19842
Further improve performance of `freeMemoryWhile`, by not invoking the callback function if nothing was reclaimed.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19843

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cache/PlainCache.cpp
+++ b/arangod/Cache/PlainCache.cpp
@@ -269,10 +269,10 @@ bool PlainCache<Hasher>::freeMemoryWhile(
       bucket.evict(candidate);
       freeValue(candidate);
       maybeMigrate |= guard.source()->slotEmptied();
-    }
 
-    if (!cb(reclaimed)) {
-      break;
+      if (!cb(reclaimed)) {
+        break;
+      }
     }
   }
 
--- a/arangod/Cache/Cache.cpp
+++ b/arangod/Cache/Cache.cpp
@@ -404,16 +404,10 @@ bool Cache::freeMemory() {
   }
 
   auto cb = [this](std::uint64_t reclaimed) -> bool {
-    if (reclaimed > 0) {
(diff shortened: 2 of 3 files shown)


neunhoef (approved):
LGTM

arangod/Cache/PlainCache.cpp
```diff
@@ -269,10 +269,10 @@ bool PlainCache<Hasher>::freeMemoryWhile(
       bucket.evict(candidate);
       freeValue(candidate);
       maybeMigrate |= guard.source()->slotEmptied();
-    }
 
-    if (!cb(reclaimed)) {
-      break;
+      if (!cb(reclaimed)) {
```
mpoeter, line 273: What it `reclaimed == 0`? Is that possible?

mpoeter (approved):

arangod/Cache/PlainCache.cpp
jsteemann: It's true, in theory `reclaimed` is not checked for `0` here.
However, we only get in here if `candidate != nullptr` and then we calculate `reclaimed` as `candidate->size()`.
And `CachedValue::size()` is defined as
```
inline std::size_t size() const noexcept {
 return _headerAllocSize + keySize() + valueSize();
}
```
(with `_headerAllocSize` being > 0).
So in practice it should not happen that `reclaimed` is `0` when we get here.

PR #19863, opened 2023-09-28
Remove weird template parameter.

maierlars:
### Scope & Purpose
This is the first clean up step for the materialize executor. The executor had two template parameters, but only two of the possible four combinations were used, while only three of them were actually compiled.

In the end, what matters is if the documents are materialized using a local document id or a arango search document id.

More simplifications will follow.


jsteemann (approved):
LGTM

PR #17465, opened 2022-10-27
Fixed issue #17394

jsteemann:
### Scope & Purpose

Fixed issue #17394: Unnecessary document-lookup instead of Index-Only query. This change improves projection handling so that more projections can be served from indexes.
This also required a few adjustments of existing tests. From the test adjustments it can be seen that now we can serve projections from indexes in more cases than before.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [x] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [x] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.10: -
  - [ ] Backport for 3.9: -
  - [ ] Backport for 3.8: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: #17394 
- [ ] Design document:

PR #19864, opened 2023-09-28
[BTS-1618] Empty AQL Query Result

apetenchea:
### Scope & Purpose

_For a more in-depth explanation, see this [recording](https://drive.google.com/file/d/1xd9Dm0H_Nnra_8gPcAXPdLPhFQleZvDG/view?usp=sharing)_

**Problem**
Assume we have a collection where each document has the field "p", with subfields "s" and "k". For example:
```
{"p": {"s": 1234, "k": "hund"}}
{"p": {"s": 1235, "k": "katze"}}
```

There's an index created on ["p.s", "p.k"].

Given the following query:
```aql
FOR doc in @@coll
  FILTER doc.p.k == 'hund'
  SORT doc.p.s DESC
  RETURN doc.p.s
```
This should return `[1234]`, but on 3.10 it returns `[]`.

**Key observations**
Notice the sub-optimal usage of the index. If we were to create the index on ["p.k", "p.s"], not only the query would be faster, but the returned result would be correct. However, the bug is still there.  
It is also important to note that this is only reproducible in a cluster. On a single-server it works just fine.

**AQL internals**
I wrote this part to facilitate an easier review. Feel free to skip it.  
The coordinator receives the above query and builds the AST. This contains a lot of nodes, and is not optimized to use any indexes. Next step is to optimize the query.  
It gets reduced to the following nodes:
```
SingletonNode
IndexNode
CalculationNode
RemoteNode
GatherNode
ReturnNode
```

Lots of optimization rules are applied, but one two of them are relevant here:
```
move-filters-into-enumerate
reduce-extraction-to-projection
```
Now, because we're using an index, the optimizer figures that it may not need to fetch the document, but use the index projection instead. There are two fields used to indicate this: _projections_ and _filterProjections_. **Note that if projections are fully covered by the index, then the filterProjections are ignored**

**Messing up the projections**
1. On the _coordinator_, the `move-filters-into-enumerate` initializes both projection fields correctly: `projections = [["p","s"], ["p","k"]];  filterProjections = [["p","k"]]`. However, upon executing `Projections::removeSharedPrefixes()`, the projections are reduced to `projections=["p"]`. This is not longer the case in devel, as it has been changed in https://github.com/arangodb/arangodb/pull/17465.
2. Still on the _coordinator_, the `reduce-extraction-to-projection`  rule kicks in. Note that currently, the projections are not covered by the index. The `IndexNode::prepareProjections()` method is where an important decision is going to be made: if the index covers the _projections_, we keep them and wipe out _filterProjections_, or we calculate new _filterProjections_ and new _projections_. At this point on the coordinator, we take the latter, which yields `projections = [["p","s"]]; filterProjections = [["p","k"]]`.
3. On the _dbserver_, we get the projections as "p.s", which are covered by the index. Therefore, we take to former condition, wipe out _filterProjections_ and (wrongly) projection only "p.s". Eventually, the dbserver checks the filter in the `DocumentProducingCallbackVariant::WithProjectionsCoveredByIndex` callback and fails to find any document, because it doesn't have access to "p.k".

**Solution**
The problem is, as noted by a comment in `IndexNode::prepareProjections()` _idx->covers(...) modifies the projections object_. This creates a misunderstanding between the coordinator and the dbserver, where the former believes that projections are not covered by the index, while the latter thinks they are.  
Therefore, we are now sending the information over, exactly as the coordinator perceives it, through the `_indexCoversProjections` field. This way, we make sure the dbserver goes on the same path as the coordinator.  
Surprisingly, some of the work was already there. The "indexCoversProjections" vpack value existed before, together with `_indexCoversFilterCondition`, although neither was used anywhere in the code. This PR "finishes" that implementation.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

Other PRs created to prevent this:
- devel https://github.com/arangodb/arangodb/pull/19869
- 3.11 https://github.com/arangodb/arangodb/pull/19871

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1618
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 v3.10.11 (XXXX-XX-XX)
 ---------------------
 
+* Fixed BTS-1618: Buggy AQL query returning empty results, due to incorrect
+  use of index projections.
+
 * Improve performance of the in-memory cache's memory reclamation procedure.
   The previous implementation acquired too many locks, which could drive system
   CPU time up.
--- a/arangod/Aql/IndexNode.cpp
+++ b/arangod/Aql/IndexNode.cpp
@@ -107,6 +107,11 @@ IndexNode::IndexNode(ExecutionPlan* plan,
     _options.ascending = !base.get("reverse").getBool();
   }
 
+  if (auto indexCoversProjections = base.get("indexCoversProjections");
+      indexCoversProjections.isBool()) {
+    _indexCoversProjections = indexCoversProjections.getBool();
+  }
+
   VPackSlice indexes = base.get("indexes");
 
   if (!indexes.isArray()) {
@@ -581,7 +586,17 @@ void IndexNode::prepareProjections() {
     return;
   }
 
-  if (idx->covers(_projections)) {
+  auto coversProjections = std::invoke([&]() {
+    if (_indexCoversProjections.has_value()) {
+      // On a DBServer, already got this information from the Coordinator.
+      return *_indexCoversProjections;
+    } else {
+      // On the Coordinator, let's check.
+      return idx->covers(_projections);
+    }
+  });
+
+  if (coversProjections) {
     _projections.setCoveringContext(collection()->id(), idx);
   } else if (this->hasFilter()) {
     // if we have a covering index and a post-filter condition,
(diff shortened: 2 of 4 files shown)


jsteemann (approved):
LGTM

maierlars (approved):

PR #19867, opened 2023-09-28
Feature/use materialize multi node for inverted index

maierlars:
### Scope & Purpose
1. Rename `MaterializeMultiNode` to `MaterializeSearchNode` because it makes more sense. (it's probably called multi because it potentially accesses multiple collections)
2. Use `MaterializeSearchNode` for inverted indexes as well.


MBkkt (approved):
I think it's ok, but will be nice to also remove code that unnecessary now

maierlars:
Will happen in a follow up PR. I'm going to remove the `MaterializeSingleNode<false>` branch.

PR #19872, opened 2023-09-28
More MaterializeExecutor cleanup.

maierlars:
### Scope & Purpose
1. Remove unused template parameter from `MaterializeSingleNode`.
2. Rename `MaterializeSingleNode` to `MaterializeRocksDBNode` because that is what it does.
3. Remove template parameter from `MaterializeExecutorInfos` and simplify. Store aql collection pointer instead of collection name.

### DIFF
--- a/arangod/Aql/MaterializeExecutor.cpp
+++ b/arangod/Aql/MaterializeExecutor.cpp
@@ -73,7 +73,11 @@ MaterializeExecutor<localDocumentId>::MaterializeExecutor(
     MaterializeExecutor<localDocumentId>::Fetcher& /*fetcher*/, Infos& infos)
     : _buffer{infos.query().resourceMonitor()},
       _trx{infos.query().newTrxContext()},
-      _readCtx{infos} {}
+      _readCtx{infos} {
+  if constexpr (localDocumentId) {
+    _collection = infos.collection()->getCollection()->getPhysical();
+  }
+}
 
 template<bool localDocumentId>
 void MaterializeExecutor<localDocumentId>::Buffer::fill(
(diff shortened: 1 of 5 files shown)


arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -73,7 +73,11 @@ MaterializeExecutor<localDocumentId>::MaterializeExecutor(
     MaterializeExecutor<localDocumentId>::Fetcher& /*fetcher*/, Infos& infos)
     : _buffer{infos.query().resourceMonitor()},
       _trx{infos.query().newTrxContext()},
-      _readCtx{infos} {}
+      _readCtx{infos} {
+  if constexpr (localDocumentId) {
```
jsteemann, line 77: If this is `false`, `_collection` is not initialized. IMO it would be good to always initialize it to a nullptr, so that we can at least easily run into issues when dereferencing it unintentionally.

jsteemann (approved):
LGTM

arangod/Aql/MaterializeExecutor.cpp
MBkkt: _collection doesn't exist if it's false

MBkkt:
```cpp
  IRS_NO_UNIQUE_ADDRESS
  irs::utils::Need<localDocumentId, PhysicalCollection*> _collection{};
```
it will be irs::utils::Empty in such case

MBkkt: And anyway it's always default initialized

PR #19873, opened 2023-09-28
better reporting for hotbackup errors

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19866

* Abandon expired local hot backup locks on coordinators a bit later than before. Previously, the timeout was `timeout + 5` seconds on each coordinator. There can be issues when multiple coordinators need to acquire the lock, so that the first coordinator acquires the lock successfully and its unlock clock starts ticking, whereas the subsequent coordinators may need to lock acquire the lock. In this case, the lock on the first coordinators could have been released already, even though the backup hasn't even started. This PR increases the per-coordinator timeout to `timeout + 30` seconds, which improves the situation, but is still not 100% safe.

* Improve logging in case hot backup locks cannot be taken on coordinators.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/19874

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 v3.11.4 (XXXX-XX-XX)
 --------------------
 
+* Slightly extended hot backup release lock timeout on coordinators from 
+  `timeout + 5` seconds to `timeout + 30` seconds, to prevent premature release
+  of hot backup commit locks on coordinators.
+
 * Improve logging in case hot backup locks cannot be taken on coordinators.
 
 * Improve performance of the in-memory cache's memory reclamation procedure.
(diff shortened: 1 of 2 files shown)


neunhoef (approved):
LGTM

maierlars (approved):

PR #19881, opened 2023-09-29
add cluster-internal connectivity checks

jsteemann:
### Scope & Purpose

* Add cluster-internal connectivity checks.
  This makes Coordinators and DB-Servers in a cluster periodically send check
  requests to each other, in order to check if all servers can connect to
  each other.
  If a cluster-internal connection to another Coordinator or DB-Server cannot 
  be established within 10 seconds, a warning will be logged. 

  The new startup option `--cluster.connectivity-check-interval` can be used
  to control the frequency of the connectivity check, in seconds.
  If set to a value greater than zero, the initial connectivity check is
  performed approximately 15 seconds after the instance start, and subsequent 
  connectivity checks are executed with the specified frequency.
  If set to a value of zero, connectivity checks are disabled.

  This change also adds the metrics
  `arangodb_network_connectivity_failures_coordinators` and
  `arangodb_network_connectivity_failures_dbservers`, which can be monitored
  for detecting temporary or permanent connectivity issues.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19882
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cluster/ClusterFeature.cpp
```diff
@@ -1132,3 +1199,112 @@ std::unordered_set<std::string> ClusterFeature::allDatabases() const {
...
+  static constexpr double timeout = 10.0;
+  network::RequestOptions reqOpts;
+  reqOpts.skipScheduler = true;
+  reqOpts.timeout = network::Timeout(timeout);
+
+  std::vector<futures::Future<network::Response>> futures;
+  futures.reserve(servers.size());
+
+  for (auto const& server : servers) {
+    futures.emplace_back(network::sendRequest(pool, "server:" + server,
```
neunhoef, line 1271: I do not know if `network::sendRequest` can run into problems if it gets a stale connection from the connection pool. This could potentially produce false positives, just because connections in the connection pool are closed without the pool noticing...

neunhoef (commented):
I think the feature is good, see one comment about dangers w.r.t. the connection pool.
One more idea for the future: In principle, one could publish a metric with a label and use the recipient's name as label, such that we could even count how often a connection to a particular server was not possible. But this is likely more than what our metrics implementation can easily provide.

arangod/Cluster/ClusterFeature.cpp
jsteemann: It probably can. But I would hope that the connection pool will someone figure out if a connection is stable or broken, and then automatically close it at least after a while.
Otherwise all sorts of other requests could fail as well.

nerpaula (approved):
Metrics & startup option LGTM!

PR #19883, opened 2023-09-29
MaterializeRocksDBExecutor is passthrough

maierlars:
### Scope & Purpose
Introduced two separate classes for the materialize executors. Furthermore, the rocksdb case has become passthrough. This should improve the performance. 

In a followup PR the rocksdb case will also be upgraded to MultiGet.


arangod/Aql/MaterializeExecutor.h
```diff
@@ -112,43 +141,13 @@ class MaterializeExecutor {
...
-   * upstream
-   */
-  [[nodiscard]] std::tuple<ExecutorState, Stats, size_t, AqlCall> skipRowsRange(
-      AqlItemBlockInputRange& inputRange, AqlCall& call);
-
-  void initializeCursor() noexcept {
-    // Does nothing but prevents the whole executor to be re-created.
-    _collection = {};
-  }
-
```
MBkkt, line 128: Why it was removed? It looks important for materialize via multiget

arangod/Aql/MaterializeExecutor.h
```diff
@@ -86,8 +86,39 @@ class MaterializerExecutorInfos {
...
+  MaterializeRocksDBExecutor(Fetcher&, Infos& infos);
+
+  std::tuple<ExecutorState, MaterializeStats, AqlCall> produceRows(
+      AqlItemBlockInputRange& inputRange, OutputAqlItemRow& output);
+
+ private:
+  PhysicalCollection* _collection{};
+};
+
+class MaterializeSearchExecutor : public MaterializeExecutorBase {
```
MBkkt, line 121: Why isn't it passthrough, although rocksdb passthrough?
Could you explain for me, please

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -204,21 +204,11 @@ MaterializeExecutor<localDocumentId>::produceRows(
...
-  auto [it, end] = [&] {
-    if constexpr (localDocumentId) {
-      return std::pair{nullptr, nullptr};
-    } else {
-      return std::pair{_buffer.docs.begin(), _buffer.docs.end()};
-    }
-  }();
-  auto& callback = _readCtx.callback;
-  auto docRegId = _readCtx.infos->inputNonMaterializedDocRegId();
+  auto [it, end] = std::pair{_buffer.docs.begin(), _buffer.docs.end()};
```
MBkkt, line 207:
```suggestion
  auto it = _buffer.docs.begin();
  auto end = _buffer.docs.end();
```

MBkkt (changes requested):
To avoid merge without answers

arangod/Aql/MaterializeExecutor.h
maierlars: Well, you probably now better than me. Afaik it can happen that search has indexed a document that no longer exists. This is not true for rocksdb indexes (or if it is, it is a bug). Thus for the rocksdb version, we produce exactly one row for each input row. This is passthrough. In the search case, the executor acts as a filter for documents not found. Thus it will never be passthrough.

MBkkt: > Afaik it can happen that search has indexed a document that no longer exists. 

We use non transactional but corresponding to inverted index rocksdb snapshot, so it shouldn't be possible

MBkkt: Can you do it passthrough too? Or I can try it in the separate PR

MBkkt (approved):

arangod/Aql/MaterializeExecutor.h
maierlars: How can it be important for the search case, when it's not used in the code.

The `initializeCursor` is called for each subquery start to reset the executor. If it is not implemented, the executor is simply rebuilt in place.

arangod/Aql/MaterializeExecutor.h
MBkkt: > Thus for the rocksdb version, we produce exactly one row for each input row.

Because same should be for search.

arangod/Aql/MaterializeExecutor.cpp
```diff
@@ -37,58 +37,62 @@
...
+                 [&, &input = input](LocalDocumentId /*id*/,
+                                     std::unique_ptr<std::string>& doc) {
+                   TRI_ASSERT(input.isInitialized());
+                   output.moveValueInto(docOutReg, input, &doc);
+                   return true;
+                 }},
+                {})
+            .ok();
+
+    TRI_ASSERT(written);
```
MBkkt, line 83: I think this can be false for rocksdb sometime 
@jsteemann What do you think?

arangod/Aql/MaterializeExecutor.h
MBkkt: Yes, but we don't need to rebuild this executor, it's rebuilding free already allocated memory and resources, and then allocate them another time

arangod/Aql/MaterializeExecutor.h
maierlars: > Can you do it passthrough too? Or I can try it in the separate PR
sure, please go ahead.

arangod/Aql/MaterializeExecutor.cpp
maierlars: If it fails, then something is seriously wrong. If it's a general IO read error, then the query should be aborted, and not silently drop that row from the result set.

arangod/Aql/MaterializeExecutor.h
maierlars: > We use non transactional but corresponding to inverted index rocksdb snapshot, so it shouldn't be possible

Just to get that right, you use a different snapshot for looking up documents found via the inverted index, than the rest of the aql query?

MBkkt: > Just to get that right, you use a different snapshot for looking up documents found via the inverted index, than the rest of the aql query?

Yes, even worse, can be different snapshot for different iresearch link/inverted index.

So in this materialize search executor or in view executor can be used multiple rocksdb snapshot for materialization.

arangod/Aql/MaterializeExecutor.cpp
MBkkt: Sounds good to me.

arangod/Aql/ExecutionNode.h
```diff
@@ -1237,6 +1237,9 @@ class MaterializeRocksDBNode : public MaterializeNode,
   ExecutionNode* clone(ExecutionPlan* plan, bool withDependencies,
                        bool withProperties) const override final;
 
+  bool alwaysCopiesRows() const override { return false; }
+  bool isIncreaseDepth() const override { return false; }
```
MBkkt, line 1241: Do I understand right it should be false because block is passthrough and we call nextDataRow at the same cycle with advanceRow?
And if it's not true, like for MultiGet, it can be true?

MBkkt: Hmm, I also not fully understand purpose of `isIncreaseDepth`.

It looks like best case is
`alwaysCopiesRows` -- true // because it's mean we can reuse registers
`isIncreaseDepth` -- false // we don't increase depth

PR #19887, opened 2023-09-29
Automate build docker container build

markuspf:
### Scope & Purpose

Automate building of the build docker containers with CircleCI. 

`config.yml` detects whether files in the directory `scripts/docker` have been changed, and

if so 
  * chooses a continuation (`docker_build.yml`) that will build the docker images for x86 and arm and push them to a Elastic Container Registry
  * creates a manifest that contains both the x86 and arm64 images; this allows just using one image name for jobs in circleci and only changing the resource class to switch between arm and x86
  * writes the generated tag into the file `BUILD_DOCKER_TAG` in the root of the git repository
  * commits this change and pushes to the current branch

if not
  * generates a continuation as before, but one that uses the `BUILD_DOCKER_TAG` to build ArangoDB and run the tests.
  * In particular, if a developer changes the Dockerfile, the first step will update the docker image and one can then build and test ArangoDB with the new docker image on the respective PR. When that PR gets merged, all future PRs that branch off `devel` will use the new image. 
  
## Why ECR and not docker.io?

ECR was chosen, because CircleCI hosts its infrastructure on AWS and therefore we expect fewer problems with test jobs failing because of failing downloads of docker images. The registry is public, and these docker images can be used by developers locally without login, too

## What is the registry and can I see the image?

The registry is  `public.ecr.aws/b0b8h2r4`, the image name is of the shape `build-alpine:$ALPINE_VERSION-clang$CLANG_VERSION-gcc$GCC_VERSION-openssl$OPENSSL_VERSION-$(git rev-parse --short)` where git is executed on the commit that changes the dockerfile.

## How long will the docker images be kept?

This is to be determined. Docker Images incur storage cost and hence have to be cleaned up regularly. Docker images that are still in use for `devel` builds and (active) PRs/branches should not be deleted (too early).


scripts/docker/build/build-image_arm64.sh
```diff
@@ -11,7 +11,7 @@ OPENSSL_VERSION=$(docker run --rm -it $image:latest-$arch sh -c "cat \${OPENSSL_
 
 OS_VERSION=$(docker run --rm -it $image:latest-$arch cat /etc/alpine-release | grep -Po "\d+\.\d+")
 
-IMAGE_TAG=${OS_VERSION}-gcc${GCC_VERSION}-openssl${OPENSSL_VERSION}-$(git rev-parse --short HEAD)-$arch
+IMAGE_TAG=${OS_VERSION}-gcc${GCC_VERSION}-openssl${OPENSSL_VERSION}-${CIRCLE_SHA1::8}-$arch
```
KVS85, line 14: Can you please set `${CIRCLE_SHA1::8}` to `$(git rev-parse --short HEAD)` in case it's empty (a local run is done)?

scripts/docker/build/build-image_x86-64.sh
```diff
@@ -11,7 +11,7 @@ OPENSSL_VERSION=$(docker run --rm -it $image:latest-$arch sh -c "cat \${OPENSSL_
 
 OS_VERSION=$(docker run --rm -it $image:latest-$arch cat /etc/alpine-release | grep -Po "\d+\.\d+")
 
-IMAGE_TAG=${OS_VERSION}-gcc${GCC_VERSION}-openssl${OPENSSL_VERSION}-$(git rev-parse --short HEAD)-$arch
+IMAGE_TAG=${OS_VERSION}-gcc${GCC_VERSION}-openssl${OPENSSL_VERSION}-${CIRCLE_SHA1::8}-$arch
```
KVS85, line 14: Can you please set `${CIRCLE_SHA1::8}` to `$(git rev-parse --short HEAD)` in case it's empty (a local run is done)?

.circleci/base_config.yml
```diff
@@ -368,6 +379,58 @@ jobs:
...
+
+  build-build-docker-manifest:
+    docker:
+      - image: cimg
+    steps:
+      - get-docker-build-scripts
+      - run:
+          name: Build manifest
+          command: |
+            IMAGE_TAG=${CIRCLE_SHA1::8} ./build-manifest.sh
```
KVS85, line 428: IMAGE_TAG for `build-manifest.sh ` is expected to be `${OS_VERSION}-gcc${GCC_VERSION}-openssl${OPENSSL_VERSION}-${CIRCLE_SHA1::8}`.

KVS85 (commented):
There are 2 submodules among changed files.

PR #19895, opened 2023-10-02
Projections for JoinNode

maierlars:
### Scope & Purpose
This PR adds support for projections to the JoinNode/JoinExecutor. Now it can handle queries like
```
for doc1 in c1
     sort doc1.c
     for doc2 in c2
          filter doc1.x == doc2.y
          return [doc1.x, doc2.y]
```
without loading the documents. It should have support for stored values already, but this I haven't tested yet.

### DIFF
--- a/arangod/Aql/JoinExecutor.cpp
+++ b/arangod/Aql/JoinExecutor.cpp
@@ -37,9 +37,23 @@ JoinExecutor::JoinExecutor(Fetcher& fetcher, Infos& infos)
   constructStrategy();
 }
 
+namespace {
+struct SpanCoveringData : arangodb::IndexIteratorCoveringData {
+  explicit SpanCoveringData(std::span<VPackSlice> data) : _data(data) {}
+  bool isArray() const noexcept override { return true; }
+  VPackSlice at(size_t i) const override { return _data[i]; }
+  arangodb::velocypack::ValueLength length() const override {
+    return _data.size();
+  }
+
+  std::span<VPackSlice> _data;
+};
+}  // namespace
+
 auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
                                OutputAqlItemRow& output)
     -> std::tuple<ExecutorState, Stats, AqlCall> {
+  VPackBuilder projectionsBuilder;
   bool hasMore = false;
   while (inputRange.hasDataRow() && !output.isFull()) {
     if (!_currentRow) {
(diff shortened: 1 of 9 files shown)


arangod/Aql/JoinExecutor.cpp
```diff
@@ -37,38 +37,73 @@ JoinExecutor::JoinExecutor(Fetcher& fetcher, Infos& infos)
...
+  }
+
+  std::span<VPackSlice> _data;
+};
+}  // namespace
+
 auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
                                OutputAqlItemRow& output)
     -> std::tuple<ExecutorState, Stats, AqlCall> {
+  VPackBuilder projectionsBuilder;
```
jsteemann, line 56: Creating VPackBuilders repeatedly for every invocation of `produceRows()` seems expensive, even if the method is only called every 1000 rows or so.
VPackBuilders are quite heavy underneath, so I suggest that we create a VPackBuilder instance variable instead, which we can recycle via `VPackBuilder::clear()` upon every iteration.

arangod/RocksDBEngine/RocksDBVPackIndex.cpp
```diff
@@ -2935,7 +2937,24 @@ struct RocksDBVPackStreamIterator : AqlIndexStreamIterator {
   }
 
   LocalDocumentId load(std::span<VPackSlice> projections) const override {
```
jsteemann, line 2939: The `span` received by this function is a by-value parameter, and this function then modifies it.
I don't see the modified span being returned by this function, so I am wondering why it is modified in the first place.
I would understand it if the span were passed by reference and not by value, but that doesn't seem to be the case here.

maierlars: A span is just a memory region. We do not modify the span itself, but the memory it points to. It's like a string_view for vector.

arangod/Aql/JoinNode.cpp
```diff
@@ -147,6 +147,8 @@ void JoinNode::doToVelocyPack(VPackBuilder& builder, unsigned flags) const {
     // condition
     builder.add(VPackValue("condition"));
     it.condition->toVelocyPack(builder, flags);
+    // projections
+    it.projections.toVelocyPack(builder);
```
jsteemann, line 151: This is fine for serialization, but AFAICS the generated "projections" velocypack attribute is never read back.
The JoinNode constructor that is supposed to create a JoinNode from a velocypack Slice should read the "projections" attribute from each IndexInfo entry somehow (e.g. here: https://github.com/arangodb/arangodb/pull/19895/files#diff-b284b202b1037fe300a1226b5a94c5b844d63bfbbfcaa3d42af091075346da6bR123-R127).
Otherwise the packaged projections info will be lost.

maierlars: Will be done in a follow up PR, when I prepare the JoinNode for execution in the cluster.

arangod/Aql/JoinExecutor.cpp
```diff
@@ -51,24 +65,46 @@ auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
...
-                                    _currentRow, &doc);
-               return true;
-             }},
-            {});
+        auto& idx = _infos.indexes[k];
+        if (idx.projections.empty()) {
+          // load the whole document
+
+          // TODO post filter based on document value
+          // TODO implement document projections
```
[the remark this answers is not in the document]
maierlars: I will do this in a follow up PR.

jsteemann (approved):
LGTM.
Both issues I mentioned will be addressed in a follow-up PR.

PR #19903, opened 2023-10-04
Add document projections for join nodes.

maierlars:
### Scope & Purpose
The projections PR only added projections for values from the index. But arangodb assumes that we also support projections from documents.

### DIFF
--- a/arangod/Aql/JoinExecutor.cpp
+++ b/arangod/Aql/JoinExecutor.cpp
@@ -68,27 +68,8 @@ auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
       std::size_t projectionsOffset = 0;
       for (std::size_t k = 0; k < docIds.size(); k++) {
         auto& idx = _infos.indexes[k];
-        if (idx.projections.empty()) {
-          // load the whole document
 
-          // TODO post filter based on document value
-          // TODO implement document projections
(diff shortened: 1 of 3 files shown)


arangod/Aql/JoinExecutor.cpp
```diff
@@ -102,9 +83,55 @@ auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
...
+                   _projectionsBuilder.openObject(true);
+                   idx.projections.toVelocyPackFromDocument(
+                       _projectionsBuilder,
+                       VPackSlice(
+                           reinterpret_cast<uint8_t const*>(doc->data())),
+                       &_trx);
+                   _projectionsBuilder.close();
+                   output.moveValueInto(
+                       _infos.indexes[k].documentOutputRegister, _currentRow,
+                       _projectionsBuilder.slice());
```
MBkkt, line 129: Maybe better to doesn't make projection for such case and keep whole document?
Btw velocypack::Buffer can be moved to aqlvalue, so you can use bufferRef for this

maierlars: You are certainly correct for single server use cases, but as soon as we send that document over the network, you want to have the projected value instead.

jsteemann (approved):
LGTM

PR #19907, opened 2023-10-04
allow compiling executables with & without V8/JavaScript support

jsteemann:
### Scope & Purpose

This introduces a CMake option `USE_V8`, which can be used to compile
our executables with and without V8/JavaScript support.
The option defaults to ON, so that V8/JavaScript are used as usual.
By setting the option to OFF, executables will be compiled without
V8/JavaScript support.

Note that V8 will always be built, regardless of the value of this option.
This is required because the V8 build step also builds ICU, which is needed
in all programs.
If set to OFF, V8 will not be linked to arangod and the client-tools,
and V8 integration code for arangod will not be compiled.
Additionally, arangosh will not be built.
Setting the option to OFF will further disable all JavaScript support in
arangod, rendering the web UI, JavaScript transactions, AQL user-defined
functions (UDFs), Foxx and some other smaller features unusable.
Setting this option to OFF also renders many tests unusable as many tests
depend on JavaScript.
Setting the option to OFF is highly experimental option and not well tested.
Use at your own risk!

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/arangod.cpp
```diff
@@ -122,7 +126,14 @@ static int runServer(int argc, char** argv, ArangoGlobalContext& context) {
         },
         [](auto& server, TypeTag<ShutdownFeature>) {
           return std::make_unique<ShutdownFeature>(
-              server, std::array{ArangodServer::id<ScriptFeature>()});
+              server,
+#ifdef USE_V8
+              std::array { ArangodServer::id<ScriptFeature>() }
+#else
+              // just a random feature specified here, so the array isn't empty.
```
MBkkt, line 133: Is it affect order? Maybe shouldn't be random feature?

arangod/VocBase/Methods/Databases.cpp
```diff
@@ -558,6 +563,9 @@ Result Databases::drop(ExecContext const& exec, TRI_vocbase_t* systemVocbase,
       return Result(TRI_ERROR_INTERNAL, dropError);
     }
   } else {
+#else
+  {
+#endif
```
MBkkt, line 568:
```suggestion
  } else 
#endif
  {
```

maierlars (approved):

arangod/VocBase/Methods/Indexes.cpp
```diff
@@ -752,10 +756,16 @@ Result Indexes::drop(LogicalCollection& collection,
 
     transaction::Options trxOpts;
     trxOpts.requiresReplication = false;
+#ifdef USE_V8
     SingleCollectionTransaction trx(transaction::V8Context::createWhenRequired(
```
MBkkt, line 760: Why is V8context used when V8 is available?

lib/CMakeLists.txt
```diff
@@ -245,7 +249,7 @@ target_link_libraries(arango
          velocypack
          linenoise-ng
          date_interface
-         ${V8_LIBS}
+         #         ${V8_LIBS}
```
MBkkt, line 252: Why not removed?

MBkkt (approved):

arangod/RestServer/arangod.cpp
jsteemann: Changed now so that the order should be preserved.

arangod/VocBase/Methods/Databases.cpp
jsteemann: Applied.

arangod/VocBase/Methods/Indexes.cpp
jsteemann: I don't know right now. In this PR I simply wanted to preserve the original code for the case we build with V8.

lib/CMakeLists.txt
jsteemann: Removed.

arangod/VocBase/Methods/Indexes.cpp
MBkkt: Ok

PR #19908, opened 2023-10-04
[CINFRA-852] Remove insert API from replicated logs

goedderz:
### Scope & Purpose

The insert API on replicated log must not be called publicly, as it will most certainly break the state machine (e.g. the document state). This PR removes this API, and reintroduces an insert API on the black hole state machine for our tests and performance tests.


apetenchea (approved):
LGTM

PR #19911, opened 2023-10-04
smaller AQL cleanups

jsteemann:
### Scope & Purpose

More miscellaneous small AQL simplifications out of another, larger PR.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/RestAqlHandler.cpp
```diff
@@ -781,11 +792,6 @@ RestStatus RestAqlHandler::handleUseQuery(std::string const& operation,
 
   answerBuilder.close();
 
-  VPackOptions const* opts = &VPackOptions::Defaults;
-  if (_engine) {  // might be destroyed on shutdown
```
MBkkt, line 785: Is it guaranteed that opts still valid here?

arangod/Aql/RestAqlHandler.cpp
```diff
@@ -709,6 +710,11 @@ RestStatus RestAqlHandler::handleUseQuery(std::string const& operation,
 
   auto const rootNodeType = _engine->root()->getPlanNode()->getType();
 
+  VPackOptions const* opts = &VPackOptions::Defaults;
+  if (_engine) {  // might be destroyed on shutdown
```
MBkkt, line 714: Do we need this check? Below it's used without this condition.
Or it's not needed only for `operation == StaticStrings::AqlRemoteExecute`?

arangod/Aql/RestAqlHandler.cpp
```diff
@@ -752,12 +758,17 @@ RestStatus RestAqlHandler::handleUseQuery(std::string const& operation,
...
     answerBuilder.add(VPackValue(StaticStrings::AqlRemoteResult));
-    result.toVelocyPack(answerBuilder, &_engine->getQuery().vpackOptions());
+    result.toVelocyPack(answerBuilder, opts);
     answerBuilder.add(StaticStrings::Code, VPackValue(TRI_ERROR_NO_ERROR));
   } else if (operation == "initializeCursor") {
+    // TODO: remove the following 2 lines if the assertion holds true
     auto pos = VelocyPackHelper::getNumericValue<size_t>(querySlice, "pos", 0);
+    TRI_ASSERT(pos == 0);
     Result res;
+    // TODO: remove the following if condition and blcok if the assertion
```
MBkkt, line 768:
```suggestion
    // TODO: remove the following if condition and block if the assertion
```

MBkkt (approved):

arangod/Aql/RestAqlHandler.cpp
jsteemann: I did not use your suggestion, but instead resolved the TODO and removed the unreachable code.

arangod/Aql/RestAqlHandler.cpp
jsteemann: It is probably true that `_engine` is always != nullptr here, but I am not 100% sure. I think the check doesn't do any harm, so I would prefer to leave it in.

MBkkt: Ok

PR #19914, opened 2023-10-05
Dump Projections

maierlars:
### Scope & Purpose

Experimental branch to test the impact of having projections in the dump api for other users than arangodump. The projection format is not yet final and subject to change. In fact we try to figure out what the best way of describing it is.


maierlars:
Still requires tests.

neunhoef (approved):
LGTM

PR #19915, opened 2023-10-05
Enable Join Rule

maierlars:
### Scope & Purpose
The Join rule is currently disabled by default. To use the join node, it has to be explicitly enabled.

This PR changes this behavior.  As a consequence of this, some queries in the tests will be optimized differently.


jsteemann (approved):
LGTM

PR #19916, opened 2023-10-05
Join Node polish

maierlars:
### Scope & Purpose
Polish the join node. This PR includes:
- fixing the calculation of the estimates.
- adding more tests for unique indexes.
- fixing a bug when using a unique index for joins (thanks to the tests)
- added more details to the explainer


jsteemann (approved):
LGTM

PR #19918, opened 2023-10-05
Improve reading document from collection

MBkkt:
### Scope & Purpose

* Resolve todo (about not using ability to store pointer to string in aql value)
* Make interface (DocumentCallback) more user friendly

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/DocumentProducingHelper.cpp
+++ b/arangod/Aql/DocumentProducingHelper.cpp
@@ -72,7 +72,8 @@ template<bool checkUniqueness, bool skip>
 IndexIterator::DocumentCallback aql::getCallback(
     DocumentProducingCallbackVariant::WithProjectionsNotCoveredByIndex,
     DocumentProducingFunctionContext& context) {
-  auto cb = [&context](LocalDocumentId token, VPackSlice slice) {
+  return [&context](LocalDocumentId token, aql::DocumentData&& data,
+                    VPackSlice slice) {
     if constexpr (checkUniqueness) {
       if (!context.checkUniqueness(token)) {
         // Document already found, skip it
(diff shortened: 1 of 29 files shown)


arangod/Aql/DocumentProducingHelper.cpp
```diff
@@ -554,34 +532,38 @@ IndexIterator::CoveringCallback aql::getCallback(
...
+          } else {
+            output.moveValueInto(registerId, input, s);
+          }
+        } else {
+          objectBuilder.clear();
+          objectBuilder.openObject(true);
+
+          // projections from the index for the filter condition
+          context.getProjections().toVelocyPackFromDocument(
+              objectBuilder, s, context.getTrxPtr());
```
jsteemann, line 555: Why is there no special handling here if `v` is set, as there is in most other cases?

arangod/Aql/MultiGet.h
```diff
@@ -58,15 +58,14 @@ class MultiGetContext {
...
-                             std::make_unique<std::string>(std::string_view{
-                                 slice.startAs<char>(), slice.byteSize()});
-                         return true;
-                       }),
-                   {.readCache = false, .fillCache = false}, snapshot);
+      base->lookup(
+          &trx, id,
+          [&](LocalDocumentId, ValueType&&, VPackSlice doc) {
+            values[_global] = std::make_unique<std::string>(
+                std::string_view{doc.startAs<char>(), doc.byteSize()});
```
jsteemann, line 65: Is it possible that `ValueType` is set here? We could then even use the unique_ptr from ValueType without creating yet another string.

arangod/Cluster/TraverserEngine.cpp
```diff
@@ -189,20 +189,19 @@ void BaseEngine::getVertexData(VPackSlice vertex, VPackBuilder& builder,
...
-          }
-          return true;
-        });
+    auto cb = [&](LocalDocumentId, aql::DocumentData&&, VPackSlice doc) {
+      // FOUND. short circuit.
+      read++;
+      builder.add(v);
+      if (!options().getVertexProjections().empty()) {
+        VPackObjectBuilder guard(&builder);
+        options().getVertexProjections().toVelocyPackFromDocument(builder, doc,
```
jsteemann, line 198: what if `DocumentData` was set?

arangod/Aql/MultiGet.h
MBkkt: Yes, but it's code only for mock collection, so I don't see the profit for optimizing it

arangod/Aql/DocumentProducingHelper.cpp
MBkkt: Because it's always some projection? v set only for empty projection case

arangod/Aql/JoinExecutor.cpp
```diff
@@ -90,51 +90,31 @@ auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
...
+              } else {
+                output.moveValueInto(reg, _currentRow, doc);
+              }
+            } else {
+              _projectionsBuilder.clear();
+              _projectionsBuilder.openObject(true);
+              idx.projections.toVelocyPackFromDocument(_projectionsBuilder, doc,
+                                                       &_trx);
+              _projectionsBuilder.close();
+              output.moveValueInto(_infos.indexes[k].documentOutputRegister,
```
jsteemann, line 108:
```suggestion
              output.moveValueInto(reg,
```

arangod/Pregel/GraphStore/GraphLoader.cpp
```diff
@@ -205,8 +205,8 @@ auto GraphLoader<V, E>::loadVertices(LoadableVertexShard loadableVertexShard)
   }
 
   std::string documentId;  // temp buffer for _id of vertex
-  auto cb = IndexIterator::makeDocumentCallbackF([&](LocalDocumentId token,
-                                                     VPackSlice slice) {
+  auto cb = [&](LocalDocumentId token, aql::DocumentData&& data,
```
jsteemann, line 208: `data` is not used inside the function.
```suggestion
  auto cb = [&](LocalDocumentId token, aql::DocumentData&&,
```

arangod/Cluster/TraverserEngine.cpp
MBkkt: We don't use DocumentData in case if some projection exist, I don't think it's good, but it's current state.

On my opinion best approach what we can to achieve is having multiple aql registers:
one with whole document
multiple with projected fields (just reference to whole document)
Note: whole document won't be serialized and sent via network

But such approach needs additional work.

arangod/Aql/JoinExecutor.cpp
MBkkt: Fixed a little different way

jsteemann (approved):
LGTM

PR #19933, opened 2023-10-09
BTS-1622: so we don't reach the ci-SUT memory limits we reduce the test pattern a little bit

dothebart:
### Scope & Purpose

This test out-does the limits; reduce its size a bit.


jsteemann (approved):
LGTM if this makes the tests pass again.

PR #19936, opened 2023-10-09
Post Filter for JoinNode

maierlars:
### Scope & Purpose
This PR adds post filter support for Join Nodes using filter projections and document lookup.


arangod/Aql/OptimizerRules.cpp
```diff
@@ -9057,18 +9057,12 @@ void arangodb::aql::joinIndexNodesRule(Optimizer* opt,
   bool modified = false;
   if (nodes.size() >= 2) {
     // not yet supported:
-    // - projections
     // - post-filtering
```
jsteemann, line 9060:
```suggestion
```

jsteemann (approved):
LGTM

PR #19946, opened 2023-10-11
JoinNode check index supports stream

maierlars:
### Scope & Purpose
Ask the index if it supports the stream interface instead of assuming its alsways the vpack index.


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #19952, opened 2023-10-11
Remove int48

MBkkt:
### Scope & Purpose

The idea is reduce count of cases, also simplify initialization, etc 

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/AqlValue.cpp
```diff
@@ -624,21 +580,19 @@ double AqlValue::toDouble(bool& failed) const {
   failed = false;
   auto t = type();
   switch (t) {
-    case VPACK_INLINE_INT48:
-      return static_cast<double>(_data.shortNumberMeta.data.int48.val);
     case VPACK_INLINE_INT64:
       // TODO(MBkkt) Not all int64_t fit to double without precision lost
-      return static_cast<double>(
-          basics::littleToHost(_data.longNumberMeta.data.intLittleEndian.val));
+      return static_cast<double>(absl::little_endian::ToHost(
```
jsteemann, line 585: I think it makes sense to add an explicit `#include` directive for the absl header that we require in this file now.
And we can now remove the include header for the endianness functionality in `Basics/Endian.h`.

jsteemann (approved):
LGTM

arangod/Aql/AqlValue.cpp
MBkkt: #include <absl/strings/numbers.h> for endian

Basics/Endian removed.

tests/js/common/shell/shell-query-cache-noncluster.js
```diff
@@ -619,7 +619,7 @@ function AqlQueryCacheSuite () {
     
     testMaxResultsSize1 : function () {
       const query = "FOR i IN 1..@value RETURN i";
-      cache.properties({ mode: "demand", maxResults: 10, maxResultsSize: 8192 });
+      cache.properties({ mode: "demand", maxResults: 10, maxResultsSize: (1000 + 1001) * 9 + 100 * 2 });
```
Dronplane, line 622: 8192 ->18209 quite a huge size growth. As AqlValue sizeof was not changed, it is most likely caused by increased VPack size produced by `initFromUint` / `initFromInt`. Is this the only consequence?

MBkkt: Yes, it's changed because vpack format changed, and I think it's good

Compression can be used to replace memory overhead via cpu overhead

jsteemann: I understand that this PR removes the int48 type and we are now using int64/uint64 instead. Fine.
But the change should only be inside `AqlValue`. And both int48 and int64/uint64 can be stored inside an AqlValue and do not need extra memory. And all the numeric AqlValue types return `0` as their memory usage.
So I don't see why the reported memory usage should be higher than before now. What am I missing?

Dronplane: @jsteemann  initFromUint / initFromInt  with this PR produce Slices with fixed format. e.g. even if number requires 1 byte  - 8 would be used. While it is inside AqlValue it makes no problem. But when this slice is copied anywhere from AqlValue it starts to overuse memory.

jsteemann: Ah ok, makes sense.
The code in `AqlValue::toVelocyPack()` adds the inline `Slice` to the result `Builder` as is. It does not convert any numeric values into the shortest possible VelocyPack number type. We could easily add that optimization in `AqlValue::toVelocyPack(...)` to save memory.

jsteemann: Something like the following should do it:
```diff
diff --git a/arangod/Aql/AqlValue.cpp b/arangod/Aql/AqlValue.cpp
index 0e7cff1ff14..a71069d299b 100644
--- a/arangod/Aql/AqlValue.cpp
+++ b/arangod/Aql/AqlValue.cpp
@@ -872,10 +872,13 @@ void AqlValue::toVelocyPack(VPackOptions const* options, VPackBuilder& builder,
         break;
       }
       [[fallthrough]];
-    case VPACK_INLINE:
-    case VPACK_INLINE_INT48:
     case VPACK_INLINE_INT64:
+      builder.add(VPackValue(_data.longNumberMeta.data.intLittleEndian.val));
+      break;
     case VPACK_INLINE_UINT64:
+      builder.add(VPackValue(_data.longNumberMeta.data.uintLittleEndian.val));
+      break;
+    case VPACK_INLINE:
     case VPACK_INLINE_DOUBLE:
     case VPACK_MANAGED_SLICE:
     case VPACK_MANAGED_STRING:
```

jsteemann: note: above code is untested, but it should convey the idea.

MBkkt: Do you want it?

MBkkt: The code is missing conversion from little endian to host btw

jsteemann: Ok, thanks for picking up the suggestion! :+1:

jsteemann (approved):
LGTM!

Dronplane (approved):
LGTM

PR #19954, opened 2023-10-12
Optimize FilterProjections

maierlars:
### Scope & Purpose
This PR tries to reduce the amount of copied data for filter projections. If a post filter can be served by projections from the indexed values or stored values, a document is built in memory and passed to the filter expression. Instead one could reference the slices directly by modifying the expression.

This PR introduces references to these external values.


maierlars:
Closed for now. We have found a simpler, but not so performant solution. We should revisit this approach later.

PR #19956, opened 2023-10-12
Introduce variables for each projection.

maierlars:
### Scope & Purpose
This is a more simple version of the idea of optimizing post filters. Instead of build the document using projections, we introduce new variables that correspond to the projected values and set them to the values stored in the index.

Also fix a bug in an formerly unused function.


arangod/Aql/JoinExecutor.cpp
```diff
@@ -312,7 +339,12 @@ void JoinExecutor::constructStrategy() {
     options.usedKeyFields = {0};
 
     auto& desc = indexDescription.emplace_back();
-
+    if (idx.filter) {
+      VPackBuilder builder;
+      idx.filter->expression->toVelocyPack(builder, true);
+      LOG_DEVEL << builder.toJson();
```
jsteemann, line 345:
```suggestion
```

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #19957, opened 2023-10-12
Optimize Projections

maierlars:
### Scope & Purpose
There are cases where the projections of index and enumerate collection nodes can be optimized after the fact. This PR tries to do this, but also introduces new problems.

If the filter projections are no longer a subset of the actual projections, an index scan with post filter will misbehave because fields are missing now. This is because the IndexExecutor uses the projected value for filtering.

This will also be addressed in this PR.


arangod/Aql/DocumentProducingHelper.cpp
```diff
@@ -505,49 +517,55 @@ template<bool checkUniqueness, bool skip>
...
-                                                     context.getTrxPtr());
+        if (context.hasFilter() && !context.checkFilter(&covering)) {
+          context.incrFiltered();
+          return false;
+        }
 
-      objectBuilder.close();
+        // recycle our Builder object
+        VPackBuilder& objectBuilder = context.getBuilder();
+        if (!skip) {
```
jsteemann, line 546:
```suggestion
        if constexpr (!skip) {
```

arangod/Aql/DocumentProducingHelper.cpp
```diff
@@ -505,49 +517,55 @@ template<bool checkUniqueness, bool skip>
...
-    }
+          // projections from a covering index
+          context.getProjections().toVelocyPackFromIndex(
+              objectBuilder, covering, context.getTrxPtr());
 
-    return true;
-  };
+          objectBuilder.close();
+        }
+        if constexpr (!skip) {
```
jsteemann, line 556: It looks like as if this condition (`! skip`) can be merged with the if above, which has the same condition.

arangod/Aql/IndexExecutor.cpp
```diff
@@ -218,6 +219,7 @@ IndexExecutorInfos::IndexExecutorInfos(
       _projections(std::move(projections)),
       _filterProjections(std::move(filterProjections)),
       _filterVarsToRegs(std::move(filterVarsToRegs)),
+      _filterCoveringVars(filterCoveringVars),
```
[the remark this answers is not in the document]
jsteemann: The caller (https://github.com/arangodb/arangodb/pull/19957/files?diff=unified#diff-6f31e3326162227b694393c2323d91800579bb1403d122351c61d88b30b194c3R417) also passes this variable via std-move, so we could also accept the parameter by value here.

arangod/Aql/DocumentProducingNode.cpp
```diff
@@ -174,3 +172,38 @@ void DocumentProducingNode::setFilterProjections(aql::Projections projections) {
...
+      LOG_DEVEL << "found projections: " << _projections;
+      _projections = Projections(std::move(attributes));
+    } else {
+      LOG_DEVEL << "too many projections: " << attributes << " ("
+                << attributes.size() << ") max: " << maxProjections();
+    }
+  }
+}
+
+bool DocumentProducingNode::isProduceResult() const {
```
jsteemann, line 206: Should this be a pure virtual function?
Asking because `IndexNode::isProduceResult()` does not call the base class `DocumentProducingNode::isProduceResult()`. 
And the only other implementation is in `EnumerateCollectionNode::isProduceResult()`. So we could move the default implementation from `DocumentProducingNode` into `EnumerateCollectionNode`.

arangod/Aql/OptimizerRules.cpp
```diff
@@ -9275,3 +9275,18 @@ void arangodb::aql::joinIndexNodesRule(Optimizer* opt,
...
+
+void arangodb::aql::removeUnnecessaryProjections(
+    Optimizer* opt, std::unique_ptr<ExecutionPlan> plan,
+    OptimizerRule const& rule) {
+  containers::SmallVector<ExecutionNode*, 8> nodes;
+  plan->findNodesOfType(nodes, {EN::INDEX, EN::ENUMERATE_COLLECTION}, true);
+
+  bool modified = false;
+  for (auto* n : nodes) {
+    modified = true;
```
jsteemann, line 9287: This rule will now show up in every plan that uses an IndexNode or an EnumerateCollectionNode, even if the projections were not changed. I think this is not ideal.
Only those optimizer rules should show up that actually modify the plan. We can make recalculateProjections() return a boolean value, indicating whether projections were changed.
In order to easily figure out if projections were changed, we could make the nodes create a hash of their projections, which can hash whatever is in `_projections`, and optionally `_filterProjections`, before and after the call to `recalculateProjections()`.

arangod/Aql/DocumentProducingNode.cpp
```diff
@@ -32,6 +32,8 @@
 #include "Aql/Variable.h"
 #include "Basics/StaticStrings.h"
 #include "Basics/VelocyPackHelper.h"
+#include "OptimizerUtils.h"
+#include "Logger/LogMacros.h"
```
jsteemann, line 36:
```suggestion
```

arangod/Aql/DocumentProducingNode.cpp
```diff
@@ -32,6 +32,8 @@
 #include "Aql/Variable.h"
 #include "Basics/StaticStrings.h"
 #include "Basics/VelocyPackHelper.h"
+#include "OptimizerUtils.h"
```
jsteemann, line 35: To be consistent with the other includes, this could use `Aql/` and be moved up so the includes are sorted alphabetically.
```suggestion
#include "Aql/OptimizerUtils.h"
```

tests/js/server/aql/aql-filter-projections.js
```diff
@@ -73,7 +73,7 @@ function filterProjectionsPlansTestSuite () {
         assertEqual(1, nodes[0].indexes.length, query);
         assertEqual(query[1], nodes[0].indexes[0].type, query);
         assertEqual(normalize(query[2]), normalize(nodes[0].projections), query);
-        assertEqual(normalize([]), normalize(nodes[0].filterProjections), query);
+        assertEqual(normalize(query[3]), normalize(nodes[0].filterProjections), query);
```
jsteemann, line 76: Some of the queries above don't have their member[3] set. I guess this is ok because we are then comparing against `normalize(undefined)`, which should produce `[]`. So it's ok, but I was slightly confused in the beginning when looking at the change.

jsteemann (approved):
LGTM

tests/js/server/aql/aql-filter-projections.js
maierlars: I double checked and I'm sure all of them have a member[3].

maierlars:
Failed tests have been disabled.

PR #19958, opened 2023-10-12
Reduce number of shared_ptr copies in cache subsystem

jsteemann:
### Scope & Purpose

Reduce number of shared_ptr copies in cache subsystem

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [x] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #19963, opened 2023-10-13
Reduce number of shared_ptr copies in cache subsystem

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/19958

Reduce number of shared_ptr copies in cache subsystem

- `Cache::migrate()` migrates data for all buckets in a cache table. for each bucket, it calls `migrateBucket(...)`. `migrateBucket` created a copy of the table's atomic shared_ptr, which internally uses a lock. this is now changed so that the table pointer is passed into the call to `migrateBucket`, so that we do not need to copy the shared_pointer repeatedly. this saves n copies of the atomic shared_ptr, where n is the size of the table.
- `Cache::requestMigrate(...)` also created a copy of the table's atomic shared_ptr, which is now saved by passing the table pointer into the function.
- `Cache::reportInsert()` could be called after inserting data into a cache table. `reportInsert()` also created a copy of the table's atomic shared_ptr, which is now also saved.
- added an internal metric in the code for tracking the number of calls to `Cache::table()`. this metric is not enabled by default to remove the overhead of the tracking. the code can be enabled easily when debugging.
- replace a modulus calculation in `freeMemoryWhile(...)` with a bit operation, as cache table sizes are always power-of-2 values.
- cache tables have a maximum size of 4 * 1024 * 1024 entries. some internals in `Cache::Table` use a uint32_t when indexing cache table entries. added code that ensures that no cache tables with larger sizes can be created.
- do not update bucket term from `freeMemoryWhile()` for every bucket.
  the term update is very expensive, but not needed for memory reclamation. 
- added trace log messages for MigrateTask and FreeMemoryTask so that their internals can be inspected.

Also make cache metrics more reliable:
The previous implementation that returned the metrics for the in-memory cache subsystem could return all metrics set to 0 if it could not acquire the cache manager lock in time.
now, instead of returning zeroed results, we simply return the same result that we returned last time. so now in the worst case we return slightly stale metrics, but this is a lot better than returning zeroed metrics.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/19962

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 v3.10.11 (XXXX-XX-XX)
 ---------------------
 
+* Reduce number of atomic shared_ptr copies in in-memory cache subsystem.
+
 * Updated arangosync to v2.19.4.
 
 * Fix updating of collection properties (schema) when the collection has a view
--- a/arangod/Cache/Cache.cpp
+++ b/arangod/Cache/Cache.cpp
@@ -128,10 +128,14 @@ void Cache::sizeHint(uint64_t numElements) {
       static_cast<double>(numElements) /
       (static_cast<double>(_slotsPerBucket) * Table::idealUpperRatio));
   std::uint32_t requestedLogSize = 0;
-  for (; (static_cast<std::uint64_t>(1) << requestedLogSize) < numBuckets;
+  for (; (static_cast<std::uint64_t>(1) << requestedLogSize) < numBuckets &&
+         requestedLogSize < Table::kMaxLogSize;
        requestedLogSize++) {
   }
-  requestMigrate(requestedLogSize);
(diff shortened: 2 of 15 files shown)


MBkkt (approved):

Dronplane (approved):
LGTM

PR #19969, opened 2023-10-16
Add CircleCI windows build on local runner

mpoeter:
The build runs in a Windows docker container on our own Windows server.


scripts/docker/build/windows/arangodb-build-win.Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    
+RUN @powershell git config --global --add safe.directory *; `
+    $env:PATH += ';C:\Program Files\NASM'; `
+    "C:/tools/install-openssl.ps1" -branch "3.1" -revision ".2"; `
+    # clean out the temp folder to reduce image size
+    Remove-Item -Path '$env:TEMP/*' -Recurse -Force -ErrorAction SilentlyContinue; `
+    # perl and nasm are only needed to build OpenSSL
+    choco uninstall strawberryperl nasm -y
+
+ENV OPENSSL_ROOT_DIR=C:\openssl-3.1.2
```
dothebart, line 53: I think this should take into account the `VERSIONS` file - at least in a final version.

dothebart (approved):
Seems good to me. 
This is build only, right? Test running will be done later?

mpoeter:
Yes, ATM this is community build only.

Still need to figure at an issue with cloning the enterprise repo on the local runner, then we can build enterprise instead (moving forward we will only need the enterprise build for Windows).

After that I will add test runs.

PR #19974, opened 2023-10-16
3.11: [BTS-1610] Fix shadowrow forwarding 

mchacki:
### Scope & Purpose

Backport of: https://github.com/arangodb/arangodb/pull/19778
Straight forward Cherry-pick

Non-relevant shadow rows (which mark the end of an iteration of an _outer_ subquery) must make an execution block return the output it has accumulated so far. When an aql item block starts with a non-relevant shadow row (i.e. it ends the iteration of an _outer_ subquery), this previously did not happen correctly, and nodes would collect more output into the same block. This would then lead pass-through nodes to return early (when encountering the non-relevant shadow row), thereby dropping some rows. In maintainer mode, this would lead to an assertion.

What was missing is that in such a scenario, a node must pop calls from all call lists with a depth lower than the encountered shadow row; i.e., from subqueries more innermost than the shadow row. Without that, a specific call on a subquery level without a default would be repeated, when it mustn't.

- [X] :hankey: Bugfix

### Checklist

- [X] Tests
  - [X] **Regression tests**
  - [ ] C++ **Unit tests**
- [ ] Backports
  - [ ] Backport for 3.11: **TODO**
  - [ ] Backport for 3.10: **TODO**

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/browse/BTS-1610

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 v3.11.5 (XXXX-XX-XX)
 --------------------
 
+* Fixed BTS-1610: In certain situations with at least three levels of nested
+  subqueries, of which some of the outer iterations don't return any results,
+  results of later iterations could be lost.
+
 * Updated arangosync to v2.19.4.
 
 * Add cluster-internal connectivity checks.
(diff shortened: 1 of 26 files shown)


jsteemann (approved):
LGTM

mchacki:
Jenkins is green:
https://jenkins01.arangodb.biz/job/arangodb-matrix-pr/31796/

For some reason it is not linked

PR #19981, opened 2023-10-17
Bug fix/fix review comments2

dothebart:
### Scope & Purpose

replace several remote js executions by API functions, js-ify namings

- [x] :hankey: Bugfix


js/client/modules/@arangodb/aql-helper.js
```diff
@@ -458,21 +447,36 @@ function unpackRawExpression (node, transform = false) {
...
+};
+
+function execute_all_json(plans, result, query) {
+  plans.forEach(function (plan) {
+    let jsonResult = db._executeJson(plan, {optimizer: {rules: ['-all']}});
+    assertEqual(jsonResult, result, query);
+  });
+};
+
+exports.execute_all_json = execute_all_json;
```
MBkkt, line 478: Is it test only?

MBkkt (approved):

js/client/modules/@arangodb/aql-helper.js
dothebart: ah, you're right. case should be adjusted.

MBkkt: My point was different, do we want to provide this API as part of our js API?

PR #19988, opened 2023-10-18
More tests for JoinNode

maierlars:
### Scope & Purpose
This PR adds tests that test all combinations in which a JoinNode can be used given two indexes.
These tests also revealed a few bugs which are also fixed:
- if an index covers the projections, but does not cover the filter condition, the document has to be loaded anyways. Thus we might as well not use the index for projections.
- the RocksDBVPackStreamIterator messed up the projections by interleaving filter projections and document projections.
- fix for the unique join strategy.


tests/js/client/aql/aql-index-join-multi.js
```diff
@@ -0,0 +1,257 @@
...
+          config.projection2, config.projectedAttribute2);
+
+      const expectedResult = computeResult(config);
+      // sorting by x is required, because the query does not imply any sorting order
+      const actualResult = db._query(query).toArray().sort((left, right) => left[0].z - right[0].z);
+
+      // db._explain(query);
+      // print(expectedResult);
+      // print(actualResult);
+
```
hkernbach, line 178:
```suggestion
```

hkernbach (approved):
LGTM (Only minor: There are some "log only" leftovers - But I guess you want them to stay there for further development, so LGTM).

PR #19989, opened 2023-10-18
Improve locking in Async Logger

mchacki:
### Scope & Purpose

*Right now the Logger holds a very large lock. The lock is held whenever the background thread flushes a queued log message, which is potentially slowed down by IO. If we try to do a new direct log entry in that period we need to wait for this lock to be released. With this PR the Lock is changed such that: We get a WriteLock whenever we modify the Loggers we use, but we only take a read-lock when we let the loggers print. This way we can have two or more threads using the loggers.
In turn the output message of the Logger, which does the IO operation, is guarded with a WriteLock.

In addition this PR adds an upper bound to the amount of messages in the queue. This is right now hard-coded and should not have effect under normal circumstances. If the logging is configured to use an extremely slow device we protect us here against accumulating too much memory in log messages. If the Limit is reached, we will start to directly log the message, which will have a performance impact. Also this may change the ordering of Log messages.
*

Note to reviewer: Please take extra care, this PR fiddles with Locks, and has potential to introduce undesired dead-locks.

This potentially fixes: https://arangodb.atlassian.net/browse/BTS-1273

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1273
- [ ] Design document:


lib/Logger/LogThread.cpp
```diff
@@ -105,6 +113,7 @@ bool LogThread::processPendingMessages() {
   MessageEnvelope env{nullptr, nullptr};
 
   while (_messages.pop(env)) {
+    _pendingMessages.fetch_sub(1);
```
jsteemann, line 116:
```suggestion
    _pendingMessages.fetch_sub(1, std::memory_order_relaxed);
```

lib/Logger/LogThread.cpp
```diff
@@ -50,9 +56,11 @@ bool LogThread::log(LogGroup& group, std::unique_ptr<LogMessage>& message) {
       (message->_level == LogLevel::FATAL || message->_level == LogLevel::ERR ||
        message->_level == LogLevel::WARN);
 
-  if (!_messages.push({&group, message.get()})) {
+  if (_pendingMessages.load() >= MAX_QUEUED_MESSAGES ||
+      !_messages.push({&group, message.get()})) {
     return false;
   }
+  _pendingMessages.fetch_add(1);
```
jsteemann, line 63: Suggestion:
instead of performing an atomic load followed by an atomic fetch_add in 99.9999% of cases, we should unconditionally execute the fetch_add, then do the comparison, and roll back the fetch_add in case we are over the limit or something fails.
This will be more efficient.
```suggestion
  auto numMessages = _pendingMessages.fetch_add(1, std::memory_order_relaxed) + 1;
  if (numMessages >= MAX_QUEUED_MESSAGES ||
      !_messages.push({&group, message.get()})) {
      /* roll back the update */
    _pendingMessages.fetch_sub(1, std::memory_order_relaxed);
    return false;
  }
```

lib/Logger/LogThread.cpp
```diff
@@ -26,6 +26,12 @@
 #include "Logger/LogAppender.h"
 #include "Logger/Logger.h"
 
+// Hard-coded limit of delayed messages. If this limit is reached, we will
+// directly log all following messages. This is to prevent the queue from
+// growing indefinitely if the log thread is not able to keep up with the
+// incoming messages.
+static constexpr size_t MAX_QUEUED_MESSAGES = 10000;
```
jsteemann, line 33: It would be really nice if this limit were configurable. LoggerFeature has a lot of options already, so we could easily add this one too.

lib/Logger/LogAppender.cpp
```diff
@@ -233,6 +234,15 @@ void LogAppender::reopen() {
   LogAppenderFile::reopenAll();
 }
 
+void LogAppender::logMessageGuarded(LogMessage const& message) {
+  // Only one thread is allowed to actually write logs to the file.
+  // We use a recusive lock here, just in case writing the log message
+  // causes a crash, in this case we may trigger another force-direct
+  // log. This is not very likely, but it is better to be safe than sorry.
+  RECURSIVE_WRITE_LOCKER(_logOutputMutex, _logOutputMutexOwner);
```
jsteemann, line 242: Not sure if it is possible, but can we have two different appenders writing to the same sink, e.g. stdout?
In this case, the previous implementation serialized the writes to the sink, and with this PR they can execute concurrently, leading to potentially garbled output.
But we should first check if it is actually possible to have two different appenders write to the same output file, fd, sink, ...

mchacki: Good Point.
This is indeed possible.
Will apply our discussed idea

mchacki: Applied, we should now only have a single Appender per File

lib/Logger/LogAppenderFile.h
```diff
@@ -78,29 +80,40 @@ class LogAppenderFile : public LogAppenderStream {
...
   static std::mutex _openAppendersMutex;
-  static std::vector<LogAppenderFile*> _openAppenders;
+  static std::vector<std::shared_ptr<LogAppenderFile>> _openAppenders;
 
   static int _fileMode;
   static int _fileGroup;
+
+ private:
+  // Static class, never construct it.
+  LogAppenderFileFactory(){};
```
jsteemann, line 116: nit: the follow would signal the intent more clearly.
```suggestion
  LogAppenderFileFactory() = delete;
```

lib/Logger/LogThread.cpp
```diff
@@ -26,11 +26,18 @@
 #include "Logger/LogAppender.h"
 #include "Logger/Logger.h"
 
+// Hard-coded limit of delayed messages. If this limit is reached, we will
+// directly log all following messages. This is to prevent the queue from
+// growing indefinitely if the log thread is not able to keep up with the
+// incoming messages.
+
```
jsteemann, line 33: this comment seems to be a left-over and should be removed now.
```suggestion
```

lib/Logger/LoggerFeature.cpp
```diff
@@ -547,6 +547,21 @@ printed in the thread that triggered the log message. This is non-optimal for
...
+  options
+      ->addOption(
+          "--log.max-queued-entries",
+          "Upper limit of log entries that are queued in background thread.",
+          new UInt32Parameter(&_maxQueuedLogMessages),
+          arangodb::options::makeDefaultFlags(
+              arangodb::options::Flags::Uncommon))
+      .setIntroducedIn(31200)
+      .setLongDescription(
+          R"(If you are using the option log.force-direct (default)
```
jsteemann, line 559:
```suggestion
          R"(If you are using the option `--log.force-direct` (default)
```

PR #19996, opened 2023-10-18
Deal with race in leader change [BTS-1647][BTS-1658]

neunhoef:
### Scope & Purpose

This deals with https://arangodb.atlassian.net/browse/BTS-1647.

Short problem description:

Due to a race during moveShard for leader change it is possible that a shard gets into a state such that both leader and follower believe they are in sync, but the FollowerTermId is incorrectly set in the follower. This means that the follower is dropped on the first write operation to the shard.

This PR fixes this by properly cleaning up `_theLeader` on the follower in case of an error.

Furthermore, we ensure that something is logged on INFO level whenever a shard synchronization fails. This is crucial for debugging.

Testing this is tricky. Therefore we introduce a new system to serialize events in a local cluster globally. This is implemented in
```
  lib/Basics/GlobalSerialization.h
  lib/Basics/GlobalSerialization.cpp
```
and can be used to stage test scenarios that depend on such a global event order.

Finally, this fixes a second problem in leader change. If A was leader and B and C follower, and we switch from leader A to B, then C was never adopted as new follower and had to resync. This is due to a wrong check for the old leader, which is fixed now. I did not find a way to reliably test this, since very occasionally it can happen that due to some sequence of events a `SynchronizeShard` is triggered in this situation.

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **Regression tests**
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20049
  - [*] Backport for 3.10: https://github.com/arangodb/arangodb/pull/20050

#### Related Information

- [*] Enterprise PR: - 
- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1647.


jsteemann (approved):
LGTM

neunhoef:
This is not quite ready, since the testing is very tedious to implement. Will continue next week.

lib/Basics/GlobalSerialization.h
```diff
@@ -0,0 +1,82 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Max Neunhoeffer
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
```
MBkkt, line 23:
```suggestion
#pragma once

#if ARANGODB_ENABLE_FAILURE_TESTS
```

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+        // Now move the shard:
+        let res = moveShard("_system", cn, collInfo.shard, collInfo.leader, collInfo.follower, false /* dontwait */);
+
+        // See if the program was executed:
+        let count = 0;
+        let nrSteps;
+        while (count <= 600) {
+          let steps = fs.readFileSync(pcPath).toString().split("\n");
+          nrSteps = steps.length - 1;
+          if (nrSteps === 12) {
```
jsteemann, line 174: Probably better to do it like this, so there is a visible relationship with `prog` for the reader of the test code.
```suggestion
          if (nrSteps === prog.length - 1) {
```

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+  } catch (err) {
+    console.error(
+      "Exception for PUT /_admin/cluster/moveShard:", err.stack);
+    return false;
+  }
+  if (dontwait) {
+    return result;
+  }
+  // Now wait until the job we triggered is finished:
+  var count = 600;   // seconds
```
jsteemann, line 79:
```suggestion
  let count = 600;   // seconds
```

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+  while (true) {
+    var job = queryAgencyJob(result.parsedBody.id);
```
jsteemann, line 81:
```suggestion
    let job = queryAgencyJob(result.parsedBody.id);
```

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+let db = arangodb.db;
+let { debugCanUseFailAt, debugRemoveFailAt, debugSetFailAt, debugClearFailAt, waitForShardsInSync } = require('@arangodb/test-helper');
+
+function queryAgencyJob (id) {
+  let query = arango.GET(`/_admin/cluster/queryAgencyJob?id=${id}`);
+  if (query.status === "Finished") {
+    return {error: false, id, status: query.status};
+  }
+  return {error: true, errorMsg: "Did not find job.", id};
+};
```
jsteemann, line 40:
```suggestion
}
```

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+  return { endpointMap, idMap, coordinator, leader, follower, shard };
+}
+
+function moveShard(database, collection, shard, fromServer, toServer, dontwait) {
+  var body = {database, collection, shard, fromServer, toServer};
+  var result;
+  try {
+    result = arango.POST_RAW("/_admin/cluster/moveShard", body);
+  } catch (err) {
+    console.error(
```
jsteemann, line 71: `console` is not declared anywhere. Using it anyway may work fine, but not sure if it can trigger warnings when linting the test.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
```diff
@@ -0,0 +1,197 @@
...
+        while (count <= 600) {
+          let steps = fs.readFileSync(pcPath).toString().split("\n");
+          nrSteps = steps.length - 1;
+          if (nrSteps === 12) {
+            break;
+          }
+          count += 1;
+          internal.wait(1);
+        }
+        assertEqual(nrSteps, 12);
```
jsteemann, line 180:
```suggestion
        assertEqual(nrSteps, prog.length - 1);
```

lib/Basics/GlobalSerialization.cpp
```diff
@@ -0,0 +1,145 @@
...
+            << "Global event " << id << " with selector " << selector
+            << " and comment " << parts[2] << " has happened...";
+      } else {
+        LOG_TOPIC("ace34", INFO, Logger::MAINTENANCE)
+            << "Global event " << id << " with selector " << selector
+            << " has happened...";
+      }
+      current.push_back('\n');
+      try {
+        arangodb::basics::FileUtils::appendToFile(pcPath, current);
```
jsteemann, line 95: If multiple threads of the same arangod instance, or even different arangod instances write to the same `globalSLP_PC` file, this could be unsafe. In theory it is possible that one writer instance writes a line only partially, and then another writer kicks in. That is at least possible if multiple writers can act upon the same trigger concurrently. 
Not sure if this is actually possible in the test.

lib/Basics/GlobalSerialization.h
neunhoef: Good suggestion, this will enforce that such global sync points are only in failure tests! Will chang.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Done, needed to do a small change to the definition of prog for this to work.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Done.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Done.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Done.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Have explicitly declared `console` now.

tests/js/client/shell/shell-move-shard-sync-fail-cluster.js
neunhoef: Done.

lib/Basics/GlobalSerialization.cpp
neunhoef: Currently it is not possible. This is, because the file is only ever modified, if we are at a specific code place which matches the current line. So it is only one code place which can modify the file in a given state. If we can now ensure that this code place is not executed by two threads concurrently, we are good. As far as I see this is currently the case.
Secondly, the code only looks at the **number of lines** in the file, that is the number of '\n' bytes and this should be the same, even if lines are mixed, but this is of course nothing we can count on.
Therefore I completely agree that this is dangerous to be misused. Therefore, I suggest to put a global mutex into this code, ensuring that only one thread at a time calls `appendToFile` on this file.
We then have "only" the problem left that in the same cluster two such tests could be executed at the same time, which is inherent in the design of this system, and I currently do not see a way around this except if we would hand in the global name of the file into the cluster for each test, which I think is overkill and overcomplicated.
Therefore I suggest to live with this problem. I will add a loud warning in the header file.

neunhoef: Note that the path to the SLP program does contain the test-suite (e.g. shell-client), which somewhat limits the problem. And since we do not execute multiple such tests in the suite in parallel, we should be good.

lib/Basics/GlobalSerialization.h
neunhoef: It turns out that this is not possible. The thing is that `TRI_IF_FAILURE` is defined to be
```
#define TRI_IF_FAILURE(what) if constexpr (false)
```
if failure tests are not compiled in. That means that the code in this block must still compile, even if failure tests are not compiled in. Therefore we cannot hide the headers in this case. Reverting change.

jsteemann (approved):
LGTM

MBkkt (approved):

PR #20003, opened 2023-10-19
[Join] SkipRowsRange

maierlars:
### Scope & Purpose
Implement support for `LIMIT` calls.

This PR introduces some code duplication. This will be fixed in a follow up PR.


hkernbach (approved):
LGTM

hkernbach:
@maierlars tests now green. Can be merged if you're happy with my test-split solution.

PR #20020, opened 2023-10-24
[BTS-1599] Adding write-concern parameter to arangorestore.

maierlars:
### Scope & Purpose
Allow arangorestore to override the writeConcern value specified in a database dump via new option --write-concern for newly creating collections.
BTS: https://arangodb.atlassian.net/browse/BTS-1599


CHANGELOG
```diff
@@ -1,6 +1,10 @@
 devel
 -----
 
+* Added --write-concern parameter similar to --replication-factor to
+  arangodump. This enabled the user to override the write concern for
```
```suggestion
  arangodump. This enables the user to override the write concern for
```

client-tools/Restore/RestoreFeature.cpp
```diff
@@ -168,6 +168,44 @@ uint64_t getReplicationFactor(arangodb::RestoreFeature::Options const& options,
...
+  if (!options.writeConcern.empty()) {
+    std::string const name = s.copyString();
+
+    for (auto const& it : options.writeConcern) {
+      auto parts = arangodb::basics::StringUtils::split(it, '=');
+      if (parts.size() == 1) {
+        result = arangodb::basics::StringUtils::uint64(parts[0]);
+      }
+
+      if (parts.size() != 2 || parts[0] != name) {
```
jsteemann, line 197: The second `if` condition looks wrong, because if `parts.size() == 1` evaluates to `true` in line 193, we fetch the numeric result in line 194, but then immediately consider it invalid in line 197.
I guess the two `if`s here should be combined by an `else`.
```suggestion
      if (parts.size() == 1) {
        result = arangodb::basics::StringUtils::uint64(parts[0]);
      } else if (parts.size() != 2 || parts[0] != name) {
```

client-tools/Restore/RestoreFeature.cpp
```diff
@@ -1979,6 +2019,13 @@ avoiding repeated memory allocations for building new in-memory buffers.)");
       "--replication-factor myCollection=3).",
       new VectorParameter<StringParameter>(&_options.replicationFactor));
 
+  options->addOption(
+      "--write-concern",
+      "Override the `writeConcern` value (can be specified "
+      "multiple times, e.g. --write-concern 2 "
+      "--write-concern myCollection=3).",
+      new VectorParameter<StringParameter>(&_options.writeConcern));
```
jsteemann, line 2027: Let's add the version number in which the option was introduced.
```suggestion
      new VectorParameter<StringParameter>(&_options.writeConcern))
      .setIntroducedIn(31200);
```

client-tools/Restore/RestoreFeature.cpp
jsteemann: Or is this actually intentional, so that we pick up the value in line 194, but keep looking for if there is a better, more precise match?

maierlars: Yes, similar to `replication-factor` we can override for all collection with `--replication-factor 2` but then add this multiple times to handle specific collections differently `--replication-factor foo=3`.

jsteemann (approved):
LGTM

Issue #16346, opened 2022-05-30
Add "ignore-collection" option to the arangodump command


AntoineAA:
## My Environment

* __ArangoDB Version__:        3.9.0
* __Deployment Mode__:       Single Server
* __Deployment Strategy__:    systemd
* __Configuration__:               
* __Infrastructure__:               own
* __Operating System__:        Ubuntu 20.04
* __Total RAM in your machine__:         256GB
* __Disks in use__:         SSD
* __Used Package__:              Ubuntu .deb

## Component, Query & Data

__Affected feature__:
arangodump

__AQL query (if applicable)__:

__AQL explain and/or profile (if applicable)__:

__Dataset__:

__Size of your Dataset on disk__:
1.7TB

__Replication Factor & Number of Shards (Cluster only)__:

__Problem__:
It would be useful to add an "ignore-collection" option to dump the whole database except a few collections.
We use a CRON to do the dump, and we would have to update it each time we add a new collection.
We just need to exclude 1/27 collections.

__Expected result__:

Zyqsempai:
Hey @AntoineAA,
Thanks a lot for sharing with us this idea, our Product Team currently reviewing your request, we will circle back with an update in the next few weeks.
Best.

matcho:
Duplicate of #12712 btw ?

hkernbach:
Status update here: Feature is in the development process. Still needs to pass CI's and Reviews, but will then be merged soon.

matcho:
Works in 3.12

dothebart:
closing Since 3.12.0.2 is available for download.

PR #20021, opened 2023-10-24
[BTS-1602] Add --ignore-collection flag to arangodump

hkernbach:
### Scope & Purpose

Adds a new flag `--ignore-collection` for `arangodump`. 
Users can now define collections which should be excluded while using arangodump. 

- [x] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **integration tests**
- [x] :books: documentation written 

#### Related Information

- [x] Docs PR: https://github.com/arangodb/docs-hugo/pull/318
- [x] Jira ticket: https://arangodb.atlassian.net/browse/BTS-1602
- [x] GitHub issue: https://github.com/arangodb/arangodb/issues/16346


client-tools/Dump/DumpFeature.cpp
```diff
@@ -806,6 +806,12 @@ void DumpFeature::collectOptions(
       "times).",
       new VectorParameter<StringParameter>(&_options.collections));
 
+  options->addOption(
+      "--ignore-collection",
+      "Ignore and exclude this collection during the dump process (can be "
+      "specified multiple times).",
+      new VectorParameter<StringParameter>(&_options.collectionsToBeIgnored));
```
jsteemann, line 813:
```suggestion
      new VectorParameter<StringParameter>(&_options.collectionsToBeIgnored))
      .setIntroducedIn(31200);
```

client-tools/Dump/DumpFeature.cpp
```diff
@@ -1167,6 +1173,16 @@ Result DumpFeature::runDump(httpclient::SimpleHttpClient& client,
     if (cid == 0 || name.empty()) {
       return ::ErrorMalformedJsonResponse;
     }
+
+    // filtered based on the user-defined list of collections to be excluded
+    if (std::find(_options.collectionsToBeIgnored.begin(),
```
jsteemann, line 1178: I am generally not very happy with this option, because now we have `--collection` and `--ignore-collection`.
The former is an allow list and the latter is a deny list.
What is the expected behavior if a user makes use of both options? And what is the expected behavior if the same collection name is specified in both options?
This should be at least defined.
In the easiest case, we can simply forbid using `--collection` and `--ignore-collection` at the same time.

hkernbach: Understandable. I would vote for the either `--collection` or `--ignore-collection` option.

jsteemann (approved):
LGTM

PR #20023, opened 2023-10-24
Extract projections into registers - EnumerateCollection

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1404

Extract document attribute projections of `ENUMERATE_COLLECTION` nodes into individual output registers, instead of storing them in a combined VelocyPack object.
Previously, the projected attributes of a document were all stored together in a newly-built VelocyPack object. For example, if the projections "a" and "b" were used, a new VelocyPack object `{"a":..., "b": ... }` was built and stored in the ExecutionNode's main output register.

While this saved registers and allowed all following query parts to remain unchanged, the approach had several drawbacks:
* the attribute names were part of the generated VelocyPack object, so even for an object with a single projected attribute, it was possible that the VelocyPack object did not fit into the SSO buffer of an AqlValue.
* looking up an attribute value from the generated VelocyPack object consumed CPU time, because in an object with multiple attributes, the correct one needed to be found. This requires a linear search or binary search in the object for the correct attribute name and potentially several string comparisons.

By storing the projected attribute values in dedicated output registers, each resulting value can be stored more compact than before. Additionally, there is no need to look up a specific attribute inside some combined result object. Now, the value can be found directly by looking into the correct output register.

The optimization is performed now by the already existing "remove-unnecessary-projections" optimizer rule, which was renamed to "optimize-projections" as part of this PR. The rule also performs the transformation of expressions such as `doc.attribute` into `#register` so that the query results remain unchanged.
When the optimization is performed, it is often also possible to remove now-redundant CalculationNodes, which previously only extracted individual attributes from the combined result object. This can also lead to shorter execution plans and thus faster overall execution.

The explain output is adjusted so that instead of
```
Execution plan:
 Id   NodeType                  Par   Est.   Comment
  1   SingletonNode                      1   * ROOT
  2   EnumerateCollectionNode     ✓      0     - FOR doc IN cv   /* full collection scan (projections: `fuchs`), scan only  */
  3   CalculationNode             ✓      0       - LET #1 = doc.`fuchs`   /* attribute expression */   /* collections used: doc : cv */
  4   ReturnNode                         0       - RETURN #1
```
it will now display
```
Execution plan:
 Id   NodeType                  Par   Est.   Comment
  1   SingletonNode                      1   * ROOT
  2   EnumerateCollectionNode     ✓      0     - FOR doc IN cv   /* full collection scan (projections: `fuchs`)  */   LET #3 = doc.`fuchs`
  4   ReturnNode                         0       - RETURN #3
```

The optimization is currently only performed for EnumerateCollectionNode, not yet for IndexNode, JoinNode and/or TraversalNode. These may follow in separate PRs.

The PR also enables enables the usage of projections when an EnumerateViewNode is used in the query. Previously, when the result variable of an EnumerateCollection node was used inside an EnumerateView node, it was always assumed that the full document needed to be used, which disabled projections.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1404/
- [ ] Design document:


arangod/Aql/Projections.h
```diff
@@ -93,6 +95,9 @@ class Projections {
   /// @brief reset all projections
   void clear() noexcept;
 
+  /// @brief erase projection at index
+  void erase(size_t index);
```
MBkkt, line 99: I think better to make method that accept boolean filter.
So this method can use std::remove_if to avoid quadratic complexity for cycles with erase

arangod/Aql/OptimizerUtils.cpp
```diff
@@ -1007,6 +1008,13 @@ bool findProjections(ExecutionNode* n, Variable const* v,
       if (!checkExpression(calculationNode, calculationNode->expression())) {
         return false;
       }
+    } else if (current->getType() == EN::ENUMERATE_IRESEARCH_VIEW) {
+      iresearch::IResearchViewNode const* viewNode =
+          ExecutionNode::castTo<iresearch::IResearchViewNode const*>(current);
+      if (!tryAndExtractProjectionsFromExpression(
+              viewNode, &viewNode->filterCondition())) {
+        return false;
+      }
```
MBkkt, line 1017: Is it for cases like this to extract `d1.x` access for collection?
```
FOR d1 IN collection
  FOR d2 IN view
    SEARCH d1.x == d2.x
    ...
```

MBkkt: > The PR also enables enables the usage of projections when an EnumerateViewNode is used in the query. Previously, when the result variable of an EnumerateCollection node was used inside an EnumerateView node, it was always assumed that the full document needed to be used, which disabled projections.

Understand

arangod/Aql/Projections.h
jsteemann: Changed

arangod/Aql/Projections.cpp
```diff
@@ -101,6 +107,175 @@ uint16_t Projections::coveringIndexPosition(
...
+
+      b.clear();
+      b.add(VPackValue(transaction::helpers::extractIdString(trxPtr->resolver(),
+                                                             slice, slice)));
+      cb(it.variable, b.slice());
+    } else if (it.type == AttributeNamePath::Type::KeyAttribute) {
+      // projection for "_key"
+      TRI_ASSERT(levelsOpen == 0);
+      TRI_ASSERT(it.path.size() == 1);
+      b.clear();
```
maierlars, line 135: Not sure why we have to clear `b` here.

jsteemann: `b` is just a temporary scratch buffer here.
The function only uses `b` to write the `_id` value into the Builder, and then calls the callback with a VPackSlice pointing to that Builder.
Because at most one `_id` value is in the Builder, we should clear it before writing into it.
The caller does not expect anything to be in `b` when the function returns, so we should be ok.

MBkkt (approved):

PR #20028, opened 2023-10-25
Agency Nodes Free list Fix

maierlars:
### Scope & Purpose
Since this bug is still present https://github.com/arximboldi/immer/issues/182 we should use a different memory policy for the agency nodes.


jsteemann (approved):
LGTM

PR #20031, opened 2023-10-26
Chore/circleci mac

CryptoNinjaGeek:
### Scope & Purpose

Support for mac-tools building on PRs

- [ ] :hankey: Bugfix
- [X] :pizza: New feature


.circleci/base_config.yml
```diff
@@ -261,6 +261,40 @@ jobs:
...
+          destination: "~/project"
+      - run:
+          name: "Install"
+          command: |
+            brew install cmake openssl@3 llvm@14
+      - run:
+          name: "Configure"
+          command: |
+            export PATH="/opt/homebrew/opt/llvm@14/bin:$PATH"
+            export MACOSX_DEPLOYMENT_TARGET=13.2
```
KVS85, line 288: Too new deployment target. Should be adjusted with PM.

.circleci/base_config.yml
```diff
@@ -261,6 +261,40 @@ jobs:
...
+      xcode: 14.2.0
+    resource_class: macos.m1.medium.gen1
+    steps:
+      - checkout-arangodb:
+          with-submodules: true
+          destination: "~/project"
+      - run:
+          name: "Install"
+          command: |
+            brew install cmake openssl@3 llvm@14
```
KVS85, line 283: Doesn't match exact OpenSSL version we use through `.\VERSIONS`.

.circleci/base_config.yml
```diff
@@ -261,6 +261,40 @@ jobs:
...
+      - run:
+          name: "Configure"
+          command: |
+            export PATH="/opt/homebrew/opt/llvm@14/bin:$PATH"
+            export MACOSX_DEPLOYMENT_TARGET=13.2
+            mkdir ~/build
+            cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -S ~/project -B ~/build
```
jsteemann, line 290: For other platforms, we can build in various configurations, e.g. Release vs. Debug, Enterprise vs. Community etc.

I guess here we intentionally do not care about `CMAKE_BUILD_TYPE`, `USE_ENTERPRISE`, `USE_STRICT_OPENSSL_VERSION`, `USE_MAINTAINER_MODE` and other CMake options, as we are only checking whether the client tools compile on macOS or not.
Probably adding `USE_ENTERPRISE` is necessary because this flag will have an influence on whether or not we are building arangobackup.

KVS85 (changes requested):
x86-64 mac and CE build are missing.

.circleci/base_config.yml
CryptoNinjaGeek: Yes because its compiling mac tools, i felt those didn't make sense.
Ill make the USE_ENTERPRISE changes

.circleci/base_config.yml
CryptoNinjaGeek: x86-64 mac on CircleCI is already deprecated and being removed. So that would need to be handles in a different way.

.circleci/base_config.yml
CryptoNinjaGeek: Its using 3.1.3, which is the same as ours.
But I get your comment that its not directly specified, will look at a solution.
All have to be build on each PR, as we cant have images like with windows and Linux

.circleci/base_config.yml
CryptoNinjaGeek: now build using enterprise-pr and community-pr preset for both m1 and x86

.circleci/base_config.yml
CryptoNinjaGeek: OpenSSL version taken from VERSIONS is used on build

jsteemann (approved):
LGTM

PR #20033, opened 2023-10-26
Clean up collection if it was created successfully.

maierlars:
### Scope & Purpose
As the comment says, it might or might not happen that the collection `cname + "2"` is created. If it is created, the test does not bother cleaning it up. And the test framework complains about a missing cleanup.

Thus, create it and then delete it again.


jsteemann (approved):
LGTM

hkernbach (approved):

PR #20036, opened 2023-10-26
Implement SupervisedScheduler::detachThread [BTS-1670]

neunhoef:
This is one measure to resolve
  https://arangodb.atlassian.net/browse/BTS-1670

In this PR we create a possiblity that a SchedulerThread can notice that 
it might be up to some longer running task (like waiting for a lock) and 
then voluntarily decide to detach itself from the scheduler to make room
for a new thread to avoid starvation of jobs on the queue.

For now, we use this for acquiring a collection level lock (as read lock 
for write operations and as write lock for exclusive access), if we 
cannot acquire the lock within 1s. This will then allow the scheduler
to create new threads to continue work on whatever priority is suitable.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] C++ **Unit tests**
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20067
  - [ ] Backport for 3.10: not planned

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1670


maierlars (approved):

CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* Allow a scheduler thread to detach itself from the scheduler if it sees 
+  that it has to perform a potentially long running task like waiting For
```
jsteemann, line 5: nit
```suggestion
  that it has to perform a potentially long running task like waiting for
```

CHANGELOG
```diff
@@ -1,6 +1,14 @@
...
+  a lock. This allows a new scheduler thread to be started and avoids
+  that it can happen that all threads are blocked waiting for a lock,
+  which has in the past lead to deadlock situations. The number of detached
```
jsteemann, line 8: nit
```suggestion
  which has in the past led to deadlock situations. The number of detached
```

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1715,6 +1716,31 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
       std::chrono::duration<double>(timeout));
 
   bool gotLock = false;
+  if (timeout_us > std::chrono::microseconds(1000000)) {
```
jsteemann, line 1719: nit: the value `std::chrono::microseconds(1000000)` is used 4 times in this code. IMHO it would be better to put the threshold value into a constexpr once, and then refer to it only by its symbol name.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -500,6 +506,44 @@ void SupervisedScheduler::shutdown() {
...
+    if ((*it)->_thread->threadNumber() == myNumber) {
+      break;
+    }
+    ++it;
+  }
+  if (it == _workerStates.end()) {
+    return Result(TRI_ERROR_INTERNAL,
+                  "scheduler thread for detaching not found");
+  }
+  std::shared_ptr<WorkerState> state = *it;
```
jsteemann, line 530: nit:
```suggestion
  std::shared_ptr<WorkerState> state = std::move(*it);
```

arangod/Scheduler/SupervisedScheduler.h
```diff
@@ -186,16 +196,19 @@ class SupervisedScheduler final : public Scheduler {
...
   std::list<std::shared_ptr<WorkerState>> _workerStates;
   std::list<std::shared_ptr<WorkerState>> _abandonedWorkerStates;
-  std::atomic<uint64_t> _numWorking;  // Number of threads actually working
-  std::atomic<uint64_t> _numAwake;    // Number of threads working or spinning
-                                      // (i.e. not sleeping)
+  std::list<std::shared_ptr<WorkerState>> _detachedWorkerStates;
+  std::atomic<uint64_t> _numWorking;   // Number of threads actually working
+  std::atomic<uint64_t> _numAwake;     // Number of threads working or spinning
+                                       // (i.e. not sleeping)
+  std::atomic<uint64_t> _numDetached;  // Number of detached threads
```
jsteemann, line 211: I don't see an initializer for `_numDetached` here or in the .cpp file. Doesn't that mean that it has an unspecified value upon initialization?

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -500,6 +506,44 @@ void SupervisedScheduler::shutdown() {
...
+  // Since the thread is effectively taken out of the pool, decrease the
+  // number of workers.
+  --_numWorkers;
+  {
+    std::unique_lock<std::mutex> guard(state->_mutex);
+    state->_stop = true;  // We will be stopped after the current task is done
+                          // We know that we are working, so we do not
+                          // have to wake the thread.
+  }
+  ++_metricsThreadsStopped;
```
jsteemann, line 541: Is it ok to count the thread as `stopped` here already? The thread is still about to perform one more task, which can take an arbitrary amount of time. And until it has finished, the `stopped` counter would be off by one.

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -500,6 +506,44 @@ void SupervisedScheduler::shutdown() {
...
+  _detachedWorkerStates.push_back(std::move(state));
```
jsteemann, line 542: If this `push_back` operation fails because OOM, does it mean we will be in an undefined state, potentially with some resources leaked?
We could avoid this by first reserving some more memory in `_detachedWorkerStates` before making any modifications to the scheduler state, and only performing any state mutations afterwards.

arangod/Transaction/Manager.cpp
```diff
@@ -785,10 +785,15 @@ std::shared_ptr<transaction::Context> Manager::leaseManagedTrx(
...
   if (!ServerState::isDBServer(role)) {  // keep end time as small as possible
-    endTime = std::chrono::steady_clock::now() +
-              std::chrono::milliseconds(int64_t(1000 * _streamingLockTimeout));
+    endTime =
+        now + std::chrono::milliseconds(int64_t(1000 * _streamingLockTimeout));
+  }
+  std::chrono::steady_clock::time_point detachTime;
+  if (_streamingLockTimeout >= 1.0) {
+    detachTime = now + std::chrono::milliseconds(1000);
   }
```
jsteemann, line 797: Right now the detaching is not restricted to the instance types single server or DB server.
Do we want it to happen on coordinators as well, or just single servers and DB servers?

CHANGELOG
neunhoef: Done.

CHANGELOG
neunhoef: Done.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1715,6 +1716,31 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
...
+          _exclusiveLock.tryLockReadFor(std::chrono::microseconds(1000000));
+    }
+    if (gotLock) {
+      // keep the lock and exit
+      return TRI_ERROR_NO_ERROR;
+    }
+    LOG_TOPIC("dd231", INFO, Logger::THREADS)
+        << "Did not get lock within 1 seconds, detaching scheduler thread.";
+    Result r = arangodb::SchedulerFeature::SCHEDULER->detachThread();
+    if (r.fail() && r.is(TRI_ERROR_TOO_MANY_DETACHED_THREADS)) {
```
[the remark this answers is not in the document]
neunhoef: Done.

arangod/Scheduler/SupervisedScheduler.cpp
neunhoef: Done.

arangod/Scheduler/SupervisedScheduler.cpp
neunhoef: I think this is correct or at least consistent with prior work. I copy/pasted it from `SupervisedScheduler::StopOneThread` and there it is exactly the same. There it is exactly the same, the thread is still running and working, but already counts as stopped for the purpose of this metric. I would prefer to leave it this way.

arangod/Scheduler/SupervisedScheduler.cpp
neunhoef: I think this is OK, since the thread itself still holds a copy of the `shared_ptr` to its own state. When the current task finishes and the thread asks for more work, it will notice that `_stop` in its own state is set to `true` and will terminate. The only thing which will happen is that `cleanupAbandonedStates` will not see it and will not count down `_numDetached`. Therefore I have now added a `try/catch` statement to ignore the error, but only count the `_numDetached`, if the `push_back` has worked.

arangod/Scheduler/SupervisedScheduler.h
neunhoef: I added a zero initializer in the constructor.

arangod/Transaction/Manager.cpp
neunhoef: I think we want detaching on all types of servers, even on agents. It is a security measure and could potentially be relevant everywhere.

arangod/Transaction/Manager.cpp
```diff
@@ -898,8 +903,19 @@ std::shared_ptr<transaction::Context> Manager::leaseManagedTrx(
...
                !ServerState::instance()->isDBServer());
 
-    if (!ServerState::isDBServer(role) &&
-        std::chrono::steady_clock::now() > endTime) {
+    auto now = std::chrono::steady_clock::now();
+    if (detachTime.time_since_epoch().count() != 0 && now > detachTime) {
+      LOG_TOPIC("dd234", INFO, Logger::THREADS)
+          << "Did not get lock within 1 seconds, detaching scheduler thread.";
+      auto res = SchedulerFeature::SCHEDULER->detachThread();
+      if (res.fail() && res.is(TRI_ERROR_TOO_MANY_DETACHED_THREADS)) {
```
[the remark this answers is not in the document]
neunhoef: Done.

jsteemann (approved):
LGTM

neunhoef:
Please do not merge yet. This seems to work, but I want to add another test or two.

tests/js/client/server_parameters/detach-threads-cluster.js
```diff
@@ -0,0 +1,177 @@
...
+          let r = arango.POST_RAW(`/_api/document/${cn}`, {_key:"L"+i}, {"x-arango-async": "store", "x-arango-trx-id": trx});
+          assertEqual(202, r.code);
+          assertTrue(r.headers.hasOwnProperty("x-arango-async-id"));
+          l.push(r.headers["x-arango-async-id"]);
+        }
+
+        // Then perform simple reads to the collection. They should soon be
+        // unblocked, because of detached threads.
+        let startTime = new Date();
+        let guck = c.document("K1");
```
jsteemann, line 139: `guck` is not used later.
```suggestion
        c.document("K1");
```

jsteemann (approved):
LGTM

tests/js/client/server_parameters/detach-threads-cluster.js
```diff
@@ -0,0 +1,177 @@
...
+        for (let i = 0; i < 100; ++i) {
+          let r = arango.POST_RAW(`/_api/document/${cn}`, {_key:"L"+i}, {"x-arango-async": "store", "x-arango-trx-id": trx});
+          assertEqual(202, r.code);
+          assertTrue(r.headers.hasOwnProperty("x-arango-async-id"));
+          l.push(r.headers["x-arango-async-id"]);
+        }
+
+        // Then perform simple reads to the collection. They should soon be
+        // unblocked, because of detached threads.
+        let startTime = new Date();
```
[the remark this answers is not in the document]
neunhoef: Removed.

tests/js/client/server_parameters/detach-threads-cluster.js
neunhoef: Removed

PR #20038, opened 2023-10-27
[BTS-1661] Added new graph (vertex related) error code

hkernbach:
### Scope & Purpose

Added more detailed error code related to the use of not available/defined vertex collections inside graph operations.
In some operations, the error code: `TRI_ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_USED` now got
replaced with the error code: `TRI_ERROR_GRAPH_VERTEX_COLLECTION_NOT_USED`. 

The difference is that the `referenced not used` means that the vertex collection is not defined within the scope of the graph definition (e.g. a vertex is used inside a `_from` or `_to` attribute of an edge document). 

The newly introduced error checks if a vertex collection is used at all in general, e.g.:
- One tries to insert a new vertex document to a vertex collection which is not part of the graph
- Same analogy applies for removals, replaces, etc.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**

#### Related Information

- [ ] Docs PR: 
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1661


hkernbach:
Review can be done, but added `DO_NOT_MERGE` label. Jenkins is in maintenance, therefore I cannot start our additional CI there

hkernbach:
Note: Also includes [BTS-1660] Fix PATCH validation in some gharial API related graph operations, see: https://github.com/arangodb/arangodb/pull/20038/commits/d798726bfeaa06c6ada7ee9c531410f01ac0236f

arangod/Graph/GraphOperations.cpp
```diff
@@ -201,7 +202,18 @@ Result GraphOperations::checkVertexCollectionAvailability(
...
+    if (edgeDocumentOrigin == VertexValidationOrigin::FROM_ATTRIBUTE ||
+        edgeDocumentOrigin == VertexValidationOrigin::TO_ATTRIBUTE) {
+      std::string originString = "_from";
+      if (edgeDocumentOrigin == VertexValidationOrigin::TO_ATTRIBUTE) {
+        originString = "_to";
+      }
+
+      return Result(TRI_ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_USED,
+                    "referenced " + originString + " collection '" +
+                        vertexCollectionName + "' is not part of the graph");
```
jsteemann, line 214: nit
```suggestion
      std::string_view originString = "_from";
      if (edgeDocumentOrigin == VertexValidationOrigin::TO_ATTRIBUTE) {
        originString = "_to";
      }

      return Result(TRI_ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_USED,
                    absl::StrCat("referenced ", originString, " collection '",
                        vertexCollectionName, "' is not part of the graph"));
```

arangod/Graph/GraphOperations.cpp
```diff
@@ -844,8 +856,74 @@ std::pair<OperationResult, bool> GraphOperations::validateEdgeContent(
...
+      if (pos != std::string::npos) {
+        fromCollectionName = fromString.substr(0, pos);
+        // check if vertex collections are part of the graph definition
+        auto foundFromRes = checkVertexCollectionAvailability(
+            fromCollectionName, VertexValidationOrigin::FROM_ATTRIBUTE);
+        if (foundFromRes.fail()) {
+          return OperationResult(foundFromRes, options);
+        }
+
+        fromCollectionKey = fromString.substr(pos + 1, fromString.length());
```
jsteemann, line 878: nit: while this works, the usage of std::strings here is not efficient, as we create up to 3 std::strings for the _from and up to 3 std::strings for the _to parts. Using std::string_views here would have done it as well, but it may require the validation function `checkVertexCollectionAvailability` to use std::string_view as well.
So I am not insisting on the change now.

js/common/bootstrap/errors.js
```diff
@@ -286,6 +286,7 @@
     "ERROR_GRAPH_NO_INITIAL_COLLECTION" : { "code" : 1946, "message" : "no valid initial collection found" },
     "ERROR_GRAPH_REFERENCED_VERTEX_COLLECTION_NOT_USED" : { "code" : 1947, "message" : "referenced vertex collection is not part of the graph" },
     "ERROR_GRAPH_NEGATIVE_EDGE_WEIGHT" : { "code" : 1948, "message" : "negative edge weight found" },
+    "ERROR_GRAPH_VERTEX_COLLECTION_NOT_USED" : { "code" : 1949, "message" : "vertex collection not used in graph" },
```
jsteemann, line 289: Is this really "not used in graph" or rather "not part of the graph"?
```suggestion
    "ERROR_GRAPH_VERTEX_COLLECTION_NOT_USED" : { "code" : 1949, "message" : "vertex collection not used in graph" },
```

jsteemann (approved):
LGTM

js/common/bootstrap/errors.js
hkernbach: Thanks, will change that and make it more clear

arangod/Graph/GraphOperations.cpp
hkernbach: I do agree, but yes this will be a bigger change. Quickly tested it and a lot of things need to be adjusted then. So let's do this in a follow up pr.

PR #20039, opened 2023-10-27
[BTS-1660] Fix PATCH validation in some gharial API related graph operations

hkernbach:
### Scope & Purpose

Original bug description:
```
When updating (PATCH) edges with Gharial and only specifying one of the two edge attributes
(_from or _to but not both), there doesn't seem to be any validation for the referenced vertex 
other than there needs to be a / with at least one character to the left and right.
```

This behaviour is now fixed. 

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**

#### Related Information

- [ ] Docs PR: 
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1660


hkernbach:
Review can be done, but added `DO_NOT_MERGE` label. Jenkins is in maintenance, therefore I cannot start our additional CI there

PR #20050, opened 2023-10-31
Deal with race in leader change [BTS-1647][BTS-1658]

neunhoef:
### Scope & Purpose

This deals with https://arangodb.atlassian.net/browse/BTS-1647.

This is a backport of https://github.com/arangodb/arangodb/pull/19996

Short problem description:

Due to a race during moveShard for leader change it is possible that a shard gets into a state such that both leader and follower believe they are in sync, but the FollowerTermId is incorrectly set in the follower. This means that the follower is dropped on the first write operation to the shard.

This PR fixes this by properly cleaning up `_theLeader` on the follower in case of an error.

Furthermore, we ensure that something is logged on INFO level whenever a shard synchronization fails. This is crucial for debugging.

Testing this is tricky. Therefore we introduce a new system to serialize events in a local cluster globally. This is implemented in
```
  lib/Basics/GlobalSerialization.h
  lib/Basics/GlobalSerialization.cpp
```
and can be used to stage test scenarios that depend on such a global event order.

Finally, this fixes a second problem in leader change. If A was leader and B and C follower, and we switch from leader A to B, then C was never adopted as new follower and had to resync. This is due to a wrong check for the old leader, which is fixed now. I did not find a way to reliably test this, since very occasionally it can happen that due to some sequence of events a `SynchronizeShard` is triggered in this situation.

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **Regression tests**
  - [*] **integration tests**
- [*] Backports
  - [*] Backport for 3.10: this is it

#### Related Information

- [*] Enterprise PR: - 
- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1647.
- [*] Devel PR: https://github.com/arangodb/arangodb/pull/19996

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,14 @@
+v3.10.12 (XXXX-XX-XX)
+---------------------
+
+* Fixed a race in controlled leader change which could lead to a situation 
+  in which a shard follower is dropped when the first write operation
+  happens. This fixes BTS-1647.
+
+* Fixed an unnecessary follower drop in controlled leader change, which will
+  speed up leader changes. This fixes BTS-1658.
+
+
 v3.10.11 (2023-10-19)
 ---------------------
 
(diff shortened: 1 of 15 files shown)


MBkkt (approved):

jsteemann (approved):
LGTM

KVS85:
- [ ] https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/31971/

PR #20051, opened 2023-10-31
Coroutines for Futures

maierlars:
### Scope & Purpose
This PR moves the quick and dirty implementation that makes futures compatible with coroutines into futures library such that they are available for all users of the futures. 

This implementation is retrofitted onto the existing futures library and can be improved once futures are designed in a coroutine first way.

### DIFF
--- a/lib/Futures/CMakeLists.txt
+++ b/lib/Futures/CMakeLists.txt
@@ -2,8 +2,12 @@ add_library(arango_futures STATIC
   src/Future.cpp)
 
 target_include_directories(arango_futures
-  PUBLIC
-  ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
-  include)
+        PUBLIC
+        ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
+        include)
+
+target_include_directories(arango_futures
+        PUBLIC
+        ${PROJECT_SOURCE_DIR}/lib/Inspection/include)
 
 target_link_libraries(arango_futures arango_assertions)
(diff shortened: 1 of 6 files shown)


lib/Futures/CMakeLists.txt
```diff
@@ -2,8 +2,12 @@ add_library(arango_futures STATIC
...
-  PUBLIC
-  ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
-  include)
+        PUBLIC
+        ${PROJECT_SOURCE_DIR}/3rdParty/iresearch/external/function2
+        include)
+
+target_include_directories(arango_futures
+        PUBLIC
+        ${PROJECT_SOURCE_DIR}/lib/Inspection/include)
```
maierlars, line 11: Include was necessary, because `ResultT` headers use some `Inspector` types to make it serializable.

MBkkt (approved):

lib/Futures/CMakeLists.txt
markuspf: Can you not just use 
```
target_link_libraries(arango_futures, INTERFACE arango_inspection)
``` 
instead of that hack?

lib/Futures/include/Futures/coro-helper.h
```diff
@@ -30,10 +30,10 @@ namespace std_coro = std;
...
 #include <optional>
 
-#include <Basics/Exceptions.h>
-#include <Basics/Result.h>
-#include <Futures/Promise.h>
-#include <Futures/Try.h>
+#include "Basics/Exceptions.h"
+#include "Basics/Result.h"
+#include "Promise.h"
+#include "Try.h"
```
markuspf, line 36: Why did you remove that `<Futures/Promise.h>` and `<Futures/Try.h>` prefix?

markuspf (changes requested):
In addition to the raised questions, would it be possible to have some tests for this functionality?

PR #20053, opened 2023-10-31
Extract projections in registers - Index

jsteemann:
### Scope & Purpose

Follow-up PR to https://github.com/arangodb/arangodb/pull/20023

Extract document attribute projections of `INDEX` nodes and `JOIN` nodes into individual output registers, instead of storing them in a combined VelocyPack object.
Previously, the projected attributes of a document were all stored together in a newly-built VelocyPack object. For example, if the projections "a" and "b" were used, a new VelocyPack object {"a":..., "b": ... } was built and stored in the ExecutionNode's main output register.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/JoinExecutor.cpp
+++ b/arangod/Aql/JoinExecutor.cpp
@@ -36,6 +36,31 @@ using namespace arangodb::aql;
 
 #define LOG_JOIN LOG_DEVEL_IF(false)
 
+RegisterId JoinExecutorInfos::registerForVariable(
+    VariableId id) const noexcept {
+  auto it = varsToRegister.find(id);
+  if (it != varsToRegister.end()) {
+    return it->second;
+  }
+  return RegisterId::maxRegisterId;
+}
+
+void JoinExecutorInfos::determineProjectionsForRegisters() {
+  if (!projectionsInitialized) {
+    for (auto& it : indexes) {
+      bool found = false;
+      for (size_t i = 0; i < it.projections.size(); ++i) {
+        if (it.projections[i].variable != nullptr) {
+          found = true;
+          break;
+        }
+      }
+      it.hasProjectionsForRegisters = found;
+    }
+    projectionsInitialized = true;
+  }
+}
+
 JoinExecutor::~JoinExecutor() = default;
 
 JoinExecutor::JoinExecutor(Fetcher& fetcher, Infos& infos)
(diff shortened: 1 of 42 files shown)


arangod/Aql/JoinExecutor.cpp
```diff
@@ -265,13 +305,28 @@ auto JoinExecutor::produceRows(AqlItemBlockInputRange& inputRange,
...
                                   _currentRow, docPtr);
           } else {
-            _projectionsBuilder.clear();
-            _projectionsBuilder.openObject(true);
-            idx.projections.toVelocyPackFromDocument(_projectionsBuilder, doc,
-                                                     &_trx);
-            _projectionsBuilder.close();
-            output.moveValueInto(_infos.indexes[k].documentOutputRegister,
-                                 _currentRow, _projectionsBuilder.slice());
+            if (!idx.hasProjectionsForRegisters) {
```
MBkkt, line 308: else if?

MBkkt (approved):

arangod/Aql/JoinNode.cpp
```diff
@@ -415,6 +478,19 @@ std::vector<JoinNode::IndexInfo>& JoinNode::getIndexInfos() {
   return _indexInfos;
 }
 
+bool JoinNode::isDeterministic() {
+  for (auto const& it : _indexInfos) {
```
mchacki, line 482: Isn't this just equal to return false?

mchacki (commented):
In general LGTM!

There is only one function which looks rather complicated but can only return false on all paths.
Is this intentional or should one exit actually be true?

arangod/Aql/JoinNode.cpp
```diff
@@ -415,6 +478,19 @@ std::vector<JoinNode::IndexInfo>& JoinNode::getIndexInfos() {
...
+    if (it.condition != nullptr && it.condition->root() != nullptr &&
+        !it.condition->root()->isDeterministic()) {
+      return false;
+    }
+    if (it.filter != nullptr && !it.filter->isDeterministic()) {
+      return false;
+    }
+  }
+  return false;
```
jsteemann, line 491:
```suggestion
  return true;
```

arangod/Aql/JoinNode.cpp
jsteemann: Indeed. Fixed!

mchacki (approved):
LGTM!

PR #20056, opened 2023-11-02
Silence TSAN w.r.t. SCHEDULER pointer on shutdown [BTS-1671]

neunhoef:
### Scope & Purpose

This addresses https://arangodb.atlassian.net/browse/BTS-1671.

It is to silence TSAN in some shutdown cases in RTA.

There is a global pointer called SchedulerFeature::SCHEDULER. On
shutdown, we set this pointer to nullptr in the
SchedulerFeature::unprepare method. In various other places, we read out
the pointer.

The logic of ApplicationFeatures should ensure that nobody will read
the pointer out after the SchedulerFeature has run its unprepare
method. Sometimes the TSAN sanitizer cannot recognize this indirect
synchronization and complains about reads that have happened before the
write there, but are not officially inter-thread synchronized. We use
atomic references in some places to silence TSAN.

- [*] :hankey: Bugfix

### Checklist

- [*] :book: CHANGELOG entry made
- [ ] Backports
  - [*] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20062
  - [ ] Backport for 3.10:

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1671

PR #20062, opened 2023-11-03
Silence TSAN w.r.t. SCHEDULER pointer on shutdown

neunhoef:
### Scope & Purpose

This is a backport of https://github.com/arangodb/arangodb/pull/20056 .

This addresses https://arangodb.atlassian.net/browse/BTS-1671.

It is to silence TSAN in some shutdown cases in RTA.

There is a global pointer called SchedulerFeature::SCHEDULER. On
shutdown, we set this pointer to nullptr in the
SchedulerFeature::unprepare method. In various other places, we read out
the pointer.

The logic of ApplicationFeatures should ensure that nobody will read
the pointer out after the SchedulerFeature has run its unprepare
method. Sometimes the TSAN sanitizer cannot recognize this indirect
synchronization and complains about reads that have happened before the
write there, but are not officially inter-thread synchronized. We use
atomic references in some places to silence TSAN.

- [*] :hankey: Bugfix

### Checklist

- [*] Backports

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1671
- [*] Devel PR: https://github.com/arangodb/arangodb/pull/20056

### DIFF
--- a/arangod/Network/Methods.cpp
+++ b/arangod/Network/Methods.cpp
@@ -259,7 +259,14 @@ void actuallySendRequest(std::shared_ptr<Pack>&& p, ConnectionPool* pool,
           return;
         }
 
-        auto* sch = SchedulerFeature::SCHEDULER;
+        // We access the global SCHEDULER pointer here via an atomic
+        // reference. This is to silence TSAN, which often detects a data
+        // race on this pointer, since this read access here can occasionally
+        // happen before the write in SchedulerFeature::unprepare, which
+        // invalidates the pointer. But even if the read here would happen
+        // later, we check for nullptr below, so all would be good.
+        std::atomic_ref<Scheduler*> schedulerRef{SchedulerFeature::SCHEDULER};
+        auto* sch = schedulerRef.load(std::memory_order_relaxed);
         // cppcheck-suppress accessMoved
         if (p->skipScheduler || sch == nullptr) {
           p->promise.setValue(network::Response{
--- a/arangod/Network/Methods.cpp
+++ b/arangod/Network/Methods.cpp
@@ -265,7 +265,8 @@ void actuallySendRequest(std::shared_ptr<Pack>&& p, ConnectionPool* pool,
         // happen before the write in SchedulerFeature::unprepare, which
         // invalidates the pointer. But even if the read here would happen
         // later, we check for nullptr below, so all would be good.
-        std::atomic_ref<Scheduler*> schedulerRef{SchedulerFeature::SCHEDULER};
+        std::atomic_ref<SupervisedScheduler*> schedulerRef{
+            SchedulerFeature::SCHEDULER};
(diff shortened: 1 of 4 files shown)


arangod/Network/Methods.cpp
```diff
@@ -259,7 +259,15 @@ void actuallySendRequest(std::shared_ptr<Pack>&& p, ConnectionPool* pool,
...
-        auto* sch = SchedulerFeature::SCHEDULER;
+        // We access the global SCHEDULER pointer here via an atomic
+        // reference. This is to silence TSAN, which often detects a data
+        // race on this pointer, since this read access here can occasionally
+        // happen before the write in SchedulerFeature::unprepare, which
+        // invalidates the pointer. But even if the read here would happen
+        // later, we check for nullptr below, so all would be good.
+        std::atomic_ref<SupervisedScheduler*> schedulerRef{
+            SchedulerFeature::SCHEDULER};
+        auto* sch = schedulerRef.load(std::memory_order_relaxed);
```
MBkkt, line 270: LGTM for code, but comments look strange for me.
If null possible when some synchronization needed, it can be explicit like here or via HB relation.
So this "silence TSAN" sounds wrong.
If exist some HB relation that TSAN ignore, feel free to point that.

dothebart (approved):

arangod/Network/Methods.cpp
neunhoef: I try to improve my comment.

MBkkt: What hb relation exists what tsan doesn't see?

neunhoef: I am not claiming that there is an official HB relation which TSAN does not see. I am only saying that it sees none. It might very well be that there exists none. This place in the code is the place which handles an incoming network response. It is entirely possible that a network response comes in when the scheduler is already shut down. Therefore, we check here if the pointer is already nullptr and act accordingly. Usually, this does not happen, not even in tests. But TSAN sees a load **before** the nullptr store and reckons that there is no HB except for some sleep and notices this. This message by TSAN is irrelevant (although technically correct). Therefore we want to silence TSAN here.

mpoeter: The wording is probably not ideal because "happens before" has a very specific meaning, but in "this read access here can occasionally happen before the write" it means something different (if we had such a "happens before" order we would not have a data race). The problem is that this read and the write that sets it to null are _unordered_. However, this does not matter in this case since we have the additional null check. So in this case we do not care about order, but just about atomicity. Nevertheless, @MBkkt is right that maybe we should adjust the wording.

KVS85:
- [x] https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/32032/

MBkkt (approved):

PR #20074, opened 2023-11-06
APM-822 Feature/apm 822 add metric for threads usage

Dronplane:
### Scope & Purpose
Added metric to expose number of  demanded threads.
Added option for setting default parallelism

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


Dronplane:
https://172.26.1.0/view/PR/job/arangodb-matrix-pr/32014/

MBkkt (commented):
I don't think it's needed and also I don't see any mention in APM-822.
So I think this shouldn't be done

MBkkt (changes requested):
I don't think it's needed and also I don't see any mention in APM-822.
So I think this shouldn't be done

arangod/IResearch/IResearchExecutionPool.cpp
```diff
@@ -0,0 +1,55 @@
...
+  uint64_t newval;
+  do {
+    newval = std::min(curr + n, _limit);
+  } while (!compare_exchange_weak(curr, newval));
+  auto add = newval - curr;
+  if (add > 0) {
+    // TODO: add a single call to thread_pool to change both
+    _pool.max_idle_delta(static_cast<int>(add));
+    _pool.max_threads_delta(static_cast<int>(add));
+  }
```
MBkkt, line 41: here can be added warn/info message about we don't have enough threads.

Dronplane:
> I don't think it's needed and also I don't see any mention in APM-822. So I think this shouldn't be done

I don't agree. We need to provide visibility to our customers. Using "htop"-like utility would not help as we are planning to use pool with pre-forked threads. So there would be no way to estimate if parallelism is used and if used  how actively.

MBkkt:
Well, I don't have better option.

But this approach also isn't option, because metrics especially with current value doesn't suite for this.

Honestly I'm not really sure why do we need provide such information at all, we don't make it for other concurrent activities, why for this?

Dronplane:
> because metrics especially with current value doesn't suite for this.
Why? 
> I'm not really sure why do we need provide such information at all, we don't make it for other concurrent activities, why for this?
You are wrong. There is  for example `arangodb_scheduler_num_awake_threads` metrics  and `arangodb_pregel_threads_number` and 'arangodb_scheduler_num_worker_threads' and so on

Dronplane:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/32049/

PR #20075, opened 2023-11-06
don't output 'undefined' - this will give NaN in the xml

dothebart:
### Scope & Purpose

we would output NaN into the xml file, this came from the value `undefined`.

- [x] :hankey: Bugfix
- [x] Backports:
  - [x] [3.11](https://github.com/arangodb/arangodb/pull/20079)
  - [x] [3.10](https://github.com/arangodb/arangodb/pull/20080)


jsteemann (approved):
LGTM

PR #20077, opened 2023-11-06
Alternative UPSERT syntax

jsteemann:
### Scope & Purpose

**Experimental**

* Support alternative syntax for UPSERT queries, so that arbitrary filter conditions can be used. The alternative syntax is:

      UPSERT FILTER <expression>
      UPDATE|REPLACE ...

  The key change here is that the UPSERT will produce a new variable `$CURRENT`
  with the to-be-found document. This variable can be used in the following 
  filter condition to apply arbitrary filter conditions and not just equality
  matches.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


MBkkt:
Is it same for execution?
```
for d in c
filter ...
upsert d ... in c
```
?

jsteemann:
> Is it same for execution?
> 
> ```
> for d in c
> filter ...
> upsert d ... in c
> ```

No, it is slightly different.
Right now our UPSERT syntax is
```
UPSERT <search-expression> INSERT ... UPDATE|REPLACE ...
```
where `<search-expression>` must be an object literal or a bind parameter.
That means it is currently possible to apply a "matching" filter condition for the UPSERT document lookup, e.g. `{ "value": 1, "foo": "bar" }`, but it is currently not possible to apply other filter conditions for the lookup, such as `<search doc>.value == 1 && <search doc>.foo != "bar" || IS_SAME_COLLECTION("foo", <search doc>._id)`.
This is exactly what this PR will add.
It is an experiment to check if the new syntax solves particular requirements with more complex lookup conditions that some users have. The outcome is yet unknown.

neunhoef (approved):
I think this is a sensible generalization. LGTM.

PR #20089, opened 2023-11-07
NonProducing case for JoinNode

maierlars:
### Scope & Purpose

If the result of one of the indexes in a join node is not used, the old implementation fell back to loading the whole document. This PR fixes this behavior.


js/common/modules/@arangodb/aql/explainer.js
```diff
@@ -2088,10 +2088,12 @@ function processQuery(query, explain, planIndex) {
...
         }
         let accessString = '';
-        if (!info.indexCoversProjections) {
+        if (!info.indexCoversProjections && info.producesOutput) {
           accessString += "index scan + document lookup";
-        } else {
+        } else if (info.indexCoversProjections) {
           accessString += "index scan";
+        } else {
+          accessString += "index scan only";
```
hkernbach, line 2096: This confuses me. The difference between "index scan" and "index scan only" is not clear to me :D

jsteemann (approved):
LGTM

js/common/modules/@arangodb/aql/explainer.js
maierlars: well, it index scan (and copy values out of the index) vs only scan through the index and don't care about the values you find...

hkernbach (approved):

PR #20093, opened 2023-11-08
[3.11] Improve locking in Async Logger (#19989)

mchacki:
### Scope & Purpose

*Right now the Logger holds a very large lock. The lock is held whenever the background thread flushes a queued log message, which is potentially slowed down by IO. If we try to do a new direct log entry in that period we need to wait for this lock to be released. With this PR the Lock is changed such that: We get a WriteLock whenever we modify the Loggers we use, but we only take a read-lock when we let the loggers print. This way we can have two or more threads using the loggers.
In turn the output message of the Logger, which does the IO operation, is guarded with a WriteLock.

In addition this PR adds an upper bound to the amount of messages in the queue. This is right now hard-coded and should not have effect under normal circumstances. If the logging is configured to use an extremely slow device we protect us here against accumulating too much memory in log messages. If the Limit is reached, we will start to directly log the message, which will have a performance impact. Also this may change the ordering of Log messages.
*

Note to reviewer: Please take extra care, this PR fiddles with Locks, and has potential to introduce undesired dead-locks.

This potentially fixes: https://arangodb.atlassian.net/browse/BTS-1273

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1273
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,14 +15,6 @@ v3.11.5 (XXXX-XX-XX)
   start to write log messages synchronously. This is to prevent the queue
   to grow indefinitely. The upper bound is configurable via the startup
   option `--log.max-queued-entries`. The default value is 10000.
-
-* Silence TSAN for shutdown for access to the SchedulerFeature::SCHEDULER
-  pointer using atomic references.
-
-* Fixed a race in controlled leader change which could lead to a situation 
-  in which a shard follower is dropped when the first write operation
-  happens. This fixes BTS-1647.
-
 * Fixed an unnecessary follower drop in controlled leader change, which will
   speed up leader changes. This fixes BTS-1658.
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,24 @@ v3.11.5 (XXXX-XX-XX)
   which a shard follower is dropped when the first write operation happens. This
   fixes BTS-1647.
 
+* Fixed BTS-1273: While queueing an async server log message, we could be
+  blocked by IO on the log-writer thread. This could slow down the main
+  path. In case the log-writer is configured to use slow device
+  (e.g. using syslog) this could have significant impact.
+
+* Introduced an upper bound of queued async log messages. If we would log
+  more messages then the background logger thread can actually process, we
+  start to write log messages synchronously. This is to prevent the queue
+  to grow indefinitely. The upper bound is configurable via the startup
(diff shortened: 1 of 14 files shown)


CHANGELOG
```diff
@@ -5,6 +5,24 @@ v3.11.5 (XXXX-XX-XX)
...
+  to grow indefinitely. The upper bound is configurable via the startup
+  option `--log.max-queued-entries`. The default value is 10000.
+
+* Silence TSAN for shutdown for access to the SchedulerFeature::SCHEDULER
+  pointer using atomic references.
+
+* Fixed a race in controlled leader change which could lead to a situation 
+  in which a shard follower is dropped when the first write operation
+  happens. This fixes BTS-1647.
+
```
jsteemann, line 25: These comments do not seem to belong to this PR.

jsteemann, line 25: These comments do not seem to belong to this PR.
```suggestion
```

lib/Logger/LoggerFeature.cpp
```diff
@@ -556,6 +556,21 @@ printed in the thread that triggered the log message. This is non-optimal for
...
 off to an extra logging thread, which asynchronously writes the log messages.)");
 
+  options
+      ->addOption(
+          "--log.max-queued-entries",
+          "Upper limit of log entries that are queued in background thread.",
+          new UInt32Parameter(&_maxQueuedLogMessages),
+          arangodb::options::makeDefaultFlags(
+              arangodb::options::Flags::Uncommon))
+      .setIntroducedIn(31200)
```
jsteemann, line 566: May be adjusted for 3.11.
```suggestion
      .setIntroducedIn(31105)
```

jsteemann (approved):
LGTM

mchacki:
Jenkins green:
https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/32037/

neunhoef (approved):
LGTM

PR #20099, opened 2023-11-08
Update jemalloc to the head of their dev branch of today.

markuspf:
### Scope & Purpose

When trying to build ArangoDB in Alpine 3.18 with clang 15.0.7, the build of jemalloc fails due to exception specifications in the jemalloc headers not coinciding with system headers.

Jemalloc 5.3.0 in the ArangoDB tree is the latest released version, but updating jemalloc to the latest development version fixes the problem. 

Unfortunately a small patch needs to be applied to jemalloc to prevent page table exhaustion through fragmentation.

One remaining question is whether jemalloc should become a git submodule like quite a few of the other 3rdParty libraries or not. This would require a fork of jemalloc at ArangoDB (because of the applied patch), and either committing the configure script to that fork as well or installing autoconf in the arango build container.

Another question is which way to choose for advertising the jemalloc version used. ArangoDB claims `5.3.0-0-g0`, before and after this patch, even though the code has changed quite a bit.


jsteemann (approved):
LGTM

dothebart (approved):

PR #20100, opened 2023-11-09
fix build files

nikhil-varma:


jsteemann (approved):
Tried it locally, and works for me.

PR #20112, opened 2023-11-09
Fix compilation with newer clang versions and musl

markuspf:
### Scope & Purpose

When compiling arangodb with clang 15 or 16 it complains

```
ISO C++11 does not allow conversion from string literal to 'char *' [-Werror,-Wwritable-strings]
    CODE* ptr = reinterpret_cast<CODE*>(facilitynames);ISO C++11 does not allow conversion from string literal to 'char *' [-Werror,-Wwritable-strings]
    CODE* ptr = reinterpret_cast<CODE*>(facilitynames);
```

This is a very reasonable complained caused by musl libc declaring `CODE` to contain a `char *`.

This PR introduces an interface function `find_syslog_facility_by_name` between the libc C code and C++ code which is compiled as C code.

### DIFF
--- a/lib/Logger/FindSyslogFacilityByName.h
+++ b/lib/Logger/FindSyslogFacilityByName.h
@@ -1,0 +1,26 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2023 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


lib/Logger/FindSyslogFacilityByName.h
```diff
@@ -0,0 +1,26 @@
...
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+extern "C" {
+int find_syslog_facility_by_name(const char* facility);
+};
```
jsteemann, line 26:
```suggestion
}
```

jsteemann (approved):
LGTM

lib/Logger/LogAppenderSyslog.cpp
```diff
@@ -59,16 +60,7 @@ LogAppenderSyslog::LogAppenderSyslog(std::string const& facility,
...
-
-    while (ptr->c_name != nullptr) {
-      if (strcmp(ptr->c_name, facility.c_str()) == 0) {
-        value = ptr->c_val;
-        break;
-      }
-
-      ++ptr;
-    }
+    value = find_syslog_facility_by_name(facility.c_str());
```
neunhoef, line 63: Could you not just have used a C-style cast here:
```
  value = (CODE*) (facilitynames.c_str());
```
? Then we would not need a separate file and an actual C compiler for it?
And then maybe a code comment to explain this?

neunhoef (approved):
LGTM, but see comment for a possible simplifaction.

PR #20116, opened 2023-11-10
Fix compilation with newer clang versions and musl (Alternative approach)

markuspf:
### Scope & Purpose

This is an alternative implementation of the fix in #20112 not requiring an extra file compiled as C and using suppressions.

### DIFF
--- a/lib/Logger/LogAppenderSyslog.cpp
+++ b/lib/Logger/LogAppenderSyslog.cpp
@@ -49,9 +49,9 @@ namespace {
 //
 // Extra fun is to be had because facilitynames is a macro, which is why this
 // code just accesses facilitynames by indexing
-// findSyslogFacilityByName returns LOG_LOCAL0 as fallback; the other choice
-// given the code below would be to ASSERT in case the facilityname cannot be
-// found.
+// findSyslogFacilityByName returns -1 as fallback; this is just dealt with
+// below by setting the facility to LOG_LOCAL0 as a last resort, as error
+// handling is not great with syslog
 #if defined(__clang__)
 #if __has_warning("-Wwritable-strings")
 #pragma clang diagnostic push
--- a/lib/Logger/LogAppenderSyslog.cpp
+++ b/lib/Logger/LogAppenderSyslog.cpp
@@ -99,9 +99,8 @@ LogAppenderSyslog::LogAppenderSyslog(std::string const& facility,
     value = findSyslogFacilityByName(name);
   }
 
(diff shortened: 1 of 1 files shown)


lib/Logger/LogAppenderSyslog.cpp
```diff
@@ -41,6 +41,32 @@ using namespace arangodb;
...
+#endif
+#endif
+auto findSyslogFacilityByName(std::string facility) -> int {
+  for (auto i = size_t{0}; i < LOG_NFACILITIES; ++i) {
+    if (strcmp(facilitynames[i].c_name, facility.c_str()) == 0) {
+      return facilitynames[i].c_val;
+    }
+    if (facilitynames[i].c_name == nullptr) {
+      return -1;
+    }
```
jsteemann, line 58: I guess these two `if`s should be exchanged.
```suggestion
    if (facilitynames[i].c_name == nullptr) {
      return -1;
    }
    if (strcmp(facilitynames[i].c_name, facility.c_str()) == 0) {
      return facilitynames[i].c_val;
    }
```

lib/Logger/LogAppenderSyslog.cpp
```diff
@@ -59,16 +96,13 @@ LogAppenderSyslog::LogAppenderSyslog(std::string const& facility,
...
-        break;
-      }
+    value = findSyslogFacilityByName(name);
+  }
 
-      ++ptr;
-    }
+  // Since the return value of ::openlog below is not checked or acted
+  // upon, try to be safe(er)
+  if (value < 0 or value > LOG_NFACILITIES) {
```
jsteemann, line 104: Should this be `>=`? Wondering if there is a valid value at `facilitynames[LOG_NFACILITIES]`.
```suggestion
  if (value < 0 or value >= LOG_NFACILITIES) {
```

lib/Logger/LogAppenderSyslog.cpp
```diff
@@ -59,16 +96,13 @@ LogAppenderSyslog::LogAppenderSyslog(std::string const& facility,
...
-      if (strcmp(ptr->c_name, facility.c_str()) == 0) {
-        value = ptr->c_val;
-        break;
-      }
+    value = findSyslogFacilityByName(name);
+  }
 
-      ++ptr;
-    }
+  // Since the return value of ::openlog below is not checked or acted
```
jsteemann, line 102: It is true, the return value of `openlog` is not checked here. Because `openlog` returns `void`.

PR #20119, opened 2023-11-13
[JOIN] More optimized projections

maierlars:
### Scope & Purpose

Optimize projections in join nodes more aggressively. With this PR the optimizer eliminates all projections in a join node that are no longer needed later in the query.


hkernbach (approved):

jsteemann (approved):
Code changes LGTM. Just wondering if any tests need to be adjusted to pass with these changes.

PR #20122, opened 2023-11-13
[JOIN] Refactor optimizer rule

maierlars:
### Scope & Purpose

This PR moves the join optimizer rule into a separate file and splits the function into multiple subfunctions to make it easier to add new futures.


arangod/Aql/OptimizerRulesJoin.cpp
```diff
@@ -0,0 +1,535 @@
...
+#define LOG_JOIN_OPTIMIZER_RULE LOG_DEVEL_IF(false)
+
+namespace {
+
+bool indexNodeQualifies(IndexNode const& indexNode) {
+  // not yet supported:
+  // - IndexIteratorOptions: sorted, ascending, evalFCalls, useCache,
+  // waitForSync, limit, lookahead
+  // - reverse iteration
+  // - support from GatherNodes
```
jsteemann, line 76: I think this is not true anymore
```suggestion
```

hkernbach (approved):
LGTM, apart from that TODO in the code: 
We should at least create a card for this on our board to not forget about it.

PR #20124, opened 2023-11-13
[EF-19] Join Node Estimations

maierlars:
### Scope & Purpose
The costs for multiple nodes are *added*, the join node multiplied the costs for indexes, which made it _very_ expensive and most of the time, i.e. indexes > 3, the optimizer would bail out and not create a join node.

### DIFF
--- a/arangod/Aql/OptimizerRulesJoin.cpp
+++ b/arangod/Aql/OptimizerRulesJoin.cpp
@@ -344,8 +344,8 @@ bool checkCandidatesEligible(
       auto* lhs = root->getMember(0);
       auto* rhs = root->getMember(1);
 
-      if (!joinConditionMatches(plan, lhs, rhs, c, candidates[i - 1]) &&
-          !joinConditionMatches(plan, lhs, rhs, candidates[i - 1], c)) {
+      if (!joinConditionMatches(plan, lhs, rhs, c, candidates[0]) &&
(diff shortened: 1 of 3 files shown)


arangod/Aql/OptimizerRulesJoin.cpp
```diff
@@ -344,8 +344,8 @@ bool checkCandidatesEligible(
       auto* lhs = root->getMember(0);
       auto* rhs = root->getMember(1);
 
-      if (!joinConditionMatches(plan, lhs, rhs, c, candidates[i - 1]) &&
-          !joinConditionMatches(plan, lhs, rhs, candidates[i - 1], c)) {
+      if (!joinConditionMatches(plan, lhs, rhs, c, candidates[0]) &&
+          !joinConditionMatches(plan, lhs, rhs, candidates[0], c)) {
```
jsteemann, line 348: This change will cause a behavior change:
Instead of comparing candidate[1] with candiate[0], and then candidate[2] with candidate[1], ... we will now compare candiate[1] with candidate[0], then candidate[2] with candidate[0] etc.
Not saying this is wrong in any way, but the behavior now can be different if there are more than 2 candidates.

maierlars: will be generalized in a follow up PR

jsteemann (approved):
LGTM

PR #20126, opened 2023-11-13
[Cinfra-822] Propagate CollectionGroup property changes

mchacki:
### Scope & Purpose

*This PR consists of 3 fixes in the same replication 2 area:
a) CollectionGroupSupervision will now update CollectionGroup in Plan whenever it's mutable attributes are changed.
b) Coordinator will now wait for ReplicatedLogs to converge when changing Properties of a Collection. This should reduce the delay between UserResponse and Change actually happening on the shards.
c) WaitForSync is now taken from CollectionProperties on Replication1, and from CollectionGroup in Replication2.
*

Bonus Fix: There was an overlap of hard-coded LogID and randomly picked logId in Supervision on test.
The test now checks a couple of ids if they are free and picks the first free one.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Cluster/ClusterCollectionMethods.cpp
```diff
@@ -878,11 +895,39 @@ ClusterCollectionMethods::createCollectionsOnCoordinator(
...
+    AgencyCallbackRegistry& callbackRegistry = ci.agencyCallbackRegistry();
+    auto currentGroup = pathCollectionGroupInCurrent(databaseName)
+                            ->group(std::to_string(col.groupID().id()))
+                            ->supervision()
+                            ->str(arangodb::cluster::paths::SkipComponents(1));
+    auto waitForSuccess = callbackRegistry.waitFor(
+        currentGroup, [versionToWaitFor](VPackSlice slice) {
+          if (slice.isNone()) {
+            // TODO: Should this actual set an "error"? It indicates
+            // that the collection is dropped if i am not mistaken
```
apetenchea, line 915: If the collection group is dropped, is this going to hang? Or is there a timeout somewhere?
I imagine that (even if the slice is not none), the collection group could still be dropped while this is ongoing. We'll have to make sure there's no trouble caused because of that, otherwise this might be a tricky hanger to investigate. Not necessarily in this PR, but that could be perhaps a low-prio ticket.

tests/Replication2/CollectionGroups/CollectionGroupsSupervision.cpp
```diff
@@ -405,3 +405,94 @@ TEST_F(CollectionGroupsSupervisionTest,
...
+    // Cannot do anything right now
+    ASSERT_TRUE(std::holds_alternative<NoActionRequired>(result))
+        << result.index();
+  }
+
+  // Now let us modify the waitForSync in target
+  group.target.attributes.mutableAttributes.waitForSync = false;
+
+  auto result = checkCollectionGroup(database, group, uniqid, health);
+  // Cannot do anything right now
```
apetenchea, line 487: Not an actual issue, but I am a bit confused by this comment. My first impression what that you used it to indicate that your expectation should be "NoActionRequired", as above.

arangod/Replication2/AgencyCollectionSpecification.h
```diff
@@ -57,6 +57,8 @@ struct CollectionGroup {
       std::size_t writeConcern{};
       std::size_t replicationFactor{};
       bool waitForSync{};
+
+      bool operator==(MutableAttributes const& other) const noexcept;
```
apetenchea, line 61: NIT
By the looks of its implementation, this could have been defaulted.

arangod/Cluster/ClusterCollectionMethods.cpp
mchacki: Absolutely. There are already other cases with exactly the same comment.
So this will be addressed in general.

tests/Replication2/CollectionGroups/CollectionGroupsSupervision.cpp
mchacki: Copy paste error...
I copied the lines above for a head-start, changed the action, but kept the comment...

apetenchea (approved):

apetenchea (approved):

PR #20128, opened 2023-11-13
Transparent compression of requests and responses

jsteemann:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1410

* Added options `--compress-transfer` and `--compress-request-threshold` to all client tools. These options can be used to enable transparent compression of the data that is sent between the client tools and the ArangoDB server.

  If the option `--compress-transfer` is set to `true`, the client-tools will add an extra HTTP header `Accept-Encoding: deflate` to all requests they make to the server. This allows the server to compress its responses before sending them back to the client tools.

  If the option is set to `true`, the client will also transparently compress their own requests to the server if the size of the request body (in bytes) is at least the value of the startup option `--compress-request-treshold`. The default value for this option is `0`, which disables compression of the request bodies in the client tools. To opt in to sending compressed data, the option must be set to a value greater than zero. Request body compression will be performed using the deflate compression algorithm. The client tools will also add a `Content-Encoding: deflate` header to the request when the request body was compressed. As compression will use CPU cycles, it should be activated only in case the network communication between the client tools and the server is slow and there is enough CPU capacity left for the extra compression/decompression work. Furthermore, requests should only be compressed when they exceed a certain minimum size, e.g. 250 bytes. Too small requests are typically not compressible very well.

  Request compression is only supported for HTTP/1.1 and HTTP/2 requests, but not when using the VST protocol.

* Added startup option `--http.compress-response-threshold` to the ArangoDB server. The value of this specifies the threshold value (in bytes) from which on response bodies will be sent out compressed by the server.

  The default value for this option is `0`, which disables sending out compressed response bodies. To enable compression, the option should be set to a value greater than 0. The selected value should be large enough to justify the compression overhead. Too small response bodies are typically not well compressible.

  Regardless of the value of this option, response body compression will only happen in case the client advertised that it expects a compressed response body by sending an `Accept-Encoding: gzip` or `Accept-Encoding: deflate` header with its request. If that header is missing, no response compression will be performed by the server. As compression will use CPU cycles, it should be activated only in case the network communication between the server and clients is slow and there is enough CPU capacity left for the extra compression/decompression work. work. Furthermore, responses should only be compressed when they exceed a certain minimum size, e.g. 250 bytes. Response compression is only supported for HTTP/1.1 and HTTP/2 responses, but not when using the VST protocol.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1410
- [ ] Design document:


client-tools/Shell/ClientFeature.cpp
```diff
@@ -198,6 +198,30 @@ arangosh without connecting to a server.)");
...
+          new BooleanParameter(&_compressTransfer))
+      .setIntroducedIn(31200)
+      .setLongDescription(R"(This option enables transport compression for data
+received by an ArangoDB server.)");
+
+  options
+      ->addOption("--compress-request-threshold",
+                  "The HTTP request body size from which on requests are "
+                  "transparently compressed when sending them to the server.",
+                  new UInt64Parameter(&_compressRequestThreshold))
```
MBkkt, line 215: Do we really need options for threshold?
At least we can combine this two.
Threshold == max (default) => no compression
Threshold == 0 => compression enabled, threshold chosen automatically
Any other threshold value => compression enabled, and manually chosen threshold

MBkkt (approved):

client-tools/Shell/ClientFeature.cpp
jsteemann: IMHO these are two different things:
1. whether or not client compresses requests based on the size of the uncompressed body
2. whether or not to send the `Accept-Encoding: deflate` along with the request
Both things are orthogonal.
If a request is eligible to be compressed, the server may still respond with an uncompressed response.  And if the request is not eligible to be compressed, the server can still send a compressed response if it is considered useful.
That's why I didn't combine the two options in the first place.
We can combine them into a single option now, but what would be the consequence for when the client should send the `Accept-Encoding: deflate` header with its requests?
1. always send it
2. never send it
3. only send it along requests for which the request body was compressed
This is unclear and thus I would favor a separate config option.

PR #20129, opened 2023-11-14
Make the actor implementation a bit more generic

mpoeter:
### Scope & Purpose

Follow up PR of #20111 in preparation of the parallel follower implementation.

Some generic messages previously relied on the `DistributedActorPID`, but other runtime implementations may use different PID types, so these messages are now templatized on the PID type. Similarly the `ActorBase` has been templatized for the same reason, and a new interface has been extracted so that the `IScheduler` has a minimal dependency (interface segregation principle).

### DIFF
--- a/lib/Actor/include/Actor/Message.h
+++ b/lib/Actor/include/Actor/Message.h
@@ -51,22 +51,22 @@ auto inspect(Inspector& f, MessagePayload<Payload>& x) {
 
 namespace message {
 
-template<class PID>
+template<typename PID>
 struct UnknownMessage {
   PID sender;
   PID receiver;
 };
-template<typename Inspector, class PID>
+template<typename Inspector, typename PID>
 auto inspect(Inspector& f, UnknownMessage<PID>& x) {
   return f.object(x).fields(f.field("sender", x.sender),
                             f.field("receiver", x.receiver));
 }
 
-template<class PID>
+template<typename PID>
 struct ActorNotFound {
   PID actor;
 };
-template<typename Inspector, class PID>
+template<typename Inspector, typename PID>
 auto inspect(Inspector& f, ActorNotFound<PID>& x) {
   return f.object(x).fields(f.field("actor", x.actor));
 }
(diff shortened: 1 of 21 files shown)


lib/Actor/include/Actor/Message.h
```diff
@@ -52,21 +51,23 @@ auto inspect(Inspector& f, MessagePayload<Payload>& x) {
...
+template<class PID>
 struct UnknownMessage {
-  DistributedActorPID sender;
-  DistributedActorPID receiver;
+  PID sender;
+  PID receiver;
 };
-template<typename Inspector>
-auto inspect(Inspector& f, UnknownMessage& x) {
+template<typename Inspector, class PID>
```
markuspf, line 59:
```suggestion
template<typename Inspector, typename PID>
```

lib/Actor/include/Actor/Message.h
```diff
@@ -52,21 +51,23 @@ auto inspect(Inspector& f, MessagePayload<Payload>& x) {
...
+auto inspect(Inspector& f, UnknownMessage<PID>& x) {
   return f.object(x).fields(f.field("sender", x.sender),
                             f.field("receiver", x.receiver));
 }
 
+template<class PID>
```
markuspf, line 65:
```suggestion
template<typename PID>
```

lib/Actor/include/Actor/Message.h
```diff
@@ -52,21 +51,23 @@ auto inspect(Inspector& f, MessagePayload<Payload>& x) {
...
 struct ActorNotFound {
-  DistributedActorPID actor;
+  PID actor;
 };
-template<typename Inspector>
-auto inspect(Inspector& f, ActorNotFound& x) {
+template<typename Inspector, class PID>
```
markuspf, line 69:
```suggestion
template<typename Inspector, typename PID>
```

lib/Actor/include/Actor/Message.h
```diff
@@ -78,14 +79,17 @@ auto inspect(Inspector& f, NetworkError& x) {
   return f.object(x).fields(f.field("server", x.message));
 }
 
-struct ActorError : std::variant<UnknownMessage, ActorNotFound, NetworkError> {
-  using std::variant<UnknownMessage, ActorNotFound, NetworkError>::variant;
+template<class PID>
```
markuspf, line 82:
```suggestion
template<typename PID>
```

markuspf (changes requested):
Nitpicks mostly about the fact that I like `typename` more than `class`, and certainly like consistency in that regard

I was wondering whether one can get away without templating the generic error messages (actor not found and friends), because underlying the message payload is the control information about sender and receiver already; then again that would somehow introduce implied knowledge (for example that the sender of the "actor not found" message is the invalid actor.

There is certainly a design discussion to be had about what should be done using templates/concepts and what with interfaces; this is outside the scope of this PR though.

markuspf (approved):
:+1: thanks for indulging in my nitpicking.

PR #20133, opened 2023-11-14
Equal Attribute Access Replacement

maierlars:
### Scope & Purpose
ENTERPRISE: https://github.com/arangodb/enterprise/pull/1411 (Fixing tests only)
Assume a query contains a filter statement like
```
FILTER doc1.x == doc2.y
```
after that statement, the optimizer can assume that `doc1.x` and `doc2.x` always refer to the same value. This PR introduces a new optimizer rule that generates a new temporary variable with the value of `doc1.x` and replaces all references of `doc1.x` and `doc2.y` with that variable.

This allows a lot of new optimizations further down the pipeline. For example, we can remove many projections from indexes and make the indexes scan only. Also the JoinNode now detects more cases, because the transitivity of equality is handled transparently.

TODO:
- [x] Only introduce one temporary variable for the equivalence class of attribute accesses. Currently whenever we introduce a new equivalence, we add a new variable. The optimizer removes all of the variable later anyways.

EXAMPLES:
Query before the optimization
```
Query String (154 chars, cacheable: true):
   for doc2 in c1
     for doc1 in c2
       for doc3 in c2
       filter doc2.x == doc1.x
       filter doc3.x == doc2.x
       return [doc1.x, doc2.x, doc3.x]

Execution plan:
 13   IndexNode           ✓      0     - FOR doc1 IN c2   /* persistent index scan, index only (projections: `x`) */    LET #13 = doc1.`x`   
 12   IndexNode           ✓      0       - FOR doc2 IN c1   /* persistent index scan, index only (projections: `x`) */    LET #11 = doc2.`x`   
 11   IndexNode           ✓      0         - FOR doc3 IN c2   /* persistent index scan, index only (projections: `x`) */    LET #9 = doc3.`x`   
  9   CalculationNode     ✓      0           - LET #7 = [ #13, #11, #9 ]   /* simple expression */
 10   ReturnNode                 0           - RETURN #7
```
Notice how the JoinRule failed to detect that joining is possible. This is because the conditions are transitive, i.e. `a == b && b == c`, but the rule expects `a == b && a == c`.

```
Query String (154 chars, cacheable: true):
   for doc2 in c1
     for doc1 in c2
       for doc3 in c2
       filter doc2.x == doc1.x
       filter doc3.x == doc2.x
       return [doc1.x, doc2.x, doc3.x]

Execution plan:
 16   JoinNode            ✓      0     - JOIN
 16   JoinNode                   0       - FOR doc1 IN c2   LET #17 = doc1.`x`   /* index scan (projections: `x`) */
 16   JoinNode                   0       - FOR doc2 IN c1   /* index scan only */
 16   JoinNode                   0       - FOR doc3 IN c2   /* index scan only */
  9   CalculationNode     ✓      0     - LET #7 = [ #17, #17, #17 ]   /* simple expression */
 10   ReturnNode                 0     - RETURN #7
 ```
 A join was detected and all but one index don't even produce an output.

### CODE REVIEW

File: arangod/Aql/OptimizerRuleReplaceEqualAttributeAccess.cpp [hunk 1]
```diff
@@ -0,0 +1,308 @@
...
+#include "Basics/ScopeGuard.h"
+#include "Basics/StaticStrings.h"
+#include "Containers/FlatHashSet.h"
+#include "Containers/NodeHashMap.h"
+#include "Containers/HashSet.h"
+#include "Containers/SmallUnorderedMap.h"
+#include "Containers/SmallVector.h"
+#include "Geo/GeoParams.h"
+#include "Graph/ShortestPathOptions.h"
+#include "Graph/TraverserOptions.h"
```
jsteemann: nit: many of these includes seem to be unnecessary in this file, so they could be cleaned up.

arangod/Aql/OptimizerRuleReplaceEqualAttributeAccess.cpp
```diff
@@ -0,0 +1,293 @@
...
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#include "ApplicationFeatures/ApplicationServer.h"
+#include "Aql/AstHelper.h"
```
jsteemann, line 23: nit: it seems that at least some of the includes are not necessary to compile this TU, so they could be removed here.

arangod/Aql/OptimizerRuleReplaceEqualAttributeAccess.cpp
```diff
@@ -0,0 +1,293 @@
...
+using EN = arangodb::aql::ExecutionNode;
+
+#define LOG_RULE LOG_DEVEL_IF(false)
+
+namespace {
+
+/*
+ * This function calls the callback for all binary equal operators found
+ * in the given expression. It will descend into nary-ors only if there is a
+ * single member. It descents into all nary-and operators.
```
jsteemann, line 59:
```suggestion
 * single member. It descends into all nary-and operators.
```

arangod/Aql/OptimizerRuleReplaceEqualAttributeAccess.cpp
```diff
@@ -0,0 +1,293 @@
...
+      }
+      modified = true;
+    }
+
+    // check if this is a filter node
+    if (node->getType() != EN::FILTER) {
+      continue;
+    }
+
+    auto* fn = static_cast<FilterNode*>(node);
```
jsteemann, line 233:
```suggestion
    auto* fn = ExecutionNode::castTo<FilterNode*>(node);
```

arangod/Aql/OptimizerRuleReplaceEqualAttributeAccess.cpp
```diff
@@ -0,0 +1,293 @@
...
+    auto* n = plan->getVarSetBy(fn->inVariable()->id);
+    TRI_ASSERT(n->getType() == EN::CALCULATION);
+    auto* cn = static_cast<CalculationNode*>(n);
```
jsteemann, line 236:
```suggestion
    auto* cn = ExecutionNode::castTo<CalculationNode*>(n);
```

jsteemann (approved):
LGTM

PR #20134, opened 2023-11-14
Added "readOwnWrites" option for AQL UPSERT operations.

jsteemann:
### Scope & Purpose

Added "readOwnWrites" option for AQL UPSERT operations.

  If set to `false` (which is the default), an UPSERT operation will process
  its inputs one by one. That way an UPSERT can observe its own writes and can
  handle modifying the same target document multiple times in the same query.

  When the "readOwnWrite" option is set to `false`, an UPSERT can process its
  input in batches (a batch is normally 1000 inputs), so it can execute a lot
  faster. However, when using batches, an UPSERT cannot observe its own writes.
  So the "readOwnWrite" option should only be set to `false` if it can be
  guaranteed that the input to the UPSERT will lead to disjoint documents
  being inserted/updated/replaced.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,19 @@
 devel
 -----
 
+* Added "readOwnWrites" option for AQL UPSERT operations.
+  
+  If set to `false` (which is the default), an UPSERT operation will process
+  its inputs one by one. That way an UPSERT can observe its own writes and can
+  handle modifying the same target document multiple times in the same query.
+
+  When the "readOwnWrite" option is set to `false`, an UPSERT can process its
+  input in batches (a batch is normally 1000 inputs), so it can execute a lot
+  faster. However, when using batches, an UPSERT cannot observe its own writes.
+  So the "readOwnWrite" option should only be set to `false` if it can be
+  guaranteed that the input to the UPSERT will lead to disjoint documents 
+  being inserted/updated/replaced.
+
 * Support alternative syntax for UPSERT queries, so that arbitrary filter
   conditions can be used. The alternative syntax is:
 
--- a/arangod/Aql/ModificationExecutorInfos.cpp
+++ b/arangod/Aql/ModificationExecutorInfos.cpp
@@ -52,6 +52,7 @@ ModificationExecutorInfos::ModificationExecutorInfos(
       _doCount(doCount),
       _isReplace(isReplace),
       _ignoreDocumentNotFound(ignoreDocumentNotFound),
+      _useBatching(true),  // used only by upsert
       _input1RegisterId(input1RegisterId),
       _input2RegisterId(input2RegisterId),
       _input3RegisterId(input3RegisterId),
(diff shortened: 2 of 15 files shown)


CHANGELOG
```diff
@@ -1,6 +1,19 @@
 devel
 -----
 
+* Added "readOwnWrites" option for AQL UPSERT operations.
+  
+  If set to `false` (which is the default), an UPSERT operation will process
```
markuspf, line 6:
```suggestion
  If set to `true` (which is the default), an UPSERT operation will process
```
?

arangod/Aql/ModificationExecutorInfos.cpp
```diff
@@ -52,6 +52,7 @@ ModificationExecutorInfos::ModificationExecutorInfos(
       _doCount(doCount),
       _isReplace(isReplace),
       _ignoreDocumentNotFound(ignoreDocumentNotFound),
+      _useBatching(true),  // used only by upsert
```
markuspf, line 55: Why is this necessary, do the other modification executors not batch as well?

arangod/Aql/ModificationNodes.h
```diff
@@ -407,19 +407,7 @@ class UpsertNode : public ModificationNode {
...
-                         outVariableNew),
-        _inDocVariable(inDocVariable),
-        _insertVariable(insertVariable),
-        _updateVariable(updateVariable),
-        _isReplace(isReplace) {
-    TRI_ASSERT(_inDocVariable != nullptr);
-    TRI_ASSERT(_insertVariable != nullptr);
-    TRI_ASSERT(_updateVariable != nullptr);
-
-    TRI_ASSERT(_outVariableOld == nullptr);
```
markuspf, line 421: Why are these removed? were they not used anymore?

markuspf: Ah I see you moved the constructor.

arangod/Aql/ModificationExecutorInfos.cpp
jsteemann: True, all other modification executors use batching.

The actual batch size however is only set in the ctors of `SimpleModifier` and `UpsertModifier`.
Probably we should remove the extra `_useBatching` flag and instead put the batch size into the `ModificationExecutorInfos`?

jsteemann: Changed.

markuspf: If I recall correctly, the only modifier that is *not* simple is the upsert modifier, and previouslyl the upsert modifier was forced to use a batchsize of 1; so it should be correct to have the batch size in the ModificationExecutorInfos.

jsteemann: I think so too. The change is implemented in the latest version of this PR.

markuspf (approved):
I am always nervous about UPSERT and its semantics, but seeing as we know that they are not very well-defined even before this change, I will not stand in its way ;)

PR #20136, opened 2023-11-15
Automatically remove finished actors

mpoeter:
### Scope & Purpose

Previously the runtime required the user to periodically call `garbageCollect` to clean out actors that are finished and idle.
However, the actor itself knows when it is finished and idle, so this change allows the actor to remove itself at that point. This means the `garbageCollect` function is no longer necessary.

### DIFF
--- a/arangod/Pregel/PregelFeature.cpp
+++ b/arangod/Pregel/PregelFeature.cpp
@@ -660,9 +660,6 @@ std::shared_ptr<Conductor> PregelFeature::conductor(
 }
 
 void PregelFeature::garbageCollectActors() {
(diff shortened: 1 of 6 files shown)


markuspf (commented):
This PR renames the `ActorRuntimeTest` to `DistributedRuntimeTest` and does not mention it in the description.

Also this rename has nothing to do with the removal of garbage collection.

markuspf (approved):

mpoeter:
True.. this was done so that the name of the test suite is consistent with the name of the class under test. I completely forgot that I made this change too before creating the PR.

PR #20140, opened 2023-11-15
Refactor: make spawn return an ActorPID

mpoeter:
### Scope & Purpose

Make spawn return a complete ActorPID instead of just the ID,


markuspf (approved):
LGTM

PR #20149, opened 2023-11-17
[BTS-1686] Get rid of data race when killing aql query

jvolmer:
Data race:
When one of the participating DBServers fails before an AQL query is executed on it, the query is killed on all DBServers and the coordinator (this happens inside `ExecutionEngine::buildEngines()`). The kill command starts a new thread and calls `Query::cleanupPlanAndEngine(...)`. Another thread on the coordinator is already working on the Query and when it also calls `Query::cleanupPlanAndEngine(...)` there is a data race on the `Query` class members `_resultCode` and `_endTime` (e.g. the TSAN noticed this data race because the coordinator called `Query::executeV8(...)`, ran into an exception there and then called `Query::cleanupPlanAndEngine(...)`). 
(Nothing worse is happening because the atomic `Query` member `_shutdownState` guards the following execution of `Query::cleanupTrxAndEngines`.

Fix:
`Query::_resultCode` is now an atomic and `_endTime` and `_resultCode` are only updated if `_resultCode` is empty. The race condition is still there - we don't know which thread wins and sets both `_endTime` and `_resultCode` but for a killed query this is not that relevant.

Reproduced race via:
```
diff --git a/arangod/Aql/ExecutionEngine.cpp b/arangod/Aql/ExecutionEngine.cpp
index eae54f8084b..4b51df35a7a 100644
--- a/arangod/Aql/ExecutionEngine.cpp
+++ b/arangod/Aql/ExecutionEngine.cpp
@@ -22,7 +22,9 @@
 
 #include "ExecutionEngine.h"
 
+#include "Network/NetworkFeature.h"
 #include "Aql/BlocksWithClients.h"
 #include "Aql/Collection.h"

@@ -571,8 +574,12 @@ struct DistributedQueryInstanciator final
         std::string comment = std::string("AQL query from coordinator ") +
                               ServerState::instance()->getId();
 
+        auto stillInThread = std::make_shared<bool>(true);
+
         std::function<void(void)> f = [srvr = server, id = _query.id(),
-                                       vn = _query.vocbase().name(), &df]() {
+                                       vn = _query.vocbase().name(), &df,
+                                       stillInThread]() {
+          TRI_ASSERT(!*stillInThread);
           LOG_TOPIC("d2554", INFO, Logger::QUERIES)
               << "killing query " << id << " because participating DB server "
               << srvr << " is unavailable";
@@ -587,6 +594,17 @@ struct DistributedQueryInstanciator final
 
         engine->rebootTrackers().emplace_back(ci.rebootTracker().callMeOnChange(
             {server, rebootId}, std::move(f), std::move(comment)));
+
+        NetworkFeature& nf =
+            _query.vocbase().server().getFeature<NetworkFeature>();
+        auto pool = nf.pool();
+
+        std::ignore = arangodb::network::sendRequest(pool, "server:" + server,
+                                                     fuerte::RestVerb::Delete,
+                                                     "_admin/shutdown")
+                          .get();
+        std::this_thread::sleep_for(std::chrono::seconds(30));
+        *stillInThread = false;
       }
     }
```
This will result in a failed coordinator due to the added TRI_ASSERT.

### DIFF
--- a/arangod/Aql/Query.cpp
+++ b/arangod/Aql/Query.cpp
@@ -394,13 +394,25 @@ void Query::prepareQuery() {
 
     enterState(QueryExecutionState::ValueType::EXECUTION);
   } catch (Exception const& ex) {
-    _resultCode.store(ex.code());
+    auto emptyResultCode = std::optional<ErrorCode>(std::nullopt);
+    if (_resultCode.compare_exchange_strong(emptyResultCode, ex.code())) {
+      ensureExecutionTime();
+    }
     throw;
   } catch (std::bad_alloc const&) {
+    auto emptyResultCode = std::optional<ErrorCode>(std::nullopt);
+    if (_resultCode.compare_exchange_strong(emptyResultCode,
+                                            TRI_ERROR_OUT_OF_MEMORY)) {
+      ensureExecutionTime();
+    }
     _resultCode.store(TRI_ERROR_OUT_OF_MEMORY);
     throw;
   } catch (...) {
-    _resultCode.store(TRI_ERROR_INTERNAL);
+    auto emptyResultCode = std::optional<ErrorCode>(std::nullopt);
+    if (_resultCode.compare_exchange_strong(emptyResultCode,
+                                            TRI_ERROR_INTERNAL)) {
+      ensureExecutionTime();
+    }
     throw;
   }
 }
--- a/arangod/Aql/Query.cpp
+++ b/arangod/Aql/Query.cpp
@@ -394,13 +394,13 @@ void Query::prepareQuery() {
 
     enterState(QueryExecutionState::ValueType::EXECUTION);
   } catch (Exception const& ex) {
-    _resultCode = ex.code();
+    _resultCode.store(ex.code());
     throw;
   } catch (std::bad_alloc const&) {
-    _resultCode = TRI_ERROR_OUT_OF_MEMORY;
+    _resultCode.store(TRI_ERROR_OUT_OF_MEMORY);
     throw;
   } catch (...) {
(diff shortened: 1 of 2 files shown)


arangod/Aql/Query.cpp
```diff
@@ -1462,7 +1462,7 @@ bool Query::canUseQueryCache() const {
 
 ErrorCode Query::resultCode() const noexcept {
   // never return negative value from here
-  return _resultCode.value_or(TRI_ERROR_NO_ERROR);
+  return _resultCode.load().value_or(TRI_ERROR_NO_ERROR);
```
jsteemann, line 1465: Is it ok to use memory order `std::memory_order_relaxed` for this load?

arangod/Aql/Query.cpp
```diff
@@ -1482,11 +1482,10 @@ void Query::enterState(QueryExecutionState::ValueType state) {
 
 /// @brief cleanup plan and engine for current query
 ExecutionState Query::cleanupPlanAndEngine(ErrorCode errorCode, bool sync) {
-  ensureExecutionTime();
```
jsteemann, line 1485: It seems that removing the unconditional update of the execution time causes assertion failures in the tests.

arangod/Aql/Query.cpp
jvolmer: I'm really unsure about memory ordering. As it is now it is at least correct, do you think that using `std::memory_order_relaxed` is really relevant here?

jsteemann: I don't think it is relevant here, because there is just one call site.
So feel free to ignore it.

PR #20150, opened 2023-11-17
Fix propagate constants rule

maierlars:
### Scope & Purpose
This PR fixes a bug in the propagate constants rule, that would propagate constant values up and not just down the query. This could result in bugs, like in this query:
```
Query String (143 chars, cacheable: true):
 for doc1 in c1
   for doc2 in c2
     filter doc1.x == doc2.x
     sort doc1.x, doc2.x
     limit 10
     filter doc1.x == 5
     return [doc1, doc2]

Execution plan:
 Id   NodeType                  Par   Est.   Comment
  1   SingletonNode                      1   * ROOT
  2   EnumerateCollectionNode            0     - FOR doc1 IN c1   /* full collection scan  */
  6   CalculationNode                    0       - LET #4 = doc1.`x`   /* attribute expression */   /* collections used: doc1 : c1 */
 14   IndexNode                          0       - FOR doc2 IN c2 LET #6 = doc2.`x`   /* persistent index scan, index scan + document lookup */    /* with late materialization */
  8   SortNode                           0         - SORT #4 ASC, #6 ASC   /* sorting strategy: standard */
  9   LimitNode                          0         - LIMIT 0, 10
 15   MaterializeNode                    0         - MATERIALIZE doc2
 10   CalculationNode             ✓      0         - LET #8 = (doc1.`x` == 5)   /* simple expression */   /* collections used: doc1 : c1 */
 11   FilterNode                  ✓      0         - FILTER #8
 12   CalculationNode             ✓      0         - LET #10 = [ doc1, doc2 ]   /* simple expression */   /* collections used: doc1 : c1, doc2 : c2 */
 13   ReturnNode                         0         - RETURN #10

Indexes used:
 By   Name                      Type         Collection   Unique   Sparse   Cache   Selectivity   Fields    Stored values   Ranges
 14   idx_1782716678583353344   persistent   c2           false    false    false      100.00 %   [ `x` ]   [  ]            (5 == doc2.`x`)
```
Notice the index usage. The query asks for 
> which of the smallest 10 have x equal to 5.

but the optimizer translates this to
> give me the 10 smallest with x equal to 5

using the index scan with `doc2.x == 5`.

Other things done in this PR:
- I moved the optimizer rule into a separate file.
- `getCollection` was moved into the optimizer utils file.


arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,302 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+#include "OptimizerRules.h"
+#include "Aql/Expression.h"
+#include "Aql/Collection.h"
+#include "Aql/Optimizer.h"
+#include "VocBase/LogicalCollection.h"
+#include "Cluster/ServerState.h"
+#include "Logger/LogMacros.h"
+#include "OptimizerUtils.h"
```
jsteemann, line 32:
```suggestion
#include "Aql/OptimizerUtils.h"
#include "Cluster/ServerState.h"
#include "Logger/LogMacros.h"
#include "VocBase/LogicalCollection.h"
```

arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,302 @@
...
+class PropagateConstantAttributesHelper {
+ public:
+  explicit PropagateConstantAttributesHelper(ExecutionPlan* plan)
+      : _plan(plan), _modified(false) {}
+
+  bool modified() const { return _modified; }
+
+  /// @brief inspects a plan and propagates constant values in expressions
+  void propagateConstants() {
+    LOG_DEVEL << "PROPAGATE CONSTANTS";
```
jsteemann, line 49:
```suggestion
```

arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  /// @brief inspects an expression (recursively) and notes constant attribute
+  /// values so they can be propagated later
+  void collectConstantAttributes(AstNode* node) {
+    if (node == nullptr) {
+      return;
+    }
+
+    LOG_DEVEL << node->toString();
+
```
jsteemann, line 97: I guess all these `LOG_DEVEL` should be removed, or changed to some `LOG_DEVEL_RULE` (as in other optimizer rules)

arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,302 @@
...
+      _constants;
+  bool _modified;
+};
+}  // namespace
+
+/// @brief propagate constant attributes in FILTERs
+void arangodb::aql::propagateConstantAttributesRule(
+    Optimizer* opt, std::unique_ptr<ExecutionPlan> plan,
+    OptimizerRule const& rule) {
+  plan->show();
```
jsteemann, line 297: Was potentially here for debugging only
```suggestion
```

arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,344 @@
...
+  LOG_RULE << "VAR = " << access;
+  auto it = replacements.find(access);
+
+  if (it == nullptr) {
+    replacements.set(access, Replacement{currentLimitCount, value});
+    return;
+  }
+}
+
+void collectConstantAttributes(AstNode* node, std::size_t currentLimitCount,
```
jsteemann, line 117: I hope that we can make `AstNode* node` const here, so we can avoid the const_cast further down below at the call site of this function.

arangod/Aql/OptimizerRulePropagateConstantAttributes.cpp
```diff
@@ -0,0 +1,344 @@
...
+      continue;
+    }
+
+    auto fn = ExecutionNode::castTo<FilterNode const*>(node);
+    auto setter = plan.getVarSetBy(fn->inVariable()->id);
+    if (setter != nullptr && setter->getType() == EN::CALCULATION) {
+      auto cn = ExecutionNode::castTo<CalculationNode*>(setter);
+      auto expression = cn->expression();
+
+      collectConstantAttributes(const_cast<AstNode*>(expression->node()),
```
jsteemann, line 296: Do we really have to const_cast away the constness of the node here? I would hope that `collectConstantAttributes` does not modify the node.

jsteemann (approved):
LGTM

hkernbach (approved):

PR #20156, opened 2023-11-17
[CINFRA-861] Implementing undo for index creation

apetenchea:
### Scope & Purpose

For most of the resource creation and resource removal operations, the leader first replicates the operation, and only then applies it locally. The rationale behind this is that unless an operations is committed into the log, we can't guarantee that it's effects will be visible after a fail-over. Then, applying the operation locally **should** succeed (some corner cases which can safely be ignored are documented in the error handler class).

However, for index creation, there are several points of failure, which may be avoided by doing the process in reverse: first trying to apply the entry locally, and only then replicating it. This is because the coordinator does not perform all the checks, but leaves some for the dbservers. For example, there's the unique constrain violated, bad parameter (only one TTL index allowed), or "shard not found".
By applying the index first, and then replicating, we ensure the index creation operation is valid. Unfortunately, it is possible that between the local application and replication, the leader resigns. In that case, we undo the operation, by dropping the index. This doesn't come without any risks, see [CINFRA-865](https://arangodb.atlassian.net/browse/CINFRA-865) for details.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

[CINFRA-865]: https://arangodb.atlassian.net/browse/CINFRA-865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mchacki (approved):
LGTM.
Especially like the comments on WHY things are this way :+1:

PR #20159, opened 2023-11-20
[PROTOTYPE] Filter Conditions on Paths

markuspf:
### Scope & Purpose

This is a prototype PR, not to be merged as is.

The goal of this work is to detect conditions that have to hold for all vertices/edges on paths, for instance as in the following query

```
FOR p IN 1..5 OUTBOUND K_PATHS "airports/LAX" TO "airports/SFO" flights
  FILTER p.vertices[* RETURN CURRENT.colour == "green"] ALL == true
  FILTER p.edges[* RETURN CURRENT.forbidden] ALL == false
  RETURN p
```

and optimize path search by abandoning all searches that would involve vertices/edges that don't satisfy this global condition.

This involves multiple things to happen:
 * The optimizer rule has to detect `FILTER` statements that access the path variable of a path enumeration using an expansion
 * The optimizer rule has to extract the expansion's map expression and rewrite it to apply to a single vertex
 * The optimizer rule has to enter the expression into the path enumeration
 * The path enumeration has to support adding global filters for vertices and edges
 * The path enumeration has to use these global filters.

At the moment all these steps happen in this one PR in various levels of completion.


neunhoef:
Infeasible to finish for now.

PR #20162, opened 2023-11-20
Use the async future lock

maierlars:
### Scope & Purpose

Final PR to fix APM-19.


arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1712,8 +1699,39 @@ Result RocksDBMetaCollection::applyUpdatesForTransaction(
   });
 }
 
+#define USE_FUTURE_MUTEX 1
```
neunhoef, line 1702: Do we really still need this switch?

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1722,58 +1740,47 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
...
-             "blockages!";
-    }
-  }
-
-  timeout_us -= detachThreshold;
-
-  if (mode == AccessMode::Type::WRITE) {
-    gotLock = _exclusiveLock.tryLockWriteFor(timeout_us);
+#else
+    gotLock = _exclusiveLockOld.tryLockWriteFor(
```
neunhoef, line 1756: If we need this case, we should go back to the thread detaching. If not, the case should be removed.

lib/Futures/include/Futures/coro-helper.h
```diff
@@ -34,6 +34,7 @@ namespace std_coro = std;
 #include "Basics/Result.h"
 #include "Promise.h"
 #include "Try.h"
+#include "Utils/ExecContext.h"
```
neunhoef, line 38: This whole class needs an explanation in the source code, such that future developers (including our own future selves) can understand what it is for.

neunhoef (changes requested):
See comments, these are only minor changes, the functionality as such should work.

lib/Futures/include/Futures/coro-helper.h
neunhoef: Done.

neunhoef:
This is good to go, except that there have been Jenkins Windows test failures: https://jenkins.arangodb.biz/view/PR/job/arangodb-matrix-pr/32276/ However, the same tests also fail in the devel nightly builds: https://jenkins.arangodb.biz/job/arangodb-ANY-windows-matrix/5372/#showFailuresLink

PR #20177, opened 2023-11-22
Teach PathValidator a global condition on edges

markuspf:
### Scope & Purpose

The PathValidator class can already enforce a global condition on all vertices; this patch also allows it to filter all edges on a path.

### DIFF
--- a/arangod/Graph/PathManagement/PathValidator.cpp
+++ b/arangod/Graph/PathManagement/PathValidator.cpp
@@ -302,6 +302,18 @@ auto PathValidator<ProviderType, PathStore, vertexUniqueness, edgeUniqueness>::
     }
   }
 
+  auto edgeExpr = _options.getEdgeExpression();
+  if (edgeExpr != nullptr) {
+    edgeBuilder.clear();
+    _provider.addEdgeToBuilder(step.getEdge(), edgeBuilder);
+
+    bool satisfiesCondition =
+        evaluateVertexExpression(edgeExpr, edgeBuilder.slice());
+    if (!satisfiesCondition) {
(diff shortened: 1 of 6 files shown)


markuspf:
@mchacki suggested a better change that enables use of indices right away; I will open another PR with this alternative code. In case the alternative code does not work out (yet), this PR can still be used.

markuspf:
Unfortunately @mchacki 's suggestion does not work, as the code makes assumptions that cannot be fulfilled. 

For the time being this code here will need to be used.

arangod/Graph/PathManagement/PathValidator.cpp
```diff
@@ -302,6 +302,22 @@ auto PathValidator<ProviderType, PathStore, vertexUniqueness, edgeUniqueness>::
...
+
+      _provider.addEdgeToBuilder(step.getEdge(), edgeBuilder);
+      bool satisfiesCondition =
+          evaluateExpression(edgeExpr, edgeBuilder.slice());
+      if (!satisfiesCondition) {
+        return ValidationResult{ValidationResult::Type::FILTER_AND_PRUNE};
+      }
+    } else {
+      // TODO: at the moment we smile and wave...
+    }
```
hkernbach, line 318: Would just remove that block - as it does nothing right now.

markuspf: I put that comment in there after debugging tests to make it clear that invalid edges just pass this check.

Tbh I don't entirely know what an "invalid" edge is and under which circumstances they would end up here. If I ADB_PROD_ASSERT() here, I think everything explodes

PR #20179, opened 2023-11-22
Try to stabilize actor test

mpoeter:
Sometimes one of the actor unit tests fails (eg. https://app.circleci.com/pipelines/github/arangodb/arangodb/19228/workflows/cf02640f-d09e-4252-86ae-c510828c264e/jobs/927780)

The reason is that the finish message is processed asynchronously, and the idle flag is set before the actor is removed and the down messages are dispatched. The `waitForAllMessagesToBeProcessed` function only indicates that the messages have been processed based on the idle state, but we don't know yet if the down messages have already been dispatched. So we have to wait until the system has converged to our expected state.

### DIFF
--- a/tests/Actor/RuntimeTest.cpp
+++ b/tests/Actor/RuntimeTest.cpp
@@ -459,6 +459,16 @@ TYPED_TEST(RuntimeTest, sends_down_message_to_monitoring_actors) {
                     FinishingActor::Message{
                         test::message::FinishingFinish{ExitReason::kShutdown}});
   waitForAllMessagesToBeProcessed(runtime);
+  for (unsigned i = 0; i < 10; ++i) {
+    if (runtime->actors.size() == 3 &&
+        runtime->template getActorStateByID<MonitoringActor>(monitor1)
+                ->deadActors.size() == 2 &&
+        runtime->template getActorStateByID<MonitoringActor>(monitor3)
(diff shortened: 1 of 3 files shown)


tests/Actor/RuntimeTest.cpp
```diff
@@ -459,6 +459,16 @@ TYPED_TEST(RuntimeTest, sends_down_message_to_monitoring_actors) {
                     FinishingActor::Message{
                         test::message::FinishingFinish{ExitReason::kShutdown}});
   waitForAllMessagesToBeProcessed(runtime);
+  for (unsigned i = 0; i < 10; ++i) {
```
markuspf, line 462: I cannot begin to say how much I dread this solution.

I wonder whether it would make anything better if at least the retries are printed ("attempt n of 10") and the test fails with "out of retries" when the number of retries is exhausted?

mpoeter: I am not a fan either, but I could not think of anything else. With the MockScheduler everything is fine and deterministic - this problem only occurs with the ThreadPoolScheduler, because then we really have async stuff happening _eventually_. If all retries are exhausted and we still don't have the expected state, then the following EXPECT_EQ will fail, and we will at least see what our current state is. But I can also print each failed test if you prefer that.

markuspf (approved):
Love it!

PR #20181, opened 2023-11-22
Make sure that temporary variables name agrees with their id.

maierlars:
### Scope & Purpose

Temporary variables have a numeric name and, similar to all other variables, a unique id. The name and the id were not the same, but offset by 1. This is very confusing. This PR fixes that.


arangod/Aql/VariableGenerator.cpp
```diff
@@ -65,21 +65,27 @@ std::unordered_map<VariableId, std::string const> VariableGenerator::variables(
...
+    }
+  }();
 
   if (isUserDefined && !isValidName(name.data(), name.data() + name.size())) {
     THROW_ARANGO_EXCEPTION_MESSAGE(
         TRI_ERROR_QUERY_PARSE,
         arangodb::basics::Exception::FillExceptionString(
-            TRI_ERROR_QUERY_VARIABLE_NAME_INVALID, temp.c_str()));
+            TRI_ERROR_QUERY_VARIABLE_NAME_INVALID, nameCopy.c_str()));
   }
```
jsteemann, line 83: In case we run into this exception, we will needlessly have incremented the id value via a call to `nextId()`.
But we don't need the id value in case a user-defined name is provided.

So we would as well move the id increment call to below this check and thus avoid increasing it in case of obvious errors. This is not critical but may even make the resulting code simpler.

markuspf (approved):

jsteemann (approved):
LGTM

PR #20182, opened 2023-11-22
Improve replace-equal-attribute-access rule.

maierlars:
### Scope & Purpose

The `replace-equal-attribute-access` rule now works for subqueries as well. Furthermore, if one side is already a single variable expression, it uses this variable instead of introducing another alias variable.

Finally the optimizer rule was moved after `moveFiltersUpRule2` which ensures that we can replace as many values as possible.


jsteemann (approved):
LGTM

PR #20187, opened 2023-11-23
Fixed MDS-1170

jsteemann:
### Scope & Purpose

Fixes https://arangodb.atlassian.net/browse/MDS-1170

* Fixed MDS-1170: Significant performance degradation when queries are executed within transactions that involve edits.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20188
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/20189

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/MDS-1170
- [ ] Design document:


arangod/RocksDBEngine/RocksDBIterators.cpp
```diff
@@ -209,7 +209,25 @@ class RocksDBAllIndexIterator final : public IndexIterator {
...
+            // when the iterator is a db snapshot-based iterator, it is ok
+            // to set iterate_lower_bound/iterate_upper_bound, because this
+            // is well supported by RocksDB.
+            // if the iterator is a multi-level iterator that merges data from
+            // the db snapshot with data from an ongoing in-memory transaction
+            // (contained in a WriteBatchWithIndex, WBWI), then RocksDB does
+            // not properly support the bounds checking using
+            // iterate_lower_bound/ iterate_upper_bound. in this case we must
+            // avoid setting the bounds here and rely on our own bounds checking
+            // using the comparator. at least one underlying issue was fixed in
```
MBkkt, line 224: Why cannot we always rely on "our own bounds checking using the comparator"?

MBkkt (approved):

arangod/RocksDBEngine/RocksDBIterators.cpp
jsteemann: Because it is much more expensive.

jsteemann: In case `iterate_upper_bound` / `iterate_lower_bound` are set, then RocksDB can filter out things on the .sst file level already, and will do some comparisons of keys against `iterate_upper_bound` / `iterate_lower_bound` itself using the comparator.

So setting the bounds is normally a good idea.
Unfortunately it is not supported for iterators that refer to the WriteBatchWithIndex (WBWI). This was a bug that we reported to RocksDB years ago, and it was only recently fixed for RocksDB 8.8, which were are not yet using.

PR #20190, opened 2023-11-23
AST Pattern Matching

markuspf:
### Scope & Purpose

Adds a small libraries of helpers to simplify matching on Expressions when writing optimiser rules.

A common pattern when dealing with Expressions in AQL is to try and match a pattern extracting certain information. In my case I was looking for 

```
paths.vertices[* RETURN CURRENT.colour == "green"] ALL == true
```
where `paths` is the output variable of a path search (ALL_SHORTEST_PATHS, K_SHORTEST_PATHS, K_PATHS), and I want to extract the expression after the `*`, and the right hand side of the `ALL ==` operator.

The code to detect this pattern is repetetive, annoying to write, and usually one does not go through the effort of adding error messages when a match fails.

There are many instances of the type of code in the optimiser which inspect ASTs, for instance, an expansion node (what happenes when you write `array[* ...]`). One has to go and look up what the separate subnodes mean, gets it wrong a couple of times, etc.

This library aims to provide something akin to "pattern matching" on an AST. The idea is that the optimizer-rule-writer writes out a pattern:

```
auto matcher = arrayEq(                                          //
        expansion(                                               //
            iterator(anyVariable(),                              //
                     attributeAccess(reference("path"),          //
                                     {"edges", "vertices"})),    //
            matchWithName("variable", Reference{.name = "3_"}),  //
            NoOp{},                                              //
            NoOp{},                                              //
            matchWithName("map", Any{})),                        //
        anyValue(),                                              //
        quantifier(::arangodb::aql::Quantifier::Type::kAll);
```  

and the library figures out the walking of the AST and assigning results to a data structure for consumption, or returning some message as to when a match has failed and why.

The library is extensible: to write one's own matcher, one has to implement a struct that fulfils the concept `Matchable`, which currently only contains one function `apply` that takes an `AstNode const*` and returns a `MatchResult`.

At this point the library is still a work in progress. The most important things that are missing for my application are more tests and being able to match on `n-ary` nodes with an `all` and a `some` match, extracting a list of matches; this will require generalising the `MatchResult` struct.

Matchers for a lot of `AstNode` types are also still missing.

Another extension would be keeping a global matcher state to do context-sensitive matching.


arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,103 @@
...
+    return std::move(*this);
+  }
+
+  auto combine(MatchResult&& rhs) -> MatchResult&& {
+    _errors.insert(std::end(_errors), std::begin(rhs._errors),
+                   std::end(rhs._errors));
+    _matches.merge(rhs._matches);
+
+    return std::move(*this);
+  }
```
mpoeter, line 65: Since these return an rvalue ref to this, they should probably have an rvalue ref-qualifier
```suggestion
  auto addError(std::string error) && -> MatchResult&& {
    _errors.emplace_back(error);
    return std::move(*this);
  }

  auto addErrorIfError(std::string error) && -> MatchResult&& {
    if (isError()) {
      _errors.emplace_back(error);
    }
    return std::move(*this);
  }

  auto addMatch(std::string name, AstNode const* value) && -> MatchResult&& {
    _matches.emplace(name, value);
    return std::move(*this);
  }

  auto combine(MatchResult&& rhs) && -> MatchResult&& {
    _errors.insert(std::end(_errors), std::begin(rhs._errors),
                   std::end(rhs._errors));
    _matches.merge(rhs._matches);

    return std::move(*this);
  }
```

arangod/Aql/Optimizer/ExpressionMatcher/Matchers/NAryAnd.h
```diff
@@ -0,0 +1,41 @@
...
+
+template<Matchable M>
+struct NAryAndAll {
+  auto apply(AstNode const* node) -> MatchResult {
+    return MatchNodeType(NODE_TYPE_OPERATOR_NARY_AND).apply(node)  //
+           | [&]() {
+               for (auto i = size_t{0}; i < node->numMembers(); ++i) {
+                 auto snr = matcher.apply(node->getMemberUnchecked(i));
+               }
+             };
```
mpoeter, line 36: This looks strange - `snr` is assigned but never used.

arangod/Aql/Optimizer/ExpressionMatcher/Matchers/Variable.cpp
```diff
@@ -0,0 +1,28 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+#include "Variable.h"
+
+#include <optional>
+#include <variant>
```
mpoeter, line 24:
```suggestion
```

arangod/Aql/Optimizer/ExpressionMatcher/Matchers/NAryAnd.h
markuspf: yes, this code is WIP to get to a minimum state where I can use it for the optimizer rule

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+
+  struct ErrorTag {};
+  struct ValueTag {};
+
+  explicit MatchResult() = default;
+  MatchResult(MatchResult&&) = default;
+  MatchResult(MatchResult const&) = delete;
+
+  auto addError(std::string error) && -> MatchResult&& {
+    _errors.emplace_back(error);
```
mpoeter, line 43:
```suggestion
    _errors.emplace_back(std::move(error));
```

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+
+  auto addErrorIfError(std::string error) && -> MatchResult&& {
+    if (isError()) {
+      _errors.emplace_back(error);
+    }
+    return std::move(*this);
+  }
+
+  auto addMatch(std::string name, AstNode const* value) && -> MatchResult&& {
+    _matches.emplace(name, value);
```
mpoeter, line 55:
```suggestion
    _matches.emplace(std::move(name), value);
```

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+  auto combine(MatchResult&& rhs) && -> MatchResult&& {
+    _errors.insert(std::end(_errors), std::begin(rhs._errors),
+                   std::end(rhs._errors));
+    _matches.merge(rhs._matches);
+
+    return std::move(*this);
+  }
+
+  static auto error(std::string message) -> MatchResult {
+    return MatchResult().addError(message);
```
mpoeter, line 68:
```suggestion
    return MatchResult().addError(std::move(message));
```

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+  }
+  static auto match(std::string key, AstNode const* value) {
+    return MatchResult().addMatch(key, value);
```
mpoeter, line 71:
```suggestion
    return MatchResult().addMatch(std::move(key), value);
```

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+    return _errors;
+  }
+  auto matches() const noexcept
+      -> std::unordered_map<std::string, AstNode const*> const& {
+    return _matches;
+  }
+
+  friend inline auto operator/(MatchResult&& lhs, std::string error)
+      -> MatchResult&& {
+    lhs._errors.emplace_back(error);
```
mpoeter, line 85:
```suggestion
    lhs._errors.emplace_back(std::move(error));
```

arangod/Aql/Optimizer/ExpressionMatcher/MatchResult.h
```diff
@@ -0,0 +1,105 @@
...
+    return std::move(lhs);
+  }
+
+  template<typename Func>
+  friend inline auto operator|(MatchResult&& lhs, Func&& func)
+      -> MatchResult&& {
+    if (lhs.isError()) {
+      return std::move(lhs);
+    }
+    return std::move(lhs).combine(func());
```
mpoeter, line 95:
```suggestion
    return std::move(lhs).combine(std::forward<Func>(func)());
```

arangod/Aql/Optimizer/ExpressionMatcher/Matchers/WithName.h
```diff
@@ -0,0 +1,50 @@
...
+    return MatchResult::match(name, node);
+  }
+  std::string name;
+  M matcher;
+};
+
+// Helper for convenient template argument deduction
+template<Matchable M>
+auto matchWithName(std::string name, M matcher) -> MatchWithName<M> {
+  return MatchWithName<M>{.name = std::move(name), .matcher = matcher};
```
mpoeter, line 47:
```suggestion
  return MatchWithName<M>{.name = std::move(name), .matcher = std::move(matcher)};
```

PR #20194, opened 2023-11-23
[CINFRA] Index Creation Race

apetenchea:
### Scope & Purpose

The replication2 index creation code (_ensureIndexCoordinatorReplication2Inner_) is affected by a race. This PR provides the fix.

**Problem we're trying to fix**

There are 3 things that have play along:
- clusterInfo.loadPlan()
- callbackRegistry.waitFor()
- clusterInfo.waitForPlan()

The race is as follows:

1. We try to create the index. Coordinator writes index to Target. Agency sees the index in Target, writes it to Plan and increments Plan/Version.
2. Coordinator is slow. Before even registering the callback for the index in Plan, ClusterInfo (via the syncer thread on coordinator) does `loadPlan()` and observes the change (because the plan version is updated). This also updates the raft index in ClusterInfo.
3. Before the coordinator registers the Plan agency callback, something random happens in the agency, say a server goes bad. This pushes the commitIndex further, without modifying the planVersion, therefore not triggering another loadPlan.
4. Coordinator finally registers callbackRegistry.waitFor() for the collection in Plan, which triggers immediately with the initial value, and returns successfully since the index is already there.
5. Subsequent clusterInfo.waitForPlan() never returns because the index observed by callbackRegistry is already higher than the last index which was seen from `loadPlan()`. Another `loadPlan()` does not happen unless Plan/Version is incremented.

For reproducing the issue reliably see https://github.com/arangodb/arangodb/pull/20163

**Solution**

Instead of waiting for the agency's raft Index, we'll make ClusterInfo catch up with the plan version. Since we already have our hands on the raft index, we still quickly check it, just in case we may be able to skip the waitFor. Otherwise, we get the plan version from the agency cache, which is guaranteed to be greater or equal to the plan version observed during index creation, and wait until ClusterInfo catches up with that.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/Cluster/ClusterIndexMethods.cpp
+++ b/arangod/Cluster/ClusterIndexMethods.cpp
@@ -886,10 +886,21 @@ auto ensureIndexCoordinatorReplication2Inner(
                   }
                   return false;
                 })
-            .thenValue([&clusterInfo](auto index) {
-              // Need to wait here, until ClusterInfo has updated to latest
-              // plan.
-              return clusterInfo.waitForPlan(index);
+            .thenValue([&clusterInfo,
+                        &server](auto index) -> futures::Future<Result> {
+              if (clusterInfo.getPlanIndex() < index) {
+                // Need to wait here, until ClusterInfo has updated to latest
+                // plan.
+                auto& agencyCache =
+                    server.getFeature<ClusterFeature>().agencyCache();
+                auto [version, _] = agencyCache.read(std::vector<std::string>{
+                    AgencyCommHelper::path("Plan/Version")});
+                auto planVersion = version->slice()[0]
+                                       .get({"arango", "Plan", "Version"})
+                                       .getNumber<uint64_t>();
+                return clusterInfo.waitForPlanVersion(planVersion);
+              }
+              return Result{};
             });
     auto res = waitOnSuccess.get();
     if (res.fail()) {
--- a/arangod/Cluster/ClusterInfo.h
+++ b/arangod/Cluster/ClusterInfo.h
@@ -883,6 +883,11 @@ class ClusterInfo final {
     return _planVersion;
   }
 
+  uint64_t getPlanIndex() const {
+    READ_LOCKER(readLocker, _planProt.lock);
+    return _planIndex;
+  }
+
   uint64_t getCurrentVersion() const {
     READ_LOCKER(guard, _currentProt.lock);
     return _currentVersion;


arangod/Cluster/ClusterIndexMethods.cpp
```diff
@@ -886,10 +886,21 @@ auto ensureIndexCoordinatorReplication2Inner(
...
+                // Need to wait here, until ClusterInfo has updated to latest
+                // plan.
+                auto& agencyCache =
+                    server.getFeature<ClusterFeature>().agencyCache();
+                auto [version, _] = agencyCache.read(std::vector<std::string>{
+                    AgencyCommHelper::path("Plan/Version")});
+                auto planVersion = version->slice()[0]
+                                       .get({"arango", "Plan", "Version"})
+                                       .getNumber<uint64_t>();
+                return clusterInfo.waitForPlanVersion(planVersion);
```
apetenchea, line 901: Note the existence of [ClusterInfo::fetchAndWaitForPlanVersion() ](https://github.com/arangodb/arangodb/blob/devel/arangod/Cluster/ClusterInfo.cpp#L5587) which serves the same purpose in [ReplicatedLogMethodsCoordinator::createReplicatedLog()](https://github.com/arangodb/arangodb/blob/devel/arangod/Replication2/Methods.cpp#L423).
The only difference is that it sends a request to the agency, while we're using the agency cache here, and it also works based on a timeout. Happy to switch to that method if this patch seems redundant.

mchacki: Please keep this version, as we save the unnecessary call to the agency.
All required information is available locally.

arangod/Cluster/ClusterIndexMethods.cpp
```diff
@@ -886,10 +886,21 @@ auto ensureIndexCoordinatorReplication2Inner(
                   }
                   return false;
                 })
-            .thenValue([&clusterInfo](auto index) {
-              // Need to wait here, until ClusterInfo has updated to latest
-              // plan.
-              return clusterInfo.waitForPlan(index);
+            .thenValue([&clusterInfo,
+                        &server](auto index) -> futures::Future<Result> {
+              if (clusterInfo.getPlanIndex() < index) {
```
mchacki, line 891: Good improvement over our discussion :+1:

mchacki (approved):
LGTM

PR #20196, opened 2023-11-23
Bug fix/replace shardid typedefstring by class

mchacki:
### Scope & Purpose

*Replaces the ShardID as string typedef by a class.
This fixes issues with numeric sorting of shard names.
This also clears up some incensistency of usages as Collections,Servers and ShardIds where all strings and soldomly the wrong one was taken.
Unfortunately moves the AgencyPaths to use std::string instead of ShardID.
*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1412
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,11 @@
 devel
 -----
 
+* EE Only: When creating a disjoint Smart Graph or a SmartJoin collection where
```
jsteemann, line 4: Our spelling is like this:
```suggestion
* EE Only: When creating a disjoint SmartGraph or a SmartJoin collection where
```

CHANGELOG
```diff
@@ -1,6 +1,11 @@
...
+  the list of shard names would go from a k digit number to a k+1 digit number,
+  some AQL queries would miss-match the shard list of one collection with the others.
```
```suggestion
  some AQL queries would mismatch the shard list of one collection with the others.
```

CHANGELOG
```diff
@@ -1,6 +1,11 @@
...
+  The data is all in correct place, this is a read-only issue.
```
```suggestion
  The data is all in the correct place, this is an issue only in read operations.
```

arangod/Agency/Supervision.cpp
```diff
@@ -3554,6 +3554,12 @@ void Supervision::checkUndoLeaderChangeActions() {
         return Result{TRI_ERROR_BAD_PARAMETER, "collection missing"};
       }
 
+      auto maybeShardID = ShardID::shardIdFromString(id);
+      if (ADB_UNLIKELY(maybeShardID.fail())) {
+        return Result{TRI_ERROR_BAD_PARAMETER,
```
jsteemann, line 3559: It would probably be better to return the error message generated by `shardIdFromString()` here. If it returns a `ResultT` and there is an error, that method should have populated the error message with something useful already, so we can simply return it here.

arangod/Aql/CollectionAccessingNode.cpp
```diff
@@ -67,7 +70,15 @@ CollectionAccessingNode::CollectionAccessingNode(
   VPackSlice restrictedTo = slice.get("restrictedTo");
 
   if (restrictedTo.isString()) {
-    _restrictedTo = restrictedTo.copyString();
+    auto maybeShardID = ShardID::shardIdFromString(restrictedTo.copyString());
+    if (ADB_UNLIKELY(maybeShardID.fail())) {
+      LOG_TOPIC("63e33", WARN, Logger::AQL)
+          << "Internal JSON representation of CollectionAccess for collection "
+          << colName << " contains invalid shard ID "
+          << restrictedTo.copyString() << " - ignoring";
```
jsteemann, line 78: We input parameter of `ShardID::shardIdFromString()` should be an std::string_view, not an std::string.
Then this code part can be made more efficient as:
```suggestion
    auto maybeShardID = ShardID::shardIdFromString(restrictedTo.stringView());
    if (ADB_UNLIKELY(maybeShardID.fail())) {
      LOG_TOPIC("63e33", WARN, Logger::AQL)
          << "Internal JSON representation of CollectionAccess for collection "
          << colName << " contains invalid shard ID "
          << restrictedTo.stringView() << " - ignoring";
```

arangod/Aql/CollectionAccessingNode.cpp
```diff
@@ -152,16 +163,17 @@ aql::Collection const* CollectionAccessingNode::collection() const {
...
 }
 
 bool CollectionAccessingNode::isRestricted() const {
-  return !_restrictedTo.empty();
+  return _restrictedTo.has_value();
 }
 
-std::string const& CollectionAccessingNode::restrictedShard() const {
-  return _restrictedTo;
+ShardID const& CollectionAccessingNode::restrictedShard() const {
```
jsteemann, line 174: As ShardIDs are now light-weight, we should now make this function return a copy of the ShardID.
```suggestion
ShardID CollectionAccessingNode::restrictedShard() const {
```

arangod/Aql/GraphNode.cpp
```diff
@@ -479,7 +479,12 @@ GraphNode::GraphNode(ExecutionPlan* plan,
         "graph needs a translation from collection to shard names");
   }
   for (auto const& item : VPackObjectIterator(collectionToShard)) {
-    _collectionToShard.insert({item.key.copyString(), item.value.copyString()});
+    auto maybeShardID = ShardID::shardIdFromString(item.value.copyString());
```
jsteemann, line 482:
```suggestion
    auto maybeShardID = ShardID::shardIdFromString(item.value.stringView());
```

arangod/Aql/OptimizerRules.cpp
```diff
@@ -4967,17 +4968,17 @@ void arangodb::aql::restrictToSingleShardRule(
         auto collection =
             ExecutionNode::castTo<ModificationNode const*>(current)
                 ->collection();
-        std::string shardId =
+        auto shardId =
             ::getSingleShardId(plan.get(), current, collection);
 
-        if (!shardId.empty()) {
+        if (shardId.has_value() && shardId.value().isValid()) {
```
jsteemann, line 4974: This looks a bit strange. If `getSingleShardId` returns a ShardID value, shouldn't it always return a valid ShardID value? If not, we should maybe change that function accordingly.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1540,22 +1540,20 @@ void ClusterInfo::loadPlan() {
...
           v->assign(p.second.begin(), p.second.end());
-          newShardsToPlanServers.insert_or_assign(
-              pmr::ShardID{p.first, _resourceMonitor}, std::move(v));
+          newShardsToPlanServers.insert_or_assign(p.first, std::move(v));
           newShardToName.insert_or_assign(p.first, newCollection->name());
         }
 
         // Sort by the number in the shard ID ("s0000001" for example):
-        ShardingInfo::sortShardNamesNumerically(*shards);
+        std::sort(shards->begin(), shards->end());
```
jsteemann, line 1556: Instead of using a vector and then sorting it, we could use a sorted container here from the get-go.
That may require changing other places too, so please treat this as optional.

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1970,9 +1968,15 @@ void ClusterInfo::loadCurrent() {
 
       for (auto const& shardSlice :
            velocypack::ObjectIterator(collectionSlice.value)) {
-        pmr::ManagedString shardID{shardSlice.key.copyString(),
-                                   _resourceMonitor};
 
+        auto maybeShardID = ShardID::shardIdFromString(shardSlice.key.copyString());
```
jsteemann, line 1972:
```suggestion
        auto maybeShardID = ShardID::shardIdFromString(shardSlice.key.stringView());
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1970,9 +1968,15 @@ void ClusterInfo::loadCurrent() {
...
+        if (ADB_UNLIKELY(maybeShardID.fail())) {
+          TRI_ASSERT(false) << "Indexed malformed shard name "<< shardSlice.key.copyString();
+          // TODO cannot handle this entry, is it better to continue or to abort here?
+          continue;
+        }
+
+        auto shardID = std::move(maybeShardID.get());
```
jsteemann, line 1979: `maybeShardID.get()` is already an rvalue. No need to add an extra `std::move()`.
```suggestion
        auto shardID = maybeShardID.get();
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -941,9 +941,17 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
                   if (dbs.isEqualString(serverId) ||
                       dbs.isEqualString(UNDERSCORE + serverId)) {
                     // at this point a shard is in plan, we have the db for it
+                    auto maybeShardID = ShardID::shardIdFromString(shard.key.copyString());
```
jsteemann, line 944:
```suggestion
                    auto maybeShardID = ShardID::shardIdFromString(shard.key.stringView());
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -941,9 +941,17 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
...
+                    if (ADB_UNLIKELY(maybeShardID.fail())) {
+                      TRI_ASSERT(false)
+                          << "Malformed shard ID in agency: "
+                          << shard.key.copyString() << " in db: " << dbname
+                          << " and collection: " << pcol.key.copyString();
```
jsteemann, line 949:
```suggestion
                          << " and collection: " << pcol.key.stringView();
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -1008,9 +1016,15 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
         }
 
         for (auto const& lcol : VPackObjectIterator(ldbslice)) {
-          auto const& colname = lcol.key.copyString();
           auto const shardMap = getShardMap(plan);  // plan shards -> servers
-          handleLocalShard(ldbname, colname, lcol.value, shardMap.slice(),
+          auto maybeShardID = ShardID::shardIdFromString(lcol.key.copyString());
```
jsteemann, line 1020:
```suggestion
          auto maybeShardID = ShardID::shardIdFromString(lcol.key.stringView());
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -1008,9 +1016,15 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
...
+          if (ADB_UNLIKELY(maybeShardID.fail())) {
+            TRI_ASSERT(false)
+                << "Malformed shard ID in agency: "
+                << lcol.key.copyString() << " in db: " << ldbname;
```
jsteemann, line 1024:
```suggestion
                << lcol.key.stringView() << " in db: " << ldbname;
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -1964,7 +1981,7 @@ arangodb::Result arangodb::maintenance::reportInCurrent(
       }
 
       for (auto const& shard : VPackObjectIterator(ldb, true)) {
-        auto const shName = shard.key.copyString();
+        auto const shName = ShardID{shard.key.copyString()};
```
jsteemann, line 1984:
```suggestion
        auto const shName = ShardID{shard.key.stringView()};
```

arangod/Cluster/TraverserEngine.cpp
```diff
@@ -122,14 +122,16 @@ BaseEngine::BaseEngine(TRI_vocbase_t& vocbase, aql::QueryContext& query,
...
     ResourceUsageAllocator<MonitoredCollectionToShardMap, ResourceMonitor>
         alloc = {_query.resourceMonitor()};
-    MonitoredStringVector shards{alloc};
+    MonitoredShardIDVector shards{alloc};
     TRI_ASSERT(collection.value.isArray());
     for (auto shard : VPackArrayIterator(collection.value)) {
       TRI_ASSERT(shard.isString());
-      std::string name = shard.copyString();
-      _query.collections().add(name, AccessMode::Type::READ,
+      auto shardString = shard.copyString();
```
jsteemann, line 129:
```suggestion
      auto shardString = shard.stringView();
```

arangod/Cluster/TraverserEngine.cpp
```diff
@@ -203,7 +205,7 @@ void BaseEngine::getVertexData(VPackSlice vertex, VPackBuilder& builder,
       return true;
     };
     for (auto const& shard : shards->second) {
-      Result res = _trx->documentFastPathLocal(shard, vertex, cb).get();
+      Result res = _trx->documentFastPathLocal(std::string{shard}, vertex, cb).get();
```
jsteemann, line 208:
```suggestion
      Result res = _trx->documentFastPathLocal(std::string_view{shard}, vertex, cb).get();
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+#include <compare>
+#include <cstdint>
+#include <fmt/format.h>
+#include <velocypack/Value.h>
+
+namespace arangodb {
+
+struct ShardID {
+
+  static ResultT<ShardID> shardIdFromString(std::string const& s) {
```
jsteemann, line 39: Additionally, the header `<string_view>` should be included here.
```suggestion
  static ResultT<ShardID> shardIdFromString(std::string_view s) {
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+    if (s.empty() || s.at(0) != 's') {
```
jsteemann, line 40:
```suggestion
    if (!s.starts_with('s')) {
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+      return Result{TRI_ERROR_BAD_PARAMETER, "Expected ShardID to start with 's'"};
+    }
+    auto res = basics::StringUtils::try_uint64(s.c_str() + 1, s.length() -1);
```
jsteemann, line 43:
```suggestion
    auto res = basics::StringUtils::try_uint64(s.data() + 1, s.size() - 1);
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+      return std::move(res.result());
+    }
+    return ShardID{res.get()};
+  }
+
+  static ShardID invalidShard() { return ShardID{0}; }
+
+  // Default constructor required for Inspectors.
+  // Note: This shardID is considered Invalid.
+  ShardID() : id(0) {}
```
jsteemann, line 54:
```suggestion
  constexpr ShardID() noexcept : id(0) {}
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+
+  explicit ShardID(uint64_t id) : id(id) {}
+
+  explicit ShardID(std::string_view id) {
+    auto maybeShardID = shardIdFromString(std::string(id));
+    if (maybeShardID.fail()) {
+      THROW_ARANGO_EXCEPTION(maybeShardID.result());
+    }
+    *this = maybeShardID.get();
```
jsteemann, line 63: While this may work, why don't we make a private function that extracts the numeric id from an input string and call it from here, then we can avoid overwriting `this`.

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+    return ShardID{res.get()};
+  }
+
+  static ShardID invalidShard() { return ShardID{0}; }
+
+  // Default constructor required for Inspectors.
+  // Note: This shardID is considered Invalid.
+  ShardID() : id(0) {}
+
+  explicit ShardID(uint64_t id) : id(id) {}
```
jsteemann, line 56:
```suggestion
  explicit ShardID(uint64_t id) noexcept : id(id) {}
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+  ShardID(ShardID&&) = default;
+  ShardID& operator=(ShardID const&) = default;
+  ShardID& operator=(ShardID&&) = default;
+
+  operator std::string() const {
+    return "s" + std::to_string(id);
+  }
+
+  operator arangodb::velocypack::Value() const {
+    return arangodb::velocypack::Value("s" + std::to_string(id));
```
jsteemann, line 78: I don't think that this will work safely.
The returned velocypack::Value string points into a temporary that will be invalid when the function scope is left.
This looks like undefined behavior!

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+
+
+  ~ShardID() = default;
+  ShardID(ShardID const&) = default;
+  ShardID(ShardID&&) = default;
+  ShardID& operator=(ShardID const&) = default;
+  ShardID& operator=(ShardID&&) = default;
+
+  operator std::string() const {
+    return "s" + std::to_string(id);
```
jsteemann, line 74: Probably better (requires including `absl/strings/str_cat.h` in addition):
```suggestion
    return absl::StrCat("s", id);
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+  friend auto operator<=>(ShardID const&, ShardID const&) = default;
+  friend bool operator==(ShardID const&, ShardID const&) = default;
+
+  bool operator==(std::string_view other) const {
+    return other == std::string{*this};
+  }
+
+  bool operator==(std::string const& other) const {
+    return other == std::string{*this};
+  }
```
jsteemann, line 90: Is the comparison for an std::string actually necessary if we already provide a comparion for std::string_view?
```suggestion
  bool operator==(std::string_view other) const noexcept {
    return other.id == id;
  }

  bool operator==(std::string const& other) const noexcept {
    return other.id == id;
  }
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+    if constexpr (Inspector::isLoading) {
+      std::string v;
+      auto res = f.apply(v);
+      if (res.ok()) {
+        auto result = shardIdFromString(v);
+        if (result.fail()) {
+          // NOTE: For some reason we cannot inline the errorMessage() into the return
+          // in that case the string_view variant will be selected, which is no valid input
+          // for status.
+          std::string msg{std::move(result.errorMessage())};
```
jsteemann, line 110: The extra std::move looks redundant.
```suggestion
          std::string msg{result.errorMessage()};
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+          return {std::move(msg)};
+        }
+        x = result.get();
+        return {};
+      }
+      return res;
+    } else {
+      return f.apply("s" + std::to_string(x.id));
```
jsteemann, line 118:
```suggestion
      return f.apply(absl::StrCat("s", x.id));
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  uint64_t id;
+};
+
+// Make ShardID logable
```
jsteemann, line 125: The implementation should probably be moved into a .cpp file.
```suggestion
// Make ShardID loggable
```

arangod/Cluster/Utils/ShardID.h
jsteemann, line 39: I would also suggest to move the implementation of this function (and several others) into a .cpp file.

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+};
+
+// Allow ShardID to be added to std::strings
+static inline std::string operator+(std::string const& text, arangodb::ShardID const& s) {
+  return text + "s" + std::to_string(s.id);
+}
+// Allow ShardID to be added to std::strings
+static inline std::string operator+(arangodb::ShardID const& s, std::string const& text) {
+  return  "s" + std::to_string(s.id) + text;
+}
```
jsteemann, line 161: Instead of declaring these static inline, we should move the implementation into a .cpp file.
```suggestion
static inline std::string operator+(std::string const& text, arangodb::ShardID const& s) {
  return absl::StrCat(text, "s", s.id);
}
// Allow ShardID to be added to std::strings
static inline std::string operator+(arangodb::ShardID const& s, std::string const& text) {
  return absl::StrCat("s", s.id, text);
}
```

arangod/Cluster/v8-cluster.cpp
```diff
@@ -985,13 +998,20 @@ static void JS_GetResponsibleServersClusterInfo(
...
   uint32_t const n = array->Length();
   for (uint32_t i = 0; i < n; ++i) {
-    shardIds.emplace(TRI_ObjectToString(
+    auto maybeShard = ShardID::shardIdFromString(TRI_ObjectToString(
         isolate, array->Get(context, i).FromMaybe(v8::Local<v8::Value>())));
+    if (maybeShard.fail()) {
+      TRI_V8_THROW_EXCEPTION_MESSAGE(
+          maybeShard.errorNumber(), "Error on entry " + std::to_string(i) +
+                                        ": " +
+                                        std::string(maybeShard.errorMessage()));
```
jsteemann, line 1012:
```suggestion
          maybeShard.errorNumber(), absl::StrCat("Error on entry ", i, ": ", maybeShard.errorMessage()));
```

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -150,7 +150,7 @@ void ClusterIndex::toVelocyPack(
       std::string const url =
           prefix + shard.first + "/" + std::to_string(_iid.id());
       futures.emplace_back(
-          network::sendRequestRetry(pool, "shard:" + shard.first,
+          network::sendRequestRetry(pool, "shard:" + std::string(shard.first),
```
jsteemann, line 153: Given that there is an operator+ defined for ShardID, do we need the explicit conversion to an std::string here?

arangod/Graph/Cache/RefactoredTraverserCache.cpp
```diff
@@ -258,7 +258,7 @@ bool RefactoredTraverserCache::appendVertex(
       };
       Result res =
           _trx->documentFastPathLocal(
-                  collectionName,
+                  shardId.c_str(),
```
jsteemann, line 261:
```suggestion
                  shardId,
```

arangod/Graph/Cache/RefactoredTraverserCache.cpp
```diff
@@ -274,8 +274,8 @@ bool RefactoredTraverserCache::appendVertex(
       if (isWithClauseMissing(ex)) {
         // turn the error into a different error
         auto message =
-            absl::StrCat("collection not known to traversal: '", collectionName,
-                         "'. please add 'WITH ", collectionName,
+            absl::StrCat("collection not known to traversal: '", shardId.c_str(),
+                         "'. please add 'WITH ", shardId.c_str(),
```
jsteemann, line 278:
```suggestion
            absl::StrCat("collection not known to traversal: '", shardId,
                         "'. please add 'WITH ", shardId,
```

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -93,7 +93,12 @@ auto DocumentFollowerState::getAssociatedShardList() const
   std::vector<ShardID> shardIds;
   shardIds.reserve(collections.size());
   for (auto const& shard : collections) {
-    shardIds.emplace_back(shard->name());
+    auto maybeShardId = ShardID::shardIdFromString(shard->name());
+    ADB_PROD_ASSERT(maybeShardId.ok())
+        << "Tried to produce shard list on Database Server for a collection "
+           "that is not a shard "
+        << shard->name();
+    shardIds.emplace_back(std::move(maybeShardId.get()));
```
jsteemann, line 101: do we need the extra std::move here?
```suggestion
    shardIds.emplace_back(std::move(maybeShardId.get()));
```

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -444,7 +444,13 @@ auto DocumentLeaderState::getAssociatedShardList() const
...
   shardIds.reserve(shards.size());
   for (auto&& shardId : shards) {
-    shardIds.emplace_back(shardId->name());
+    auto maybeShardId = ShardID::shardIdFromString(shardId->name());
+    ADB_PROD_ASSERT(maybeShardId.ok())
+        << "Tried to produce shard list on Database Server for a collection "
+           "that is not a shard "
+        << shardId->name();
+    shardIds.emplace_back(std::move(maybeShardId.get()));
+    shardIds.emplace_back(std::move(maybeShardId.get()));
```
jsteemann, line 453: Why are we added the shard twice, especially after moving it out of `maybeShardId`? Looks like a copy&paste error!
```suggestion
    shardIds.emplace_back(std::move(maybeShardId.get()));
```

arangod/Sharding/ShardDistributionReporter.cpp
```diff
@@ -201,14 +201,14 @@ static void ReportPartialNoProgress(
...
-  containers::SmallVector<std::string_view, 8> sortedShards;
+  containers::SmallVector<ShardID, 8> sortedShards;
   sortedShards.reserve(shardIds.size());
 
   for (auto const& s : shardIds) {
     sortedShards.emplace_back(s.first);
   }
   // sorts sortedShards in place in a deterministic order
-  ShardingInfo::sortShardNamesNumerically(sortedShards);
+  std::sort(sortedShards.begin(), sortedShards.end());
```
jsteemann, line 211: Why not make `sortedShards` a sorted container in the first place?

arangod/Sharding/ShardingInfo.cpp
```diff
@@ -161,7 +161,8 @@ ShardingInfo::ShardingInfo(arangodb::velocypack::Slice info,
   if (shardsSlice.isObject()) {
     for (auto const& shardSlice : VPackObjectIterator(shardsSlice)) {
       if (shardSlice.key.isString() && shardSlice.value.isArray()) {
-        ShardID shard = shardSlice.key.copyString();
+        // NOTE: Can throw if shard is not a valid shard name
+        ShardID shard{shardSlice.key.copyString()};
```
jsteemann, line 165:
```suggestion
        ShardID shard{shardSlice.key.stringView()};
```

arangod/Sharding/ShardingInfo.cpp
```diff
@@ -478,7 +479,7 @@ std::shared_ptr<std::vector<ShardID>> ShardingInfo::shardListAsShardID() const {
   for (auto const& mapElement : *_shardIds) {
     vector->emplace_back(mapElement.first);
   }
-  sortShardNamesNumerically(*vector);
+  std::sort(vector->begin(), vector->end());
```
jsteemann, line 482: Would be better if we could turn `vector` into a sorted container.

arangod/VocBase/Methods/Collections.cpp
```diff
@@ -802,7 +798,7 @@ Collections::create(         // create collection
     auto res = validateCreationInfo(
         infos.front(), vocbase, allowEnterpriseCollectionsOnSingleServer,
         enforceReplicationFactor, isLocalCollection(infos.front()),
-        NameValidator::isSystemName(name), allowSystem);
+        false, allowSystem);
```
jsteemann, line 801: The change from `NameValidator::isSystemName(name)` may be ok here, but I am not 100% clear.
Is this function only called for shards, not collections?

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+        x = result.get();
+        return {};
+      }
+      return res;
+    } else {
+      return f.apply("s" + std::to_string(x.id));
+    }
+  }
+
+  uint64_t id;
```
jsteemann, line 122: I suggest to make this value const, so the id of an existing ShardID object cannot be changed.
This will implicitly delete `operator=`. Not sure if it's worth it, but it prevents all kinds of schabernack.
Alternatively, we can make the id private and make it accessible (read-only) via a member function only.
```suggestion
  uint64_t const id;
```

arangod/Cluster/ClusterMethods.cpp
```diff
@@ -873,7 +886,7 @@ futures::Future<OperationResult> revisionOnCoordinator(
   for (auto const& p : *shards) {
     auto future = network::sendRequestRetry(
         pool, "shard:" + p.first, fuerte::RestVerb::Get,
-        "/_api/collection/" + StringUtils::urlEncode(p.first) + "/revision",
+        "/_api/collection/" + StringUtils::urlEncode(std::string{p.first}) + "/revision",
```
jsteemann, line 889: If we are using ShardIDs here only, we can avoid the URL-encoding, because ShardIDs are always URL-safe.
This is true for this occurrence and all the following ones where we only refer to ShardIDs, not collection names.

arangod/Agency/Supervision.cpp
mchacki: Fixed with 134fb22ab67

arangod/Aql/CollectionAccessingNode.cpp
mchacki: Fixed by: 553ed9e25e7

arangod/Aql/GraphNode.cpp
mchacki: Fixed by 553ed9e25e7

arangod/Cluster/ClusterInfo.cpp
mchacki: Fixed by 553ed9e25e7

arangod/Cluster/Maintenance.cpp
mchacki: Fixed by: 553ed9e25e7

arangod/Cluster/Maintenance.cpp
mchacki: Fixed by 553ed9e25e7

arangod/Cluster/Maintenance.cpp
mchacki: Fixed by 553ed9e25e7

arangod/Cluster/Maintenance.cpp
mchacki: Fixed by 553ed9e25e7

arangod/Cluster/Utils/ShardID.h
mchacki: I will go for the Private variant.
const here forces me to implement the String -> ShardID twice.

arangod/Cluster/Utils/ShardID.h
mchacki: Done

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,161 @@
...
+
+// Make ShardID fmt::formatable
+template<>
+struct fmt::formatter<arangodb::ShardID> {
+  template <typename ParseContext>
+  constexpr auto parse(ParseContext& ctx) { return ctx.begin(); }
+
+  template<typename FormatContext>
+  auto format(arangodb::ShardID const& shardId, FormatContext& ctx) {
+    return format_to(ctx.out(), "{}", "s" + std::to_string(shardId.id));
```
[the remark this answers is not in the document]
mchacki: Changed the format_to statement, can get away without a concat.

arangod/Cluster/Utils/ShardID.h
mchacki: Done

mchacki: Done

arangod/Cluster/Utils/ShardID.h
mchacki: It indeed is!

mchacki: Deleted by: 8da353b6215

The behavior i required this for is actually covered by the std::string operator, there without undefined behaviour.
Added another Test in the Above commit (FlatHashMap) that uses the same codepath. This test passes witu ASAN/UBSAN now, did not pass with the original VPack::Value operator.

arangod/Cluster/Utils/ShardID.h
mchacki: Done

arangod/Cluster/TraverserEngine.cpp
mchacki: Collections.add requires the std::string.
Would like to avoid changing it's signature with this PR as well.
Therefore the copy has to happen anyways.

arangod/Aql/OptimizerRules.cpp
mchacki: I agree it either returns nullopt or a valid shard.
Moved to Assertion

arangod/Cluster/ClusterInfo.cpp
mchacki: I agree using a sorted container here would be better.
But this is also used outside of Cluster Info, probably also the reason why _shards is the only one not captured in a Managed vector.

This is some medium size change i would like to avoid right now.

arangod/Cluster/Utils/ShardID.h
mchacki: Done

arangod/Cluster/Utils/ShardID.h
mchacki: Done

arangod/Cluster/Utils/ShardID.h
mchacki: Done

arangod/Cluster/Utils/ShardID.h
mchacki: To your suggestion:
The other is a string-type. It does not have the id value.

So i either need to turn this into a string.
Or do the string parsing on other, to extract an id.

Probably i will indeed go to string-parsing other.

Will try to get away with string_view comparison only.

mchacki: Done

arangod/Cluster/Utils/ShardID.h
mchacki: Both done

arangod/ClusterEngine/ClusterIndex.cpp
mchacki: Right, done

arangod/Graph/Cache/RefactoredTraverserCache.cpp
mchacki: Good catch, that was unintentional

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
mchacki: No, we do not, removed

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
mchacki: done

arangod/Sharding/ShardDistributionReporter.cpp
mchacki: Done

arangod/Sharding/ShardingInfo.cpp
mchacki: Done

arangod/Sharding/ShardingInfo.cpp
mchacki: Done

arangod/VocBase/Methods/Collections.cpp
mchacki: The method is called "createShard" and takes a new ShardID as input => It can only be called for shards.

A Shard never matches the isSystemName pattern, hence i think this change is correct.

arangod/Cluster/ClusterMethods.cpp
mchacki: All done, they are all safe.

arangod/Agency/Supervision.cpp
```diff
@@ -3596,7 +3602,8 @@ void Supervision::checkUndoLeaderChangeActions() {
         return result;
       }
 
-      auto logId = replication2::LogId::fromString(id);
+      // TODO: Check if this is correct
+      auto logId = replication2::LogId::fromString(id.c_str());
```
jsteemann, line 3606: If `id` is an `std::string`, no need to pass its `c_str()` value here.
```suggestion
      auto logId = replication2::LogId::fromString(id);
```

arangod/Cluster/ClusterInfo.cpp
```diff
@@ -1970,9 +1968,17 @@ void ClusterInfo::loadCurrent() {
 
       for (auto const& shardSlice :
            velocypack::ObjectIterator(collectionSlice.value)) {
-        pmr::ManagedString shardID{shardSlice.key.copyString(),
-                                   _resourceMonitor};
+        auto maybeShardID =
+            ShardID::shardIdFromString(shardSlice.key.stringView());
+        if (ADB_UNLIKELY(maybeShardID.fail())) {
+          TRI_ASSERT(false)
+              << "Indexed malformed shard name " << shardSlice.key.copyString();
```
jsteemann, line 1975:
```suggestion
              << "Indexed malformed shard name " << shardSlice.key.stringView();
```

arangod/Cluster/Utils/ShardID.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "ShardID.h"
+
+#include "Basics/Exceptions.h"
+#include "Basics/StringUtils.h"
+
+#include <absl/strings/str_cat.h>
+
+arangodb::ResultT<arangodb::ShardID> arangodb::ShardID::shardIdFromString(
+    std::string_view s) noexcept {
```
jsteemann, line 31: I would not declare this function `noexcept`, because in theory the creation of the `arangodb::Result` object can throw an exception in `std::make_unique<Error>()`.

arangod/Cluster/Utils/ShardID.cpp
```diff
@@ -0,0 +1,83 @@
...
+arangodb::ShardID arangodb::ShardID::invalidShard() noexcept { return ShardID{0}; }
+arangodb::ShardID::ShardID(std::string_view id) {
+  auto maybeShardID = shardIdFromString(id);
+  if (maybeShardID.fail()) {
+    THROW_ARANGO_EXCEPTION(maybeShardID.result());
+  }
+  *this = maybeShardID.get();
+}
+arangodb::ShardID::operator std::string() const {
+  return absl::StrCat("s", std::to_string(_id));
```
jsteemann, line 51:
```suggestion
  return absl::StrCat("s", _id);
```

arangod/Cluster/Utils/ShardID.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// NOTE: This is not noexcept because of shardIdFromString
+bool arangodb::ShardID::operator==(std::string_view other) const noexcept {
```
jsteemann, line 55: Exactly, it is not noexcept because of the call to `shardIdFromString`. So we should remove the `noexcept` here.

arangod/Cluster/Utils/ShardID.cpp
```diff
@@ -0,0 +1,83 @@
...
+  return false;
+}
+
+bool arangodb::ShardID::isValid() const noexcept {
+  // We can never have ShardID 0. So we use it as invalid value.
+  return _id != 0;
+}
+
+std::ostream& arangodb::operator<<(std::ostream& o,
+                                   const arangodb::ShardID& r) {
```
jsteemann, line 69:
```suggestion
                                   arangodb::ShardID const& r) {
```

arangod/Cluster/Utils/ShardID.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+std::string operator+(const std::string& text, const arangodb::ShardID& s) {
+  return absl::StrCat(text, "s", std::to_string(s.id()));
+}
+std::string operator+(const arangodb::ShardID& s, const std::string& text) {
+  return absl::StrCat("s", std::to_string(s.id()), text);
+}
+
+auto std::hash<arangodb::ShardID>::operator()(
+    const arangodb::ShardID& v) const noexcept -> std::size_t {
```
jsteemann, line 81:
```suggestion
    arangodb::ShardID const& v) const noexcept -> std::size_t {
```

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <fmt/format.h>
+
+#include <cstdint>
+#include <string_view>
+
+namespace arangodb {
+
+struct ShardID {
+  static ResultT<ShardID> shardIdFromString(std::string_view s) noexcept;
```
jsteemann, line 36: As the method can throw, we should rather remove the `noexcept` declaration.

arangod/Cluster/Utils/ShardID.h
```diff
@@ -0,0 +1,126 @@
...
+  template<typename FormatContext>
+  auto format(arangodb::ShardID const& shardId, FormatContext& ctx) {
+    return format_to(ctx.out(), "s{}", std::to_string(shardId.id()));
+  }
+};
+
+// Allow ShardID to be added to std::strings
+std::string operator+(std::string const& text, arangodb::ShardID const& s);
+// Allow ShardID to be added to std::strings
+std::string operator+(arangodb::ShardID const& s, std::string const& text);
```
jsteemann, line 126: See suggestions in the .cpp file.

arangod/Cluster/v8-cluster.cpp
```diff
@@ -1005,7 +1025,8 @@ static void JS_GetResponsibleServersClusterInfo(
   v8::Handle<v8::Object> responsible = v8::Object::New(isolate);
   for (auto const& it : result) {
     responsible
-        ->Set(context, TRI_V8_ASCII_STRING(isolate, it.first.data()),
+        ->Set(context,
+              TRI_V8_ASCII_STRING(isolate, std::string{it.first}.c_str()),
```
jsteemann, line 1029:
```suggestion
              TRI_V8_ASCII_STD_STRING(isolate, std::string{it.first}),
```

arangod/VocBase/LogicalCollection.cpp
```diff
@@ -1282,20 +1278,19 @@ void LogicalCollection::decorateWithInternalValidators() {
...
 std::optional<replication2::LogId> LogicalCollection::tryShardIdToStateId(
-    std::string_view shardId) {
-  if (shardId.empty()) {
+    ShardID const& shardId) {
+  if (!shardId.isValid()) {
     return {};
   }
-  auto stateId = shardId.substr(1, shardId.size() - 1);
-  return replication2::LogId::fromString(stateId);
+  return replication2::LogId::fromString(std::to_string(shardId.id()));
```
jsteemann, line 1293: This looks a bit strange. We are generating a temporary string from the shard number (without the "s"), and then are parsing that temporary string again.
Can't we use `return replication2::LogId(shardId.id());` here?

arangod/Agency/Supervision.cpp
mchacki: Done

arangod/Cluster/ClusterInfo.cpp
mchacki: Done

arangod/Cluster/Utils/ShardID.cpp
mchacki: Done

arangod/Cluster/Utils/ShardID.cpp
mchacki: Done

arangod/Cluster/Utils/ShardID.cpp
mchacki: Done

arangod/Cluster/v8-cluster.cpp
```diff
@@ -957,15 +963,24 @@ static void JS_GetResponsibleServerClusterInfo(
...
+    // Asking for non-shard name pattern.
+    // Compatibility with original API return empty array.
+    v8::Handle<v8::Array> list = v8::Array::New(isolate, 0);
+    TRI_V8_RETURN(list);
+  } else {
+    auto result = ci.getResponsibleServer(maybeShardID.get());
+    v8::Handle<v8::Array> list = v8::Array::New(isolate, (int)result->size());
+    uint32_t count = 0;
+    for (auto const& s : *result) {
+      list->Set(context, count++, TRI_V8_STD_STRING(isolate, s.c_str()))
```
jsteemann, line 978:
```suggestion
      list->Set(context, count++, TRI_V8_STD_STRING(isolate, s))
```

jsteemann (approved):
LGTM

PR #20208, opened 2023-11-24
replace some memory function shims for Windows

jsteemann:
### Scope & Purpose

Remove `memmem` and `memrchr` shims for Windows, and replace the usage of `memmem` and `memrchr` with std::string_view::find/rfind calls.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/Functions.cpp
+++ b/arangod/Aql/Functions.cpp
@@ -2366,14 +2366,15 @@ AqlValue functions::Contains(ExpressionContext* ctx, AstNode const&,
     size_t const searchLength = buffer->length() - valueLength;
 
     if (searchLength > 0) {
-      char const* found = static_cast<char const*>(
-          memmem(buffer->data(), valueLength, buffer->data() + searchOffset,
-                 searchLength));
+      std::string_view haystack(buffer->data(), valueLength);
+      std::string_view needle(buffer->data() + searchOffset, searchLength);
 
-      if (found != nullptr) {
+      size_t found = haystack.find(needle);
+
+      if (found != std::string_view::npos) {
         if (willReturnIndex) {
           // find offset into string
-          int bytePosition = static_cast<int>(found - buffer->data());
+          int bytePosition = static_cast<int>(found);
           char const* p = buffer->data();
           int pos = 0;
           while (pos < bytePosition) {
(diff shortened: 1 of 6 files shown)


arangod/Aql/Functions.cpp
```diff
@@ -2366,14 +2366,15 @@ AqlValue functions::Contains(ExpressionContext* ctx, AstNode const&,
...
 
     if (searchLength > 0) {
-      char const* found = static_cast<char const*>(
-          memmem(buffer->data(), valueLength, buffer->data() + searchOffset,
-                 searchLength));
+      std::string_view haystack(buffer->data(), valueLength);
+      std::string_view needle(buffer->data() + searchOffset, searchLength);
 
-      if (found != nullptr) {
+      size_t found = haystack.find(needle);
```
MBkkt, line 2372: It works for O(n*m), maybe we want to use https://en.cppreference.com/w/cpp/utility/functional/boyer_moore_horspool_searcher?

MBkkt (approved):

arangod/Aql/Functions.cpp
jsteemann: I don't think that `memmem` or `memrchr` make any guarantees about the complexity of the operations. I think they are not even defined in Posix standards, but glibc extensions.
Also, AFAIK C++ does not make any guarantees that find/rfind will have always have the mentioned worst-case complexity, so the STL implementers should be allowed to use optimized versions as well.

Anyway, we can switch to Boyer-Moore-Horspool here, but I am wondering if we should do so in other places as well then? I think we are using find/rfind a lot in our code.

MBkkt: My main point that we know memmem is using two-way and find using naive algorithm.

PR #20211, opened 2023-11-24
[CINFRA] Extra logging in case of failure

apetenchea:
### Scope & Purpose

There's been a few cases (quite rarely) when I noticed something really weird: we try to insert a key into a newly created replication2 collection and fail. The key is definitely unique, there's no way it could have been there before, yet we get a unique constraint violation.
For example:
```
ArangoError: unique constraint violated - in index primary of type primary over '_key'; conflicting key: testDoubleCreateDropIndex-0
```
Inspecting the agency shows nothing wrong (I would've expected to see at least some failover).

**This PR adds extra logging information that should aid in investigating such failures.**

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM. we shall get more info this way :+1:

PR #20242, opened 2023-11-28
Bug fix/cinfra fix test follower code read api tests

mchacki:
### Scope & Purpose

*
Fixes:

    [FAILED]  tests/js/client/shell/api/multi/document-internal-http-return-codes-cluster.js

      "testFollowerCodeRead" failed: Error: at assertion #1: assertEqual: (200) is not equal to (421)
(Error

This happens only on Replication2 databases for now.

Background: The dirty-read check if a Shard is a Follower was Replication1 specific.
This is now replaced with a generic variant that works in both versions.
*

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/VocBase/LogicalCollection.cpp
```diff
@@ -1298,7 +1298,28 @@ std::optional<replication2::LogId> LogicalCollection::tryShardIdToStateId(
...
+    // Stubs are no shards, they cannot be leaders
+    return false;
+  }
+  if (replicationVersion() == replication::Version::ONE) {
+    // Replication version 1 does not have leaders
+    auto const& followerInfo = followers();
+    // If the Leader is empty, we are the leader
+    return followerInfo->getLeader().empty();
+  } else {
+    auto const& docState = getDocumentState();
```
apetenchea, line 1316: This method works. I want to point out that `getDocumentState()` may throw if the replicated log is not there. I doubt this is possible with the current use-case, as the transaction is already there.

A way to make sure we would never crash is to wrap this into a `catchToResultT`. You may use `LogicalCollection::getDocumentStateLeader()` for that, which should always result in `ok` if the leader is there.

apetenchea (approved):
LGTM! 

My comment is just an optional remark, as your code is already correct.

Another observation to bear in mind is that looking up the state acquires the `VocBaseLogManager::_guardedData` lock. I don't see any problem with that now, but I'm pointing this out in case we start seeing flaky deadlocks in the near future.

PR #20245, opened 2023-11-29
[CINFRA] Database not found hanger

apetenchea:
### Scope & Purpose

This PR fixes a rare hanger in JS tests.

The problem occurred in a [CircleCI Windows Run](https://app.circleci.com/pipelines/github/arangodb/arangodb/19700/workflows/44067109-eb00-4f8a-b2bb-2b088716c7b8/jobs/990219).

**Description**
The `testCheckPingLog` from `replicatedLogEntrySuite` aims to test the _ping_ functionality of replication2. It creates a replicated log, inserts an entry, does some checks, then deletes the replicated log from Target. Since it is the last test from its suite, `tearDownAll` is triggered immediately and the database is dropped. This causes a race between `DropDatabase` and `UpdateReplicatedLog`. If the `UpdateReplicatedLog` maintenance action gets executed, but in the meantime `DropDatabase` makes the DB unavailable, the former will keep throwing exceptions and re-schedule itself indefinitely.
**Note that removing a replicated log from Target is not something a user would ever do - this is a stunt we make for testing purposes.** In production, when dropping a database, all collections are being dropped first, along with their subsequent replicated logs. However, when the coordinator drops a database, it may create a similar situation, because it removes the database entry from Plan along with the replicated log entries from Target and Plan. But note that in this particular situation, the `UpdateReplicatedLog` action is not going to be triggered, because the database is no longer in Plan.

**Solution**
I'm wrapping the main logic (including the DatabaseGuard) inside a `catchToResult`, such that potential exceptions won't cause this problem again.
In order to prevent maintenance actions from being re-scheduled indefinitely when they throw, I'm adding the `nextState(false)` line to the exception handling code. This causes the `_loopState` to switch to `eFIND_ACTION` after such failures.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM!

mchacki:
Red test is the known unwind bug.

PR #20246, opened 2023-11-29
Fix K_SHORTEST_PATHS

markuspf:
### Scope & Purpose

This PR attempts to be a small solution to the problem that presents itself with path searches and global conditions on vertices and edges.

The code that executes path/vertex conditions while enumerating paths was as yet untested and hence crashed on the first use.

This is due to the fact that path validation, inspecting the vertex documents on the path, was called at a time when the vertex document wasn't known yet.

This PR moves the full path validation that inspects a vertex document to a point where the vertex document is known.

This is after the vertex has been popped from the (priority-)queue, but retains redundancy checks on expansion, that is if a vertex-id has been seen already it is not put on the queue again for the enumerator without weights.

The weighted two sided enumerator is simplified a little bit.

### DIFF
--- a/tests/Graph/WeightedShortestPathTest.cpp
+++ b/tests/Graph/WeightedShortestPathTest.cpp
@@ -433,7 +433,7 @@ TEST_P(WeightedShortestPathTest, shortest_path_A_F_outbound) {
     aql::TraversalStats stats = finder.stealStats();
     // We have to lookup the vertex
     // 4x vertices, 3x edges
-    EXPECT_EQ(stats.getScannedIndex(), 25U);
+    EXPECT_EQ(stats.getScannedIndex(), 26U);
   }
 
   {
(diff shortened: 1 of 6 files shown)


tests/Graph/WeightedShortestPathTest.cpp
```diff
@@ -433,7 +433,7 @@ TEST_P(WeightedShortestPathTest, shortest_path_A_F_outbound) {
     aql::TraversalStats stats = finder.stealStats();
     // We have to lookup the vertex
     // 4x vertices, 3x edges
```
hkernbach, line 435: minor, could just get rid of those already outdated comments here

tests/Graph/WeightedShortestPathTest.cpp
```diff
@@ -482,7 +482,7 @@ TEST_P(WeightedShortestPathTest, shortest_path_A_F_inbound) {
...
```
hkernbach, line 484: minor, could just get rid of those already outdated comments here

hkernbach (approved):
LGTM

PR #20250, opened 2023-11-30
FE-397:  Query UI errors with multiple queries & explain/profile 

palashkaria:
### Scope & Purpose

Fixes a couple of issues

- Query profile/explain overrides the last results
- When multiple queries are triggered, results don't show up

These got fixed by making `asyncJobId` required, & by removing `index` from the sync job useEffect deps

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] Manually tested

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/FE-397


js/apps/system/_admin/aardvark/APP/react/src/views/query/queryResults/useSyncQueryExecuteJob.tsx
```diff
@@ -153,5 +152,5 @@ export const useSyncQueryExecuteJob = ({
     };
     // disabled because these functions don't need to be in deps
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [asyncJobId, index]);
```
palashkaria, line 156: this fixed it because asyncJobId is stable, index keeps changing when we execute new queries

nikhil-varma:
I noticed a peculiar behaviour
- Write a erroneous query
- Click on Explain button
- It shows the Profile section in the result area.

It shows both the toast and the the Profile section in the result area. Is that expected?
image

nikhil-varma (approved):

nikhil-varma (approved):

PR #20261, opened 2023-12-01
Set build-docker-image default value in base config

mpoeter:
### Scope & Purpose

If a pipeline parameter is set explicitly, this value will implicitly be forwarded to the continuation pipeline. If a value is not explicitly set. but has a default value, then this value is _not_ implicitly forwarded. It can be explicitly forwarded by using the continuation's parameters, but if we do that unconditionally, this causes a problem once the parameter is again set explicitly, because then we get a "Conflicting pipeline parameters." error because the parameter is forwarded twice (once implicitly and once explicitly).

### DIFF
--- a/.circleci/base_config.yml
+++ b/.circleci/base_config.yml
@@ -9,7 +9,7 @@ parameters:
     default: ""
   build-docker-image:
     type: string
-    default: ""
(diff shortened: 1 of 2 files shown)


mchacki (approved):
LGTM

mpoeter:
Not waiting for the full CI to finish, because this fix affects only the dynamic config generation.

PR #20264, opened 2023-12-01
query optimization for queries using `LIKE`

jsteemann:
### Scope & Purpose

* Optimize equality lookups and prefix matches that use the AQL like function. The optimization will turn `LIKE(doc.attribute, 'fixed string')` into an equality lookup, and `LIKE(doc.attribute, 'prefix%')` into a range query.
  That allows indexes to be used in case they are present on the underlying attribute.
  The optimization is applied only if the second parameter to the `LIKE` function is a fixed string and if third parameter to the `LIKE` function is either not set or set to a value of `false`.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* Optimize equality lookups and prefix matches that use the AQL like function.
+  The optimization will turn `LIKE(doc.attribute, 'fixed string')` into an
+  equality lookup, and `LIKE(doc.attribute, 'prefix%')` into a range query.
+  That allows indexes to be used in case they are present on the underlying
+  attribute.
+  The optimization is applied only if the third parameter to the `LIKE`
+  function is either not set or set to a value of `false`.
```
```suggestion
  The optimization is applied only if the second parameter to the `LIKE`
  function is a fixed string and if third parameter to the `LIKE` function 
  is either not set or set to a value of `false`.
```

PR #20268, opened 2023-12-04
[CINFRA] Adapting server-parameters tests

apetenchea:
### Scope & Purpose

This PR is part of our efforts to run the tests using replication2 as default. The following tests are being changed:

_max-collections-per-query-cluster.js_
The `allShardsOnSameServer` failure point is moved into common code (EventDistribution.cpp). The place of invocation depends on the replication version: `PlanCollectionToAgencyWriter` (replication1, on coordinators), `CollectionGroupSupervision` (replication2, on agents). 

_test-failed-write-concern-cluster-403.js_
_test-failed-write-concern-cluster-503.js_
Normally, in replication2, not meeting the write concern threshold blocks the insert, because the leader cannot commit. I've added checks in `transaction::Methods::determineReplication2TypeAndFollowers` that determine whether we can use the log, based on the participants health. As a consequence, a test from `shell-transaction-replication2-recovery.js` had to be adapted accordingly.

_auto-refill-index-caches-on-followers-on-cluster.js_
Some tests were failing because replication2 ignored any specific options for the `RefillIndexCaches` optimization. In this PR, we're adding support for that, allowing the leader to specify a custom setting during replication.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


arangod/Transaction/Methods.cpp
```diff
@@ -2409,9 +2410,72 @@ Result transaction::Methods::determineReplication2TypeAndFollowers(
...
+    if (canReachWriteConcern.fail()) {
+      LOG_TOPIC("bdce1", DEBUG, Logger::REPLICATED_STATE)
+          << "Cannot reach write concern for transaction " << tid()
+          << ", concerning shard " << collection.name() << ": "
+          << canReachWriteConcern.result();
+    } else if (*canReachWriteConcern) {
+      return {};
+    }
+
+    if (clusterFeature.statusCodeFailedWriteConcern() == 403) {
```
mchacki, line 2475: Should be the according error static value

mchacki (changes requested):
Minor change, should replace a number with matching static value.

Code looks good to me

arangod/Transaction/Methods.cpp
mchacki: Nevermind, the config in the value is hard-coded to 403 and the error int type is actually different, let's keep it like this.

mchacki (approved):
LGTM

PR #20277, opened 2023-12-06
CINFRA: BugFix recovery of views under replication2

mchacki:
### Scope & Purpose

*Recovery tests on Views are red in Replication2. We found the issue that InvertedIndexes cannot always repeat an operation (replace document with itself) which is what Replication2 Protocol relies on. This all raced with what is still to be recovered from replicated log and what is recovered from rocksdb, and sometimes Replication2 recovers a small amount of operations multiple times.*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
+++ b/tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
@@ -1,0 +1,412 @@
+/* jshint globalstrict:false, strict:false, maxlen: 200 */
+/* global assertEqual, assertIdentical */
+
+// //////////////////////////////////////////////////////////////////////////////
+// / @brief Arango View Recovery Tests
+// /
+// /
+// / DISCLAIMER
+// /
+// / Copyright 2023 ArangoDB GmbH, Cologne, Germany
+// /
+// / Licensed under the Apache License, Version 2.0 (the "License")
+// / you may not use this file except in compliance with the License.
+// / You may obtain a copy of the License at
+// /
+// /     http://www.apache.org/licenses/LICENSE-2.0
+// /
+// / Unless required by applicable law or agreed to in writing, software
+// / distributed under the License is distributed on an "AS IS" BASIS,
+// / WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// / See the License for the specific language governing permissions and
+// / limitations under the License.
+// /
+// / Copyright holder is ArangoDB GmbH, Cologne, Germany
+// /
+// / @author Michael Hackstein
+// //////////////////////////////////////////////////////////////////////////////
+
+const jsunity = require('jsunity');
+const db = require('@arangodb').db;
+const lh = require('@arangodb/testutils/replicated-logs-helper');
+const rh = require('@arangodb/testutils/restart-helper');
+const {getCtrlDBServers} = require('@arangodb/test-helper');
+
+
+const testSuite = (config) => {
+  const documentCount = 100;
+
+  const restartAndWaitForRebootId = function (servers) {
+    const rebootIds = servers.map(s => [s.id, lh.getServerRebootId(s.id)]);
+    rh.restartServers(servers);
+    lh.waitFor(function () {
+      for (const [server, rid] of rebootIds) {
+        const current = lh.getServerRebootId(server);
+        if (rid > current) {
+          return Error(`server ${server} reboot id is ${current}, expected > ${rid}`);
+        }
+      }
+    });
+  };
+
+  const disableMaintenanceMode = function () {
+    const response = db._connection.PUT('/_admin/cluster/maintenance', '"off"');
+    assertIdentical(false, response.error);
+    assertIdentical(200, response.code);
+    if (response.hasOwnProperty('warning')) {
+      console.warn(response.warning);
+    }
+  };
+
+  const enableMaintenanceMode = function () {
+    const response = db._connection.PUT('/_admin/cluster/maintenance', '"on"');
+    assertIdentical(false, response.error);
+    assertIdentical(200, response.code);
+    if (response.hasOwnProperty('warning')) {
+      console.warn(response.warning);
+    }
+  };
+
+  const restartServers = () => {
+    const dbServers = getCtrlDBServers();
+
+    enableMaintenanceMode();
+    rh.shutdownServers(dbServers);
+
+    restartAndWaitForRebootId(dbServers);
+    disableMaintenanceMode();
+  };
+
+
+  let collection;
+  let shardsToLogs;
+  let shards;
+  let logs;
+
+  const {setupCollection, dropCollection, genDoc, namePostfix} = config;
+
+  const dbn = 'UnitTestsDatabase';
+  const viewName = 'UnitTestView';
+  const {
+    setUpAll,
+    tearDownAll,
+    setUpAnd,
+    tearDownAnd,
+  } = lh.testHelperFunctions(dbn, {replicationVersion: '2'});
+
+  const compactLogs = () => {
+    for (const l of logs) {
+      l.compact();
+    }
+  };
+
+  const insertDummyData = (number) => {
+    // NOTE: This is on purpose single calls.
+    // We need those to be different entries in the replicated log
+    // Otherwise we would only have a single entry for the batch
+    for (let i = 0; i < number; ++i) {
+      collection.save(genDoc());
+    }
+  };
+
+  const insertDummyUpdates = (number, updates) => {
+    // NOTE: This is on purpose single calls.
+    // We need those to be different entries in the replicated log
+    // Otherwise we would only have a single entry for the batch
+    const docs = [];
+    for (let i = 0; i < number; ++i) {
+      const doc = genDoc();
+      const {_key} = collection.save(doc);
+      docs.push({doc, _key});
+    }
+    for (let i = 0; i < updates; ++i) {
+      for (const {_key, doc} of docs) {
+        collection.update(_key, doc);
+      }
+    }
+  };
+
+  const insertDummyReplaces = (number, updates) => {
+    // NOTE: This is on purpose single calls.
+    // We need those to be different entries in the replicated log
+    // Otherwise we would only have a single entry for the batch
+    const docs = [];
+    for (let i = 0; i < number; ++i) {
+      const doc = genDoc();
+      const {_key} = collection.save(doc);
+      docs.push({doc, _key});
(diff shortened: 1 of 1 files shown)


tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,545 @@
...
+    const result = db._query(`FOR doc IN ${viewName} SEARCH doc.c >= 0 OPTIONS {waitForSync: true} RETURN doc._key`).toArray();
+    const expectedResult = db._query(`FOR doc IN ${collection.name()} FILTER doc.c >= 0 RETURN doc._key`).toArray();
+
+    const uniqView = new Set(result);
+    const uniqCol = new Set(expectedResult);
+    if (result.length !== expectedResult.length) {
+      {
+        const dh = require("@arangodb/testutils/document-state-helper");
+        let intermediateCommitEntries = dh.getDocumentEntries(dh.mergeLogs(logs), "Insert");
+        require("console").error(JSON.stringify(intermediateCommitEntries));
```
mchacki, line 206: Note to Reviewer:
The console output here is for more precise error reporting.
Therefore I would like to keep it

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
```diff
@@ -215,4 +219,27 @@ auto DocumentStateShardHandler::lockShard(ShardID const& shard,
...
+auto DocumentStateShardHandler::prepareShardsForLogReplay() noexcept -> void {
+  for (auto const& shard : getAvailableShards()) {
+    // The inverted indexes cannot work with duplicate LocalDocumentIDs within the same
+    // commit interval. They however can if we have a commit in between the two.
+    // If we replay one log we know there can never be a duplicate LocalDocumentID.
+    for (auto const& index : shard->getIndexes()) {
+      {
+        auto maybeInvertedIndex =
+            std::dynamic_pointer_cast<iresearch::IResearchInvertedIndex>(index);
+        if (maybeInvertedIndex) {
```
MBkkt, line 231: ArangoDB indexes has type

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
```diff
@@ -63,7 +63,10 @@ DocumentFollowerState::DocumentFollowerState(
           handlersFactory->createShardHandler(core->getVocbase(), core->gid)),
       _errorHandler(handlersFactory->createErrorHandler(core->gid)),
       _guardedData(std::move(core), handlersFactory, loggerContext,
-                   _errorHandler) {}
+                   _errorHandler) {
+  // Get ready to replay the log
+  _shardHandler->prepareShardsForLogReplay();
```
Dronplane, line 68: Does it makes sense to call it here as I see that  this is anyway called after "acquiring" snapshot?

mchacki: Yes, this is actually necessary.

This code place is hit whenever there is a term change (e.g. a LeaderElection / Failover).
As long as followers are in-sync (or close) we will not enter the Snapshot code-path.

The Snapshot code-path is only for new and Empty servers.
Or for servers that runter after a long period of time.

It is also required after the Snapshot.
Because there we first see this constructor.
Then we do the Snapshot.
And without calling this constructor again, we start applying log.
The log may overlap with data from Snapshot.

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
mchacki: Added ifs for the type.

arangod/Replication2/StateMachines/Document/DocumentFollowerState.cpp
Dronplane: I see. Thx

arangod/Replication2/StateMachines/Document/DocumentStateShardHandler.cpp
Dronplane: Then we don't need to check cast result. More to say we have a more robust machinery to do this: 

``` cpp
    if (index->type() == Index::TRI_IDX_TYPE_INVERTED_INDEX) {
       basics::downCast<IResearchRocksDBInvertedIndex>(*index).commit(true);
    } else if (index->type() == Index::TRI_IDX_TYPE_IRESEARCH_LINK) {
       basics::downCast<IResearchRocksDBLink>(*index).commit(true);
    }
```
It does the same and has all necessary asserts inside.

mchacki: Perfect, exactly what I was looking for :+1:

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,542 @@
...
+
+  const assertViewInSync = () => {
+    const result = db._query(`FOR doc IN ${viewName} SEARCH doc.c >= 0 OPTIONS {waitForSync: true} RETURN doc._key`).toArray();
+    const expectedResult = db._query(`FOR doc IN ${collection.name()} FILTER doc.c >= 0 RETURN doc._key`).toArray();
+
+    const uniqView = new Set(result);
+    const uniqCol = new Set(expectedResult);
+    if (result.length !== expectedResult.length) {
+      {
+        const dh = require("@arangodb/testutils/document-state-helper");
```
apetenchea: `dh` is already imported

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,542 @@
...
+        let intermediateCommitEntries = dh.getDocumentEntries(dh.mergeLogs(logs), "Insert");
```
apetenchea, line 166: If you're looking for intermediate commit entries, you should pass the "IntermediateCommit" string instead of "Insert"

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,542 @@
...
+
+    /*
+    * Recover the view if the log is contains entries
+    */
+    [`testRecoverInsertsInLog_${namePostfix}`]() {
+      compactLogs();
+      insertDummyData(documentCount);
+      lh.bumpTermOfLogsAndWaitForConfirmation(dbn, collection);
+      assertViewInSync();
+    },
```
apetenchea: Since `compactLogs()` doesn't really do anything (it's an empty for loop), I would argue that these two tests are the same 🤔

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
```diff
@@ -64,7 +64,10 @@ DocumentLeaderState::DocumentLeaderState(
           _handlersFactory->createSnapshotHandler(core->getVocbase(), gid)),
       _errorHandler(_handlersFactory->createErrorHandler(gid)),
       _guardedData(std::move(core), _handlersFactory),
-      _transactionManager(transactionManager) {}
+      _transactionManager(transactionManager) {
+  // Get ready to replay the log
+  _shardHandler->prepareShardsForLogReplay();
```
apetenchea, line 69: NIT
This can be moved to `DocumentLeaderState::recoverEntries()` - that's the place where the leader may replay the log (unlike the follower). I think that might be a bit cleaner, but surely, the constructor must happen before that, so feel free to keep it here if you it better fits your needs.

apetenchea (changes requested):
C++ part looks good.
I left a couple of remarks on the JS test. Nothing major, but we should clarify before merging.

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,542 @@
...
+      const commitIndexAfterCompaction = status.participants[leader].response.local.commitIndex + 1;
+
+
+      let logContents = log.head(1000);
+
+      // We need to make sure that the CreateShard entries are being compacted before doing recovery.
+      // Otherwise, the recovery procedure will try to create the shards again. The point is to test how
+      // the recovery procedure handles operations referring to non-existing shards.
+      // By waiting for lowestIndexToKeep to be greater than the current commitIndex, and for the index to be released,
+      // we make sure the CreateShard operation is going to be compacted.
```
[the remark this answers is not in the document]
mchacki: Removed the compact all together

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
mchacki: Fixed

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
mchacki: forgot to rename the variable
i actualy want to see InsertEntries

tests/js/client/restart/test-restart-replication2-view-recovery-cluster.js
```diff
@@ -0,0 +1,542 @@
...
+    * Recover the view on leader change
+    */
+    [`testRecoverInsertsInLogSwitchLeader_${namePostfix}`]() {
+      compactLogs();
+      insertDummyData(documentCount);
+      // Try to switch the leader
+      for (const l of logs) {
+        lh.unsetLeader(dbn, l.id());
+      }
+      lh.bumpTermOfLogsAndWaitForConfirmation(dbn, collection);
```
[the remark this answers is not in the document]
mchacki: It's not, both replay the log anyhow.

arangod/Replication2/StateMachines/Document/DocumentLeaderState.cpp
mchacki: I would like to keep it in the Constructor.
Just to have little difference between leader and follower and to make it clear for anyone looking at this in the future that this operation is required for both, and required round and about the same time.

mchacki:
Red test is unrelated to change.

PR #20281, opened 2023-12-06
adjust V8 integration code to work with old and new V8 versions

jsteemann:
### Scope & Purpose

Adjust our V8 integration code to work with old and new V8 versions.
The code currently uses an `#ifdef V8_UPGRADE` to differentiate between the old and the new V8 version.
The `V8_UPGRADE` define is currently not set, so the old integration code will still be used.
However, when we move to a newer version of V8, we can easily find the few code places we will need to adjust for the new V8 version and remove the `#else` cases there.
The goal of this PR is to merge back part of the necessary work for upgrading as quickly as possible, so that we can avoid having a single very large upgrade PR.

Note that in this PR a few files/class have been renamed to avoid name confusion between v8's `v8::Context` and ArangoDB's `V8Context`. It was hard to talk about these things without causing confusion.
To avoid confusion in the future, our internal `V8Context` is now renamed to `V8Executor` in the code, and all mentions of "contexts" have been renamed to "executors" where it makes sense. Metrics and API output that mentioned "v8 contexts" has intentionally not been modified to keep metrics and APIs stable.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


lib/V8/JSLoader.cpp
```diff
@@ -73,32 +62,40 @@ JSLoader::eState JSLoader::loadScript(v8::Isolate* isolate,
...
-  // Enter the newly created execution environment.
-  v8::Context::Scope context_scope(context);
+#if 0
+  v8::Handle<v8::Context> context = isolate->GetCurrentContext();
+#endif
+  {
+#if 0
+    // Enter the newly created execution environment.
+    v8::Context::Scope context_scope(context);
+#endif
```
markuspf, line 72: can this just go?

lib/V8/v8-globals.cpp
```diff
@@ -217,30 +217,30 @@ TRI_v8_global_t::SharedPtrPersistent::emplace(
 
 TRI_v8_global_t::~TRI_v8_global_t() = default;
 
-v8::Local<v8::Object> TRI_GetObject(v8::Local<v8::Context>& context,
+v8::Local<v8::Object> TRI_GetObject(v8::Local<v8::Context> context,
```
markuspf, line 220: This looks weird; is the pass by reference a historic thing?

lib/V8/JSLoader.cpp
jsteemann: Yes!

lib/V8/v8-globals.cpp
jsteemann: Normally it is good to pass things by reference to avoid copies.
All the v8 objects however can (and should) be passed by value, as they are just nice wrappers around pointers.
So moving from by-reference to by-value here was intentional and should not make things any worse.

lib/V8/JSLoader.cpp
jsteemann: Removed

PR #20289, opened 2023-12-08
Invalid Shard Ids

maierlars:
### Scope & Purpose

Test uses invalid shard ids that can trigger an assertion in `loadCurrent`.


jsteemann (approved):
LGTM

PR #20291, opened 2023-12-08
[CINFRA] Replicated Log and Shard bound together

apetenchea:
### Scope & Purpose

We've noted a few crashes in the `isReplication2Leader()` method, due to the replicated log being dropped, while the corresponding shard was still there.

_Drop the replicated log and corresponding shards together_
Before, while dropping the vocbase, we had to use `shutdownReplicatedLogs()`. Only after that, we dropped collections. This leaves a small time window where the collections can still be found locally, but the replicated logs are gone. To fix this problem, I introduced the `prepareDropAll()` method, which cleans up the corresponding resources of all replicated states, similar to how `dropReplicatedState()` would do.

_Add preconditions while reporting to Current_
We're taking the replication version from local databases. This is safer, because previously, we used to take it from Plan, but fallback on replication ONE if database was not to be found in Plan.
Also, note that in order to report a shard in Current, we're adding a precondition that the server is indeed the leader of the corresponding replicated log in Plan.

_Do not crash in production_
I am replacing the assertion in `isReplication2Leader()` with a regular `TRI_ASSERT`, because, while the covered scenario is still problematic, I don't see a good enough reason to crash. Emitting a warning and returning `false` would lead the maintenance to believe that the current server is not the leader, which in the worst case does nothing (believes it is a follower) and best case is correct (it is indeed a follower).

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/Cluster/Maintenance.cpp
+++ b/arangod/Cluster/Maintenance.cpp
@@ -268,16 +268,19 @@ static std::string CreateLeaderString(std::string const& leaderId,
 
 static bool isReplication2Leader(ShardID const& shname,
                                  ReplicatedLogStatusMap const& localLogs,
-                                 ShardIdToLogIdMap const& shardsToLogs) {
-  auto it = shardsToLogs.find(shname);
+                                 ShardIdToLogIdMap const& localShardsToLogs) {
+  auto it = localShardsToLogs.find(shname);
   // If the shard is not found, then it has not been created yet.
-  if (it != std::end(shardsToLogs)) {
+  if (it != std::end(localShardsToLogs)) {
     auto logId = it->second;
     auto logStatus = localLogs.find(logId);
-    ADB_PROD_ASSERT(logStatus != std::end(localLogs))
-        << "Log " << logId
-        << " not found. If the shard is found, then the log must be also "
-           "there.";
+    if (logStatus == std::end(localLogs)) {
+      LOG_TOPIC("c1d8d", WARN, Logger::MAINTENANCE)
+          << "Replicated log " << logId << " not found. Since the shard "
+          << shname << " exists, so must the log";
+      TRI_ASSERT(false) << logId << " " << shname << " " << localShardsToLogs;
+      return false;
+    }
     return logStatus->second.status.role ==
            replication2::replicated_log::ParticipantRole::kLeader;
   }
@@ -732,7 +735,7 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
     std::vector<std::shared_ptr<ActionDescription>>& actions,
     MaintenanceFeature::ShardActionMap const& shardActionMap,
     ReplicatedLogStatusMapByDatabase const& localLogsByDatabase,
-    ShardIdToLogIdMapByDatabase const& shardIdToLogId) {
+    ShardIdToLogIdMapByDatabase const& localShardIdToLogId) {
   // You are entering the functional sector.
   // Vous entrez dans le secteur fonctionel.
   // Sie betreten den funktionalen Sektor.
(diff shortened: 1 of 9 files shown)


arangod/VocBase/VocBaseLogManager.cpp
```diff
@@ -121,46 +121,44 @@ auto VocBaseLogManager::updateReplicatedState(
   }
 }
 
+void VocBaseLogManager::prepareDropAll() noexcept {
+  auto guard = _guardedData.getLockedGuard();
+  guard->resignAllWasCalled = true;
```
mchacki, line 126: do we need an (if (guard->resignAllWasCalled == true) { return; }) 
Before this line?
To protect against 2 Threads triggering DropAll() ?
Then one will wait on the other, but both will actually excecute below calls.

arangod/VocBase/VocBaseLogManager.cpp
```diff
@@ -186,6 +184,42 @@ auto VocBaseLogManager::dropReplicatedState(arangodb::replication2::LogId id)
...
+  auto metadata = storage->readMetadata();
+  if (metadata.fail()) {
+    THROW_ARANGO_EXCEPTION(
+        std::move(metadata).result());  // state untouched after this
+  }
+
+  // Invalidate the snapshot in persistent storage.
+  metadata->snapshot.updateStatus(
+      replication2::replicated_state::SnapshotStatus::kInvalidated);
+  // TODO make sure other methods working on the state, probably meaning
```
mchacki, line 208: TODO still open?
I have seen this is just moved from old code

arangod/Cluster/Maintenance.cpp
```diff
@@ -268,16 +268,19 @@ static std::string CreateLeaderString(std::string const& leaderId,
...
-    ADB_PROD_ASSERT(logStatus != std::end(localLogs))
-        << "Log " << logId
-        << " not found. If the shard is found, then the log must be also "
-           "there.";
+    if (logStatus == std::end(localLogs)) {
+      LOG_TOPIC("c1d8d", WARN, Logger::MAINTENANCE)
+          << "Replicated log " << logId << " not found. Since the shard "
+          << shname << " exists, so must the log";
+      TRI_ASSERT(false) << logId << " " << shname << " " << localShardsToLogs;
+      return false;
```
mchacki, line 282: Like it

arangod/Cluster/Maintenance.cpp
```diff
@@ -752,17 +755,25 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
     auto const& pdb = pb.get(std::vector<std::string>{AgencyCommHelper::path(),
                                                       PLAN, DATABASES, dbname});
 
+    if (!pdb.isObject()) {
```
mchacki, line 758: Have you seen this?
Or "just" security?

:+1: for the change

arangod/Cluster/Maintenance.cpp
```diff
@@ -752,17 +755,25 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
...
-      // if the "replicationVersion" field is missing this has to be an old
-      // DB which defaults to version ONE.
+      // If the "replicationVersion" field is missing this has to be an old
+      // DB which defaults to version ONE. Note that while processing shards,
+      // we'll take decisions based on the replication version.
       replicationVersion.emplace(dbname, replication::Version::ONE);
     }
 
-    if (pdb.isObject() && local.find(dbname) == local.end()) {
+    if (local.find(dbname) == local.end()) {
```
mchacki, line 776: Ah i See you moved the object piece upwards.
Good change.

arangod/VocBase/VocBaseLogManager.cpp
apetenchea: Yes, it is from old code, was already there. It is still open, although not critical. I think we should keep it.

arangod/VocBase/VocBaseLogManager.cpp
apetenchea: Although it wouldn't hurt, that's not necessary. A second `prepareDropAll()` would just find and empty `statesAndLogs` map and basically do nothing.

PR #20305, opened 2023-12-12
[BTS-1636] Fix path found by SHORTEST_PATH with weight is not path of lowest weight

markuspf:
### Scope & Purpose

When searching for (in particular weighted) shortest paths, ArangoDB searches forward from the `start` and backward from the `target` vertex with something that looks quite a lot like a Dijkstra search.

An important aspect is when a search is finished, that is when a lowest weight path has been found.  Whenever a vertex is relaxed, every newly discovered vertex is checked whether it has been seen on the other side of the search. If it has been *a* path has been found, but there is no guarantee that this is the lowest weight one.

The search has to continue until the sum of the diameters, the weight of the highest *proven* weight paths, exceeds the weight of the best candidate found.

This patch implements this terminating condition.

In addition, since the two sided enumerator supports enumerating all paths between two vertices in increasing order by weight, by extension all weights below the sum of the diameters threshold are deemed to be optimal and hence all paths with weight below the threshold can be returned.

In addition, some simplifications are made
 * a (supposed) optimisation that complicated search handling is removed.
 * when the source vertex is equal to the target vertex, it is only checked whether that vertex fullfils a possible global vertex condition; no searches are performed.


arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.cpp
```diff
@@ -824,14 +838,6 @@ typename WeightedTwoSidedEnumerator<QueueType, PathStoreType, ProviderType,
...
-    _leftInitialFetch = true;
-    return BallSearchLocation::LEFT;
-  } else if (!_rightInitialFetch) {
-    _rightInitialFetch = true;
-    return BallSearchLocation::RIGHT;
-  }
-
   if (!_left.isQueueEmpty() && !_right.isQueueEmpty()) {
     auto leftStep = _left.peekQueue();
     auto rightStep = _right.peekQueue();
```
jsteemann, line 843: nit: `peekQueue()` returns a reference, so these lines will copy the `Step` objects.
I don't think that the copies are expensive here (looks like `Step` is a 24 byte object in most cases), but as `Step` is a template parameter, we may at some point have `Step`s which are more expensive to copy.
In this case it may be better to change `auto leftStep` and `auto rightStep` to `auto const& ...`.

arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.cpp
```diff
@@ -824,14 +838,6 @@ typename WeightedTwoSidedEnumerator<QueueType, PathStoreType, ProviderType,
                                     PathValidator>::BallSearchLocation
 WeightedTwoSidedEnumerator<QueueType, PathStoreType, ProviderType,
                            PathValidator>::getBallToContinueSearch() {
-  if (!_leftInitialFetch) {
```
jsteemann, line 827: nit: it also looks like that the `getBallToContinueSearch()` method doesn't mutate the object and thus could be made `const`.

arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.cpp
```diff
@@ -621,50 +658,49 @@ void WeightedTwoSidedEnumerator<QueueType, PathStoreType, ProviderType,
...
+      // than the sum of the two diameters are valid shortest paths that must
+      // be returned.
+      // K-SHORTEST-PATH search has to continue until the queues on both
+      // sides are empty
+
+      auto leftDiameter = _left.getDiameter();
+      auto rightDiameter = _right.getDiameter();
+      auto sumDiameter = leftDiameter + rightDiameter;
+
+      if (sumDiameter >= bestWeight) {
```
jsteemann, line 701: Is this safe w.r.t. floating-point inaccuracies?

jsteemann (approved):
LGTM for everything but the change to the algorithm. The algorithmic change I cannot comment on.

arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.cpp
markuspf: I am trying to understand what the potential problems are that could pop up.

If you are referring to the `>=` part of this inequality, I don't think it will make a difference to put a `>` there; sure, one can construct cases in which one then searches a really long time to terminate instead of getting an answer quickly but you'd have to put effort in to do it. 

Further,

* There cannot be (obvious) catastrophic cancellation, as we sum up and compare non-negative numbers 
* Weights along paths are added up and as such, in particular if the magnitude of those weights varies, there'll be loss of precision. This is in particular also a problem for the diameter. There are ways of addressing this, none of which would fit into this PR.
* The diameters are of different magnitude and the diameter of one of the balls is rounded away; This would indeed make sumDiameter smaller than it should be this only makes a difference if this "rounding error" leads to early termination and potentially an incorrect result. 

I will have to do a slightly more careful numerical analysis to see whether there are any more pitfalls; There is certainly a small pandora's box that could be opened with respect to floating points and weighted path search;

In all, though, this will not introduce worse behaviour than before!

maierlars (approved):

PR #20306, opened 2023-12-13
Fix jemalloc cmake setup for building with glibc images

neunhoef:
It turns out that we have to give the include path for jemalloc to
find our own libunwind in the CPPFLAGS rather than the CFLAGS. Otherwise
it can happen that the configure/make system cannot compute the
dependencies correctly. This problem hits if there is no system wide
libunwind installed.

- Minor fix for jemalloc build setup.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

### DIFF
--- a/3rdParty/jemalloc/CMakeLists.txt
+++ b/3rdParty/jemalloc/CMakeLists.txt
@@ -35,7 +35,8 @@ if (LINUX OR DARWIN)
   if (USE_JEMALLOC_PROF)
       if (USE_LIBUNWIND)
           SET(JEMALLOC_PROF "--enable-prof" "--enable-prof-libunwind" "--with-static-libunwind=${LIBUNWIND_LIB}")
-          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -I${LIBUNWIND_HOME}/include")
+          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -isystem ${LIBUNWIND_HOME}/include")
+          SET(MY_CPPFLAGS "-isystem ${LIBUNWIND_HOME}/include")
       else ()
           SET(JEMALLOC_PROF "--enable-prof")
       endif()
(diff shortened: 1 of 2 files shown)


maierlars (approved):

3rdParty/jemalloc/CMakeLists.txt
```diff
@@ -35,7 +35,8 @@ if (LINUX OR DARWIN)
   if (USE_JEMALLOC_PROF)
       if (USE_LIBUNWIND)
           SET(JEMALLOC_PROF "--enable-prof" "--enable-prof-libunwind" "--with-static-libunwind=${LIBUNWIND_LIB}")
-          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -I${LIBUNWIND_HOME}/include")
+          SET(JEMALLOC_C_FLAGS "${CMAKE_C_FLAGS} -isystem ${LIBUNWIND_HOME}/include")
+          SET(MY_CPPFLAGS "-isystem ${LIBUNWIND_HOME}/include")
```
jsteemann, line 39: `MY_CPPFLAGS` is only set when `USE_LIBUNWIND` is used.
Otherwise it is not set.
Should we set `MY_CPPFLAGS` to
```
if (USE_LIBUNWIND)
  SET(MY_CPPFLAGS "$ENV{CPPFLAGS} -isystem ${LIBUNWIND_HOME}/include")
else ()
  SET(MY_CPPFLAGS "$ENV{CPPFLAGS}")
endif ()
```
instead, to not clobber whatever `CPPFLAGS` were set from the outside?
Maybe this is not necessary. Just wondering.

neunhoef: Let's leave this for now, since this could also be a danger of breaking something (if flags from the outside leak into our build of jemalloc). In `devel`, we have basically the same setup.

kvahed (commented):
LGTM

kvahed (approved):
LGTM

PR #20310, opened 2023-12-13
Further simplify weighted path search

markuspf:
### Scope & Purpose

* Removes `ResultCache` as it is not needed anymore after recent simplifications
* Removes `_result` and  `ResultList`
* Simplifies returning of paths

### DIFF
--- a/arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
+++ b/arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
@@ -30,12 +30,10 @@
 #include "Graph/Options/TwoSidedEnumeratorOptions.h"
 #include "Graph/PathManagement/PathResult.h"
 #include "Containers/FlatHashMap.h"
-#include "Containers/FlatHashSet.h"
 
 #include <limits>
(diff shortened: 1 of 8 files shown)


arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
```diff
@@ -101,9 +98,29 @@ class WeightedTwoSidedEnumerator {
 
     [[nodiscard]] bool isEmpty() const { return _queue.empty(); }
 
-    [[nodiscard]] std::vector<CalculatedCandidate> getQueue() const& {
-      return _queue;
-    };
+    [[nodiscard]] std::vector<Step*> getLeftLooseEnds() {
```
jsteemann, line 101: These are the same changes as the ones in the separate PR https://github.com/arangodb/arangodb/pull/20366
Once the other PR is merged, then the changes will be in devel and this PR will have a smaller change surface.

markuspf: removed that change in favour of #20366

PR #20313, opened 2023-12-14
Improvements to the actor library

mpoeter:
### Scope & Purpose

- Add synchronous shutdown method that waits until all actors have finished
- Add getter in HandlerBase so we can access the runtime
- Instantiate actor with the derived Runtime type, not the BaseRuntime
- Avoid copying messages
- Fix potential data race when finishing actor

- [x] :hankey: Bugfix
- [x] :pizza: New feature


markuspf (approved):
Might use a test for the shutdown functionality?

markuspf:
:heart_eyes:

PR #20320, opened 2023-12-15
[BTS-1642] Failover Jobs checking for clones

maierlars:
### Scope & Purpose

Check that collections that are shared like the prototype collection still exist when the job is moving to pending. Otherwise we write to an already delete collection.

### DIFF
--- a/tests/Agency/FailedLeaderTest.cpp
+++ b/tests/Agency/FailedLeaderTest.cpp
@@ -499,7 +499,7 @@ class FailedLeaderTest
     ASSERT_TRUE(obj.isObject());
     // Will be set to false if ommited, or actively se t to false
     bool oldEmpty = VelocyPackHelper::getBooleanValue(obj, "oldEmpty", false);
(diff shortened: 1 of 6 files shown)


tests/Agency/FailedLeaderTest.cpp
```diff
@@ -488,6 +495,14 @@ class FailedLeaderTest
     EXPECT_TRUE(oldEmpty);
   }
 
+  void AssertOldNotEmptyObject(VPackSlice obj) {
+    ASSERT_TRUE(obj.isObject());
+    // Will be set to false if ommited, or actively se t to false
+    bool oldEmpty = VelocyPackHelper::getBooleanValue(obj, "oldEmpty", false);
+    // Required to be TRUE!
+    EXPECT_FALSE(oldEmpty);
```
jsteemann, line 503: The code comment does not match what the test expects.

markuspf (approved):
You wouldn't believe that this actually works!

PR #20323, opened 2023-12-15
Fix some shell_client tests when run with replication2

mpoeter:
Test Only change:

1. This makes sure we clean up if a test fails, this removes some false negative follow up tests in the Suite. This does not change anything in the test logic itself.
2. Do not run a Replication1 specific test in Replication2


mchacki (approved):
Code Looks good to me.

To add a description for the PR:

1) This makes sure we clean up if a test fails, this removes some false negative follow up tests in the Suite.
2) Do not run a Replication1 specific test in Replication2

PR #20329, opened 2023-12-18
Prefix Values for ZKD

maierlars:
### Scope & Purpose

There are two limitations when using a zkd index in a traversal:
1. The zkd index only allows indexing numerical values, but `_from` or `_to` are strings.
2. Even if strings were allowed, traversals use a condition containing `_from == <some-id>`. Equality checks are supported by the zkd index, but are not very fast, because the zcurve leaves the search box after every found document, which causes a huge amount of seeks.

This PR introduces are `prefixValues`. These values are stored as a velocypack array as prefix to the curve data, similar to the persistent indexes. In the traversal case, these values would contain `_from` or `_to` and can easily handle conditions that contain checks for equality. One drawback is, that the user has to specify *all* prefix values by equality.
 
TODO:
- [x] Introduce static strings for `sortedPrefixValues`.
- [x] Rename to `prefixFields`.
- [ ] Allow projections from the prefix fields. (Moved into separate tickets)
- [ ] Add a separate column family for prefixed zkd index. (Moved into separate tickets)


arangod/Indexes/IndexFactory.cpp
```diff
@@ -752,6 +752,63 @@ Result IndexFactory::enhanceJsonIndexFulltext(VPackSlice definition,
...
+        }
+        auto normalFields = definition.get(StaticStrings::IndexStoredValues);
+        TRI_ASSERT(normalFields.isArray());
+        for (VPackSlice it : VPackArrayIterator(normalFields)) {
+          if (!fields.insert(it.stringView()).second) {
+            res.reset(TRI_ERROR_BAD_PARAMETER,
+                      "duplicate attribute name (overlap between index sorted "
+                      "prefix fields "
+                      "and index "
+                      "stored values list)");
```
jsteemann, line 785: nit: we can as well break after the first error.
```suggestion
            res.reset(TRI_ERROR_BAD_PARAMETER,
                      "duplicate attribute name (overlap between index sorted "
                      "prefix fields "
                      "and index "
                      "stored values list)");
            break;
```

arangod/RocksDBEngine/RocksDBZkdIndex.h
```diff
@@ -44,8 +44,13 @@ class RocksDBZkdIndexBase : public RocksDBIndex {
   std::vector<std::vector<basics::AttributeName>> const& coveredFields()
       const override {
     // index does not cover the index attributes!
+    // return _coveredFields;
```
jsteemann, line 47:
```suggestion
```

arangod/RocksDBEngine/RocksDBZkdIndex.h
```diff
@@ -44,8 +44,13 @@ class RocksDBZkdIndexBase : public RocksDBIndex {
...
     return _storedValues;
```
jsteemann, line 48: If the ZKD index has prefix fields, it will also cover them, provided it is chosen by the optimizer for eligible queries. So here we could return the union of stored values and prefix fields.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -424,11 +484,41 @@ void zkd::extractBoundsFromCondition(
...
+      return false;
+    }
+
+    for (auto&& [idx, field] : enumerate(getSortedPrefixFields(index))) {
+      if (attributeData.second != field) {
+        continue;
+      }
+
+      auto [it, inserted] = extractedPrefix.emplace(idx, other);
+      TRI_ASSERT(inserted) << "duplicate access for " << attributeData.second;
```
jsteemann, line 505: It is possible that the user specifies the same attribute multiple times in a filter condition by mistake, e.g.
`FILTER doc._from == 'abc' && doc._from == 'def'`, or `FILTER doc._from == 'abc' && doc._from == 'abc'`.
I think this is detected earlier on already, but maybe there are some cases under which such obviously senseless filter conditions will make it into this function. If that happens, I think we should handle it gracefully and not fail with an assertion failure (in maintainer mode).

arangod/RocksDBEngine/RocksDBZkdIndex.h
maierlars: As I wrote under ToDos
> Allow projections from the prefix fields. (Moved into separate tickets)

jsteemann: Vale

jsteemann (approved):
LGTM

PR #20330, opened 2023-12-18
Fix: check replicationFactor against number of available servers

mpoeter:
### Scope & Purpose

In replication2 we plan the distribution of shards to the individual servers in the supervision. However, once the coordinator has written the relevant information to target, the agency has no way to report back any errors, but has to go through with it. So on the coordinator we at least have to check if the replicationFactor can be fulfilled with the number of available servers, before we write to target.

This should fix a couple of tests when run with replication2 as default.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**


mchacki (approved):
LGTM!

PR #20332, opened 2023-12-18
Chore/circleci windows ccache

CryptoNinjaGeek:
### Scope & Purpose

Get ccache working on our window builds, to cut down compile time.


dothebart (approved):
it would be good if there was a way to DRY the docker container names. 
else LGTM.

CMakeLists.txt
```diff
@@ -937,10 +937,52 @@ if (MSVC)
...
+
+      if(CCACHE_RESULT_CODE)
+        set( CCACHE ${CCACHE_REAL} )
+        message(STATUS "Resolved ccache SHIM: '${CCACHE_REAL}'")
+      endif()
+
+      file(COPY_FILE 
+        ${CCACHE} ${CMAKE_BINARY_DIR}/cl.exe
+        ONLY_IF_DIFFERENT
+      )
```
mpoeter, line 962: I think this deserves a comment about why we need that.

.circleci/base_config.yml
```diff
@@ -344,24 +344,35 @@ jobs:
...
           name: "Build"
           shell: powershell.exe
           command: |
-            docker run --rm --isolation process -v D:/CircleCI/project:C:/project -w C:/project/build arangodb/build-windows-x64:10.0.17763.4737-VS-17.3.3-OpenSSL-3.1.4-f3c77755ece cmake --build . --config `"RelWithDebInfo`" --parallel -- /p:CL_MPcount=24
+            docker run --rm --isolation process -v D:/CircleCI/project:C:/project -v D:/CircleCI-ccache:C:/ccache -w C:/project/build arangodb/build-windows-x64:10.0.17763.4737-VS-17.3.3-OpenSSL-3.1.4-399cc6a6e26 cmake --build . --config `"RelWithDebInfo`" --parallel -- /p:CL_MPcount=24 
+      - run:
+          name: "Print CCache statistics"
+          shell: powershell.exe
+          command: |
+            docker run --rm --isolation process -v D:/CircleCI/project:C:/project -v D:/CircleCI-ccache:C:/ccache -w C:/project/build arangodb/build-windows-x64:10.0.17763.4737-VS-17.3.3-OpenSSL-3.1.4-399cc6a6e26 ccache -s
```
mpoeter, line 375: We probably don't need the project directory, and we also don't care about the work dir
```suggestion
            docker run --rm --isolation process -v D:/CircleCI-ccache:C:/ccache arangodb/build-windows-x64:10.0.17763.4737-VS-17.3.3-OpenSSL-3.1.4-399cc6a6e26 ccache -s
```

mpoeter (approved):
I think the cmake could use a comment, otherwise LGTM

.circleci/base_config.yml
CryptoNinjaGeek: Change in multiple places

PR #20335, opened 2023-12-18
Fix: db can already be gone when we process the reconfigureReplicatedLog action

mpoeter:
### Scope & Purpose

The isReplication2 check can return false when the db is not longer found. However, the returned undo action already handles the case when the DB is already gone, and in contrast to replication1, replication2 does not operate on plan, but on target, so this check was superfluous.

This sometimes lead to assertion failures in rebalance tests.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests**


mchacki (approved):
LGTM

PR #20340, opened 2023-12-19
Fix: cannot use rf=3 when we have only 2 servers

mpoeter:
### Scope & Purpose

Adjust a test that uses an invalid replication factor. Previously this worked, because in replication2 we were missing an according check. This was fixed recently, which broke this test.


mchacki (approved):
LGTM

PR #20341, opened 2023-12-19
Fix more shell_client tests for replication2

mpoeter:
### Scope & Purpose

- move a failure point so it works for replication 1 and 2
- disable tests that are not relevant for replication2
- with replication2 a replication factor update is propagated to all collections in the collection group


mchacki (approved):
LGTM

PR #20346, opened 2023-12-20
work around wintendo pid grabbing CI issue

dothebart:
### Scope & Purpose

CI would try to kill svchost.exe - don't do that.

- [x] :hankey: Bugfix

### DIFF
--- a/scripts/test/async_client.py
+++ b/scripts/test/async_client.py
@@ -200,12 +200,13 @@ def kill_children(identifier, params, children):
         if one_child.pid in killed:
             continue
         try:
-            killed.append(one_child.pid)
-            err += add_message_to_report(
-                params,
-                f"{identifier}: killing {one_child.name()} - {str(one_child.pid)}",
-            )
-            one_child.resume()
+            if one_child.name() != "svchost.exe":
+                killed.append(one_child.pid)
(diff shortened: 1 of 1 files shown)


mpoeter (approved):

scripts/test/async_client.py
```diff
@@ -200,12 +200,14 @@ def kill_children(identifier, params, children):
...
             continue
         try:
-            killed.append(one_child.pid)
-            err += add_message_to_report(
-                params,
-                f"{identifier}: killing {one_child.name()} - {str(one_child.pid)}",
-            )
-            one_child.resume()
+            pname = one_child.name()
+            if pname not in ["svchost.exe", "conhost.exe", "mscorsvw.exe"]:
```
mpoeter, line 204: No... please check instead if the executable starts with "arango" or something like that.

mpoeter (changes requested):

scripts/test/async_client.py
dothebart: No. 
we rather attempt to kill one process to much and have a false positive, than kill one to less, and get no proper results.

mpoeter (approved):

PR #20347, opened 2023-12-20
VPack Zkd Column Family

maierlars:
### Scope & Purpose

Add a separate column family for zkd indexes that use prefixed values. They use the same configurations but have different key formats.


markuspf (approved):
lgtm

arangod/RocksDBEngine/RocksDBOptionFeature.cpp
```diff
@@ -1498,6 +1500,38 @@ of outgrowing the maximum number of file descriptors the ArangoDB process
...
+                  arangodb::options::makeFlags(
+                      arangodb::options::Flags::Uncommon,
+                      arangodb::options::Flags::Experimental,
+                      arangodb::options::Flags::DefaultNoComponents,
+                      arangodb::options::Flags::OnAgent,
+                      arangodb::options::Flags::OnDBServer,
+                      arangodb::options::Flags::OnSingle))
+      .setIntroducedIn(31200)
+      .setLongDescription(R"(Enabling this option will make RocksDB's
+  compaction write the persistent index data for different zkd
```
jsteemann, line 1517:
```suggestion
  compaction write the zkd index data for different zkd
```

arangod/RocksDBEngine/RocksDBOptionFeature.cpp
```diff
@@ -1498,6 +1500,38 @@ of outgrowing the maximum number of file descriptors the ArangoDB process
...
+  indexes (also indexes from different collections/shards) into different
+  .sst files. Otherwise the persistent index data from different
```
jsteemann, line 1519:
```suggestion
  .sst files. Otherwise the zkd index data from different
```

arangod/RocksDBEngine/RocksDBOptionFeature.h
```diff
@@ -172,6 +172,7 @@ class RocksDBOptionFeature final : public ArangodFeature,
   bool _partitionFilesForPrimaryIndexCf;
   bool _partitionFilesForEdgeIndexCf;
   bool _partitionFilesForVPackIndexCf;
+  bool _partitionFilesForZkdIndexCf;
```
jsteemann, line 175: It looks like this variable is not initialized anywhere, so it has an undefined startup value.

jsteemann (changes requested):
The member variable `_partitionFilesForZkdIndexCf` does not seem to be initialized.
This should be fixed!

arangod/RocksDBEngine/RocksDBKeyBounds.cpp
```diff
@@ -415,6 +423,9 @@ RocksDBKeyBounds::RocksDBKeyBounds(RocksDBEntryType type, uint64_t first,
     : _type(type) {
   switch (_type) {
     case RocksDBEntryType::ZkdIndexValue:
+    case RocksDBEntryType::UniqueZkdIndexValue:
+    case RocksDBEntryType::ZkdVPackIndexValue:
+    case RocksDBEntryType::UniqueZkdVPackIndexValue:
     case RocksDBEntryType::VPackIndexValue:
     case RocksDBEntryType::UniqueVPackIndexValue: {
       uint8_t const maxSlice[] = {0x02, 0x03, 0x1f};
```
jsteemann, line 431: Is it actually ok to use a VelocyPack `max` value for the upper bound (after the index id) in case of the traditional ZkdIndex?
I was under the impression that the traditional Zkd index does not store VelocyPack data in the keys, and not use a VelocyPack-aware comparator. So using a VelocyPack `max` value as part of the upper will likely not work as expected.

arangod/RocksDBEngine/RocksDBOptionFeature.cpp
maierlars: I did this change initially, but I though it is refering to the persistent data of that index, which just happens to be called `persistent` index and in this case `zkd` index.

PR #20351, opened 2023-12-20
Projections from Prefix Values

maierlars:
### Scope & Purpose

This PR adds the capability of projection values out of the prefix fields of the zkd index.


arangod/RocksDBEngine/RocksDBZkdIndex.h
```diff
@@ -44,7 +44,7 @@ class RocksDBZkdIndexBase : public RocksDBIndex {
   std::vector<std::vector<basics::AttributeName>> const& coveredFields()
       const override {
     // The index only covers stored values
-    return _storedValues;
+    return _coveredFields;
```
markuspf, line 47: The comment above is now out of date, I presume

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -219,12 +219,24 @@ class RocksDBZkdIndexIterator final : public IndexIterator {
   bool nextCoveringImpl(CoveringCallback const& callback,
                         uint64_t limit) override {
     struct CoveringData : IndexIteratorCoveringData {
-      explicit CoveringData(velocypack::Slice data) : _data(data) {}
-      VPackSlice at(size_t i) const override { return _data.at(i); }
+      CoveringData(VPackSlice prefixValues, VPackSlice storedValues)
```
markuspf, line 222: is there a reason that this constructor is not marked explicit anymore?

markuspf (approved):
LGTM, modulo a question and a comment.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
maierlars: yes, it has now two parameters. They can't be used in implicit conversions, right?

PR #20354, opened 2023-12-20
Validate that smartgraph attribute exists and is not changed

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/1422

Previously this check was performed in the KeyGenerator and therefore only performed for inserts, but not for replace or update operations. Therefore replace or update operations could modify or even remove this attribute.
Now we use a separate validator to ensure that this cannot happen. In addition, this PR also changes the returned error code from `ERROR_KEY_MUST_BE_PREFIXED_WITH_SMART_GRAPH_ATTRIBUTE` to `ERROR_NO_SMART_GRAPH_ATTRIBUTE` if the attribute is not set.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **integration tests** 


arangod/VocBase/LogicalCollection.h
```diff
@@ -182,6 +182,7 @@ class LogicalCollection : public LogicalDataSource {
   bool isSmartEdgeCollection() const noexcept;
   bool isSatToSmartEdgeCollection() const noexcept;
   bool isSmartToSatEdgeCollection() const noexcept;
+  bool isSmartVertexCollection() const noexcept;
```
jsteemann, line 185: There is no implementation for this new method in the Community Edition.
This may be ok if we don't call it, but it looks fishy. I suggest adding a stub implementation for the method as we did for the other methods above too.

mpoeter: This method is only (and should only ever be) called from within EE code. I prefer to not have a definition in the community, so that we get a linker error in case this is ever called within a CE build.

jsteemann: Ok. The other methods above do have an implementation though.
Not saying that these should be adjusted now. Just for information.

jsteemann (approved):
LGTM

mchacki (approved):
LGTM

PR #20355, opened 2023-12-20
Disable a test that is not relevant for replication2

mpoeter:
### Scope & Purpose

This test coves an update scenario that is only relevant for old versions that predate replication2.


jsteemann (approved):
LGTM

PR #20361, opened 2023-12-21
Disable some chaos failure points for replication2

mpoeter:
### Scope & Purpose
 
In replication2 the modification operations on the follower MUST NOT fail!
If they do, we conclude that something else must have gone terribly wrong and therefore abort the process.


mchacki (approved):
LGTM!

PR #20365, opened 2023-12-21
only waitpid for the processes we actually sent a kill signal

dothebart:
### Scope & Purpose

subsequent runs would show throws by the `wait_procs` invocation which would wait for not killed processes (svchost...) and throw.

- [x] :hankey: Bugfix


mpoeter (approved):

dothebart:
windows support dropped.

PR #20366, opened 2023-12-21
Fix copy + use after free in path search code

markuspf:
### Scope & Purpose

The old code took a copy of the queue of candidates and then stored pointers into that copy for fetching results. This blew up in recent tests after simplification.

This fix isn't exactly pretty, but is modeled after the `getLooseEnds` in other Queue constructs in path search and addresses the immediate problem.

I am working on further patches that simplify this code but they will take longer to materialise.

### DIFF
--- a/arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
+++ b/arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
@@ -101,9 +101,29 @@ class WeightedTwoSidedEnumerator {
 
     [[nodiscard]] bool isEmpty() const { return _queue.empty(); }
 
-    [[nodiscard]] std::vector<CalculatedCandidate> getQueue() const& {
-      return _queue;
-    };
+    [[nodiscard]] std::vector<Step*> getLeftLooseEnds() {
+      std::vector<Step*> steps;
+
+      for (auto& [_, step, __] : _queue) {
+        if (!step.isProcessable()) {
+          steps.emplace_back(&step);
+        }
+      }
+
+      return steps;
+    }
+
+    [[nodiscard]] std::vector<Step*> getRightLooseEnds() {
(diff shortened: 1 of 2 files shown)


arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
```diff
@@ -101,9 +101,29 @@ class WeightedTwoSidedEnumerator {
 
     [[nodiscard]] bool isEmpty() const { return _queue.empty(); }
 
-    [[nodiscard]] std::vector<CalculatedCandidate> getQueue() const& {
-      return _queue;
-    };
+    [[nodiscard]] std::vector<Step*> getLeftLooseEnds() {
```
jsteemann, line 104: nit: This method can be made const.

arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
```diff
@@ -101,9 +101,29 @@ class WeightedTwoSidedEnumerator {
...
+      for (auto& [_, step, __] : _queue) {
+        if (!step.isProcessable()) {
+          steps.emplace_back(&step);
+        }
+      }
+
+      return steps;
+    }
+
+    [[nodiscard]] std::vector<Step*> getRightLooseEnds() {
```
jsteemann, line 116: nit: This method can be made const.

jsteemann (approved):
LGTM

arangod/Graph/Enumerators/WeightedTwoSidedEnumerator.h
markuspf: afaict, it can't, because then the `Step *` want to become `const` and hilarity ensues. I am on the case getting this code straightened out; it's just taking a bit. In particular I need to sort out lifetimes of these steps....

PR #20367, opened 2023-12-21
Fix more shell_client replication2 test

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/1426

Collection creation/update agency operations must skip the scheduler
This is necessary to ensure that the collection creation works even when the medium queue is full

The collection property cacheEnabled is actually mutable.

Note: the enterprise companion also contains some test fixes, but these are actually unrelated to the ones in this PR.


mchacki (approved):
LGTM

PR #20370, opened 2023-12-22
fix crash upon exception during recursive AstNode creation [BTS-1720]

jsteemann:
### Scope & Purpose

Fix a crash that could occur when an exception was thrown during recursive creation of AstNodes from VelocyPack.
In this case, the FixedSizeAllocator that was responsible for creating and destroying the AstNodes inside its own memory buffers was assuming that the exception was thrown by the creation of the current AstNode. But in fact the exception could have been thrown by a nested AstNode as well. In this case, the cleanup procedure may not have worked.
This PR fixes it by not attempting to roll back the last allocation, but rather to patch the memory hole with a default-constructed AstNode.
This is a relatively small fix that can easily be backported, and it can be improved later.

The issue fixed by this bugfix happened when AstNodes were created on DB servers from a VelocyPack query plan snippet, and the memory usage exceeded the configured query memory usage limit. This can happen especially with larger bind parameter values.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20368
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/20371

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/AstNode.cpp
+++ b/arangod/Aql/AstNode.cpp
@@ -413,8 +413,19 @@ int compareAstNodes(AstNode const* lhs, AstNode const* rhs, bool compareUtf8) {
   }
 }
 
+// private ctor, only called during by FixedSizeAllocator in case of emergency
+// to properly initialize the node
+AstNode::AstNode() noexcept(noexcept(decltype(members)::allocator_type()))
+    : type(NODE_TYPE_NOP), flags(0), _computedValue(nullptr), members{} {
+  // properly zero-initialize all members
+  value.value._int = 0;
+  value.length = 0;
+  value.type = VALUE_TYPE_NULL;
+}
+
 /// @brief create the node
-AstNode::AstNode(AstNodeType type)
+AstNode::AstNode(AstNodeType type) noexcept(
+    noexcept(decltype(members)::allocator_type()))
     : type(type), flags(0), _computedValue(nullptr) {
   // properly zero-initialize all members
   value.value._int = 0;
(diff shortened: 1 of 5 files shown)


arangod/Aql/AstNode.cpp
```diff
@@ -413,8 +413,19 @@ int compareAstNodes(AstNode const* lhs, AstNode const* rhs, bool compareUtf8) {
   }
 }
 
+// private ctor, only called during by FixedSizeAllocator in case of emergency
+// to properly initialize the node
```
neunhoef, line 417:
```suggestion
// private ctor, only called during by FixedSizeAllocator in case of emergency
// to properly initialize the node
// Note that since C++17 the default constructor of `std::vector` is
// `noexcept` iff and only if the  default constructor of its `allocator_type`
// is. Therefore, we can say that `AstNode::AstNode()` is noexcept, if and
// only if the default constructor of the allocator type of
// `std::vector<AstNode*>` is noexcept, which is exactly what this fancy
// `noexcept` expression does.
```

arangod/Aql/AstNode.h
```diff
@@ -589,6 +596,10 @@ struct AstNode {
   AstNodeValue value;
 
  private:
+  // private ctor, only called during by FixedSizeAllocator in case of emergency
+  // to properly initialize the node
```
neunhoef, line 600:
```suggestion
  // private ctor, only called during by FixedSizeAllocator in case of emergency
  // to properly initialize the node
  // Note that since C++17 the default constructor of `std::vector` is
// `noexcept` iff and only if the  default constructor of its `allocator_type`
// is. Therefore, we can say that `AstNode::AstNode()` is noexcept, if and
// only if the default constructor of the allocator type of
// `std::vector<AstNode*>` is noexcept, which is exactly what this fancy
// `noexcept` expression does.
```

neunhoef (approved):
LGTM. See two suggestions about additional source code comments which are actually already present in 3.11 and 3.10.

markuspf (approved):

PR #20373, opened 2023-12-23
Simplify path search templates

markuspf:
### Scope & Purpose

This PR aims to untangle the slightly unholy template mess that reigns in the path search code, and make it more similar to the traversal code


markuspf:
Completely out of date.

PR #20384, opened 2023-12-27
Revert revert 20366 bug fix/dangling references

maierlars:
### Scope & Purpose
Revert the revert of #20366 . And fix the actual issue.


neunhoef (approved):
LGTM

PR #20390, opened 2023-12-28
ZKD Index Estimates

maierlars:
### Scope & Purpose

This PR adds index estimates to the zkd index for
- unique indexes.
- indexes using prefix values.

The non prefixed case is rather hard to give an accurate estimate, similar to why we don't estimate sorted indexes when using the compare operator.


arangod/Aql/TraversalNode.cpp
```diff
@@ -1176,11 +1176,6 @@ void TraversalNode::prepareOptions() {
 
   TraverserOptions* opts = this->TraversalNode::options();
   TRI_ASSERT(opts != nullptr);
-  /*
-   * HACK: DO NOT use other indexes for smart BFS. Otherwise, this will produce
-   * wrong results.
-   */
-  bool onlyEdgeIndexes = this->isSmart() && opts->isUseBreadthFirst();
```
maierlars, line 1183: This is an old fix that became unnecessary due to the graph refactoring a few years ago. It essentially disabled the usage of non-edge indexes for smart graph bfs.

arangod/Graph/TraverserOptions.cpp
```diff
@@ -97,7 +97,7 @@ TraverserOptions::TraverserOptions(arangodb::aql::QueryContext& query,
       THROW_ARANGO_EXCEPTION_MESSAGE(
           TRI_ERROR_BAD_PARAMETER,
           "uniqueVertices: 'global' is only "
-          "supported, with mode: bfs|weighted due to "
+          "supported, with order: bfs|weighted due to "
```
maierlars, line 100: The error message is confusing, because the option is actually called `order`.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -579,10 +578,36 @@ auto zkd::supportsFilterCondition(
...
+
+  double const unusedDimensionCost =
+      0.5 * static_cast<double>(unusedDimensions * costs.estimatedItems);
+  auto const unusedExpressionCost =
+      static_cast<double>(costs.estimatedItems * unusedExpressions.size());
+
+  // account for post filtering
+  costs.estimatedCosts = static_cast<double>(costs.estimatedItems) +
+                         unusedDimensionCost + unusedExpressionCost;
+
```
maierlars, line 610: I'm happy to discuss this estimation and to completely geek out about it. But for now I would like to merge this PR and improve the estimations later on.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -579,10 +578,36 @@ auto zkd::supportsFilterCondition(
...
-  // TODO -- actually return costs
-  auto costs = Index::FilterCosts::defaultCosts(itemsInIndex);
-  costs.coveredAttributes = extractedBounds.size() + extractedPrefix.size();
+  Index::FilterCosts costs;
   costs.supportsCondition = true;
+  costs.coveredAttributes = extractedBounds.size() + extractedPrefix.size();
+
+  // we look up a single point using the prefix values
+  const double estimatedElementsOnCurve =
+      index->hasSelectivityEstimate() ? 1. / index->selectivityEstimate()
```
jsteemann, line 587: The selectivity estimate could be `0` here, so we would attempt to divide by zero.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -579,10 +578,36 @@ auto zkd::supportsFilterCondition(
...
+  const double volumeReductionFactor = 1.4;  // guessed, 2 might be too much
+  const double searchBoxVolume =
+      1. /
+      pow(volumeReductionFactor, static_cast<double>(extractedBounds.size()));
+
+  costs.estimatedItems =
+      static_cast<size_t>(estimatedElementsOnCurve * searchBoxVolume);
+
+  const size_t unusedDimensions =
+      index->fields().size() - extractedBounds.size();
```
jsteemann, line 600: I guess `extractedBounds` is always `<=` to `index->fields().size()` here, but I am not 100% sure.
It is fine if this is already guarenteed by other means.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -667,24 +692,26 @@ auto extractAttributeValues(
...
 
   RocksDBKey rocksdbKey;
+  uint64_t hash = 0;
   if (_prefixFields.empty()) {
     rocksdbKey.constructZkdIndexValue(objectId(), keyValue, documentId);
   } else {
     auto prefixValues = extractAttributeValues(trx, _prefixFields, doc);
     rocksdbKey.constructZkdIndexValue(objectId(), prefixValues->slice(),
                                       keyValue, documentId);
+    hash = prefixValues->slice().normalizedHash();
```
jsteemann, line 714: nit: calculating the hash is unnecessary if `!estimates`.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -694,33 +721,70 @@ Result RocksDBZkdIndexBase::insert(transaction::Methods& trx,
...
```
jsteemann, line 771: nit: calculating the hash is unnecessary if `!estimates`.

arangod/Aql/TraversalNode.cpp
```diff
@@ -1197,16 +1192,14 @@ void TraversalNode::prepareOptions() {
...
+              builder->getInboundCondition()->clone(ast), depth, false, dir);
           break;
         case TRI_EDGE_OUT:
-          opts->addDepthLookupInfo(_plan, _edgeColls[i]->name(),
-                                   StaticStrings::FromString,
-                                   builder->getOutboundCondition()->clone(ast),
-                                   depth, onlyEdgeIndexes, dir);
+          opts->addDepthLookupInfo(
+              _plan, _edgeColls[i]->name(), StaticStrings::FromString,
+              builder->getOutboundCondition()->clone(ast), depth, false, dir);
```
jsteemann, line 1202: These two callers have been the only ones to `addDepthLookupInfo`. 
As we are now hard-coding the `onlyEdgeIndexes` call parameter to `false` here, we could as well remove it from the function `addDepthLookupInfo` function signature.

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
maierlars: yes, it is guaranteed.

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -502,7 +513,6 @@ ClusterIndex::coveredFields() const {
     case TRI_IDX_TYPE_GEO2_INDEX:
     case TRI_IDX_TYPE_FULLTEXT_INDEX:
     case TRI_IDX_TYPE_TTL_INDEX:
-    case TRI_IDX_TYPE_ZKD_INDEX:
```
jsteemann, line 505: Not sure if the removal is correct here.
Otherwise, if we have a ZKD index that does not cover anything, it will now return `_fields` whereas before it returned `[]`.

arangod/RocksDBEngine/RocksDBIndexFactory.cpp
```diff
@@ -286,11 +286,61 @@ struct ZkdIndexFactory : public DefaultIndexFactory {
           arangodb::velocypack::Value(std::to_string(TRI_NewTickServer())));
     }
 
+    if (definition.hasKey(StaticStrings::IndexPrefixFields)) {
+      return Result(TRI_ERROR_BAD_PARAMETER,
+                    "`mdi` index does not support prefixed fields. use "
+                    "`mdi-prefixed` as type instead.");
+    }
+    // a zdk index never uses index estimates
```
jsteemann, line 294:
```suggestion
    // a zkd index never uses index estimates
```

arangod/RocksDBEngine/RocksDBIndexFactory.cpp
```diff
@@ -403,6 +455,7 @@ RocksDBIndexFactory::indexAliases() const {
   return {
       {"hash", "persistent"},
       {"skiplist", "persistent"},
+      {"mdi", "zkd"},
```
jsteemann, line 458: The aliases in this object are `{ old name, new name }`. So here we would add `mdi` as an alias for `zkd`.
Not sure if this is intentional or if it should be the other way round. Potentially it is ok for this PR but can be changed later if zkd is renamed to mdi.

maierlars: that was my plan

jsteemann (approved):
LGTM

PR #20393, opened 2024-01-02
[CINFRA] Attempt to "fix" ReplicatedLog shutdown blockage

mchacki:
### Scope & Purpose

*This is an attempt to mitigate an issue we have in our version of RocksDB.*
We do NOT solve the issue, we will simply not wait if the issue on compact shows up, but continue normally.
If the issue kicks in the following happens:
1) We drop a replicated state
2) For this we do a range delete in rocksdb over all entries in the states log.
3) After the range delete we do a "compact" call, which can cause the below bug.

Before: We do the compact in a Maintenance thread, and wait for it to complete, if the issue shows up the maintenance thread is blocked.
Now: We schedule the compact on low-priority queue and move on. If the issue shows up we temporarily loose a low-pro thread.
 
In Server shutdown scenario, we wait on the Maintenance to complete (which can not happen after the bug), now there is a "cancel" flag handed in to the LowPrio thread, basically telling it to give up and shutdown.

The issue seems to be the following:

1) We schedule a compaction of a column family in RocksDB.
2) The ColumnFamily has one meltable_list instance, which has a `bool flush_requested_` flag, if we schedule the compaction this flag is set to `true`.
3) Now RocksDBBackground Thread starts working on the compaction task, internally this does a FlushMemTablesToOutputFiles(), and compacts the Memtables known now. (Say this has ID: 5)
4) We add a new MemTable to the column family. This is not known and relevant to (3). And gets ID 6.
5) FlushMemTablesToOutputFiles does several fsync operations, for those it releases a rather global lock, allowing the next operation to pass, while it is still ongoing.
6) We use a different thread the schedule a compaction of the same column family (non-overlapping range)
6a) Side effect: This sets the `flush_requested_ ` from `true` to `true` again and gets added to the `flush_queue_`
7) This compaction waits for Memtable ID 6 to be flushed. And the CompactRange blocks until ID 6 is compacted.
8) Now RocksDBBackground continues. It successfully compacts MemTable with ID 5 as it already started.
8a) This triggers the following code:
`  if (!atomic_flush || num_flush_not_started_ == 0) {
    flush_requested_ = false;  // start-flush request is complete
  }`
8b) In our case we do not do an atomic_flush => We always set flush_requested to `false`.
9) The RocksDBBackgroundThread finds that `flush_queue_` is not empty and takes the second compaction of ColumnFamily as a work item. There it tests `!cfd->imm()->IsFlushPending()` to decide if the flush is still relevant. This is only true if `flush_requested_ == true` or if we have "Enough" files to flush.
9a) In our case we do not have "enough" files. So the flush operation completes with "nothing to do".
10) Now the Compact() waiting is blocked because the Memtable 6 is not yet compacted.
11) This is unblocked as soon as we have "enough" MemTables. Or as soon as someone triggers another Compaction on the column family.

In Production: 11) should be eventually triggered if we replicate any data entries, or if we drop a different shard.

- [x] 💩  💩 
- [] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Replication2/Storage/RocksDB/LogPersistor.cpp
```diff
@@ -110,13 +113,11 @@ auto LogPersistor::drop() -> Result {
...
-  auto start = range.start();
-  auto end = range.end();
-  auto res = db->CompactRange(
-      ::rocksdb::CompactRangeOptions{.exclusive_manual_compaction = false,
-                                     .allow_write_stall = false},
-      logCf, &start, &end);
-  return rocksutils::convertStatus(res);
+  _engine->compactRange(range);
+  // The CompactRange of the engine does not have
+  // a return value and is asynchroneous. So we
```
mpoeter, line 118:
```suggestion
  // a return value and is asynchronous. So we
```

arangod/Replication2/Storage/RocksDB/LogPersistor.h
```diff
@@ -68,6 +73,7 @@ struct LogPersistor final : ILogPersistor {
   std::shared_ptr<AsyncLogWriteBatcherMetrics> const _metrics;
   ::rocksdb::DB* const db;
   ::rocksdb::ColumnFamilyHandle* const logCf;
+  arangodb::ICompactKeyRange* _engine;
```
mpoeter, line 76: I would prefer a different name since we don't really have the engine - after all you did the right thing and applied the [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle), so we have an `ICompactKeyRange` instance that can only do `compactRange`. So can we call it "compactor" or something like that?

mpoeter (commented):
Just some nitpicking - otherwise LGTM!

arangod/Replication2/Storage/RocksDB/LogPersistor.h
mchacki: Absoutely, will rename

PR #20399, opened 2024-01-03
fix non-deterministic behavior on Windows

jsteemann:
### Scope & Purpose

The register planning could insert different variables with the same register id into the VarInfoMap.
The info from the VarInfoMap was later copied into a vector. VarInfoMap is an unordered container with no guaranteed iteration order. Thus it was not deterministic which of the inserted variables with the same register id value was found first.
The issue did not occur in any of the Linux builds, potentially because there the iteration order for unordered containers is more deterministic than on Windows, for better or worse. It was a bug anyway, and the iteration order must not play a role here.

The issue was uncovered by a recent change in EnumerateListExecutor, which was made only in devel. Thus no backports needed.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/EnumerateListExecutor.cpp
+++ b/arangod/Aql/EnumerateListExecutor.cpp
@@ -62,11 +62,11 @@ EnumerateListExpressionContext::EnumerateListExpressionContext(
     VariableId outputVariableId)
     : QueryExpressionContext(trx, context, cache),
       _varsToRegister(varsToRegister),
-      _outputVariableId(outputVariableId) {}
+      _outputVariableId(outputVariableId),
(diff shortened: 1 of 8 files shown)


max-itzpapalotl (approved):
LGTM

PR #20405, opened 2024-01-04
[ZKD] Sparse Property

maierlars:
### Scope & Purpose
`mdi` (or `zkd`) and `mdi-prefixed` now support `sparse: true`.


arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -763,16 +788,32 @@ Result RocksDBZkdIndex::remove(transaction::Methods& trx,
...
+    keyValue = std::move(result.get());
+  }
 
   RocksDBKey rocksdbKey;
   uint64_t hash = 0;
   if (!isPrefixed()) {
     rocksdbKey.constructZkdIndexValue(objectId(), keyValue, documentId);
   } else {
-    auto prefixValues = extractAttributeValues(trx, _prefixFields, doc);
+    auto result = extractAttributeValues(trx, _prefixFields, doc, !_sparse);
```
jsteemann, line 810: The new parameter `nullAllowed` is now set to true if `!_sparse`, i.e. we allow null values if the index is not sparse.
Shouldn't this be opposite, i.e. 
```suggestion
    auto result = extractAttributeValues(trx, _prefixFields, doc, _sparse);
```
?

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
```diff
@@ -691,6 +694,11 @@ auto extractAttributeValues(
     if (s.isNone()) {
       s = VPackSlice::nullSlice();
     }
+
+    if (s.isNull() && !nullAllowed) {
+      return {TRI_ERROR_ARANGO_DOCUMENT_KEY_MISSING};
```
jsteemann, line 699: Is this the best error code we can use here? In other cases we are using it to indicate that the `_key` attribute is missing.
This is not the case here. I am happy if we use this error code here too, but it should be a conscious decision.

maierlars: The code is only used internally and never leaves the database. It's only used within this file. Happy to change it, I just don't want to add a new error code here. Maybe something more generic?

jsteemann: If it is used only internally, I think it is fine as is. Would be useful to add a code comment somewhere to indicate that this is only an internal code to signal a certain condition.

jsteemann (approved):
LGTM

arangod/RocksDBEngine/RocksDBZkdIndex.cpp
maierlars: I will add that comment in a follow up PR.

PR #20408, opened 2024-01-04
Back port index progress visibility

kvahed:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1428

Backport of index [visibility progress](https://github.com/arangodb/arangodb/pull/18516) on `_api/index`*

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1428
- [ ] Design document:


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

arangod/VocBase/Methods/Indexes.cpp
```diff
@@ -171,7 +171,7 @@ arangodb::Result Indexes::getAll(
       }
     }
     tmp.close();
-
+    std::cout << tmp.toJson() << std::endl;
```
jsteemann, line 174: Must be removed before merge!
```suggestion
```

tests/js/common/shell/shell-index.js
```diff
@@ -55,6 +55,22 @@ function IndexSuite() {
...
+      for (var i = 0; i < 10000; ++i) {
+        docs.push({name : "name" + i});
+      }
+      c.insert(docs);
+      internal.debugSetFailAt("fillIndex::pause");
+      c.ensureIndex({ type: "persistent", fields:["name"]});
+      console.error("hu");
+      require('internal').sleep(1);
+      console.error("ha");
+      internal.debugSetFailAt("fillIndex::unpause");
```
jsteemann, line 71: This is not a "real" test, in the sense of that does not have any conditions under which it will fail
So it executes something that one can attend and watch, but for automated tests it is not so helpful.
Apart from that, the test suite will be execute against a single server and cluster alike, and in the cluster it will likely not be any useful, because the failure points would need to be propagated to all DB servers for this (which I guess is not what is done here).

jsteemann (changes requested):
See my last 2 comments

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
       return RestStatus::DONE;
     }
 
+    std::stringstream ap;
+    ap << "Plan/Collections/" << _vocbase.name() << "/" << coll->planId()
+       << "/indexes";
+    auto& ac = _vocbase.server().getFeature<ClusterFeature>().agencyCache();
+    auto [plannedIndexes, idx] = ac.get(ap.str());
```
jsteemann, line 204: I think we must not execute this code if we are not on a coordinator. At least it should not be executed on a single server, most likely also not on DB servers.

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
             VPackValue(static_cast<int>(ResponseCode::OK)));
-    tmp.add("indexes", indexes.slice());
+    tmp.add(VPackValue("indexes"));
+    {
+      VPackArrayBuilder guard(&tmp);
+      for (auto const& i : VPackArrayIterator(indexes.slice())) {
+        tmp.add(i);
+      }
+      for (auto const& pi : VPackArrayIterator(plannedIndexes->slice())) {
+        if (pi.hasKey("isBuilding") && pi.get("isBuilding").getBool()) {
```
jsteemann, line 219: can be simplified to just
```suggestion
        if (pi.hasKey("isBuilding").isTrue()) {
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+          VPackObjectBuilder o(&tmp);
+          for (auto const& source : VPackObjectIterator(pi)) {
+            tmp.add(source.key.copyString(), source.value);
+          }
```
jsteemann, line 223: can be improved to
```suggestion
          for (auto source : VPackObjectIterator(pi, /*useSequentialIteration*/ true)) {
            tmp.add(source.key.stringView(), source.value);
          }
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+          double progress = 0;
+          double success = 0;
+          auto const shards = coll->shardIds();
+          auto const body = VPackBuffer<uint8_t>();
+          auto* pool =
+              coll->vocbase().server().getFeature<NetworkFeature>().pool();
+          std::vector<Future<network::Response>> futures;
+          futures.reserve(shards->size());
+          std::string const prefix = "/_api/index/";
+          network::RequestOptions reqOpts;
```
jsteemann, line 234: The requests will always be executed inside the system database, which will not work if the underlying collection is executed in a different database.
The setup of the RequestOptions is missing the `database` attribute here, i.e. `reqOptions.database = _vocbase.name();`

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+          reqOpts.param("withHidden", "true");
+          // best effort. only displaying progress
+          reqOpts.timeout = network::Timeout(10.0);
+          for (auto const& shard : *shards) {
+            std::string const url = prefix + shard.first + "/" + iid;
```
jsteemann, line 239: Better alternative:
```suggestion
            std::string const url = absl::StrCat(prefix, shard.first, "/", pi.get("id").stringView());
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+                reqOpts));
+          }
+          for (Future<network::Response>& f : futures) {
+            network::Response const& r = f.get();
+
+            // Only best effort accounting. If something breaks here, we just
+            // ignore the output. Account for what we can and move on.
+            if (r.fail()) {
+              LOG_TOPIC("afde4", INFO, Logger::CLUSTER)
+                  << "Communication error while collecting figures for "
```
jsteemann, line 251:
```suggestion
                  << "Communication error while fetching index data for "
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+            if (!resSlice.isObject() ||
+                !resSlice.get(StaticStrings::Error).isBoolean()) {
+              LOG_TOPIC("aabe4", INFO, Logger::CLUSTER)
+                  << "Result of collecting figures for collection "
+                  << coll->name() << " from " << r.destination << " is invalid";
+              continue;
+            }
+            if (resSlice.get(StaticStrings::Error).getBoolean()) {
+              LOG_TOPIC("a4bea", INFO, Logger::CLUSTER)
+                  << "Failed to collect figures for collection " << coll->name()
```
jsteemann, line 266:
```suggestion
                  << "Failed to collect index data for collection " << coll->name()
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+                  << " from " << r.destination;
+              continue;
+            }
+            if (resSlice.get("progress").isNumber()) {
+              progress += resSlice.get("progress").getNumber<double>();
+              success++;
```
jsteemann, line 272: I see that the `success` variable is increased here only, but apart from that it is not used.
Do we need to divide the total progress value by the number of shards that replied successfully? I guess if we don't do this, the progress value could exceed 100% easily.

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
...
+            } else {
+              LOG_TOPIC("aeab4", INFO, Logger::CLUSTER)
+                  << "No progress entry on index " << iid << "  from "
+                  << r.destination << ": " << resSlice.toJson();
+            }
+          }
+          tmp.add("progress", VPackValue(progress));
```
jsteemann, line 279:
```suggestion
          if (success > 0) {
            tmp.add("progress", VPackValue(progress / success));
          }
```

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -152,14 +155,38 @@ Result fillIndexSingleThreaded(
...
+            docsProcessed.load(std::memory_order_relaxed) * 100.0 / count;
+        ridx.progress(p);
+        if (progress != nullptr) {
+          (*progress)(p);
+        }
+      }
+
+#ifdef ARANGODB_ENABLE_FAILURE_TESTS
+      TRI_IF_FAILURE("fillIndex::pause") {
+        std::cout << "fillIndex::pause" << std::endl;
```
jsteemann, line 179: The print statements here should really be removed, even though they are only included inside a failure point.

tests/js/common/shell/shell-index.js
```diff
@@ -55,6 +55,27 @@ function IndexSuite() {
...
+      internal.debugSetFailAt("fillIndex::pause");
+      c.ensureIndex({ type: "persistent", fields:["name"], inBackground: true});
+      //arango.POST("_api/index?collecion=c0l", { type: "persistent", fields:["name"], inBackground: true}, {"x-arango-async": "store"})
+      console.error("hu");
+      require('internal').sleep(1000);
+      console.error("ha");
+      internal.debugSetFailAt("fillIndex::unpause");
+      internal.debugRemoveFailAt("fillIndex::pause");
+      internal.debugRemoveFailAt("fillIndex::unpause");
+      require('internal').sleep(1000);
```
jsteemann, line 75: While this test may work when running it manually and watching its output, it won't help us in any automated test runs.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
neunhoef, line 179: This will eventually have to be removed.

arangod/RocksDBEngine/RocksDBBuilderIndex.cpp
```diff
@@ -152,14 +155,38 @@ Result fillIndexSingleThreaded(
...
+        while (true) {
+          std::this_thread::sleep_for(std::chrono::milliseconds(3000));
+          // TRI_IF_FAILURE("fillIndex::unpause") {
```
neunhoef, line 182: This is not ready. We need the two failure points for controlled unpause.

tests/js/common/shell/shell-index.js
```diff
@@ -55,6 +55,27 @@ function IndexSuite() {
...
+    testIndexProgress : function () {
+      var c = internal.db._create("c0l");
+      var docs = [];
+      for (var i = 0; i < 10000; ++i) {
+        docs.push({name : "name" + i});
+      }
+      c.insert(docs);
+      internal.debugSetFailAt("fillIndex::pause");
+      c.ensureIndex({ type: "persistent", fields:["name"], inBackground: true});
+      //arango.POST("_api/index?collecion=c0l", { type: "persistent", fields:["name"], inBackground: true}, {"x-arango-async": "store"})
```
neunhoef, line 68: This isn't finished.

neunhoef (changes requested):
First read done, some parts missing, in particular in the testing.

neunhoef (changes requested):
I have tried this out manually. For indexes with `inBackground::true` it now seems to work in the cluster. Without `inBackground:true` progress is not reported, just the `isBuilding` index, at least, from what I seem to see, progress is at first 0 for a long time, then it grows (for 6 shards) towards the end to 599.999 or so and then it vanishes when the index is ready.
Furthermore, in both cases, the hidden index is (on the coordinator) only reported in the `indexes` attribute and not in the `identifiers` object. This should behave consistently.
In single server, the APi does not seem to work at all currently, I get this response:
```{
  "code": 500,
  "error": true,
  "errorMessage": "VPack error: Expecting Array slice (exception location: /home/neunhoef/ArangoDB310/arangod/GeneralServer/RestHandler.cpp:548). Please reportthis error to arangodb.com",
  "errorNum": 4
}``` I believe this is because the new code must only be executed on coordinators.

All in all, this is not yet finished.

File: arangod/Cluster/ActionBase.cpp [hunk 16]
```diff
@@ -171,7 +171,9 @@ void ActionBase::startStats() {
 
 /// @brief show progress on Action, and when that progress occurred
 void ActionBase::incStats() {
-  ++_progress;
+  auto progress = _progress.load();
```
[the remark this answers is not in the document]
kvahed: yes sir

arangod/RestHandler/RestIndexHandler.cpp
kvahed: on response is divided by number of servers.

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -193,13 +197,89 @@ RestStatus RestIndexHandler::getIndexes() {
       return RestStatus::DONE;
     }
 
+    std::stringstream ap;
```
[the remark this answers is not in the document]
kvahed: This was at least the plan originally. Only indexes created in background and only, when isHidden is true. Can of course be changed.

arangod/Indexes/Index.h
```diff
@@ -440,6 +440,13 @@ class Index {
 
   static size_t sortWeight(aql::AstNode const* node);
 
+  void progress(double p) noexcept {
```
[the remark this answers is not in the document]
neunhoef: OK, not really necessary.

arangod/RestHandler/RestIndexHandler.cpp
neunhoef: Done in the new place.

arangod/Cluster/ActionBase.cpp
```diff
@@ -171,9 +171,12 @@ void ActionBase::startStats() {
 
 /// @brief show progress on Action, and when that progress occurred
 void ActionBase::incStats() {
-  ++_progress;
+  // OSX clang cannot increment double very rare calls
+  std::mutex m;
```
neunhoef, line 175: Using a local variable of type std::mutex is nonsense. This protects nothing at all. Use a mutex as a member variable and please use a mutex guard for locking.

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -199,7 +205,97 @@ RestStatus RestIndexHandler::getIndexes() {
...
+                } else {
+                  // Obviously, the index is already ready there.
+                  progress += 100.0;
+                  LOG_TOPIC("aeab4", DEBUG, Logger::CLUSTER)
+                      << "No progress entry on index " << iid << "  from "
+                      << r.destination << ": " << resSlice.toJson()
+                      << " index already finished.";
+                }
+              }
+              tmp.add("progress", VPackValue(progress / shards->size()));
```
neunhoef, line 289: Let's add this only if `progress` is non-zero, since otherwise we will always see 0 in the cluster when `inBackground` is set to `false`.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,116 @@
...
+          assertTrue(idx.progress >= progress);
+          progress = idx.progress;
+          print(progress.toPrecision(4) + "%");
+          sleep(0.1);
+        } else {
+          break;
+        }
+      }
+
+      internal.debugRemoveFailAt("fillIndex::pause");
```
neunhoef, line 107: These two should go into a finally clause, otherwise they could remain behind if an exception is thrown.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,116 @@
...
+      internal.debugSetFailAt("fillIndex::pause");
+      var idx = { name:"progress", type: "persistent", fields:["name"], inBackground: true};
+      arango.GET(`/_api/index?collection=c0l`);
+
+      var job = arango.POST_RAW(`/_api/index?collection=c0l`, idx,
+                                {"x-arango-async": "store"}).headers["x-arango-async-id"];
+      let count = 0;
+      var progress = 0.0;
+      var idxs = [];
+      while (true) {
```
neunhoef, line 86: This whole test will not fail if one day no progress is reported at all. in fact, if would not have failed without this feature. The reason is that no assertion ever in here requires that `progress` is actually set!
We agreed to do this differently: We said that we set the pause failure point and then wait in the test until at least for once we see `progress` with a value between 0 and 100 (exclusive). Then we would release the pause with the second failure point and let it finish. In the meantime we can still check and if we see `progress`, it must be at least as large as the previous time (check non-decreasing). We should also check that `isBuilding` is set if and only if `progress` is set.

tests/js/common/shell/shell-index.js
```diff
@@ -32,6 +32,7 @@ const errors = internal.errors;
 const { helper, versionHas } = require("@arangodb/test-helper");
 const platform = require('internal').platform;
 
+
```
neunhoef, line 35: This is an unnecessary change.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,116 @@
...
+      const sleep = require('internal').sleep;
+      for (var j = 0; j < 16; ++j) {
+        docs=[];
+        for (var i = 0; i < 10240; ++i) {
+          docs.push({name : "name" + i});
+        }
+        c.insert(docs);
+      }
+      
+      internal.debugSetFailAt("fillIndex::pause");
```
neunhoef, line 77: This will work in single server, because `debugSetFailAt` will set the failure point on the point of contact, which is the coordinator. For the cluster, it is more difficult: We have to first find the endpoints of all dbservers and then use the two argument version in `test-helpers` to activate the failure points there. And likewise in the end switch off the failure points again. I bet that on cluster this only works since the test can never fail.

neunhoef (changes requested):
See further comments, the test is still not all right.

neunhoef (approved):
LGTM

arangod/Cluster/ActionBase.cpp
```diff
@@ -171,9 +171,10 @@ void ActionBase::startStats() {
 
 /// @brief show progress on Action, and when that progress occurred
 void ActionBase::incStats() {
-  ++_progress;
+  // OSX clang cannot increment double very rare calls
+  std::lock_guard<std::mutex> guard(_progressMutex);
+  _progress = _progress + 1.0;
   _actionLastStat = secs_since_epoch();
```
jsteemann, line 177: nit: `_actionLastStat` is now updated under the mutex, but read without it.
This is not racy, as `_actionLastStat` is an atomic variable. However, technically it is not necessary to update it under the mutex, and it is slightly confusing to readers of the code.

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -199,7 +205,102 @@ RestStatus RestIndexHandler::getIndexes() {
     tmp.add(StaticStrings::Error, VPackValue(false));
     tmp.add(StaticStrings::Code,
             VPackValue(static_cast<int>(ResponseCode::OK)));
-    tmp.add("indexes", indexes.slice());
+
+    if (ServerState::instance()->isCoordinator()) {
+      std::string ap =
+          absl::StrCat("Plan/Collections/", _vocbase.name(), "/",
+                       std::to_string(coll->planId().id()), "/indexes");
```
jsteemann, line 212: The extra `std::to_string(number)` is not necessary here: absl::StrCat() handles that automatically.
```suggestion
                       coll->planId().id(), "/indexes");
```

arangod/RestHandler/RestIndexHandler.cpp
```diff
@@ -199,7 +205,102 @@ RestStatus RestIndexHandler::getIndexes() {
...
+                      << "No progress entry on index " << iid << "  from "
+                      << r.destination << ": " << resSlice.toJson()
+                      << " index already finished.";
+                }
+              }
+              if (progress != 0) {
+                // Don't show progress 0, this is in particular relevant
+                // when isBackground is false, in which case no progress
+                // is reported by design.
+                tmp.add("progress", VPackValue(progress / shards->size()));
```
jsteemann, line 293: nit: if one of the shards fails to answer, dividing by `shards->size()` may underreport the actual progress.
I don't have a better solution here, except counting the number of shards for which we successfully received a response. But as the progress value may be skewed anyway in this case, it does not matter much.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+        for (let i = 0; i < 10240; ++i) {
+          docs.push({name : "name" + i});
+        }
+        c.insert(docs);
+      }
+      
+      try {
+        internal.debugSetFailAt("fillIndex::pause");
+        let idx = { name:"progress", type: "persistent", fields:["name"], inBackground: true};
+        arango.GET(`/_api/index?collection=c0l`);
```
jsteemann, line 63: I don't think this HTTP GET has any purpose in the current version of the code, and thus probably can be removed.
```suggestion
```

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+        let progress = 0.0;
+        let idxs = [];
+        // Wait until the index is there (with withHidden):
+        while (true) {
+          idxs = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes;
+          if (idxs.length > 1) {
+            break;
+          }
+          sleep(0.1);
+        }
```
jsteemann, line 77: Nit: this will be an unbounded loop if the index doesn't show up for whatever reason.
We could make the loop only try for a certain amount of time or a certain number of attempts to retrieve the index info.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+        let job = arango.POST_RAW(`/_api/index?collection=c0l`, idx,
+                                  {"x-arango-async": "store"}).headers["x-arango-async-id"];
+        let count = 0;
+        let progress = 0.0;
+        let idxs = [];
+        // Wait until the index is there (with withHidden):
+        while (true) {
+          idxs = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes;
+          if (idxs.length > 1) {
+            break;
```
jsteemann, line 74: We should assert here that the index data we have got actually is for the correct index, by peeking into `idxs[1]` and validating its contents.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+      try {
+        internal.debugSetFailAt("fillIndex::pause");
+        let idx = { name:"progress", type: "persistent", fields:["name"], inBackground: true};
+        arango.GET(`/_api/index?collection=c0l`);
+
+        let job = arango.POST_RAW(`/_api/index?collection=c0l`, idx,
+                                  {"x-arango-async": "store"}).headers["x-arango-async-id"];
+        let count = 0;
+        let progress = 0.0;
+        let idxs = [];
```
jsteemann, line 69: nit: `idxs` is a variable that is only used inside the following loop. So it should not be declared outside of the loop.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+          idxs = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes;
+          if (idxs.length > 1) {
+            break;
+          }
+          sleep(0.1);
+        }
+
+        let unpaused = false;
+        while (true) {
+          idx = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes[1];
```
jsteemann, line 81: nit: the `idx` variable that was declared ~20 lines above is used here for some different purpose.
I suggest making this a local variable inside this loop, so the variable scope & purpose become more clear.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+          if (idx.hasOwnProperty("progress")) {
+            assertTrue(idx.progress >= progress);
+            assertTrue(idx.isBuilding);
```
jsteemann, line 84: nit: in case the assertions don't hold true, let's already provide all relevant information for debugging.
```suggestion
            assertTrue(idx.progress >= progress, {idx, progress});
            assertTrue(idx.isBuilding, idx);
```

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+        while (true) {
+          idxs = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes;
+          if (idxs.length > 1) {
+            break;
+          }
+          sleep(0.1);
+        }
+
+        let unpaused = false;
+        while (true) {
```
jsteemann, line 80: nit: This loop is only left by `break` once certain conditions become true.
If these conditions do not become true for whatever reason, this loop is unbounded. We should limit the amount of tries or the total time that can spent inside the loop.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+          idx = arango.GET(`/_api/index?collection=c0l&withHidden=true`).indexes[1];
+          if (idx.hasOwnProperty("progress")) {
```
jsteemann, line 82: Are there any expectations as to how often we should get into this branch? Right now it may be executed 0 times, and the test will still pass.
Do we expect a minimum amount of times this particular branch is executed? Then we should count the number of invocations and add an assertion for that following the loop body.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+///
+/// @author Kaveh Vahedipour
+////////////////////////////////////////////////////////////////////////////////
+
+const jsunity = require("jsunity");
+const internal = require("internal");
+const arango = internal.arango;
+const errors = internal.errors;
+const { helper, versionHas } = require("@arangodb/test-helper");
+const platform = require('internal').platform;
```
jsteemann, line 33: not used
```suggestion
```

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+      for (let j = 0; j < 16; ++j) {
+        docs=[];
+        for (let i = 0; i < 10240; ++i) {
+          docs.push({name : "name" + i});
+        }
+        c.insert(docs);
+      }
+      
+      try {
+        internal.debugSetFailAt("fillIndex::pause");
```
jsteemann, line 61: suggestion: use `setUp` and `tearDown` functions instead. This will also unclutter the actual test function.

arangod/Cluster/ActionBase.cpp
neunhoef: Will fix.

arangod/RestHandler/RestIndexHandler.cpp
neunhoef: Cool, I did not know this. Will fix.

arangod/RestHandler/RestIndexHandler.cpp
neunhoef: This is correct, but unlikely to occur. Dividing only by the number of correct results also has issues. Let's leave it like it is for now.

tests/js/client/shell/shell-index.js
neunhoef: The expectation is that we see intermediate progress at least once. I have added an assertion after the loop.

tests/js/client/shell/shell-index.js
```diff
@@ -0,0 +1,109 @@
...
+          } else {
+            assertFalse(idx.hasOwnProperty("isBuilding"));
+            break;
+          }
+        }
+      } finally {
+        internal.debugRemoveFailAt("fillIndex::pause");
+        internal.debugRemoveFailAt("fillIndex::unpause");
+        c.drop();
+      }
```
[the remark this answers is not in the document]
neunhoef: Done.

jsteemann (approved):
LGTM

PR #20067, opened 2023-11-03
Implement SupervisedScheduler::detachThread [BTS-1670]

neunhoef:
This is a backport of https://github.com/arangodb/arangodb/pull/20036

This is one measure to resolve
  https://arangodb.atlassian.net/browse/BTS-1670

In this PR we create a possiblity that a SchedulerThread can notice that 
it might be up to some longer running task (like waiting for a lock) and 
then voluntarily decide to detach itself from the scheduler to make room
for a new thread to avoid starvation of jobs on the queue.

For now, we use this for acquiring a collection level lock (as read lock 
for write operations and as write lock for exclusive access), if we 
cannot acquire the lock within 1s. This will then allow the scheduler
to create new threads to continue work on whatever priority is suitable.

### Scope & Purpose

*(Please describe the changes in this PR for reviewers, motivation, rationale - **mandatory**)*

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] C++ **Unit tests**
  - [*] **integration tests**
- [*] :book: CHANGELOG entry made
- [*] Backports
  - [*] Backport for 3.11: This is it.

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1670
- [*] devel PR: https://github.com/arangodb/arangodb/pull/20036

PR #20411, opened 2024-01-05
Implement SupervisedScheduler::detachThread [BTS-1670]

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/20067

This is one measure to resolve https://arangodb.atlassian.net/browse/BTS-1670

In this PR we create a possiblity that a SchedulerThread can notice that it might be up to some longer running task (like waiting for a lock) and then voluntarily decide to detach itself from the scheduler to make room for a new thread to avoid starvation of jobs on the queue.

For now, we use this for acquiring a collection level lock (as read lock for write operations and as write lock for exclusive access), if we cannot acquire the lock within 1s. This will then allow the scheduler to create new threads to continue work on whatever priority is suitable.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20067

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
+++ b/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
@@ -1643,6 +1643,8 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
       // keep the lock and exit
       return TRI_ERROR_NO_ERROR;
     }
+
+    timeout_us -= detachThreshold;
     LOG_TOPIC("dd231", INFO, Logger::THREADS)
         << "Did not get lock within 1 seconds, detaching scheduler thread.";
     uint64_t currentNumberDetached = 0;
@@ -1659,8 +1661,6 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
              "blockages!";
     }
   }
-
-  timeout_us -= detachThreshold;
 
   if (mode == AccessMode::Type::WRITE) {
     gotLock = _exclusiveLock.tryLockWriteFor(timeout_us);
--- a/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
+++ b/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
@@ -42,6 +42,7 @@
 #include "RocksDBEngine/RocksDBSettingsManager.h"
 #include "RocksDBEngine/RocksDBTransactionCollection.h"
 #include "RocksDBEngine/RocksDBTransactionState.h"
+#include "Scheduler/SchedulerFeature.h"
 #include "StorageEngine/EngineSelectorFeature.h"
 #include "Transaction/Context.h"
 #include "Transaction/Methods.h"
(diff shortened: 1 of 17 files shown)


maierlars (approved):

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -485,6 +492,59 @@ void SupervisedScheduler::shutdown() {
...
+  // Now we have access to the _workerStates and _detachedWorkerStates
+  // Let's first find ourselves in the _workerStates:
+  uint64_t myNumber = Thread::currentThreadNumber();
+  auto it = _workerStates.begin();
+  while (it != _workerStates.end()) {
+    if ((*it)->_thread->threadNumber() == myNumber) {
+      break;
+    }
+    ++it;
+  }
```
mpoeter, line 518: This could be simplified (but still needs to be clang formatted)
```suggestion
  auto it = std::find_if(_workerStates.begin(), _workerStates.end(), 
    [&](auto& v) { return v._thread->threadNumber() == myNumber; });
```

arangod/Scheduler/SupervisedScheduler.cpp
```diff
@@ -485,6 +492,59 @@ void SupervisedScheduler::shutdown() {
...
+  std::shared_ptr<WorkerState> state = std::move(*it);
+  _workerStates.erase(it);
+  // Since the thread is effectively taken out of the pool, decrease the
+  // number of workers.
+  --_numWorkers;
+  {
+    std::unique_lock<std::mutex> guard(state->_mutex);
+    state->_stop = true;  // We will be stopped after the current task is done
+                          // We know that we are working, so we do not
+                          // have to wake the thread.
```
mpoeter, line 532: This lock should not be necessary since stop is an atomic.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1629,7 +1630,38 @@ ErrorCode RocksDBMetaCollection::doLock(double timeout, AccessMode::Type mode) {
...
 
+  constexpr auto detachThreshold = std::chrono::microseconds(1000000);
+
   bool gotLock = false;
+  if (timeout_us > detachThreshold) {
+    if (mode == AccessMode::Type::WRITE) {
+      gotLock = _exclusiveLock.tryLockWriteFor(detachThreshold);
+    } else {
+      gotLock = _exclusiveLock.tryLockReadFor(detachThreshold);
+    }
```
mpoeter, line 1641: What it `timeout` is less than 1s? Then this could would still block for 1s. And if we it is greater than 1s but we do not get the lock hear, then the later tryLock does not account for the time spent here already, so in the worst case we block for `timeout + 1s`. Is that intentional?

kvahed (approved):
LGTM

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
jsteemann: There was already a subtraction of `detachThreshold` from `timeout_us`, but IMHO it happened at the wrong location, so it could have made the timeout negative. I pushed a fixed for this now.

arangod/Scheduler/SupervisedScheduler.cpp
jsteemann: Let's do the beautification only in devel, but let's not delay the backport for it.

arangod/Scheduler/SupervisedScheduler.cpp
jsteemann: Same here (devel only).

mpoeter (approved):

neunhoef (approved):
LGTM

PR #20413, opened 2024-01-05
[ZKD] BugFix PR

maierlars:
### Scope & Purpose

This PR fixes a few smaller bugs:
- Indexes did not compare correctly and did not account for `prefixFields`.
- `fieldValueTypes: double` was required for creation but not stored when serialized.

### DIFF
--- a/tests/js/client/aql/aql-zkdindex-prefix-fields.js
+++ b/tests/js/client/aql/aql-zkdindex-prefix-fields.js
@@ -26,7 +26,7 @@ const jsunity = require("jsunity");
 const arangodb = require("@arangodb");
 const db = arangodb.db;
 const aql = arangodb.aql;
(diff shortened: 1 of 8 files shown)


tests/js/client/aql/aql-zkdindex-prefix-fields.js
```diff
@@ -35,9 +35,10 @@ function optimizerRuleZkd2dIndexTestSuite() {
   let col;
 
   return {
-
     tearDown: function () {
-      db[colName].drop();
+      if (db[colName]) {
+        db[colName].drop();
+      }
```
jsteemann, line 41: nit: can be simplified to just
```suggestion
      db._drop(colName);
```
`db._drop()` will not fail if no such collection exists.

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #20423, opened 2024-01-05
Rename ZKD to MDI

maierlars:
### Scope & Purpose
The zkd index has been renamed to `mdi`.

Some things could not be renamed, for example the old `zkdIndex` column family.

Also the zkd library was not renamend, because it is closely tied to the z curve.


jsteemann (approved):
LGTM

PR #20425, opened 2024-01-05
Cleanup dump contexts in tests.

mpoeter:
### Scope & Purpose

Otherwise the context holds the collection lock which prevents us from dropping the database.

- [x] :hankey: Bugfix


mchacki (approved):
LGTM

PR #20439, opened 2024-01-09
Prometheus labels should be snake_case

mpoeter:
### Scope & Purpose

... instead of camelCase

Running Jenkins on this because the CircleCI test image currently does not contain the promtool (see #20440)

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #20440, opened 2024-01-09
Add promtool and net-tools to test container

mpoeter:
### Scope & Purpose

The promtool is used in one of our js tests to validate our metrics against the prometheus format. If this tool is missing, the test is simply not executed.

This PR add does not install a complete prometheus, but only adds the promtool, and only for the amd64 image. I suppose it is enough if this test runs for x64 only (the metrics should be the same for all architectures).

It also installs net-tools since the testing framework tries to start netstat to gather some information about resource usage.


jsteemann (approved):
LGTM

PR #20441, opened 2024-01-09
Keep zkd index type.

maierlars:
### Scope & Purpose

Fix backward compatibility issues by keeping the zkd index type.


jsteemann (approved):
LGTM

PR #20447, opened 2024-01-10
Get rid of getpwuid and friends to avoid nsswitch

neunhoef:
This PR addresses the problem that our glibc-based static binaries
so far us functions which are governed by the /etc/nsswitch.conf .
Namely, so far we use the following functions:
  getpwuid
  getpwnam
  getgrgid
  getgrnam
These have the potential to crash static binaries when running on older
Linux distributions. It turns out that we can replace all uses of These
functions by a call to the external `id` and `getent` executables, which
are available in all Linux distributions. For Mac, `id` works as well,
but for `getgrgid` and `getgrnam` we stay with the old functions,
because `getent` is not available. For Windows, there is no change since
all features these functions are used for are not available anyway.

No backports are necessary, since for older versions we will only ever
publish Docker images with glibc-based static binaries.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist


lib/Basics/FileUtils.cpp
```diff
@@ -804,4 +804,100 @@ std::string slurpProgram(std::string const& program) {
...
+    std::vector<std::string> args{"-u", nameOrId};
+    std::string output = slurpProgramInternal("/usr/bin/id", args);
+    StringUtils::trimInPlace(output);
+    bool valid = false;
+    uid_t uidNumber = NumberUtils::atoi_positive<int>(
+        output.data(), output.data() + output.size(), valid);
+    if (valid) {
+      return {uidNumber};
+    }
+  } catch (std::exception const& exc) {
```
jsteemann, line 830: I have seen MSVC complaining about unused exception variables before. Better not use it.
```suggestion
  } catch (std::exception const&) {
```

lib/Basics/FileUtils.cpp
```diff
@@ -804,4 +804,100 @@ std::string slurpProgram(std::string const& program) {
...
+  // /etc/nsswitch.conf).
+  try {
+    std::vector<std::string> args{"passwd", std::to_string(id)};
+    std::string output = slurpProgramInternal("/usr/bin/getent", args);
+    StringUtils::trimInPlace(output);
+    auto parts = StringUtils::split(output, ':');
+    if (parts.size() >= 1) {
+      return {std::move(parts[0])};
+    }
+  } catch (std::exception const& exc) {
```
jsteemann, line 854:
```suggestion
  } catch (std::exception const&) {
```

lib/Basics/FileUtils.cpp
```diff
@@ -804,4 +804,100 @@ std::string slurpProgram(std::string const& program) {
...
+    auto parts = StringUtils::split(output, ':');
+    if (parts.size() >= 3) {
+      bool valid = false;
+      uid_t gidNumber = NumberUtils::atoi_positive<int>(
+          parts[2].data(), parts[2].data() + parts[2].size(), valid);
+      if (valid) {
+        return {gidNumber};
+      }
+    }
+  } catch (std::exception const& exc) {
```
jsteemann, line 897:
```suggestion
  } catch (std::exception const&) {
```

lib/Basics/FileUtils.cpp
```diff
@@ -756,17 +756,17 @@ void makePathAbsolute(std::string& path) {
   }
 }
 
-std::string slurpProgram(std::string const& program) {
+namespace {
+
+std::string slurpProgramInternal(std::string const& program,
+                                 std::vector<std::string>& moreArgs) {
```
jsteemann, line 762: That "moreArgs" is passed by mutable reference makes this API look a bit weird. Why not use a const reference for it?
```suggestion
                                 std::vector<std::string> const& moreArgs) {
```

lib/Basics/FileUtils.h
```diff
@@ -154,4 +164,16 @@ std::string dirname(std::string const&);
 // returns the output of a program
 std::string slurpProgram(std::string const& program);
 
+// returns the output of a program
+std::string slurpProgramWithArgs(std::string const& program,
+                                 std::vector<std::string> args);
+
```
jsteemann, line 170: I think in the changes in FileUtils.cpp, this function has not been added, so it probably can be removed from here now.
```suggestion
```

jsteemann (approved):
LGTM

jsteemann (approved):

PR #20453, opened 2024-01-10
Get rid of initgroups avoid nsswitch

neunhoef:
This PR addresses another problem that our glibc-based static binaries
so far use functions which are governed by the /etc/nsswitch.conf .
Namely, so far we use the following function:
  initgroups
These have the potential to crash static binaries when running on older
Linux distributions. It turns out that we can replace all uses of this
function by a call to the external `groups`, followed by a call to
`setgroups`, which is OK. `groups` is available in all Linux distributions.
For other Unixes (including Mac), we can keep `initgroups`.
For Windows, there is no change since all features these functions are
used for are not available anyway.

No backports are necessary, since for older versions we will only ever
publish Docker images with glibc-based static binaries.

### Scope & Purpose

- [*] :hankey: Bugfix


lib/Basics/FileUtils.cpp
```diff
@@ -900,4 +900,38 @@ std::optional<gid_t> findGroup(std::string const& nameOrId) noexcept {
...
+#ifdef ARANGODB_HAVE_INITGROUPS
+void initGroups(std::string const& userName, gid_t groupId) noexcept {
+#ifdef __linux__
+  // For Linux, calling initgroups poses problems with statically linked
+  // binaries, since /etc/nsswitch.conf can then lead to crashes on
+  // older Linux distributions. Therefore, we need to do the groups lookup
+  // ourselves using the groups command. Then we can use setgroups to
+  // achieve the desired result.
+  try {
+    std::vector<std::string> args{"groups", userName};
```
jsteemann, line 913: Is it necessary to pass the extra `groups` argument to the `/usr/bin/groups` executable? On my local Linux this does not work, but the user name needs to be passed without additional parameters.
```suggestion
    std::vector<std::string> args{userName};
```

jsteemann (approved):
LGTM

PR #20463, opened 2024-01-11
Fix: counting with tombstones in iresearch can produce wrong results

mpoeter:
### Scope & Purpose

In replication2 recovery we can replay the inserts + truncate, but the truncate is then executed with normal remove operations instead of a range-delete. The resulting tombstones in iresearch can lead to incorrect counting. This can be circumvented by returning the complete documents which uses late materialization.

### DIFF
--- a/tests/js/client/recovery/search/view-arangosearch-link-populate-truncate-no-flushthread.js
+++ b/tests/js/client/recovery/search/view-arangosearch-link-populate-truncate-no-flushthread.js
@@ -77,14 +77,17 @@ function recoverySuite () {
     // //////////////////////////////////////////////////////////////////////////////
 
     testIResearchLinkPopulateTruncate: function () {
-      var v = db._view(vn);
+      const v = db._view(vn);
       assertEqual(v.name(), vn);
       assertEqual(v.type(), 'arangosearch');
-      var p = v.properties().links;
+      const p = v.properties().links;
       assertTrue(p.hasOwnProperty(cn));
-      var result = db._query("FOR doc IN " + vn + " SEARCH doc.c >= 0 OPTIONS {waitForSync: true} COLLECT WITH COUNT INTO length RETURN length").toArray();
-      var expectedResult = db._query("FOR doc IN " + cn + " FILTER doc.c >= 0 COLLECT WITH COUNT INTO length RETURN length").toArray();
-      assertEqual(result[0], expectedResult[0]);
+      // For the search query we do not use COLLECT WITH COUNT, because we have
+      // need late materialization to ensure proper counting. COLLECT WITH COUNT
+      // can return wrong results if the truncate used normal removes.
+      const result = db._query("FOR doc IN " + vn + " SEARCH doc.c >= 0 OPTIONS {waitForSync: true} RETURN doc._key").toArray();
+      const expectedResult = db._query("FOR doc IN " + cn + " FILTER doc.c >= 0 COLLECT WITH COUNT INTO length RETURN length").toArray();
+      assertEqual(result.length, expectedResult[0]);
     }
  };
 }
--- a/tests/js/client/recovery/search/view-arangosearch-link-populate-truncate.js
+++ b/tests/js/client/recovery/search/view-arangosearch-link-populate-truncate.js
@@ -73,14 +73,17 @@ function recoverySuite () {
     // //////////////////////////////////////////////////////////////////////////////
 
     testIResearchLinkPopulateTruncate: function () {
-      var v = db._view(vn);
+      const v = db._view(vn);
       assertEqual(v.name(), vn);
       assertEqual(v.type(), 'arangosearch');
-      var p = v.properties().links;
+      const p = v.properties().links;
       assertTrue(p.hasOwnProperty(cn));
-      var result = db._query("FOR doc IN " + vn + " SEARCH doc.c >= 0 OPTIONS {waitForSync: true} COLLECT WITH COUNT INTO length RETURN length").toArray();
-      var expectedResult = db._query("FOR doc IN " + cn + " FILTER doc.c >= 0 COLLECT WITH COUNT INTO length RETURN length").toArray();
-      assertEqual(result[0], expectedResult[0]);
+      // For the search query we do not use COLLECT WITH COUNT, because we have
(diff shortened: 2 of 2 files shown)


tests/js/client/recovery/search/view-arangosearch-link-populate-truncate-no-flushthread.js
```diff
@@ -77,14 +77,17 @@ function recoverySuite () {
...
+      const v = db._view(vn);
       assertEqual(v.name(), vn);
       assertEqual(v.type(), 'arangosearch');
-      var p = v.properties().links;
+      const p = v.properties().links;
       assertTrue(p.hasOwnProperty(cn));
-      var result = db._query("FOR doc IN " + vn + " SEARCH doc.c >= 0 OPTIONS {waitForSync: true} COLLECT WITH COUNT INTO length RETURN length").toArray();
-      var expectedResult = db._query("FOR doc IN " + cn + " FILTER doc.c >= 0 COLLECT WITH COUNT INTO length RETURN length").toArray();
-      assertEqual(result[0], expectedResult[0]);
+      // For the search query we do not use COLLECT WITH COUNT, because we have
```
mchacki, line 85:
```suggestion
      // For the search query we do not use COLLECT WITH COUNT, because we
```

tests/js/client/recovery/search/view-arangosearch-link-populate-truncate.js
```diff
@@ -73,14 +73,17 @@ function recoverySuite () {
...
+      const v = db._view(vn);
       assertEqual(v.name(), vn);
       assertEqual(v.type(), 'arangosearch');
-      var p = v.properties().links;
+      const p = v.properties().links;
       assertTrue(p.hasOwnProperty(cn));
-      var result = db._query("FOR doc IN " + vn + " SEARCH doc.c >= 0 OPTIONS {waitForSync: true} COLLECT WITH COUNT INTO length RETURN length").toArray();
-      var expectedResult = db._query("FOR doc IN " + cn + " FILTER doc.c >= 0 COLLECT WITH COUNT INTO length RETURN length").toArray();
-      assertEqual(result[0], expectedResult[0]);
+      // For the search query we do not use COLLECT WITH COUNT, because we have
```
mchacki, line 81:
```suggestion
      // For the search query we do not use COLLECT WITH COUNT, because we
```

mchacki (approved):
LGTM

MBkkt (changes requested):
I don't think it's desired change, because 
1. We have recovery tests with common removes
2. If I understand right and this change (materialize from rocksdb snapshots) helps you with tests, replication2 reintroduce issue with inconsistency between search and rocksdb snapshots.

mchacki:
1.) Yes we know for replication1 there is a difference between Recovery on truncate and recovery on removes, that is explicitly tested.

For Replication2 however the Recovery is actually in two steps:
a) RocksDBWAL recovery, which includes Search (and everything else) as in Replication1.
b) Then we still have entries in-flight for the Replication2 Log, which are recovered on top of the RocksDB.

For step a) the RocksDBSnapshot and Search are fully in-sync.
If before the forced crash, the replicated log is compacted up to the point that it does not contain the "truncate" anymore this test is green.
If the log is not compacted yet, we need to do the recovery b), during this operation we on purpose do not hold the exclusive lock, hence truncate will be executed like single removes.

=> So I think RocksDB Snapshot and Search are consistent before Replication2 recovery.
From there on they are in the eventual consistency area.


How was the inconstancy of snapshot & search solved before? 
Maybe we can try and do the same thing during ReplicatedLog recovery.

MBkkt:
@mchacki 

> How was the inconstancy of snapshot & search solved before?

ArangoSearch index commit looks like:
1. get rocksdb snapshot
2. make new iresearch snapshot
3. mask in iresearch snapshot documents that not still available in rocksdb snapshot (by tick). These documents will be unmasked in the next iresearch commit # this how bug was solved

So in general you just should call remove/insert and everything will be work.
I think the main difference with rocksdb that we more rely on that new insert generate new localdocumentid.

I made request change because this change doesn't solve a some bug with replication2 with search indexes and instead tried to hide it.
Also in replication1 not for all truncate range delete used, sometime it's common removes. And we tested search (and fix some bug before) with it too.

mpoeter:
@MBkkt this "fixes" a flaky test, but you are right that it may hide some actual issue. The problem is that sometimes a query with `COLLECT WITH COUNT` may return a non-zero count while a query that returns the actual documents returns an empty result.
AFAIK this can be because of an "optimization" that does not really look at the documents, but may return incorrect results with tombstones. That's what we try to work around here.

MBkkt:
@mpoeter 

As I mentioned this change force to read from rocksdb snapshot, for an example if we will make search late materialize passthrough (which is possible) this workaround won't work.

I doubt it's some "count" issue but you can check it, with using view without materialization instead of "count" (read results from stored values).

So I don't think this change desired, whole search snapshot can be incorrect in this test because assumptions in replication2 doesn't work for search and this should be fixed.

mpoeter:
@MBkkt I changed the query to return just the `_key` - does this not count as a stored value? It is definitely a count issue, because as I said, the two queries produces different results. I don't know enough about the search internals to make a qualified statement about _why_ we see this difference, but my tests irrefutably showed that we have such a difference.

FYI - during replication2 recovery we replay part of the log which can result in re-inserts of documents (using the same _rev/LocalDocumentId). I cannot tell if this is a problem for iresearch or not.

MBkkt:
>  I changed the query to return just the _key - does this not count as a stored value?

No it's not, please read documentation about "storedValues" (they should be explicitly specified in view definition), for now this value read from rocksdb snapshot.

> It is definitely a count issue, because as I said, the two queries produces different results

No, it's not definitely, because as I mentioned it's possible that whole search snapshot is incorrect

> why we see this difference, but my tests irrefutably showed that we have such a difference.

We don't have it for replication1, so we should to fix for replication2.
This change just try to hide the bug in my opinion.

> FYI - during replication2 recovery we replay part of the log which can result in re-inserts of documents (using the same _rev/LocalDocumentId). I cannot tell if this is a problem for iresearch or not.

As already was discussed yes it's not supported by default, exists few ways to support this.

PR #20466, opened 2024-01-11
fgrep: warning: fgrep is obsolescent; use grep -F

Rustmilian:
replaces fgrep with grep -F


dothebart:
Thanks for the sugestion. @KVS85 do we still package the generated file? or could we delete this script alltogether?

PR #20467, opened 2024-01-12
Get rid of initgroups avoid nsswitch, take 3

neunhoef:
It turns out that the date library we use (3rdParty) performs a user
database lookup for no good reason. Namely, when we set the folder
for tzdata, then a static string variable is first automatically
initialized to ~/Downloads, which is then immediately overwritten
by the path we set. Unfortunately, the ~ in the path is looked up
via the getpwuid glibc-internally. This poses problems on static
glibc builds when /etc/nsswitch.conf is configured in a certain way
and a different glibc version is available on the host system than
the one with which we build.

This PR fixes this by replacing `~/Downloads` with
`/home/arangodb/Downloads` which has no further consequences.

### Scope & Purpose

- [*] :hankey: Bugfix

### DIFF
--- a/3rdParty/README_maintainers.md
+++ b/3rdParty/README_maintainers.md
@@ -35,7 +35,9 @@ Custom boost locator
 
 ## date
 
-Forward port of C++20 date/time class
+Forward port of C++20 date/time class. We have patched this slightly to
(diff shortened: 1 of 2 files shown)


jsteemann (approved):
LGTM

PR #20470, opened 2024-01-12
[Bug-Fix] Attempt to fix windows frontend build

hkernbach:
### Scope & Purpose

For some reason, windows does not like two commands called in a cmake `add_custom_target`.
Before:
```
COMMAND yarn install
COMMAND yarn build
```

Now: 
```
COMMAND cmd /c "yarn install && yarn build"
```

This approach has worked locally on my windows dev machine. Now need to verify this on both CIs. 

- [x] :hankey: Bugfix

### DIFF
--- a/cmake/frontend/aardvark.cmake
+++ b/cmake/frontend/aardvark.cmake
@@ -10,8 +10,7 @@ set(FRONTEND_DESTINATION ${PROJECT_SOURCE_DIR}/js/apps/system/_admin/aardvark/AP
 add_custom_target(frontend ALL
   COMMENT "create frontend build"
   WORKING_DIRECTORY ${FRONTEND_DESTINATION}
-  COMMAND yarn install
-  COMMAND yarn build
+  COMMAND cmd /c "yarn install && yarn build"
 )
 
 add_custom_target(frontend_clean


hkernbach:
Jenkins Verification Run: https://jenkins.arangodb.biz/job/arangodb-ANY-windows-matrix/5397/

mpoeter:
Updated the fix with a different approach.

On Windows cmake creates a cmd script for the custom command. However, yarn itself is also a cmd script. We have to use the CALL command to run a nested batch script, because otherwise the nested batch script will terminate the _parent_ script.

Jenkins run: https://jenkins.arangodb.biz/job/arangodb-ANY-windows-matrix/5398/

hkernbach:
> Updated the fix with a different approach.
> 
> On Windows cmake creates a cmd script for the custom command. However, yarn itself is also a cmd script. We have to use the CALL command to run a nested batch script, because otherwise the nested batch script will terminate the _parent_ script.
> 
> Jenkins run: https://jenkins.arangodb.biz/job/arangodb-ANY-windows-matrix/5398/

LGMT! :)

dothebart (approved):

cmake/frontend/aardvark.cmake
```diff
@@ -2,19 +2,28 @@
...
+else ()
+  set(YARN_PREFIX "")
+endif ()
+
 add_custom_target(frontend ALL
   COMMENT "create frontend build"
   WORKING_DIRECTORY ${FRONTEND_DESTINATION}
-  COMMAND yarn install
-  COMMAND yarn build
+  COMMAND ${YARN_PREFIX} yarn install
```
neunhoef, line 22: The PR description says that one must not use two COMMAND arguments, but the PR does not fix this. I am confused...

neunhoef: Maybe you can adjust the PR comment?

neunhoef (approved):
LGTM

cmake/frontend/aardvark.cmake
hkernbach: I will. The original statement is not true anymore since mpoeters last commit :D

PR #20471, opened 2024-01-12
Validate that smartgraph attribute exists and is not changed

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/20354

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1433

Previously this check was performed in the KeyGenerator and therefore only performed for inserts, but not for replace or update operations. Therefore replace or update operations could modify or even remove this attribute.
Now we use a separate validator to ensure that this cannot happen. In addition, this PR also changes the returned error code from `ERROR_KEY_MUST_BE_PREFIXED_WITH_SMART_GRAPH_ATTRIBUTE` to `ERROR_NO_SMART_GRAPH_ATTRIBUTE` if the attribute is not set.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/20472

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1433
- [ ] Design document:


neunhoef (approved):
LGTM

mpoeter (approved):

CHANGELOG
```diff
@@ -1,6 +1,14 @@
 v3.11.7 (XXXX-XX-XX)
 --------------------
 
+* Validate that smartgraph attribute exists and is not changed. Previously this
+  check was performed in the KeyGenerator and therefore only performed for
+  inserts, but not for replace or update operations. Therefore replace or update
+  operations could modify or even remove this attribute.
+  In addition, we now return the error code `ERROR_NO_SMART_GRAPH_ATTRIBUTE`
+  instead of `ERROR_KEY_MUST_BE_PREFIXED_WITH_SMART_GRAPH_ATTRIBUTE` if the
+  attribute is not set.
```
```suggestion
* Validate that the attribute stored in the `smartGraphAttribute` of SmartGraph
  vertex collections exists and is not changed afterwards by update or replace
  operations. Previously the `smartGraphAttribute` value was checked only when
  inserting documents into a SmartGraph vertex collection, but not for update
  or replace operations, although the documentation always stated that the 
  `smartGraphAttribute` value must not be changed after the initial creation of
  the document.
  The missing checks on update/replace allowed users to retroactively modify
  the value of the `smartGraphAttribute` for existing documents, which could 
  have led to problems when the data of such a SmartGraph vertex collection was
  replicated to a new follower shard. On the new follower shard, the documents
  went through the full validation, and led to documents with modified
  `smartGraphAttribute` values being rejected on the follower. This could have
  led to follower shards not getting in sync.
  Now the value of the `smartGraphAttribute` will be fully validated with every
  insert, update or replace operation, and every attempt to modify the value of
  the `smartGraphAttribute` retroactively will fail with error 4003
  (`ERROR_KEY_MUST_BE_PREFIXED_WITH_SMART_GRAPH_ATTRIBUTE`, error message
  "in smart vertex collections _key must be a string and prefixed with the 
  value of the smart graph attribute").
  Additionally, if upon insert the `smartGraphAttribute` is missing for a
  SmartGraph vertex, the error code will be error 4001 (
  `ERROR_NO_SMART_GRAPH_ATTRIBUTE`, error message "smart graph attribute not 
  given") instead of error 4003.
  To retroactively repair the data in any of the affected collections, it is
  possible to update every (affected) document with the correct value of the
  `smartGraphAttribute` via an AQL query as follows:
  
      FOR doc IN @@collection 
        LET expected = SUBSTRING(doc._key, 0, FIND_FIRST(doc._key, ':')) 
        LET actual = doc.@attr 
        FILTER expected != actual 
        UPDATE doc WITH {@attr: expected} IN @@collection 
        COLLECT WITH COUNT INTO updated 
        RETURN updated

  This will update all documents with the correct (expected) value of the
  `smartGraphAttribute` if it deviates from the expected value. The query will
  return the number of updated documents as well.
  The bind parameters necessary to run this query are:
  - `@@collection`: name of a SmartGraph vertex collection to be updated
  - `@attr`: attribute name of the `smartGraphAttribute` of the collection
```

jsteemann:
**Important**: before merging, please apply the suggestion for the CHANGELOG. Thanks!

PR #20478, opened 2024-01-15
[Bug-Fix] Do not copy UI's node_modules folder in V8

hkernbach:
### Scope & Purpose

This PR fixes the unnecessary copy of the `node_modules` folder located in the frontend area.
This issue has been (accidentally) found by our Windows test runs.
It got uncovered as Windows run into issues with the maximum path length:

```
Microsoft Windows has a MAX_PATH limit of ~256 characters.
```

This will now also speed up the server start as the V8Feature will NOT copy those files anymore
(As usually any node_modules directory consists of a lot of files and a lot of sub-directories).

- [x] :hankey: Bugfix

### Checklist

- [x] :book: CHANGELOG entry made

PR #20480, opened 2024-01-16
[Bug-Fix] Do not copy UI's node_modules folder in ArangoShell (V8)

hkernbach:
### Scope & Purpose

Fix `node_modules` copy in arangosh. 
Quite close to the previous PR: https://github.com/arangodb/arangodb/pull/20478
But now "client" part. 

- [x] :hankey: Bugfix

### Checklist


client-tools/Shell/V8ShellFeature.cpp
```diff
@@ -310,9 +310,20 @@ void V8ShellFeature::copyInstallationFiles() {
...
+      FileUtils::buildFilename("js", "apps", "system", "_admin", "aardvark",
+                               "APP", "react", "node_modules");
   std::regex const binRegex("[/\\\\]\\.bin[/\\\\]", std::regex::ECMAScript);
 
-  auto filter = [&nodeModulesPath, &nodeModulesPathVersioned,
+  auto filterPath = [](std::string_view normalizedPath,
+                       std::string_view filterPath) -> bool {
+    return (!filterPath.empty() && normalizedPath.size() >= filterPath.size() &&
+            normalizedPath.substr(normalizedPath.size() - filterPath.size(),
+                                  filterPath.size()) == filterPath);
```
jsteemann, line 322: Isn't this the same as `return !filterPath.empty() && normalizedPath.ends_with(filterPath);`?

hkernbach:
Windows run can be found here:
- https://jenkins.arangodb.biz/job/arangodb-matrix-pr-windows/24123/

For some reason it did not get linked automatically here from jenkins.

jsteemann (changes requested):
While this fix works, I would be in favor of not even trying to copy all these files when starting the arangosh.
What we actually only need to copy when starting arangosh is `js/client` and `js/common`. Maybe also `js/node`, but I don't know.
Neither `js/server` nor `js/apps` nor `js/actions` are needed to boot up the arangosh, so we should not even attempt to copy files in these directories.

hkernbach:
Jenkins Windows Run: https://jenkins.arangodb.biz/job/arangodb-matrix-pr-windows/24124/

jsteemann (approved):
LGTM

hkernbach:
Another Jenkins Windows Run: https://jenkins.arangodb.biz/job/arangodb-matrix-pr-windows/24136/

PR #20492, opened 2024-01-17
Improve Batch Materialize

maierlars:
### Scope & Purpose

- Allow IndexNodes that use the same index for multiple conditions.

PR #20493, opened 2024-01-17
Batch Materialize with Projections

maierlars:
### Scope & Purpose
Allow projections during late materialization.
Depends on https://github.com/arangodb/arangodb/pull/20492.

```
Query String (109 chars, cacheable: true):
         FOR doc IN MyTestCollection
           FILTER doc.x > 5
           RETURN [doc.y, doc.z, doc.a]
       

Execution plan:
  7   IndexNode           ✓   1000     - FOR doc IN MyTestCollection   /* persistent index scan, index scan + document lookup */    /* with late materialization */
  8   MaterializeNode         1000       - MATERIALIZE doc /* (projections: `a`, `y`, `z`) */
  5   CalculationNode     ✓   1000       - LET #2 = [ doc.`y`, doc.`z`, doc.`a` ]   /* simple expression */   /* collections used: doc : MyTestCollection */
  6   ReturnNode              1000       - RETURN #2

Indexes used:
 By   Name                      Type         Collection         Unique   Sparse   Cache   Selectivity   Fields    Stored values   Ranges
  7   idx_1788354556957032448   persistent   MyTestCollection   false    false    false      100.00 %   [ `x` ]   [ `b` ]         (doc.`x` > 5)

Optimization rules applied:
 Id   Rule Name                                 Id   Rule Name                                 Id   Rule Name                        
  1   move-calculations-up                       5   use-indexes                                9   batch-materialize-documents      
  2   move-filters-up                            6   remove-filter-covered-by-index            10   async-prefetch                   
  3   move-calculations-up-2                     7   remove-unnecessary-calculations-2
  4   move-filters-up-2                          8   reduce-extraction-to-projection  

Optimization rules with highest execution times:
 RuleName                                    Duration [s]
 use-indexes                                      0.00021

53 rule(s) executed, 1 plan(s) created, peak mem [b]: 0, exec time [s]: 0.00213
```

If an index scan is used but the documents have to be loaded for the projections, this PR now makes sure that the documents are loaded in batches.

This PR only supports the old style of projections, where the document is built up and stored in a single variable. A follow up PR will improve this situation.


tests/js/client/aql/aql-index-batch-materialize.js
```diff
@@ -182,6 +192,47 @@ function IndexBatchMaterializeTestSuite() {
...
     },
 
+    testMaterializeIndexScanProjections: function () {
+      const query = `
+        FOR doc IN ${collection}
+          FILTER doc.x > 5
+          RETURN [doc.y, doc.z, doc.a]
+      `;
+      const {materializeNode, indexNode} = expectOptimization(query);
+      assertTrue(indexNode.projections, undefined);
```
jsteemann, line 202: I can't make sense of this assertion.
Should it rather be
```suggestion
      assertUndefined(indexNode.projections);
```

jsteemann (approved):
LGTM

PR #20496, opened 2024-01-18
Suppress all data races coming from within V8 and improve build settings

mpoeter:
### Scope & Purpose

ATM we build V8 with sanitizers _disabled_, so the V8 code is not instrumented which means the sanitizer cannot observe the synchronizations inside that code. So when the V8 code calls back into instrumented code (e.g. the C runtime), TSan might report potential data races because of these missing synchronizations. For that reason we simply need to ignore _all_ data races coming from V8.

In addition this PR changes the build settings in V8 to only generate minimal debug information. For V8 we don't care about local variables etc.; function names are enough so that we can generate back traces. Also, this PR enables compression of debug sections for all executables to significantly reduce their sizes.


jsteemann (approved):
LGTM

PR #20509, opened 2024-01-18
Push Down Late Materialization

maierlars:
### Scope & Purpose

This PR adds an optimizer rule that pushes down `Materialize` nodes to defer the materialization and reduce the amount of materialize documents.


jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #20516, opened 2024-01-19
[CINFRA] Fail to create databases if automatic syncing is disabled

apetenchea:
### Scope & Purpose

For replication2 databases, automatic syncing must be enabled, otherwise we cannot use the `waitForSync=false` setting. In case automatic syncing is disabled, we previously used to emit a warning stating that replication2 can no longer be used. Now, on top of that, any attempts to create a replication2 database will fail.
Note that once we introduce the custom WAL, this will be no longer needed.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature


arangod/VocBase/Methods/Databases.cpp
```diff
@@ -402,15 +404,24 @@ Result Databases::create(ArangodServer& server, ExecContext const& exec,
...
+        // Should not happen during testing
+        TRI_ASSERT(false);
+        return Result(TRI_ERROR_NOT_IMPLEMENTED, message);
+      }
+
+      auto& selector = server.getFeature<EngineSelectorFeature>();
+      auto& engine = selector.engine<RocksDBEngine>();
+      if (engine.syncThread() == nullptr) {
+        return Result(TRI_ERROR_ILLEGAL_OPTION,
+                      "Unless using the custom WAL, automatic syncing must be "
```
mchacki, line 422: Please not name the Custom WAL in this publicly visible message.
This is nothing our customers can control.

I would prefer a message that helps the user, so it should in my opinion contain info on the setting to avoid this situation
(taken from above:)
"--rocksdb.sync-interval"

mchacki (changes requested):
Technical good.
I would like to get a more user helping error message.

goedderz (commented):
Is it still possible to create a replication 2 database, and then turn off the sync thread? If so, I think we should error out during startup.

apetenchea:
> Is it still possible to create a replication 2 database, and then turn off the sync thread? If so, I think we should error out during startup.

I have added a second check in `RocksDBEngine::openExistingDatabase()` and manually checked that it will not let us start a cluster without a syncer thread, once there are replication2 databases on disk.

mchacki (approved):
LGTM.

If it is possible to formulate I would like an automatic test for the startup with existing Replication2  Database.

Technical this looks good to me :+1:

apetenchea:
Added test cases in `test-restart-replication2-syncer-thread-off.js`

mchacki (approved):
LGTM! Good test

PR #20517, opened 2024-01-19
Materialize Projections into Registers

maierlars:
### Scope & Purpose

This PR adds the support for projections into registers for the materialize node.


arangod/Aql/OptimizerRuleBatchMaterializeDocuments.cpp
```diff
@@ -120,6 +122,12 @@ void arangodb::aql::batchMaterializeDocumentsRule(
         plan.get(), plan->nextId(), indexNode->collection(), *docIdVar,
         *indexNode->outVariable());
     plan->insertAfter(indexNode, materialized);
+    if (!indexNode->projections().empty()) {
+      TRI_ASSERT(!indexNode->projections().usesCoveringIndex());
+      TRI_ASSERT(!indexNode->projections().hasOutputRegisters());
+      // move projections from index node into materialize node
+      materialized->projections(std::move(indexNode->projections()));
```
jsteemann, line 129: nit:
Is it intended here to remove the projections from the IndexNode?
I think the current code transfer them, so that the IndexNode is left without projections. The code comment above also indicates that this is the intended behavior.
I am fine with this, but I would like to make this even more obvious, probably by either some explicit call to `indexNode->projections().clear()` (if such method exists), or even a new method `Projections DocumentProducingNode::stealProjections()`. 
That would make it absolutely obviously that projections are removed from the IndexNode. Otherwise, when casually reading the code, it was not 100% clear to me.

jsteemann (approved):
LGTM

maierlars:
Closed in favor of https://github.com/arangodb/arangodb/pull/20650

PR #20526, opened 2024-01-23
 [Bug-Fix] Fix comparison between two GeoIndices

hkernbach:
### Scope & Purpose

Compare method delivered wrong results while comparing two geo indices. 

- [x] :hankey: Bugfix


jsteemann (approved):
LGTM

PR #20530, opened 2024-01-24
improvements for TTL index removal queries

jsteemann:
### Scope & Purpose

Improvements for TTL index removal queries:

- do not have max query runtime affect TTL index removal queries
- do not favor certain databases/collections/shards in the removals, but give them more or less equal probability of being picked
- reduce the default amount of documents to be removed per collection (`--ttl.max-collection-removes` startup option`) from 1m to 40k. this allows different databases/collections/shards to be picked in a single TTL background thread iteration
- do not use an AQL query for removing documents from the TTL index. using an AQL query initiated on the DB server is not ideal because it will intermediate commit on followers but can be rolled back on leaders

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RestServer/TtlFeature.cpp
+++ b/arangod/RestServer/TtlFeature.cpp
@@ -337,6 +337,10 @@ class TtlThread final : public ServerThread<ArangodServer> {
           if (index->type() != Index::TRI_IDX_TYPE_TTL_INDEX) {
             continue;
           }
+          if (index->isHidden()) {
+            // Filter out indexes that are currently still being built
+            continue;
+          }
 
           // serialize the index description so we can read the "expireAfter"
           // attribute
--- a/arangod/RestServer/TtlFeature.cpp
+++ b/arangod/RestServer/TtlFeature.cpp
@@ -43,6 +43,10 @@
 #include "ProgramOptions/ProgramOptions.h"
 #include "RestServer/DatabaseFeature.h"
 #include "Transaction/StandaloneContext.h"
+#include "Transaction/Methods.h"
+#include "Utils/OperationOptions.h"
+#include "Utils/OperationResult.h"
+#include "Utils/SingleCollectionTransaction.h"
 #include "VocBase/LogicalCollection.h"
 #include "VocBase/vocbase.h"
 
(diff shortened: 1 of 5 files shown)


arangod/RestServer/TtlFeature.cpp
```diff
@@ -302,6 +320,9 @@ class TtlThread final : public ServerThread<ArangodServer> {
             continue;
           }
 
+          // TODO: Filter out indexes that are currently still being built
+          // (checking isBuilding flag)
```
neunhoef, line 324:
```suggestion
          if (index->isHidden()) {
            // Filter out indexes that are currently still being built
            continue;
          }
```

arangod/RestServer/TtlFeature.cpp
```diff
@@ -320,66 +341,160 @@ class TtlThread final : public ServerThread<ArangodServer> {
...
-                        << " documents for collection '" << collection->name()
-                        << "'";
-                    if (limitLeft >= removed) {
-                      limitLeft -= removed;
-                    } else {
-                      limitLeft = 0;
-                    }
-                  }
+
+            auto trx = std::make_unique<SingleCollectionTransaction>(
```
neunhoef, line 422: For a shard belonging to a smart edge collection, we must proceed as follows:
 1. If it is a `_local` shard: to what we do here normally (local removal with replication).
 2. If it is a `_from` shard: send removal as batch to the corresponding global edge collection on some coordinator.
 3. If it is a `_to` shard: skip it.

neunhoef (changes requested):
This is great! I think it will solve our problems. But two fixes are still missing, see comments. We must skip indexes, which are still isBuilding, this is easy. Secondly, we must deal with smart edge collections correctly.

neunhoef (approved):
LGTM

arangod/RestServer/TtlFeature.cpp
```diff
@@ -320,65 +356,252 @@ class TtlThread final : public ServerThread<ArangodServer> {
...
+              }
+
+              size_t count = 0;
+              for (auto it : VPackArrayIterator(val.slice())) {
+                VPackSlice error = it.get(StaticStrings::Error);
+                if (!error.isTrue()) {
+                  ++count;
+                  continue;
+                }
+              }
```
mpoeter, line 517:
```suggestion
              for (auto it : VPackArrayIterator(val.slice())) {
                VPackSlice error = it.get(StaticStrings::Error);
                if (!error.isTrue()) {
                  ++count;
                }
              }
```

mpoeter (approved):
I cannot even begin to express how much I dislike that this adds a lot more complexity (including duplicated code and mixing different abstraction levels) to a function that is already rather long. 

But I guess it works and has to be "quick" again.

arangod/RestServer/TtlFeature.cpp
```diff
@@ -320,65 +356,252 @@ class TtlThread final : public ServerThread<ArangodServer> {
...
+                         "collection '"
+                      << collection->name() << "': " << res.errorMessage();
+                }
+                break;
+              }
+
+              OperationOptions opOptions;
+              opOptions.ignoreRevs = false;
+              opOptions.waitForSync = false;
+              // opOptions.silent = true;
```
mpoeter, line 552: Remove?
```suggestion
```

neunhoef (approved):
LGTM

PR #20533, opened 2024-01-24
Fix data race in actor framework

mpoeter:
### Scope & Purpose

Concurrent finish requests could race on setting the exitReason.

- [x] :hankey: Bugfix


neunhoef (approved):
LGTM

PR #20541, opened 2024-01-25
Feature/index visibility

kvahed:
### Scope & Purpose

*Forward port from 3.10 of index visibility*
Enterprise companion https://github.com/arangodb/enterprise/pull/1442

- [ ] :hankey: Bugfix
- [X] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [X] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/ActionBase.cpp
+++ b/arangod/Cluster/ActionBase.cpp
@@ -275,7 +275,10 @@ void ActionBase::result(ErrorCode errorNumber, std::string const& errorString) {
   _result.reset(errorNumber, errorString);
 }
 
-arangodb::Result ActionBase::setProgress(double d) { return {}; }
+arangodb::Result ActionBase::setProgress(double d) {
+  _progress.store(d, std::memory_order_relaxed);
+  return {};
+}
 
 /**
  * progress() operation is an expected future feature.  Not supported in the
--- a/arangod/Cluster/ActionBase.cpp
+++ b/arangod/Cluster/ActionBase.cpp
@@ -170,11 +170,8 @@ void ActionBase::startStats() {
 
 /// @brief show progress on Action, and when that progress occurred
 void ActionBase::incStats() {
-  auto progress = _progress.load();
-  progress += 1.;
-  _progress.store(progress);
+  _progress.fetch_add(1.);
   _actionLastStat = secs_since_epoch();
-
 }  // ActionBase::incStats
 
 void ActionBase::endStats() {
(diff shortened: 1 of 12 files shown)


arangod/Cluster/ActionBase.cpp
```diff
@@ -170,11 +170,8 @@ void ActionBase::startStats() {
 
 /// @brief show progress on Action, and when that progress occurred
 void ActionBase::incStats() {
-  auto progress = _progress.load();
-  progress += 1.;
-  _progress.store(progress);
+  _progress.fetch_add(1.);
```
jsteemann, line 173: We are using atomic increment here, whereas in 3.10 & 3.11 we are increasing the value under a mutex.
This is ok for me, because the reason to use a mutex in 3.10 & 3.11 was that the atomic increment did not compile on macOS. This should not be a problem for our devel branch, but wanted to confirm here that this intentionally deviates from 3.10 & 3.11.

jsteemann (approved):
LGTM

arangod/Cluster/ActionBase.cpp
neunhoef: This was exactly my understanding, that we no longer need to support the Mac compiler in 3.12.

arangod/ClusterEngine/ClusterIndex.cpp
```diff
@@ -169,66 +169,6 @@ void ClusterIndex::toVelocyPack(
     builder.close();
   }
```
neunhoef, line 170: We probably can remove a few header includes here, too.

neunhoef: AgencyCache.h, LogMacros.h, Utilites.h, Network/Methods.h, Network/NetworkFeature.h I would expect from what we have seen in 3.11.

neunhoef (changes requested):
LGTM except for the few header includes which should again be removed, see comment.

arangod/Cluster/ActionBase.cpp
KVS85: We support macOS compilation for `client-tools` target only.

neunhoef (approved):
LGTM

PR #20547, opened 2024-01-25
potentially fix UB

jsteemann:
### Scope & Purpose

Fix potential UB in non-released code

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #20549, opened 2024-01-25
potentially fix UB

jsteemann:
### Scope & Purpose

Fix potential UB in non-released code

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20548
  - [x] Backport for 3.10: https://github.com/arangodb/arangodb/pull/20547

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/ClusterQuery.cpp
+++ b/arangod/Aql/ClusterQuery.cpp
@@ -97,7 +97,7 @@ std::shared_ptr<ClusterQuery> ClusterQuery::create(
 
 void ClusterQuery::prepareClusterQuery(
     VPackSlice querySlice, VPackSlice collections, VPackSlice variables,
-    VPackSlice snippets, VPackSlice traverserSlice, std::string_view user,
+    VPackSlice snippets, VPackSlice traverserSlice, std::string const& user,
     VPackBuilder& answerBuilder,
     QueryAnalyzerRevisions const& analyzersRevision, bool fastPathLocking) {
   LOG_TOPIC("9636f", DEBUG, Logger::QUERIES)
--- a/arangod/Aql/ClusterQuery.h
+++ b/arangod/Aql/ClusterQuery.h
@@ -53,8 +53,8 @@ class ClusterQuery : public Query {
                            velocypack::Slice collections,
                            velocypack::Slice variables,
                            velocypack::Slice snippets,
-                           velocypack::Slice traversals, std::string_view user,
-                           velocypack::Builder& answer,
+                           velocypack::Slice traversals,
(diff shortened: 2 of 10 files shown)


kvahed (approved):
LGTM

arangod/Aql/ClusterQuery.cpp
```diff
@@ -97,7 +97,7 @@ std::shared_ptr<ClusterQuery> ClusterQuery::create(
 
 void ClusterQuery::prepareClusterQuery(
     VPackSlice querySlice, VPackSlice collections, VPackSlice variables,
-    VPackSlice snippets, VPackSlice traverserSlice, std::string_view user,
+    VPackSlice snippets, VPackSlice traverserSlice, std::string const& user,
```
neunhoef, line 100: I do not understand why this helps. If the UB is that the `user` `string_view` pointed to memory which was later no longer valid, then using a `std::string const&` does not help. I looked at the call sites of this function and found one place, where `ExecContext::current().user()` was used, which returns a **reference** to a string in the `ExecContext`. Now this is thread local and in the new era of coroutines, it is possible that the current `ExecContext` is different, when the coroutine is resumed in a different thread. I fear we must make this argument a `std::string` by value and then move it to its destination to fix the UB.

neunhoef (changes requested):
See comment, I do not understand how this change helps.

arangod/Aql/ClusterQuery.cpp
jsteemann: This may be a possible issue in devel. But in 3.10 we don't use coroutines at all.

MBkkt: >  I fear we must make this argument a std::string by value and then move it to its destination to fix the UB.

It also can be better because for now code copying string under mutex lock

neunhoef (approved):
LGTM

arangod/Aql/ClusterQuery.cpp
mpoeter: I wouldn't care about the copying - this is the username which in virtually all cases will be short enough to fit in the small buffer, so it doesn't matter whether we move or copy, we have to do the same work.

mpoeter (approved):

PR #20552, opened 2024-01-26
[Bug-Fix][BTS-1747] Windows CI do not produce cores (case: hardKills)

hkernbach:
### Scope & Purpose

This PR fixes the behaviour where we want to crash the server on purpose and do not want core files creation. 
This already works on macOS and Linux, but needs this subtle change to also work on Windows operating systems. 

- [x] :hankey: Bugfix

#### Related Information

- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1747


jsteemann (approved):
LGTM

hkernbach:
NOTE: arangodb-matrix-pr failed overall - **but the expected single-server fix worked.**

PR #20565, opened 2024-01-29
Fix more leftover windows ci issues

hkernbach:
### Scope & Purpose

Some tests do fail, especially under windows, due to the test setup. 
This branch will fix some (or all) of them. 

- [x] :hankey: Bugfix


apetenchea (approved):
LGTM

PR #20573, opened 2024-01-30
Fix issue that prefetch tasks could flood the scheduler queue

mpoeter:
### Scope & Purpose

Async prefetching could post a task onto the scheduler. If this task is not picked up in time, we could claim that task for ourselves and simply to the work directly. However, we could post yet another task for the next batch. This could go on and on, filling up the scheduler queue. As a side effect, there was also a race where an old task could be executed with an old AqlCallStack. To avoid this, we now rearm the already posted task with the current AqlCallStack and set it to pending. This way we can ensure that either the updated task is available to be picked up, or if it was picked up before we could make it available again, we will realize and can then repost it on the scheduler.

- [x] :hankey: Bugfix

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**

### DIFF
--- a/arangod/Aql/ExecutionBlockImpl.h
+++ b/arangod/Aql/ExecutionBlockImpl.h
@@ -389,6 +389,8 @@ class ExecutionBlockImpl final : public ExecutionBlock {
       bool abandoned;
     };
     std::atomic<State> _state{{Status::Pending, false}};
+    static_assert(decltype(_state)::is_always_lock_free == true);
+
     std::mutex mutable _lock;
     std::condition_variable mutable _bell;
     std::optional<PrefetchResult> _result;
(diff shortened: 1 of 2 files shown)


arangod/Aql/ExecutionBlockImpl.h
```diff
@@ -360,31 +361,43 @@ class ExecutionBlockImpl final : public ExecutionBlock {
...
+    void execute();
 
    private:
-    std::atomic<State> _state{State::Pending};
+    void updateStatus(Status status, std::memory_order memoryOrder) noexcept;
+    struct State {
+      Status status;
+      bool abandoned;
+    };
+    std::atomic<State> _state{{Status::Pending, false}};
```
jsteemann, line 391: I verified that with all relevant compilers (g++-11, g++-13, clang++-16) this atomic is always lock-free (i.e. `std::atomic<State>::is_always_lock_free == true`). Shall we also add an assertion to have this guaranteed during the compile process?

mpoeter: I figured it should be lock-free since this is a POD type of (probably) size 2, but yes, let's add an assertion for that!

PR #20574, opened 2024-01-30
Dependency cleanup (reduce getFeature calls)

mpoeter:
### Scope & Purpose

At the moment the ApplicationServer basically acts as a [service locator](https://en.wikipedia.org/wiki/Service_locator_pattern) which is usually seen as an [antipattern](https://deviq.com/antipatterns/service-locator). We currently have ~1940 calls to getFeature to get some feature from the server. This approach hides the dependencies and has a negative impact on maintainability, testability, and the clarity of the code. Instead, we should use proper dependency injection, which usually means to take all the necessary dependencies as parameters (e.g. in the constructor).

This PR is a first step to reduce the number of getFeature calls to make dependencies more explicit.

#### Related Information

- [x] Jira ticket: https://arangodb.atlassian.net/jira/software/c/projects/TD/issues/TD-1


markuspf:
This change looks good to me. It's a long overdue step to make the ArangoDB codebase better structured.

I wonder about the risk that comes with a change this size; is there anything that could subtly break and it is not noticed in tests? I can't think of anything (yet).

markuspf (approved):
See my other comment. Long overdue change.

PR #20577, opened 2024-01-31
RocksDBEngine is only available on dbservers

mpoeter:
### Scope & Purpose

A new check that was added recently could result in undefined behavior on the coordinator, because we tried to access the RocksDBEngine which is not available on Coordinators.
To avoid such mistakes in the future this PR also adds an assertion that we do not cast to the wrong engine type.

- [x] :hankey: Bugfix


markuspf (approved):
LGTM

PR #20578, opened 2024-01-31
reduce cluster-internal traffic

jsteemann:
### Scope & Purpose

Save some traffic in cluster-internal requests:

* Remove disk usage metrics from heartbeat messages (`freeDiskSpaceBytes` and `freeDiskSpacePercent`), as these are not processed by the agency supervision at all.
* Do not return `id` and `endpoint` attributes in responses to empty `appendEntries` requests in the agency, as these attributes are not picked up by the response message handler.
* Do not include `x-arango-source` HTTP header in any outgoing requests that the agency makes to other servers.
* Do not include `x-arango-hlc` HTTP header in heartbeat messages to the agency's transient store.

* Do not return the following HTTP response headers in cluster-internal traffic:
  * `x-arango-queue-time-seconds`: this is not picked up by any handler in cluster-internal traffic
  * `x-content-type-options`
  * `content-security-policy`
  * `cache-control`
  * `pragma`
  * `expires`
  * `strict-transport-security`
  * `server`
  
  The following responses do **not** count as cluster internal traffic, so the HTTP response headers are fully retained:
  * responses from single server instances
  * responses from coordinators if the `x-arango-source` HTTP header was not set in the request
  The following responses count as cluster-internal traffic, so the HTTP response headers will be removed:
  * all responses from agents and DB servers, regardless from where the requests were initiated
  * all responses from coordinators, if the `x-arango-source` HTTP header was set in the request and is set to the id of a coordinator or DB server

* Removed unused `resetRAFTTimes` method in agency.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


MBkkt (approved):

neunhoef (approved):
LGTM

PR #20580, opened 2024-02-01
[CINFRA] Handle shutdown gracefully

apetenchea:
### Scope & Purpose

Normally, before the log leader tries to apply a resource update locally (e.g. drop a shard), it first has to replicate the same update on followers. If replication succeeds, but the leader itself is unable to apply the operation locally, we consider the leader unable to continue and kill it.
However, if the operation fails due to the leader being shut down, there's no reason to panic. All we have to do is make sure to keep the log entry around (not calling release anymore), and let the shutdown continue. That is the purpose of this PR.

Additionally, a couple of `std::move` around shard IDs were removes, as they became obsolete since now ShardID is trivially copyable.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM

PR #20581, opened 2024-02-01
[CINFRA-885] Replicate CreateIndex while holding the exclusive collection lock

goedderz:
### Scope & Purpose

https://arangodb.atlassian.net/browse/CINFRA-885


apetenchea (approved):
LGTM

PR #20582, opened 2024-02-01
Fixed check if a Shard is a leader for Replication2 in TTL BG thread

mchacki:
### Scope & Purpose

*The TTL background thread to skip over followers only worked on Replication1. Now we use the shared mechanism that works in both cases.

Also slightly adapted the test to be more resilient against "bad luck" the test can only wait if any DBServer has performed a TTL cleanup, if things are unlucky we have a server that only has Followers, so it does nothing, which would cause the test to fail. Now we wait for at least 4 runs of TTL, which increases the odds that at least one leader hat the chance to perform one.*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RestServer/TtlFeature.cpp
+++ b/arangod/RestServer/TtlFeature.cpp
@@ -332,7 +332,7 @@ class TtlThread final : public ServerThread<ArangodServer> {
         }
 
         if (ServerState::instance()->isDBServer() &&
-            !collection->followers()->getLeader().empty()) {
+            !collection->isLeadingShard()) {
           // we are a follower for this shard. do not remove any data here, but
           // let the leader carry out the removal and replicate it
           continue;
--- a/tests/js/client/shell/multi/shell-ttl-nocov.js
+++ b/tests/js/client/shell/multi/shell-ttl-nocov.js
@@ -1014,15 +1014,15 @@ function TtlSuite () {
             docs = [];
           }
         }
-      
+
         let oldStats = internal.ttlStatistics();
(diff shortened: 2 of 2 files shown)


arangod/RestServer/TtlFeature.cpp
```diff
@@ -332,7 +332,7 @@ class TtlThread final : public ServerThread<ArangodServer> {
         }
 
         if (ServerState::instance()->isDBServer() &&
-            !collection->followers()->getLeader().empty()) {
+            !collection->isLeadingShard()) {
```
apetenchea, line 335: The `isLeadingShard` return `false` if the collection `isAStub()`. I'm pointing it out, although as far as I know that can only happen on a coordinator.

apetenchea (approved):
LGTM

arangod/RestServer/TtlFeature.cpp
mchacki: Good Spot!
That is actually expected and I think the correct behaviour.
isAStub indicates we are talking to the "Cluster-Wide" Collection, that does not have data itself, but only delegates to shards.
This collections cannot do any of the below operations, so we are good, and have another layer of protection.

PR #20584, opened 2024-02-02
nightly fix startup 

maierlars:
When there is high load on the test system, 30s might not be enough for the single server to come up.


jsteemann (approved):
LGTM

PR #20585, opened 2024-02-02
[CINFRA] Adjust replication2 test

apetenchea:
### Scope & Purpose

The test is meant to check that we disallow the creation of replication2 databases in case the syncer thread is off. However, if there are already replication2 databases on disk (which is the case of `_system` when replication2 is on by default), it is not allowed to switch the syncer thread off, and therefore the test does not make sense in that context.
This PR checks the version of the `_system` database and skips the test accordingly.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


mchacki (approved):
LGTM

PR #20589, opened 2024-02-03
improve performance of constrained sort executor

jsteemann:
### Scope & Purpose

Improve performance of constrained sort executor.

As the final step in the constrained sort executor, the min/max heap is sorted to get a totally ordered result back.
The previous implementation used std::sort() for this, which is over the top. as we have already built up a heap, we can use std::sort_heap() to do the sorting. This can be algorithmically more efficient.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/ConstrainedSortExecutor.cpp
+++ b/arangod/Aql/ConstrainedSortExecutor.cpp
@@ -211,7 +211,10 @@ ExecutorState ConstrainedSortExecutor::consumeInput(
     if (_returnNext == 0) {
       // Only once sort the rows again, s.t. the
       // contained list of elements is in the right ordering.
-      std::sort(_rows.begin(), _rows.end(), *_cmpHeap);
+      // we can get away with std::heap_sort() here as _rows
+      // already satisfies the heap property. this is less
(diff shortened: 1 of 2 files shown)


MBkkt:
it's just n pop_heap, so it's slower

jsteemann:
> it's just n pop_heap, so it can be slower

Yes, it can be slower in theory.
**Worst** case performance for `std::sort()` is n * log(n) comparisons, whereas the **worst** case performance for `std::sort_heap()` is 2 * n * log(n) comparisons.
But in practice it seems that `std::sort_heap()` can substantially outperform `std::sort()` for several inputs (i.e. non-worst case performance).
Probably still not worth doing the change, because it may cause a slowdown in some fringe cases.

PR #20602, opened 2024-02-06
[CINFRA] Index Creation Error

apetenchea:
### Scope & Purpose

In a [previous PR](https://github.com/arangodb/arangodb/pull/20581), the replication step of the `CreateIndex` operation has been moved into the `RocksDBCollection::createIndex` method, mainly with the purpose of avoiding deadlocks caused by the ordering of transaction and index operations in the replicated log.

However, this now leaves a time window during which the replicated state is fetched during index creation, but there is nothing to stop the state from becoming unavailable. Dropping the state in this case doesn't cause any harm - it is ok to abort the index creation and move on, it's not going to matter anymore whether the index has been finished or not.

This PR adds a quick fix to that, by marking the `REPLICATED_STATE_NOT_FOUND` error as safe in the context of a `CreateIndex` operation. Ideally, the shared pointed of the `DocumentState` should be passed all the way to the `RocksDBCollection::createIndex` method, but that would require a more significant refactoring.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### DIFF
--- a/arangod/Replication2/StateMachines/Document/DocumentStateErrorHandler.cpp
+++ b/arangod/Replication2/StateMachines/Document/DocumentStateErrorHandler.cpp
@@ -105,6 +105,21 @@ auto DocumentStateErrorHandler::handleOpResult(
         << " failed because a TTL index already exists, ignoring: " << res;
     return TRI_ERROR_NO_ERROR;
   }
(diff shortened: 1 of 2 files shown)


goedderz (approved):

mchacki (approved):
LGTM

PR #20603, opened 2024-02-06
CINFRA: Hot Restore BugFix Part 2

mchacki:
### Scope & Purpose

*When we take a hot backup, then add a collection, and then do a hot restore from before, there was the ReplicatedLog information left behind of the Collection added in between, however no one was responsible for it, this caused havoc when we tried to remove servers from the Cluster. Now we make sure Current/ReplicatedLogs get's a factory reset. Including a Test (see Enterprise Part)
This PR also fixes a race in Supervision code, which would get into Assertion when it sees an Empty Current.
*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [x] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1449
- [ ] Design document:


arangod/Replication2/ReplicatedLog/Supervision.cpp
```diff
@@ -969,7 +972,9 @@ auto checkConfigCommitted(SupervisionContext& ctx, Log const& log) -> void {
 
   auto const& plan = *log.plan;
   auto const& current = *log.current;
-
+  if (!current.supervision.has_value()) {
+    return;
+  }
   ADB_PROD_ASSERT(current.supervision.has_value());
```
apetenchea, line 978: NIT, these assertions are now redundant, because of the check right above.

apetenchea (approved):
LGTM

arangod/Replication2/ReplicatedLog/Supervision.cpp
mchacki: Agreed.
I doubt that this will have measurable performance impact, also there are other places exactly like this.
S would like to keep it for now, may improve in future.

PR #20606, opened 2024-02-07
BTS-1783

maierlars:
### Scope & Purpose

Add more information to an assertion.

Furthermore, this PR removes to place that made invalid or harmful modifications to the merkel tree state. See comments.

### DIFF
--- a/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
+++ b/arangod/RocksDBEngine/RocksDBMetaCollection.cpp
@@ -473,8 +473,9 @@ RocksDBMetaCollection::computeRevisionTree(uint64_t batchId) {
 
 Result RocksDBMetaCollection::takeCareOfRevisionTreePersistence(
     LogicalCollection& coll, RocksDBEngine& engine, rocksdb::WriteBatch& batch,
-    rocksdb::ColumnFamilyHandle* const cf, rocksdb::SequenceNumber maxCommitSeq,
-    bool force, std::string const& context, std::string& scratch,
+    rocksdb::ColumnFamilyHandle* const cf,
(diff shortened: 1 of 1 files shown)


arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -1008,10 +1008,6 @@ futures::Future<Result> RocksDBMetaCollection::rebuildRevisionTree() {
     while (true) {
       std::unique_lock<std::mutex> guard(_revisionTreeLock);
 
-      if (_revisionTreeApplied < beginSeq) {
-        _revisionTreeApplied = beginSeq;
-      }
```
maierlars, line 1013: This code could potentially move `_revisionTreeApplied` past a blocker. Even worse, if the timeout of two minutes is hit, we leave the old merkel tree in an invalid state.

arangod/RocksDBEngine/RocksDBMetaCollection.cpp
```diff
@@ -962,14 +970,6 @@ futures::Future<Result> RocksDBMetaCollection::rebuildRevisionTree() {
           std::chrono::milliseconds(RandomGenerator::interval(uint32_t(2000))));
     }
 
-    {
-      // we may still have some buffered updates, so let's remove them.
-      std::unique_lock<std::mutex> guard(_revisionTreeLock);
-      removeBufferedUpdatesUpTo(blockerSeq - 1);
-
-      _revisionTreeApplied = blockerSeq - 1;
-    }
```
maierlars, line 971: Removing buffered updates will break the existing merkel tree, if the replacement procedure fails later on. For example due to a timeout.

MBkkt (approved):

PR #20620, opened 2024-02-14
[BTS-1774] Fix potential nullptr access in StateHandleManager

mchacki:
### Scope & Purpose

*We use a protected structure in the StateHandleManager where we make sure it cannot be access in parallel.
However there exists a path where the protected element gets a nullptr shortly before it is freed, namely resign(). Between resign and deallocation however there can be another shared_ptr instance that access the protected element. The functions fixed here just dereferenced it without doing nullptr tests.

For two of them it is save because they share the same outside lock, so I added an ASSERT that we do not have a nullptr.
For one (the case we triggered) I added a throw.
For the last one (acquireSnapshot) the method is actually a noexcept and does not return anything. So I made it only do action if it is still possible.
*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1774
- [ ] Design document:

### DIFF
--- a/arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
+++ b/arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
@@ -93,7 +93,11 @@ void StateHandleManager::acquireSnapshot(const ParticipantId& leader,
 auto StateHandleManager::getInternalStatus() const -> replicated_state::Status {
   auto guard = guardedData.getLockedGuard();
   if (guard->stateHandle == nullptr) {
-    THROW_ARANGO_EXCEPTION(TRI_ERROR_REPLICATION_REPLICATED_LOG_FOLLOWER_RESIGNED);
+    // We can only have a nullptr here if we have resigned,
+    // also this stateHandleManager is only used in the follower
+    // So we can directly return FollowerResigned here.
+    return {replicated_state::Status::Follower{
+        replicated_state::Status::Follower::Resigned{}}};
   }
   return guard->stateHandle->getInternalStatus();
 }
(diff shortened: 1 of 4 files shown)


goedderz (approved):

arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
```diff
@@ -69,16 +74,32 @@ StateHandleManager::StateHandleManager(
...
   guard->stateHandle->becomeFollower(std::move(ptr));
 }
 
 void StateHandleManager::acquireSnapshot(const ParticipantId& leader,
                                          std::uint64_t version) noexcept {
-  auto guard = guardedData.getLockedGuard();
-  // TODO is the correct log index required here?
-  guard->stateHandle->acquireSnapshot(leader, LogIndex{0}, version);
+  if (auto guard = guardedData.getLockedGuard(); guard->stateHandle) {
+    // TODO is the correct log index required here?
```
apetenchea, line 90: This is fine for now, the `LogIndex` is not even used.

apetenchea (approved):

arangod/Replication2/ReplicatedLog/Components/StateHandleManager.cpp
mchacki: I moved the comment only ;)

PR #20625, opened 2024-02-15
[BTS-1789] Fix local expansion for join nodes

maierlars:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1450

Local expansion for joins for shards on the same dbserver was not enabled. This could lead to wrong results, because only one shard was read from each server.

### DIFF
--- a/tests/js/client/aql/aql-index-join.js
+++ b/tests/js/client/aql/aql-index-join.js
@@ -1060,6 +1060,22 @@ const IndexJoinTestSuite = function () {
       }
     },
 
+    testJoinInClusterShardedLocalExpansions: function () {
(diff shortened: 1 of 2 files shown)


tests/js/client/aql/aql-index-join.js
```diff
@@ -1060,6 +1060,22 @@ const IndexJoinTestSuite = function () {
       }
     },
 
+    testJoinInClusterShardedLocalExpansions: function () {
+      const A = db._create("A", {numberOfShards: 10, replicationFactor: 1, writeConcern: 1, waitForSync: true});
```
jsteemann, line 1064: Do we need the `waitForSync: true` here? I guess it is not needed and will only slow down filling the collection with data.
Same for the following occurrence of `waitForSync: true` below.

jsteemann (approved):
LGTM

PR #20628, opened 2024-02-16
[CINFRA] Ignore ParticipantGone

apetenchea:
### Scope & Purpose

There are a some cases when "ParticipantGone" can be ignored. In `LogLeader.cpp`, this caused an exception to be thrown in a `noexcept` function.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature


arangod/Replication2/ReplicatedLog/Components/CompactionManager.cpp
```diff
@@ -116,7 +116,18 @@ void CompactionManager::triggerAsyncCompaction(
...
+
+      auto storeRes = basics::catchToResultT([&]() -> auto {
+        return guard->storage.transaction();
+      });
+      if (storeRes.fail()) {
+        // The participant might be gone
+        LOG_CTX("d7724", DEBUG, self->loggerContext)
+            << "error during compaction: " << storeRes.result();
+        break;
+      }
```
goedderz, line 128: You should report the error, roughly as follows:
```suggestion
      if (storeRes.fail()) {
        // The participant might be gone
        LOG_CTX("d7724", ERR, self->loggerContext)
            << "error during compaction: " << storeRes.result();
        cresult.error = std::move(storeRes).result().error();
        guard->status.lastCompaction->error = cresult.error;
        clearCompactionRunning.fire();
        break;
      }
```
Please double-check the code, it was written blindly.

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -312,6 +310,14 @@ void replicated_log::LogLeader::executeAppendEntriesRequests(
...
                           return {};
                         });
+                    if (responseResult.fail()) {
+                      LOG_TOPIC("a32de", DEBUG, Logger::REPLICATION2)
+                          << "appendEntries failed, follower already gone: "
+                          << responseResult.result();
+                    }
+
+                    auto [preparedRequests, resolvedPromises] =
+                        std::move(responseResult.get());
```
goedderz, line 320: I don't think it's okay to silently stop here - the leader will stop working in a situation where it is not expected to. I rather think we should explicitly ignore resigned errors, but continue to crash with everything else.

goedderz (changes requested):

arangod/Replication2/ReplicatedLog/Components/CompactionManager.cpp
apetenchea: Addressed in https://github.com/arangodb/arangodb/pull/20628/commits/444c825fbff19e5f971ef36f315a0c7c2003a184

arangod/Replication2/ReplicatedLog/LogLeader.cpp
apetenchea: Addressed in https://github.com/arangodb/arangodb/pull/20628/commits/5b019267229a77636c84ae0a2b0e65035f7c3c36

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -273,69 +273,80 @@ void replicated_log::LogLeader::executeAppendEntriesRequests(
...
+                            << "received response from follower but leader "
+                               "already resigned, messageId = "
+                            << messageId;
+                      }
+                      return {};
+                    });
+                if (responseResult.fail()) {
+                  LOG_TOPIC("a32de", DEBUG, Logger::REPLICATION2)
+                      << "appendEntries failed, follower already gone: "
+                      << responseResult.result();
```
goedderz, line 317: The log message might not be accurate, unless you move it into the `if` below

arangod/Replication2/ReplicatedLog/LogLeader.cpp
```diff
@@ -273,69 +273,80 @@ void replicated_log::LogLeader::executeAppendEntriesRequests(
...
+                          TRI_ERROR_REPLICATION_REPLICATED_LOG_PARTICIPANT_GONE)) {
+                    // Ignore resign errors
+                    return;
                   }
-                });
+
+                  TRI_ASSERT(false)
+                      << "appendEntries failed, follower already gone: "
+                      << responseResult.result();
+                  FATAL_ERROR_EXIT();
```
goedderz, line 328:
```suggestion
                  ADB_PROD_ASSERT(responseResult.ok())
                      << "appendEntries failed with an unhandled error: "
                      << responseResult.result();
```

arangod/Replication2/ReplicatedLog/LogLeader.cpp
apetenchea: Good catch, addressed in https://github.com/arangodb/arangodb/pull/20628/commits/7f9c83b10f065a508acf68dc96cb815bd452dcd0

arangod/Replication2/ReplicatedLog/Components/CompactionManager.cpp
```diff
@@ -126,16 +126,19 @@ void CompactionManager::triggerAsyncCompaction(
...
-                    CompactionStopReason::StorageManagerNotReady{}},
-            .compactedRange = {}};
-        guard->status.lastCompaction->error = cresult.error;
+
+        result::Error err{storeRes.result().errorNumber(),
+                          storeRes.result().errorMessage()};
+        auto& compaction = guard->status.inProgress.emplace();
+        guard->status.lastCompaction = guard->status.inProgress;
+        guard->status.inProgress.reset();
+        compaction.error = std::move(err);
```
goedderz, line 135: This looks a little convoluted, maybe a c&p relic? Why do you emplace `inProgress`, then copy it to `lastCompaction`, then reset it again, and _then_ write the `.error` field? Isn't this writing in the middle of a  freshly reset optional?

The way I see it: If `inProgress` is there, use that. If it's not, construct a new one, but don't forget to set the `.time` field to `CompactionStatus::clock::now()`. E.g.
```suggestion
        if (guard->status.inProgress.has_value()) {
          guard->status.lastCompaction = guard->status.inProgress;
          guard->status.inProgress.reset();
        } else {
          guard->status.lastCompaction = Compaction();
          guard->status.lastCompaction.time = CompactionStatus::clock::now();
        }
        guard->status.lastCompaction.error = std::move(err);
```
(which suggests that now `Compaction::range` should be an optional, but let's keep it at that ;).

apetenchea: Thanks, I simplified it in https://github.com/arangodb/arangodb/pull/20628/commits/f480fa0c1aa8c6cb6615763e7aafc7bf2e60b58d

PR #20630, opened 2024-02-16
Unify Projections Into Registers

maierlars:
### Scope & Purpose
Projections into registers was first introduced for the late materialize mode for indexes. If was built on the side and did not integrate with the _Projections_ used in other places. Recently projections into registers was added to the normal projections and the old special handling can now be removed.


arangod/Aql/OptimizerUtils.cpp
```diff
@@ -1113,6 +1113,44 @@ bool findProjections(ExecutionNode* n, Variable const* v,
...
+  Projections projections{std::move(projectedAttributes)};
+
+  std::size_t i = 0;
+  for (auto [var, fieldIndex] : indexVars.second) {
+    auto& proj = projections[i++];
+    proj.coveringIndexPosition = fieldIndex;
+    proj.coveringIndexCutoff = fieldIndex;
+    proj.variable = var;
+    proj.levelsToClose = proj.startsAtLevel = 0;
+    proj.type = proj.path.size() > 1 ? AttributeNamePath::Type::MultiAttribute
```
jsteemann, line 1146: Is it possible that we had `_key`, `_id`, `_from` or `_to` here? Then we probably need to turn the type into `AttributeNamePath::Type::KeyAttribute` etc.

maierlars: Fixed it. This could is supposed to be removed anyways.

arangod/Aql/IndexNode.cpp
```diff
@@ -698,13 +652,17 @@ void IndexNode::setLateMaterialized(aql::Variable const* docIdVariable,
                                     IndexId commonIndexId,
                                     IndexVarsInfo const& indexVariables) {
   _outNonMaterializedDocId = docIdVariable;
-  _outNonMaterializedIndVars.first = commonIndexId;
-  _outNonMaterializedIndVars.second.clear();
-  _outNonMaterializedIndVars.second.reserve(indexVariables.size());
+  IndexValuesVars oldIndexValuesVars;
+  oldIndexValuesVars.first = commonIndexId;
+  oldIndexValuesVars.second.clear();
```
jsteemann, line 657: `.second()` should already be empty here. Clearing it does not harm, but also does not help.
```suggestion
```

PR #20632, opened 2024-02-19
Add sanitizer builds for CircleCI

mpoeter:
### Scope & Purpose

Enterprise companion: https://github.com/arangodb/enterprise/pull/1451

This PR adds the `sanitizer` pipeline parameter which can be set to either "tsan" or "alubsan". If this parameter is set, the build will only contain workflows for x64 community/enterprise, but no aarch64, and also not non-V8 or non-maintainer builds.

As part of this, I removed all the workflows from the base config, and instead generate them from the `generate_config` python script. Unfortunately CircleCI currently has no way to show the set of parameters which where used to start a pipeline, so the generated workflows (and also the jobs) will have an according suffix, e.g., `x64-enterprise-tsan`.

Because sanitizer builds require a lot more resources, we scale up the resource classes accordingly, and since the sanitizer allocations are not accounted for in our memory tracking, we override the available memory for each instance. This also fixes a long standing issue that in the CI the individual instances may think they have much more memory available than they actually do, because ATM we only consider the physical memory and do not account for cgroup limits imposed by the subordinate docker container. In addition, we also have to increase the test timeout to account for much slower performance.

Last but not least this PR deletes a bunch of tests that were only relevant for MMF, but caused timeout issues when ran with a tsan build, because of their inefficient implementation. Since nowadays these are no longer relevant with the rocksdb engine, in accordance with @jsteemann I have removed them.

Unfortunately the tsan/alubsan builds are still mostly read, because we still have some tests that are timing sensitive and can easily flake, and there is one other test can frequently runs into an assertion in `commitCounts`, but these issues should be addressed separately.

- [x] :pizza: New feature


js/client/modules/@arangodb/testsuites/recovery.js
```diff
@@ -175,7 +175,7 @@ function runArangodRecovery (params, useEncryption, isKillAfterSetup = true) {
   let agentArgs = pu.makeArgs.arangosh(params.options);
   agentArgs['server.endpoint'] = params.instanceManager.findEndpoint();
   if (params.args['log.level']) {
-    agentArgs['log.level'] = params.args['log.level'];
+    agentArgs['log.level'] = ['warning', 'httpclient=debug', 'V8=debug'];
```
jsteemann, line 178: It is unclear if this change was made intentionally, or if it is a leftover from some debugging attempt?
If it is intentional, it is fine. The motivation is just unclear from looking at the change.

jsteemann (approved):
LGTM

js/client/modules/@arangodb/testsuites/recovery.js
mpoeter: I had already reverted that, but hadn't pushed it yet to avoid cancelling the running build. Pushed it just now,

jsteemann (approved):
LGTM

PR #20633, opened 2024-02-19
Fix testRestoreIndexesFulltextLengthZero

mpoeter:
### Scope & Purpose

This PR fixes a sporadic failure of the testRestoreIndexesFulltextLengthZero test, but the underlying issue that caused this sporadic failure is even more severe.

Similar to a previous bugfix (#20526), the fulltext index is always sparse, which must be considered in the index compare function. Because this was not the case, the maintenance would first create the index and report to the agency, but in the next iteration when we compare the local indexes against the plan, the comparison would yield false, so we drop the index again - so the db servers keep creating, dropping, creating, dropping, ....

In case the supervision observes the situation where all servers have reported the index in current, it clears the `isBuilding` flag which indicates to the client that the request has been finished successfully. That is why in many cases the test is green. But even then, the servers keep dropping and recreating the index indefinitely.

- [x] :hankey: Bugfix


hkernbach (approved):
LGTM

jsteemann (approved):
LGTM

PR #20640, opened 2024-02-20
add progress reporting for RocksDB .sst file checksum computation

jsteemann:
### Scope & Purpose

Add progress reporting for RocksDB .sst file checksum computation.
Upon startup, ArangoDB Enterprise Edition will check if for every RocksDB .sst file there is a corresponding .hash file with the SHA256 hash value of the .sst file's contents. Otherwise such .hash file will be created at startup.
Normally, the .hash file creation is triggered by RocksDB whenever an .sst file is written to disk. So there should not be many .hash files missing on a restart. In case of a crash, a few .hash files may be missing and may be recomputed. This should be quick and unintrusive.
However, when upgrading an existing database from the Community Edition to the Enterprise Edition, there will be no .hash files initially, as they are only written out in the Enterprise Edition. So the startup will lead to all these files being computed, which can take a long time. For exactly this scenario there is now a progress reporting, so whoever runs into this situation can self-diagnose what is going on.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Add startup progress reporting for RocksDB .sst file checksum computation.
+
 * Upgrade OpenSSL to 3.2.1.
 
 * Remove arangosync binary.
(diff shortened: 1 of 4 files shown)


apetenchea (approved):
LGTM, nice description

PR #20645, opened 2024-02-21
CircleCI full nightly builds

mpoeter:
### Scope & Purpose

Previously the CircleCI nightly builds ran normal PR builds. This PR introduces a `nightly` parameter which allows one to run a completely nightly build including all long-running tests.

The workflows for a build with this flag will have a `-nightly` suffix to indicate that.

In addition, this PR removes some outdated/no longer relevant tests/suites.

For normal PR runs we now only run enterprise build+test on aarch64 to reduce CI cost, and in return we will run daily sanitizer tests.

- [x] :pizza: New feature


scripts/test/site_config.py
```diff
@@ -109,15 +109,17 @@ def __init__(self, definition_file):
             self.portbase = int(os.environ["PORTBASE"])
         self.port_offset = 100
         self.timeout = 1800
+        if "timeLimit".upper() in os.environ:
+            self.timeout = int(os.environ["timeLimit".upper()])
+        elif "timeLimit" in os.environ:
+            self.timeout = int(os.environ["timeLimit"])
```
jsteemann, line 115: Why do we need to check for both the environment variables `timeLimit` and `TIMELIMIT` here?
Isn't one enough?

jsteemann (approved):
LGTM

scripts/test/site_config.py
mpoeter: Yes, it was like that before (I just moved the code block to a different place). I didn't want to change the logic because I was not sure whether this would have some unintended side effects. But AFAIK this script is only run from CircleCI, so it should probably be safe.

CryptoNinjaGeek (approved):

PR #20648, opened 2024-02-22
Allow limiting async prefetching

jsteemann:
### Scope & Purpose

Allow limiting async prefetching

* Allow limiting the amount of async prefetching operations for AQL queries. Since version 3.12, AQL queries can make use of async prefetching. While this is normally beneficial for query performance, the async prefetch operations may cause congestion in the ArangoDB scheduler and interfere with other operations. In order to control the maximum amount of prefetch operations, the following startup options have been added:

  - `--query.max-total-async-prefetch-slots`: the maximum total number of slots available for asynchronous prefetching, across all AQL queries.
  - `--query.max-query-async-prefetch-slots`: the maximum per-query number of slots available for asynchronous prefetching inside any AQL query.

  That means the total number of concurrent prefetch operations across all AQL
  queries can be limited using the first option, and the maximum number of
  prefetch operations in every single AQL query can be capped with the second
  option. These options prevent that running a lot of AQL queries with async
  prefetching fully congests the scheduler queue, and also they prevent large
  AQL queries to use up all async prefetching capacity on their own.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/ExecutionEngine.cpp
```diff
@@ -257,16 +259,26 @@ ExecutionEngine::~ExecutionEngine() {
   if (_sharedState) {  // ensure no async task is working anymore
     _sharedState->invalidate();
   }
+
+  if (_asyncPrefetchSlotsReserved > 0) {
+    // return leased slots to QueryRegistryFeature
+    _query.vocbase()
+        .server()
+        .getFeature<QueryRegistryFeature>()
+        .returnAsyncPrefetchSlots(_asyncPrefetchSlotsReserved);
```
mpoeter, line 268: Can we store a reference to the feature instead of using `getFeature` here? Ideally we would take a reference to the feature in the constructor, but I would also accept a single `getFeature` call in the constructor to populate the member.

Actually, after think about this a bit more, I think this tracking logic should not be part of the `QueryRegistryFeature` at all. I would rather put it into a separate class, and have the ExecutionEngine keep a reference to that class. The `QueryRegistryFeature` can still be the owner of that instance (since it also defines the program options, and we currently have no simple way to move them somehwere that is not a feature), but that way the ExecutionEngine at least has no dependency on the the QueryRegistryFeature.

Even better would be to use RAII and have a guard class that returns the slots,

arangod/RestServer/QueryRegistryFeature.cpp
```diff
@@ -794,6 +832,42 @@ void QueryRegistryFeature::unprepare() {
...
+      return 0;
+    }
+    TRI_ASSERT(target > expected);
+    if (_asyncPrefetchSlotsUsed.compare_exchange_weak(
+            expected, target, std::memory_order_relaxed,
+            std::memory_order_relaxed)) {
+      return target - expected;
+    }
+    TRI_ASSERT(expected <= _maxAsyncPrefetchSlotsTotal);
+  } while (true);
```
mpoeter, line 859: I would like to reformulate/simplify this a bit:
```suggestion
  size_t expected = _asyncPrefetchSlotsUsed.load(std::memory_order_relaxed);
  TRI_ASSERT(expected <= _maxAsyncPrefetchSlotsTotal);
  // as long as we have slots left, try to allocate some
  while (expected < _maxAsyncPrefetchSlotsTotal) {
    size_t target = std::min(_maxAsyncPrefetchSlotsTotal, expected + value);
    TRI_ASSERT(target > expected);
    if (_asyncPrefetchSlotsUsed.compare_exchange_weak(
            expected, target, std::memory_order_relaxed)) {
      return target - expected;
    }
    TRI_ASSERT(expected <= _maxAsyncPrefetchSlotsTotal);
  };
```

arangod/Aql/ExecutionEngine.cpp
```diff
@@ -744,6 +763,31 @@ void ExecutionEngine::instantiateFromPlan(Query& query, ExecutionPlan& plan,
...
+    // the QueryRegistryFeature keeps an overview of how many slots have been
+    // leased by the currently running queries.
+    size_t numAsyncPrefetchNodesInPlan = plan.asyncPrefetchNodes();
+    if (numAsyncPrefetchNodesInPlan > 0) {
+      size_t asyncPrefetchSlots =
+          query.vocbase()
+              .server()
+              .getFeature<QueryRegistryFeature>()
+              .leaseAsyncPrefetchSlots(numAsyncPrefetchNodesInPlan);
+      // we may get less slots that we asked for
```
mpoeter, line 782:
```suggestion
      // we may get less slots than we asked for
```

arangod/RestServer/QueryRegistryFeature.h
```diff
@@ -24,9 +24,12 @@
 #pragma once
 
 #include "RestServer/arangod.h"
+#include "Aql/AsyncPrefetchSlotsManager.h"
 #include "Aql/QueryRegistry.h"
 #include "Metrics/Fwd.h"
 
+#include <atomic>
+
```
mpoeter, line 32: No longer necessary
```suggestion
#include <atomic>

```

arangod/Aql/AsyncPrefetchSlotsManager.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+AsyncPrefetchSlotsReservation::AsyncPrefetchSlotsReservation() noexcept
+    : _manager(nullptr), _slotsReserved(0) {}
+
+AsyncPrefetchSlotsReservation::AsyncPrefetchSlotsReservation(
+    AsyncPrefetchSlotsManager* manager, size_t slotsReserved) noexcept
+    : _manager(manager), _slotsReserved(slotsReserved) {
+  // if number of slots is > 0, we must have a manager
+  TRI_ASSERT(_slotsReserved == 0 || _manager != nullptr);
+}
```
mpoeter, line 40: Can we just take the manager by reference? Then it is always guaranteed to be non-null.

arangod/Aql/AsyncPrefetchSlotsManager.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // as long as we have slots left, try to allocate some
+    while (expected < _maxSlotsTotal) {
+      size_t target = std::min(_maxSlotsTotal, expected + value);
+      TRI_ASSERT(target > expected);
+      if (_slotsUsed.compare_exchange_weak(expected, target,
+                                           std::memory_order_relaxed)) {
+        return {this, target - expected};
+      }
+      TRI_ASSERT(expected <= _maxSlotsTotal);
+    }
```
mpoeter, line 100: Maybe we should rename `expected` to `current`. IMHO this makes the code clearer and easier to understand
```suggestion
    size_t current= _slotsUsed.load(std::memory_order_relaxed);
    TRI_ASSERT(current <= _maxSlotsTotal);
    // as long as we have slots left, try to allocate some
    while (current < _maxSlotsTotal) {
      size_t target = std::min(_maxSlotsTotal, current + value);
      TRI_ASSERT(target > current);
      if (_slotsUsed.compare_exchange_weak(current, target,
                                           std::memory_order_relaxed)) {
        return {this, target - current};
      }
      TRI_ASSERT(current <= _maxSlotsTotal);
    }
```

mpoeter (approved):
A few small suggestions, but otherwise LGTM

mpoeter (approved):

PR #20652, opened 2024-02-23
Add LGPL greeting at startup to fulfill LGPL terms

neunhoef:
### Scope & Purpose

This PR adds a greeting message at the start of each executable we ship,
which is linked statically against the glibc (provided we build with
glibc).

- [*] Legal requirement


fceller (approved):
LGTM

PR #20656, opened 2024-02-23
Change license of velocypack to BSL, adjust headers

neunhoef:
This updates Velocypack to a version which has the BSL 1.1 license
change. No enterprise part.


jsteemann (approved):
LGTM

PR #20664, opened 2024-02-28
arangodump: Set all parallel dump versions

Simran-B:
Backport of #20663

### DIFF
--- a/client-tools/Dump/DumpFeature.cpp
+++ b/client-tools/Dump/DumpFeature.cpp
@@ -933,8 +933,7 @@ servers running version 3.12 or higher.
 If the dump should be restored into versions of ArangoDB older than 3.12, this
 option should be turned off.)")
       .setIntroducedIn(31008)
-      .setIntroducedIn(31102)
-      .setIntroducedIn(31200);
+      .setIntroducedIn(31102);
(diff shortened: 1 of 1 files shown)


KVS85 (approved):

jsteemann:
Note that the `--parallel-dump` option did not exist before 3.12.0.
It was named `--experimental-dump` in the versions before that. 3.12.0 renames it to `--parallel-dump`, because it is not experimental anymore since 3.12.0.
I don't think we should advertise the option for versions older than 3.12.0, because it had a different name there and it was experimental and is likely unsupported there.

CryptoNinjaGeek (approved):

Simran-B:
@jsteemann  In the docs PR, the experimental status and that the option is named `--use-parallel-dump` are pointed out:
https://github.com/arangodb/docs-hugo/pull/484/files
The PR series addresses the issue of the option showing v3.12.0 as the version it was introduced in, under the original name in 3.10 and 3.11. A user specifically asked about `--split-files` which shows up in the 3.10 docs with the hint that it has been introduced in 3.12, which is pretty confusing:
https://arangodb-community.slack.com/archives/C07NV9NAH/p1709112670636019

PR #20687, opened 2024-03-05
Verify transaction database name

mpoeter:
### Scope & Purpose

Transactions are bound to the user who started it, and the database it was started for. However, there was a code path that was missing the database check, which allowed the transaction id to be used in a request to a different database. The request would not return an error, but it would actually be executed in the context of the original database, and not the one specified in the request.

With this fix, such a request will now return an ERROR_TRANSACTION_NOT_FOUND.

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] **Regression tests**
- [ ] Backports

#### Related Information

- [x] Jira ticket: https://arangodb.atlassian.net/browse/BTS-1807

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,13 @@
 devel
 -----
 
-* 
+* Fix BTS-1807: transactions are bound to the user who started it, and the
+  database it was started for. However, there was a code path that was missing
+  the database check, which allowed the transaction id to be used in a request
+  to a different database. The request would not return an error, but it would
+  actually be executed in the context of the original database, and not the one
+  specified in the request.
+  With this fix, such a request will now return an ERROR_TRANSACTION_NOT_FOUND.
 
 
 3.12.0-alpha.1 (XXXX-XX-XX)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,7 @@ devel
 -----
 
 * Fix BTS-1807: transactions are bound to the user who started it, and the
-  database it was started for. However, there was a code path that was missing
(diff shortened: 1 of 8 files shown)


MBkkt (approved):

CHANGELOG
```diff
@@ -1,7 +1,13 @@
 devel
 -----
 
-* 
+* Fix BTS-1807: transactions are bound to the user who started it, and the
+  database it was started for. However, there was a code path that was missing
```
```suggestion
  database they were started for. However, there was a code path that was missing
```

tests/js/client/shell/transaction/shell-transaction.inc
```diff
@@ -5719,6 +5719,20 @@ function transactionDatabaseSuite() {
...
+      const opts = { collections: { write: [cn] } };
+      const trx = db._createTransaction(opts);
+      trx.collection(cn).insert({ _key: "test" });
+      try {
+        db._useDatabase(dbn2);
+        trx.collection(cn).document("test");
+        fail();
+      } catch(err) {
+        assertEqual(internal.errors.ERROR_TRANSACTION_NOT_FOUND.code, err.errorNum);
+      }
```
jsteemann, line 5734: Should there be a `finally` clause here to switch back to the original database?
Otherwise the test case may reset some global state (i.e. the current database) and thus have side-effects for following tests. (note: there are currently no tests failing the new one, but I consider it good practice if a test cleans up after itself).

mpoeter: Good point! 👍

jsteemann (approved):
LGTM

PR #20694, opened 2024-03-06
Merge current 3.12.0 state back to devel

KVS85:
Merge current 3.12.0 state back to devel carefully.


KVS85:
Closed as not needed.

PR #20698, opened 2024-03-06
Expose number of read-only shards because of write-concern

maierlars:
### Scope & Purpose

This PR adds a metric, that counts the number of shards per database that are currently in read-only mode, because there are not enough in-sync followers (less than write concern).


arangod/Metrics/InstrumentedBool.h
```diff
@@ -0,0 +1,86 @@
...
+  InstrumentedBool& operator=(bool value) {
+    set(value);
+    return *this;
+  }
+
+  void set(bool value) {
+    if (value == _value) {
+      return;
+    }
+    _value = value;
```
jsteemann, line 61: Does this part need to be thread-safe?
What if different threads try to push concurrently to the same InstrumentedBool instance?

maierlars: It is not thread safe, similar to any other `bool` variable.

jsteemann: Ok, thanks for confirming. It think it is good as it is.

jsteemann (approved):
LGTM

PR #20714, opened 2024-03-11
[EF-51] Extend ExecContext's lifetime

goedderz:
### Scope & Purpose

The ExecContext's lifetime was tied to the request's lifetime; however, some task run after a request is finished. For example, in Replication 2.0, an appendEntries request can trigger a bunch of different background tasks, like applying entries or acquiring a snapshot. This led to UB, and sporadic test failures. Most often I've seen this in `replication2-replicated-state-document-store.js` and `testFollowerSnapshotTransfer`, but also in others. This can happen (at least in the particular case I've investigated) since the NetworkFeature uses coroutines, which save and restore the old ExecContext.

This PR makes it possible to extend the ExecContext's lifetime, by making it a shared_ptr, which is then copied when a coroutine is suspended.

- [X] :hankey: Bugfix

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/browse/EF-51

### DIFF
--- a/arangod/Aql/ExecutionBlockImpl.h
+++ b/arangod/Aql/ExecutionBlockImpl.h
@@ -447,7 +447,7 @@ class ExecutionBlockImpl final : public ExecutionBlock {
       LogContext logContext;
     };
 
-    void run(std::unique_ptr<ExecContext> execContext);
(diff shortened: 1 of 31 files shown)


jsteemann (approved):
LGTM

PR #20719, opened 2024-03-11
Repair shard rebalancer [BTS-1814]

neunhoef:
This fixes a problem in shard rebalancing. Here is a description of the
problem:

  https://arangodb.atlassian.net/browse/BTS-1814

If for some collection a server is not the leader for any of its
shards, but has some follower shards. Then the leader imbalance score
does not take into account this server. Therefore, the system does not
consider leader moves to this server, since they often do not
improve the score. Therefore the rebalancer often only rebalances
leaders between servers which already have leader shards.

Furthermore, there is another small problem that some parameter of the
algorithm is only set from the user's input after the computation of the
imbalance score "before" has already happened. This is also fixed in
this PR.

- Set pifactors before computing starting imbalances.
- Repair piCoefficients computation.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **Regression tests**
  - [*] C++ **Unit tests**
- [*] Backports
  - [*] Backport for 3.12.0: https://github.com/arangodb/arangodb/pull/20730
  - [*] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20723

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1814


arangod/Cluster/AutoRebalance.cpp
```diff
@@ -230,29 +230,36 @@ ShardImbalance AutoRebalanceProblem::computeShardImbalance() const {
 
 std::vector<double> AutoRebalanceProblem::piCoefficients(
     Collection const& c) const {
-  std::vector<double> tmp;
-  tmp.resize(dbServers.size(), 0);
+  std::vector<double> leaders;
+  leaders.resize(dbServers.size(), 0);
+  std::vector<bool> haveShards;
+  haveShards.resize(dbServers.size(), 0);
```
jsteemann, line 236: For improved readability, I guess we should set each member to `false` in a vector<bool> and not to `0`.
```suggestion
  haveShards.resize(dbServers.size(), false);
```

arangod/Cluster/AutoRebalance.cpp
```diff
@@ -230,29 +230,36 @@ ShardImbalance AutoRebalanceProblem::computeShardImbalance() const {
...
+  uint32_t dbServersAffected = 0;
   if (c.shards.size() == 1) {
```
jsteemann, line 238: This if condition can now be moved up before the initialization of `haveShards` and `dbServersAffected`, because the latter have no influence on the if condition and what's happening in the if condition's execution block.

jsteemann (approved):
LGTM

PR #20722, opened 2024-03-11
refactor ternary operator condition evaluation

jsteemann:
### Scope & Purpose

Refactor, simplify and improve the handling of conditional operands for the following AQL constructs:

* ternary operator: `condition ? truePart : falsePart`: the `condition` part is now guaranteed to be evaluated only a single time. the `truePart` is evaluated only if `condition` is truthy. The `falsePart` is evaluated only if `condition` is falsy.
* logical AND: `condition && truePart`: the `truePart` is evaluated only if `condition` is truthy.
* logical OR: `condition || falsePart`: the `falsePart` is evaluated only if `condition` is falsy.

This holds true for all operand types, even if the operands are subqueries.

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Aql/LazyConditions.cpp
```diff
@@ -0,0 +1,205 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2014-2024 ArangoDB GmbH, Cologne, Germany
+/// Copyright 2004-2014 triAGENS GmbH, Cologne, Germany
```
mchacki, line 5: The copyright note looks a bit outdated 😁

arangod/Aql/Parser.h
```diff
@@ -24,29 +24,38 @@
...
 class Parser {
+ public:
+  struct LazyCondition {
+    // the actual condition expression (always non-null)
+    AstNode* condition;
+    // the variable node that is assigned the condition expression's result.
+    // this is null initially, but can be populated on demand
+    AstNode* variable = nullptr;
+  };
+
```
mchacki, line 46:
```suggestion
```

This struct is not used, it has been moved to LazyConditions.h

tests/js/client/aql/aql-ternary.js
```diff
@@ -468,6 +491,35 @@ RETURN NOOPT(true) ?
       assertEqual(true, result[0]);
     },
 
+    testShortcutTernary : function () {
```
mchacki, line 494: I would like the same test with swapped positions.

`RETURN NOOPT(false) ? ASSERT(false, 'fail') : `

To make sure that also the "positive" code path is not taken if the condition yields FALSE

mchacki (changes requested):
Looks very good to me.

Two Minor things:
1) There is a struct that is not used anymore.
2) I would like another very small test.

arangod/Aql/LazyConditions.cpp
jsteemann: It is actually a current one, including a reference to the "Business Source License" a few lines below.

Reading the copyright years, I can change it to "Copyright 2024 ArangoDB GmbH, Cologne, Germany" only. I don't know if this is legally correct or if for some reason we need to retain the reference to triAGENS in there.
My personal preference would be to remove the reference to the years entirely, so we could unify the header across all files and don't need to bother about it when a new year begins etc.

mchacki (approved):
LGTM! Thank you for the changes

PR #20723, opened 2024-03-12
Fix shard rebalancer [BTS-1814]

neunhoef:
This is a backport from devel: https://github.com/arangodb/arangodb/pull/20719

This fixes a problem in shard rebalancing. Here is a description of the
problem:

  https://arangodb.atlassian.net/browse/BTS-1814

If for some collection a server is not the leader for any of its
shards, but has some follower shards. Then the leader imbalance score
does not take into account this server. Therefore, the system does not
consider leader moves to this server, since they often do not
improve the score. Therefore the rebalancer often only rebalances
leaders between servers which already have leader shards.

Furthermore, there is another small problem that some parameter of the
algorithm is only set from the user's input after the computation of the
imbalance score "before" has already happened. This is also fixed in
this PR.

- Set pifactors before computing starting imbalances.
- Repair piCoefficients computation.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [*] Tests
  - [*] **Regression tests**
  - [*] C++ **Unit tests**
- [*] Backports

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1814

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 v3.11.9 (XXXX-XX-XX)
 --------------------
 
+* Repair shard rebalancer if some server has no leaders of a collection
+  but followers. Previously, this server was then not always considered
+  for leader movements.
+
 * Updated ArangoDB Starter to v0.18.4 and arangosync to v2.19.7.
 
 * BTS-1808: fix arangodump format for ZKD indexes. Previously the structural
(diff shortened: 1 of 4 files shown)


jsteemann (approved):
LGTM

PR #20726, opened 2024-03-12
Feature/instrumented mutex wrapper

maierlars:
### Scope & Purpose
This PR adds an `InstrumentedMutex` wrapper class that collects metrics about the mutex it instruments. These metrics contain:
- number of pending shared/exclusive lock requests
- number of shared/exclusive locks.

Later I want to add Histograms tracking the lock and wait time. But this PR is already big enough.

As an application the exclusive lock in the `RocksDBMetaCollection` has been instrumented and new metrics were introduced. As a side effect, the locking methods now return a guard object that is stored in the `RocksDBTransactionCollection` object.

This guard object is necessary for reconstructing the lock times. Otherwise, one can not assign the lock and unlock of a shared lock.


Documentation/Metrics/arangodb_vocbase_meta_collection_lock_locked_exclusive.yaml
```diff
@@ -0,0 +1,14 @@
...
+unit: number
+type: gauge
+category: Transactions
+complexity: advanced
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  Counts the exclusively locked collection across all shards in a database.
```
jsteemann, line 14:
```suggestion
  Counts the exclusively locked collections across all shards in a database.
```

Documentation/Metrics/arangodb_vocbase_meta_collection_lock_locked_exclusive.yaml
```diff
@@ -0,0 +1,14 @@
+name: arangodb_vocbase_meta_collection_lock_locked_exclusive
```
jsteemann, line 1: I don't think we should name a publicly visible metric something like "*vocbase*". "vocbase" is some internal term that doesn't make much sense to end users. The same is true for "meta_collection". 
Why not go with some simpler name such as "arangodb_collection_locks_locked_*"? If that name is already in use, maybe some similar easy alternative?

Documentation/Metrics/arangodb_vocbase_meta_collection_lock_locked_exclusive.yaml
jsteemann: Maybe also note that on a DB server, the number reported by the metric is the number of exclusively locked _shards_, not _collections_.

Documentation/Metrics/arangodb_vocbase_meta_collection_lock_locked_shared.yaml
```diff
@@ -0,0 +1,14 @@
...
+unit: number
+type: gauge
+category: Transactions
+complexity: advanced
+exposedBy:
+  - dbserver
+  - agent
+  - single
+description: |
+  Counts the shared locked collection across all shards in a database.
```
jsteemann, line 14:
```suggestion
  Counts the shared locked collections across all shards in a database.
```

arangod/Metrics/CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ add_library(arango_metrics STATIC
   MetricsFeature.cpp
   ClusterMetricsFeature.cpp
   ${PROJECT_SOURCE_DIR}/arangod/RestHandler/RestMetricsHandler.cpp
-  ${PROJECT_SOURCE_DIR}/arangod/RestHandler/RestUsageMetricsHandler.cpp)
+  ${PROJECT_SOURCE_DIR}/arangod/RestHandler/RestUsageMetricsHandler.cpp
+        InstrumentedMutex.cpp)
```
jsteemann, line 18: nit: the indentation here is somewhat unexpected, but it doesn't matter much.

arangod/RocksDBEngine/RocksDBReplicationContext.cpp
```diff
@@ -264,17 +264,13 @@ RocksDBReplicationContext::bindCollectionIncremental(TRI_vocbase_t& vocbase,
...
-    if (!_patchCount.empty() && _patchCount == cname &&
-        co_await rcoll->lockWrite(to) == TRI_ERROR_NO_ERROR) {
-      // fetch number docs and snapshot under exclusive lock
-      // this should enable us to correct the count later
-      documentCountAdjustmentTicket =
-          rcoll->meta().documentCountAdjustmentTicket();
-    } else {
-      lockGuard.cancel();
+    if (!_patchCount.empty() && _patchCount == cname) {
+      auto [guard, res] = co_await rcoll->lockExclusive(to);
```
jsteemann, line 268: I think this leads to changed behavior.
In the old implementation, the lock acquired by `rcoll->lockWrite(to)` acquired on line old:269 could be held until the end of the entire code block. The lock may have been released only on line old:289.

In the new implementation, we are acquiring the lock in line 268 and release it on line 273 already, which is way earlier than before.
So with the new implementation, even if the lock acquistion is successful, we don't execute the following code under the lock, but previously we did:
``` 
    // re-acquire the mutex. we need it for checking and modifying _iterators.
    writeLocker.lock();

    it = _iterators.find(cid);
    if (it !=
        _iterators.end()) {  // someone else inserted the iterator in-between
      co_return std::make_tuple(Result{}, it->second->logical->id(),
                                it->second->numberDocuments);
    }
    numberDocuments = rcoll->meta().numberDocuments();
    lazyCreateSnapshot();
```

arangod/RocksDBEngine/RocksDBTransactionCollection.h
```diff
@@ -177,5 +178,8 @@ class RocksDBTransactionCollection : public TransactionCollection {
 
   bool _usageLocked;
   bool _exclusiveWrites;
+  std::variant<std::monostate, RocksDBMetaCollection::ExclusiveLock,
```
jsteemann, line 181: nit: as we are now using std::variant in this header file, we should also add `#include <variant>` at the top of the file.

lib/Basics/FutureSharedLock.h
```diff
@@ -273,3 +273,64 @@ struct FutureSharedLock {
...
+template<typename Scheduler>
+struct InstrumentedMutexLockGuardTraits<
+    futures::FutureSharedLock<Scheduler>,
+    typename futures::FutureSharedLock<Scheduler>::LockGuard> {
+  using LockGuard = futures::FutureSharedLock<Scheduler>::LockGuard;
+  void unlock_exclusive(LockGuard&& guard) { guard.unlock(); }
+
+  void unlock_shared(LockGuard&& guard) { guard.unlock(); }
+
+  bool owns_lock(LockGuard& guard) { return guard.isLocked(); }
```
jsteemann, line 334: nit:
```suggestion
  bool owns_lock(LockGuard const& guard) const noexcept { return guard.isLocked(); }
```

tests/Metrics/InstrumentedMutexTest.cpp
```diff
@@ -0,0 +1,117 @@
...
+  ASSERT_FALSE(guard2);
+
+  auto guard2_1 = m.try_lock_shared();
+  ASSERT_FALSE(guard2.owns_lock());
+  ASSERT_FALSE(guard2);
+
+  guard.unlock();
+  ASSERT_FALSE(guard.owns_lock());
+  ASSERT_EQ(lockExclusive.load(), 0);
+  ASSERT_EQ(pendingExclusive.load(), 0);
```
jsteemann, line 102: same here

tests/js/client/shell/transaction/shell-transaction-rocksdb-meta-collection-lock-metrics-cluster.js
```diff
@@ -0,0 +1,143 @@
...
+    assertEqual(expected.arangodb_vocbase_meta_collection_lock_locked_exclusive, actual[2]);
+    assertEqual(expected.arangodb_vocbase_meta_collection_lock_locked_shared, actual[3]);
+  };
+
+  return {
+    setUpAll: function () {
+      db._createDatabase(database);
+      db._useDatabase(database);
+      db._create(collection, {numberOfShards: 1, replicationFactor: 1});
+      waitForShardsInSync(collection);
```
jsteemann, line 63: nit: with numberOfShards == 1 && replicationFactor == 1, I don't think we need to wait for any shard to get into sync.

tests/js/client/shell/transaction/shell-transaction-rocksdb-meta-collection-lock-metrics-cluster.js
```diff
@@ -0,0 +1,143 @@
...
+    },
+
+    tearDownAll: function () {
+      db._useDatabase('_system');
+      db._dropDatabase(database);
+    },
+
+    testSharedLock: function () {
+      {
+        assertMetrics({
```
jsteemann, line 73: I think this entire test can spuriously fail if there is a write operation ongoing in some other collection in the background. During the tests, this can possibly happen due to some statistics background thread writing data into any of the statistics collections or running a cleanup query on them.

What we could do to work around interferences by temporary background operations is to add a few retries to `assertMetrics`, until the desired target state has been reached or some timeout has been exceeded.

tests/js/client/shell/transaction/shell-transaction-rocksdb-meta-collection-lock-metrics-cluster.js
```diff
@@ -0,0 +1,143 @@
...
+          arangodb_vocbase_meta_collection_lock_pending_shared: 0,
+          arangodb_vocbase_meta_collection_lock_locked_exclusive: 0,
+          arangodb_vocbase_meta_collection_lock_locked_shared: 0,
+        });
+      }
+    },
+
+    testExclusiveLock: function () {
+      {
+        assertMetrics({
```
jsteemann, line 107: same issue here as mentioned for above test.

arangod/RocksDBEngine/RocksDBEngine.cpp
```diff
@@ -2058,11 +2058,6 @@ Result RocksDBEngine::dropCollection(TRI_vocbase_t& vocbase,
   bool const prefixSameAsStart = true;
   bool const useRangeDelete = rcoll->meta().numberDocuments() >= 32 * 1024;
 
-  auto resLock = rcoll->lockWrite().get();  // technically not necessary
```
jsteemann, line 2061: While the comments mentions that this is "technically not necessary", acquiring the lock here may previously have led to synchronization with all ongoing write operations to the collection being fully finished.
With the lock being removed here now, that behavior may or may not change. I can't tell.
To be on the safe side, we could add the lock back even if we can't prove it is actually needed. Or we try and see.

PR #20734, opened 2024-03-12
Fix a small issue in a bash-if that causes a confusing warnings

mpoeter:
### Scope & Purpose

Without this fix the CircleCI output contains these errors/warnings:
```
/bin/bash: line 2: [: =: unary operator expected
/bin/bash: line 7: [: =: unary operator expected
```


jsteemann (approved):
LGTM

PR #20735, opened 2024-03-13
Update default log format

maierlars:
### Scope & Purpose
Update default log format. Include thread id and utc mirco seconds timestamp.


jsteemann (approved):
LGTM

PR #20739, opened 2024-03-13
small optimizations for AqlItemBlock::toVelocyPack()

jsteemann:
### Scope & Purpose

small optimizations for AqlItemBlock::toVelocyPack()

- use FlatHashMap and not std::unordered_map to remember already seen objects. FlatHashMap should be slightly faster because the table is stored in contiguous memory
- instead of using a sequence of table.find(value) and table.emplace(value, pos), we are now doing unconditional try_emplaces into the table. in case the target element was already in the table, this does nothing. in case the target element was not yet in the table, the call will create it, and we will have saved the additional call to find()
- simplify the API signature of `sanitizeNonClientTypes` by removing the boolean parameters that were always set to `true` anyway on all call sites. Additionally, pass the velocypack::Options parameter by reference so that we can avoid nullptr-checks later.

these changes should not alter the observable behavior in any way.

observed performance improvements for toVelocyPack() are around ~1% with these changes for a test set of values.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

goedderz (approved):

PR #20748, opened 2024-03-15
Accept legcy LANGUAGE detections for C locale [BTS-1824]

neunhoef:
This addresses https://arangodb.atlassian.net/browse/BTS-1824

The problem is that the ICU library was updated for 3.12. The new
version detects the language settings differently for the case that the
locale is set to "C". The new version then uses "en_US_POSIX", whereas
the old version sometimes takes "" (empty string) and sometimes uses
"en_US" (depending on the charset settings). This is a problem in
upgrades, which use the LanguageFeatureCheck feature to FATAL exit,
if the detected language setting is different from the one found during
database directory creation.

This PR "fixes" this problem by being more tolerant in the check, thus
allowing seamless upgrades. Furthermore, new databases in 3.12 would
then be generated with the newer, more correct detection.

See https://unicode-org.github.io/icu/userguide/locale/

- CHANGELOG.
- Accept legacy LANGUAGE settings for "C" locale.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [ ] Backports
  - [ ] Backport for 3.12.0:

#### Related Information

- [*] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1824

### DIFF
--- a/arangod/RestServer/LanguageCheckFeature.cpp
+++ b/arangod/RestServer/LanguageCheckFeature.cpp
@@ -169,6 +169,20 @@ std::tuple<std::string, LanguageType> getOrSetPreviousLanguage(
 
   return {std::string{collatorLang}, currLangType};
 }
+
+bool compareLangWithPrev(std::string const& lang, std::string const& prevLang) {
+  if (lang == prevLang) {
+    return true;
+  }
+  // Earlier ICU versions (for example in ArangoDB <= 3.11) derived different
+  // values for the "C" locale. For upgrades to work, we have to accept these
+  // old settings, too:
+  if (lang == "en_US_POSIX" && (prevLang == "en_US" || prevLang == "")) {
+    return true;
+  }
+  return false;
+}
+
 }  // namespace
 
 namespace arangodb {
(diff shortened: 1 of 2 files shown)


arangod/RestServer/LanguageCheckFeature.cpp
```diff
@@ -169,6 +169,20 @@ std::tuple<std::string, LanguageType> getOrSetPreviousLanguage(
...
 }
+
+bool compareLangWithPrev(std::string const& lang, std::string const& prevLang) {
+  if (lang == prevLang) {
+    return true;
+  }
+  // Earlier ICU versions (for example in ArangoDB <= 3.11) derived different
+  // values for the "C" locale. For upgrades to work, we have to accept these
+  // old settings, too:
+  if (lang == "en_US_POSIX" && (prevLang == "en_US" || prevLang == "")) {
```
jsteemann, line 180: I understand that `en_US_POSIX` and `en_US` should be treated identically.
But should other locales with a `_POSIX` suffix also compare equally to their non-posix counterparts, e.g. `de_DE_POSIX` and `de_DE`?

neunhoef: As far as I see, a locale setting with `de_DE` or anything (and in particular also `de_DE_POSIX`  is always shortened to "de" with respect to what goes into the LANGUAGE file. Therefore, this is not necessary. And I would like to restrict the check to the bare necessary minimum.

jsteemann (approved):
LGTM

PR #20096, opened 2023-11-08
remove cluster-only attributes from single server dumps

jsteemann:
### Scope & Purpose

Previously the arangodump collection properties from single server dumps contained the following attributes:

- replicationFactor
- writeConcern
- numberOfShards
- minReplicationFactor
- shards
- shardKeys

It is not really useful to write these attributes out when taking single server dumps, because they don't have useful values.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [x] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #20760, opened 2024-03-18
Repair dump/restore tests

dothebart:
### Scope & Purpose

We would return false while the rta-checkdata test result is true. This would skip further tests. It was introduced in #19718, and since then those tests haven't been run.

Necessary changes to fix the by-now failing tests are also part of this PR. The most notable changes are the following:

- Change the default of shardingStrategy from `enterprise-compat`/`community-compat` to `hash` in the restore API, and adjust the tests accordingly. This is now consistent with the collection API, and the more sensible default. An inconsistent default would lead to test failures.
- Revert https://github.com/arangodb/arangodb/pull/20096: there are a bunch of tests that check that restoring a cluster-dump to a single server preserves the cluster-specific attributes that were removed in that PR. Other improvements from that PR are preserved. (only applies to devel and 3.12.0)

In addition, the tests got the following fixes:

- Restoring foxx apps used the wrong dump-path to restore
- Two more tests calculated the wrong input path (testHiddenCollectionsOmitted and testShadowCollectionsOmitted)
- `testKeygenAutoInc` got the wrong expected document count in `dump-multiple.js`
- One of the teardown files threw an exception when trying to remove a non-existent database, making the teardown fail
- Another teardown file was missing completely

- [x] :hankey: Bugfix

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -569,17 +569,6 @@ devel
 
 * Fix propagation of collection `writeConcern` changes to DB servers.
 
-* Make arangodump exclude the following cluster collection properties when
-  writing collection parameters files in a single server dump:
-  - replicationFactor
-  - writeConcern
-  - numberOfShards
-  - minReplicationFactor
-  - shardKeys
-  - shards
-  It is not really useful to write these attributes out when taking single
-  server dumps, because they don't have useful values.
-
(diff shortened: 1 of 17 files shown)


mchacki (approved):
LGTM

jsteemann (approved):
LGTM

PR #20764, opened 2024-03-19
do not move uncorrelated subqueries down

jsteemann:
### Scope & Purpose

Do not move uncorrelated subqueries down via optimizer rule `move-calculations-down`.
Doing so can cause performance degradations during query plan optimization for no good reason.

For example, if there is the following query
```
FOR doc IN ...
  LET result1 = (...)
  LET result2 = (...)
  ...
```
and `result1` and `result2` are uncorrelated, then there is no benefit in moving `result1` down past `result2`. 
Doing so anyway will trigger a recalculation of which variables are valid in the plan at any step, which is expensive and thus we want to avoid doing this unnecessarily.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Aql/OptimizerRules.cpp
+++ b/arangod/Aql/OptimizerRules.cpp
@@ -1811,7 +1811,8 @@ void arangodb::aql::moveCalculationsDownRule(
         continue;
       }
       variable = nn->outVariable();
-    } else if (n->getType() == EN::SUBQUERY) {
+    } else {
+      TRI_ASSERT(n->getType() == EN::SUBQUERY);
       auto nn = ExecutionNode::castTo<SubqueryNode*>(n);
       if (!nn->isDeterministic() || nn->isModificationNode()) {
         // we will only move subqueries down that are deterministic and are not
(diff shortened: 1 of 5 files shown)


arangod/Aql/OptimizerRules.cpp
```diff
@@ -1819,6 +1819,8 @@ void arangodb::aql::moveCalculationsDownRule(
         continue;
       }
       variable = nn->outVariable();
+    } else {
+      TRI_ASSERT(false);
```
MBkkt, line 1823: just replace `else if` with `else` and write assert in it.
Can be production assert if you're not sure

MBkkt (approved):

PR #20769, opened 2024-03-20
Feature/circleci rta

dothebart:
### Scope & Purpose

work on the ability to launch rta ui tests from circle-ci.

- one codepath now used to download aux binaries
- configuration switches added to enable and filter for test cases + deployments

added Switches:
- `ui`: 
  - `off` or not specified: disabled.
  - `only` : run just the enterprise UI tests => Skip `community`, `ARM`, `testing.js` tests
  - `community`: run the community UI tests as well
  - anything else: run enterprise UI tests
- `ui-testsuites`: coma separated list of UI testsuites to run (if not all). Cavehead: to parallize, this repo needs to know the list of tests as default. 
- `ui-deployments`: corrosponds to `--starter-mode` from RTA, atm only SG is specified by default.

- [x] :pizza: New feature


KVS85 (changes requested):
`off` should be added (set by default in case of empty `ui` value)

.circleci/base_config.yml
```diff
@@ -42,7 +51,7 @@ parameters:
   # these are basically global constants
   build-targets:
     type: string
-    default: "arangod arangoimport arangoexport arangodump arangorestore arangobench frontend"
+    default: "arangod arangoimport arangoexport arangodump arangorestore arangobench frontend arangovpack"
```
KVS85, line 54: Why do we need `arangovpack` here?

.circleci/base_config.yml
```diff
@@ -233,6 +242,8 @@ jobs:
       s3-prefix:
         type: string
         default: ""
+      arch:
```
KVS85, line 245: Which are possible? What is default?

.circleci/base_config.yml
```diff
@@ -475,6 +511,73 @@ jobs:
...
+             -v $(pwd)/allure-report:/allure-report \
+             -v $(pwd)/allure-results:/allure-results \
+             -v $(pwd)/allure-config:/allure-config \
+             smartcosmos/allure-commandline generate /allure-results -o /allure-report --clean
+             exit $(($exit_code + $(cat ./test_dir/status.json) ))
+      - run:
+          name: Copy test results
+          when: always
+          command: |
+            echo TODO
```
KVS85, line 571: ?

.circleci/generate_config.py
```diff
@@ -72,6 +72,9 @@ def parse_arguments():
     )
     parser.add_argument("-o", "--output", type=str, help="filename of the output")
     parser.add_argument("-s", "--sanitizer", type=str, help="sanitizer to use")
+    parser.add_argument("--ui", type=str, help="which test of UI job to run")
+    parser.add_argument("--ui-testsuites", type=str, help="which test of UI job to run")
```
KVS85, line 76: Both are `which test of UI job to run`?

.circleci/generate_config.py
```diff
@@ -351,7 +368,42 @@ def add_test_definition_jobs_to_workflow(
...
+        "LogInTestSuite",
+        "DashboardTestSuite",
+        "SupportTestSuite",
+        "ServiceTestSuite",
+    ]
+    if args.ui_testsuites != "":
+        ui_testsuites = args.ui_testsuites.split(',')
+    deployments = ['SG'
+                   #"CL",
+                   ]
```
KVS85, line 390: Why is "CL" commented out?

.circleci/generate_config.py
```diff
@@ -351,7 +368,42 @@ def add_test_definition_jobs_to_workflow(
...
+            jobs.append(create_rta_test_job(build_config, build_job, deployment, test_suite))
+
+
+def add_test_jobs_to_workflow(args, workflow, tests, build_config, build_job, repl2):
+    if build_config.arch == "x64" and args.ui != "" and args.ui != "off":
+        if build_config.enterprise:
+            add_rta_ui_test_jobs_to_workflow(args, workflow, build_config, build_job)
+        elif args.ui == "community":
+            add_rta_ui_test_jobs_to_workflow(args, workflow, build_config, build_job)
+        if args.ui == "only":
```
KVS85, line 405: Why do we have logically inequal values within the same `ui` parameter?

KVS85 (changes requested):

.circleci/base_config.yml
dothebart: RTA expects to find all binaries that are there in distribution packages. `arangovpack` is part of that. We probably should have tests for it?

.circleci/generate_config.py
dothebart: should we run both by default or not? since its doubling the number of jobs. Basically a management decision.

.circleci/generate_config.py
dothebart: - we will allways run enterprise.
- if `community` we will those as well. 
- if `only` we won't run `testing.js` tests that would be added below.

.circleci/base_config.yml
dothebart: this is generated by `generate_config.py`; it will depend on the current target architecture. 
Its string is later on adjusted to download the go aux-binaries.

.circleci/generate_config.py
palashkaria: Questions/suggestions

1. What is `testing.js`?
2. what do you think about making `ui` an array, which can have any of these values: `'enterprise' | 'community' | 'ARM' | 'testing.js'`. So if we want to run both, we will set it to `['enterprise', 'community']`.

dothebart: `testing.js` is the framework which runs all the regular integration tests. Its using arangosh. 
The selenium tests are created under the python written RTA framework, which attempts to test installations and deployments - we use it during the release testing.
Regarding ARM, I don't think that running the tests on ARM yields different results, unless maybe we run them with instrumented binaries (which then tend to be slooouuwww)

KVS85 (approved):

PR #20770, opened 2024-03-21
improve debug facilities for shard synchronization, transactions and locks

jsteemann:
### Scope & Purpose

* Dump state of r/w lock in case we cannot acquire the lock in time. The lock state can be used to reason about pending readers/writers and thus can be used for internal debugging.
* Improve debug log messages for shard synchronization.
* Improve certain error messages when there are failures when looking up transactions in the transaction manager.
* Log a message into arangod log for all requests that spent >= 30 seconds on the scheduler queue.
* Add `details` flag for HTTP REST API GET `/_api/transaction`, which will make it fetch the transaction manager internals from all servers in the cluster (not just coordinators) and expose a lot more details about the transactions. The `details` flag is intended to be used for debugging only. It is not part of the officially supported API.
* Added internal debug API for dumping AQL query registry contents. The API is GET `/_api/query/registry` and can only be executed by superusers. This API is intended to be used for debugging only. It is not an officially supported API.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestHandler/RestQueryHandler.cpp
```diff
@@ -80,6 +92,105 @@ RestStatus RestQueryHandler::execute() {
...
+      }
+
+      auto f = network::sendRequestRetry(
+          pool, "server:" + server, fuerte::RestVerb::Get,
+          "/_api/query/registry", body, options, std::move(headers));
+      futures.emplace_back(std::move(f));
+    }
+
+    if (!futures.empty()) {
+      auto responses = futures::collectAll(futures).get();
```
maierlars, line 164: We actively try to get rid of blocking code and it was never easier to write non-blocking code using coroutines. 
I will provide a PR that makes writing async rest handler more easily.

maierlars (approved):

hkernbach (approved):

hkernbach:
@jsteemann one thing I forgot during the review: The changelog entry is missing.

PR #20772, opened 2024-03-21
[EF-60] Remove CollectionID from SmartGraph Schema

mchacki:
### Scope & Purpose

*EE Only: For SmartGraphs the Schema Definition document contained a CollectionID for one of the SmartGraphCollections. When you dump the graph and restore it into a different Database the CollectionID may be regenerated, but the Restore process did not reflect the change into the Schema definition. In the restored version the Schema could not be changed because of the missmatch. This PR removes the CollectionID from the Schema, and ignores it if it exists. This way we can restore dumps from old versions and allow to Change the SmartGraph schema again as intended.*

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [x] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [x] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [x] Enterprise PR: https://github.com/arangodb/enterprise/pull/1463
- [x] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/EF-60, https://arangodb.atlassian.net/browse/MDS-1190
- [ ] Design document:


CHANGELOG
```diff
@@ -1,6 +1,14 @@
 devel
 -----
 
+* EE-Only bug-fix MDS-1190: When we use arangodump and arangorestore while
+  using a SmartGraph the restore process may generate new internal CollectionIds.
+  This generated Ids are not always updated in the also restored GraphSchema.
+  In the situation the restored graph still works as expected, all data is there
+  and can be queried and modified. Only the GraphSchema on the restored version
+  could not be changed anymore. We erased the dependency on the CollectionId
+  in the GraphSchema, now the Schema can again be modified.
```
```suggestion
* EE-Only bug-fix MDS-1190: When using arangorestore to restore a SmartGraph,
  the restore process may generate new internal collection ids.
  These generated ids were not always updated in the restored SmartGraph schema.
  If the ids diverge, the restored SmartGraph still works as expected, all data is there
  and can be queried and modified. Only the schema of the restored SmartGraph
  could not be changed anymore, e.g. no collections could be added to or removed
  from the graph. 
  The SmartGraph schema handling now does not consider collection ids anymore,
  so that schemas can again be modified, even those of existing SmartGraphs that
  were originally affected by the issue.
```

tests/js/client/dump/dump-test.inc
```diff
@@ -1563,6 +1563,38 @@
         assertEqual(graph.orphanCollections.length, 0);
       },
 
+      testSmartGraph3_11_compat: function () {
+      return;
```
jsteemann, line 1567: This line looks like an unintentional change.
```suggestion
```

tests/js/client/dump/dump-test.inc
```diff
@@ -1563,6 +1563,38 @@
...
+        if (!isEnterprise) {
+          return;
+        }
+        const smartGraphName = "UnitTestDumpSmartGraphCompat";
+        const graphsColl = getCollection('_graphs');
+        assertTrue(!!graphsColl.exists(smartGraphName));
+
+        const graphDoc = graphsColl.document(smartGraphName);
+        // This is the requirement to test if the compat cid has effect
+        assertTrue(graphDoc.hasOwnProperty('initialCid'));
```
jsteemann, line 1577: IIRC, the enterprise part of the PR will remove this attribute. Is this assertion still valid now?
```suggestion
        assertTrue(graphDoc.hasOwnProperty('initialCid'));
```

tests/js/client/dump/dump-test.inc
```diff
@@ -1563,6 +1563,38 @@
...
+          newVertexName
+        );
+        // New collections do not yet exist, that would be forbidden.
+        assertTrue(db._collections().indexOf(newRelationName) === -1);
+        assertTrue(db._collections().indexOf(newVertexName) === -1);
+        const graph = gm._graph(smartGraphName);
+        graph._extendEdgeDefinitions(newRelation);
+        // New collections are added
+        assertTrue(db._collections().indexOf(newRelationName) !== -1);
+        assertTrue(db._collections().indexOf(newVertexName) !== -1);
```
jsteemann, line 1592: I thought that `db._collections()` returns an array of collection objects, not an array of collection names.
So I think these assertions will always succeed, but they do something different than expected.
I think it should rather be `assertFalse(db._collections().map((c) => c.name()).includes(...));`.
Same for the following assertions.

tests/js/client/dump/dump-cluster.js
```diff
@@ -88,6 +74,19 @@ jsunity.run(function dump_cluster_testsuite() {
...
       "testHybridDisjointSmartGraphSingleServer"
+    ];
+  }
+  deriveTestSuite(
+    baseTests(),
+    suite,
+    "_cluster",
+    [ // <-- Blacklisted Tests
+      // Magic Hint: Those tests are tests which you need to additional blacklist in case they are not supported
+      // in that specific environment. Those blacklist is separately and manually maintained per test-suite.
```
jsteemann, line 85:
```suggestion
    [ // <-- Deactivated Tests
      // Magic Hint: Those tests are not supported in the specific environment. 
      // The exclusion list is separately and manually maintained per test-suite.
```

tests/js/client/dump/dump-test.inc
mchacki: 😨 oh noe this is an oversight on my end, while I was trying to figure out when the test is run.
Very good catch this disabled the test!!!

tests/js/client/dump/dump-test.inc
mchacki: Yes in this case it absolutely is.
Because we are simulating an old graph here.
The corresponding setup could should inject the initialCid value like an old version would. (So after restore of old data, or rolling upgrade, we don't want to break either)
And this test should then validate that we do not have negative side effects by having this value set.

I updated the comment above

jsteemann (approved):
LGTM

PR #20791, opened 2024-03-27
Fix potential sleeping barber in async prefetching

mpoeter:
### Scope & Purpose

Previously we disabled async prefetching for any node that had a (indirect) dependency on a RemoteNode, because of the following problem: If an async prefetch task returns WAITING (e.g. from a RemoteNode), the wakeup could arrive at a point were the RestHandler has already been removed and is thus swallowed. When the next RestHandler makes the next call, we simply return the result from the prefetch task. But if that is WAITING, we now wait for a wakeup that will never come.

To fix that, the next call will now check if the prefetch task has returned WAITING, and in that case will immediately make the next call to the upstream dependency. If the wakeup has already occurred, we will get the new data, otherwise we will get another WAITING, but then we know that the wakeup is still going to happen.

- [x] :pizza: New feature

### Checklist

- [x] Tests - this is covered by existing tests

#### Related Information

- [x] Jira ticket: https://arangodb.atlassian.net/browse/EF-66


arangod/Aql/SharedQueryState.cpp
```diff
@@ -72,32 +72,33 @@ void SharedQueryState::waitForAsyncWakeup() {
...
-  _wakeupCb = nullptr;
-  _numWakeups = 0;
-  _cbVersion++;
-}
-
-void SharedQueryState::resetNumWakeups() {
-  std::lock_guard<std::mutex> guard(_mutex);
-  _numWakeups = 0;
-  _cbVersion++;
+  setWakeupHandler(nullptr);
```
jsteemann, line 101: This calls `setWakeupHandler(nullptr);` while having acquired the mutex.
But `setWakeupHandler()` also acquires the mutex. I guess this will deadlock.

mpoeter: Argl... yes, of course! 🤦‍♂️ That was one of the last cleanup changes I did.

jsteemann (approved):
LGTM

maierlars (approved):

PR #20803, opened 2024-04-02
Trigger callbacks waiting on raft indexes in the ClusterInfo timely

goedderz:
### Scope & Purpose

ClusterInfo allows to register callbacks that are executed when it reaches a certain recency; either based on Plan- or Current-Version, or on a raft index. Currently, the ClusterInfo is only updated (in loadPlan and loadCurrent) when the respective *Version changes. This means if a callback is registered with a raft index more recent than the latest *Version, it won't get executed until something unrelated changes and updates the particular *Version again.

This PR makes changes mainly to the SyncerThread to get updates of raft indexes to the ClusterInfo. Additional changes in the AgencyCache were necessary to continue to avoid the Scheduler, which is the main reason for having the SyncerThread in the first place, so load on the system won't stall updates to the ClusterInfo. loadPlan and loadCurrent also get slight changes One, they now return the raft index they've updated to, for the SyncerThread to work with. And two, avoid any unneccessary work before the early exit, as they'll be called more often now.

- [X] :hankey: Bugfix


arangod/Cluster/AgencyCache.h
```diff
@@ -211,7 +214,8 @@ class AgencyCache final : public ServerThread<ArangodServer> {
 
   /// @brief Waiting room for indexes during office hours
   mutable std::mutex _waitLock;
-  std::multimap<consensus::index_t, futures::Promise<arangodb::Result>>
+  std::multimap<consensus::index_t,
+                std::pair<futures::Promise<arangodb::Result>, Executor>>
```
maierlars, line 218: Maybe define a struct instead of a pair? You can still use structured bindings with structs.

maierlars (approved):

arangod/Cluster/AgencyCache.h
goedderz: Good idea, done.

PR #20804, opened 2024-04-02
Remove the detach thread code from transaction manager.

maierlars:
### Scope & Purpose

Remove the detach thread thingy from the `leaderManagedTrx` function. Instead use a delayed future to make the current execution fiber sleep for some time. Previously we would block the thread and then eventually detach it.


jsteemann (approved):
LGTM

PR #20808, opened 2024-04-03
Fixed size ThreadPool scheduler

maierlars:
### Scope & Purpose

This PR implements a new `ThreadPoolScheduler` with a fixed number of threads, and a number of different thread pools. The  `ThreadPoolScheduler` creates a separate thread pool for each priority lane.

We have the following thread pools:
  - SimpleThreadPool - this is a very simple thread pool with a std::queue, mutex + condition variable
  - LockFreeThreadPool - this one is similar to the SimpleThreadPool but replaces the std::queue with a boost::lockfree::queue
  - WorkStealingThreadPool - this is a first version of a work stealing pool where instead of a globally shared queue, each thread has a local queue

This PR also implements a number of synthetic benchmarks as tests (by default these are disabled so that they do not run durng regular builds). These benchmarks showed, that the WorkStealingThreadPool scales significantly better than the other two thread pools, and also much better then the current SupervisedScheduler.


arangod/Scheduler/SimpleThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+  std::unique_lock guard(_mutex);
+  _stop.request_stop();
+  _cv.notify_all();
+}
+
+void ThreadPool::push(std::unique_ptr<WorkItem>&& task) noexcept {
+  std::unique_lock guard(_mutex);
+  _tasks.emplace_back(std::move(task));
+  _cv.notify_one();
+  statistics.queued += 1;
```
goedderz, line 67: Wouldn't it be better to release the mutex before notifying?

arangod/Scheduler/SimpleThreadPool.cpp
```diff
@@ -0,0 +1,82 @@
...
+#endif
+
+    return thread;
+  });
+}
+
+ThreadPool::~ThreadPool() {
+  std::unique_lock guard(_mutex);
+  _stop.request_stop();
+  _cv.notify_all();
```
goedderz, line 60: Wouldn't it be better to release the mutex before notifying?

goedderz (commented):
Soo much simpler 😍

arangod/Scheduler/LockfreeThreadPool.cpp
maierlars, line 1: do we want to keep the other implementations? Like Lockfree and Simple?

maierlars (commented):
LGTM

arangod/Scheduler/LockfreeThreadPool.cpp
mpoeter: I would keep them around for now (at least the lockfree one). Theoretically we could even make them accessible via the command line parameter.

arangod/Scheduler/LockfreeThreadPool.h
```diff
@@ -0,0 +1,78 @@
...
+  void push(F&& fn) noexcept {
+    struct LambdaTask : WorkItem {
+      F _func;
+      explicit LambdaTask(F&& func) : _func(std::forward<F>(func)) {}
+
+      static_assert(std::is_nothrow_invocable_r_v<void, F>);
+      void invoke() noexcept override { std::forward<F>(_func)(); }
+    };
+
+    push(std::make_unique<LambdaTask>(std::forward<F>(fn)));
```
jsteemann, line 54: The `make_unique` here can throw an exception. The method is marked `noexcept`, so when in an OOM scenario this will terminate the entire server process. If this is intentional, fine. But we could also make this method not `noexcept` and make it throw a regular exception, so that callers can handle it.

arangod/Scheduler/LockfreeThreadPool.h
```diff
@@ -0,0 +1,78 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+#pragma once
+#include <memory>
```
jsteemann, line 22: nit: I suggest also including the atomic header here, as `std::atomic<T>` is used here.

arangod/Scheduler/SimpleThreadPool.cpp
```diff
@@ -0,0 +1,98 @@
...
+    std::unique_lock guard(_mutex);
+    _stop.request_stop();
+  }
+  _cv.notify_all();
+}
+
+void SimpleThreadPool::push(std::unique_ptr<WorkItem>&& task) noexcept {
+  {
+    std::unique_lock guard(_mutex);
+    _tasks.emplace_back(std::move(task));
```
jsteemann, line 79: this can throw an exception, but the method is marked `noexcept`. If this is intentional, we should at least leave a code comment that we are willing to terminate the entire server process here if the `emplace_back` goes wrong.

arangod/Scheduler/LockfreeThreadPool.cpp
```diff
@@ -0,0 +1,133 @@
...
+        if (item.get() == &stopItem) {
+          std::ignore = item.release();
+          return;
+        }
+
+        try {
+          item->invoke();
+        } catch (...) {
+          LOG_TOPIC("d5fb4", WARN, Logger::FIXME)
+              << "Scheduler just swallowed an exception.";
```
jsteemann, line 63: Knowing that an arbitrary exception has happened by not knowing what kind of exception is not as helpful as having exception details here.
IMHO it would be better here to catch all exceptions of type `std::exception` and log a proper exception message here for exceptions that have a message attached.

arangod/Scheduler/LockfreeThreadPool.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (item != &stopItem) {
+      delete item;
+    }
+  }
+}
+
+void LockfreeThreadPool::push(std::unique_ptr<WorkItem>&& task) noexcept {
+  _queue.push(task.release());
+  statistics.queued.fetch_add(1);
+  auto old = statistics.inQueue.fetch_add(1);
```
jsteemann, line 101: Are we intentionally using the strictest memory order here, or can it be relaxed?

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+void ThreadPoolScheduler::trackCreateHandlerTask() noexcept {}
+void ThreadPoolScheduler::trackBeginOngoingLowPriorityTask() noexcept {
+  _metrics->_ongoingLowPriorityGauge += 1;
+}
+void ThreadPoolScheduler::trackEndOngoingLowPriorityTask() noexcept {
+  _metrics->_ongoingLowPriorityGauge -= 1;
+}
+void ThreadPoolScheduler::trackQueueTimeViolation() {}
+void ThreadPoolScheduler::trackQueueItemSize(std::int64_t int64) noexcept {}
```
jsteemann, line 50: nit
```suggestion
void ThreadPoolScheduler::trackQueueItemSize(std::int64_t) noexcept {}
```

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,123 @@
...
+      .jobsQueued = &_metrics->_metricsJobsSubmittedTotal,
+      .jobsDequeued = &_metrics->_metricsJobsDequeuedTotal,
+  };
+
+  _threadPools.reserve(4);
+
+  poolMetrics.queueLength = _metrics->_metricsQueueLengths[0];
+  poolMetrics.dequeueTimes = _metrics->_metricsDequeueTimes[0];
+  _threadPools.emplace_back(std::make_unique<WorkStealingThreadPool>(
+      "SchedMaintenance", std::max(std::ceil(maxThreads * 0.1), 2.),
```
jsteemann, line 101: nit: the `std::max` here returns (and in the following occasions) returns a double value, but the constructor argument for `WorkStealingThreadPool` is a `size_t` `threadCount`. While this works, an explicit cast to size_t may be more readable here.

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  poolMetrics.queueLength = _metrics->_metricsQueueLengths[2];
+  poolMetrics.dequeueTimes = _metrics->_metricsDequeueTimes[2];
+  _threadPools.emplace_back(std::make_unique<WorkStealingThreadPool>(
+      "SchedMedium", std::max(std::ceil(maxThreads * 0.4), 8.), poolMetrics));
+
+  poolMetrics.queueLength = _metrics->_metricsQueueLengths[3];
+  poolMetrics.dequeueTimes = _metrics->_metricsDequeueTimes[3];
+  _threadPools.emplace_back(std::make_unique<WorkStealingThreadPool>(
+      "SchedLow", std::max(std::ceil(maxThreads * 0.6), 16.), poolMetrics));
```
jsteemann, line 117: The multipliers of `maxThreads` sum up to `0.6 + 0.4 + 0.4 + 0.1 == 1.5`. I would have expected them to sum up to just `1.0` here. That we use 1.5 times the number of configured threads may be intentional, but the rationale for it is not immediately clear. I suggest adding code comments here to explain the rationale.

arangod/Scheduler/ThreadPoolScheduler.h
```diff
@@ -0,0 +1,60 @@
...
+  void shutdown() override;
+
+ protected:
+  bool queueItem(RequestLane lane, std::unique_ptr<WorkItemBase> item,
+                 bool bounded) override;
+  bool isStopping() override { return _stopping; }
+
+ private:
+  std::shared_ptr<SchedulerMetrics> _metrics;
+  std::atomic<bool> _stopping;
```
jsteemann, line 55: It looks like `_stopping` is not initialized to `false` anywhere. We should definitely initialize it!

arangod/Scheduler/WorkStealingThreadPool.h
```diff
@@ -0,0 +1,84 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+#pragma once
+#include <cstddef>
```
jsteemann, line 22: nit: I suggest to also include `atomic` and `type_traits` here, as they are used by the header file.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+#include "WorkStealingThreadPool.h"
+
+#include <algorithm>
+#include <atomic>
+#include <chrono>
+#include <cstddef>
+#include <memory>
```
jsteemann, line 27: nit: this should also include `deque`, as it is used by this file.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+void WorkStealingThreadPool::ThreadState::pushBack(
+    std::unique_ptr<WorkItem>&& work) noexcept {
+  TRI_ASSERT(work != nullptr);
+  TRI_ASSERT(work->next == nullptr);
+  auto p = work.release();
+  p->next = pushQueue.load(std::memory_order_acquire);
+  bool hasWork = p->next != nullptr;
+  // the compare_exchange implicitly reloads the next pointer
+  while (
+      !pushQueue.compare_exchange_weak(p->next, p, std::memory_order_seq_cst)) {
```
jsteemann, line 144: nit: it seems that this is pushing to the _front_ of the queue, although the method is named `pushBack`. I found that slightly confusing, but it does not need to be changed.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+    -> std::unique_ptr<WorkItem> {
+  TRI_ASSERT(item != nullptr);
+  while (item->next != nullptr) {
+    auto next = item->next;
+    item->next = nullptr;
+    _queue.push_front(std::unique_ptr<WorkItem>(item));
+    item = next;
+    TRI_ASSERT(item != nullptr);
+  }
+  queueSize.store(_queue.size(), std::memory_order_relaxed);
```
jsteemann, line 289: nit: we could save this store operation in case we did not push anything into `_queue` in this method. this would save an atomic store, but not sure if it's worth optimizing it.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+    auto d =
+        std::chrono::duration_cast<std::chrono::microseconds, double>(queueTime)
+            .count();
+    pool._metrics.dequeueTimes->count(d);
+  }
+  pool.statistics.dequeued.fetch_add(1, std::memory_order_relaxed);
+  try {
+    work.invoke();
+  } catch (...) {
+    LOG_TOPIC("d5fb3", WARN, Logger::FIXME)
```
jsteemann, line 308: I would prefer catching all `std::exception`s here and then logging their proper exception message. Otherwise it is hard to figure out what exactly went wrong.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+              TRI_ASSERT(other->_queue.size() > 0);
+              work = std::move(other->_queue.back());
+              TRI_ASSERT(work != nullptr);
+              other->_queue.pop_back();
+            }
+            queueSize.store(_queue.size(), std::memory_order_relaxed);
+          }
+          other->queueSize.store(other->_queue.size(),
+                                 std::memory_order_relaxed);
+
```
jsteemann, line 365: nit: `guard` could be prematurely unlocked here, so that it does not need to wait for the following atomic update to complete.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+      _threads(),
+      _threadStates(threadCount),
+      _latch(threadCount) {
+  auto coprimes = calculateCoprimes(threadCount);
+  std::size_t coprimeIdx = 0;
+  for (std::size_t i = 0; i < threadCount; ++i) {
+    _threads.emplace_back(std::jthread([this, i, &coprimes, coprimeIdx]() {
+      // we intentionally create the ThreadStates _inside_ the thread
+      _threadStates[i] =
+          std::make_unique<ThreadState>(i, coprimes[coprimeIdx], *this);
```
jsteemann, line 400: nit: the separate `coprimeIdx` variable is not needed and could be substituted by just `i % coprimes.size()`.

arangod/Scheduler/WorkStealingThreadPool.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+void WorkStealingThreadPool::push(std::unique_ptr<WorkItem>&& task) noexcept {
+  if (auto* ts = ThreadState::currentThread;
+      ts != nullptr && ts->belongsToPool(*this)) {
+    ts->pushFront(std::move(task));
+  } else {
+    auto h = hint.load(std::memory_order_relaxed);
+    if (h != NoHint) {
+      auto prevHint =
+          _threadStates[h]->prevHint.load(std::memory_order_relaxed);
```
jsteemann, line 439: Can there be a situation in which `hint` is set to anything else but `NoHint`, but `prevHint` still is `NoHint`?
In this case this would cause an out-of-bounds read, but I am not sure if it is guaranteed that `prevHint` always has a valid value when we get here.

tests/Scheduler/ThreadPool.cpp
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+
+  ASSERT_EQ(counter, 100);
+}
+
+TEST(ThreadPoolTest, stop_when_sleeping) {
+  {
+    SimpleThreadPool pool{"test-sched", 3};
+    std::this_thread::sleep_for(std::chrono::seconds{3});
```
jsteemann, line 73: Also unclear what exactly is tested here.

tests/Scheduler/ThreadPool.cpp
```diff
@@ -0,0 +1,350 @@
...
+    durationMs = std::chrono::duration_cast<std::chrono::milliseconds>(
+                     std::chrono::steady_clock::now() - start)
+                     .count();
+
+    // wait a bit so we don't run into an assertion in the SupervisedScheduler
+    // that we tried to queue an item after the SchedulerFeature was stopped
+    std::this_thread::sleep_for(std::chrono::milliseconds{100});
+  }
+  auto numOps = counter.load();
+  std::cout << std::setw(2) << numThreads << " threads: " << std::setw(5)
```
jsteemann, line 211: note: this test only prints out timing information, but does not contain any test assertions/expectations.
It is rather a performance test than a functional test. This is fine, but we may want to create separate benchmarks for it rather than running it as part of the functional tests. To be clarified.

arangod/Scheduler/LockfreeThreadPool.cpp
mpoeter: The LockfreeThreadPool turned out to scale less well so I did not spend time on such low level optimizations - it will probably be removed at some point.

arangod/Scheduler/WorkStealingThreadPool.cpp
mpoeter: The `pushQueue` is basically a stack, so you are right that we are pushing to the front. However, we are not consuming items directly from the `pushQueue`. Instead, any thread that grabs these items pushes them onto it's own queue using `push_front`, effectively reversing the list again.

arangod/Scheduler/WorkStealingThreadPool.cpp
mpoeter: `pushMany` should not be called with a nullptr (note the assertion), so we should always add at least one item to the queue.

arangod/Scheduler/WorkStealingThreadPool.cpp
mpoeter: > Can there be a situation in which hint is set to anything else but `NoHint`, but `prevHint` still is `NoHint`?
Sure, but I don't see how this can cause any issues? `hint` and `prevHint` must always either contain a valid index or `NoHint`. If `hint != NoHint` but `_threadStates[h]->prevHint == NoHint` we simply update `hint` to `NoHint`.

arangod/Scheduler/ThreadPoolScheduler.cpp
maierlars: This has been done, because there is no longer thread sharing between the different priorities.

arangod/Scheduler/WorkStealingThreadPool.cpp
jsteemann: I agree. But even if `pushMany` is called with a non-nullptr `item`, it is possible and valid that `item->next` is a nullptr, and then the queue size will not be modified. Thus we also don't need to perform the atomic store on the queue size variable. But this is an optimization and not necessary for correctness.

arangod/Scheduler/WorkStealingThreadPool.cpp
jsteemann: Ah true, please ignore my comment then.

arangod/Scheduler/WorkStealingThreadPool.cpp
jsteemann: Ok, fine for me.

arangod/Scheduler/ThreadPoolScheduler.cpp
mpoeter: I will add a comment. FWIW - in the future I want to get rid of the separate thread pools and have a WorkStealingThreadPool where we have separate partitions for the threads to steal from.

arangod/Scheduler/WorkStealingThreadPool.cpp
mpoeter: Ah, right, because we keep the last item to be work on immediately. However, I don't think this is worth the optimization.

jsteemann: Fine for me too.

js/client/modules/@arangodb/testutils/san-file-handler.js
```diff
@@ -94,8 +94,15 @@ class sanHandler {
       print("processing ", value);
       const { upstream, local } = value;
       let fn = `${local}.${this.binaryname}.${pid}`;
-      if (this.extremeVerbosity) {
+      if (true || this.extremeVerbosity) {
```
jsteemann, line 97: This change looks unintentional.

arangod/Scheduler/ThreadPoolScheduler.h
```diff
@@ -0,0 +1,62 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#include "Scheduler.h"
+#include "SimpleThreadPool.h"
+#include "LockfreeThreadPool.h"
```
jsteemann, line 24: Are these two actually used by this header file?
```suggestion
```

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+void ThreadPoolScheduler::trackCreateHandlerTask() noexcept {}
+void ThreadPoolScheduler::trackBeginOngoingLowPriorityTask() noexcept {
+  _metrics->_ongoingLowPriorityGauge += 1;
+}
+void ThreadPoolScheduler::trackEndOngoingLowPriorityTask() noexcept {
+  _metrics->_ongoingLowPriorityGauge -= 1;
+}
+void ThreadPoolScheduler::trackQueueTimeViolation() {}
+void ThreadPoolScheduler::trackQueueItemSize(std::int64_t) noexcept {}
```
jsteemann, line 50: Just wondering why these two methods are not implemented here.

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,132 @@
...
+  b.add("blocked", VPackValue(0));            // obsolete
+  b.add("queued", VPackValue(0));             // scheduler queue length
+  b.add("in-progress",
+        VPackValue(0));                 // number of working (non-idle) threads
+  b.add("direct-exec", VPackValue(0));  // obsolete
+}
+Scheduler::QueueStatistics ThreadPoolScheduler::queueStatistics() const {
+  return QueueStatistics();
+}
+void ThreadPoolScheduler::trackCreateHandlerTask() noexcept {}
```
jsteemann, line 42: Should this be implemented here, or somewhere else? It is probably ok as it is now, but a code comment about why the implementation is empty would help me here.

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,132 @@
...
+      _threadPools[int(RequestPriority::LOW)]->statistics.queued.load();
+  auto remaining = dequeued > queued ? 0 : queued - dequeued;
+  return std::make_pair(
+      _metrics->_ongoingLowPriorityGauge.load(std::memory_order_relaxed),
+      remaining);
+}
+
+double ThreadPoolScheduler::approximateQueueFillGrade() const { return 0; }
+
+double ThreadPoolScheduler::unavailabilityQueueFillGrade() const { return 0; }
```
jsteemann, line 75: Code comments here would help me understand why both functions return 0.

arangod/Scheduler/ThreadPoolScheduler.cpp
mpoeter: Fixed

arangod/Scheduler/ThreadPoolScheduler.cpp
mpoeter: Fixed

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,138 @@
...
+      _threadPools[int(RequestPriority::LOW)]->statistics.dequeued.load();
+  auto queued =
+      _threadPools[int(RequestPriority::LOW)]->statistics.queued.load();
+  auto remaining = dequeued > queued ? 0 : queued - dequeued;
+  return std::make_pair(
+      _metrics->_ongoingLowPriorityGauge.load(std::memory_order_relaxed),
+      remaining);
+}
+
+// queue fill grad tracking is currently not implemented for this scheduler
```
jsteemann, line 79:
```suggestion
// queue fill grade tracking is currently not implemented for this scheduler
```

arangod/Scheduler/ThreadPoolScheduler.cpp
```diff
@@ -0,0 +1,138 @@
...
+double ThreadPoolScheduler::approximateQueueFillGrade() const { return 0; }
+double ThreadPoolScheduler::unavailabilityQueueFillGrade() const { return 0; }
```
jsteemann, line 81: Note that the queue fill grade may be checked by a load balancer periodically to determine if the instance is still in good shape.
Consider this calling code, which determines the instance's availability for further requests, from a load balancer's perspective:
```
        // if the scheduler's queue is more than x% full, render
        // the server unavailable
        double unavailabilityFillGrade =
            scheduler->unavailabilityQueueFillGrade();
        if (unavailabilityFillGrade > 0.0) {
          double fillGrade = scheduler->approximateQueueFillGrade();
          if (fillGrade >= unavailabilityFillGrade) {
            // oops, queue is relatively full
            available = false;
          }   
        }   
```
While this code works if the unavailability fill grade is 0, it will lead to more requests being sent to the instance now even if the queue is relatively full.
If this is an intentional change, fine. If not, is it much work to implement the fill grade methods in the new scheduler, or is there a reason why we shouldn't do this?

jsteemann (approved):
LGTM

arangod/Scheduler/ThreadPoolScheduler.cpp
mpoeter: These are functions defined in the Scheduler interface, but currently only make sense for the SupervisedScheduler (leaky abstraction). We will have to discuss if/how to add this to the new scheduler.

PR #20811, opened 2024-04-04
Async trx commit/abort

maierlars:
### Scope & Purpose
`commitManagedTrx` and `abortManagedTrx` now return a future. Instead of blocking and spinning in a busy loop, they yield the current thread and wait. However, this waiting is only done, if the transaction is in use by a different request, while also being committed/aborted.

### DIFF
--- a/arangod/Scheduler/Scheduler.h
+++ b/arangod/Scheduler/Scheduler.h
@@ -242,6 +242,22 @@ class Scheduler {
     });
   }
 
+  // Yield the current thread
+  futures::Future<futures::Unit> yield(
+      RequestLane lane = RequestLane::CONTINUATION) {
+    struct awaitable {
+      bool await_ready() { return false; }
+      void await_suspend(std::coroutine_handle<> coro) {
+        sched->queue(lane, [coro] { coro.resume(); });
+      }
+      void await_resume() {}
+      Scheduler* sched;
+      RequestLane lane;
+    };
+
+    co_await awaitable{this, lane};
+  }
+
   // ---------------------------------------------------------------------------
   // CronThread and delayed tasks
   // ---------------------------------------------------------------------------
--- a/arangod/Transaction/Manager.cpp
+++ b/arangod/Transaction/Manager.cpp
@@ -1036,9 +1036,15 @@ futures::Future<Result> Manager::statusChangeWithTimeout(
     TransactionId tid, std::string const& database,
     transaction::Status status) {
   double startTime = 0.0;
-  constexpr double maxWaitTime = 3.0;
+  constexpr double maxWaitTime = 9.0;
   Result res;
   while (true) {
+    // does not acquire the lock
+    CONDITIONAL_READ_LOCKER(guard, _hotbackupCommitLock, false);
+    if (status == Status::COMMITTED) {
+      // only for commit acquire the lock
+      guard.lock();
+    }
     res = updateTransaction(tid, status, false, database);
     if (res.ok() || !res.is(TRI_ERROR_LOCKED)) {
       break;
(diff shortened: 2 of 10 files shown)


arangod/Transaction/Manager.cpp
```diff
@@ -1018,19 +1026,21 @@ Result Manager::statusChangeWithTimeout(TransactionId tid,
...
+  co_return res;
 }
 
-Result Manager::commitManagedTrx(TransactionId tid,
-                                 std::string const& database) {
+futures::Future<Result> Manager::commitManagedTrx(TransactionId tid,
+                                                  std::string const& database) {
   READ_LOCKER(guard, _hotbackupCommitLock);
-  return statusChangeWithTimeout(tid, database, transaction::Status::COMMITTED);
+  co_return co_await statusChangeWithTimeout(tid, database,
```
jsteemann, line 1038: Can the `_hotbackupCommitLock` now be held while the execution is suspendend, and then be woken up in a potentially different thread?

arangod/Transaction/Manager.cpp
```diff
@@ -1018,19 +1026,21 @@ Result Manager::statusChangeWithTimeout(TransactionId tid,
       // timeout
       break;
     }
-    std::this_thread::yield();
+    co_await sleepUsingSchedulerIfAvailable("trx-status-change-backup",
+                                            std::chrono::milliseconds{10});
```
jsteemann, line 1030: Previously we did a `this_thread::yield()` here and then continued in the same thread if the status cannot be changed.
Now we are sleeping 10ms here, and we are using the scheduler for continuations. That may take substantially longer than in the old implementation if there is a lot of activity in the scheduler.
Thus I think we should also increase the `maxWaitTime` (line 1015) from 3s to some substantially higher value here.

arangod/Scheduler/Scheduler.h
```diff
@@ -242,6 +242,22 @@ class Scheduler {
     });
   }
 
+  // Yield the current thread
+  futures::Future<futures::Unit> yield(
```
jsteemann, line 246: No objections against this method, but is it being used in this PR?

arangod/Transaction/Manager.cpp
maierlars:
```
basics::ReadWriteLock _hotbackupCommitLock;
```
afaik this is our own lock and is it capable of being released in a different thread. I will double check that.

arangod/Scheduler/Scheduler.h
maierlars: It was initially used in `statusChangeWithTimeout` instead of the `std::this_thread::yield()`. But then I figured that the loop in that function is essentially a busy loop and that it might not be the best idea to hammer on the scheduler. That is why I introduce the delay. 

In particular this delay only happens if someone aborts or commits a transaction, while there is also an ongoing request _using_ that transaction.

arangod/Transaction/Manager.cpp
jsteemann: Yes, it is our own ReadWriteLock implementation, which supports locking/unlocking being performed by different threads.
The question was also more like: do we actually want to hold the hot backup lock here even if we cannot change the transaction's status and need to retry?
In this case, we could give up the hot backup lock temporarily and reacquire it later, when the operation resumes. Otherwise we would keep holding the hot backup lock for no good reason. 
Obviously the previous implementation already had the same issue, so this change would rather be a orthogonal and doesn't necessarily need to be done in this PR, if at all.

arangod/Transaction/Manager.cpp
```diff
@@ -784,6 +784,25 @@ futures::Future<Result> Manager::ensureManagedTrx(
   co_return res;
 }
 
+namespace {
+futures::Future<futures::Unit> sleepUsingSchedulerIfAvailable(
+    const char* reason, std::chrono::milliseconds duration) {
```
jsteemann, line 789: nit: we can as well use std::string_view here. we are only calling it with C string literals as of now, which are null-terminated.
```suggestion
    std::string_view reason, std::chrono::milliseconds duration) {
```

jsteemann (approved):
LGTM

PR #20813, opened 2024-04-04
Use std::source_location, and make paths relative to the content root

goedderz:
### Scope & Purpose

Enterprise companion PR: https://github.com/arangodb/enterprise/pull/1470

1) Make use of std::source_location in SourceLocation, instead of __FILE__ and __LINE__, to make it more usable
2) Strip the prefix of the content root from the file name.

A little explanation for 2): SourceLocation is generally used to show where an exception originated from. There's usually no use to have the absolute path of that file on the machine that built the binary. So SourceLocation now makes the path relative to the content root; e.g., instead of `/root/project/arangodb/lib/Basics/SourceLocation.h`, the file_name will contain `lib/Basics/SourceLocation.h`.

- [X] :pizza: New feature

### Checklist

- [X] Tests
  - [X] C++ **Unit tests**

### DIFF
--- a/lib/Basics/SourceLocation.h
+++ b/lib/Basics/SourceLocation.h
@@ -24,33 +24,82 @@
 #pragma once
 
 #include <iosfwd>
+#include <source_location>
+#include <string_view>
 
 namespace arangodb::basics {
+consteval auto stripPrefix(const char* pathStr) -> const char*;
 
-// Poor-man's replacement for std::source_location, until we get C++20.
+// This is mostly a custom type around std::source_location with the same
+// interface; However, it changes the filename's path, so it's relative to the
+// git root directory.
 struct SourceLocation {
  private:
-  char const* _fileName;
-  int _line;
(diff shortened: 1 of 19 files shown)


lib/Basics/SourceLocation.h
```diff
@@ -23,34 +23,84 @@
...
+  // We can allow this if we need to, but this way it prevents accidental usage
+  // of uninitialized SourceLocations.
   SourceLocation() = delete;
-  // cppcheck-suppress unknownMacro
-  constexpr SourceLocation(decltype(_fileName) file,
-                           decltype(_line) line) noexcept
-      : _fileName(file), _line(line) {}
+
+  consteval explicit SourceLocation(
+      std::source_location const location) noexcept
```
jsteemann, line 56: nit: as we are passing by value here, I don't think that the `const` actually matters.
```suggestion
      std::source_location location) noexcept
```

PR #20817, opened 2024-04-05
GetLeaderForShard Fix

maierlars:
### Scope & Purpose
Stop waiting when the collection/shard is dropped and return the _datasource not found_ error.


jsteemann (approved):
LGTM

PR #20827, opened 2024-04-09
QA-531: start implementing single-shard dump/restore

dothebart:
Backport of https://github.com/arangodb/arangodb/pull/20838 
Enterprise PR: https://github.com/arangodb/enterprise/pull/1494

### DIFF
--- a/tests/js/server/dump/dump-setup-common.inc
+++ b/tests/js/server/dump/dump-setup-common.inc
@@ -44,6 +44,12 @@ if (isEnterprise) {
   egm = require("@arangodb/enterprise-graph");
 }
 
+function getNoShards(defaultNo) {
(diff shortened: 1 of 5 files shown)


naushniki (approved):

tests/js/server/dump/dump-setup-common.inc
```diff
@@ -44,6 +44,12 @@ if (isEnterprise) {
   egm = require("@arangodb/enterprise-graph");
 }
 
+function getNoShards(defaultNo) {
```
jsteemann, line 47: I am not against adding this function and using it in the tests.
But the devel version of this PR did not add such function, and I currently do not see an open devel PR that would add it.
So why is this function added only in 3.11? Do we need to add it in a separate devel PR?

PR #20831, opened 2024-04-10
ExecuteAsync for RestHandlers

maierlars:
### Scope & Purpose
Add the `executeAsync` function to rest handlers. It can be implemented instead of the `execute` function. The future returned is automatically awaited and the rest handlers lifetime will exceed that of the execution of the future.

### DIFF
--- a/arangod/GeneralServer/RestHandler.cpp
+++ b/arangod/GeneralServer/RestHandler.cpp
@@ -781,3 +781,9 @@ void RestHandler::resetResponse(rest::ResponseCode code) {
   TRI_ASSERT(_response != nullptr);
   _response->reset(code);
 }
+
+futures::Future<futures::Unit> RestHandler::executeAsync() {
+  THROW_ARANGO_EXCEPTION(TRI_ERROR_NOT_IMPLEMENTED);
+}
+
+RestStatus RestHandler::execute() { return waitForFuture(executeAsync()); }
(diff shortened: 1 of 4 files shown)


goedderz (approved):

arangod/GeneralServer/RestHandler.cpp
```diff
@@ -781,3 +781,9 @@ void RestHandler::resetResponse(rest::ResponseCode code) {
   TRI_ASSERT(_response != nullptr);
   _response->reset(code);
 }
+
+futures::Future<futures::Unit> RestHandler::executeAsync() {
+  THROW_ARANGO_EXCEPTION(TRI_ERROR_NOT_IMPLEMENTED);
+}
+
+RestStatus RestHandler::execute() { return waitForFuture(executeAsync()); }
```
jsteemann, line 789: I am a bit confused about how this can actually work.
This PR adds `executeAsync()` as a new virtual method, but it is only overridden in a single RestHandler.

And the default implementation of RestHandler::execute() now changed to call executeAsync(), which be default will call the base class executeAsync(), which will throw an exception.

maierlars: Either implement `execute` or `executeAsync`. If you implement both `execute` will be called.

jsteemann (approved):
LGTM

PR #20833, opened 2024-04-11
[3.11] AsyncNode must not forward waiting result

mpoeter:
### Scope & Purpose

Backport of #20768

Fix an issue that can cause AQL related RestHandlers to wind up in a WAITING state and never get woken up. This implies that the associated query snippet and transaction cannot be released. If the query contains modification operations, this also implies that the associated collection write lock is not released and can therefore prevent other threads from acquiring the exclusive lock (which is required by e.g., the replication).

Fix an issue that can cause some background load in parallel traversals while waiting for data from upstream. For full details see https://github.com/arangodb/arangodb/pull/20768

- [x] :hankey: Bugfix

### Checklist

- [x] Tests
  - [x] C++ **Unit tests**


jsteemann (approved):
LGTM

hkernbach (approved):

PR #20837, opened 2024-04-11
Fix Locking for ResignLeadership

maierlars:
### Scope & Purpose
While streaming transactions and aql queries carefully lock multiple shards in a deterministic order, the maintenance did not care about such a order and just locked shards when resigning in any order. In fact the maintenance has no idea about distributeShardsLike and handles each shard of the shard group individually. A move shard requests locking of all shards in a shard group. This can lead to deadlocks with aql queries, since the locks are acquired in different orders.


arangod/Cluster/Maintenance.cpp
```diff
@@ -649,6 +663,77 @@ VPackBuilder getShardMap(VPackSlice const& collections) {
...
+
+  std::set<ShardID> result;
+  for (auto shard : VPackObjectIterator(collection.get(SHARDS))) {
+    result.insert(ShardID::shardIdFromString(shard.key.stringView()).get());
+  }
+  return result;
+}
+
+std::tuple<DistributeShardsLikeMapping, ClonePrototypeMapping>
+getDistributeShardsLike(VPackSlice const& collections) {
```
jsteemann, line 680:
```suggestion
getDistributeShardsLike(VPackSlice collections) {
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -649,6 +663,77 @@ VPackBuilder getShardMap(VPackSlice const& collections) {
...
+  DistributeShardsLikeMapping mapping;
+  ClonePrototypeMapping reverseMapping;
+
+  // maps a collection to the sorted set of shards
+  std::unordered_map<std::string_view, std::set<ShardID>> prototypeShards;
+
```
jsteemann, line 686: nit: these objects are needed only after the initial `if` statement in this function.
So we can move their declarations down to after the `if`.

arangod/Cluster/Maintenance.cpp
```diff
@@ -649,6 +663,77 @@ VPackBuilder getShardMap(VPackSlice const& collections) {
...
+      if (auto s = collection.value.get("distributeShardsLike"); s.isString()) {
+        return s.stringView();
+      }
+      return std::string_view{};
+    }();
+
+    if (prototype.empty()) {
+      continue;
+    }
+    // check is we have collected information about this prototype already
```
jsteemann, line 704:
```suggestion
    // check if we have collected information about this prototype already
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -649,6 +663,77 @@ VPackBuilder getShardMap(VPackSlice const& collections) {
...
+    auto it = prototypeShards.find(prototype);
+    if (it == prototypeShards.end()) {
+      auto protoCol = collections.get(prototype);
```
jsteemann, line 707: nit: I had a hard time read `protoCol` because I was intuitively thinking of `protocol` (as in HTTP, HTTPS, etc.). I get that it means `protoCollection` here, but naming it `protoCol` is slightly ambiguous, so I suggest using `protoCollection` instead.

arangod/Cluster/Maintenance.cpp
```diff
@@ -649,6 +663,77 @@ VPackBuilder getShardMap(VPackSlice const& collections) {
...
+
+    while (iterA != it->second.end()) {
+      TRI_ASSERT(iterB != cloneShards.end());
+      mapping[*iterA].insert(*iterB);
+      reverseMapping.emplace(*iterB, *iterA);
+      ++iterA;
+      ++iterB;
+    }
+    // prototype and clone have the same number of shards
+    TRI_ASSERT(iterB == cloneShards.end());
```
jsteemann, line 731: nit: I would be very happy if we made such assertion before we execute the while loop.

arangod/Cluster/ResignShardLeadership.cpp
```diff
@@ -130,15 +119,31 @@ bool ResignShardLeadership::first() {
     // for now but we will not accept any replication operation from any
     // leader, until we have negotiated a deal with it. Then the actual
     // name of the leader will be set.
-    col->followers()->setTheLeader(LeaderNotYetKnownString);  // resign
-    res = methods->abort();                                   // unlock
+    for (auto const& s : shards) {
+      auto col = vocbase->lookupCollection(s);
+      if (col == nullptr) {
+        std::stringstream error;
+        error << "Failed to lookup local collection " << collection
```
jsteemann, line 126: Probably we should include the shard name in the error message here as well, to have more detail should this error happen, e.g.
"Failed to lookup shard _s_ of local collection _collection_"

arangod/Cluster/ResignShardLeadership.cpp
```diff
@@ -130,15 +119,31 @@ bool ResignShardLeadership::first() {
...
+              << " in database " << database;
+        LOG_TOPIC("e06ca", ERR, Logger::MAINTENANCE)
+            << "ResignLeadership: " << error.str();
+        result(TRI_ERROR_ARANGO_DATA_SOURCE_NOT_FOUND, error.str());
+        return false;
+      }
+
+      col->followers()->setTheLeader(LeaderNotYetKnownString);  // resign
```
jsteemann, line 134: If we have multiple shards in `shards` and loop has multiple iterations, we will call `setTheLeader(LeaderNotYetKnownString)` for all shards until we run into an error and abort the loop with `false`.
So may have resign for some shards but not others. While this may be perfectly fine, I don't know if it actually is and if there is some way that this mixed state is automatically resolved later.

arangod/Cluster/ResignShardLeadership.cpp
```diff
@@ -130,15 +119,31 @@ bool ResignShardLeadership::first() {
...
     if (res.fail()) {
       LOG_TOPIC("10c35", ERR, Logger::MAINTENANCE)
           << "Failed to abort transaction during resign leadership: " << res;
     }
 
-    transaction::cluster::abortLeaderTransactionsOnShard(col->id());
-
+    for (auto const& s : shards) {
+      auto col = vocbase->lookupCollection(s);
+      transaction::cluster::abortLeaderTransactionsOnShard(col->id());
```
jsteemann, line 145: nit: it is theoretically possible that the lookup call _now_ produces a nullptr, because the shard was dropped on this server after the exclusive lock was released.
This may be an edge case, but I think it is better to check here if `col` is a nullptr or not.

PR #20777, opened 2024-03-25
[BTS-1034]: Make ArangoLanguageFeature unit test independent from system locale

alexbakharew:
### Scope & Purpose

This PR fixes bug for ArangoLanguageFeature test when current system locale is "DE".
 

- [X] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.12.0: *(Please link PR)*
  - [ ] Backport for 3.11: *(Please link PR)*
  - [ ] Backport for 3.10: *(Please link PR)*

#### Related Information

*(Please reference tickets / specification / other PRs etc)*

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [X] GitHub issue / Jira ticket: https://arangodb.atlassian.net/browse/BTS-1034
- [ ] Design document:

PR #20844, opened 2024-04-15
[3.11]: Make ArangoLanguageFeature unit test independent from system locale

alexbakharew:
Make ArangoLanguageFeature unit test independent from system locale (#20777)

* Initial commit

* test debug

* wip

* wip

* fix clang-format

* clear debug output

### Scope & Purpose

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/tests/RestServer/LanguageFeatureTest.cpp
+++ b/tests/RestServer/LanguageFeatureTest.cpp
@@ -136,9 +136,8 @@ std::string_view getNonSysLang() {
   const std::string_view currSysLang =
       std::string_view(setlocale(LC_ALL, NULL)).substr(0, 2);
 
-  auto sg = arangodb::scopeGuard([&]() noexcept {
-    setlocale(LC_ALL, before.c_str());
-  });
+  auto sg = arangodb::scopeGuard(
+      [&]() noexcept { setlocale(LC_ALL, before.c_str()); });
 
   if (currSysLang == "de") {
     return kRussian;
--- a/tests/RestServer/LanguageFeatureTest.cpp
+++ b/tests/RestServer/LanguageFeatureTest.cpp
@@ -131,13 +131,13 @@ constexpr std::string_view kGerman = "de";
 constexpr std::string_view kRussian = "ru";
 
 std::string_view getNonSysLang() {
-  auto before = std::string_view(setlocale(LC_ALL, NULL)).substr(0, 2);
+  auto before = std::string(setlocale(LC_ALL, NULL));
   setlocale(LC_ALL, "");
   const std::string_view currSysLang =
       std::string_view(setlocale(LC_ALL, NULL)).substr(0, 2);
 
   auto sg = arangodb::scopeGuard([&]() noexcept {
-    setlocale(LC_ALL, before.data());
+    setlocale(LC_ALL, before.c_str());
   });
 
   if (currSysLang == "de") {
--- a/tests/RestServer/LanguageFeatureTest.cpp
+++ b/tests/RestServer/LanguageFeatureTest.cpp
@@ -36,14 +36,14 @@
 #include "Basics/files.h"
 #include "Basics/icu-helper.h"
 #include "IResearch/common.h"
+#include <locale.h>
 
 // -----------------------------------------------------------------------------
 // --SECTION--                                                 setup / tear-down
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


tests/RestServer/LanguageFeatureTest.cpp
```diff
@@ -36,22 +36,22 @@
...
+void checkCollatorSettings(std::string_view language, bool isDefaultLanguage) {
   // Create collator with expected language
   UErrorCode status = U_ZERO_ERROR;
   icu::Collator* expectedColl = nullptr;
   if (language == "") {
     // get default collator for empty language
     expectedColl = icu::Collator::createInstance(status);
   } else {
-    icu::Locale locale(language.c_str());
+    icu::Locale locale(language.data());
```
jsteemann, line 54: Is the value of this string_view guaranteed to be null-terminated?
Because std::string_views are not, so creating a locale from a raw data pointer here may read over the std::string_view's bounds.

tests/RestServer/LanguageFeatureTest.cpp
```diff
@@ -127,6 +126,24 @@ void checkLanguageFile(const arangodb::ArangodServer& server,
     ASSERT_TRUE(false);
   }
 }
+
+constexpr std::string_view kGerman = "de";
+constexpr std::string_view kRussian = "ru";
+
+std::string_view getNonSysLang() {
+  setlocale(LC_ALL, "");
```
jsteemann, line 134: This changes the locale for not only this particular test, but for all that follow. So that modifies global state.
It would be good if the test would clean up after itself and reset the locale to what it was before the test executed.

tests/RestServer/LanguageFeatureTest.cpp
```diff
@@ -308,8 +325,7 @@ TEST_F(ArangoLanguageFeatureTest, testDefaultLangCheckTrue) {
 
   constexpr std::string_view firstLang = "sv";
   constexpr std::string_view secondLang = "de";
-  constexpr std::string_view defaultParameter = "default-language";
-  constexpr std::string_view icuParameter = "icu-language";
+  ;
```
jsteemann, line 328:
```suggestion
```

tests/RestServer/LanguageFeatureTest.cpp
alexbakharew: yes, `language` variables here are created like this:
```
constexpr std::string_view language2 = "sv";
```

So they are guaranteed to be null terminated.

tests/RestServer/LanguageFeatureTest.cpp
```diff
@@ -127,6 +126,30 @@ void checkLanguageFile(const arangodb::ArangodServer& server,
...
+constexpr std::string_view kRussian = "ru";
+
+std::string_view getNonSysLang() {
+  auto before = std::string_view(setlocale(LC_ALL, NULL)).substr(0, 2);
+  setlocale(LC_ALL, "");
+  const std::string_view currSysLang =
+      std::string_view(setlocale(LC_ALL, NULL)).substr(0, 2);
+
+  auto sg = arangodb::scopeGuard([&]() noexcept {
+    setlocale(LC_ALL, before.data());
```
jsteemann, line 140: This won't work as expected.
Although `before` now has a length of (up to) 2 bytes, using only its data pointer will point to a string of that is not necessarily just 2 bytes long.

Try running this program:
```
#include <string_view>
#include <iostream>

int main() {
  char const* loca = "de_DE";
  auto before = std::string_view(loca).substr(0, 2);

  std::cout << "before: "        << before << "\n";
  std::cout << "before.data(): " << before.data() << "\n";
}
```
This prints
```
$ g++ -std=c++20 sv.cpp && ./a.out 
before: de
before.data(): de_DE
```
Whenever you need to guarantee that the strings are null-terminated, it is best to use std::string and not std::string_view.
And you need the null-termination because the pointer is passed into a C function.

jsteemann (approved):
LGTM

PR #20864, opened 2024-04-23
cURL for internal requests

maierlars:
### Scope & Purpose
Instead of rolling our own http2 client, we use curl, which supports multiplexing via http2 and connection pools. This allows us to remove a lot of our (not soo reliable) network code.

**This PR is a proof of concept**: it quickly hacks in curl at the lowest level and hands over all network requests to the library. After we ran tests and performance analysis, I will put more effort into remove the old code and all traces of fuerte.


maierlars:
Curl is simply too slow.

PR #20868, opened 2024-04-25
Adding parenthesis around error message

apetenchea:
### Scope & Purpose

While working on the Python driver, I noticed that an error message looks rather strange: `Value cannot be empty. on attribute distributeShardsLike`. It's nothing bad really, but due to the way the string gets formatted in some cases we get a dot before the on _attribute..._, which looks as if the sentence is incomplete.
This PR simply adds some parenthesis around that, to make it look more like an additional explanation of the previous sentence: `Value cannot be empty. (on attribute "distributeShardsLike")`

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


jsteemann (approved):
LGTM

dothebart (approved):

PR #20870, opened 2024-04-26
add fixes for socket shutdown

jsteemann:
### Scope & Purpose

* Fix connection retry attempts for cluster-internal TLS connections that ran into the 15 seconds timeout during the connection establishing attempt. In this case, the low-level socket was repurposed, but not reset properly. This could leave the connection in an improper state and lead to callbacks for some requests to not being called as expected.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/3rdParty/fuerte/src/AsioSockets.h
+++ b/3rdParty/fuerte/src/AsioSockets.h
@@ -146,7 +146,7 @@ struct Socket<fuerte::SocketType::Ssl> {
     bool verify = config._verifyHost;
     resolveConnect(
         config, resolver, socket.next_layer(),
-        [=, this, done(std::forward<F>(done))](auto const& ec) mutable {
+        [=, this](auto const& ec) mutable {
           if (ec) {
             done(ec);
             return;
--- a/3rdParty/fuerte/src/AsioSockets.h
+++ b/3rdParty/fuerte/src/AsioSockets.h
@@ -103,7 +103,7 @@ struct Socket<SocketType::Tcp> {
   void connect(detail::ConnectionConfiguration const& config, F&& done) {
     resolveConnect(config, resolver, socket, [this, done = std::forward<F>(done)](asio_ns::error_code ec) mutable {
       FUERTE_LOG_DEBUG << "executing tcp connect callback, ec: " << ec.message() << ", canceled: " << this->canceled;
-      if (this->canceled) {
+      if (canceled) {
         // cancel() was already called on this socket
         ec = asio_ns::error::operation_aborted;
       }
@@ -116,6 +116,7 @@ struct Socket<SocketType::Tcp> {
   }
 
   void cancel() {
+    canceled = true;
     try {
       timer.cancel();
       resolver.cancel();
@@ -174,7 +175,7 @@ struct Socket<fuerte::SocketType::Ssl> {
         config, resolver, socket.next_layer(),
         [=, this](asio_ns::error_code ec) mutable {
           FUERTE_LOG_DEBUG << "executing ssl connect callback, ec: " << ec.message() << ", canceled: " << this->canceled;
-          if (this->canceled) {
+          if (canceled) {
            // cancel() was already called on this socket
            ec = asio_ns::error::operation_aborted;
           }
@@ -223,7 +224,7 @@ struct Socket<fuerte::SocketType::Ssl> {
   }
 
   void cancel() {
-    this->canceled = true;
+    canceled = true;
     try {
       timer.cancel();
       resolver.cancel();
@@ -297,7 +298,7 @@ struct Socket<fuerte::SocketType::Unix> {
       : socket(ctx), timer(ctx) {}
 
   ~Socket() { 
-    this->canceled = true;
+    canceled = true;
     try {
       this->cancel(); 
     } catch (std::exception const& ex) {
@@ -307,7 +308,7 @@ struct Socket<fuerte::SocketType::Unix> {
 
   template <typename F>
   void connect(detail::ConnectionConfiguration const& config, F&& done) {
-    if (this->canceled) {
+    if (canceled) {
       // cancel() was already called on this socket
       done(asio_ns::error::operation_aborted);
       return;
@@ -322,7 +323,7 @@ struct Socket<fuerte::SocketType::Unix> {
   }
 
   void cancel() {
-    this->canceled = true;
+    canceled = true;
     try {
       timer.cancel();
       if (socket.is_open()) {  // non-graceful shutdown
--- a/3rdParty/fuerte/src/AsioSockets.h
+++ b/3rdParty/fuerte/src/AsioSockets.h
@@ -35,18 +35,8 @@ void resolveConnect(detail::ConnectionConfiguration const& config,
                     asio_ns::ip::tcp::resolver& resolver, SocketT& socket,
                     F&& done, IsAbortedCb&& isAborted) {
   auto cb = [&socket, 
-#ifdef ARANGODB_USE_GOOGLE_TESTS
-             fail = config._failConnectAttempts > 0,
-#endif
              done = std::forward<F>(done),
              isAborted = std::forward<IsAbortedCb>(isAborted)](auto ec, auto it) mutable {
-#ifdef ARANGODB_USE_GOOGLE_TESTS
-    if (fail) {
-      // use an error code != operation_aborted
-      ec = boost::system::errc::make_error_code(boost::system::errc::not_enough_memory);
-    }
-#endif
-
     if (isAborted()) {
       ec = asio_ns::error::operation_aborted;
     }
@@ -121,10 +111,6 @@ struct Socket<SocketType::Tcp> {
     }, [this]() {
       return canceled;
     });
-  }
-
-  void rearm() {
-    canceled = false;
   }
 
   void cancel() {
@@ -233,11 +219,6 @@ struct Socket<fuerte::SocketType::Ssl> {
         });
   }
   
-  void rearm() {
-    socket = asio_ns::ssl::stream<asio_ns::ip::tcp::socket>(this->ctx, this->sslContext);
-    canceled = false;
-  }
-
   void cancel() {
     canceled = true;
     try {
@@ -333,10 +314,6 @@ struct Socket<fuerte::SocketType::Unix> {
     socket.async_connect(ep, std::forward<F>(done));
   }
   
-  void rearm() {
-    canceled = false;
-  }
-
   void cancel() {
     canceled = true;
     try {
--- a/3rdParty/fuerte/src/AsioSockets.h
+++ b/3rdParty/fuerte/src/AsioSockets.h
@@ -34,7 +34,11 @@ template <typename SocketT, typename F>
 void resolveConnect(detail::ConnectionConfiguration const& config,
                     asio_ns::ip::tcp::resolver& resolver, SocketT& socket,
                     F&& done) {
-  auto cb = [&socket, done = std::forward<F>(done), fail = config._failConnectAttempts > 0](auto ec, auto it) mutable {
+  auto cb = [&socket, 
+#ifdef ARANGODB_USE_GOOGLE_TESTS
+             fail = config._failConnectAttempts > 0,
+#endif
+             done = std::forward<F>(done)](auto ec, auto it) mutable {
 #ifdef ARANGODB_USE_GOOGLE_TESTS
     if (fail) {
       // use an error code != operation_aborted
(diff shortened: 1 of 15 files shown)


neunhoef (approved):
LGTM

neunhoef (approved):
LGTM. This looks great, it is much simplified and I am now reasonably convinced that this is going to work and we are not leaking anything. Thanks!

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -77,11 +91,28 @@ struct Socket<SocketType::Tcp> {
...
+      }
+      done(ec);
+    });
+  }
+
+  void rearm() {
+    canceled = false;
   }
 
   void cancel() {
```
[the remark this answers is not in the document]
neunhoef: Or else we do not need to check canceled in the `connect` method above.

3rdParty/fuerte/src/GeneralConnection.h
```diff
@@ -300,57 +315,79 @@ class GeneralConnection : public fuerte::Connection {
...
-            [self(std::move(self)), start, retries](auto ec) mutable {
+            [self = std::move(self), retries](asio_ns::error_code ec) mutable {
+              if (ec) {
+                FUERTE_LOG_DEBUG << "tryConnect, retry timer canceled. this=" << self.get() << "\n";
+                // timer canceled.
+                return;
+              }
               auto& me = static_cast<GeneralConnection<ST, RT>&>(*self);
-              me.tryConnect(!ec ? retries - 1 : 0, start, ec);
+              if (retries - 1 == 0) {
```
neunhoef, line 359: Looks strange, why not `retries == 1`?

3rdParty/fuerte/src/GeneralConnection.h
```diff
@@ -300,57 +315,79 @@ class GeneralConnection : public fuerte::Connection {
...
+                std::string msg("connecting failed: timeout");
+                FUERTE_LOG_DEBUG << "tryConnect, timeout this=" << self.get() << "\n";
+                me.shutdownConnection(Error::CouldNotConnect, msg);
```
neunhoef, line 362: I do not understand why we suddenly need two places which call `shutdownConnection`, where we had only one before. But it is now too late to see this clearly. Will have another look tomorrow morning.
It seems to me that either we catch the case of exhausted retries here before calling ourselves again with `retries-1`, or we can catch it outside, but I do not think we need both.

neunhoef (changes requested):
I need to look at this with a fresh mind in the morning. The changes requested are all tiny, but in the main async connect callback, I think a simplification is possible.

3rdParty/fuerte/src/GeneralConnection.h
```diff
@@ -300,57 +315,79 @@ class GeneralConnection : public fuerte::Connection {
...
       }
-      FUERTE_LOG_DEBUG << "connecting failed: " << ec.message() << "\n";
+#ifdef ARANGODB_USE_GOOGLE_TESTS
+      if (me._failConnectAttempts > 0) {
+        --me._failConnectAttempts;
+      }
+#endif
+
+      FUERTE_LOG_DEBUG << "tryConnect (" << retries << "), connecting failed: " << ec.message() << "\n";
       if (retries > 0 && ec != asio_ns::error::operation_aborted) {
```
neunhoef, line 347: `retries` can never be zero here after the recent changes, since it is a copy of the argument `retries` to `tryConnect`.

neunhoef: But see comment below that I suggest to change this to `retries > 1`.

3rdParty/fuerte/src/GeneralConnection.h
neunhoef: With fresh mind in the morning: I think in the case that `retries == 1`, we should not wait for `_connectRetryPause` before we give up without a further try. Therefore, if we change the above condition to
```
if (retries > 1 && ec != asio_ns::error::operation_aborted) {
```
then we would not even do the `async_wait` here in this case. That is, after the delay in this closure, we know that `retries > 1` and can directly call `rearm` and `tryConnect(retries-1);`.

This has the additional benefit that we do not need the shutdown code twice.

3rdParty/fuerte/src/GeneralConnection.h
neunhoef: This comment has become obsolete with the above one.

neunhoef (changes requested):
On second inspection in the morning I believe I found a simplification. Let's discuss in a meeting.

3rdParty/fuerte/src/GeneralConnection.h
```diff
@@ -300,57 +315,73 @@ class GeneralConnection : public fuerte::Connection {
...
+        std::string msg("connecting failed: ");
+        msg.append((ec != asio_ns::error::operation_aborted) ? ec.message()
+                                                             : "timeout");
+        FUERTE_LOG_DEBUG << "tryConnect, calling shutdownConnection(2): " << msg << " this=" << self.get();
+        me.shutdownConnection(Error::CouldNotConnect, msg);
       }
     });
+    
+    auto start = Clock::now();
+    _proto.timer.expires_at(start + _config._connectTimeout);
```
mpoeter, line 374: We could simply use `expires_after` (same obviously goes for the other `expires_at` calls we have in this file).

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -34,8 +34,16 @@ template <typename SocketT, typename F>
 void resolveConnect(detail::ConnectionConfiguration const& config,
                     asio_ns::ip::tcp::resolver& resolver, SocketT& socket,
                     F&& done) {
-  auto cb = [&socket, done(std::forward<F>(done))](auto ec, auto it) mutable {
+  auto cb = [&socket, done = std::forward<F>(done), fail = config._failConnectAttempts > 0](auto ec, auto it) mutable {
```
mpoeter, line 37: This won't build if `ARANGODB_USE_GOOGLE_TESTS` is not defined, because then we have no member `_failConnectAttempts`
```suggestion
  auto cb = [&socket, done = std::forward<F>(done), fail = config._failConnectAttempts > 0](auto ec, auto it) mutable {
```

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -77,22 +91,41 @@ struct Socket<SocketType::Tcp> {
   Socket(EventLoopService&, asio_ns::io_context& ctx)
       : resolver(ctx), socket(ctx), timer(ctx) {}
 
-  ~Socket() { this->cancel(); }
+  ~Socket() { 
+    try {
+      this->cancel(); 
+    } catch (std::exception const& ex) {
+      FUERTE_LOG_ERROR << "caught exception during tcp socket shutdown: " << ex.what();
```
mpoeter, line 98: By default error logging is off, so we are again silently swallowing exceptions. :/

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -77,22 +91,41 @@ struct Socket<SocketType::Tcp> {
...
+  }
 
   template <typename F>
   void connect(detail::ConnectionConfiguration const& config, F&& done) {
-    resolveConnect(config, resolver, socket, std::forward<F>(done));
+    resolveConnect(config, resolver, socket, [this, done = std::forward<F>(done)](asio_ns::error_code ec) mutable {
+      FUERTE_LOG_DEBUG << "executing tcp connect callback, ec: " << ec.message() << ", canceled: " << this->canceled;
+      if (canceled) {
+        // cancel() was already called on this socket
+        ec = asio_ns::error::operation_aborted;
```
mpoeter, line 108: I would like to add an assertion that `is_open() == false` here.

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -77,22 +91,41 @@ struct Socket<SocketType::Tcp> {
...
     try {
       timer.cancel();
       resolver.cancel();
       if (socket.is_open()) {  // non-graceful shutdown
         asio_ns::error_code ec;
         socket.close(ec);
       }
-    } catch (...) {
+    } catch (std::exception const& ex) {
+      FUERTE_LOG_ERROR << "caught exception during tcp socket cancelation: " << ex.what();
```
mpoeter, line 128: By default error logging is off, so we are again silently swallowing exceptions. :/

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -108,29 +141,44 @@ struct Socket<SocketType::Tcp> {
...
   Socket(EventLoopService& loop, asio_ns::io_context& ctx)
-    : resolver(ctx), socket(ctx, loop.sslContext()), timer(ctx), cleanupDone(false) {}
+    : resolver(ctx), socket(ctx, loop.sslContext()), timer(ctx), ctx(ctx), sslContext(loop.sslContext()), cleanupDone(false) {}
 
-  ~Socket() { this->cancel(); }
+  ~Socket() { 
+    try {
+      this->cancel(); 
+    } catch (std::exception const& ex) {
+      FUERTE_LOG_ERROR << "caught exception during ssl socket shutdown: " << ex.what();
```
mpoeter, line 167: By default error logging is off, so we are again silently swallowing exceptions. :/

mpoeter (changes requested):

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -44,7 +52,12 @@ void resolveConnect(detail::ConnectionConfiguration const& config,
       // A successful resolve operation is guaranteed to pass a
       // non-empty range to the handler.
       asio_ns::async_connect(socket, it,
```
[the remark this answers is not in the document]
neunhoef: "Need" or "need not"? I think this should read "need to check".

3rdParty/fuerte/src/AsioSockets.h
neunhoef: Isn't it possible that both the connect succeed and the timeout is triggered, so both callbacks are scheduled. In this case, it is possible that the timeout is scheduled first and closes the socket and your assertion here would fail?

neunhoef: No, this is nonense, please ignore. THis is the "cancelled == true" case, in this case the socket must be closed.

3rdParty/fuerte/src/AsioSockets.h
```diff
@@ -108,29 +141,44 @@ struct Socket<SocketType::Tcp> {
         ec.clear();
         socket.close(ec);
       }
-    } catch (...) {
+    } catch (std::exception const& ex) {
+      // an exception is unlikely to occur here, as we are using the error-code
+      // variants of cancel/shutdown/close above 
+      FUERTE_LOG_ERROR << "caught exception during tcp socket shutdown: " << ex.what();
```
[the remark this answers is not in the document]
neunhoef: But what should we do instead?

3rdParty/fuerte/src/AsioSockets.h
neunhoef: The only way out I see is to switch on error logging by default and leave this here as it is.

3rdParty/fuerte/src/AsioSockets.h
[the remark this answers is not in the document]
mpoeter: Of course "need to" 🤦‍♂️

neunhoef (changes requested):
I think Manuel is right, we should check the `canceled` flag before we do `async_connect` in `resolveConnect`. Other than that: LGTM.
Should we enable error logging for fuerte by default on log level ERROR?

3rdParty/fuerte/src/AsioSockets.h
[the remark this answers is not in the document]
jsteemann: Done

mpoeter (approved):

neunhoef (approved):
LGTM

PR #20874, opened 2024-04-26
Transactions reaching max size

apetenchea:
### Scope & Purpose

The `maxTransactionSize` option is there to limit the memory usage of a given transaction. Upon reaching this limitation, any subsequent operations are being rejected, but the transaction can be continued and even committed. However, the error message is misleading, as it states the the transaction is being *aborted*, which is not the case.

This PR modifies the error message and adds a clear test, showcasing the expected behavior.

- [ ] :hankey: Bugfix
- [ ] :pizza: New feature


jsteemann (approved):
LGTM

PR #20885, opened 2024-05-02
Maintenance Speedup + Small Refactoring

maierlars:
### Scope & Purpose
First compute the shard map outside of the loop. Otherwise we end up with a quadratic algorithm.

Furthermore, while I'm at it, I replaced some adhoc velocypack with some proper c++ data structure.

### DIFF
--- a/arangod/Cluster/Maintenance.cpp
+++ b/arangod/Cluster/Maintenance.cpp
@@ -1063,6 +1063,7 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
           continue;
         }
 
+        // NOTE: shardMap only stays valid as long as plan is not deleted. Do not return it.
         auto const shardMap = getShardMap(plan);  // plan shards -> servers
         for (auto const& lcol : VPackObjectIterator(ldbslice)) {
           auto maybeShardID = ShardID::shardIdFromString(lcol.key.stringView());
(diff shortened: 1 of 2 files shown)


arangod/Cluster/Maintenance.cpp
```diff
@@ -1959,7 +1969,7 @@ arangodb::Result arangodb::maintenance::reportInCurrent(
       return replication::Version::ONE;
     });
 
-    VPackBuilder shardMap;
+    ShardMapView shardMap;
```
mchacki, line 1972:
```suggestion
    // NOTE: shardMap only stays valid as long as plan is not deleted. Do not return it.
    ShardMapView shardMap;
```

arangod/Cluster/Maintenance.cpp
```diff
@@ -1054,19 +1063,19 @@ arangodb::Result arangodb::maintenance::diffPlanLocal(
           continue;
         }
 
+        auto const shardMap = getShardMap(plan);  // plan shards -> servers
```
mchacki, line 1066:
```suggestion
        // NOTE: shardMap only stays valid as long as plan is not deleted. Do not return it.
        auto const shardMap = getShardMap(plan);  // plan shards -> servers
```

mchacki (changes requested):
Technical: Very good :+1:

Please add a CHANGELOG entry.
For consistency with 3.11 you can take:


```
* Improved the time required to create a new Collection in a database with
  hundreds of collections. This also improves times for indexes and dropping
  of collections.
```

mchacki (approved):
LGTM!

jsteemann (approved):
LGTM

PR #20908, opened 2024-05-07
remove retry logic for fuerte connections

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/20907

* Fix connection retry attempts for cluster-internal TLS connections that ran
  into the 15 seconds timeout during the connection establishing attempt.
  In this case, the low-level socket was repurposed, but not reset properly.
  This could leave the connection in an improper state and lead to callbacks
  for some requests to not being called as expected.
  The fix is to remove the retry logic, and instead prolong the connection
  timeout to a multiple of the previous timeout value.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


jsteemann:
We will abandon this solution without retries in favor of a solution with retries.

PR #20929, opened 2024-05-14
Wait for ClusterInfo to catch up after creating a collection

goedderz:
### Scope & Purpose

Currently, a Coordinator that's creating a Replication 2 collection does not wait for its own ClusterInfo to catch up, which can sometimes lead to "collection not found" errors on that same coordinator after a successful collection creation. This also leads to some spurious test failures.

This PR waits for CI to catch up.

- [X] :hankey: Bugfix


maierlars (approved):

jsteemann (approved):
LGTM

PR #20932, opened 2024-05-14
Remove server dependency of Logger and Thread

goedderz:
### Scope & Purpose

Remove dependency on the Server of Logger and Thread; this makes it easier to instantiate them in tests.

This removes a check whether a thread may be started depending on the server's state (in `Thread::start`).


maierlars (approved):

jsteemann (approved):
LGTM

PR #20945, opened 2024-05-16
Make MetricsFeature's dependencies explicit

goedderz:
### Scope & Purpose

Make the dependencies of MetricsFeature explicit, so it can get by without getFeature calls, so it can be an ApplicationFeature instead of an ArangodFeature, so it can be used separately from it.

In order to achieve that, this PR introduces LazyApplicationFeatureReference to handle potential circular references in the feature dependencies (which we should try to get rid of later, whenever possible).

#### Related Information

- [X] Enterprise PR: https://github.com/arangodb/enterprise/pull/1492
- [X] Jira ticket: https://arangodb.atlassian.net/jira/software/c/projects/TD/issues/TD-1


arangod/Metrics/MetricsFeature.cpp
```diff
@@ -365,4 +388,12 @@ void MetricsFeature::batchRemove(std::string_view name,
...
 }
 
+void MetricsFeature::prepare() {
+  _queryRegistryFeature = std::move(_lazyQueryRegistryFeatureRef).get();
+  _statisticsFeature = std::move(_lazyStatisticsFeatureRef).get();
+  _engineSelectorFeature = std::move(_lazyEngineSelectorFeatureRef).get();
+  _clusterMetricsFeature = std::move(_lazyClusterMetricsFeatureRef).get();
+  _clusterFeature = std::move(_lazyClusterFeatureRef).get();
+}
+
```
alexbakharew, line 398: Should we assert somehow that `prepare()` and therefore `get()` methods are really called only once?

jsteemann (approved):
LGTM

arangod/Metrics/MetricsFeature.cpp
goedderz: It is asserted that `get()` is called only once: https://github.com/arangodb/arangodb/pull/20945/files?show-viewed-files=true&file-filters%5B%5D=#diff-ba7c32ee3716b0719387545111a837bd0a213f052957a14db2d32b9ee24be3ccR103

alexbakharew (approved):
LGTM

PR #20946, opened 2024-05-16
Remove non-local AgencyCallbacks

goedderz:
### Scope & Purpose

Deleted the non-local variant of AgencyCallbacks - they are unused and don't make sense any longer. Also got rid of ArangodServer as a dependency of AgencyComm, AgencyCallback, and AgencyCallbackRegistry. The latter as part of a larger undertaking to make features reusable again (note that features needing ArangodServer cannot be used without a complete ArangodServer, which makes them practically unusable for other tools or certain tests).

#### Related Information

- [X] Jira ticket: https://arangodb.atlassian.net/jira/software/c/projects/TD/issues/TD-1


arangod/Cluster/AgencyCallback.cpp
```diff
@@ -89,35 +89,18 @@ void AgencyCallback::refetchAndUpdate(bool needToAcquireMutex,
...
-    if (!result.isArray()) {
-      if (!_server.isStopping()) {
-        // only log errors if we are not already shutting down...
-        // in case of shutdown this error is somewhat expected
-        LOG_TOPIC("ec320", ERR, arangodb::Logger::CLUSTER)
-            << "Callback to get agency cache was not successful: "
-            << result.toJson();
-      }
-      return;
+  std::tie(builder, idx) =
```
jsteemann, line 92: There is no need anymore to use `std::tie(builder, idx)` here to assign data to variables declared above.
Instead we could use `auto [builder, idx] = ...` here and remove the separate declarations above.
We then also can remove the now-unused variable `tmp` above.

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

neunhoef (approved):
LGTM

PR #20952, opened 2024-05-21
Fix potential race in collectRebalanceInformation

mpoeter:
### Scope & Purpose

shardIds returns a shared_ptr copy which must stay alive long enough. However, if we dereference the shared_ptr as part of the range expression, then we only keep the reference to the ShardMap, but the temporary is immediately destroyed and does not guarantee that the map stays around.

- [x] :hankey: Bugfix

### Checklist

- [x] Backports
  - [x] Backport for 3.11: #20953

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 devel
 -----
 
+* Fix a potential race in collectRebalanceInformation.
+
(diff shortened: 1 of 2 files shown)


jsteemann (approved):
LGTM

dothebart (approved):

maierlars (approved):

PR #20961, opened 2024-05-23
Use posix_spawn instead of fork+exec

maierlars:
### Scope & Purpose
Use `posix_spawn` instead of fork + exec calls.


kvahed (approved):
LGTM

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+        close(pipe_child_to_server[0]);
+        close(pipe_child_to_server[1]);
+      }
+
+      return;
+    }
+  }
+
+  LOG_TOPIC("ac58b", DEBUG, arangodb::Logger::FIXME)
+      << "spawn succeeded, child pid: " << external->_pid;
```
mpoeter, line 312: This log message is a bit confusing if `executableNotFound` is set.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+  if (usePipes) {
+    posix_spawn_file_actions_adddup2(&file_actions, pipe_server_to_child[0], 0);
+    posix_spawn_file_actions_adddup2(&file_actions, pipe_child_to_server[1], 1);
+
+    posix_spawn_file_actions_addclose(&file_actions, pipe_server_to_child[0]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_server_to_child[1]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_child_to_server[0]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_child_to_server[1]);
+  } else {
+    posix_spawn_file_actions_addopen(&file_actions, 0, "/dev/null", O_RDONLY,
```
dothebart, line 259: By this you probably change the current behaviour to derive the callers stdout/stderr file handles?
Probably `usePipes` should be a tri-state of
- use Pipes
- derive STDOUT/STDERR
- STFU

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+  // file actions are performed in order they were added.
+
+  if (usePipes) {
+    posix_spawn_file_actions_adddup2(&file_actions, pipe_server_to_child[0], 0);
+    posix_spawn_file_actions_adddup2(&file_actions, pipe_child_to_server[1], 1);
+
+    posix_spawn_file_actions_addclose(&file_actions, pipe_server_to_child[0]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_server_to_child[1]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_child_to_server[0]);
+    posix_spawn_file_actions_addclose(&file_actions, pipe_child_to_server[1]);
```
jsteemann, line 257: Note: all of these function calls can fail, which is indicated by any of the calls returning a value other than zero. Do we want to add error handling here? I guess in case of an error all we can do set `external->_status = TRI_EXT_PIPE_FAILED;`, log a message and error out.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+                                     0);
+  }
+
+  if (!fileForStdErr.empty()) {
+    posix_spawn_file_actions_addopen(&file_actions, 2, fileForStdErr.c_str(),
+                                     O_CREAT | O_WRONLY | O_TRUNC, 0644);
+  }
+
+  posix_spawnattr_t spawn_attrs;
+  posix_spawnattr_init(&spawn_attrs);
```
jsteemann, line 269: Do we want to handle a potential error returned by `posix_spawnattr_init`?

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+    envs.push_back(*e);
+  }
+
+  envs.reserve(additionalEnv.size() + 1);
+  std::transform(
+      additionalEnv.begin(), additionalEnv.end(), std::back_inserter(envs),
+      [](auto& str) -> char* { return const_cast<char*>(str.data()); });
+  envs.emplace_back(nullptr);
+
+  int result = posix_spawnp(&external->_pid, external->_executable.c_str(),
```
jsteemann, line 282: I am not 100% clear about how signal handling for the child process will work with this call.
The previous code that uses `fork` and `exec` called `arangodb::signals::unmaskAllSignals();` for the child, to unblock all signals.
If this is still happening by default when using `posix_spawnp` is unknown to me. I read the man page (https://man7.org/linux/man-pages/man3/posix_spawn.3p.html) but found it confusing:

        If the POSIX_SPAWN_SETSIGMASK flag is set in the spawn-flags
       attribute of the object referenced by attrp, the child process
       shall initially have the signal mask specified in the spawn-
       sigmask attribute of the object referenced by attrp.

       If the POSIX_SPAWN_SETSIGDEF flag is set in the spawn-flags
       attribute of the object referenced by attrp, the signals
       specified in the spawn-sigdefault attribute of the same object
       shall be set to their default actions in the child process.
       Signals set to the default action in the parent process shall be
       set to the default action in the child process.

       Signals set to be caught by the calling process shall be set to
       the default action in the child process.

       Except for SIGCHLD, signals set to be ignored by the calling
       process image shall be set to be ignored by the child process,
       unless otherwise specified by the POSIX_SPAWN_SETSIGDEF flag
       being set in the spawn-flags attribute of the object referenced
       by attrp and the signals being indicated in the spawn-sigdefault
       attribute of the object referenced by attrp.

       If the SIGCHLD signal is set to be ignored by the calling
       process, it is unspecified whether the SIGCHLD signal is set to
       be ignored or to the default action in the child process, unless
       otherwise specified by the POSIX_SPAWN_SETSIGDEF flag being set
       in the spawn_flags attribute of the object referenced by attrp
       and the SIGCHLD signal being indicated in the spawn_sigdefault
       attribute of the object referenced by attrp.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+                            &file_actions, &spawn_attrs, external->_arguments,
+                            envs.data());
+
+  posix_spawnattr_destroy(&spawn_attrs);
+  posix_spawn_file_actions_destroy(&file_actions);
+
+  bool executableNotFound = false;
+
+  if (result != 0) {
+    if (errno == ENOENT) {
```
jsteemann, line 292: I am not very happy about `errno` being used here for checking the error code returned by the failed `posix_spawnp` call. While checking `errno` is correct, I would prefer to do it directly after the call to `posix_spawnp`. Otherwise it is not guaranteed that the calls to `posix_spawnattr_destroy` or `posix_spawn_file_actions_destroy` do not clobber `errno` and overwrite it with something else. 
The same applies for the usage of `errno` later in the `strerror(errno)`. 
I suggest to simply assign the original `errno` value to a variable, and then use that variable later.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+      external->_status = TRI_EXT_PIPE_FAILED;
+      return;
+    }
+  }
+
+  posix_spawn_file_actions_t file_actions;
+  posix_spawn_file_actions_init(&file_actions);
+  // file actions are performed in order they were added.
+
+  if (usePipes) {
```
jsteemann, line 250: The original code also called 
```
fcntl(0, F_SETFD, 0);
fcntl(1, F_SETFD, 0);
fcntl(2, F_SETFD, 0);
```
in the child process. I guess this removed the `FD_CLOEXEC` flag from these file descriptors. I am unsure if this is still needed.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +225,115 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
...
+
+  if (executableNotFound) {
+    external->_status = TRI_EXT_TERMINATED;
+    external->_exitStatus = 1;
+  } else {
+    external->_status = TRI_EXT_RUNNING;
+  }
+}
+
+[[maybe_unused]] static void StartExternalProcess(
```
jsteemann, line 333: I think that the `[[maybe_unused]]` is not necessary.
```suggestion
static void StartExternalProcess(
```

lib/Basics/process-utils.cpp
neunhoef: And we leak open file descriptors in this case...

neunhoef: Refactored.

lib/Basics/process-utils.cpp
neunhoef: I have added error handling.

lib/Basics/process-utils.cpp
maierlars: I don't understand. As far as I can tell, the old code did precisely that
```
      {  // "close" stdin, but avoid fd 0 being reused!
        int fd = open("/dev/null", O_RDONLY);
        if (fd >= 0) {
          dup2(fd, 0);
          close(fd);
        }
      }
```
if `usePipes` was false.

lib/Basics/process-utils.cpp
neunhoef: I have added code to handle this.,

neunhoef (approved):
My answers. LGTM.

lib/Basics/process-utils.cpp
```diff
@@ -86,6 +88,13 @@
 #include "Logger/Logger.h"
 #include "Logger/LoggerStream.h"
 
+#ifdef __APPLE__
+// The following hack is required to get access to the environment variables
+// in the same way as on Linux:
+#include <crt_externs.h>
+#define environ (*_NSGetEnviron())
+#endif
```
jsteemann, line 96: Technically not necessary, as we do not support building on macOS anymore.

lib/Basics/process-utils.cpp
```diff
@@ -225,9 +234,141 @@ static bool CreatePipes(int* pipe_server_to_child, int* pipe_child_to_server) {
 /// @brief starts external process
 ////////////////////////////////////////////////////////////////////////////////
 
-static void StartExternalProcess(ExternalProcess* external, bool usePipes,
-                                 std::vector<std::string> const& additionalEnv,
-                                 std::string const& fileForStdErr) {
+#include <spawn.h>
```
jsteemann, line 237: As devel is Linux-only, we could move this include to the top of the file, to where the other includes are.

lib/Basics/process-utils.cpp
neunhoef: Will remove. This is an artifact from the 3.11 forward port.

jsteemann (approved):
LGTM

jsteemann (approved):
LGTM

PR #20972, opened 2024-05-27
ENTRIES for AQL

maierlars:
### Scope & Purpose
`ENTRIES` takes a single object argument and returns a list of key-value pairs, similar to JS `Object.entries()`.
This PR is part of a bigger effort to make iteration over objects in aql more memory efficient and performant.

Currently the only way of iteration over objects is
```
LET keys = ATTRIBUTES(object)
FOR key IN keys
    LET value = object[key]
```

This requires `object` to be copied past the `FOR` loop. Secondly, we have to build the `ATTRIBUTES` array and then do a binary search in the object to look up the value.

This PR alone does not improve the situation by much, however, it allows to write
```
FOR pair in ENTRIES(object)
   let key = pair[0]
   let value = pair[1]
```
Which already eliminates the copy of `object` if it is not used after the `FOR` loop.

A follow up PR will introduce a optimizer rule and a new executor, that takes in an object and produces two registers, containing the keys and values.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Added ENTRIES function, that takes an object and produces a list of key-value
+  pairs.
+
 * Prevent spurious rclone errors in S3.
 
 * Use posix_spawn instead of fork/exec for subprocesses. This solves a
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,7 @@ devel
 -----
 
 * Added ENTRIES function, that takes an object and produces a list of key-value
-  pairs.
+  pairs. The order of pairs is unspecified.
 
 * Prevent spurious rclone errors in S3.
 
--- a/arangod/Aql/Function/ObjectFunctions.cpp
+++ b/arangod/Aql/Function/ObjectFunctions.cpp
@@ -798,7 +798,7 @@ AqlValue functions::Entries(ExpressionContext* expressionContext,
   transaction::BuilderLeaser builder(trx);
   builder->openArray();
 
-  for (auto [key, value] : VPackObjectIterator(objectSlice)) {
+  for (auto [key, value] : VPackObjectIterator(objectSlice, true)) {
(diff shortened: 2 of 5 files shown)


arangod/Aql/Function/ObjectFunctions.cpp
```diff
@@ -774,4 +774,39 @@ AqlValue functions::Zip(ExpressionContext* expressionContext, AstNode const&,
...
+  transaction::Methods* trx = &expressionContext->trx();
+  auto* vopts = &trx->vpackOptions();
+
+  AqlValueMaterializer objectMaterializer(vopts);
+  VPackSlice objectSlice = objectMaterializer.slice(object);
+
+  transaction::BuilderLeaser builder(trx);
+  builder->openArray();
+
+  for (auto [key, value] : VPackObjectIterator(objectSlice)) {
```
jsteemann, line 801: I strongly suggest that we iterate over the attributes of the object in _object build order_ here. This is indicated by passing the `useSequentialIteration = true` flag to the velocypack::ObjectIterator. This allows a linear scan over the object's underlying memory.
Not passing this flag means we will be iterating over the attributes in _index table order_ here. This will mean random accesses to the object entries for each index table entry, which can be a lot less efficient for larger objects or objects with many keys.

```suggestion
  for (auto [key, value] : VPackObjectIterator(objectSlice, true)) {
```

CHANGELOG
```diff
@@ -1,6 +1,9 @@
 devel
 -----
 
+* Added ENTRIES function, that takes an object and produces a list of key-value
+  pairs.
```
jsteemann, line 5: I suggest that we add here that the order in which attributes are iterated is not guaranteed and thus is an implementation detail.

arangod/Aql/Function/ObjectFunctions.cpp
maierlars: Valid point. I will fix that.

jsteemann (approved):
LGTM

PR #20978, opened 2024-05-28
Async AQL Modification

maierlars:
This is just a draft for now, to figure out what lifecycle issues are still present.

### DIFF
--- a/arangod/Aql/SimpleModifier.cpp
+++ b/arangod/Aql/SimpleModifier.cpp
@@ -186,13 +186,6 @@ ExecutionState SimpleModifier<ModifierCompletion, Enable>::transact(
 
   auto result = _completion.transact(trx, _accumulator.closeAndGetContents());
 
-  // we are currently waiting here for the `result` future to get
-  // ready before we continue. this makes the AQL modification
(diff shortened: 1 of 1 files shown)


jsteemann (changes requested):
I did the same modification a few weeks ago to check if the `wait()` call would be still relevant. I can't recall if I opened a test PR for this or only made a local modification and tested it locally.
But I do remember that the lifecycle issues were still present, at least when running some specific workloads. This requires a cluster and a replicationFactor > 1, so that `transaction::Methods::replicateOperations(...)` chains some lambdas to the future. These lambdas may be executed only after the query / transaction has finished, which will lead to undefined behavior.
The lambdas in `replicationOperations` currently do not ensure that the transaction is still valid when the lambda gets executed later. This has to be ensured by external means. In AQL this is achieved by calling `wait()` on the returned future.
Not doing this would mean that a coordinator could send the DB server an `/_api/aql/finish` call to terminate the query, which would destroy the query object (and thus the current transaction object) on the DB server. Additionally, the DB server's query registry could detect an inactivity timeout on the query and remove it, which would also delete the underlying transaction object and thus making the execution of the lambda UB.

PR #20986, opened 2024-06-03
move network retry operations to a separate thread

jsteemann:
### Scope & Purpose

Move network retry requests to a separate thread, so that retries do not depend on the availability/state of the scheduler.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20985
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/Network/NetworkFeature.cpp
```diff
@@ -23,30 +23,232 @@
...
+          if (!_retryRequests.empty()) {
+            _nextRetryTime = _retryRequests.top().retryTime;
+          } else {
+            _nextRetryTime = now + kDefaultSleepTime;
+          }
+
+          guard.unlock();
+
+          if (isStopping()) {
+            break;
```
neunhoef, line 188: If we hit this, we would still go into the `wait_until`, so this could be a large delay, if the next retry is far in the future. Maybe we want to `return` here directly?

arangod/Network/NetworkFeature.cpp
```diff
@@ -609,54 +816,13 @@ void NetworkFeature::finishRequest(network::ConnectionPool const& pool,
...
-      _retryRequests.emplace(std::move(req), std::move(item));
-      // the request successfully made it into _retryRequests.
-      // now abandon the automatic cancelation.
-      cancelGuard.cancel();
-    } catch (...) {
-    }
-  }
-  // if we get here and haven't canceled cancelGuard,
-  // the request will be automatically canceled
+  TRI_ASSERT(_retryThread);
```
neunhoef, line 823: What if a retryable response comes back after the `NetworkFeature` has destroyed the retry thread? Could this then be a shutdown crash if we try to retry here? Should we check `_retryThread` here again (and maybe even use an atomic to check?)?

neunhoef (commented):
LGTM

arangod/Network/NetworkFeature.cpp
```diff
@@ -23,30 +23,213 @@
...
-#include "Network/ConnectionPool.h"
-#include "Network/Methods.h"
-#include "ProgramOptions/ProgramOptions.h"
-#include "ProgramOptions/Section.h"
 #include "Metrics/CounterBuilder.h"
 #include "Metrics/FixScale.h"
 #include "Metrics/GaugeBuilder.h"
 #include "Metrics/HistogramBuilder.h"
 #include "Metrics/MetricsFeature.h"
+#include "Network/ConnectionPool.h"
```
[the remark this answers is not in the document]
jsteemann: Done

arangod/Network/NetworkFeature.cpp
jsteemann: `return`ing is not sufficient, because we also want to call `cancelAll()` once. Adjusted the code now to break the loop properly without waiting afterwards.

arangod/Network/NetworkFeature.cpp
jsteemann: Slightly adjusted it.

kvahed (approved):
LGTM

arangod/Network/NetworkFeature.cpp
```diff
@@ -23,30 +23,236 @@
...
 namespace {
+// executes network request retry operations in a separate thread,
+// so that they do not have be executed via the scheduler.
+// the reason to execute them from a dedicated thread is that a
+// dedicated thread will always have capacity to execute, whereas
+// pushing retry operations to the scheduler needs to use the correct
+// priority lanes and also could be blocked by scheduler threads
+// not pulling any more new tasks due to overload/overwhelm.
+class RetryThread : public ServerThread<ArangodServer> {
+  static constexpr auto kDefaultSleepTime = std::chrono::milliseconds(10'000);
```
goedderz, line 67:
```suggestion
  static constexpr auto kDefaultSleepTime = std::chrono::seconds(10);
```

arangod/Network/NetworkFeature.cpp
goedderz: Should we call `req->cancel()` here? Otherwise this request will just be dropped, won't it?

neunhoef: OK, the `break` is good, it leads to termination of the outer loop, too and thus to `cancelAll`, agreed. But Tobias comment is also good, because `req` is already off the retry queue...

neunhoef (approved):
LGTM

arangod/Network/NetworkFeature.cpp
jsteemann: Done

goedderz (approved):

PR #20990, opened 2024-06-03
reduce the possibility of stampedes

jsteemann:
### Scope & Purpose

Reduce the possibility of stampedes during collection count cache refilling.
This is not a perfect solution, but it should greatly reduce the opportunity for request stampedes to happen when updating the cached value for collection counts.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/20991
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/Cluster/ClusterInfo.cpp
+++ b/arangod/Cluster/ClusterInfo.cpp
@@ -83,6 +83,7 @@
 #include "Sharding/ShardingInfo.h"
 #include "StorageEngine/EngineSelectorFeature.h"
 #include "StorageEngine/PhysicalCollection.h"
+#include "Transaction/CountCache.h"
 #include "Utils/Events.h"
 #include "VocBase/LogicalCollection.h"
 #include "VocBase/LogicalView.h"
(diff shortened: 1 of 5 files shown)


kvahed (approved):
LGTM

goedderz (approved):
I'm OK with this, but I do want to note that I find the resulting behavior unnecessarily inconsistent; as far as I understand, a request that bumps the cache expiry not only triggers an update of the cache, but also waits for its completion before returning. While other requests may get arbitrarily outdated cache values, but return fast.

In principle, I'd favor a solution where either _all_ requests wait for the cache to be updated (in a way that guarantees a bound on how old the cached value is), or _none_ of the requests wait (therefore "guaranteeing" a fast response).

jsteemann:
> I'm OK with this, but I do want to note that I find the resulting behavior unnecessarily inconsistent; as far as I understand, a request that bumps the cache expiry not only triggers an update of the cache, but also waits for its completion before returning. While other requests may get arbitrarily outdated cache values, but return fast.
> 
> In principle, I'd favor a solution where either _all_ requests wait for the cache to be updated (in a way that guarantees a bound on how old the cached value is), or _none_ of the requests wait (therefore "guaranteeing" a fast response).

You are right, the behavior before and also after this PR is not yet optimal. What this PR tried to fix is that multiple concurrent threads can come to the conclusion that the cache value is outdated, and all ask the DB servers for the current collection count values. This would be a cache stampede, and I think that this PR will successfully avoid that. At least in case a cached value for the collection was seen at any point before - if no cache value was ever set, then they will all ask the DB server for the count value.

We could improve on this current state (after this PR) by fetching the counts from the DB server asynchronously (e.g. via the scheduler) and still return the outdated cache value for the thread that triggers fetching the update. This would be an improvement, and it would only leave the issue of what value to return for the initial request to the count cache, when no cached value was ever fetched before.

PR #17644, opened 2022-11-23
Fix issue #16451: Don't stop an outer subquery early due to inner LIMIT being fulfilled

goedderz:
### Scope & Purpose

Fix [issue #16451](https://github.com/arangodb/arangodb/issues/16451): In certain situations, a LIMIT inside a subquery could actually reduce the results of the containing (sub)query.

The code skipping shadow rows due to outer LIMITs calculates the depth to skip. However, at SubqueryStart nodes, this was off-by-one, as it was calculated in relation to its call, but applied to its input. Note that SubqueryStart nodes increase the depth of all incoming shadow rows by one. This has to be taken into account when using the depth of the call stack to skip input rows.

- [X] :hankey: Bugfix

### Checklist

- [X] Tests
  - [X] **Regression tests**
  - [X] C++ **Unit tests**
  - [X] **integration tests**
- [x] :book: CHANGELOG entry made
- [ ] Backports
  - [ ] Backport for 3.10: **TODO**
  - [ ] Backport for 3.9: **TODO**
  - [ ] Backport for 3.8: **TODO**

#### Related Information

This relates to the following earlier bugfix: https://github.com/arangodb/arangodb/pull/13308

- [X] GitHub issue / Jira ticket: https://github.com/arangodb/arangodb/issues/16451

PR #20995, opened 2024-06-04
[3.11] [BTS-1813] Adjust SubqueryEnd depth offset

goedderz:
Backport of https://github.com/arangodb/arangodb/pull/20992

### DIFF
--- a/arangod/Aql/SkipResult.cpp
+++ b/arangod/Aql/SkipResult.cpp
@@ -40,13 +40,18 @@ auto SkipResult::didSkip(size_t skipped) -> void {
   _skipped.back() += skipped;
 }
 
-auto SkipResult::didSkipSubquery(size_t skipped, size_t depth) -> void {
+template<int depthOffset>
+requires(depthOffset == 0 || depthOffset == -1)  //
+    auto SkipResult::didSkipSubquery(size_t skipped, size_t depth) -> void {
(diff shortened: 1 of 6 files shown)


arangod/Aql/SkipResult.cpp
```diff
@@ -40,14 +40,19 @@ auto SkipResult::didSkip(size_t skipped) -> void {
   _skipped.back() += skipped;
 }
 
-auto SkipResult::didSkipSubquery(size_t skipped, size_t depth) -> void {
+template<int depthOffset>
+requires(depthOffset == 0 || depthOffset == -1)  //
```
jsteemann, line 44: let's see if all compilers used for 3.11 support this yet...

jsteemann (approved):
code LGTM, but wouldn't be surprised if some of the compilers used for 3.11 run into compile issues on the relatively new syntax. Fingers crossed!

arangod/Aql/SkipResult.cpp
goedderz: I guess so, we've had e.g. this in for a while: https://github.com/arangodb/arangodb/pull/17644/files#diff-d516d61017cef8d3f4ba883997c93b18e838008f49bd03add75d5ce0caa44cf4

mpoeter (approved):

PR #21002, opened 2024-06-04
Optimize cache performance

mpoeter:
### Scope & Purpose

Protect cache table pointer with ReadWriteLock instead of using atomic load. Atomically loading a shared_ptr acquires a mutex even though we don't really need exclusive access. To avoid contention on that mutext we instead use a ReadWriteLock.


lib/Basics/ReadWriteSpinLock.cpp
```diff
@@ -145,14 +145,17 @@ bool ReadWriteSpinLock::lockWrite(std::size_t maxAttempts) noexcept {
...
+  if ((state & ~::ReaderMask) == 0) {
+    // if no one holds the write lock (and no writers are queued), we simply go
+    // ahead and optimistically increment the reader count
+    auto old = _state.fetch_add(::ReaderIncrement, std::memory_order_acquire);
+    // if there are still no writers, we got the lock! \o/
+    if ((old & ~::ReaderMask) == 0) {
       return true;
     }
+    // otherwise we have to revert our change
+    _state.fetch_sub(::ReaderIncrement, std::memory_order_relaxed);
```
jsteemann, line 158: If we get here, we will have (temporarily) incremented the lock state value by one reader, so other observers of the lock state could see that the lock is used by this reader. But we are going to revert the lock state modification in the next line.
This will have implications on other lock methods, e.g. `tryLockWrite()`, `isLocked()` and `isLockedRead()` which may now behave differently because of the spurious reader state modification. For example, is `isLockedRead()` can now return true, although the lock is actually not read locked.
I don't see a concrete problem with this, we should double-check this.

jsteemann (approved):
LGTM

PR #21004, opened 2024-06-05
avoid dropping of followers in some case

jsteemann:
### Scope & Purpose

Backport of https://github.com/arangodb/arangodb/pull/21006

Avoid dropping of followers in the following scenario:
- leader has in-sync follower(s) for a given shard
- resignLeadership is called for current leader
- resignLeadership is reverted, so that old leader is now again the leader
- write to the shard

Previously this led to dropping the followers when the write operation was executed, because the leader's session id changed when taking over shard leadership again. The new session id was propagated to all followers, but it wasn't picked up because the leader server id did not change.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature
- [ ] :fire: Performance improvement
- [ ] :hammer: Refactoring/simplification

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] :book: CHANGELOG entry made
- [ ] :books: documentation written (release notes, API changes, ...)
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: this PR
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] GitHub issue / Jira ticket:
- [ ] Design document:

PR #21006, opened 2024-06-05
avoid dropping of followers in some case

jsteemann:
### Scope & Purpose

Avoid dropping of followers in the following scenario:
- leader has in-sync follower(s) for a given shard
- resignLeadership is called for current leader
- resignLeadership is reverted, so that old leader is now again the leader
- write to the shard

Previously this led to dropping the followers when the write operation was executed, because the leader's session id changed when taking over shard leadership again. The new session id was propagated to all followers, but it wasn't picked up because the leader server id did not change.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: https://github.com/arangodb/arangodb/pull/21004
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RestHandler/RestReplicationHandler.cpp
+++ b/arangod/RestHandler/RestReplicationHandler.cpp
@@ -2707,6 +2707,11 @@ void RestReplicationHandler::handleCommandSetTheLeader() {
     if (followingTermId.isNumber() && !leaderId.empty() &&
         currentLeaderCopy !=
             maintenance::ResignShardLeadership::LeaderNotYetKnownString) {
+      Result res = checkPlanLeaderDirect(col, leaderId);
+      if (res.fail()) {
+        THROW_ARANGO_EXCEPTION(res);
+      }
+
       newLeader =
           absl::StrCat(leaderId, "_", followingTermId.getNumber<uint64_t>());
     }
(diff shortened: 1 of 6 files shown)


neunhoef (approved):
LGTM

arangod/RestHandler/RestReplicationHandler.cpp
```diff
@@ -2691,12 +2696,28 @@ void RestReplicationHandler::handleCommandSetTheLeader() {
...
-          StringUtils::itoa(followingTermId.getNumber<uint64_t>()));
+      newLeader =
+          absl::StrCat(leaderId, "_", followingTermId.getNumber<uint64_t>());
     } else {
-      col->followers()->setTheLeader(leaderId);
+      newLeader = leaderId;
     }
+  } else {
+    // no leader change, but maybe the session id for the same leader has
+    // changed
```
goedderz, line 2706: Do we have to do a `checkPlanLeaderDirect(col, leaderId);` in this branch as well, or is it not necessary? The leader might have changed in the plan already.

jsteemann: Probably yes, but I think there is still a potential for a race even with the additional check in place.
Assuming the following sequence of events:
1. new leaderId == current leaderId
2. call to checkPlanLeaderDirect(...), and the leader hasn't changed in the plan
3. leader gets changed
4. we carry out the modification in the FollowerInfo

I think to avoid a race in the new case, we should add a function to FollowerInfo that only changes the session id in case the server id has not changed.

goedderz: I think I am not yet quite sure what the check accomplishes in the original code. Does it prevent a bug, or is it a best-effort check to improve the likelihood of the leadership change succeeding? And do we have to, or want to, accomplish the same here - or is it different?

jsteemann: @goedderz : I have added some more guardrails.

goedderz (approved):

PR #21011, opened 2024-06-05
honor special case "OFF" for ARANGODB_BUILD_DATE cmake var

jsteemann:
### Scope & Purpose

By default, CMake variables default to "OFF". This case was not handled separately for the `ARANGODB_BUILD_DATE` cmake variable. We only compared the value of this variable to the empty string. This PR adjusts the cmake code to also take into account the special value of "OFF", so that the build date for an ArangoDB executable will not be set to "OFF" if the build date is not explicitly set via the `ARANGODB_BUILD_DATE` cmake variable.

- [ ] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: 3.11 part is contained in https://github.com/arangodb/arangodb/pull/21009
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,7 @@ configure_file(
 )
 
 option(ARANGODB_BUILD_DATE "Specific build date set from the outside (leave empty to auto-generate)" "")
-if (ARANGODB_BUILD_DATE STREQUAL "")
+if (ARANGODB_BUILD_DATE STREQUAL "" OR ARANGODB_BUILD_DATE STREQUAL "OFF")
   if (NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/lib/Basics/build-date.h")
     # auto-generate build date
     string(TIMESTAMP ARANGODB_BUILD_DATE "%Y-%m-%d %H:%M:%S")


CMakeLists.txt
```diff
@@ -238,7 +238,7 @@ configure_file(
 )
 
 option(ARANGODB_BUILD_DATE "Specific build date set from the outside (leave empty to auto-generate)" "")
-if (ARANGODB_BUILD_DATE STREQUAL "")
+if (ARANGODB_BUILD_DATE STREQUAL "" OR ARANGODB_BUILD_DATE STREQUAL "OFF")
```
goedderz, line 241: I would assume that
```
if (ARANGODB_BUILD_DATE)
```
works as well, and would also include the other false-y values. From https://cmake.org/cmake/help/latest/command/if.html#variable:
>if(<variable>)
>True if given a variable that is defined to a value that is not a false constant.

and above:

> False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND.

jsteemann: @goedderz : I tried your suggestion, but it did not work for me, whereas the version in my PR works.
When I use `if (ARANGODB_BUILD_DATE)` only, I get a build date of `OFF`.

goedderz (approved):

neunhoef (approved):
LGTM

PR #21018, opened 2024-06-06
extend chaos tests and fix a potential Merkle tree corruption issue

jsteemann:
### Scope & Purpose

Extend chaos tests and fix a potential data corruption in a collection's Merkle tree, in case a write operation in a streaming transaction ran into the transaction's maximum size limit and failed. If the transaction was continued afterwards and more operations were pushed into it, the leftovers of the failed operation were not removed from the collection's in-memory buffers that contained all modifications for the collection's Merkle tree upon commit of that transaction.

Sequence of events needed:

1. streaming transaction
2. write operation in streaming transaction ran into transaction size limit error
3. the write operation error was ignored by the user
4. the transaction was committed anyway

The bugfix in this PR is to call the RocksDBSavepoint::finish() operation before revision ids are added to the collection's in-memory Merkle tree state in RocksDBCollection.cpp. Previously it was the other way around.
The problem is that RocksDBSavepoint::finish() can fail if the transaction runs into the maximum size limit. In this case the in-memory Merkle tree state was not cleaned up before. We now avoid this problem by calling finish() and checking its return value before we add the changes to the in-memory Merkle tree state.

The PR also extends the CI chaos tests to cover more cases:

- add artificial delay during query shutdown via failure point
- add artificial low timeouts for count cache refreshing via failure
- limit number of server threads during chaos and deadlock tests, in order to pronounce any potential deadlocks
- use more concurrent clients in chaos and deadlock tests
- improve inconsistency error messages in chaos and deadlock tests
- fix passing of AQL query options from tests, distinguishing between the OPTIONS AQL keyword and the separate query options parameter
- add fullcount and `LIMIT x, y` queries to the mix
- make deadlock tests stop on errors such as timeout/queue full etc. to actually find deadlocks

The PR also contains an improvement for an already existing assertion, so that it will provide more information if it's triggered again.

The PR also adds an assertion that will be triggered if documents cannot be removed successfully from the collection's Merkle tree.
Additionally, tests are added for a few more Merkle tree cases (insert-then-remove and rollback).

- [x] :hankey: Bugfix
- [x] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [x] Backport for 3.11: https://github.com/arangodb/arangodb/pull/21028
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


neunhoef (approved):
LGTM

arangod/RocksDBEngine/RocksDBCollection.cpp
```diff
@@ -948,15 +948,19 @@ Result RocksDBCollection::truncateWithRemovals(transaction::Methods& trx,
...
   };
 
-  auto iter =
-      mthds->NewIterator(documentBounds.columnFamily(), [&](ReadOptions& ro) {
-        if (!mthds->iteratorMustCheckBounds(ReadOwnWrites::no)) {
+  // TODO: decide if we need readOwnWrites here!
+  auto iter = mthds->NewIterator(
+      documentBounds.columnFamily(),
+      [&, readOwnWrites = false /*state->hasHint(
+              transaction::Hints::Hint::GLOBAL_MANAGED)*/](ReadOptions& ro) {
```
jsteemann, line 955: note: still needs to be decided if we need to enable or revert this change.

neunhoef (approved):
LGTM

PR #21020, opened 2024-06-07
Resolve blockage by detaching a thread in AQL write on leader waiting for replication

neunhoef:
This addresses a blockage which can occur if many threads on a leader
perform AQL writes at the same time, all replicate their operations on
their followers and are all waiting for responses.

Since this is AQL execution, this usually happens on MEDIUM priority.
But due to a (for now unresolvable) bug in the scheduler, it is
sometimes possible that more threads than allowed engage in this
activity. If the number grows to more than the number of HIGH priority
operations which are allowed concurrently, we have a blockage, since 
no thread can make progress any more as no HIGH priority tasks are
executed any more (due to a shortage of free threads).

This is a special bandaid for 3.11. The problem is solved more
thoroughly in 3.12 by making this particular operation truly
asynchronous.

Here, we have to live with a hack: Namely, we detach a thread which
remains for some time in this waiting loop. This allows the scheduler to
spawn another new thread and resolve the blockage by pulling some
response from the HIGH priority queue.

Note that we detach the thread after a random time. This will lead to
the fact that if many threads are blocked, only few will at first detach
and thus resolve the blockage, so that the others can make normal
progress without detaching.

- Detach thread waiting on leader in AQL write for replication.
- CHANGELOG.

### Scope & Purpose

- [*] :hankey: Bugfix

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- This is 3.11 only.

#### Related Information

- [*] Design document: https://github.com/arangodb/documents/blob/master/DesignDocuments/02_PLANNING/GlobalOutagePrevention.md
      This should already help with the problems described in this
      document.

### DIFF
--- a/arangod/Aql/SimpleModifier.cpp
+++ b/arangod/Aql/SimpleModifier.cpp
@@ -203,19 +203,19 @@ ExecutionState SimpleModifier<ModifierCompletion, Enable>::transact(
   // we choose a random timeout for the detaching. But first we spin a
   // while to avoid delays:
   if (!result.isReady()) {
-    auto const spinTime = std::chrono::microseconds(10'000);
+    auto const spinTime = std::chrono::milliseconds(10);
     auto start = std::chrono::steady_clock::now();
     while (!result.isReady() &&
            std::chrono::steady_clock::now() - start < spinTime) {
       basics::cpu_relax();
     }
     if (!result.isReady()) {
-      auto const detachTime = std::chrono::microseconds(
-          RandomGenerator::interval(uint32_t(100)) * 100'000);
+      auto const detachTime = std::chrono::milliseconds(
+          1000 + RandomGenerator::interval(uint32_t(100)) * 100);
       auto start = std::chrono::steady_clock::now();
       while (!result.isReady() &&
              std::chrono::steady_clock::now() - start < detachTime) {
-        std::this_thread::sleep_for(std::chrono::microseconds(1'000));
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
       }
       if (!result.isReady()) {
         LOG_TOPIC("afe32", INFO, Logger::THREADS)
--- a/arangod/Aql/SimpleModifier.cpp
+++ b/arangod/Aql/SimpleModifier.cpp
@@ -35,8 +35,11 @@
 #include "Basics/StringUtils.h"
 #include "Basics/VelocyPackHelper.h"
 #include "Basics/application-exit.h"
+#include "Basics/cpu-relax.h"
 #include "Cluster/ServerState.h"
 #include "Logger/LogMacros.h"
+#include "Random/RandomGenerator.h"
+#include "Scheduler/SchedulerFeature.h"
 #include "VocBase/LogicalCollection.h"
 
 #include <velocypack/Collection.h>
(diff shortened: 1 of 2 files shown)


arangod/Aql/SimpleModifier.cpp
```diff
@@ -192,8 +195,53 @@ ExecutionState SimpleModifier<ModifierCompletion, Enable>::transact(
...
-  result.wait();
 
+  // If we simply wait, it can happen that we get into a blockage in which
+  // all threads wait in the same place here and none can make progress,
+  // since the scheduler is full. This means we must detach the thread
+  // after some time. To avoid that all are detaching at the same time,
+  // we choose a random timeout for the detaching. But first we spin a
+  // while to avoid delays:
+  if (!result.isReady()) {
+    auto const spinTime = std::chrono::microseconds(10'000);
```
jsteemann, line 206: I would suggest to use milliseconds here to make it more intuitive.
```suggestion
    auto const spinTime = std::chrono::milliseconds(10);
```

arangod/Aql/SimpleModifier.cpp
```diff
@@ -192,8 +195,53 @@ ExecutionState SimpleModifier<ModifierCompletion, Enable>::transact(
...
+    auto start = std::chrono::steady_clock::now();
+    while (!result.isReady() &&
+           std::chrono::steady_clock::now() - start < spinTime) {
+      basics::cpu_relax();
+    }
+    if (!result.isReady()) {
+      auto const detachTime = std::chrono::microseconds(
+          RandomGenerator::interval(uint32_t(100)) * 100'000);
```
jsteemann, line 214: I think that this can lead to a detachTime of `0` in the worst case.
If the rng returns `0`, then the effective wait time is only 10ms (because of the 10ms spinning before).
I think we should at a minimum offset value of at least 1 second here, maybe even more, e.g.
```
      auto const detachTime = std::chrono::microseconds(
          1'000'000 + RandomGenerator::interval(uint32_t(100)) * 100'000);

File: arangod/Aql/SimpleModifier.cpp [hunk 3]
```diff
@@ -192,8 +195,53 @@ ExecutionState SimpleModifier<ModifierCompletion, Enable>::transact(
...
+      auto start = std::chrono::steady_clock::now();
+      while (!result.isReady() &&
+             std::chrono::steady_clock::now() - start < detachTime) {
+        std::this_thread::sleep_for(std::chrono::microseconds(1'000));
```
jsteemann, line 218:
```suggestion
        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```

arangod/Aql/SimpleModifier.cpp
neunhoef: Done.

arangod/Aql/SimpleModifier.cpp
neunhoef: Done with 1s.

jsteemann (approved):
LGTM

kvahed (approved):
Such a simple trick LGTM.

PR #21030, opened 2024-06-11
Fix processing of sanitizer report files

mpoeter:
### Scope & Purpose

Previously sanitizer reports from TSan/ASan etc. were not processed correctly in the CI. As a result, jobs would fail due to reported warnings, but we were missing the details about what went wrong.

This PR should fix these issues. The reported warnings should now be in the testfailures.txt and also published as CI artifacts.


js/client/modules/@arangodb/testutils/instance-manager.js
```diff
@@ -334,6 +334,15 @@ class instanceManager {
       this.launchFinalize(startTime);
       return true;
     } catch (e) {
+      // disable watchdog
+      internal.SetGlobalExecutionDeadlineTo(0.0);
```
dothebart, line 338: should we also set `arangod.serverCrashedLocal`  there? or is this instance only?

js/client/modules/@arangodb/testutils/instance-manager.js
```diff
@@ -334,6 +334,15 @@ class instanceManager {
...
+      this.arangods.forEach(arangod => {
+        try {
+          arangod.terminateInstance();
```
dothebart, line 341: we should have `true` here for forceterminate?

js/client/modules/@arangodb/testutils/instance-manager.js
```diff
@@ -1001,10 +1010,10 @@ class instanceManager {
           }
           return true;
         }
-        if ((arangod.exitStatus !== null) && (arangod.exitStatus.status === 'RUNNING')) {
-          arangod.exitStatus = statusExternal(arangod.pid, false);
+        if (arangod.isRunning()) {
```
dothebart, line 1013: can we rather have 
```
arangod.checkIsRunning() /* -> instance.js */ {
if (this.isRunning()) {
  this.exitStatus = this.status(false)
}
```
here?

dothebart (approved):
I think we should push that functionality to instance.js already. 
else LGTM.

dothebart (approved):

KVS85 (approved):

PR #21035, opened 2024-06-11
  Fix some transaction::Methods lifetime issues

goedderz:
### Scope & Purpose

There are some lifetime issues in the transaction methods, which in some situations can lead to UB during certain document operations on DBServers.

- [X] :hankey: Bugfix

### Checklist


jsteemann (approved):
LGTM

PR #21042, opened 2024-06-12
Use correct memory order to properly synchronize with stopping thread

mpoeter:
### Scope & Purpose

We often use a pattern like `while (thread.isRunning()) { /* wait */ }` during shutdown. `isRunning` returns false once the internal thread state has been set to `STOPPED`. However, previously this load used `memory_order_relaxed` and therefore did not synchronize with the operation that updated the state. Thus, we did not have a happens-before relation and strictly speaking any access to data that was previously accessed by the just stopped thread could potentially result in a data race. 

- [x] :hankey: Bugfix

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

### DIFF
--- a/lib/Basics/Thread.cpp
+++ b/lib/Basics/Thread.cpp
@@ -260,8 +260,10 @@ void Thread::shutdown() {
 
 /// @brief checks if the current thread was asked to stop
 bool Thread::isStopping() const noexcept {
-  auto state = _state.load(std::memory_order_relaxed);
-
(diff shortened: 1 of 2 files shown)


goedderz (approved):

neunhoef (approved):
LGTM

PR #21045, opened 2024-06-12
fix a potential use-after-free error

jsteemann:
### Scope & Purpose

Fix a potential use-after-free error

The `calculationVocbase` object in the TU `DatabaseFeature.cpp` is a `TRI_vocbase_t` instance that can outlive the lifetime of the application server in case `DatabaseFeature::unprepare()` is not called. In this case, the calculationVocbase's dtor is called only by the executable's static finalizers. This must be prevented, as `TRI_vocbase_t` contains other objects, which upon destruction may call into other ApplicationFeatures, e.g. the MetricsFeature transitively via the vocbase's `CursorRepository`.

This PR works around the problem by adding a new `isInternal` flag to the constructor call for the `calculationVocbase`, and only there. Setting this flag tells the constructor to not use the metrics from the MetricsFeature when setting up the vocbase's `CursorRepository` object.
A cleaner solution would be to remove the global `calculationVocbase` object and turn it into an instance variable of the `DatabaseFeature` class. While this is possible, it would also require to change the constructor arguments for `IResearchAqlAnalyzer`, so that it gets the `DatabaseFeature` instance passed to it and does not need to rely on the global variable.
This is a lot more work to do and should be addressed separately. I left code comments in the PR with that suggestion.

No backport for 3.11 needs to be created, as the `CursorRepository` object does not access the MetricsFeature.

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [ ] Tests
  - [ ] **Regression tests**
  - [ ] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports
  - [ ] Backport for 3.12.0: -
  - [ ] Backport for 3.11: -
  - [ ] Backport for 3.10: -

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:


arangod/RestServer/DatabaseFeature.cpp
```diff
@@ -272,7 +272,13 @@ DatabaseFeature::DatabaseFeature(Server& server)
   startsAfter<metrics::MetricsFeature>();
 }
 
-DatabaseFeature::~DatabaseFeature() = default;
+DatabaseFeature::~DatabaseFeature() {
+  // must clean this up, as otherwise we would may destroying the
+  // calculation vocbase until the end of the entire program.
```
goedderz, line 277: This sentence seems a little borked

goedderz (approved):

neunhoef (approved):
LGTM

arangod/RestServer/DatabaseFeature.cpp
jsteemann, line 277:
```suggestion
  // must clean this up, as otherwise we may delay destroying the
  // calculation vocbase until the end of the entire program.
```

jsteemann: Indeed :facepalm:

arangod/VocBase/vocbase.h
```diff
@@ -126,8 +126,16 @@ struct TRI_vocbase_t {
   friend class arangodb::StorageEngine;
 
   explicit TRI_vocbase_t(arangodb::CreateDatabaseInfo&& info);
+
+  // note: isInternal=true is currently only used for the special internal
+  // vocbase object that is used to execute IResearchAqlAnalyzer computations.
+  // all other vocbases do not set it.
```
goedderz, line 132:
```suggestion
  // note: isInternal=true is currently only used for the special internal
  // vocbase object that is used to execute IResearchAqlAnalyzer computations,
  // namely calculationVocbase.
  // all other vocbases do not set it.
```

goedderz (approved):

PR #21050, opened 2024-06-12
Fix bug in LanguageFeature for C.UTF8 locale 

alexbakharew:
### Scope & Purpose

- [x] :hankey: Bugfix
- [ ] :pizza: New feature

### Checklist

- [X] Tests
  - [ ] **Regression tests**
  - [X] C++ **Unit tests**
  - [ ] **integration tests**
  - [ ] **resilience tests**
- [ ] Backports

#### Related Information

- [ ] Docs PR: 
- [ ] Enterprise PR:
- [ ] Design document:

### DIFF
--- a/arangod/RestServer/LanguageCheckFeature.cpp
+++ b/arangod/RestServer/LanguageCheckFeature.cpp
@@ -155,7 +155,7 @@ ErrorCode writeLanguage(arangodb::ArangodServer& server, std::string_view lang,
 std::tuple<std::string, LanguageType> getPreviousLanguage(
     arangodb::ArangodServer& server) {
   std::string prevLanguage;
-  LanguageType prevType;
+  LanguageType prevType = LanguageType::EMPTY;
   [[maybe_unused]] auto res = ::readLanguage(server, prevLanguage, prevType);
 
   return {prevLanguage, prevType};
--- a/arangod/RestServer/LanguageCheckFeature.cpp
+++ b/arangod/RestServer/LanguageCheckFeature.cpp
@@ -182,7 +182,8 @@ void LanguageCheckFeature::start() {
   using namespace arangodb::basics;
 
   auto& feature = server().getFeature<LanguageFeature>();
-  auto [currLang, currLangType] = feature.getLanguage();
+  auto [currLang, currLangType] =
+      feature.getLanguage();  // language that was provided by user
   auto collatorLang = feature.getCollatorLanguage();
 
   auto [prevLang, prevLangType] =
--- a/arangod/RestServer/LanguageCheckFeature.cpp
+++ b/arangod/RestServer/LanguageCheckFeature.cpp
@@ -50,7 +50,6 @@ arangodb::Result readLanguage(arangodb::ArangodServer& server,
   std::string filename = databasePath.subdirectoryName("LANGUAGE");
 
   if (!TRI_ExistsFile(filename.c_str())) {
-    type = LanguageType::EMPTY;
     return TRI_ERROR_FILE_NOT_FOUND;
   }
 
--- a/arangod/RestServer/LanguageCheckFeature.cpp
+++ b/arangod/RestServer/LanguageCheckFeature.cpp
@@ -50,6 +50,7 @@ arangodb::Result readLanguage(arangodb::ArangodServer& server,
   std::string filename = databasePath.subdirectoryName("LANGUAGE");
 
   if (!TRI_ExistsFile(filename.c_str())) {
+    type = LanguageType::EMPTY;
     return TRI_ERROR_FILE_NOT_FOUND;
   }
 
@@ -87,7 +88,7 @@ arangodb::Result readLanguage(arangodb::ArangodServer& server,
   if (LanguageType::DEFAULT == type) {
     LOG_TOPIC("c499e", TRACE, arangodb::Logger::CONFIG)
         << "using default language: " << language;
-  } else {
+  } else if (LanguageType::ICU == type) {
     LOG_TOPIC("c490e", TRACE, arangodb::Logger::CONFIG)
         << "using icu language: " << language;
   }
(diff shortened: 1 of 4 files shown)


arangod/RestServer/LanguageCheckFeature.cpp
```diff
@@ -151,21 +152,13 @@ ErrorCode writeLanguage(arangodb::ArangodServer& server, std::string_view lang,
...
 }
 
-std::tuple<std::string, LanguageType> getOrSetPreviousLanguage(
-    arangodb::ArangodServer& server, std::string_view collatorLang,
-    LanguageType currLangType) {
+std::tuple<std::string, LanguageType> getPreviousLanguage(
+    arangodb::ArangodServer& server) {
   std::string prevLanguage;
   LanguageType prevType;
   arangodb::Result res = ::readLanguage(server, prevLanguage, prevType);
```
jsteemann, line 159: The result of `readLanguage` is now fully ignored. If this is intentional, we should change this line to either
```suggestion
  res = ::readLanguage(server, prevLanguage, prevType);
```
or
```suggestion
  [[maybe_unused]] auto res = ::readLanguage(server, prevLanguage, prevType);
```

arangod/RestServer/LanguageCheckFeature.cpp
```diff
@@ -181,23 +174,33 @@ LanguageCheckFeature::LanguageCheckFeature(Server& server)
...
-      ::getOrSetPreviousLanguage(server(), collatorLang, currLangType);
+  auto& serverRef = server();
+  auto& feature = serverRef.getFeature<LanguageFeature>();
+  auto [currLang, currLangType] =
+      feature.getLanguage();  // language and type that were provided by user
 
   if (LanguageType::INVALID == currLangType) {
     LOG_TOPIC("7ef61", FATAL, arangodb::Logger::CONFIG)
-        << "Specified language '" << collatorLang << " has invalid type";
+        << "Specified language '" << currLang << " has invalid type";
```
jsteemann, line 184:
```suggestion
        << "Specified language '" << currLang << "' has invalid type";
```

lib/Basics/Utf8Helper.cpp
```diff
@@ -128,15 +128,16 @@ bool Utf8Helper::setCollatorLanguage(std::string_view lang,
   status = U_ZERO_ERROR;
 
   if (_coll) {
-    ULocDataLocaleType type = ULOC_ACTUAL_LOCALE;
+    ULocDataLocaleType type = ULOC_VALID_LOCALE;
```
jsteemann, line 131: I _guess_ this change is useful to fix the problem at hand, but I do not overlook if there could be other undesired consequences of this change.
What is the difference between `ULOC_VALID_LOCALE` and `ULOC_ACTUAL_LOCALE`? I don't find the ICU documentation particularly useful.

alexbakharew: I made following assumption:
1. in `getCollatorLanguage` and in `tokenize` methods we use `ULOC_VALID_LOCALE` to obtain the locale and check some properties in it. 
2. In `setCollatorLanguage` we also obtaining collator and check the language inside it. That is why it looks like that we also need to use it here. 

https://unicode-org.github.io/icu/userguide/locale/#valid-locale
```
Valid Locale
A populated locale is one for which ICU has data, or one in which client code has registered a service. If the requested locale is not populated, then ICU will fallback until it reaches a populated locale. The first populated locale it reaches is the valid locale. The valid locale is reachable from the requested locale via zero or more fallback steps.
```

jsteemann: The problem is that I cannot tell from the ICU documentation alone if this change will get us into trouble in case a _different_ locale is used now than before. 
I am under the impression that the locale must not be changed after it was originally selected/configured. Because we store locale-dependent sort orders etc. on disk.

alexbakharew: OK, I agree. Right now I tried to remove this change and seems that it still works. So my small refactoring in logic has fixed this issue. Will launch Jenkins again

arangod/RestServer/LanguageCheckFeature.cpp
```diff
@@ -151,21 +152,13 @@ ErrorCode writeLanguage(arangodb::ArangodServer& server, std::string_view lang,
...
 
-std::tuple<std::string, LanguageType> getOrSetPreviousLanguage(
-    arangodb::ArangodServer& server, std::string_view collatorLang,
-    LanguageType currLangType) {
+std::tuple<std::string, LanguageType> getPreviousLanguage(
+    arangodb::ArangodServer& server) {
   std::string prevLanguage;
   LanguageType prevType;
-  arangodb::Result res = ::readLanguage(server, prevLanguage, prevType);
+  [[maybe_unused]] auto res = ::readLanguage(server, prevLanguage, prevType);
```
neunhoef, line 159: There are code paths in `::readLanguage` which do not initialize `prevType`, so I suggest to initialize `prevType` here before calling `::readLanguage`.

neunhoef: We probably want to initialize it with type `EMPTY`.

lib/Basics/Utf8Helper.h
```diff
@@ -44,7 +44,7 @@ class RegexMatcher;
 namespace arangodb {
 namespace basics {
 
-enum class LanguageType { INVALID, DEFAULT, ICU };
+enum class LanguageType { INVALID, DEFAULT, ICU, EMPTY };
```
neunhoef, line 47: I can' t see the change "promised" in the description of the PR about `ULOC_ACTUAL_LOCALE`, is this reverted? Is it needed?

alexbakharew: Yes, it was reverted after discussion here: https://github.com/arangodb/arangodb/pull/21050#discussion_r1639777585

neunhoef (approved):
LGTM

PR #21052, opened 2024-06-13
Make sure we always report crash output

mpoeter:
### Scope & Purpose

Previously some details like sanitizer reports were only printed when `writeXmlReport` was set to true, which was confusing, especially when running sanitizer tests locally.


dothebart (approved):
LGTM

